Issue #6, opened 2015-08-28
Installation failure due to missing files


comentarinformal:
While most of the installation is done automatically by 'vagrant up', there is a file (/vagrant/logs/nginx.log) that is not created, and makes nginx unable to start, making the rest of the installation fail.
When created, the next missing folder is (/data/osu!web/)bootstrap/cache;  once you create said folder, the rest of the installation goes without a problem.

(Of course, you can create then yourself via ssh and then vagrant provision, but to set up PuTTY you have to migrate the private key Vagrant generates in puttygen, then save it somewhere, then set it as auth, which greatly differs from the four-steps installation guide.)

omkelderman:
i was already creating a textfile with all the problems i got while setting things up :P, but this was indeed one of them, and i think there are more cause i cant get shit to work.

Whenever I have shit running i'll provide a list of steps to do do get things running that worked for me

TheHowl:
So I guess this can easily be fixed by adding to the vagrantfile `touch /vagrant/logs/nginx.log; mkdir /data/osu!web/bootstrap/cache`?

On 28 August 2015 19:22:58 CEST, Comentarinformal <notifications@github.com> wrote:
>While most of the installation is done automatically by 'vagrant up',
>there is a file (/vagrant/logs/nginx.log) that is not created, and
>makes nginx unable to start, making the rest of the installation fail.
>When created, the next missing folder is
>(/data/osu!web/)bootstrap/cache;  once you create said folder, the rest
>of the installation goes without a problem.
>
>(Of course, you can create then yourself via ssh and then vagrant
>provision, but to set up PuTTY you have to migrate the private key
>Vagrant generates in puttygen, then save it somewhere, then set it as
>auth, which greatly differs from the four-steps installation guide.)
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/ppy/osu-web/issues/6

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

omkelderman:
@TheHowl creating the `vagrant/logs` dir is enough, nginx can create the file itself

nanaya:
Sorry, I missed some files when importing the repository. Please check if there's any problem left.

PR #7, opened 2015-08-28
Should fix installation fails (#6)

comentarinformal:


omkelderman:
isn't this fixed by 852c0fa?

comentarinformal:
Yeah, it is

Issue #4, opened 2015-08-28
No login method in mobile view


peppy:
As reported on public slack, there's no way to login if you are viewing the mobile view of the page.

![](http://puu.sh/jpFWq/0fff7bd06b.png)

peppy:
test

PR #11, opened 2015-08-29
Login method for mobile users

reesebiz:
Added div and class `login-nav` to show the username/avatar and also allows for logging in. The login modal is already responsive so nothing needed to be changed there. For issue #4 

**Guest:**
Guest

**Logged in:**
Logged

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,4 @@ sftp-settings.json
 /resources/assets/js/messages.js
 
 composer-installer
+prepros.cfg
--- a/.gitignore
+++ b/.gitignore
@@ -78,4 +78,3 @@ sftp-settings.json
 /resources/assets/js/messages.js
 
 composer-installer
-prepros.cfg
--- a/resources/assets/less/layout/nav.less
+++ b/resources/assets/less/layout/nav.less
@@ -17,6 +17,7 @@
 *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
 *
 */
+
 @nav-xs-height: 50px;
 // top padding + menu + page title
 @nav-sm-height: @grid-gutter-width/2 + @nav-menu-height + @nav-page-title-height;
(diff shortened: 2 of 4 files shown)


nanaya:
there isn't enough space in there on portrait mode :(

peppy:
To be honest we could probably remove the page title (it feels less important) to make room...

nanaya:
or the mobile header could use an actual design. 

Also the view can't be reused as is because it is used for login box positioning (on desktop) and contains element with id

reesebiz:
I think it looks better without the page title too. I'll work on redesigning the mobile nav a bit too.

reesebiz:
Ok in these commits I removed the page title, rearranged the mobile buttons, and put the logo in the middle. Here's what it looks like:

arflyte:
Putting the logo in the centre would break the alignment. http://puu.sh/jUJ44/6f48e59d2a.png (it's not even aligned currently but it will be polished in the future lol). Even if the logo is fitted within bar, it'll look messy, and the logo will look out of place.

peppy:
After further discussion, @arflyte said the logo should be at the left (completely inside the bar), and the menu button should be moved to the right (to the right of the user login piece).

Feel free to give this a shot.

reesebiz:
Newest commit looks like this:

.gitignore
```diff
@@ -76,5 +76,6 @@ sftp-settings.json
 /bootstrap/cache/
 
 /resources/assets/js/messages.js
-
-composer-installer
+
+composer-installer
+prepros.cfg
```
nanaya: you probably should check [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

Various other editor-specific files in here should also be removed as well but at least we shouldn't add a new one.

reesebiz:
I'll use the global gitignore instead thanks

resources/assets/less/layout/nav.less
```diff
@@ -24,121 +25,127 @@
...
+    height: @logo-side;
+    // general padding is aligned to inner circle instead of outline circle
+    margin-top: -10px;
+    .at2x('/images/layout/osu-logo.png', @logo-side, @logo-side);
 }
-
 #nav-links {
-	padding: 0px @grid-gutter-width/4;
-	flex: 0 1 100%;
+    padding: 0px @grid-gutter-width/4;
```
peppy: you seem to have (unintentionally?) reformatted tabs to spaces here.

arflyte:
I got this, I'll take care of it as it requires more work than simple touch ups.

peppy:
Didn't you say the above was fine yesterday? If you're going to change your mind then please actually mention it somewhere.

arflyte:
mobile-header
mobile-header-expanded
screen shot 2015-09-01 at 3 46 21 pm

Here's the new design for mobile header. The logo is changed from the full colour version to 2 colour version.

By default, user avatar appears next to the hamburger. Clicking the avatar opens the user profile.

Header expands on clicking the hamburger. Hamburger icon changed to active state. User avatar fades away and reappears in it's new section. Sign in/out button underneath the hamburger.

Navigation links are all in white. When expanding a link, other links are greyed out.

peppy:
@reesebiz just following up on whether you have intentions on implementing the updated design provided above!

reesebiz:
@peppy I'm still planning on working on this i've just been busy lately. Sorry for the delay!

PR #17, opened 2015-08-31
Cover Hover Previews

MasterBugPatch:
When a user hovers over a preview of a cover, their cover will turn into
it until their mouse leaves the preview.

### DIFF
--- a/resources/assets/js/jsx/profile_page.jsx
+++ b/resources/assets/js/jsx/profile_page.jsx
@@ -53,8 +53,6 @@
 					dataType: 'json'
 			})
 			.done(function(userData) {
-				this.props.user.cover.id = userData.data.cover.id;
-				this.props.user.cover.url = "/images/headers/profile-covers/c" + userData.data.cover.id + ".jpg";
 				this.updateData(null, userData.data);
 			}.bind(this));
 		},
@@ -65,7 +63,7 @@
 		},
 		
 		previewCover: function(_e, p) {//handles hover event
-			var obj = this.props.user;
+			var obj = this.state.user;
 			obj.cover.url = "/images/headers/profile-covers/c" + p + ".jpg";
 			this.updateData(null, obj);
 		},
@@ -92,7 +90,7 @@
 		unlistenAll: function() {
 			$(document).off('profile:cover:select');
 			$(document).off('profile:updated');
-			$(document).off('profile:previewcover');
+			$(document).off('profile:cover:preview');
 			$(document).off('profile:cover:upload:start');
 			$(document).off('profile:cover:upload:complete');
 			$(document).off('profile:page:update');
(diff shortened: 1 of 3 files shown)


peppy:
This project is using UK/Australian spelling by default, so please remove the spelling commit (it shouldn't be grouped together with other changes anyway).

resources/assets/js/jsx/profile_page.jsx
```diff
@@ -53,16 +53,23 @@
 					dataType: 'json'
 			})
 			.done(function(userData) {
+				this.props.user.cover.id = userData.data.cover.id;
```
nanaya: this and the following line shouldn't be necessary. Additionally, props may never be modified.

resources/assets/js/jsx/profile_page.jsx
```diff
@@ -53,16 +53,23 @@
...
 
 		changeMode: function(e) {
 			e.preventDefault();
-
 			this.setState({ mode: e.target.getAttribute('data-mode') });
 		},
-
+		
+		previewCover: function(_e, p) {//handles hover event
+			var obj = this.props.user;
```
nanaya: props.user only used for initial state. Use state.user instead.

resources/assets/js/jsx/profile_page/header.jsx
```diff
@@ -25,6 +25,15 @@
 			$(document).trigger('profile:cover:select', this.props.name);
 		},
 
+		onMouseEnter: function() {
+			$(document).trigger('profile:previewcover', this.props.name);
+		},
+		
+		onMouseLeave: function() {
+			if(this.props.selectedName != this.props.name){
+			$(document).trigger('profile:previewcover', this.props.selectedName);
```
nanaya: this doesn't return custom profile image back.

resources/assets/js/jsx/profile_page/header.jsx
```diff
@@ -25,6 +25,15 @@
 			$(document).trigger('profile:cover:select', this.props.name);
 		},
 
+		onMouseEnter: function() {
+			$(document).trigger('profile:previewcover', this.props.name);
+		},
+		
+		onMouseLeave: function() {
+			if(this.props.selectedName != this.props.name){
```
nanaya: never use loose comparison (`!=`, `==`) in javascript.

resources/assets/js/jsx/profile_page.jsx
```diff
@@ -94,6 +102,7 @@
 			this.unlistenAll();
 			$(document).on('profile:cover:select', this.changeCoverDefault);
 			$(document).on('profile:updated', this.updateData);
+			$(document).on('profile:previewcover', this.previewCover);
```
nanaya: should be `profile:cover:preview` to follow other events naming pattern.

MasterBugPatch:
Alright, I'll have all of these fixed by tonight. Thanks for the feedback ^.^

TheHowl:
Some of the spelling fixes are actually right though.

http://dictionary.cambridge.org/dictionary/english/repository (search for respository, it won't find it)

http://dictionary.cambridge.org/dictionary/english/guarantee

resources/assets/js/jsx/profile_page.jsx
MasterBugPatch: Yes but for some reason their was an error where when a user selects a cover, then re-enters a different box the selected cover will go back to its previous one. This solved that for me.

nanaya: that's probably because you use props.user instead of state.user below.

PR #19, opened 2015-08-31
Autorotate and strip exif.

### LINKED ISSUE #3: Consider stripping EXIF metadata from cover image uploads
Probably worth considering just so users doing accidentally their location.

credit where due:

nanaya:
The best method ( ﾟ◡ﾟ)

`jhead` is required (otherwise nothing will happen).

Fixes #3.

### DIFF
--- a/app/Libraries/ProfileCover.php
+++ b/app/Libraries/ProfileCover.php
@@ -145,6 +145,11 @@ class ProfileCover
 			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


peppy:
I'd probably do the filesize check before the `jhead` call (so people can't theoretically upload a 100mb jpeg and increase server load from exec calls)

nanaya:
it's after upload limit is checked.

peppy:
oops, i was looking at the second filesize check.

Issue #20, opened 2015-08-31
Sample Data


reesebiz:
It would be really convenient for there sample data for people working on the site. When people clone the repo it should available from the get-go. Preferably already on the site too, and not just in the DB.

Looking for:
- Beatmaps
- Beatmap Sets
- Users
  - User Stats
- Forums
  - Forum Posts
- Wiki
- Tournaments
- Store
  - Store Items
- Changelog

peppy:
We need to setup proper [seeding](http://culttt.com/2013/12/16/seeding-laravel-4-database/). If anyone is interested in looking into it, I'm happy to offer some bounty for a good implementation (amount depends on quality and completeness)

reesebiz:
I'll look into it, but no promises.

peppy:
**[Bounty](https://goo.gl/nFdoyI) offered**: 20

Only offered for complete implementation as described in https://github.com/ppy/osu-web/pull/26#issuecomment-136953592, and documentation added to the readme.

ameliaikeda:
Feel free to assign this to me; I'm going to implement [Model Factories](http://laravel.com/docs/5.1/testing#model-factories) for everything so people have have a full set of data from the start.

reesebiz:
@ameliaikeda I'm working on the pull request that I have open now and it's almost done. I'm using model factories for it too.

Hughp135:
As someone wanting to help out on this project, some sample data would be really useful to start out with. Even if it's just a DB export with some sample users/beatmaps/score data that would be extremely helpful.

Has anyone finished implementing model factories / the seeder yet? I'd be willing to make these if I could get some sample data. @ameliaikeda  @reesebiz

peppy:
I'll go out on a limb and say it hasn't been done yet.

I'll see what I can do with some data export. I'll presume you just need a few rows from each table to understand the format?

Hughp135:
Yeah that would be ideal. Thanks.

ameliaikeda:
I got a new job a while back and haven't been able to do much of anything
(deadline is in just over a month for me). This pretty much just needs a
small db dump with a user or two, a forum, a few topics and some posts in
it (and a sample beat map) and people can jump in way faster.

peppy:
Increasing bounty to 300.

Hughp135:
Currently working on this. I'm grabbing some sample data using the osu API atm, since I still don't have an export of peppys database :'(

peppy:
[...]

PR #26, opened 2015-09-02
Table seeding for sample data

reesebiz:
Here's my approach to seeding sample data for people mentioned in issue #20. In the database/seeds/ folder there will be a seeder class file for each table in the database. The main file is `DatabaseSeeder.php`. 

The file uses Faker to generate rows in the DB (so far I have it set to generate 5 rows each table). I set the Faker generator to run from a specific seed so that everyone gets the exact same sample data. You will also see that columns like user_id are identical in different tables which is useful in the long run.

Run `hhvm artisan db:seed` to seed your database.

So far I only added 3 seed classes for these tables:
- phpbb_users (Barebone, simply too many columns I don't know about)
- osu_beatmaps
- osu_beatmapsets

I can add sample data for the rest of the tables mentioned in #20, but first I would like to fix 1 problem:

I need help finding a way to pass variables into these seed files from a different file. You'll see for example that I have `$userid` set in both the `UserTableSeeder.php` and `BeatmapTableSeeder.php`. The variables are also arranged in a specific order so that when the Faker generator outputs a number it uses the same one for each file. If I could just store all these variables somewhere else it would be really helpful. I could just do php include but I wanted to know if there's a more suitable way for Laravel.


database/seeds/BeatmapTableSeeder.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+		Model::unguard();
+		$faker = Faker\Factory::create();
+		
+        for ($i = 0; $i < 5; $i++) {
+			
+			$faker->seed($i);
+			
+			$username = $faker->userName;
+			$userid = $faker->numberBetween($min = 1, $max = 9999);
```
peppy: i'd probably rely on the auto_increment of the database for primary keys

database/seeds/BeatmapsetTableSeeder.php
```diff
@@ -0,0 +1,55 @@
...
+     * @return void
+     */
+    public function run()
+    {
+		Model::unguard();
+		$faker = Faker\Factory::create();
+		
+		for ($i = 0; $i < 5; $i++) {
+			
+			$faker->seed($i);
```
peppy: i'm *pretty* sure you only need to set the seed once per creation of `Faker`. no need to update it each loop here.

database/seeds/BeatmapTableSeeder.php
reesebiz: Oh yea I forgot to remove that after the commit made to to repo today.

database/seeds/BeatmapTableSeeder.php
```diff
@@ -0,0 +1,60 @@
...
+			$beatmapsetid = $faker->numberBetween($min = 1, $max = 1000000);
```
nanaya: is it some new style assigning variable when calling a function.

nanaya:
Probably should check this as well: (Model Factories)[http://laravel.com/docs/5.1/testing#model-factories)

database/seeds/BeatmapsetTableSeeder.php
reesebiz: Putting the seed inside the loop was intended so that everyone would generate the same sample data. It also allows for me to have the same user_id in multiple tables.

reesebiz:
@nanaya I'll look into that, thanks

ameliaikeda:
This should really be using a Model Factory, mostly so we have the ability to generate any sample data for every single model, reliably for tests (via a Faker seed)

reesebiz:
With the latest commit I pushed it uses the Model Factories created `database/factories/ModelFactory.php`

Some info:
* Before I fill out the other tables I just want to make sure this format is fine so far 
* The user_id is purposely generated by Faker so that I can use the same user_id in multiple tables.
  * As long as you're not running `db:seed` hundreds of times it rare to get a conflict. You won't need to run it more than once because i'll eventually set it to generate more than 5 later.
* In `ModelFactory.php` is where all the magic happens. For each table there is a factory describing how each row is to be generated, and also very flexible.
* DatabaseSeeder.php calls each factory set up and creates the rows. 
 * Specific columns that are used in multiple tables (such as user_id and username) will share the same value with each iteration of the seeding.

ameliaikeda:
The purpose of model seeders is for testing, as well – we should be able to define a full model with all relations and just call `make` on it, and have usable data.

Each model should be self-contained, and the ID for each one doesn't matter, as you can take advantage of model relationships via calling the relation as a method instead of a property (`->sets()` vs `->sets`, etc)

ameliaikeda:
The build passes now (see #69), so if you fancy updating this, go ahead. There's a few issues, but we *really* need a way to generate random but usable data soon, so we can start testing properly. I can get this done pretty quickly if you like, but try and keep on top of it otherwise.

reesebiz:
Thanks for your feedback I implemented them all. I'm having trouble running db:seed though. I got the error `[BadMethodCallException] Call to undefined method Illuminate\Database\Query\Builder::add()`. Any idea how to fix that? Looks like it's coming from the line that says `$set->beatmaps()->add($maps);`

TheHowl:
Perhaps it's create() and not add()? Just suggesting.

On 12 September 2015 01:00:05 CEST, reesebiz <notifications@github.com> wrote:
>Thanks for your feedback I implemented them all. I'm having trouble
>running db:seed though. I got the error `[BadMethodCallException] Call
>to undefined method Illuminate\Database\Query\Builder::add()`. Any idea
>how to fix that? Looks like it's coming from the line that says
>`$set->beatmaps()->add($maps);`
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/ppy/osu-web/pull/26#issuecomment-139682526

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

reesebiz:
@TheHowl I tried create() and that probably wasn't it either. Gave me an even more complicated error.

ameliaikeda:
Yeah, I kinda forgot how to associate a one-to-many without saving. Try `$set->beatmaps->add()` without the parens to directly attach the beatmaps

ameliaikeda:
Alternatively, you can do the inverse of the relationship, and do

```php
$beatmaps->each(function (Beatmap $beatmap) use ($set) {
    $beatmap->set()->associate($set);
});
```

This part always trips me up, heh.

ameliaikeda:
Also, if you fancy actually hitting the database before creating stuff, you can use `$set->beatmaps->saveMany()` (not `create`), but the idea is to be able to create everything without touching the database. 
Also, push your WIP stuff so that people can actually fix it and comment on it, and so people can run it. That's what a pull request is for. You can always rebase and force-push to your branch if you want to tidy up mistakes.

reesebiz:
For the inverse of the relationship code snippet: Where is that $beatmaps variable coming from?

reesebiz:
I'm going to have to close my pull request i've been a lot more busy in the past few weeks so i'll let ameliaikeda take over.

PR #34, opened 2015-09-03
Update all files to use PSR-2

ameliaikeda:
Also I changed my name in the composer file. This is a pretty huge update; merge conflicts on existing PRs ahoy.


nanaya:
it doesn't fix blade templates, does it? :(

ameliaikeda:
It should avoid them, because tabs vs spaces in those files can cause visual changes (like replacing tab sequences with `&nbsp;`)

nanaya:
tabs to spaces is okay, I think. Not sure how it can go from tabs to html entity though.

ameliaikeda:
Tabs don't appear as a space in a browser, while a space does (I meant to say *similar to `&nbsp;`*). It's a small edge-case, but it's probably why the styleci guys chose not to run on those files.

nanaya:
pretty sure spaces and tabs are squashed (`    ` behaves the same as ` `) in html.

nanaya:
I'm not sure about `align_double_arrow` fixer. It looks nice but adding key longer than existing keys means realigning entire thing.

nanaya:
enable `unalign_double_arrow` as well, I guess.

ameliaikeda:
Will do~

nanaya:
and then squash them all \o/

peppy:
\o:

Issue #29, opened 2015-09-03
Bullet points look wrong in some instances


peppy:
Likely bbcode parsing failure: http://jizz.ppy.sh/forum/t/259747?start=3508242#forum-post-3508242

![](http://puu.sh/jXZEu/6d7ae40601.png)

**[Bounty](https://goo.gl/nFdoyI) offered**: 4

tewki:
For anyone who wanna fix as I don't have my normal environment, that caused by HTMLPurifier's strict parsing because of the tags that is in it (any tag) as it interprets it as only having the *ol* tag with no *li* tag surrounding it which is going against the strict parsing,
Thus it pushes the end stack to the beginning of that tag (This case *em*) and wraps the content with *li*.

So it's either to try to make HTMLPurifier ignore it or surround the content if has none with a *li* or the bbcode needed for it.

kcaze:
The exact issue we're seeing is http://stackoverflow.com/questions/29433364/htmlpurifier-prepending-extra-list-item-to-all-lists.

There are extra `<br/>`'s being added to the list because we're appending a `<br/>` to all `\n` characters. The easiest fix for this is to append `<p></p>` instead of `<br/>` elements. Here's a screenshot where `<p></p>` is inserted instead of `<br/>`:
http://up.ppy.sh/files/screenshot2015-09-03at8.13.44pm.png

peppy:
Another case:

![](http://puu.sh/k43Zv/d0ebe764b5.png)

http://jizz.ppy.sh/forum/t/364888?start=4490864#forum-post-4490864

VeilStar:
Posting this as it might be useful.

If you open the "List of Confirmed Threads by Category" in this thread https://jizz.ppy.sh/forum/t/338103. It's a spaghetti mess while it's fine on the old forum. https://osu.ppy.sh/forum/t/338103. There is nothing wrong with the way the bbcode is used in that post as far as I'm aware.

Also leaving these screenshots here as examples of when these issues happen.

![01](https://cloud.githubusercontent.com/assets/14024130/10602578/7f2b7fc2-7719-11e5-8634-c3c6d4625804.jpg)

![02](https://cloud.githubusercontent.com/assets/14024130/10602580/7f2dd6d2-7719-11e5-8993-f5a3a943694f.jpg)

![03](https://cloud.githubusercontent.com/assets/14024130/10602579/7f2d22e6-7719-11e5-8635-20efa8a94f77.jpg)

[...]

PR #38, opened 2015-09-04
Fix BBCode parsing to insert '<p></p>' instead of '<br />'

kcaze:
Fix for ppy/osu-web#29.

### DIFF
--- a/app/Libraries/BBCodeFromDB.php
+++ b/app/Libraries/BBCodeFromDB.php
@@ -251,7 +251,7 @@ class BBCodeFromDB
         $text = $this->parseYoutube($text);
         $text = $this->parseProfile($text);
 
-        $text = preg_replace('/\n/', "\n<br />", $text);
+        $text = preg_replace('/\n/', "\n<p></p>", $text);
         $text = CleanHTML::purify($text);
 
         return "<div class='bbcode'>{$text}</div>";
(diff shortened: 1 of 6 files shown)


peppy:
Able to rebase this on the latest master to fix the CI build error?

kcaze:
I think I am based on the latest master?

nanaya:
create test case for what you're fixing and update the other test cases

nanaya:
and seems like travis is ignoring the addons settings as well :( will look into it later

kcaze:
Updated the BBCode test cases to reflect my changes. And yeah, the other pull requests are also having the same Travis build issues.

nanaya:
Not sure having an empty `<p>` a good idea.

nanaya:
also, where's the new test case for the problem being fixed?

Tewki:
Just a question, why not use a real bbcode engine/library like well jBBCode instead of purifying?

Also kcaze, empty dom elements is not a good idea

nanaya:
because phpbb transforms input bbcode into its custom format (suffixing tags with uid `[b]` is stored as `[b:$uid]`).

Tewki:
Yea but that can be accounted for in jBBCode + other libraries.

nanaya:
feel free to try. It will need to at least pass the test without modification at `.db.txt` files and few modifications to `.html`.

Tewki:
I can do it in a few hours when I get back from work if no one else has, also later on wouldn't it be good to drop those uid suffixes since peppy is pretty much leaving phpbb fully.

nanaya:
yeah, but unfortunately there are currently quite a lot of missing functionalities in new forum.

kcaze:
Yeah, you're right that using `<p></p>` is probably not a good idea. Go ahead and try jBBCode since I'm going to go to sleep soon.

PR #41, opened 2015-09-04
Mysql Sucks.

nanaya:
This is test part 2.


nanaya:
so a pull request from local branch (same user) sets the addons (mariadb 10.0) correctly but not when from another user.

kcaze:
Looks like it's still failing? I'm also not seeing the logs anymore, but I'm guessing that's a Travis issue: https://www.traviscistatus.com/

peppy:
Closing for now :-1:

PR #40, opened 2015-09-04
O hai

nanaya:
Sorry, I lied.

PR #42, opened 2015-09-04
Travis CI test

kcaze:


kcaze:
Oh, I guess I can't actually test and see whether or not this fixes the Travis issue. I'm wondering if using MariaDB 10.1 fixes the issue since it looks like that's the version that was whitelisted in https://github.com/travis-ci/travis-core/pull/469/files

nanaya:
it does work in #40. Only when from different user seems to be broken.

PR #44, opened 2015-09-04
Add name header to store listings

comentarinformal:


peppy:
I have a feeling this was omitted on purpose in the design. @arflyte?

peppy:
Can you provide a screenshot showing how this changes the page?

peppy:
Talking through this with @arflyte and @nanaya, this is actually intended so we can display different text in these places.

comentarinformal:
Hmh, alright then

PR #47, opened 2015-09-04
CSS fixes

nanaya:
originally part of profile page updates branch but figured might as well put this out ahead since I'm going to update forum buttons/links/whatever first.


nanaya:
I noticed some styles are still broken (mainly forum area).

Issue #45, opened 2015-09-04
User card/menu design


peppy:
Trialling some larger bounties. This one involves implementing a design concept fully.

![](http://puu.sh/jZttw/0d4f5faa35.png)

This is the user card and menu, which will appear when you click your avatar/username at the top-right of any page.

Requirements:
- Implement design with interaction logic (should appear and dim page when avatar is clicked).
- No need to actually connect up the data source; we need to do some further work for this.
- Follow existing styling naming standards.
- Reuse existing styles where possible.

**[Bounty](https://goo.gl/nFdoyI) offered**: 60

As per bounty guidelines, it is recommended that you claim this issue if you are serious about it.

Ask away for further clarifications.

PSD: http://puu.sh/jZtBI/2a69c21449.psd

ThePooN:
Interesting. I may try to do it if I find enough time.

claudiodekker:
Sign me up for this one, I don't care about the bounty.

peppy:
On a semi-related note: I understand many people do not want financial compensation. That's why we are offering unique rewards (flare for your profiles etc.) which are only available by accumulating bounty. If I was you I'd start saving :).

claudiodekker:
Alright, fair enough. Count me in, then.

PieInOblivion:
This can be done in very simple JS, I'll see what I can throw together in my spare moments. (I understand you want it completed, not a half finished one)

PieInOblivion:
[Re-Adding Comment]
I'll give it a shot. But question, when you say "No need to actually connect up the data source", why?
"we need to do some further work for this.", it would just be so simple to 'Get <a> text inside div 'nav-user-bar'' then paste that at the end of link "http://jizz.ppy.sh/u/" but JUST load the index.html and take the information from that. I don't see the need for another data source.

I recognize that my view is of no importance what so ever, I'm just babbling on wondering why, and I find it comforting to understand these sorts of things, especially coming from someone working on this project. So, if I may ask, why a different data source?

peppy:
You might want to check the pull request above, where @claudiodekker has already submitted a completed implementation for this!

peppy:
Increasing bounty to 100 due to the final complexity of the implementation and awarding to @claudiodekker. Thanks for your contributions! We still need to hook up the implementation before pushing out to the release branch, but that should happen this week.

PR #50, opened 2015-09-05
User menu

claudiodekker:
This should (hopefully) mark #45 as completed.


nekodex:
Found an issue - if you repeatedly click on your avatar to show/hide the dialog too fast, the page progressively gets dimmer and dimmer.

claudiodekker:
Well spotted. Turns out that's an issue already in existence long before this feature was added, namely with the login modal. (Look at the currently live site, which doesn't include this commit)

To be fair, I don't think this issue is *that* important, as it doesn't block user-experience, but I'll still take a look at it, as it should also resolve the issue on the guest-side of things.

If possible, I'll see if I can create a separate bugtracker-issue for this & claim it, although I'm not sure if I've got the rights to do so.

nanaya:
- ctb (and other play mode) logo is available as font `<i class="fa osu fa-ctb-o"></i>`
- country flags are in `images/flags/`
- check this regarding css: #53
- apparently having the modal button in front of overlay breaks the overlay ;_;
- try using localization when possible

claudiodekker:
@nanaya Alright, thanks. I'll get onto it in a bit. 

Is there any quick(er) way for me to reach you regarding quick reviews/feedback on things, or should I just be a patient man? ;)

nanaya:
you can reach me in slack. How fast I will response is a bit of mystery though.

claudiodekker:
Fixed most issues mentioned, and did some (on slack) discussed stuff..

- Replaced gamemode icon (png) with the font version.
- Now including country flags automatically (NOTE: I'm using the path "/images/flags/{{ Auth::user()->country_acronym }}.png", let me know if this is correct or not)
- Reworked the ENTIRE dropdown modal for both login and when logged in, to be based upon the BEM concept.
- Now using localization

As discussed with @nanaya on Slack, this entire has been reworked to be based upon the BEM concept:
This also means that there is now a BEM folder, and a bem-index.less file, in which each BEM-block is being included by it's own .less file.
This can be changed in the future, but for now I suppose this is okay. It's pretty much a playtest for now, doesn't really matter as it's easily moveable.

nanaya:
sorry for late response,

- there are merge conflicts
- files should have end of lines
- less files will have two spaces indentations from now on (following common style)

resources/assets/less/bem/badges.less
```diff
@@ -0,0 +1,48 @@
...
+*/
+
+.badges {
+
+	// Elements
+	&__badge {
+		text-align: center;
+		display: block;
+
+		&--small {
```
nanaya: just fold into `badges__badge`?

resources/assets/less/bem/badges.less
```diff
@@ -0,0 +1,48 @@
...
+			background-size: 38px;
```
nanaya: `contain` will work here

resources/assets/less/bem/badges.less
```diff
@@ -0,0 +1,48 @@
...
+			height: 38px;
+			width: 38px;
+			font-size: 16px;
+			font-weight: bold;
+			color: #fff;
+			padding-top: 2px;
+		}
+
+		&--level {
+			.at2x("/images/badges/levelbadge.png");
```
nanaya: `.at2x` will override background-size at retina.

resources/assets/less/bem/bootstrap/modal-body.less
```diff
@@ -0,0 +1,79 @@
...
+*    osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+*    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+*    See the GNU Affero General Public License for more details.
+*
+*    You should have received a copy of the GNU Affero General Public License
+*    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+*
+*/
+
+.modal-body {
```
nanaya: probably better use a different class name since it's not obvious it's overriding bootstrap styling.

resources/assets/less/bem/bootstrap/modal-body.less
```diff
@@ -0,0 +1,79 @@
...
+	background-color: #FFF;
+	padding: 10px;
+	border-radius: 0 0 5px 5px;
+
+	// Elements
+	&__compartment {
+		display: flex;
+		height: 140px;
```
nanaya: I don't think fixed height should be used

resources/assets/less/bem/bootstrap/modal-body.less
```diff
@@ -0,0 +1,79 @@
...
+		padding: 10px;
```
nanaya: `@grid-gutter-width/4`?

resources/assets/less/bem/bootstrap/modal-header.less
```diff
@@ -0,0 +1,81 @@
...
+	// Elements
+	&__badges {
+		display: flex;
+		position: absolute;
+		top: -20px;
+		left: 10px;
+	}
+
+	&__dimmer {
+		background: -moz-linear-gradient(top, rgba(0,0,0,0.5) 0%, rgba(0,0,0,0.75) 100%); /* FF3.6+ */
```
nanaya: no need for old browsers compatibility

resources/assets/less/bem/bootstrap/modal-header.less
```diff
@@ -0,0 +1,81 @@
...
+		color: #FFF;
+
+		position: absolute;
+		margin: @grid-gutter-width/4;
+		bottom: 0px;
+		left: 0px;
+	}
+
+	&__userinfo {
+		height: 49px;
```
nanaya: why fixed height

resources/assets/less/bem/bootstrap/modal-header.less
```diff
@@ -0,0 +1,81 @@
...
+	&--login {
+		.at2x("/images/headers/login-box.jpg", 430px, 60px);
+		height: 60px;
+	} 
+
+	&--authenticated {
+		align-items: flex-end;
+		.at2x("/images/headers/profile-covers/c3.jpg", 430px, 70px);
+		background-size: 430px auto;
+		background-position: center center;
```
nanaya: apparently single value means same value for x and y

resources/assets/less/bem/bootstrap/modal.less
```diff
@@ -0,0 +1,50 @@
...
+	&__dialog {
+		width: 430px;
+		color: @text-color;
+	}
+
+	// Modifiers
+}
+
+// Javascript-based hooks/styling
+.js-user-dropdown-modal {
```
nanaya: this probably shouldn't be here

resources/assets/less/bem/quick-info.less
```diff
@@ -0,0 +1,60 @@
...
+	&__statistics {
+		display: flex;
+		flex-flow: column nowrap;
+		font-size: 12px;
+		font-weight: 300;
+		margin: 0;
+		padding: 0;
+		width: 100%;
+
+		> li {
```
nanaya: `> *` probably better here (no increase in specificity)

resources/assets/less/bem/quick-info.less
```diff
@@ -0,0 +1,60 @@
...
+			line-height: 16px;
+			display: flex;
+			justify-content: space-between;
+
+			.text-right {
+				font-weight: 600;
+				text-align: right;
```
nanaya: pretty sure not needed

resources/assets/less/bem/quick-info.less
```diff
@@ -0,0 +1,60 @@
...
+		margin: 0;
+		padding: 0;
+		width: 100%;
+
+		> li {
+			line-height: 16px;
+			display: flex;
+			justify-content: space-between;
+
+			.text-right {
```
nanaya: `&__statistics--value`?

resources/assets/less/bem/quick-info.less
```diff
@@ -0,0 +1,60 @@
...
+*
+*/
+
+.quick-info {
+	
+	// Elements
+	&__level {
+		font-style: normal;
+		color: #555;
+		font-family: @font-family-sans-serif;
```
nanaya: ?

resources/views/objects/user-dropdown.blade.php
```diff
@@ -16,50 +16,90 @@
...
+							<a href="#/teams/ppy" class="userinfo-small__team" style="background-image: url('/temp-issue-35-related/default-team.png');"></a>
+						</div>
+						<div class="modal-header__ranking rankinginfo-small">
+							<span class="rankinginfo-small__gamemode"><i class="fa osu fa-ctb-o"></i> #37542</span>
+							<span class="rankinginfo-small__country">Malaysia #37542</span>
+						</div>
+					</div>
+					<div class="modal-body modal-body--compartimentalized">
+						<div class="modal-body__compartment modal-body__compartment--left quick-info">
+							<h1 class="quick-info__level">Level 5</h1>
```
nanaya: this isn't a heading.

resources/assets/less/bem/bootstrap/modal-body.less
```diff
@@ -0,0 +1,79 @@
...
+*    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+*    See the GNU Affero General Public License for more details.
+*
+*    You should have received a copy of the GNU Affero General Public License
+*    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+*
+*/
+
+.modal-body {
+	background-color: #FFF;
```
nanaya: this overrides styling used in forum "search" modal

resources/assets/less/bem/badges.less
claudiodekker: The reason I did this, is to create a R-E-M structure, nested as such.

resources/assets/less/bem/bootstrap/modal-body.less
claudiodekker: Actually, it is. Notice the file's path name.
Besides, normal REM-blocks use a different structure (.modal__body)

resources/assets/less/bem/bootstrap/modal-header.less
claudiodekker: No kittens were killed while generating this code.

resources/assets/less/bem/bootstrap/modal-header.less
claudiodekker: Is that in the spec / does it work cross browser?

resources/assets/less/bem/bootstrap/modal.less
claudiodekker: Hm, you might be right, however I can't seem to think of a more suitable place.

It's related to the modal as it [already was] (https://github.com/ppy/osu-web/blob/master/resources/assets/less/login-modal.less), so I think this should be okay-ish?

resources/assets/less/bem/quick-info.less
claudiodekker: ¿

That's used to pick the Exo v2 w/ fallback or whatever the font is, instead of typing it out fully.

resources/assets/less/bem/quick-info.less
claudiodekker: lol, correct. #bootstrap

resources/assets/less/bem/bootstrap/modal-header.less
nanaya: apparently [the actual behavior](https://developer.mozilla.org/en-US/docs/Web/CSS/position_value) is the other value is always 50% (x or y if keyword, always y if value)

resources/assets/less/bem/bootstrap/modal-body.less
claudiodekker: Absolutely right.

resources/assets/less/bem/bootstrap/modal-header.less
nanaya: my eyes kind of harmed +_+ also changing it is going to be difficult.

And please use `fade(#abc, x%)` instead of rgba.

resources/assets/less/bem/quick-info.less
nanaya: isn't it the default already?

resources/assets/less/bem/bootstrap/modal.less
nanaya: I think we should expect everything here is from same block otherwise it'll get fun real fast.

resources/assets/less/bem/quick-info.less
claudiodekker: It wasn't for this H1, so it's actually overriding something. But I suppose this'll be changed considering *that* will be changed.

resources/assets/less/bem/badges.less
claudiodekker: Ah, I see. Right. Guess we'll have to '!important' on the --small one then.

resources/assets/less/bem/bootstrap/modal-body.less
claudiodekker: Ah, right, didn't look at this one properly..

resources/assets/less/bem/bootstrap/modal.less
claudiodekker: Yeah, probably, although since we're already having a custom prefix for `js-`, which might be styled sometimes (like now), then the only other solution is creating new files, which I don't know is a good idea when you (maybe in the future) port the entire CSS to BEM-style.

Until this is decided, I think it's safe to simply keep it like this..

resources/assets/less/bem/quick-info.less
claudiodekker: Yeah, the idea behind this, is that you can go overboard on custom BEM files. I think this really should be discussed.

The problem here is, is that I could create another item on the element `quick-info-statistics__statistic` called something like `quick-info-statistics-statistic`, with the `_statistics--value` can be `quick-info-statistics-statistic--value`.

See what I'm getting at? This happens at one or two more places (maybe).

resources/views/objects/user-dropdown.blade.php
claudiodekker: Right. Line 51 is? (`<h1 class="userinfo-small__username">{{ Auth::user()->username }}</h1>`)

resources/assets/less/bem/bootstrap/modal.less
nanaya: I think we'll keep at least few files outside of bem anyway (utilities, variables, globals).

resources/assets/less/bem/quick-info.less
nanaya: oh right, should be just `user-statistics__value`. Reusable in profile page as well.

claudiodekker: So, `<li class="quick-info-statistics__statistic user-statistics"><span>{{ trans("users.show.stats.ranked_score") }}</span><span class="user-statistics__value">{{ number_format(29955840356) }}</span></li>` then?

nanaya: there isn't anything in `quick-info-statistics__statistic`, is there?

claudiodekker: `quick-info-statistics` is the block/wrapper element. `quick-info-statistics__statistics` is the bit at line 37.
But no, there isn't anything in `user-statistics`.

nanaya: display and justify content probably belongs to user statistics.

resources/lang/en/users.php
```diff
@@ -63,7 +60,8 @@
             ],
         ],
         'first_members' => 'here since the beginning',
-        'is_supporter' => 'osu!supporter',
+        'is_supporter' => 'osu!Supporter',
```
nanaya: oh right, this and one blow should be in all lowercase. The one in design image is wrong.

resources/assets/less/bem/quick-info.less
claudiodekker: Actually, this is being used fo the following: 
screenshot 2015-09-14 04 51 02

resources/assets/less/bem/bootstrap/modal.less
claudiodekker: Very well, moving it.

Issue #31, opened 2015-09-03
Disallow double-posting on forum


peppy:
- Reply dialog should be visibly disabled with a message in its place telling user to edit last post if they want to add more content.
- A double post should be allowed after n days from original post (let's start with 3, but this is probably best a definable constant).

**[Bounty](https://goo.gl/nFdoyI) offered**: 60

tewki:
Wouldn't a automated merge system be better user side wise?

Talking about merging the new post with the old one (Not making a new post then merging)

peppy:
Automatic merging would probably work too. Not sure which feels better to a user. Most people would probably rethink/rephrase their last post if they were aware it was going to be merged automatically.

Having a link to enter edit mode in the last post from within the "can't-double-post" message is a must, though. Gotta make it accessible.

Repflez:
I suggest going with disabling posting new replies in the thread if the last post was made by user and instead encourage it to edit their post. Editing it would bump the thread obviously.

neico:
Bumping should only happen in interval's though, else people would abuse it to keep their favorite thread on top.

And I'm also for a auto-merge option, which each user can set on their own personal preference, something like:

Option: (checkbox) Enable auto-merge on double-post
Desc: Disabling this will disable the reply dialog as long as a reply would count as a double-post

peppy:
No user-side options, please. We decide on what is best for the user for cases like this.

dvcrn:
Are there any style mockups for disabled textareas and / or disabled buttons? 
I'm poking around with the code right now and ended up with 2 mockups

Just disabling the textarea and writing some text into it + hiding reply button:
![screen shot 2015-09-03 at 1 43 02 pm](https://cloud.githubusercontent.com/assets/688326/9650707/bd552f98-5241-11e5-991c-b4a9dba20bc2.png)

Like above but hiding the avatar as well. Feels more like a infobox by taking all control from the user
![screen shot 2015-09-03 at 1 42 23 pm](https://cloud.githubusercontent.com/assets/688326/9650701/a56d8a1a-5241-11e5-8bd1-3a2879217472.png)

Maybe giving the element that contains the text a smaller height and different color so it looks like a appended 'info/warning box'?

arflyte:
Let's go with something like this 
![screen shot 2015-09-03 at 3 14 57 pm](https://cloud.githubusercontent.com/assets/3632750/9651687/c00ff206-524e-11e5-9a55-04e0122c6e85.png)

[...]

PR #52, opened 2015-09-05
Feature/disallow double posts

dvcrn:
PR for https://github.com/ppy/osu-web/issues/31

Man, this turned out to be a lot more difficult than I expected. While working on it I ran into more and more edge cases. In any way, here's what's done: 

- Style on mobile devices (still needs some edits) 
- Reply links will focus the infobox
- Posting results in the infobox to always show (because the doublepost box is only relevant for the last poster) 
- Clicking the 'edit your previous post' link will always edit the last post of the user

In any way, please let me know if changes are needed.

### DIFF
--- a/app/Http/Controllers/Forum/TopicsController.php
+++ b/app/Http/Controllers/Forum/TopicsController.php
@@ -51,18 +51,6 @@ class TopicsController extends Controller
         ]]);
     }
 
-    private function isDoublePost($post)
-    {
-        $postTime = $post['post_time'];
-        $now = Carbon::now();
-        $diffSinceLastPost = $now->diffInDays($postTime);
-        if (! is_null(Auth::user()) && $post['poster_id'] == Auth::user()->user_id && $diffSinceLastPost < 3) {
-            return true;
-        }
-
-        return false;
-    }
-
     public function create($forum_id)
     {
         $forum = Forum::findOrFail($forum_id);
@@ -171,7 +159,7 @@ class TopicsController extends Controller
         }
 
         $postsPosition = $topic->postsPosition($posts);
-        $isDoublePost = $this->isDoublePost($posts[count($posts) - 1]);
+        $isDoublePost = Topic::isDoublePost($posts[count($posts) - 1]);
 
         Event::fire(new TopicWasViewed($topic, $posts->last(), Auth::user()));
 
@@ -185,7 +173,7 @@ class TopicsController extends Controller
         $topic = Topic::findOrFail($id);
         $posts = Post::where('post_id', $topic->topic_last_post_id)->get();
 
-        if ($this->isDoublePost($posts[count($posts) - 1])) {
+        if (Topic::isDoublePost($posts[count($posts) - 1])) {
             return 'not allowed';
         }
 
--- a/app/Http/Controllers/Forum/TopicsController.php
+++ b/app/Http/Controllers/Forum/TopicsController.php
@@ -51,11 +51,12 @@ class TopicsController extends Controller
         ]]);
     }
 
-    private function isDoublePost($post) {
+    private function isDoublePost($post)
+    {
         $postTime = $post['post_time'];
         $now = Carbon::now();
         $diffSinceLastPost = $now->diffInDays($postTime);
-        if (!is_null(Auth::user()) && $post['poster_id'] == Auth::user()->user_id && $diffSinceLastPost < 3) {
+        if (! is_null(Auth::user()) && $post['poster_id'] == Auth::user()->user_id && $diffSinceLastPost < 3) {
             return true;
         }
 
@@ -182,10 +183,10 @@ class TopicsController extends Controller
     public function reply(HttpRequest $request, $id)
     {
         $topic = Topic::findOrFail($id);
-        $posts = Post::where("post_id", $topic->topic_last_post_id)->get();
+        $posts = Post::where('post_id', $topic->topic_last_post_id)->get();
 
         if ($this->isDoublePost($posts[count($posts) - 1])) {
-            return "not allowed";
+            return 'not allowed';
         }
 
         $this->authorizePost($topic->forum, $topic);
--- a/app/Http/Controllers/Forum/TopicsController.php
+++ b/app/Http/Controllers/Forum/TopicsController.php
@@ -51,6 +51,17 @@ class TopicsController extends Controller
         ]]);
     }
 
+    private function isDoublePost($post) {
+        $postTime = $post['post_time'];
+        $now = Carbon::now();
+        $diffSinceLastPost = $now->diffInDays($postTime);
+        if (!is_null(Auth::user()) && $post['poster_id'] == Auth::user()->user_id && $diffSinceLastPost < 3) {
+            return true;
+        }
+
+        return false;
+    }
+
     public function create($forum_id)
     {
         $forum = Forum::findOrFail($forum_id);
@@ -159,17 +170,23 @@ class TopicsController extends Controller
         }
 
         $postsPosition = $topic->postsPosition($posts);
+        $isDoublePost = $this->isDoublePost($posts[count($posts) - 1]);
 
         Event::fire(new TopicWasViewed($topic, $posts->last(), Auth::user()));
 
         $template = $skipLayout ? '_posts' : 'show';
 
-        return view("forum.topics.{$template}", compact('topic', 'posts', 'postsPosition', 'jumpTo'));
+        return view("forum.topics.{$template}", compact('topic', 'posts', 'postsPosition', 'jumpTo', 'isDoublePost'));
     }
 
     public function reply(HttpRequest $request, $id)
     {
         $topic = Topic::findOrFail($id);
+        $posts = Post::where("post_id", $topic->topic_last_post_id)->get();
+
+        if ($this->isDoublePost($posts[count($posts) - 1])) {
+            return "not allowed";
+        }
 
         $this->authorizePost($topic->forum, $topic);
 
--- a/app/Http/Controllers/Forum/TopicsController.php
+++ b/app/Http/Controllers/Forum/TopicsController.php
@@ -159,7 +159,7 @@ class TopicsController extends Controller
         }
 
         $postsPosition = $topic->postsPosition($posts);
-        $isDoublePost = $topic->isDoublePostBy(Auth::user()->user_id);
+        $isDoublePost = $topic->isDoublePostBy(Auth::user() ? Auth::user()->user_id : null);
 
         Event::fire(new TopicWasViewed($topic, $posts->last(), Auth::user()));
 
--- a/app/Http/Controllers/Forum/TopicsController.php
+++ b/app/Http/Controllers/Forum/TopicsController.php
@@ -159,7 +159,7 @@ class TopicsController extends Controller
         }
 
         $postsPosition = $topic->postsPosition($posts);
-        $isDoublePost = Topic::isDoublePost($posts[count($posts) - 1]);
+        $isDoublePost = $topic->isDoublePostBy(Auth::user()->user_id);
 
         Event::fire(new TopicWasViewed($topic, $posts->last(), Auth::user()));
 
@@ -173,7 +173,7 @@ class TopicsController extends Controller
         $topic = Topic::findOrFail($id);
         $posts = Post::where('post_id', $topic->topic_last_post_id)->get();
 
-        if (Topic::isDoublePost($posts[count($posts) - 1])) {
+        if ($topic->isDoublePostBy(Auth::user()->user_id)) {
             return 'not allowed';
         }
 
--- a/app/Models/Forum/Topic.php
+++ b/app/Models/Forum/Topic.php
@@ -70,18 +70,6 @@ class Topic extends Model
         return $topic;
     }
 
-    public static function isDoublePost($post)
-    {
-        $postTime = $post['post_time'];
-        $now = Carbon::now();
-        $diffSinceLastPost = $now->diffInDays($postTime);
-        if (! is_null(Auth::user()) && $post['poster_id'] == Auth::user()->user_id && $diffSinceLastPost < 3) {
-            return true;
-        }
-
-        return false;
-    }
-
     public function addPost($poster, $body, $notifyReplies)
     {
         DB::transaction(function () use ($poster, $body, $notifyReplies) {
@@ -207,6 +195,22 @@ class Topic extends Model
     {
         return $this->posts()->skip(intval($n) - 1)->first();
     }
+
+    public function isDoublePostBy($userId)
+    {
+        $post = $this->posts()->orderBy('post_id', 'desc')->first();
+        $postTime = $post['post_time'];
+        $now = Carbon::now();
+
+        $diffSinceLastPost = $now->diffInDays($postTime);
+
+        if (! is_null($userId) && $post['poster_id'] == $userId && $diffSinceLastPost < 3) {
+            return true;
+        }
+
+        return false;
+    }
+
 
     public function postPosition($postId)
     {
--- a/app/Models/Forum/Topic.php
+++ b/app/Models/Forum/Topic.php
@@ -24,6 +24,7 @@ namespace App\Models\Forum;
 use Carbon\Carbon;
 use DB;
 use Illuminate\Database\Eloquent\Model;
+use Auth;
 
 class Topic extends Model
 {
@@ -67,6 +68,18 @@ class Topic extends Model
         });
 
         return $topic;
+    }
+
+    public static function isDoublePost($post)
+    {
+        $postTime = $post['post_time'];
+        $now = Carbon::now();
+        $diffSinceLastPost = $now->diffInDays($postTime);
+        if (! is_null(Auth::user()) && $post['poster_id'] == Auth::user()->user_id && $diffSinceLastPost < 3) {
+            return true;
+        }
+
+        return false;
     }
 
     public function addPost($poster, $body, $notifyReplies)
--- a/app/Models/Forum/Topic.php
+++ b/app/Models/Forum/Topic.php
@@ -198,10 +198,10 @@ class Topic extends Model
     public function isDoublePostBy($userId)
     {
         $post = $this->posts()->orderBy('post_id', 'desc')->first();
-        $postTime = $post['post_time'];
-        $diffSinceLastPost = Carbon::now()->diffInDays($postTime);
+        $postTime = $post->post_time;
+        $daysSinceLastPost = Carbon::now()->diffInDays($postTime);
 
-        if (! is_null($userId) && $post['poster_id'] == $userId && $diffSinceLastPost < 3) {
+        if (! is_null($userId) && $post->poster_id == $userId && $daysSinceLastPost < 3) {
             return true;
         }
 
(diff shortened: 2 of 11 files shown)


app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -157,17 +168,23 @@ public function show($id) {
...
 
+		$isDoublePost = $this->isDoublePost($lastPost = $posts[count($posts) - 1]);
 		$postsPosition = $topic->postsPosition($posts);
 
 		Event::fire(new TopicWasViewed($topic, $posts->last(), Auth::user()));
 
 		$template = $skipLayout ? "_posts" : "show";
 
-		return view("forum.topics.{$template}", compact("topic", "posts", "postsPosition", "jumpTo"));
+		return view("forum.topics.{$template}", compact("topic", "posts", "postsPosition", "jumpTo", 'isDoublePost'));
```
peppy: probably should be using double quotes here to match the others on this line.

dvcrn: Oh whoops. Gonna correct that later when I'm back at my PC. Anything else?

ameliaikeda: Simple strings are replaced with single quotes by StyleCI. Click details and apply the patch that StyleCI provides (below) after rebasing against master to automatically fix your files.

dvcrn:
Rebasing against master turned out to be a good chunk more difficult than expected since close to every line has changed in between. I'll replay my commits manually this weekend to fix this one up

dvcrn:
All right, manually replayed all the changes on top of latest master

app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -51,6 +51,18 @@ public function __construct()
         ]]);
     }
 
+    private function isDoublePost($post)
```
nanaya: this should be part of Topic model.

resources/assets/coffee/forum/topic-ajax.coffee
```diff
@@ -42,6 +42,11 @@ $(document).on 'ajax:success', '.delete-post-link', (_event, data) ->
 
 $(document).on 'ajax:success', '.reply-post-link', (e, data) ->
   data += '\n'
+  if window.isDoublePost
+    url = location.href
+    history.replaceState(null, null, url)
+    location.href = "#doublepost-box"
```
nanaya: this won't work if last post isn't yet loaded.

resources/assets/coffee/forum/topic-ajax.coffee
```diff
@@ -42,6 +42,11 @@ $(document).on 'ajax:success', '.delete-post-link', (_event, data) ->
 
 $(document).on 'ajax:success', '.reply-post-link', (e, data) ->
   data += '\n'
+  if window.isDoublePost
```
nanaya: probably should intercept much earlier instead of waiting for server reply. `ajax:before` or `ajax:beforeSend`. [Reference](https://github.com/rails/jquery-ujs/wiki/ajax).

resources/views/forum/topics/_post.blade.php
```diff
@@ -71,7 +71,7 @@ class="row-page forum-post flex-row js-forum-post__shrunk post-viewer"
 
     <div class="post-viewer__actions">
         @if ($options["editLink"] === true)
-            <a href="{{ route("forum.posts.edit", $post) }}" class="post-viewer__action edit-post-link" data-remote="1">
+            <a href="{{ route("forum.posts.edit", $post) }}" class="post-viewer__action edit-post-link" data-remote="1" target="{{ $post->post_id }}">
```
nanaya: `target` does different thing. Use something else like `data-target` or `data-post-id`.

resources/views/forum/topics/_doublepostbox.blade.php
```diff
@@ -0,0 +1,30 @@
...
+
+    osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+    See the GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+--}}
+<?php
+$display = "inherit";
```
nanaya: inherit means inherit parent dom. Should be empty string.

resources/views/forum/topics/_doublepostbox.blade.php
```diff
@@ -0,0 +1,30 @@
...
+
+if ($hide)
+    $display = "none";
+?>
+<div id="{{ $id }}" class="row-page flex-row forum-post-box-warning" style="display:{{ $display }}">
```
nanaya: the check isn't long enough that a simple `$hide ? 'none' : ''` is okay.

resources/assets/less/forum/base.less
```diff
@@ -928,6 +928,27 @@ body.community {
...
+  p {
+    line-height: 22px;
+    margin: 0;
+    padding: @grid-gutter-width/2;
+  }
+  .iconbox {
+    background: @yellow-dark;
+    padding: @grid-gutter-width/2;
+    i {
+      font-size: 30px;
```
nanaya: part of `.iconbox` works.

resources/assets/less/forum/base.less
```diff
@@ -928,6 +928,27 @@ body.community {
...
+  color: white;
+  min-height: 0;
+  padding: 0;
+
+  p {
+    line-height: 22px;
+    margin: 0;
+    padding: @grid-gutter-width/2;
+  }
+  .iconbox {
```
nanaya: make it `&__iconbox` instead to avoid increasing specificity and following bem.

resources/assets/less/forum/base.less
```diff
@@ -928,6 +928,27 @@ body.community {
...
 }
 
+.forum-post-box-warning {
+  &:extend(.forum-post);
+  background: @dark-grey;
+  color: white;
+  min-height: 0;
+  padding: 0;
+
+  p {
```
nanaya: a div probably work better since you don't need to then reset margin.

resources/assets/less/forum/base.less
```diff
@@ -928,6 +928,27 @@ body.community {
...
+    line-height: 22px;
```
nanaya: a `display: flex` on the parent and `align-self: center` here works better, I believe (and guarantees centering).

resources/views/forum/topics/_doublepostbox.blade.php
```diff
@@ -0,0 +1,30 @@
...
+    <div class="iconbox">
+        <i class="fa fa-info-circle"></i>
+    </div>
+    <p>You just posted. Please wait a bit or <a data-remote="1" target="{{ $lastPostId }}" class="edit-post-link" href="{{ route("forum.posts.edit", $lastPostId) }}">edit your previous post</a>.
```
nanaya: i18n?

resources/assets/coffee/forum/topic-ajax.coffee
```diff
@@ -70,17 +75,31 @@ $(document).on 'ajax:success', '#forum-topic-reply-box', (_event, data) ->
...
+    editLastPostLink.attr("target", lastPostId)
+    doublepostBox.show()
+
   else
     osu.navigate $(data).find('.js-post-url').attr('href')
 
 
 $(document).on 'ajax:success', '.edit-post-link', (e, data, status, xhr) ->
+  targetId = $(e.target).attr("target");
+  $postBox = $("[data-post-id='"+targetId+"']")
```
nanaya: use coffeescript's string interpolation. `"[data-post-id='#{targetId}']"`.

resources/assets/coffee/forum/topic-ajax.coffee
```diff
@@ -70,17 +75,31 @@ $(document).on 'ajax:success', '#forum-topic-reply-box', (_event, data) ->
     newPost = window.forum.endPost()
     newPost.classList.remove('js-forum-post__shrunk')
     newPost.scrollIntoView()
+    window.isDoublePost = true
+
+    $('#forum-topic-reply-box').hide()
+
+    doublepostBox = $("#doublepost-box")
+    editLastPostLink = doublepostBox.find("a")
```
nanaya: finding the `edit-post-link` class is better instead of any links.

resources/assets/coffee/forum/topic-ajax.coffee
```diff
@@ -70,17 +75,31 @@ $(document).on 'ajax:success', '#forum-topic-reply-box', (_event, data) ->
...
+    lastPostId = $(".forum-post").last().attr("data-post-id")
```
nanaya: at this point might as well have the double post box target as `lastPost` instead of specifying id. Or a different link type altogether (`edit-last-post` link or something like that which clicks last post's edit link).

resources/views/forum/topics/show.blade.php
```diff
@@ -111,5 +117,8 @@ class="jump-link last-post"
 
     <script data-turbolinks-eval="always">
         window.postJumpTo = {{ $jumpTo }};
+        @if ($isDoublePost)
+            window.isDoublePost = true;
```
nanaya: this should always be set otherwise turbolinks will keep old value.

app/Http/Controllers/Forum/TopicsController.php
dvcrn: I didn't add this block to the Topic model because it contains a `Auth::` check which doesn't really fit in there. Passing `user_id` might work but seems unnecessary. But anyway, I moved it.

resources/assets/coffee/forum/topic-ajax.coffee
dvcrn: 👌

resources/views/forum/topics/_post.blade.php
dvcrn: 👌

resources/views/forum/topics/_doublepostbox.blade.php
dvcrn: 👌

resources/views/forum/topics/_doublepostbox.blade.php
dvcrn: That was mainly for readability

resources/assets/less/forum/base.less
dvcrn: What exactly do you mean by `part of .iconbox works`?

resources/assets/less/forum/base.less
dvcrn: Picked `p` mainly because of semantic reasons

resources/views/forum/topics/_doublepostbox.blade.php
dvcrn: I had it in i18n before but because the text format is `text text <link with more text>` I wasn't sure how to structure it correctly. I could define 2 elements, 1 for the link and 1 for the text surrounding the link but thought in some languages there might be some more text after the link as well. 

If you are ok with the 2 element approach, I'll move it into i18n

resources/assets/coffee/forum/topic-ajax.coffee
dvcrn: 👌

resources/assets/coffee/forum/topic-ajax.coffee
dvcrn: Changed it to `a.edit-post-link` 👌

resources/assets/less/forum/base.less
nanaya: `i` (or more accurately, `.fa`) has `font-size` set to `inherit`.

resources/views/forum/topics/show.blade.php
dvcrn: 👌

app/Models/Forum/Topic.php
```diff
@@ -69,6 +70,18 @@ public static function createNew($forum, $title, $poster, $body, $notifyReplies)
         return $topic;
     }
 
+    public static function isDoublePost($post)
```
nanaya: Why not just always check against last post of itself?

resources/views/forum/topics/_doublepostbox.blade.php
```diff
@@ -0,0 +1,30 @@
...
+$display = "";
+
+if ($hide)
+    $display = "none";
+?>
+<div id="{{ $id }}" class="row-page flex-row forum-post-box-warning" style="display:{{ $display }}">
+    <div class="iconbox">
+        <i class="fa fa-info-circle"></i>
+    </div>
+    <p>You just posted. Please wait a bit or <a data-remote="1" data-target-post-id="{{ $lastPostId }}" class="edit-post-link" href="{{ route("forum.posts.edit", $lastPostId) }}">edit your previous post</a>.
```
nanaya: you can pass the translated link as part of full sentence parameter.

resources/views/forum/topics/_doublepostbox.blade.php
ameliaikeda: You should never be assigning things in a view, ever.

app/Models/Forum/Topic.php
```diff
@@ -195,6 +195,21 @@ public function nthPost($n)
...
+    public function isDoublePostBy($userId)
+    {
+        $post = $this->posts()->orderBy('post_id', 'desc')->first();
+        $postTime = $post['post_time'];
+        $now = Carbon::now();
+
+        $diffSinceLastPost = $now->diffInDays($postTime);
+
+        if (! is_null($userId) && $post['poster_id'] == $userId && $diffSinceLastPost < 3) {
+            return true;
```
nanaya: return ! is_null ....

?

app/Models/Forum/Topic.php
```diff
@@ -195,6 +195,21 @@ public function nthPost($n)
         return $this->posts()->skip(intval($n) - 1)->first();
     }
 
+    public function isDoublePostBy($userId)
+    {
+        $post = $this->posts()->orderBy('post_id', 'desc')->first();
+        $postTime = $post['post_time'];
```
nanaya: attributes can be accessed as property (`$post->post_time`, etc)

app/Models/Forum/Topic.php
```diff
@@ -195,6 +195,21 @@ public function nthPost($n)
...
+        $now = Carbon::now();
+
+        $diffSinceLastPost = $now->diffInDays($postTime);
```
nanaya: `$daysSinceLastPost` makes more sense, I believe

app/Models/Forum/Topic.php
```diff
@@ -195,6 +195,21 @@ public function nthPost($n)
         return $this->posts()->skip(intval($n) - 1)->first();
     }
 
+    public function isDoublePostBy($userId)
+    {
+        $post = $this->posts()->orderBy('post_id', 'desc')->first();
+        $postTime = $post['post_time'];
+        $now = Carbon::now();
```
nanaya: calling it inline isn't too long

app/Models/Forum/Topic.php
dvcrn: Validation in the following order: 
- Check if the passed userId is `null` to rule out errors
- Check if the last poster id is the id that got passed in
- Check if the diff is bigger than 3

Or what is your question?

nanaya: why not immediate return

dvcrn: Style choice. Adding a immediate return would have 3 separate returns inside a very short function. This way it is a tick less performant but all validation happens at the same spot. Either it passes -> return true, or it doesn't -> return false. 

Of course I could have it like 
```
    public function isDoublePostBy($userId)
    {
        if (is_null($userId)) {
            return false;
        }

        $post = $this->posts()->orderBy('post_id', 'desc')->first();

        if ($post['poster_id'] != $userId) {
            return false;
        }

        $postTime = $post['post_time'];
        $now = Carbon::now();
        $diffSinceLastPost = $now->diffInDays($postTime);

        if ($diffSinceLastPost < 3) {
            return true;
        }

        return false;
    }
```

but that is a lot longer and harder to read

ameliaikeda: As an additional note, use property access (`$post->post_time`) instead of array access, as this is a model.

dvcrn:
@nanaya rebased latest develop and did some more code style changes based on your codereview. I had to rebase this blind (no way I can run this on my work machine) so I'll verify later if everything still looks good but based on the code are there any other things you'd like me to change?

app/Models/Forum/Topic.php
```diff
@@ -195,6 +195,19 @@ public function nthPost($n)
         return $this->posts()->skip(intval($n) - 1)->first();
     }
 
+    public function isDoublePostBy($userId)
+    {
+        $post = $this->posts()->orderBy('post_id', 'desc')->first();
+        $postTime = $post->post_time;
+        $daysSinceLastPost = Carbon::now()->diffInDays($postTime);
+
+        if (! is_null($userId) && $post->poster_id == $userId && $daysSinceLastPost < 3) {
```
nanaya: `is_null($userId)` check is too late, it should be the first thing checked and return immediately if null.

Also I think it should always return true if `$userId` is null.

And no need to bother doing `if (condition) { return true; }`, just write it like this:

```
return $post->poster_id === ... && ...;
```

Extra: don't use `==` operator.

resources/assets/coffee/forum/topic-ajax.coffee
```diff
@@ -40,6 +40,14 @@ $(document).on 'ajax:success', '.delete-post-link', (_event, data) ->
       $(document).trigger('osu:page:change')
 
 
+$(document).on 'ajax:before', '.reply-post-link', (e, data) ->
+  if window.isDoublePost
```
nanaya:
```
return true unless window.isDoublePost

url = ...
```

resources/assets/coffee/forum/topic-ajax.coffee
```diff
@@ -40,6 +40,14 @@ $(document).on 'ajax:success', '.delete-post-link', (_event, data) ->
...
+    url = location.href
+    history.replaceState(null, null, url)
+    location.href = "#doublepost-box"
```
nanaya: I mentioned it before, this doesn't work if last posts hasn't been loaded. It'll just scroll to bottom and load more posts.

dvcrn: must have overseen that, hmm. Any suggestions how to load / find the necessary post if it's not loaded yet?

peppy:
This PR seems to have stagnated. @dvcrn interested in seeing it through still? Else I'll close it for now.

dvcrn:
oh - I'll try to give it another shot on the weekend!
If I remember correctly the last problem was the case when the user is double posting but the previous post is not loaded yet (for example when replying to someone). We show the error message with a link to edit the previous post but if that has not been loaded yet, it won't resolve to anything. 

Could we for example show a permalink to the last post by the user in that case or should we load all pages / posts in between until we found that particular one? (It could be 10+ pages away)

dvcrn:
Looks like I have less time for this than I wish I'd have. I'm closing this until I'm a bit more free but please feel free to use any of the provided code if needed. 

In the meantime, @nanaya any thoughts on what I wrote above in how to retrieve / display the correct post?

PR #60, opened 2015-09-06
Wiki Page Design

MasterBugPatch:
Help->Wiki page design.


peppy:
Hey, sorry for the delay on response here. I'd like to see this through and get it merged, but it looks like your remote doesn't resolve any more. Do you still have a copy of your branch you can push to your osu-web clone?

MasterBugPatch:
Hmm I thought since the code was still viewable under "changes" that the merge would still work. Once I get home in around 30 minutes I will be able to commit it again to a new clone :)

peppy:
It shows in changes, but I'm unable to actually check it out locally (as far as I know). Thanks in advance!

MasterBugPatch:
Done :) https://github.com/ppy/osu-web/pull/298

PR #68, opened 2015-09-09
Npm updates

nanaya:
\<insert swear words here\>


package.json
```diff
@@ -1,17 +1,17 @@
...
     "font-awesome": "*",
-    "gulp": "^3.8.11",
-    "jquery": "^2.1.4",
-    "laravel-elixir": "^2.0.0",
-    "layzr.js": "^1.4.3",
-    "lodash": "^3.10.1",
-    "photoswipe": "^4.1.0",
-    "react": "^0.13.3",
-    "retina.js": "^1.1.0"
+    "gulp": "*",
```
GrahamCampbell: wildcards are normally a very bad idea

nanaya: there's shrinkwrap to handle this.

PR #71, opened 2015-09-10
Profile custom selection

nanaya:
It will allow people to re-select their custom cover past their supporter period (but still can't upload a new one).


nanaya:
weird, I didn't get notification this being merged.

PR #80, opened 2015-09-14
Use coffeescript for react

nanaya:
Doesn't seem too bad?


peppy:
will definitely take some getting used to.. but at least it's consistent across the board now.

PR #82, opened 2015-09-16
composer.json modifications

peppy:
As discussed in public slack, the authors section of composer.json is going to get out of hand and should probably be disposed of.


nanaya:
sure thing ┐(･_･)┌

nanaya:
Also, are agpl3+ or agpl3 now?

peppy:
3+ until someone `sed`s every file :pager:

PR #103, opened 2015-10-05
Fix ugly image cover spinner.

### LINKED ISSUE #99: Cover image upload spinner is ugly
User reported.

Should probably use the standard loading animation.

TheHowl:
This fixes #99.


peppy:
This is a pretty half-assed fix. At very least, I'd want to be able to target where the loading screen is displayed (so it was only overlaying the cover image, since the rest of the page is still usable).

A better solution is probably to remove the spinner and replace it with a better overlay graphic or just have the cover dim during upload.

TheHowl:
I agree. Will make it only dim the cover pic as suggested when loading when I'm home.

TheHowl:
(closing as i'm going to develop the mentioned system on another branch)

PR #120, opened 2015-10-08
Fix ugly image cover spinner, v2

TheHowl:
#103 reworked as suggested by peppy.


nanaya:
Fixed by replacing with another spinner.

PR #132, opened 2015-10-14
added explicit bower install

ErunamoJAZZ:
For some reason, when i try to build the assets with gulp, this fail because it don't found some js from bower. So... i think will be better if it is explicit in instructions.


nanaya:
it's supposed to be part of dependencies installation together with npm and composer.

peppy:
You can check it on https://github.com/ppy/osu-web/blob/master/build.sh#L25.

Best figure why it didn't run/failed for you.

peppy:
Closing this for now.

PR #164, opened 2015-10-26
New forum navigation

nanaya:
The font resize and wider post need ~slightly~ more work.


resources/assets/less/radial-progress.less
```diff
@@ -1,128 +0,0 @@
-/**
```
peppy: might be worth keeping this around for future use?

nanaya: I'll probably forget how to use it if the need ever comes anyway.

peppy: fair enough

PR #168, opened 2015-10-26
Async sucks ;_;

nanaya:
Sometimes the setTimeout in `in` function is ran after another, unrelated event.
- mouseenter called, settimeout queued
- mouseleave called
- settimeout event ran <- whoops

Noticed on firefox.


nanaya:
Also, I shall just leave this here https://developer.mozilla.org/en/docs/Web/CSS/pointer-events

PR #177, opened 2015-10-29
Misc fixes

nanaya:
Mostly mobile and safari.

(ノﾟ ヮﾟ)ノ︵ ıɹɐɟɐs


resources/assets/less/bem/bootstrap/modal.less
```diff
@@ -24,7 +24,8 @@
 
   // Elements
   &__dialog {
-    width: 430px;
+    width: calc(100% ~'-' (@spacing * 2));
```
nanaya: best syntax

Issue #109, opened 2015-10-08
Ability to change store orders from admin page


peppy:
Right now I'm making changes manually at a database level when a user requests an address or t-shirt design/size changes. We need the ability to edit orders from the admin page (/store/admin/).

Requirements:
- Ability to edit address.
- Ability to change product to another product of same value (different size/design) using a dropdown.
- Changes should be saved instantly.
- Should only be allowed for orders in a "paid" state. Shipped or Delivered orders should not be changeable.
- Remaining stock should be checked (products without stock shouldn't be shown in the dropdown) and updated on change.

**[Bounty](https://goo.gl/nFdoyI) offered**: 50

echojc:
I'm gonna give this a shot.

@peppy I'd be grateful if you could take a look at the current state of my PR and tell me if I'm on the right track style- and architecture-wise since it's my first time working with such a heavily ORM-based framework!

echojc:
Should editing the address save to the same row or create a new row? I can see use cases for both ("I entered my address wrong" and "send this order to this other address").

peppy:
I ended up doing this myself, so closing the issue.

PR #185, opened 2015-10-30
Ability to change store orders from admin page

echojc:
This is a work in progress for ppy/osu-web#109

- [x] Ability to edit address.
  - [x] Changes should be saved instantly.
- [ ] Ability to change product to another product of same value (different size/design) using a dropdown.
  - [ ] Changes should be saved instantly.
- [ ] Should only be allowed for orders in a "paid" state. Shipped or Delivered orders should not be changeable.
- [ ] Remaining stock should be checked (products without stock shouldn't be shown in the dropdown)
  - [ ] and updated on change.

### DIFF
--- a/app/Http/Controllers/Store/Admin/OrdersController.php
+++ b/app/Http/Controllers/Store/Admin/OrdersController.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace App\Http\Controllers\Store\Admin;
+
+use App\Http\Controllers\Controller;
+use App\Models\Store;
+use Auth;
+use Request;
+
+class OrdersController extends Controller
+{
+    protected $section = 'storeAdmin';
+
+    public function __construct()
+    {
+        $this->middleware('auth', ['only' => [
+            'index',
+            'show',
+            'update',
+        ]]);
+
+        return parent::__construct();
+    }
+
+    public function index()
+    {
+        return $this->show();
+    }
+
+    public function show($id = null)
+    {
+        if (!Auth::user()->isAdmin()) {
+            abort(403);
+        }
+
+        $orders = Store\Order::with('user', 'address', 'address.country', 'items.product');
+
+        if ($id) {
+            $orders->where('orders.order_id', $id);
+        } else {
+            $orders->where('orders.status', 'paid');
+        }
+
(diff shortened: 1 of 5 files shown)


nanaya:
This controller is awesome™.

/s

Anyway, I believe it should at least look like this, route-wise

```
GET /store/admin/orders/{orderId}
PUT /store/admin/orders/{orderId}
PUT /store/admin/orders/{orderId}/items/{orderItemId}
```

And uses separate, namespaced controller

(now I look again, the work to change just this section probably isn't too bad)

echojc:
With the PUT on order item, does it matter that `orderItemId` can already be used to directly reference the right row without needing the `orderId`? Would you check that they match (e.g., `where order_id = $a and id = $b`) before updating?

When you say use a separate namespaced controller, do you mean a new controller class with new routes in routes.php?

nanaya:
> With the PUT on order item, does it matter that orderItemId can already be used to directly reference the right row without needing the orderId? Would you check that they match (e.g., where order_id = $a and id = $b) before updating?

Good point. I'm not sure if it's better to have the check one from OrderItem instead of done through Order, like, adding a method to update OrderItem from Order model - together the order state check etc.

> When you say use a separate namespaced controller, do you mean a new controller class with new routes in routes.php?

Yeah. Current `StoreController` scope is getting a bit too... big (it's currently doing p.much everything related to store).

nanaya:
I just noticed there's already such method in Order model (`updateItem`). Comes with product stock updating logic as well.

echojc:
Thanks, I'll take a look.

nanaya:
Though looking what it does, it's probably not usable for this problem :confused:

echojc:
@nanaya can you take a look at my store controller split before I carry on and add more stuff to it?

nanaya:
I'm thinking something along this line http://laravel.com/docs/5.1/controllers#restful-resource-controllers with the controllers namespaced in `App\Http\Controllers\Store\Admin`

echojc:
Third time lucky?

nanaya:
looks about right (controller part).

peppy:
I'm continuing work on this for what it's worth (can't wait any longer sorry!).

peppy:
Closing this in favour of https://github.com/ppy/osu-web/pull/264. More to come.

PR #187, opened 2015-11-02
Add support for Docker as a provider for Vagrant.

echojc:
Useful for development on Linux.

I'm not sure how the default provider is determined. The docs say that VirtualBox is always the default, but on my work machine, Vagrant automatically chooses Docker even if I don't explicitly pass `--provider docker` to `vagrant up`.

It could be that Vagrant detects that VirtualBox isn't installed and so tries the next provider?

Will need to test for compatibility before merge.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -25,8 +25,6 @@ cd /data/osu\!web/
 mkdir -p "public/uploads"
 chmod 777 "public/uploads"
 
-./bin/db_setup
-
 # replace mysql config vars to be reachable from vm host
 sed -i 's/bind-address/#bind-address/' /etc/mysql/my.cnf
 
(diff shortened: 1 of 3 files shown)


peppy:
docker should be better in every way, anyway.

echojc:
Note: Docker requires `--privileged` in order to run the `mount` command in the provisioning script in the Vagrantfile. Running a container in privileged mode is somewhat of a red flag. Is that mount necessary? Since node_modules is in the synced folder, shouldn't that cache be preserved between container restarts anyway?

nanaya:
> Since node_modules is in the synced folder, shouldn't that cache be preserved between container restarts anyway?

npm uses quite a few symlinks and virtualbox shared folder hates symlinks

PR #194, opened 2015-11-03
Explicitly import app.less into other less files (fixes linting issues).

peppy:
Without these there are many variable and function declarations that cannot be resolved by a naive less linter.


nanaya:
For some reason it looks like a circular dependency (though probably less keeps track which files it has imported).

peppy:
I've tested and it doesn't cause any issues with compilation, so should be okay.

PR #198, opened 2015-11-03
Use (js) Menu for header menu as well

nanaya:
Less code, more sharing. Maybe.


peppy:
pity the licence headers increase the line count >_>

PR #208, opened 2015-11-07
Allow a custom_class to be defined but not exist.

peppy:
I'm using custom_class in the fulfillment process to determind certain types of items. In the case of the new tournament banners, they don't (yet?) have a custom blade, but they do need a custom class. This allows that to be possible without error to face.

Needs to be merged ASAP.


peppy:
plz someone ;_;

PR #215, opened 2015-11-13
Add ability to search by rank achieved

nekodex:
Now using DB + ElasticSearch to enable searching by rank achieved (supporter required)

Things still remaining:
- Search by rank status (for favourites, my maps, etc)
- Re-fix playmode/difficulty display
- Persist/restore state via URL
- Thumbnails
- \clean and optimize up all the codes/

### DIFF
--- a/app/Models/BeatmapSet.php
+++ b/app/Models/BeatmapSet.php
@@ -22,6 +22,8 @@ namespace App\Models;
 use Request;
 use Es;
 use Illuminate\Database\Eloquent\Model;
+use Auth;
+use DB;
 
 class BeatmapSet extends Model
 {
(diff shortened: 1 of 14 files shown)


app/Models/BeatmapSet.php
```diff
@@ -169,63 +171,69 @@ public function isQualified()
...
+
+        $valid_ranks = ['A', 'B', 'C', 'D', 'S', 'SH', 'X', 'XH'];
+        foreach ($params['rank'] as $rank) {
+            if (!in_array($rank, $valid_ranks)) {
+                unset($params['rank'][$rank]);
+            }
+        }
+
+        if (presence($params['query']) != null) {
+            $params['query'] = preg_replace('/\s\s+/', ' ', $params['query']);
```
nanaya: is it a fancy way to write `\s{2,}`

app/Models/BeatmapSet.php
```diff
@@ -169,63 +171,69 @@ public function isQualified()
...
+        unset($params['sort']);
+
+        $valid_ranks = ['A', 'B', 'C', 'D', 'S', 'SH', 'X', 'XH'];
+        foreach ($params['rank'] as $rank) {
+            if (!in_array($rank, $valid_ranks)) {
+                unset($params['rank'][$rank]);
+            }
+        }
+
+        if (presence($params['query']) != null) {
```
nanaya: not sure why use `presence` since `'' == null` is already true ¯\\_(ツ)_/¯

app/Models/BeatmapSet.php
```diff
@@ -242,8 +250,18 @@ public static function search(array $params = [])
             }
         }
 
-        if (presence($query)) {
-            $matchParams[] = ['query_string' => ['query' => implode(' AND ', explode(' ', $query))]];
+        if (presence($query) != null) {
+            $matchParams[] = ['query_string' => ['query' => $query]];
+        }
+
+        if (!empty($rank)) {
```
nanaya: A reminder that `empty('0')` is true :D (though tracing back to its caller does guarantee that it's array).

nekodex: php ヽ( ﾟ ワ 。)ノ

app/Models/BeatmapSet.php
```diff
@@ -242,8 +250,18 @@ public static function search(array $params = [])
...
+            $klass = presence($mode != null) ? Score\Model::getClass($mode) : Score\Combined::class;
+            $scores = $klass::forUser(Auth::user())->whereIn('rank', $rank)->lists('beatmapset_id');
+            $matchParams[] = ['ids' => ['type' => 'beatmaps', 'values' => $scores]];
+        }
+
+        if (presence($mode) != null && @presence($rank) == null) {
```
nanaya: wait, what's `@` sign doing there

nekodex: holdover from when it used to be $params['rank']

app/Models/BeatmapSet.php
```diff
@@ -169,63 +171,67 @@ public function isQualified()
...
+            if (!in_array($rank, $valid_ranks)) {
+                unset($params['rank'][$rank]);
+            }
+        }
+
+        if ($params['query'] !== null) {
+            $params['query'] = preg_replace('/\s{2,}/', ' ', $params['query']);
+            $params['query'] = trim($params['query']);
+
+            $query_parts = explode(' ', $params['query']);
```
nanaya: apparently up to this part can be done with single `preg_split`.

nekodex: If you're talking about using regex instead of explode, [obligatory joke about using regex](http://www.xkcd.com/1171/). Also [this](http://micro-optimization.com/explode-vs-preg_split.html) and [this](http://stackoverflow.com/questions/27303235/in-php-which-is-faster-preg-split-or-explode) >_>

FWIW I haven't really bothered optimising stuff yet. just getting it all going first~

nanaya: did you forget the preg_replace and trim before this explode <_<

app/Models/BeatmapSet.php
```diff
@@ -251,47 +267,51 @@ public static function search(array $params = [])
...
+            'rank' => [],
+            'page' => 1,
+            'sort' => ['ranked', 'desc'],
+        ];
 
-        $searchParams['index'] = env('ES_INDEX', 'osu');
-        $searchParams['size'] = $max;
-        $searchParams['body']['sort'] = ['approved_date' => ['order' => 'desc']];
-        $searchParams['type'] = 'beatmaps';
+        self::sanitizeSearchParams($params);
```
nanaya: how do you decide when to use `self` and when to use `static` (actual question)

nekodex: `static` looks like some workaround for the dodgy "object orientation" in php and `self` not behaving as you would expect? [see this](http://php.net/manual/en/language.oop5.late-static-bindings.php)

...so I guess use it when you have overridden methods on inherited classes?
dunnolol `¯\_(ツ)_/¯`

app/Models/BeatmapSet.php
nanaya: something like this

    preg_split('/\s+/', '   her!p   dekl=  lksdfjlkj ', PREG_SPLIT_NO_EMPTY);

nanaya: updated with trim because the previous one looked dumb

PR #229, opened 2015-11-19
Blank line code and immediate object parsing

nanaya:
- allows creation of blank line in code block again
- pass json as javascript object immediately instead of telling javascript to parse json text


peppy:
fark off tarvis

PR #234, opened 2015-11-25
Forum user ribbon

nanaya:
@peppy: I don't remember how long each level is :D (and is it even stored in days)


peppy:
looks close enough to what we discussed

nanaya:
I vaguely remember the highest one but I don't remember about the second one

PR #235, opened 2015-11-26
store admin fixes

peppy:


nanaya:
The note probably belongs right above the checkout button?

peppy:
looks better

PR #254, opened 2015-12-02
Speed up animations

nanaya:
Smoother animations \o/

<sub>at cost of readability</sub>


nanaya:
<del>scrolling the page is slow instead :(</del> or maybe not too bad

PR #267, opened 2015-12-09
Reimplement old API on new framework

nekodex:
This is a first step towards the greater goal of reimplementing osu! web features (and eventually bancho) via a shared API.

Should be functionally equivalent to the [existing API](https://osu.ppy.sh/api/) except for a few small things, namely:
 - return types should be more 'correct', i.e. numbers are actually numbers instead of strings
 - api/get_beatmaps return order should be more "correct" (i.e. ordered by approved_date in most cases)

The returned data should otherwise match the existing API, quirks included.

Moving forward, a new version of the API will be created (likely a different endpoint) to be more consistent and "correct".


app/Models/UserStatistics/Model.php
```diff
@@ -91,27 +91,27 @@ public function countryRank()
             ->count() + 1;
     }
 
-    public function __construct($attributes = [])
+    public function __construct($attributes = [], $null = false)
```
nekodex: @nanaya This is kinda dodgy - didn't want to break existing profile stuff. Thoughts on changing the profile stuff to do null->0 for display outside of this constructor?

app/Transformers/API/StatisticsTransformer.php
```diff
@@ -0,0 +1,51 @@
...
+use App\Models;
+use App\Models\UserStatistics;
+use League\Fractal;
+
+class StatisticsTransformer extends Fractal\TransformerAbstract
+{
+    public function transform(Models\UserStatistics\Model $stats = null)
+    {
+        if ($stats === null) {
+            $stats = new UserStatistics\Osu([], null);
```
nanaya: ...do you mean `([], true)` here?

    >>> $a = null ? '0' : '1';
    => "1"

PR #274, opened 2015-12-17
Tooltips again

nanaya:
Now with customisable position.


peppy:
probably don't need this one

PR #280, opened 2015-12-23
Play mode

nanaya:
Unifies play modes constants, switches, etc.

<sub>hopefully nothing breaks</sub>

### DIFF
--- a/app/Models/Beatmap.php
+++ b/app/Models/Beatmap.php
@@ -53,12 +53,6 @@ class Beatmap extends Model
 
     protected $hidden = ['checksum', 'filename', 'orphaned'];
 
-    // playmodes
-    const OSU = 0;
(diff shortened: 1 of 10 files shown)


app/Models/Beatmap.php
```diff
@@ -53,17 +53,37 @@ class Beatmap extends Model
...
-    const TAIKO = 1;
-    const FRUITS = 2;
-    const MANIA = 3;
-
     public function mods()
     {
         return $this->hasMany(Mod::class, 'beatmap_id', 'beatmap_id');
     }
 
+    public static function modes()
```
nanaya: ideally I'd use `const MODES = [...];` but [hhvm hates array constants](https://github.com/facebook/hhvm/issues/4277).

Issue #251, opened 2015-12-01
osu!web to-do list


peppy:
A top-level overview of what is left to be completed for the new osu!web. Still a WIP, which means there's stuff to add which hasn't been added, but also potentially out-of-date items which need to be updated.

While this issue will exist for now, I plan on exploding each of these tasks into individual issues, so we can tackle (and discuss) them in a more optimal manner.
## Live Stream Page
- [x] Highlighted Stream (it should embed a player)
- [x] Stream list
## Change Logs
- [ ] Calendar
- [x] Version History
- [x] Comments
## Home (after login)
- [x] News coming from twitter.
- [x] Top beatmaps
- [ ] User activity (recent plays etc.)
- [ ] Game chat feed
## Home (before login)
- [x] Design
- [x] Implementation
- [ ] Fixes
## Beatmap
- [x] Beatmap Index
  - [x] List view (needs optimisation)
  - [ ] Beatmap Hover Popup
- [x] Beatmap Details
  - [x] Description
  - [x] Success Rate Graph
  - [x] Scoreboard
  - [x] Discussion Thread
  - [x] Spider graph
  - [x] Sound preview
- [x] Beatmap Pack Listing
- [x] Generate new thumbnail images for EVERYWHERE (get list)
## osu!status
- [x] Basic page layout
- [x] Incident history (Active, Recent)
- [x] Online users / scores per second graphs (http://pp.sakisan.be/graph/d3.html)
- [x] Country region map
- [ ] Database implementation (current we only have a design).
## User preferences
- [x] Change avatar
- [x] Change profile details
- [x] Change email or password
- [x] Change signature
## Profile Page
- [x] Supporter Banner
- [x] Recent Activities
- [x] Kudosu
- [x] Top Ranks
  - [x] Best Performance
  - [x] First Place Ranks
- [x] Beatmaps
  - [x] Favourite Beatmaps
  - [x] Ranked & Approved
- [x] Historical
  - [x] Play History
  - [ ] Rank
  - [x] Most Played Beatmaps
- [x] User hover popup
- [ ] Score hover popup
## Sign-up Process
- [ ] New user registration
- [x] Password recovery process
- [ ] Email recovery process
- [ ] Verification gateway (IRC / moderation systems)
- [x] Email verification (2FA)
## Forums
- [x] New Post Design, with Supporter Banner
- [x] Custom Thread Header
- [x] New Post Design (Forum Poster B)
- [x] Width Adjust Button (<1 day)
[...]

PR #284, opened 2015-12-30
Remaining parts for user profile

lunafreya:
Only three parts of the profile to go (according to #251)

- [x] Top Ranks
 - [x] Best Performance
 - [x] First Place Ranks

- [ ] Beatmaps
 - [ ] Favourite Beatmaps
 - [ ] Ranked & Approved

- [ ] Historical
 - [ ] Play History
 - [ ] Rank
 - [ ] Most Played Beatmaps


nanaya:
Adding this slows down the profile page which is why I haven't created pull request...

lunafreya:
https://codeclimate.com/github/ppy/osu-web/compare/nanaya/user-profile-updates

Two complexity issues and seven clarity issues, "app/Models/User.php" reported as "F" as reported by Code Climate.

nanaya:
@peppy I think the top ranks part is done.

PR #289, opened 2016-01-12
Add unique and foreign key constraints to profile customization table

nanaya:
Raw SQL because there's no support for changing column type to `mediumint` :D

Oh and the table needs fixing as well since there seems to be some users having multiple customisations.

```sql
SELECT COUNT(1), user_id FROM user_profile_customizations GROUP BY user_id HAVING COUNT(1) > 1;
```

### DIFF
--- a/database/migrations/2016_01_12_074035_unique_user_customization_user_id.php
+++ b/database/migrations/2016_01_12_074035_unique_user_customization_user_id.php
@@ -1,0 +1,38 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+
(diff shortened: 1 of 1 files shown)


peppy:
Probably a good idea to just use INT, since we're approaching the MEDIUMINT boundary and will need to switch over DB-wide soon.

nanaya:
foreign key requires same datatypes... Also more like bigint?

nanaya:
TIL int is larger than mediumint ಠ_ಠ

peppy:
Hmm yeah. we're still 10 million users off this being an issue, too, so let's just go with this for now.

PR #297, opened 2016-01-21
Disallow restricted users from placing orders at the osu!store.

peppy:
They like to fuck things up (surprise surprise)

### DIFF
--- a/app/Http/Middleware/CheckUserRestricted.php
+++ b/app/Http/Middleware/CheckUserRestricted.php
@@ -54,7 +54,7 @@ class CheckUserRestricted
     public function handle($request, Closure $next)
     {
         if ($this->auth->check() && $this->auth->user()->isRestricted()) {
-            return error_popup('You are not able to perform this action while your account is in a restricted state, sorry!');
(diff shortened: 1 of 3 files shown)


nanaya:
I have something really similar to this, just the difference is mine simply blocks entire non-`GET` requests instead with this additional check: `!$request->isMethod('GET')` (and just add the middleware without specifying the actions).

peppy:
where?

app/Http/Middleware/CheckUserRestricted.php
```diff
@@ -0,0 +1,62 @@
...
+     * Handle an incoming request.
+     *
+     * @param \Illuminate\Http\Request $request
+     * @param \Closure                 $next
+     *
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        if ($this->auth->check() && $this->auth->user()->isRestricted()) {
```
nanaya: here (the `GET` method check).

peppy: potentially it may be used for get requests too though, right?

PR #298, opened 2016-01-22
Wiki Page

MasterBugPatch:
This is the initial wiki page creation


peppy:
Does it still display correctly for you? Seems like some upstream changes may have broken it slightly:

MasterBugPatch:
It looks like everything works in your screenshot except for the header image I have in there, do you have the header image locally?

peppy:
The page is no longer in the correct containers, and spews out to the sides. Here's a better shot:

MasterBugPatch:
Okay, I'll make changes accustomed to it. Should I make the side navigation hidden on mobile or smaller screens?

peppy:
Yeah, that's probably a good direction to go.

MasterBugPatch:
Got it, thanks for the help :).

TPGPL:
@peppy need any help with transporting articles to new.ppy.sh?

flashwave:
@TPGPL pretty sure it's just going to use the mediawiki database, like how the forum uses the phpbb database

nanaya:
Design and stuff have been changed and then implemented.

PR #299, opened 2016-01-22
Store checkout threshold

peppy:
Until now, we've been negating stock of items temporarily while we are overwhelmed with orders. This is ugly from a user's perspective (the whole store is marked OUT OF STOCK) and a pain to manage.

This automates the process by adding a confirmation step for users placing orders to let them know that there will be delays on their orders. It also means that users can now place orders even when we are running behind, and they will be shipped in sequential order.


resources/assets/coffee/store.coffee
```diff
@@ -77,6 +77,17 @@ $(document).on 'input', '#username.form-control', ->
     $status.html "Checking availability of #{requestedUsername}..."
     debouncedCheckUsernameValidity()
 
+#checkout checks
+checkCheckoutConfirmations = ->
+  $checkboxes = $('.checkout-confirmation-step')
+  $checkboxesChecked = $('.checkout-confirmation-step:checked')
+  $('#checkout-with-paypal').toggleClass('disabled', $checkboxesChecked.length < $checkboxes.length)
+
+$(document).on 'ready page:load', ->
```
nanaya: passing just function name also work (`$(document).on 'ready page:load', checkCheckoutConfirmations`)

resources/assets/coffee/store.coffee
```diff
@@ -77,6 +77,17 @@ $(document).on 'input', '#username.form-control', ->
...
+  checkCheckoutConfirmations()
+
+$(document).on 'change', '.checkout-confirmation-step', ->
```
nanaya: javascript related classes should be prefixed with `js-`...

resources/assets/coffee/store.coffee
```diff
@@ -77,6 +77,17 @@ $(document).on 'input', '#username.form-control', ->
     $status.html "Checking availability of #{requestedUsername}..."
     debouncedCheckUsernameValidity()
 
+#checkout checks
+checkCheckoutConfirmations = ->
+  $checkboxes = $('.checkout-confirmation-step')
+  $checkboxesChecked = $('.checkout-confirmation-step:checked')
```
nanaya: not really matter but `$checkboxes.filter(':checked')` also work.

resources/assets/coffee/store.coffee
```diff
@@ -77,6 +77,17 @@ $(document).on 'input', '#username.form-control', ->
...
+  $('#checkout-with-paypal').toggleClass('disabled', $checkboxesChecked.length < $checkboxes.length)
```
nanaya: Using `.js-checkout-confirmation--button` for this and `.js-checkout-confirmation--step` for the checkboxes will perhaps be easier to understand? (or some better class names)

config/osu.php
```diff
@@ -26,6 +26,10 @@
             'topic_ids' => array_map('intval', explode(' ', env('SLACK_WATCH_TOPIC_IDS', '259747'))),
         ],
     ],
+    'store' => [
+        'delayed_shipping_order_threshold' => 100,
+        'delayed_shipping_order_message' => '', //default uses checkout.delayed_shipping localised message.
```
nanaya: take from env instead? Saves modifying this file when needing custom message.

app/Http/Controllers/StoreController.php
```diff
@@ -115,7 +115,9 @@ public function getCheckout()
 
         $addresses = Auth::user()->storeAddresses()->with('country')->get();
 
-        return view('store.checkout', compact('order', 'addresses'));
+        $delayedShipping = Store\Order::where('orders.status', 'paid')->count() > env('DELAYED_SHIPPING_ORDER_THRESHOLD', 100);
```
nanaya: I vaguely remember having some kind of pain figuring out environment variables outside of config files so I usually prefer this flow

```
env -> config -> app
```

resources/views/store/checkout.blade.php
```diff
@@ -62,12 +62,30 @@
             <div class="osu-layout__sub-row osu-layout__sub-row--lg1">
                 <h1>Payment</h1>
 
+                @if($delayedShipping)
+                <div class="alert alert-warning">
+                    <p><strong>IMPORTANT: SHIPPING DELAYS</strong></p>
+
+                    <p>
+                        {!! Markdown::convertToHtml(env('DELAYED_SHIPPING_ORDER_MESSAGE', config('store.delayed_shipping_order_message', trans('store.checkout.delayed_shipping')))) !!}
```
nanaya: what's... happening here? ಠ_ಠ

`env(config(), trans())`?

PR #309, opened 2016-01-26
Red color take two

nanaya:
I also noticed that attempting to mix in non-US English when the rest of libraries use US English results in useless confusion (variables, method names, etc).

Thereby I propose using US English for everything except as display text (e.g. values in `resources/lang/en`) ┐(･_･)┌

### DIFF
--- a/resources/assets/less/app.less
+++ b/resources/assets/less/app.less
@@ -20,7 +20,10 @@
 @node_root: "../../../node_modules";
 
 @import "@{node_root}/font-awesome/less/font-awesome";
+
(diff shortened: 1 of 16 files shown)


peppy:
Was the US english thing related to something in this commit or just a general statement? I guess I don't mind as long as we're consistent..

nanaya:
General statement. Especially noticeable when working with CSS properties and bootstrap variables.

Though at least constructor in php isn't called `initialize` or otherwise it'll be even more jarring.

PR #311, opened 2016-01-26
Add enabled state to store items.

peppy:
We need to be able to have hidden store items for internal use (tournament prizes for example) to generate invoices and tracking information. We also want to be able to hide items which are no longer relevant, such as the 2015 OWC banners.

This adds the basic ability to disable store products.


nanaya:
looks about right though there might be more state needed especially if user can view their previous purchases (linking to a 404 or just not linked at all isn't quite nice).

peppy:
we can add proper checking when that happens. need this merged today.

PR #317, opened 2016-01-29
User medals and historical

nanaya:
Rank history graph is coming soon™.


database/migrations/2016_01_29_090810_add_description_to_achievements.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('osu_achievements', function ($table) {
+            $table->text('description')->after('name');
```
peppy: needs a default of '' or nullable + default NULL else i think this will fail when applying (has no default to apply to existing entries.

nanaya: weird but it ran for me and somehow just filled them in with empty strings. I'll update it.

nanaya:
mysql quality, I guess ¯\_(ツ)_/¯

```
MariaDB [osu]> create table dragons (id int not null, herp varchar(255) not null);
Query OK, 0 rows affected (0.16 sec)

MariaDB [osu]> show create table dragons\G
*************************** 1. row ***************************
       Table: dragons
Create Table: CREATE TABLE `dragons` (
  `id` int(11) NOT NULL,
  `herp` varchar(255) NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4
1 row in set (0.00 sec)

MariaDB [osu]> insert into dragons (id) values (1);
Query OK, 1 row affected, 1 warning (0.02 sec)

MariaDB [osu]> show warnings;
+---------+------+-------------------------------------------+
| Level   | Code | Message                                   |
+---------+------+-------------------------------------------+
| Warning | 1364 | Field 'herp' doesn't have a default value |
+---------+------+-------------------------------------------+
1 row in set (0.00 sec)

MariaDB [osu]> select * from dragons;
+----+------+
| id | herp |
+----+------+
|  1 |      |
+----+------+
1 row in set (0.00 sec)
```

PR #323, opened 2016-02-02
Achievement fixes

peppy:
Achievements are now ordered correctly into sub-groups (database `ordering`).


nanaya:
table structure dump needs updating... :fireworks: 

I'll update this later.

PR #325, opened 2016-02-02
Misc fixes

nanaya:
- error handlers
- less page reinitialization
- bug fixes

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -303,17 +303,18 @@ function open_image($path, $dimensions = null)
         $dimensions = getimagesize($path);
     }
 
-    if ($dimensions === false) {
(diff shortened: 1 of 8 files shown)


app/helpers.php
```diff
@@ -303,13 +303,22 @@ function open_image($path, $dimensions = null)
...
     }
 
-    switch ($dimensions[2]) {
-        case IMAGETYPE_GIF:
+    if ($dimensions === false) {
+        return false;
+    }
+
+    try {
+        if ($dimensions[2] === IMAGETYPE_GIF) {
```
peppy: what's the reasoning behind if over switch here?

nanaya: something to do with avoiding php loose comparison.

PR #331, opened 2016-02-04
Add osu_languages structure and data

nanaya:
Didn't know there's a language called "Instrumental".


peppy:
learn something new every day :100:

PR #332, opened 2016-02-04
Use correct icon for nightcore@2x

nanaya:
Let's see if this works.

Nope :(


peppy:
i can skip CI anyway :grapes:

PR #334, opened 2016-02-05
Vagrant 2016

nanaya:
Now with php 7. I think this mostly works. The whole setup kind of sucks though ¯\\\_(ツ)_/¯


Nicarim:
Current vagrant has old node version (I had 0.10.25v when provisioning last time), maybe you should consider changing it?

nanaya:
it's already there

PR #341, opened 2016-02-09
Allow viewing own profile when restricted. Resolves #216.

### LINKED ISSUE #216: Unable to see own profile page when restricted
Would still be nice to be able to see one's own profile page, even when restricted.

peppy:
Also allow admins to view restricted profiles, as expected.

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -20,6 +20,7 @@
 namespace App\Models;
 
 use App\Transformers\UserTransformer;
+use Auth;
 use Cache;
 use Carbon\Carbon;
 use DB;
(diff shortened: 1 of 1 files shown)


nanaya:
the said user and admins won't be able to tell if the user is restricted or not (in forum, etc) this way. Is that all right?

peppy:
yeah, that's fine for now. we need to add a proper "restricted" display for admins (and user) going forward. probably gotta get a new design for that too.

app/Models/User.php
```diff
@@ -860,7 +872,7 @@ public function hasProfile()
     {
         return
             $this->user_id !== null
-            && !$this->isRestricted()
+            && $this->canView()
```
nanaya: I'd separate `hasProfile` to simply mean "is it actual user or bot/guest" and do separate call for authorization.

app/Models/User.php
```diff
@@ -220,10 +221,15 @@ public static function lookup($username_or_id, $lookup_type = null, $find_all =
         }
 
         if (!$find_all) {
-            $user = $user->where('user_type', 0)->where('user_warnings', 0);
+            $user = $user->where('user_type', 0);
         }
 
-        return $user->first();
+        $user = $user->first();
+        if (!$user || !$user->canView()) {
```
nanaya: why is there even this check in a lookup function? Doesn't seem to belong here.

peppy: it should be there. restricted/banned users shouldn't be returned anywhere system-wide.

PR #344, opened 2016-02-09
Make user card appear faster. Fixes #172

### LINKED ISSUE #172: User card appears too slow
Feels like the page fade and the user card fade are being done synchronously.

Bobo1239:
I'm not really satisfied with using `setTimeout triggerModal, 10` but as I'm new to webdev, I don't know better alternatives. It gets the job done but somebody more experienced can improve this.

### DIFF
--- a/resources/assets/coffee/bootstrap-modal.coffee
+++ b/resources/assets/coffee/bootstrap-modal.coffee
@@ -30,6 +30,13 @@ $(document).on 'hidden.bs.modal', '.modal', ->
 
 
 $(document).on 'show.bs.modal', '.modal', ->
+  # trigger modal display immediately instead of waiting for backdrop animation to finish
+  triggerModal = ->
+    $('.modal-backdrop').trigger 'bsTransitionEnd'
+
+  # have to wait a bit until the backdrop is created
+  setTimeout triggerModal, 10
+
   alignments = []
 
   for el, i in fixedElements


Nicarim:
Uh, where else was that animation also triggered? :open_mouth:

Bobo1239:
Which animation are you referring to?

Nicarim:
I mean, there are two animations, right? one is background shadow, and second is modal dropdown? Or am I wrong?

Bobo1239:
Yeah; I'm not sure if I understand your question but I can try to explain how this works.

After a `$(...).modal("show")` call, a callback function (which starts the modal dropdown) is passed to the function which creates the backdrop. The backdrop function waits for the 'bsTransitionEnd' event (or 150ms (hardcoded)) and calls the modal dropdown function which was passed to it.

Here I'm just calling `$(...).modal("show")`, wait a bit, and trigger the `bsTransitionEnd` event so the modal dropdown function gets called nearly instantly by the backdrop function.

Nicarim:
Ok, let me check something out in this code.

peppy:
This indeed seems to work, but the `setTimeout` does feel a touch hacky. @nanaya got a better proposal before I hit the merge button?

nanaya:
Nope. This seems to be the least hacky way to do it. At worst it'll fall back to original behavior.

Issue #343, opened 2016-02-09
Implement supporter page design


peppy:
This bounty is for the complete implementation of the supporter page design. This one is a good task for people who don't want to involve themselves with the back-end, as it's mostly just css/template design.

- Each of the "why" and "what" items should be localisable and extensible (ie. we should be able to add more and change them easily).
- The triangles in the PSD are outdated. Please use the new triangle pattern as used elsewhere on the [live site](http://new.ppy.sh/forum).
- As there are no interactive components yet, this page can be created without react as a direct template.
- CSS should be clean and in line with existing formatting see the [BEM disussion here](https://github.com/ppy/osu-web/issues/53) I guess.

![](https://puu.sh/n1tSe/61ea403b2a.jpg)

PSD: https://www.dropbox.com/s/m5n3hxvnyfwxws9/osu%21next%20-%20Support%20Page%20C.psd?dl=0

**[Bounty](https://goo.gl/nFdoyI) offered**: 150-300 depending on implementation quality.

As always, ask if you have any questions, and feel free to jump on the [osu!dev slack](http://osu.ppy.sh/p/slack) for more real-time discussion.

UnnoTed:
ill do it

RagnarLaud:
I'm gonna give it a shot

peppy:
Awarding the bounty here to @UnnoTed as his implementation is basically complete. Nice job :+1:.

PR #346, opened 2016-02-10
implemenTed supporter page design

UnnoTed:
https://github.com/ppy/osu-web/issues/343

supports localization
and easily extensible (supporter perks and support's blocks)
template kinda responsive

preview: https://i.imgur.com/MsIkF5r.png


peppy:
On a quick pass-over, one thing to note is that help.support is actually a different page from this one. This would probably fit better as community.support.

UnnoTed:
anything else so i can fix all at once?

peppy:
I'll need to check it out and look over it. Will definitely do so today, just a bit later!

peppy:
Here's some things I've noticed:

- Button cut off when window width is mobile-sized

- The logic behind the top header section is weird. Have a look how it's done on the forum and/or store and do it in a similar way. You don't need absolute positioning or custom z-index.
- We're moving away from bootstrap, so it'd be better if you use flexbox for the boxes on the page.

Nicarim:
It doesn't use BEM, does it.

UnnoTed:
updaTed

peppy:
I'll bring up a few remaining issues which would be great if you can look into, but I'm also going to award the bounty to you in advance.

- The top image has the drop shadow included, where as it should actually be in the css style (already present on http://new.ppy.sh/store/product/1 for reference).
- Maximum height of header should be limited in pixels, as per the design. At the moment it gets way too tall.
- Images should have @2x versions.

peppy:
Will need to make a few changes to the assets, but merging this for now!

PR #351, opened 2016-02-10
Chart fixes

nanaya:
First day is removed for now so the furthest back is 89 days (not that anyone can tell yet because there's no tooltip). The heuristic in old site is... _nice try_.

Not fixing the rank scale because I can't think of better solution yet.

Here's another case of log vs linear:

current idea is to check if difference between max and min rank is over an (or two) order of magnitude (coming soon™).

### DIFF
--- a/app/Models/RankHistory.php
+++ b/app/Models/RankHistory.php
@@ -38,8 +38,8 @@ class RankHistory extends Model
     {
         $data = [];
 
-        $startOffset = Count::currentRankStart();
-        $end = $startOffset + 90;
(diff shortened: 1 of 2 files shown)


peppy:
Yeah, there are some issues with exponential scaling but my reasoning was that being at that low rank will see very quick jumps if you actually play (the example you're showing is a mostly inactive player).

Tom94:
The issue in your screenshot is just, that the range of the log graph is way too large for the small changes. If the range was clamped to the data like your linear case, then the two graphs would look very similar in fact.

PR #355, opened 2016-02-11
Cast twitch's ID to float, since int was overflowing.

Nicarim:
Fixes situation where certain streams weren't possible to be highlighted.

### DIFF
--- a/app/Http/Controllers/CommunityController.php
+++ b/app/Http/Controllers/CommunityController.php
@@ -65,15 +65,15 @@ class CommunityController extends Controller
 
         $featuredStreamId = Cache::get('featuredStream');
         if ($featuredStreamId !== null) {
+            $featuredStreamId = floatval($featuredStreamId);
             foreach ($streams as $stream) {
-                if ($stream->_id !== $featuredStreamId) {
+                if (floatval($stream->_id) !== $featuredStreamId) {
                     continue;
                 }
                 $featuredStream = $stream;
                 break;
             }
         }
-
         return view('community.live', compact('streams', 'featuredStream'));
     }
 
@@ -84,7 +84,7 @@ class CommunityController extends Controller
         }
 
         if ($request->has('promote')) {
-            Cache::forever('featuredStream', intval($request->promote));
+            Cache::forever('featuredStream', floatval($request->promote));
         }
 
         if ($request->has('demote')) {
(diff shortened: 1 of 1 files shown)


nanaya:
why not just string.

Nicarim:
that was first that came to my mind, it doesn't matter that much at the end i guess.

Tom94:
This is a bad idea. Floats have less precision than ints, even if they have higher maximum numbers. Read up on them on wikipedia if you want more details.

You really should use either string, or if those intgers were 32bit, use 64bit integers.

nanaya:
(not related) and then javascript doesn't even have int :dancer:

app/Http/Controllers/CommunityController.php
```diff
@@ -65,8 +65,9 @@ public function getLive(Request $request)
 
         $featuredStreamId = Cache::get('featuredStream');
         if ($featuredStreamId !== null) {
+            $featuredStreamId = floatval($featuredStreamId);
             foreach ($streams as $stream) {
-                if ($stream->_id !== $featuredStreamId) {
+                if (floatval($stream->_id) !== $featuredStreamId) {
```
nanaya: btw, if int overflows, what's in `$stream->_id` to begin with? If it's already int, I don't see how casting will help?

I think it's more like victim of strict comparison upgrade and `Cache` casts everything to string.

Nicarim:
copy paste from slack:
sec.

[19:03] 
 ```      +"_id": 19511430960.0```<- twitch
```19511430960.0```
my casted float,

File: app/Http/Controllers/CommunityController.php (hunk 1 above)
Comment (Nicarim) [inline app/Http/Controllers/CommunityController.php]: Yeah, I think depending on cache driver, it casts to string. I also casted _id to float (but now I checked it, it is already float), just to be safe.

File: app/Http/Controllers/CommunityController.php [hunk 2]
```diff
@@ -84,7 +85,7 @@ public function postLive(Request $request)
         }
 
         if ($request->has('promote')) {
-            Cache::forever('featuredStream', intval($request->promote));
+            Cache::forever('featuredStream', floatval($request->promote));
```
nanaya: my guess says the cast will add '.0' to the stored string and thus will never pass the check. But maybe I'm wrong ¯\\\_(ツ)_/¯

peppy: i'd rather push to for 64-bit php for this project either way..

PR #358, opened 2016-02-12
Chart part 2

nanaya:
I don't think I've tested enough but heyyy ヽ( ﾟ ヮﾟ)ﾉ

Now using same scales and tick values for y axis (I think).

I'll let you guys fight over how to format the tooltip and axis values.

Unrelated commits included because this branch branches off #357.


peppy:
Something feels weird about the alignment in some cases:

PR #359, opened 2016-02-13
Correct migrations and update database dump

Bobo1239:
Add `osu.forum_forum_covers` and `osu.forum_topic_covers` to `bin/db_dump`.

Fix some inconsistencies between the migrations and the database dump:
- `osu_store.products`: the columns `promoted` and `display_order` have default values in the dump but the migration doesn't specify them
- `osu_store.orders`: the database has `paid_at` as an index but the migration doesn't
- the two oldest migrations don't exist (anymore) so they shouldn't be be in `osu.migrations`

After these changes `php artisan migrate:refresh` works (again).
The dump was created after that command in the vagrant box.

Regarding the change from floats to doubles:
Apparently this is done intentionally by Laravel and there is no difference.
https://github.com/laravel/framework/issues/9103

### DIFF
--- a/bin/db_dump
+++ b/bin/db_dump
@@ -11,6 +11,8 @@ _mysqldump() {
 
 _mysqldump osu \
   migrations \
+  forum_forum_covers \
+  forum_topic_covers \
   osu_achievements \
   osu_apikeys \
   osu_beatmaps \
--- a/database/db-osu-data.sql
+++ b/database/db-osu-data.sql
@@ -1,10 +1,11 @@
--- MySQL dump 10.15  Distrib 10.0.23-MariaDB, for debian-linux-gnu (x86_64)
+-- MySQL dump 10.13  Distrib 5.7.11, for Linux (i686)
 --
 -- Host: localhost    Database: osu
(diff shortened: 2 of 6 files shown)


Bobo1239:
As a heads-up: I'm currently moving all tables to migrations so we don't don't rely on SQL dumps anymore and follwing that I'll try to set up proper seeding (https://github.com/ppy/osu-web/issues/20, https://github.com/ppy/osu-web/pull/26).

peppy:
That could work well. I hope you come up with a semi-automated way to do it; there's quite a few tables!

Bobo1239:
I'm using [this](https://github.com/Xethron/migrations-generator) but there's still a lot to be done until the dump is equivalent to the original.

nanaya:
merge with master again~

Hughp135:
BTW: I don't think osu_genres shouldn't have auto increment set on the genre_id column as it won't allow me to insert a genre with an id of 0.

Bobo1239:
Closing this in favor of https://github.com/ppy/osu-web/pull/393.

PR #360, opened 2016-02-14
Added ability to reorder profile sections

### LINKED ISSUE #340: Profiles: Ability to re-order sections.
Need to add drag and drop support to reorder sections in your own user profile. The order should be saved to database so it is used whenever viewing your profile (as yourself or someone else).

Each section should have a "hamburger" button that can be used to begin the drag. While being dragged, it should be very obvious that the section is being dragged, via display of a "ghost" box attached to the cursor (and the existing one dimmed out to show what is being moved.

Check out the "Sortable list with handles" example [here](http://farhadi.ir/projects/html5sortable/) for a sample implementation. Not necessarily promoting this plugin as the best choice; just one possibility.

Keep in mind the fixed top menu will also need to update. As an extra, you could also make the fixed menu draggable for easy reordering (have to see how it feels, though).

**[Bounty](https://goo.gl/nFdoyI) offered**: 150-250 depending on implementation quality.

LiquidPL:
Fixes #340

We should also consider adding something like Selenium, so we can test what React renders.


database/migrations/2016_02_11_155642_add_extras_order_on_user.php
```diff
@@ -0,0 +1,56 @@
...
+class AddExtrasOrderOnUser extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('phpbb_users', function ($table) {
```
nanaya: probably should be part of `user_profile_customizations`.

peppy: definitely agree

app/Models/User.php
```diff
@@ -338,6 +338,33 @@ public function setApiKey($key)
         $this->api->save();
     }
 
+    /**
+     * Constants referring to the user profile extras (kudosu,
+     * recent activities, etc).
+     */
+    const USER_PAGE = 1;
```
nanaya: Also part of `UserProfileCustomization` and see how thing like this is handled in beatmap modes.

app/Http/Controllers/AccountController.php
```diff
@@ -72,4 +74,72 @@ public function updatePage()
 
         return ['html' => $user->userPage->bodyHTML];
     }
+
+    /**
+     * Gets the new order of user profile page elements from request,
+     * and saves it into the database.
+     */
+    public function updateProfileExtrasOrder(HttpRequest $request)
```
nanaya: Probably can be merged with `updateProfileCover`?

app/Http/Controllers/AccountController.php
```diff
@@ -72,4 +74,72 @@ public function updatePage()
...
+        // so here we convert them back
+        $order = array_map(function ($x) {
+            if (gettype($x) === 'string') {
+                return intval($x);
+            } else {
+                return $x;
+            }
+        }, $order);
+
+        $count = array_fill(1, User::EXTRAS_COUNT, 0);
```
nanaya: not used?

app/Http/Controllers/AccountController.php
```diff
@@ -72,4 +74,72 @@ public function updatePage()
...
+            abort(403, trans('errors.codes.http-403'));
+        }
+
+        $user = Auth::user();
+        $order = Request::input('order');
+
+        // jQuery's AJAX methods convert ints to strings unfortunately,
+        // so here we convert them back
+        $order = array_map(function ($x) {
+            if (gettype($x) === 'string') {
```
nanaya: `intval`-ing an integer does nothing so just intval everything.

app/Http/Controllers/AccountController.php
```diff
@@ -72,4 +74,72 @@ public function updatePage()
...
+                            trans('errors.account.profile-order.duplicate'),
+                        ],
+                    ]);
+            }
+        }
+
+        $user->extras_order = $order;
+        $user->save();
+
+        return response()
```
nanaya: doesn't matter if to combined with profile update but returning a json in laravel can be done by just returning a php array (`return ['status' => 'OK'];` in this case).

app/Http/Controllers/AccountController.php
```diff
@@ -72,4 +74,72 @@ public function updatePage()
...
+            }
+        }
+
+        // Checking whether the values in the order array are unique,
+        // so we don't get repeating blocks on the profile page.
+        $occurences = array_count_values($order);
+
+        foreach ($occurences as $i) {
+            if ($i > 1) {
+                return response()
```
nanaya: input error is handled by returning a 422 (see `error_popup` helper).

app/Http/Controllers/AccountController.php
```diff
@@ -72,4 +74,72 @@ public function updatePage()
...
+                        'errors' => [
+                            trans('errors.account.profile-order.invalid-id', ['count' => User::EXTRAS_COUNT]),
+                        ],
+                    ]);
+            }
+        }
+
+        // Checking whether the values in the order array are unique,
+        // so we don't get repeating blocks on the profile page.
+        $occurences = array_count_values($order);
```
nanaya: did I miss a check if all pages are included? Not sure if allowing people to remove a section is a good idea.

resources/assets/coffee/react/profile-page/extra.coffee
```diff
@@ -19,21 +19,57 @@
 el = React.createElement
 
 class ProfilePage.Extra extends React.Component
+  # Mirrored from App/Models/User.php
+  # The numbers corresponding to particular profile section.
+  USER_PAGE = 1
```
nanaya: now I look again, why even bother having an integer constant mapping and not just use the section id string everywhere?

resources/assets/coffee/react/profile-page/extra.coffee
```diff
@@ -90,20 +126,51 @@ class ProfilePage.Extra extends React.Component
 
   _removeListeners: ->
     $.unsubscribe '.profileContentsExtra'
+    $.unsubscribe 'profilePageExtraToggle:toggleDragging'
```
nanaya: if you suffixed the name with `.profileContentsExtra`, you don't need to add `$.unsubscribe` here.

resources/assets/coffee/react/profile-page/extra.coffee
```diff
@@ -90,20 +126,51 @@ class ProfilePage.Extra extends React.Component
...
     $(window).off '.profileContentsExtra'
 
 
   _tabsStick: (_e, target) =>
     @setState tabsSticky: (target == 'profile-extra-tabs')
 
+  _toggleDragging: =>
+    if @state.draggingEnabled
+
+      csrfParam = $('meta[name=csrf-param]').attr('content')
```
nanaya: I believe csrf parameters are already added as long you use jquery's ajax (globally by jquery-ujs).

resources/assets/coffee/react/profile-page/extra.coffee
nanaya: Additionally, if it's for db storage saving, the integer can be converted immediately when storing/retrieving from db during casting phase so nothing else will need to do the mapping. If it's for bandwidth saving, then, well ┐('～`；)┌ (not sure if worth the effort)

resources/assets/coffee/react/profile-page/extra.coffee
```diff
@@ -90,20 +126,51 @@ class ProfilePage.Extra extends React.Component
...
+      csrfToken = $('meta[name=csrf-token]').attr('content')
+
+      osu.showLoadingOverlay
```
nanaya: this doesn't actually get called here since you forgot the parentheses.

resources/assets/coffee/react/profile-page/extra.coffee
nanaya: (also this variable isn't actually used because you did it wrong (should be `"#{csrfParam}": csrfToken` instead of `csrfParam: csrfToken`)

resources/assets/coffee/react/profile-page/extra.coffee
```diff
@@ -122,46 +189,55 @@ class ProfilePage.Extra extends React.Component
...
+        @props.user.profileOrder.map (m) =>
+          switch m
+            when USER_PAGE
+              if withMePage
+                div
+                  className: 'osu-layout__row js-profile-page-extra--scrollspy'
+                  id: 'me'
+                  el ProfilePage.UserPage, userPage: @props.userPage, withEdit: @props.withEdit, user: @props.user
+            when RECENT_ACTIVITIES
+              div
```
nanaya: I have a feeling this can be simplified.

resources/assets/coffee/react/profile-page/dragdrop-toggle.coffee
```diff
@@ -0,0 +1,35 @@
...
+    if @context.withEdit
+      span
+        className: 'fa fa-bars profile-extra__dragdrop-toggle'
+        onClick: @handleClick
+    else
+      # empty selector so react stops complaining that render() returns nothing
+      span
+        className: ''
+
+ProfilePage.DragDropToggle.contextTypes =
```
nanaya: it works in coffeescript by placing it inside the class declaration.

```coffeescript
class ... extends React.Component
  @contextTypes:
    withEdit: ...
```

LiquidPL:
i guess that would be it

app/Http/Controllers/UsersController.php
```diff
@@ -131,6 +132,11 @@ public function show($id)
             ])
         );
 
-        return view('users.show', compact('user', 'userArray', 'achievements'));
+        // The default order of the profile sections,
+        // used when the particular user has no profile customization
+        // object created.
+        $defaultProfileOrder = UserProfileCustomization::$sections;
```
nanaya: just set the default in `UserProfileCustomization`'s `__construct`. Look how cover's default is set.

package.json
```diff
@@ -10,6 +10,7 @@
     "font-awesome": "^4.5.0",
     "gulp": "^3.9.0",
     "jquery": "~2.1.0",
+    "jquery-ui": "^1.10.5",
```
nanaya: please update `npm-shrinkwrap.json` as well by rerunning `npm shrinkwrap`.

resources/assets/coffee/react/profile-page/extra.coffee
```diff
@@ -20,21 +20,40 @@ el = React.createElement
 
 class ProfilePage.Extra extends React.Component
   constructor: (props) ->
+    withMePage = props.userPage.html != '' || props.withEdit
+
+    if not withMePage
+      index = props.profileOrder.indexOf 'me'
+      props.profileOrder.splice index, 1
```
nanaya: props should not be modified.

nanaya:
I'm quite sure the reordering should only be possible by dragging the hamburger icon, not anywhere in the page ಠ_ಠ

Also the box is weirdly moved by quite a lot to the left when dragging starts.

https://puu.sh/n9MIy/233281292f.mp4

LiquidPL:
Also, should dragging be enabled on page load and order saved after each drag, or togglable like it is right now?

LiquidPL:
Also, should the dragging be enabled on page load, and order saved each time a section is moved, or togglable like it is right now?

LiquidPL:
provided that sections are draggable only by the toggle

peppy:
On page load sounds correct. Since it is (should be) restricted to a very small draggable area, there's no need for a toggle. Saving should happen in the background.

nanaya:
oh, I see what's happening now. Yeah, the hamburger should be just the dragging handle, not the toggle to initiate dragging mode.

nanaya:
moving around a big page is quite painful without mouse scroll though...

peppy:
Optimally the page would scroll when you drag near the edges.

As an alternative we can also make the header items draggable as I mentioned in the issue.



Another easier alternative we could try is collapsing each section down to just the header once the user begins to drag like this:

LiquidPL:
if an error occurs and order is not saved, should it revert to what was there before the save, or retry saving?

peppy:
Keep it simple; just revert for now.

LiquidPL:
>Another easier alternative we could try is collapsing each section down to just the header once the user begins to drag

Well based on my current coffeescript knowledge I would have to duplicate code for this in every sections component. Do you have any idea how to avoid it, as I tried creating a separate class and extending from it, but it seems that React doesn't like inheritance.

LiquidPL:
There seems to be a concept called [higher-order components](https://gist.github.com/sebmarkbage/ef0bf1f338a7182b6775), but it seems to work only in ECMAScript 6

nanaya:
collapsing the pages will cause problem with scroll position...

LiquidPL:
>Optimally the page would scroll when you drag near the edges

It should do this already...

peppy:
Let's just make it work without collapsing or anything. Can refactor after that.

LiquidPL:
Regarding 'moved to left' bug, it seems to:
* appear only in Chrome (or Blink, i guess),
* disappears if I remove `width: calc(100% ~'-' (@extra-margin * 2));` from `osu-layout__row`, and replace it with 40px padding on left and right, but I guess this will break the layout.

LiquidPL:
Everything done aside from getting the actualy drag'n'drop, because I can't find a working library.

LiquidPL:
I moved that code to PHP so props should not be modified anymore.

Also about the dragging problem, i solved it by applying the `width` and `margin: auto` to the container instead of the row, please let me know if this is ok.

resources/assets/less/bem/profile-extra.less
```diff
@@ -49,10 +49,38 @@
...
   }
+
+  &__dragdrop-toggle {
+    font-size: 18px;
+    position: relative;
+    float: right;
+    cursor: move;
+  }
+
+  &__list {
```
nanaya: bem violation aside, using `osu-layout__row` instead of this (at same div) seems to do same thing?

LiquidPL:
yes

peppy:
While we're here, can the default order of profiles be set to:

```
me
recent
top
medals
historical
beatmaps
kudosu!
```

peppy:
In fact, if you're happy to do a bit more work helping tidy up the ordering, here's a few more tasks that need being done (I'll adjust bounty accordingly):

- Rename "Recent Activities" to "Recent"
- Rename "Top Ranks" to "Ranks"
- Add a new section (after "me!") called "Performance"
- Move the Rank History graph from Historical to Performance (and remove the Rank History title).

app/Models/UserProfileCustomization.php
```diff
@@ -46,9 +46,41 @@ public function setCover($id, $file)
...
     }
 
+    /**
+     * An array of all possible profile sections, also in their default order.
+     */
+    public static $sections = ['me', 'recent_activities', 'kudosu', 'top_ranks', 'beatmaps', 'medals', 'historical'];
+
+    public function getExtrasOrderAttribute($value)
+    {
+        return unserialize($value);
```
nanaya: sorry, I keep forgetting about this but this should be stored as json. (just add the column to `$casts` variable above and remove the setter/getter)

nanaya: don't forget to update the migration as well.

database/migrations/2016_02_11_155642_add_extras_order_on_user_profile_customization.php
```diff
@@ -0,0 +1,48 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('user_profile_customizations', function ($table) {
+            $table->string('extras_order')->default(serialize(UserProfileCustomization::$sections));
```
nanaya: actually, keep the default empty so user will always have default ordering until they explicitly set the order.

LiquidPL:
done, i guess

peppy:
Looks good here, except for the missing localisation string:

app/Http/Controllers/UsersController.php
```diff
@@ -131,6 +131,13 @@ public function show($id)
             ])
         );
 
+        if ($userArray['data']['page']['data']['html'] === '' &&
+            (Auth::check() && Auth::user()->user_id !== $user->user_id)) {
+            if (($index = array_search('me', $userArray['data']['profileOrder'], true)) !== false) {
+                array_splice($userArray['data']['profileOrder'], $index, 1);
+            }
+        }
```
nekodex: I feel like this should be done at a react level not php?

LiquidPL: I did it there previously, but @nanaya complained that I was editing props

LiquidPL:
About the missing locale:

It seems to be here: http://m.imgur.com/dyjaxSU

Did you regenerate js strings?

peppy:
Yep, my bad. Works fine now.

app/Http/Controllers/UsersController.php
nekodex: Uh yeah, you shouldn't be editing props... but I'm pretty sure @nanaya meant for you to implement it differently, not for it to be done at a PHP level :P

LiquidPL:
Uaerpage logic should be in react now, could you check it? I'm on mobile and writing from a browser

PR #371, opened 2016-02-17
Limit number of top played maps shown in user profile to 100.

peppy:
Let's not return infinite results.


nanaya:
more like approximately 100 if there's any deleted beatmaps (unless the records are deleted as well when the beatmap is deleted). Doesn't really matter though, I guess? ┐(･_･)┌

PR #370, opened 2016-02-16
Beatmap updates

nanaya:
Namespaced react files, variable-width beatmapset panel, 10 difficulties icon before the +.

PR #372, opened 2016-02-17
React Turbolinks

nanaya:
As hinted by reactjs/react-rails#205. Now we have proper `componentWillUnmount` \o/

Branched off #370 .


nanaya:
broken, etc. Need some investigation.

nanaya:
I think it's fixed for now. And then look into turbolinks 5 because it seems to save some pain regarding a lot of stuff.

PR #374, opened 2016-02-18
Fixed a bug where only the user page was displayed

LiquidPL:


resources/assets/coffee/react/profile-page/main.coffee
```diff
@@ -79,7 +79,7 @@ class ProfilePage.Main extends React.Component
     profileOrder = @props.user.profileOrder.map (m) ->
       if m is not 'me'
```
nanaya: for reference the problem is this line which generates `m === !'me'`

PR #376, opened 2016-02-18
Support multiple tracking codes for store orders

peppy:
We ship orders in more than one parts when it makes sense from a logistics perspective. This allows that.


resources/views/store/invoice.blade.php
```diff
@@ -167,16 +166,15 @@
...
+                </p>
             @else
-                <div>
-                    <p><strong>Your order has been shipped!</strong></p>
-                    <p>
-                        We don't have tracking details as it was send in a small envelope, but you can expect to receive it within 1-2 weeks at the most. If you have any concerns, please reply to the order confirmation email you received (or <a href='mailto:osustore@ppy.sh'>send us an email</a>).
-                    </p>
-                </div>
+                <p>
+                    We don't have tracking details as it was send in a small envelope, but you can expect to receive it within 1-2 weeks at the most. If you have any concerns, please reply to the order confirmation email you received (or <a href='mailto:osustore@ppy.sh'>send us an email</a>).
```
smoogipooo: Typo: Should read "as it was sent".

peppy:
LOOKS GOOD PEPPY GREAT JOB :) (: :)

peppy:
(sorry, but i need this live now)

PR #384, opened 2016-02-19
Profile Page Anchors

CMS-Falcon:
Added the profile page anchors. I'm a bit new to git, so hopefully I'm doing this correctly :)

### DIFF
--- a/resources/assets/coffee/react/profile-page/extra.coffee
+++ b/resources/assets/coffee/react/profile-page/extra.coffee
@@ -38,26 +38,27 @@ class ProfilePage.Extra extends React.Component
     @_modeScan()
 
     if window.location.hash
+      title_id = @props.user.sections
       url = window.location.href
       anchor = url.substring(url.indexOf('#') + 1)
       mode = ''
       switch anchor
         when Lang.get('users.show.extra.me.title')
-          mode = 'me'
+          mode = @props.user.sections[0]
         when Lang.get('users.show.extra.performance.title')
-          mode = 'performance'
+          mode = @props.user.sections[1]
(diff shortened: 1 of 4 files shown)


resources/assets/coffee/react/profile-page/extra.coffee
```diff
@@ -37,6 +37,29 @@ class ProfilePage.Extra extends React.Component
     osu.pageChange()
     @_modeScan()
 
+    if window.location.hash
+      url = window.location.href
+      anchor = url.substring(url.indexOf('#') + 1)
+      mode = ''
+      switch anchor
+        when 'me!'
```
peppy: this should definitely not be hardcoded here.

CMS-Falcon:
Just to clarify, do you mean that piece of code belongs somewhere else, or that there is a problem with the way it is written (or both)?

peppy:
As in you shouldn't be adding redundant copies of header names/identifiers that could change at any point. Get it from its source (in the php classes).

CMS-Falcon:
Alright, will do. Thanks

CMS-Falcon:
I think this is what you were referring to, let me know if I'm still off.

peppy:
Still missing the point.. We have a list of all sections already, so you shouldn't be duplicating it redundantly:

CMS-Falcon:
Yeah, I saw that list and for some reason thought that wasn't what I needed to do. I'll just blame the lack of sleep. Thanks for bearing with me, if there's anything else missing I'll try to get to it tomorrow.

peppy:
See PR for a more correct implementation. Closing this one!

PR #392, opened 2016-02-20
Add Seeders for Users/Stats/Beatmaps/Forum/Misc data #20

Hughp135:
Added a seeder. This will use the osu api (with the help of your api key) to seed 500 real beatmaps, some from all game modes. It will then generate fake users and fill our their profiles with appropriate data (using the beatmaps imported from the api). If an API key isn't set, it will still seed what it can (e.g. users and some profile data).

Note: this deletes some tables before reseeding so if you want to make sure you don't lose any of your test data, back up your db first. Most tables (beatmaps/users/beatmaps/forums) aren't deleted unless they by chance have the same primary key(s) as the seed data.

Summary of changes:

- Tables that are seeded: phpbb_users, phpbb_user_stats, osu_beatmaps, osu_beatmapsets, osu_events, osu_leaders, osu_scores, osu_scores_high,  phpbb_forums, phpbb_topics, phpbb_posts, osu_user_beatmap_playcount, osu_favouritemaps

- A few model changes that were necessary for the seeder (and possibly in general)

- Added more statements to the sql dumps, as those fields were required.

- Added "php artisan db:seed" command to build.sh [edit: removed this command] .

- Environment variable 'OSU_API' will take your osu api key to seed beatmaps.

- Seeder makes a log file of all data imported from the jsons (in /database/data/json/log.txt)

- The password for every user is 'password' and they are all admins

Things left to seed: Everything osu-store related. More user profile details (website, playstyle, playmode, personal editable profile section etc)

P.S. if anyone's interested in seeing how I used data obtained from the osu API to generate jsons/zip files, check out https://gist.github.com/Hughp135/8386cde4fb0e196f85b7


Hughp135:
Good idea Bob1239, I will add a seeder for osu_genres and osu_languages as well.

database/seeds/UserBeatmapSeeder.php
```diff
@@ -0,0 +1,350 @@
+<?php
+
+use Illuminate\Database\Seeder;
+
+class UserBeatmapSeeder extends Seeder
```
LiquidPL: You moved user generating from the model factory to a seeder, but you forgot to change the tests appropriately.

Also, wouldn't a model factory be more useful to generate a single model on demand, as in testing? @nanaya?

database/seeds/UserBeatmapSeeder.php
```diff
@@ -0,0 +1,350 @@
...
+            $savedbeatmaps = [];
+            $savedscores = [];
+            $overbeatmaps = [];
+            $savedbeatmapsets = [];
+            $overbeatmapsets = [];
+            $error_log = [];
+            $exceptions = [];
+            $saveduserids = [];
+
+            $users = json_decode(file_get_contents($datapath.'users.json'));
```
LiquidPL: You moved user generating from the model factory to a seeder, but you forgot to change the tests appropriately.

Also, shouldn't we have a model factory to generate random clean users for unit testing? Picking an user from database at random seems unreliable.

build.sh
```diff
@@ -23,6 +23,8 @@ php composer.phar install
 php artisan migrate --force
 php artisan lang:js resources/assets/js/messages.js
 
+php artisan db:seed --force
```
nanaya: build.sh is run during deployment so this will probably destroy production data if run

database/db-osu-structure.sql
```diff
@@ -323,7 +323,6 @@ CREATE TABLE `osu_favouritemaps` (
   `user_id` mediumint(9) unsigned NOT NULL,
   `beatmapset_id` mediumint(9) unsigned NOT NULL,
   `dateadded` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
-  PRIMARY KEY (`user_id`,`beatmapset_id`),
```
nanaya: ?

nanaya:
I don't see why the seed data need to be zipped.

build.sh
Hughp135: Will remove this line then

database/db-osu-structure.sql
Hughp135: Readded the line back in as it should be there.

Hughp135:
There isn't really a need to zip it. It's just to save a few hundred KBs of space and reduce the number of files in the repository :P

database/seeds/UserBeatmapSeeder.php
Hughp135: I was not aware that the factory was being used in a test. Is this test still necessary?

I will redefine the user factory so that this test can still pass.

nanaya:
just use json so it can be edited easily and properly handled by git. Also, make sure to prettyprint it instead of one big line.

Hughp135:
Done!

database/seeds/UserBeatmapSeeder.php
LiquidPL: Actually, you could try improving it a little, as my implementation was quite shit.

For instance, `username` and `username_clean` should contain the same thing, I guess? I'd it actually used somewhere in the new non-phpBB codebase?

nanaya:
need @peppy's confirmation regarding data included in the seed files.

database/seeds/UserBeatmapSeeder.php
```diff
@@ -0,0 +1,350 @@
...
+               $scores = json_decode(file_get_contents($datapath.'scores.json'));
+            foreach ($scores as $score) {
+                if ($existing_score = \App\Models\Score\Osu::where('user_id', $score->user_id)->where('beatmap_id', $score->beatmap_id)->first()) {
+                    $existing_score->delete(); // overwrite existing score
+                 $overscores[] = $score->user_id.'-'.$score->beatmap_id;
+                }
+
+                //  echo 'Over-riding: ';
+                     try {
+                         $sc = new \App\Models\Score\Osu;
```
nanaya: this should will achieve same thing without needing to manually assign everything

```
App\Models\Score\Osu::unguard();
new App\Models\Score\Osu($score);
App\Models\Score\Osu::reguard();
```

.gitignore
```diff
@@ -79,3 +79,5 @@ sftp-settings.json
 /resources/assets/js/build/
 
 composer-installer
+
+/database/data/json/
```
nanaya: not needed anymore?

peppy:
- Is there any reason not to just generate random user data? Seems like we'd only need the username anyway.
- Can we have the beatmap data fetched from the API as part of the seeding script rather than from a .json? Might be a better solution.

Hughp135:
- With front end devs in mind, having the entire user profile filled with useful data instead of just an empty profile with only a username makes it easier to jump in and fix stuff on the profile page. 

- The user profile data is kinda hard to generate (and most of it is linked to actual beatmaps for example) so I thought it would be easier to just use real data instead of thinking up generators for each field.

- Having a beatmaps json based on real life beatmaps makes it possible for front end devs to compare their local test site to the existing osu.ppy.sh, and looks nicer than having random words in there.

- If we had it all in the seeding script, we may need to cut down drastically on the 35k lines of json data being imported

peppy:
I mean actually make the API request from the seeding script, so none of the json needs to exist. And yes, I agree beatmap data should be pulled from the API. User data probably shouldn't; we can easily seed beatmaps first and then create fake user data based on those available beatmaps).

Hughp135:
Sure sounds ok but 2 issues. 1 apo key. 2 how many calls would be
acceptable? The amount of data in the jsons took about 600 calls and
several minutes

peppy:
1. I'd recommend having the dev specify their own API key in the their local .env.
2. any number of calls are fine. the API limits itself if you call too fast (which shouldn't be the case with synchronous calls). I'd be more worried about the time it takes to run, rather than number of calls.

Hughp135:
Ok then. This should be fine as I already have the code written for that. I
will probably cut down the number of calls though so it fits within about
30secs. It will result in a smaller sample set but it's not a big deal.

Also it is an extra setup step for new users who might not even have a key
yet

Hughp135:
Since we've decided to make API calls directly from the seeder, i'm going to make some changes to the order in which things are seeded.

1st: Use the get_beatmaps API method which allows you to get up to 500 beatmaps in one call. 
2nd: Seed Beatmaps and their Sets in a BeatmapSeeder
3rd: Seed fake users in a UserSeeder, using randomly beatmaps for their top plays/recent activity/events etc. 
4th: Player stats e.g. rank, experience, country etc will be randomly but (hopefully :P) intelligently generated.

As an added bonus I'm hoping this will also allow me to seed stats and maps for all game modes (taiko, ctb, mania) as well as standard. This will all take some time but should result in a decent seeder.

peppy:
If an API key isn't provided you can just skip the beatmap population step? Maybe throw a red error message in there.

Honestly I don't see this as being a huge extra step (and for someone who is seriously looking at getting involved, it should be good as a "sanity check" hoop they have to jump though before going further, really).

As long as it's documented it should be fine.

Hughp135:
Ok I've finished all the mentioned changes, all went well!

The current state of this pull request is completely different from how it started out :P in fact most of the files and code are completely different. Was even considering making a new pull request with the fresh files :P

Lemmmy:
you should squash the commits

Hughp135:
my fav. part of the seeder was when generating users' rank history. it gives each user their own learning curve (e.g. slow or fast learner), how often they played in the last 90 days, with a chance of having a 'good day' every now and then where they jump up ranks, plus applies a base rank decay when they're inactive :D

.env.example
```diff
@@ -8,6 +8,8 @@ DB_DATABASE=osu
 DB_USERNAME=osuweb
 # DB_PASSWORD=
 
+OSU_API=
```
peppy: I'd probably call this `OSU_API_KEY` just for clarity.

Hughp135:
Done. I was wondering whether or not it would be better to delete the db tables before seeding them? If for example someone doesn't enter an API key before running the seeder, then add the key and re-run the seeder, it will cause a bunch of old users without beatmap stats to still be in there if the tables weren't deleted.

Hughp135:
I did a fresh reinstall of vagrant to test if everything was working and ran into some errors when it tried to run composer install. This is unrelated to this pull request, but I noticed the php extensions mbstring and dom were missing.

Adding these lines to bootstrap.sh fixed it. (adding this in as a commit)
```
php7.0-mbstring \
    php7.0-xml \
```

Bobo1239:
If https://github.com/ppy/osu-web/pull/393 gets accepted, developers can just run `php artisan migrate:refresh --seed` to repopulate the db and seed it.

Hughp135:
Any chance of some feedback on what needs to be improved here if anything?

peppy:
@Hughp135 for a start it would be great if you can resolve the conflicts :).

Hughp135:
Done - it's been a while since I've worked on this so wasn't aware of the conflict!

peppy:
and... again >_>

app/Models/User.php
```diff
@@ -540,7 +540,7 @@ public function rank()
 
     public function rankHistories()
     {
-        return $this->hasMany(RankHistory::class);
+        return $this->hasMany("App\Models\RankHistory", 'user_id', 'user_id');
```
nanaya: ?

database/db-osu-data.sql
```diff
@@ -67,6 +67,13 @@ INSERT INTO `osu_languages` VALUES (8,'German',4);
 INSERT INTO `osu_languages` VALUES (9,'Swedish',9);
 INSERT INTO `osu_languages` VALUES (10,'Spanish',8);
 INSERT INTO `osu_languages` VALUES (11,'Italian',5);
+
```
nanaya: remove this file if already done by seed?

app/Models/User.php
Hughp135: the relation between user and rankhistories was broken, so I changed the statement to match how all the other relations are defined. is this ok?

nanaya: broken how?

Hughp135: the statement was unlike all the others, and the 'user_id' column needed to be defined for it to work.

nanaya: It is using new style. Also the default already understands the column name.

What exactly the error you're entering?

Hughp135: It was when executing $user->rankHistories->save($hist) in the seeder

Hughp135: Seems like the primary key isn't set in RankHistory - returns 

```
>>> $r->delete()
Illuminate\Database\QueryException with message 'SQLSTATE[42S22]: Column not found: 1054 Unknown column 'id' in 'where clause' (SQL: delete from `osu_user_performance_rank` where `id` is null)'
```

Maybe should just add to RankHistory: `protected $primaryKey = 'user_id';`?

nanaya: laravel doesn't support composite primary key so the only way is to use conditional delete on class level.

(also fwiw I tried reverting this line and ran the seeder again; it went fine)

Hughp135: All right, I've reverted it back then.

Bobo1239:
Perhaps we should add a couple of words to the README regarding repopulating and reseeding the db. (`php artisan migrate:refresh --seed`)

Hughp135:
Ok I'll add some more words in there.

Also as the genre_id and language_id field is still incrementing, I'm getting this error when trying to seed genres/languages

`SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '1' for key 'PRIMARY' (SQL: insert into `osu_languages` (`display_order`, `language_id`, `name`) values (0, 0, Any), (11, 1, Other), (1, 2, English), (6, 3, Japanese), (2, 4, Chinese), (10, 5, Instrumental), (7, 6, Korean), (3, 7, French), (4, 8, German), (9, 9, Swedish), (8, 10, Spanish), (5, 11, Italian))`

Incrementing fields won't allow you to set a value of 0 in them....

nanaya:
it does, just not using insert. Until it's fixed just create and follow by updating the id, I guess?

Hughp135:
All right, it'll update the id back to 0 on those two entries after creating them.

nanaya:
some files have weird indentations and unnecessary comments.

Hughp135:
that's what styleci wanted

nanaya:
the comments probably confused styleci. Try reindenting them again.

Hughp135:
Is the indentation ok now?

nanaya:
nope. Set whatever editor you're using to indent by 4 spaces.

Hughp135:
I've gone through and reindented everything :O

database/seeds/ModelSeeders/BeatmapSeeder.php
```diff
@@ -0,0 +1,125 @@
...
+
+            foreach ($beatmaps as $bm) {
+                // Here we are going to check if the current beatmap belongs to a new set, and make the set if necessary
+                if ($last_beatmapset === $bm->beatmapset_id || $first_map === true) {
+                    ++$beatmapset_versions;
+                    $beatmap_diff_names = $beatmap_diff_names.$bm->version.',';
+                    $set_playcount += $bm->playcount;
+                } else {
+                    // Create new beatmapset based on the PREVIOUS beatmap (since current one is in the next set)
+                    rtrim($beatmap_diff_names, ',');
```
nanaya: this doesn't actually do anything since the result isn't saved anywhere

database/seeds/ModelSeeders/BeatmapSeeder.php
```diff
@@ -0,0 +1,125 @@
...
+            $beatmapset_versions = 0;
+            $set_playcount = 0;
+
+            $first_map = true;
+
+            foreach ($beatmaps as $bm) {
+                // Here we are going to check if the current beatmap belongs to a new set, and make the set if necessary
+                if ($last_beatmapset === $bm->beatmapset_id || $first_map === true) {
+                    ++$beatmapset_versions;
+                    $beatmap_diff_names = $beatmap_diff_names.$bm->version.',';
```
nanaya: difficulty is saved in form `{version}@{mode}`. Also might as well save in array first and then join/implode when inserting to table.

```php
$beatmap_diff_names[] = $bm->version.'@'.$bm->playmode;
```

database/seeds/ModelSeeders/BeatmapSeeder.php
```diff
@@ -0,0 +1,125 @@
...
+                    $set_playcount += $bm->playcount;
+                } else {
+                    // Create new beatmapset based on the PREVIOUS beatmap (since current one is in the next set)
```
nanaya: doesn't this mean the last beatmap won't have corresponding beatmapset? Or am I missing something?

Hughp135: Yeah, seems you're right. Guess I'll need to write some extra code for the last beatmapset.

database/seeds/ModelSeeders/BeatmapSeeder.php
Hughp135: Ok, for the mode is it in string format e.g. osu/fruits/ctb/mania?

nanaya: in integer format.

Hughp135:
Ammended the beatmap with the corrections you pointed out  - I haven't actually been able to run this code to test if it works since I'm on a different pc atm but will check there are no errors later.

Hughp135:
Tested it and working

nanaya:
- fix styleci error
- remove `bin/db_seed`
- remove call to that file in `Vagrantfile` and `.travis.yml`
- rebase/merge master

Hughp135:
When merging in master, the user factory was defined again in ModelFactory.php.

I have therefore given the user factory in UserFactory.php an alias of 'userseeder' to resolve this conflict.

LiquidPL:
Well I don't think that the old user factory is necessary, provided that you add the `user_warnings` and `user_type` fields to your one.

Hughp135:
Ok, i'll add in those fields and remove it from the model factory.

database/seeds/ModelSeeders/UserSeeder.php
```diff
@@ -0,0 +1,120 @@
...
+        // DB::table('osu_user_performance_rank')->delete();
+
+        $this->beatmapCount = App\Models\Beatmap::count();
+        $this->faker = Faker::create();
+        if ($this->beatmapCount === 0) {
+            $this->command->info('Can\'t seed events, scores, events or favourite maps due to having no beatmap data.');
+        }
+
+        $this->stats_count = 0;
+        $this->hist_count = 0;
```
nanaya: is this and `stats_count` above used anywhere?

database/seeds/ModelSeeders/UserSeeder.php
```diff
@@ -0,0 +1,120 @@
...
+            }
+            $st3 = $u->statisticsOsu()->save(factory(App\Models\UserStatistics\Mania::class)->create(['country_acronym' => $country_code, 'rank' => $rank4, 'rank_score_index' => $rank4]));
+            if ($st3) {
+                ++$this->stats_count;
+            }
+            // END USER STATS
+
+            // RANK HISTORY
+
+            // Create rank histories for all 3 modes
```
nanaya: you mean four.

database/seeds/ModelSeeders/EventSeeder.php
```diff
@@ -0,0 +1,93 @@
...
+    {
+        // DB::table('osu_events')->delete();
+        App\Models\Event::unguard();
+
+        $beatmapCount = App\Models\Beatmap::count();
+        $faker = Faker::create();
+        $users = App\Models\User::all();
+
+        $generateEventText = function ($bm, $bms, $u, $rank) {
+        switch ($bm->playmode) {
```
nanaya: the indentation is wrong - it looked like the `run` function returns `$string`.

database/seeds/ModelSeeders/UserSeeder.php
Hughp135: was meant to be used as a kind of check to make sure it was created and avoid errors, but I don't think it's necessary so i"ll remove it.

Hughp135:
sigh. no idea why this has happened but somehow in the last couple of days the indentation messed up again.

database/seeds/ModelSeeders/UserSeeder.php
```diff
@@ -0,0 +1,106 @@
...
+            // USER STATS
+            $common_countries = ['US', 'JP', 'CN', 'DE', 'TW', 'RU', 'KR', 'PL', 'CA', 'FR', 'BR', 'GB', 'AU'];
+            $country_code = $common_countries[array_rand($common_countries)];
+
+            $rank1 = rand(1, 500000);
+            $rank2 = rand(1, 500000);
+            $rank3 = rand(1, 500000);
+            $rank4 = rand(1, 500000);
+            $st = $u->statisticsOsu()->save(factory(App\Models\UserStatistics\Osu::class)->create(['country_acronym' => $country_code, 'rank' => $rank1, 'rank_score_index' => $rank1]));
+            $st1 = $u->statisticsOsu()->save(factory(App\Models\UserStatistics\Taiko::class)->create(['country_acronym' => $country_code, 'rank' => $rank2, 'rank_score_index' => $rank2]));
```
nanaya: this doesn't seem right to attach Taiko stats through Osu relation but actually works.

Though probably the whole thing can be looped like this?

```php
foreach (array_keys(Beatmap::modes()) as $modeStr) {
    $modeStr = studly_case($modeStr);

    $rank = rand(...)
    $st = $u->statistics($modeStr)->save(factory("...\\{$modeStr}")->...

    ...create rank here...
}
```

database/seeds/ModelSeeders/UserProfileSeeder.php
```diff
@@ -0,0 +1,74 @@
...
+                        }
+                    }
+                    $leader = new App\Models\BeatmapLeader\Osu;
+                    $leader->beatmap_id = $bm['beatmap_id'];
+                    $leader->user_id = $usr_id;
+                    $leader->score_id = $bm['score_id'];
+                    $leader->save();
+                }
+            }
+        } catch (\Illuminate\Database\QueryException $e) {
```
nanaya: this and the catch below seems the same? (since `QueryException` inherits `Exception`)

database/seeds/ModelSeeders/UserProfileSeeder.php
```diff
@@ -0,0 +1,74 @@
...
+            echo "Error: Unable to save User Profile Data\r\n".$e;
```
nanaya: `$this->command->error`?

database/seeds/ModelSeeders/UserProfileSeeder.php
```diff
@@ -0,0 +1,74 @@
...
+            }
+
+            // FAVOURITE BEATMAPS AND BEATMAP PLAYCOUNTS FOR EACH USER
+
+            foreach (App\Models\User::all()as $usr) {
+                $bms = $usr->scoresBestOsu()->get();
+                $usr_id = $usr->user_id;
+
+                foreach ($bms as $bm) {
+                    if (DB::table('osu_favouritemaps')->where('user_id', $usr_id)->where('beatmapset_id', $bm['beatmapset_id'])->first()) {
```
nanaya: might as well just delete immediately without first checking

database/seeds/ModelSeeders/UserProfileSeeder.php
```diff
@@ -0,0 +1,74 @@
...
+                    }
+                    $fav = new App\Models\FavouriteBeatmapSet;
+                    $fav->beatmapset_id = $bm['beatmapset_id'];
+                    $fav->user_id = $usr_id;
+                    $fav->save();
+
+                    // Add a random couple few first place ranks
+
+                    $bm = $bms[rand(0, count($bms) - 1)];
+                    if (DB::table('osu_user_beatmap_playcount')->where('user_id', $usr_id)->where('beatmap_id', $bm['beatmap_id'])->first()) {
```
nanaya: this as well

database/seeds/ModelSeeders/UserProfileSeeder.php
```diff
@@ -0,0 +1,74 @@
...
+                    $playcount = new App\Models\BeatmapPlaycount;
+
+                    $playcount->user_id = $usr_id;
+                    $playcount->beatmap_id = $bm['beatmap_id'];
+                    $playcount->playcount = rand(0, 1500);
+                    $playcount->save();
+
+                    $bm = $bms[rand(0, count($bms) - 1)];
+                    if (DB::table('osu_leaders')->where('beatmap_id', $bm['beatmap_id'])->first()) {
+                        $bm = $bms[rand(0, count($bms) - 1)];
```
nanaya: why is it tried twice?

database/seeds/ModelSeeders/UserSeeder.php
```diff
@@ -0,0 +1,106 @@
...
+                        }
+                        $new_rank = round($hist->$prev_r * $improvement_modifier);
+                        if ($new_rank < 1) {
+                            $new_rank = 1;
+                        }
+                        // User rank will be modified by somewhere between 0.97 and 1.1 of current rank (realistic amount)
+                        $hist->$r = $new_rank;
+                    } else {
+                        $new_rank = round($hist->$prev_r * (rand(998, 999) / 1000));
+                        if ($new_rank < 1) {
```
nanaya: the duplicated logic can be move outside.

database/seeds/ModelSeeders/UserSeeder.php
```diff
@@ -0,0 +1,106 @@
...
+                $hist->r89 = $rank;
+
+                for ($i = 88; $i >= 0; $i--) {
+                    $r = 'r'.$i;
+                    $prev_r = 'r'.($i + 1);
+                    $prev_rank = $hist->$prev_r;
+                    // We wouldn't expect the user to improve every day
+                    $does_improve = $this->faker->boolean($play_freq);
+                    if ($does_improve === true) {
+                        $extreme_improvement = $this->faker->boolean(2); // 1% chance of extreme improvement today
```
nanaya: isn't `boolean(2)` means 2% of it being true, not 1%?

database/seeds/ModelSeeders/UserProfileSeeder.php
Hughp135: Because twice means there's less chance of a leader score being overwritten. I couldn't be bothered to write a while loop as the chance is already small.

Hughp135:
I've fixed the things that could potentially cause confusion, however some of the small changes you have suggested do beg the question whether they are over analysing that don't matter, e.g. code optimization to save a few lines of code when this code isn't performance critical.

database/seeds/ModelSeeders/UserSeeder.php
```diff
@@ -0,0 +1,106 @@
...
+        ];
+
+        // Create 10 users and their stats
+        factory(App\Models\User::class, 10)->create()->each(function ($u) {
+
+            // USER STATS
+            $common_countries = ['US', 'JP', 'CN', 'DE', 'TW', 'RU', 'KR', 'PL', 'CA', 'FR', 'BR', 'GB', 'AU'];
+            $country_code = $common_countries[array_rand($common_countries)];
+
+            $rank1 = rand(1, 500000);
```
nanaya: if you don't want to loop it at least change this to actually correlate with `$st` numbers. It's currently off by one (`rank1` is for `st`, `rank2` is for `st1`, etc).

database/seeds/ModelSeeders/UserSeeder.php
```diff
@@ -0,0 +1,106 @@
...
+                        if ($extreme_improvement === true) {
+                            $improvement_modifier = 1.5;
+                        } else {
+                            $improvement_modifier = $this->improvement_speeds[array_rand($this->improvement_speeds)];
+                        }
+                        $new_rank = round($hist->$prev_r * $improvement_modifier);
+                        if ($new_rank < 1) {
+                            $new_rank = 1;
+                        }
+                        // User rank will be modified by somewhere between 0.97 and 1.1 of current rank (realistic amount)
```
nanaya: is this correct? Looks more like either 1.5 or somewhere between 1.02 and 1.15

database/seeds/ModelSeeders/UserSeeder.php
```diff
@@ -0,0 +1,106 @@
...
+                        $hist->$r = $new_rank;
+                    } else {
+                        $new_rank = round($hist->$prev_r * (rand(998, 999) / 1000));
+                        if ($new_rank < 1) {
+                            $new_rank = 1;
+                        }
+                        $hist->$r = $new_rank; // Slight decay of between 0.98 and 0.99
```
nanaya: 0.998 and 0.999?

nanaya:
the intention isn't to make it faster but to make it less confusing to read.

database/seeds/ModelSeeders/UserSeeder.php
```diff
@@ -0,0 +1,106 @@
...
+                    $prev_r = 'r'.($i + 1);
+                    $prev_rank = $hist->$prev_r;
+                    // We wouldn't expect the user to improve every day
+                    $does_improve = $this->faker->boolean($play_freq);
+                    if ($does_improve === true) {
+                        $extreme_improvement = $this->faker->boolean(2); // chance of extreme improvement today
+                        if ($extreme_improvement === true) {
+                            $improvement_modifier = 1.5;
+                        } else {
+                            $improvement_modifier = $this->improvement_speeds[array_rand($this->improvement_speeds)];
```
nanaya: why is it done in roundabout way? The user doesn't seem to be going to be constantly being any type of player this way? (also the value of `improvement_speeds` itself is static).

database/seeds/ModelSeeders/UserSeeder.php
Hughp135: A decay of 0.01-0.02% per day may be a little forgiving, it could perhaps be increased to 0.02-0.03%

database/seeds/ModelSeeders/UserSeeder.php
```diff
@@ -0,0 +1,108 @@
...
+                // Start with current rank, and move down (back in time) to r0
+                $hist->r89 = $rank;
+
+                for ($i = 88; $i >= 0; $i--) {
+                    $r = 'r'.$i;
+                    $prev_r = 'r'.($i + 1);
+                    $prev_rank = $hist->$prev_r;
+                    // We wouldn't expect the user to improve every day
+                    $does_improve = $this->faker->boolean($play_freq);
+                    if ($does_improve === true) {
```
nanaya: from what I read this in the end seems to be just for deciding the rank modifier.

```php
if ($this->faker->boolean($playFreq)) {
    if ($this->faker->boolean(2)) {
        $modifier = 1.5;
    } elseif (
       $modifier = rand($userImprovementSpeed[0], $userImprovementSpeed[1]) / 100;
} else {
    $modifier = rand(97, 99) / 100;
}

$newRank = round($hist->$prevR * $modifier);
if ($newRank < 1) {
    $newRank = 1;
}

$hist->$currentR = $newRank;
```

database/seeds/ModelSeeders/UserSeeder.php
```diff
@@ -0,0 +1,108 @@
...
+
+                $improvement_speed = $this->improvement_speeds[array_rand($this->improvement_speeds)];
+
+                // Start with current rank, and move down (back in time) to r0
+                $hist->r89 = $rank;
+
+                for ($i = 88; $i >= 0; $i--) {
+                    $r = 'r'.$i;
+                    $prev_r = 'r'.($i + 1);
+                    $prev_rank = $hist->$prev_r;
```
nanaya: this variable isn't used.

database/seeds/ModelSeeders/UserSeeder.php
nanaya: also updated variable names since I didn't realize that "previous" means the next day's rank, not previous.

Hughp135: I use variables like $does_improve and $extreme_improvement, because it makes it easier to understand the logic behind the code. This is what I would find easier to read.

For me the line  `if ($this->faker->boolean($playFreq) {` doesn't make as much sense as:

```
$does_improve = $this->faker->boolean($play_freq);
       if ($does_improve === true) {
```

The variable $does_improve actually explains the concept of what the random chance applies to, and for me this is easier to understand.

database/seeds/ModelSeeders/ScoreSeeder.php
```diff
@@ -0,0 +1,214 @@
...
+                    'beatmapset_id' => $bm->beatmapset_id,
+                    'score' => rand(50000, 100000000),
+                    'maxcombo' => $maxcombo,
+                    'count300' => round($maxcombo * 0.8),
+                    'count100' => rand(0, round($maxcombo * 0.15)),
+                    'count50' => rand(0, round($maxcombo * 0.05)),
+                    'countgeki' => round($maxcombo * 0.3),
+                    'countmiss' => round($maxcombo * 0.05),
+                    'countkatu' => round($maxcombo * 0.05),
+                    'enabled_mods' => $possible_mods[array_rand($possible_mods)],
```
nanaya: at the number of this pattern is used, please just make a helper function in `app/helpers.php`:

```php
array_rand_value($array)
{
    return $array[array_rand($array)];
}
```

database/seeds/ModelSeeders/UserSeeder.php
nanaya: just put a comment?

```
// $playFreq chance of user playing today
// 2% chance of extreme improvement
// not playing, today's rank is higher (lower number) than next day's rank
```

nanaya: mixture with cleanup and assignment doesn't help with readability either

database/seeds/ModelSeeders/ScoreSeeder.php
```diff
@@ -0,0 +1,214 @@
...
+        foreach ($users as $k => $u) {
+            $osuBeatmaps = $beatmaps->where('playmode', 0)->take(20);
+            $taikoBeatmaps = $beatmaps->where('playmode', 1)->take(20);
+            $fruitsBeatmaps = $beatmaps->where('playmode', 2)->take(20);
+            $maniaBeatmaps = $beatmaps->where('playmode', 3)->take(20);
+
+            //add 20 osu! Standard scores
+            foreach ($osuBeatmaps as $bm) {
+                $bms = $allBeatmapSets->find($bm->beatmapset_id);
+                $maxcombo = rand(1, 5000);
```
nanaya: shouldn't this be `rand(1, $bm->count_total)`?

database/seeds/ModelSeeders/ScoreSeeder.php
Hughp135: Fully agree with you there! :P

database/seeds/ModelSeeders/BeatmapSeeder.php
```diff
@@ -0,0 +1,139 @@
...
+                    $set->bpm = $the_beatmap->bpm;
+                    $set->approved = $the_beatmap->approved;
+                    $set->approved_date = $the_beatmap->approved_date;
+                    $set->genre_id = $the_beatmap->genre_id;
+                    $set->language_id = $the_beatmap->language_id;
+                    $set->versions_available = $beatmapset_versions;
+                    $set->difficulty_names = $beatmap_diff_names;
+                    $set->play_count = $set_playcount;
+                    $set->favourite_count = $the_beatmap->favourite_count;
+                    $set->save();
```
LiquidPL: You should set `user_id` on beatmaps and beatmap sets because i keep getting errors if I try accessing those.

LiquidPL: I guess this applies to any model where there are foreign keys.

Hughp135: Right, I was wondering whether that field was necessary or not. I suppose I could put in random user IDs after seeding the users?

Hughp135:
I have changed the order in which the beatmap and user seeders run, so that the beatmap seeder is run after the user seeder. The beatmap and beatmapsets now set their user_id fields to a random user id.

Currently looking into whether any other foreign keys need to be set that haven't been.

Hughp135:
Would like to get people's input on other tables that could potentially be seeded e.g. achievements/user_achievements, countries, kudos_exchange, phpbb_ranks/smilies, user_profile_customizations.

Possible ammendments as well: 
- Should beatmapset thread_id be set to a random forum thread?
- I will also need to improve the leaders seeder (in userprofileseeder) to include other game modes.

LiquidPL:
For the time being the beatmapset's thread id should link to a thread whose first post follows this scheme:

    forum post
    --------------- (exactly 15 dashes)
    beatmapset description

Just keep in mind that peppy said that this is going to change at some point in the future.

Hughp135:
Thanks for the LiquidPL, shouldn't be too hard to add a forum thread for each beatmap.

database/seeds/ModelSeeders/UserSeeder.php
Sakisan: Here's my take on the readability. It's always possible to do more, but this seems like a good baseline. 
https://github.com/Sakisan/osu-web/blob/hugh-seeder-review/database/seeds/ModelSeeders/UserSeeder.php

LiquidPL:
Also, the seeded beatmaps are not appearing in the listing somehow.

Sakisan:
'forum_id' is not set on the forum posts. Can't edit a post because of that

I also wasn't able to create a new thread or post yet, but that may not be related to this. I haven't figured it out yet.

Hughp135:
Thanks Sakisan, will ensure it sets a forum_id. You should be able to create a new thread or post if logged in as admin

Hughp135:
just checking in and merging in the master branch.

nanaya:
merge once more and I'll merge it.

Hughp135:
done :)

PR #393, opened 2016-02-21
Move base tables to migrations

Bobo1239:
Based on https://github.com/ppy/osu-web/pull/359. Together with https://github.com/ppy/osu-web/pull/392, we don't rely on SQL dumps anymore.

I've written a small script to compare the db to the reference/original. The script and its outputs are [here](https://gist.github.com/Bobo1239/9101009618557ab33c19) (db_store.diff is empty).

Regarding the change from floats to doubles:
Apparently this is done intentionally by Laravel and there is no difference.
laravel/framework#9103


Bobo1239:
Need some help here. I've no idea why the build is failing on an unchanged file and only with hhvm.

Bobo1239:
Ok, so removing the `??` fixes the build problem but why???

Bobo1239:
[According to peppy, hhvm support will be dropped](https://osudev.slack.com/archives/jizz/p1456287833000332), so this PR will have to wait until that happens or somebody finds a fix. Should I close the PR for the time being?

Bobo1239:
Just a note: After https://github.com/ppy/osu-web/pull/392 gets accepted, the remaining `db_osu_data.sql` dump can be removed as well as the `bin/db_seed` script.

Hughp135:
Just tested this out on a new vagrant install in conjunction with my seeder and all seems to be created and seeded without errors :)

database/migrations/2015_01_01_133337_base_tables.php
```diff
@@ -0,0 +1,1298 @@
...
+            $table->timestamp('dateadded')->useCurrent();
+            $table->primary(['user_id', 'beatmapset_id']);
+        });
+        $this->setRowFormat('osu_favouritemaps', 'DYNAMIC');
+
+        Schema::create('osu_genres', function (Blueprint $table) {
+            $table->charset = 'utf8';
+            $table->collation = 'utf8_general_ci';
+
+            $table->smallIncrements('genre_id');
```
Hughp135: shouldn't increment

Bobo1239:
I agree with you but my target was to replicate the original db as closely as possible. The reason for that is to not diverge from the state of the db on the actual osu webserver. I believe that those fixes should be made in a seperate migration so that they are also made on the actual server.
(The exception being the `osu_leaders_taiko` table as I asked peppy on slack and he sent me the relevant part of their sql create table command which didn't include autoincrement)

I'm not sure if that reason is a valid concern. Maybe we can get more input on this issue? (especially from the main team)

database/db-osu-data.sql
```diff
@@ -13,28 +14,28 @@
...
-INSERT INTO `migrations` VALUES ('2015_07_09_091745_add_tournament_description_text',12);
-INSERT INTO `migrations` VALUES ('2015_07_09_105144_create_tournaments_registrations_table',12);
-INSERT INTO `migrations` VALUES ('2015_08_10_083016_add_slug_to_achievements',14);
-INSERT INTO `migrations` VALUES ('2015_11_16_065319_create_topic_covers_table',15);
-INSERT INTO `migrations` VALUES ('2015_11_20_122003_add_unique_index_to_forum_topic_covers_forum_id',16);
-INSERT INTO `migrations` VALUES ('2015_12_07_061307_create_forum_covers_table',17);
-INSERT INTO `migrations` VALUES ('2016_01_12_074035_unique_user_customization_user_id',18);
-INSERT INTO `migrations` VALUES ('2016_01_26_194005_disable_products',19);
-INSERT INTO `migrations` VALUES ('2016_01_29_090810_add_description_to_achievements',20);
+INSERT INTO `migrations` VALUES ('2015_01_01_133337_base_tables',1);
```
nanaya: this should be removed because `artisan migrate` already adds it.

database/migrations/2015_01_01_133337_base_tables.php
```diff
@@ -0,0 +1,1298 @@
...
+    {
+        Schema::create('osu_achievements', function (Blueprint $table) {
+            $table->charset = 'utf8';
+            $table->collation = 'utf8_general_ci';
+
+            $table->mediumIncrements('achievement_id');
+            $table->string('name', 40);
+            $table->string('image', 50);
+            $table->string('grouping', 30)->default('-');
+            $table->unsignedTinyInteger('ordering');
```
nanaya: size 10? (this one generates size 3)

database/migrations/2015_01_01_133337_base_tables.php
```diff
@@ -0,0 +1,1298 @@
...
+            $table->unsignedTinyInteger('progression');
```
nanaya: size 10? (this one generates size 3)

database/migrations/2015_01_01_133337_base_tables.php
```diff
@@ -0,0 +1,1298 @@
...
+            $table->index(['grouping', 'ordering'], 'display_order');
+            $table->index('quest_ordering', 'quest_ordering');
+        });
+        $this->setRowFormat('osu_achievements', 'DYNAMIC');
+
+        Schema::create('osu_apikeys', function (Blueprint $table) {
+            $table->charset = 'utf8';
+            $table->collation = 'utf8_general_ci';
+
+            $table->increments('key');
```
nanaya: size 10? (this one generates size 11)

database/migrations/2015_01_01_133337_base_tables.php
```diff
@@ -0,0 +1,1298 @@
...
+            $table->unsignedInteger('user_id');
```
nanaya: size 10? (this one generates size 11)

nanaya:
Looks fine. @peppy just need to remember adding the migration entries manually before deploying this change.

```
INSERT INTO `migrations` VALUES ('2015_01_01_133337_base_tables',1);
INSERT INTO `migrations` VALUES ('2015_01_01_133337_store_base_tables',1);
```

peppy:
done

peppy:
@Bobo1239 did you have a script which you used to generate the migration code from sql dumps? if so, it'd probably be a good idea to include it for (possible) future use as we add more tables to the osu-web distribution.

I'm awarding 300 bounty for your effort here, even though there was no bounty explicitly assigned to the task.

Bobo1239:
I used [this](https://github.com/Xethron/migrations-generator) but it (currently) doesn't support:
- mediumints
- table/field charsets/collations
- floats get replaced with doubles
- The indices don't maintain their order. (Does this even matter?)
- Other stuff I forgot about

Additionally, Laravel migrations are unable to replicate the scheme perfectly so there are SQL statements for:
- table/field comments
- composite autoincrement indices
- `ROW_FORMAT`
- sized binary fields
- specific triggers (timestamp)
- ...

Thanks for the bounty.

Issue #379, opened 2016-02-18
Add ability to rearrange profile sections from header menu


peppy:
Rearrange own profile sections by dragging the links at the top of the page:

![](https://puu.sh/nciOn/bc3e2e39a0.png)

**[Bounty](https://goo.gl/nFdoyI) offered**: 30

Bounty may increase depending on whether we need design iterations. Initial implementation doesn't need any specific design, just functionality.

LiquidPL:
This is actually somewhat tricky. 

In my current implementation, only the tabs' order is dependent on the state, as they need to be reordered with the sections.

However, when I make both tabs and sections state-dependent, something breaks, as when you reorder any of those, it snaps back to original, with the other behaving as expected.

I assume it's because React doesn't expect anything besides itself to change the DOM, which is what the sortable does.

I guess I will have to find a more React-compatible library, or write a drag and drop implementation from scratch.

peppy:
Has been completed by @nanaya.

Issue #378, opened 2016-02-18
Profile section changes should add #anchor to URL


peppy:
For instance, clicking on the "Medals" section should generate a URL like `http://new.ppy.sh/u/2#medals` which can be linked to other users (or bookmarked).

**[Bounty](https://goo.gl/nFdoyI) offered**: 40

As always, ask if you have any questions, and feel free to jump on the [osu!dev slack](http://osu.ppy.sh/p/slack) for more real-time discussion.

LiquidPL:
should this also happen when user scrolls to the section on his own?

peppy:
hmm, i don't think it needs to. anyone else have opinions on this?

Bobo1239:
Imo, it should only happen when clicking on the section button.

CMS-Falcon:
Should have this done very soon.

shavitush:
Does it need to be dynamic or static and just for the current available sections?

peppy:
Completed by @nanaya. @CMS-Falcon check out the solution for the kind of implementation we were looking for. Notice how it is not duplicating any logic which will cause headaches down the road when we add/reorders/remove sections.

PR #394, opened 2016-02-22
React fixes

nanaya:
- some cleanups
- some ~~refactors~~ rewrites
- fix reselecting custom cover
- hash in profile page. Implements #378?
- apparently now comes with draggable tab. Kind of implements #379 without actual UI
- one less bug when using php 7!!


nanaya:
did I say refactors? I mean rewrites :dancer:

PR #401, opened 2016-02-25
Ran all the PNG files through PNGOUT to reduce filesize.

kevinwlau:
Most of the PNGs were already optimized. Total size reduction of about 100KB.


peppy:
How many of these had significant reduction? I'd rather not clutter up file history for 3 byte differences.

Also we should probably settle on a tool to use for this which produces reproducible results.

kevinwlau:
I was considering that too. Only about 15 files were cut by more than 10-15%.

peppy:
10% is still a good reduction

PR #418, opened 2016-03-05
Stopped store being used for non-store requests and vice-versa

### LINKED ISSUE #399: Stop store.ppy.sh being used for non-store routes
This needs to be fixed (and vice-versa).

**[Bounty](https://goo.gl/nFdoyI) offered**: 30

As always, ask if you have any questions, and feel free to jump on the [osu!dev slack](http://osu.ppy.sh/p/slack) for more real-time discussion.

LiquidPL:
Fixes #399.

The domain settings need to be set in the `.env` file, notably the `BASE_DOMAIN`, `MAIN_PREFIX`, `STORE_PREFIX`, and `STAT_PREFIX` variables.

Afterwards, the entire thing is enabled by setting `USE_SUBDOMAINS` to `true`.


LiquidPL:
updated this to work with the slack page

README.md
```diff
@@ -54,6 +54,40 @@ For the time being, using React is generally preferred for pages which involve i
 
 Much of the CSS present abides to [BEM](http://getbem.com/) conventions but there is also a fair deal that doesn't. We haven't come to a unanimous decision about how to proceed going forward, so please use your own discretion or continue the discussion in the relevant [issue](https://github.com/ppy/osu-web/issues/53).
 
+### Working with separate osu and store subdomains
```
nekodex: I'd just put a single-line comment in `.env.example` saying "enable this and add hosts entries to..."

I don't think we don't need a walkthrough in the readme on how to add a hosts entry :stuck_out_tongue:

LiquidPL: fair point

LiquidPL:
updated :fireworks:

nanaya:
how does the laroute javascript thing handle this?

LiquidPL:
it looks like it disregards the domains :confused:

as in, `laroute.route('home')` returns `/` while laravel `route('home')` returns `http://osu.localhost:8080` (or, well, whatever is configured)

LiquidPL:
Apparently, in `config/laroute.php` there is an `absolute` option, however, it doesn't respect the subdomains anyway (`laroute.route('home')` -> `http://localhost:8080/`)

LiquidPL:
alright, after patching laroute a bit i got this to work.

however, that guy who made it seems not to respond to issues and stuff, so we might consider forking it into our org, i guess.

peppy:
Did you PR your fix or anything?

LiquidPL:
not, yet, but i've seen nanaya posting something there with no response, so i assumed he wouldn't bother anyway

peppy:
go for your life https://github.com/ppy/laroute

app/Http/routes.php
```diff
@@ -27,113 +27,197 @@
...
+        Route::get('/country', ['as' => 'country', 'uses' => 'RankingController@getCountry']);
+        Route::get('/mapper', ['as' => 'mapper', 'uses' => 'RankingController@getMapper']);
+    });
 
-Route::get('/beatmaps/{beatmaps}/scores', ['as' => 'beatmaps.scores', 'uses' => 'BeatmapsController@scores']);
-Route::get('/b/{beatmaps}', ['as' => 'beatmaps.show', 'uses' => 'BeatmapsController@show']);
+    // Livestreams section
+    Route::resource('livestreams', 'LivestreamsController', ['only' => ['index']]);
+    Route::group(['prefix' => 'livestreams', 'as' => 'livestreams.'], function () {
+        Route::post('livestreams/promote', ['as' => 'livestreams.promote', 'uses' => 'LivestreamsController@promote']);
```
nanaya: duplicated prefix

app/Http/routes.php
```diff
@@ -207,16 +280,9 @@
...
-    Route::group(['domain' => 'stat.ppy.sh'], function () {
-        Route::get('/', ['uses' => 'StatusController@getMain']);
-    });
-}
-
 // OAuth2 (for API)
-Route::get('oauth/authorize', ['as' => 'oauth.authorize.get', 'middleware' => ['check-authorization-params'], 'uses' => 'OAuthController@authorizeForm']);
-Route::post('oauth/authorize', ['as' => 'oauth.authorize.post', 'middleware' => ['check-authorization-params'], 'uses' => 'OAuthController@authorizePost']);
-Route::post('oauth/access_token', ['uses' => 'OAuthController@getAccessToken']);
+Route::group(['prefix' => 'oauth', 'as' => 'oauth.'], function () {
```
nanaya: apparently this should be only for main domain

app/Http/routes.php
```diff
@@ -27,113 +27,197 @@
...
+    // Authentication section (Temporarily set up as replacement/improvement of config("osu.urls.*"))
+    Route::get('/forgot-password', ['as' => 'forgot-password', function () {
+        return Redirect::to('https://osu.ppy.sh/p/forgot');
+    }]);
+    Route::get('/register', ['as' => 'register', function () {
+        return Redirect::to('https://osu.ppy.sh/p/register');
+    }]);
 });
 
+// Admin section
```
nanaya: should be in main domain as well. Though will give unfortunate effect of having store admin page in main domain :dancer:

nanaya:
- `APP_URL` now kind of duplicates `BASE_DOMAIN`
- remember that turbolinks doesn't work cross browser
- only login-related url should be shared across domains
- cookie/session needs to have correct domain (in `config/session.php`)

LiquidPL:
Isn't `APP_URL` used mostly by artisan actually?

nanaya:
it's at least in `config/app.php` (and hopefully the only place using it?)

LiquidPL:
I think so (and that's what the comment that is there says)

nanaya:
- `APP_URL` still needs to be removed (and replaced with combination of `USE_SSL` and `BASE_DOMAIN`)
- session cookie domain needs to be set to `.BASE_DOMAIN` when using subdomain

nanaya:
and `config('some.thing')` is slightly shorter than `Config::get('some.thing')`

LiquidPL:
apparently setting `session.domain` to `.localhost` (on dev environment) causes the website to throw a token mismatch on login (╯°□°）╯︵ ┻━┻

nanaya:
That's because browser doesn't allow a single word (tld-like) cookie domain (except in certain cases).

LiquidPL:
thanks for pointing that one out.

i think we should be good to go with this one.

nanaya:
the config ended up looking pretty dumb with "default" being set multiple times. And default/null (80/http or 443/https) port probably shouldn't have the port part at all in url.

nanaya:
oh and app.url should have the main subdomain as well otherwise it doesn't make sense.

nanaya:
assets will be loaded from each subdomain, won't they?

LiquidPL:
They should be, elixir generates relative asset URLs i think (is that bad?)

nanaya:
just doubles the stuff one will need download.

LiquidPL:
actually it looks like it always pulls stuff from the main domain (at least on production, i don't have access to my dev server right now), and I don't think this changed

nanaya:
looks like there's no good way to not repeat default values short of using helper or something else which loads before config is parsed.

config/app.php
```diff
@@ -1,5 +1,9 @@
 <?php
 
+$protocol = (env('USE_SSL', false) ? 'https' : 'http');
+$prefix = env('USE_SUBDOMAINS', false) ? '.'.env('MAIN_PREFIX', 'osu') : '';
```
nanaya: position of dot seems wrong.

LiquidPL:
should i do something like appending `:8080` to the app url when in the debug environment?

nanaya:
my environment doesn't have `:8080` so it should be a settable config.

nanaya:
might as well add wrapper function for `elixir` which adds assets domain to prevent loading same files from different domains.

LiquidPL:
>it should be a settable config

yes i had this in mind, blame my terrble wording

LiquidPL:
> wrapper function for `elixir()`

actually this breaks [fontawesome on store domain](http://puu.sh/pK7kL.png)

PR #419, opened 2016-03-08
Add better messaging for username availability.

peppy:
Had several people complain they aren't sure which minute/hour the new username would become available (and annoyance at the display staying on "1 day remaining" for much longer than one day). This fixes that, while not adding localisation support just yet.


nanaya:
is reducing the period to two years intended?

peppy:
yeah, decided 2 years is probably still worth displaying as "unavailable"

PR #421, opened 2016-03-09
Retune the Docker provider

Sakisan:
I tried to setup `vagrant up` with the docker provider but it does not work for me.
mysql and php7.0-fpm cannot be installed because of this:
```
==> default: Setting up mysql-community-server (5.7.11-1ubuntu14.04) ...
==> default: /usr/sbin/mysqld: error while loading shared libraries: libaio.so.1: cannot open shared object file: Permission denied
```

I solve that by removing the --privileged arg in the Vagrantfile, as is explained here: https://github.com/docker/docker/issues/5430

The build seems to run fine, even though it exits with a non-zero status.
The last message of the vagrant output is
```
==> default: mount: block device /data/node_modules is write-protected, mounting read-only
==> default: mount: cannot mount block device /data/node_modules read-only
```
but that does not seem to be an issue. (If it turns out otherwise after I've done some actual developing, I'll update this PR)

However I still need to start it up by doing `vagrant ssh` and `sudo service php7.0-fpm start`.
But I don't suppose that's related to this.


Sakisan:
Simplified summary: 

- Removed the --priviliged argument from the docker provider
- The mount command is now only configured for the virtualbox provider
- Restart all 3 services on start up: mysql, php7.0-fpm and nginx

PR #424, opened 2016-03-12
disabling-doubleposts feature

Risers:
Doubleposting is disabled server-side and user-side. Users can doublepost 3 days after last post.
User that created topic can doublepost every 24h.


app/Models/Forum/Topic.php
```diff
@@ -425,4 +426,30 @@ public function deleteWithCover()
...
         $this->delete();
     }
+
+    public function isADoublePost(User $user)
+    {
+        if ($user === null) {
+            return false;
+        }
+        if ($user->user_id === $this->topic_poster) {
+            if (Carbon::now()->subdays(1) > $this->topic_last_post_time) {
```
peppy: this should be config level

app/Models/Forum/Topic.php
```diff
@@ -425,4 +426,30 @@ public function deleteWithCover()
...
+                return false;
+            }
+            if ($user->user_id !== $this->topic_last_poster_id) {
+                return false;
+            }
+
+            return true;
+        } else {
+            if (Carbon::now()->subdays(3) > $this->topic_last_post_time) {
```
peppy: this should be config level

peppy:
what's with all the file permission changes? might want to check your git setup.

LiquidPL:
This gets created on a clean vagrant provision iirc

Sakisan:
I also get these modifications automatically right after the installation by vagrant.
```
~/projects/osu-web/code on master !
➜ gss
 M bootstrap/cache/.keep
 M storage/.gitignore
 M storage/app/.gitignore
 M storage/clockwork/.gitignore
 M storage/framework/.gitignore
 M storage/framework/cache/.gitignore
 M storage/framework/sessions/.gitignore
 M storage/framework/views/.gitignore
 M storage/htmlpurifier/.keep
 M storage/logs/.gitignore
```

No one noticed this before?

LiquidPL:
I did, but I didn't really care to bring that up

Sakisan:
I guess the quick fix is to just let these files be merged. I don't think that would be a problem.

Risers:
Kinda up?

app/Models/Forum/Topic.php
```diff
@@ -425,4 +426,30 @@ public function deleteWithCover()
 
         $this->delete();
     }
+
+    public function isADoublePost(User $user)
```
nanaya: should be `isDoublePostBy`

app/Models/Forum/Topic.php
```diff
@@ -425,4 +426,30 @@ public function deleteWithCover()
...
+    {
+        if ($user === null) {
+            return false;
+        }
+        if ($user->user_id === $this->topic_poster) {
+            if (Carbon::now()->subhours(config('osu.forum.hours_before_topic_author_doublepost')) > $this->topic_last_post_time) {
+                return false;
+            }
+            if ($user->user_id !== $this->topic_last_poster_id) {
```
nanaya: this should be checked before the outer conditional

app/Models/Forum/Topic.php
```diff
@@ -425,4 +426,30 @@ public function deleteWithCover()
 
         $this->delete();
     }
+
+    public function isADoublePost(User $user)
+    {
+        if ($user === null) {
+            return false;
+        }
+        if ($user->user_id === $this->topic_poster) {
```
nanaya: this conditional should just assign `$minimumHours` and do the actual check afterward.

resources/assets/coffee/forum/post-box.coffee
```diff
@@ -1,17 +1,13 @@
 ###
 Copyright 2015 ppy Pty. Ltd.
-
```
nanaya: ??

resources/assets/less/bem/forum-post.less
```diff
@@ -383,4 +385,24 @@
...
+        position:absolute;
+        top: 0;
+        left: 0;
+        height: 100%;
+        width: 100%;
+        text-align: center;
+        font-size: 30px;
+        color: white;
+        padding-top: 100px;
+	      filter:alpha(opacity=50);
```
nanaya: broken indentation, also there's no need for this ie crap

nanaya:
- what's with so many "style fixes" commit?
- don't change permission of unrelated files. And, no, those files should not be executable. 
- once activated (and is double post) the box can't be hidden since close button is obstructed by the overlay
- if the user has slow connection, he will be able to write something and then later the overlay appears
- every reply box activation triggers network round trip. Especially noticeable when scrolling back and forth from the bottom of the page while writing a reply

Risers:
Thanks

Risers:
Done

app/Models/Forum/Topic.php
```diff
@@ -403,4 +404,27 @@ public function deleteWithCover()
...
+    public function isDoublePostBy(User $user)
+    {
+        if ($user === null) {
+            return false;
+        }
+        if ($user->user_id !== $this->topic_last_poster_id) {
+            return false;
+        }
+        if ($user->user_id === $this->topic_poster) {
+            if (Carbon::now()->subhours(config('osu.forum.authorDoublePostTime')) > $this->topic_last_post_time) {
```
nanaya:
```php
if ($user->user_id === $this->topic_poster) {
    $minTime = config(...);
} else {
    $minTime = config(...);
}

return Carbon::now()->subhours(...) < $this->...;
```

config/osu.php
```diff
@@ -26,7 +26,8 @@
     'forum' => [
         'admin_forum_id' => intval(env('ADMIN_FORUM_ID', 28)),
         'help_forum_ids' => array_map('intval', explode(' ', env('HELP_FORUM_IDS', '4 5 29 30 101'))),
-
+        'doublePostTime' => 72,
```
nanaya: `=> ['normal' => 72, 'author' => 24]`?

nanaya:
Just make it like this



In `resources/views/forum/topics/show.blade.php` around line 137:

```php
<div class="forum-post__content forum-post__content forum-post__content--edit-bar">
    @if (priv_check('ForumTopicReply', $topic)->can())
        @include("forum.topics._post_box_footer", ["submitText" => trans("forum.topic.post_reply")])
    @else
        <span>
            <i class="fa fa-warning"></i>
            {{ priv_check('ForumTopicReply', $topic)->message() }}
        </span>
    @endif
</div>
```

And add relevant check in `app/Libraries/OsuAuthorize.php`, function `checkForumTopicReply`.

Risers:
Done

config/osu.php
```diff
@@ -29,6 +29,10 @@
     'forum' => [
         'admin_forum_id' => intval(env('ADMIN_FORUM_ID', 28)),
         'help_forum_ids' => array_map('intval', explode(' ', env('HELP_FORUM_IDS', '4 5 29 30 101'))),
+        'doublePostTime' => [
```
nanaya: everything else here is `snake_cased` so this one should be as well.

Risers:
Done

PR #428, opened 2016-03-17
Add StatsD support

nekodex:
:chart_with_upwards_trend:


composer.json
```diff
@@ -35,7 +50,8 @@
       "database"
     ],
     "psr-4": {
-      "App\\": "app/"
+      "App\\": "app/",
+      "League\\StatsD\\": "vendor/league/statsd/src/"
```
nanaya: is this actually needed

nekodex: sure is

PR #429, opened 2016-03-17
Simplify(?) storage usage

nanaya:
Now with more plain laravel storage and less `s3-v2`.


config/filesystems.php
```diff
@@ -46,6 +46,7 @@
         'local' => [
             'driver' => 'local',
             'root' => public_path().'/uploads',
+            'url_prefix' => '/uploads',
```
nekodex: why isn't this configurable too?

nanaya:
and then @peppy will need to update the config to include `S3_BASE_URL=https://assets.ppy.sh` ┐(･_･)┌

nanaya:
the s3 still works :dancer:

PR #438, opened 2016-03-31
Use regular expressions to pull out tracking codes specifically.

peppy:
Now that we are sending packages via airmail and other methods, the tracking code field can regularly contain notes for us and for the user which aren't valid tracking numbers.

This change matches tracking codes using the JP Post specification for them via regex.


peppy:
no response = good to go

PR #440, opened 2016-04-05
Fix shipped orders showing in a "pending" state

peppy:


peppy:
Contains... StyleCI fixes for commits outside the scope of this PR to make StyleCI happy.

nanaya:
ಠ_ಠ can I just rebase that instead

PR #443, opened 2016-04-07
Seeder updates

Hughp135:
I've updated a bunch of code and added in more seeders for extra juicy sample data.

Summary of changes:
- Seed osu store products
- Seed osu_countries table with random country names
- Seed osu_achievements and osu_user_achievements with some example medals
- Seed additional user profile fields including playstyle, kudosu, interests, join date, twitter/website page etc.
- Users are now made as an admin & supporter
- Slight ammend to readme file for the seeding section
- Reduce number of favourite beatmaps to 6 per user.


nanaya:
can you move the seeding logic to factory as many as possible? The seed is useless for writing tests.

Hughp135:
Are you referring to any particular classes, or just in general?

Mostly, the reason for so many fields being added by the seeders instead of in the factory is because of their reliance on previously seeded data or external data (e.g. for child/parent relationships), api data and fields that cant be randomly generated and still work).

nanaya:
in general. For example in this pr, the achievements (slug/grouping) and the countries (all attributes).

Hughp135:
I see what you mean with the above two examples. They could indeed be inside the factories.

If there are any others that stand out, let me know. I will have a quick look through the others to see if anything else can be moved over to factories.

Hughp135:
I've added a lang file with a list of country codes/names since it is a reliable way of getting a country name from a randomly generated code, but let me know if you would prefer not to use a lang file.

Hughp135:
This has been lying around for a while now, is there anything else that needs to be done, or any issues with this?

database/seeds/ModelSeeders/ProductSeeder.php
```diff
@@ -0,0 +1,61 @@
...
+                'size' => $sizes[$i],
+                'colour' => $colour,
+            ];
+            ++$i;
+            ++$y;
+        }
+        $master_tshirt->type_mappings_json = json_encode($type_mappings_json, JSON_PRETTY_PRINT);
+        $master_tshirt->save();
+
+        try {
```
nanaya: empty try?

database/seeds/ModelSeeders/ProductSeeder.php
```diff
@@ -0,0 +1,61 @@
...
+        $i = 1;
+        $y = 1;
+        $type_mappings_json[$master_tshirt->product_id] = [
+            'size' => 'S',
+            'colour' => 'White',
+        ];
+        foreach ($this->product_ids as $id) {
+            if ($i === 4) {
+                // Reset the counter back to 0 after every 4 so sizes are cycled
+                $i = 0;
```
nanaya: I thought I mentioned this before but why don't just use mod?

resources/lang/en/countries.php
```diff
@@ -0,0 +1,257 @@
+<?php
```
nanaya: this shouldn't be here yet. It'll also increase size of built javascript.

database/factories/CountryFactory.php
```diff
@@ -0,0 +1,38 @@
...
+
+$factory->define(App\Models\Country::class, function (Faker\Generator $faker) {
+    $acronym = $faker->unique()->countryCode;
+
+    while (App\Models\Country::where('acronym', $acronym)->count() > 0 || strlen($acronym) < 2) {
+        $acronym = $faker->unique()->countryCode;
+    }
+
+    // Get the country name from the countries list
+    if (trans('countries.'.$acronym)) {
```
nanaya: this will always be true.

database/factories/CountryFactory.php
```diff
@@ -0,0 +1,38 @@
...
+| Here you may define all of your model factories. Model factories give
+| you a convenient way to create models for testing and seeding your
+| database. Just tell the factory how a default model should look.
+|
+*/
+
+$factory->define(App\Models\Country::class, function (Faker\Generator $faker) {
+    $acronym = $faker->unique()->countryCode;
+
+    while (App\Models\Country::where('acronym', $acronym)->count() > 0 || strlen($acronym) < 2) {
```
nanaya: doesn't this mean this will never finish if the table contains all of faker's country code entry?

Hughp135: that's right, but why would you call the factory that many times?

nanaya: a bunch of users each with its own countries? Or just filling up the countries and the number is off by one?

Hughp135: All right, i'll fix this up soon

peppy:
why so closed? :(

Hughp135:
Hoping to have time to work on it next week or so, in which case I'll reopen it.

Hughp135:
Fixed all the things mentioned above (note that country code and country name are now generated randomly but unique).

I'm getting a strange bug when migrating the osu_achievements table. It doesn't set auto increment on the 'achievement_id'  field after migrating. It uses a mediumIncrements field, which appears to be working on other tables (e.g. forums, beatmaps), so I'm not sure why it isn't working. This causes the achievement seeder to fail for me. Is this the same for everyone?

Hughp135:
To get around the error with the achievement_id not incrementing, I set the id manually in the factory (unqiue number between 1-5000).

peppy:
I believe we removed the autoincrement from that table recently.

Hughp135:
Ok. Not sure why the build check is failing, it's giving some kind of network error?

LiquidPL:
@Hughp135 are you still there? what are the chances that you're going to finish this up?

Hughp135:
As far as I know I've done all the things that were suggested, if I have missed anything please let me know.

Hughp135:
I'm not sure how to fix this error with the test BeatmapControllerTest.php line 53. It is not returning the correct error code apparently. Not sure what this has to do with what I've done.

LiquidPL:
strange, i thought that was fixed ages ago :confused:

Hughp135:
Could this be because I changed the user factory to give the created user supporter?

The test could be using this factory to make the user, hence causing the fail?

LiquidPL:
That would be it.

Maybe don't set the supporter status in the factory for now, i think that it's quite easy to change anyways.

LiquidPL:
(or maybe create a second factory inheriting from the first one that has the supporter enabled)

Hughp135:
Ok I've moved supporter out of the factory and into the seeder instead, so it is fixed now.

LiquidPL:
:ok_hand:

Issue #466, opened 2016-04-14
Beatmap Info Page


peppy:
This bounty is for the new beatmap details page.

![](http://puu.sh/nPzv6/8fdd6dfe0c.png)

### Difficulty Selector

![](http://puu.sh/nPAkW/c987dad24f.png)

This is the mode and difficulty selector. 

The tabs are for the modes. It’s red when selected and gray if the mode is not available.

The buttons are the available difficulties, represented by difficulty level.

The title of the difficulty is stated under the buttons.

### Mapper Details and Downloads

![](http://puu.sh/nPAg0/740d73961b.png)

Area for mapper details. Submit and ranked date stated here.

osu!direct button will download the via the game client. Download button will download the map as a normal file.

There’s an checkbox if the user wants to download the map without video.

### Map Details

![](http://puu.sh/nPAPg/9ccd73c4f7.png)

Star chart in the middle. CS, SD, AR, OD and HP represents, Circle Size, Star Difficulty, Approach Rate, Overall Difficulty, and HP Drain.

Upper right is the standard map details. Lower left is music details.

![](http://puu.sh/nPB08/edf76e20ae.png)

It’s the same as the new profile page.

## Success Rate Chart

![](http://puu.sh/nPBBf/67aa5a4640.png)

## Scoreboard

![](http://puu.sh/nPBBY/e36f41645e.png)

Tabs changes the ranking mode. Please ignore the Mods:All.

Player ranked #1 displayed in a box with result details.

Rank list of #2—#50 listed under the #1 box. Grey highlights when hovered.

This time around, we'll be providing technical information required for implementation as we go, rather than fleshing this out in the initial post itself. Please show interest and we will work through it in this issue or in #osu-web on [slack](https://new.ppy.sh/community/slack)

Do note that @LiquidPL showed interest in this bounty, so he has priority on it should he still be wishing to take it up.

**[Bounty](https://goo.gl/nFdoyI) offered**: 500 upwards, depending on implementation quality.

LiquidPL:
Actually almost done with that, the only thing left is to plug the scoreboard to the backend, so the tabs actually do something ¯\_(ツ)_/¯.

Having that in mind, could you tell me how can I retrieve the user's friend list? (I asked you several times on slack but you didn't respond)

peppy:
Friends is retrieved from the phpbb_zebra table for now.

```CREATE TABLE phpbb_zebra
(
[...]

Issue #422, opened 2016-03-09
Use a helper library for routes in coffeescript instead of mirroring PHP


LiquidPL:
We could use something like [this](https://github.com/aaronlord/laroute), similarly to locales, instead of having routes duplicated in coffee.

LiquidPL:
why the hell was this never closed

PR #470, opened 2016-04-15
Beatmapset page

LiquidPL:
_fixes #466 yadda yadda_

Stuff done:
* beatmapset page
* reorganized certain profile page css classes to be reusable across pages
* moved all profile page css to bem while at it

Things to consider:
* The no video toggle looks ugly af when downscaled to @1x
* There is still no way to tell the client to download through osu!direct on a per song basis (i mean, through an `osu://` link)
* There was a hidden [beatmap rating](http://puu.sh/oiTN5.png) within the psd. Should this be implemented too?
* The [error thingy](http://puu.sh/oiTGP.png) displayed when a non-supporter tries to retrieve the country/friend rankings might need a design.
* I don't like the way how URLs are hardcoded in a CoffeeScript class. I recall bringing up an issue (#422 iirc) suggesting using a helper for retrieving Laravel routes, similarly to what we have now in locales. How about doing that?


LiquidPL:
i have no idea where did I break those goddamn tests

LiquidPL:
I can't seem to reproduce this outside of travis.

_plshalp ;_;_

resources/assets/coffee/react/beatmapset-page/success-rate.coffee
```diff
@@ -0,0 +1,76 @@
...
+    @_renderChart()
+
+  componentDidUpdate: ->
+    @_renderChart()
+
+  componentWillUnmount: ->
+    $(window).off '.beatmapSetPageSuccessRate'
+
+  _renderChart: ->
+    data = @props.failtimes.map (m) -> m.data
```
nanaya: seems to fail when failtimes data is empty.

app/Models/BeatmapSet.php
```diff
@@ -684,4 +686,14 @@ public function approver()
...
     }
+
+    public function description()
+    {
+        $topic = Topic::find($this->thread_id);
+        $post = Post::find($topic->topic_first_post_id);
+
+        $split = preg_split('[-{15}]', $post->bodyRaw);
+
+        return bbcode($split[1], $post->bbcode_uid, true);
```
nanaya: fail when description is missing/not following format (older beatmapsets)

app/Models/UserZebra.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Model;
+
+class UserZebra extends Model
```
nanaya: get a better class name (doesn't have to follow actual table name).

app/Transformers/BeatmapSetTransformer.php
```diff
@@ -39,12 +39,19 @@ public function transform(BeatmapSet $beatmap = null)
             'beatmapset_id' => $beatmap->beatmapset_id,
             'title' => $beatmap->title,
             'artist' => $beatmap->artist,
+            'description' => $beatmap->description(),
             'play_count' => $beatmap->play_count,
             'favourite_count' => $beatmap->favourite_count,
+            'submitted' => $beatmap->submit_date->formatLocalized('%b %d, %Y'),
```
nanaya: format time in the view, not in json. Use iso8601* here.

app/Transformers/ScoreTransformer.php
```diff
@@ -67,4 +72,20 @@ public function includeWeight(Score $score)
             ];
         });
     }
+
+    public function includeUser(Score $score)
+    {
+        return $this->item($score, function ($score) {
```
nanaya: there's `UserCompactTransformer`

app/Transformers/ScoreTransformer.php
```diff
@@ -67,4 +72,20 @@ public function includeWeight(Score $score)
...
+            $user = $score->user;
+
+            return [
+                'username' => $user->username,
+                'avatarUrl' => $user->user_avatar,
+                'country' => [
+                    'code' => $user->country_acronym,
+                    'name' => $user->countryName(),
```
nanaya: this will query database for every single scores. Find a better way.

database/migrations/2015_01_01_133338_create_beatmap_failtimes_table.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        // Checking whether the table already exists
+        // so it doesn't break when migrating on production
+        if (Schema::hasTable('osu_beatmap_failtimes')) {
+            return;
+        }
+
+        Schema::create('osu_beatmap_failtimes', function (Blueprint $table) {
+            $table->mediumInteger('beatmap_id');
+            $table->enum('type', ['FAIL', 'EXIT']);
```
nanaya: the actual enum values are downcased.

LiquidPL:
should be good now.

nanaya:
I'll write up some more tomorrow but for now please make sure to use `Beatmapset` instead of `BeatmapSet` anywhere new.

Also the url should be `/beatmapsets/{id}` (or just a single action `Route::resource`).

LiquidPL:
Should I change all previous instances to `Beatmapset` , or just leave them as they are?

nanaya:
feel free to try.

LiquidPL:
also, should i move the `/beatmaps/map/{id}` route under `/beatmaps/{id}` to be in line with beatmapsets?

nanaya:
yes, and use `Route::resource`.

LiquidPL:
provided that i'm moving beatmapsets into a resource controller, shouldn't [this](https://github.com/ppy/osu-web/blob/master/app/Http/Controllers/BeatmapController.php#L35) live in `BeatmapsetsController`?

nanaya:
sure is

app/Http/Controllers/BeatmapController.php
```diff
@@ -21,106 +21,60 @@
...
-
-        $ranks = [];
-        foreach (['XH', 'X', 'SH', 'S', 'A', 'B', 'C', 'D'] as $rank) {
-            $ranks[] = ['id' => $rank, 'name' => trans("beatmaps.rank.{$rank}")];
-        }
-
-        $filters = ['data' => compact('modes', 'statuses', 'genres', 'languages', 'extras', 'ranks')];
-
-        return view('beatmaps.index', compact('filters', 'beatmaps'));
+        return Redirect::to(route('beatmapsets.show', ['id' => $set->beatmapset_id]).'#'.$id);
```
nanaya: use `ujs_redirect` helper to redirect

app/Http/Controllers/BeatmapController.php
```diff
@@ -21,106 +21,60 @@
...
-                'rank' => array_filter(explode('-', Request::input('r')), 'strlen'),
-                'page' => Request::input('page'),
-                'sort' => explode('_', Request::input('sort')),
-            ];
+        $user = Auth::user();
 
-            if (!$current_user->isSupporter()) {
-                unset($params['rank']);
-            }
+        if ($type !== 'global' && !$user->isSupporter()) {
```
nanaya: looks like can be combined with previous check

```
if type != global
  if !user
    return 403
  else if !user->supporter
    return supporter only error message
```

resources/assets/coffee/react/beatmapset-page/main.coffee
```diff
@@ -0,0 +1,131 @@
...
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+{div} = React.DOM
+el = React.createElement
+
+class BeatmapsetPage.Main extends SwitchableModePage
```
nanaya: use mixin instead of inheritance

app/Transformers/BeatmapSetTransformer.php
```diff
@@ -39,12 +39,19 @@ public function transform(BeatmapSet $beatmap = null)
...
             'title' => $beatmap->title,
             'artist' => $beatmap->artist,
+            'description' => $beatmap->description(),
             'play_count' => $beatmap->play_count,
             'favourite_count' => $beatmap->favourite_count,
+            'submitted' => $beatmap->submit_date->toIso8601String(),
+            'ranked' => $beatmap->approved_date->toIso8601String(),
             'creator' => $beatmap->creator,
             'user_id' => $beatmap->user_id,
+            'avatarUrl' => $beatmap->user->user_avatar,
```
nanaya: use UserCompact transformer?

resources/assets/coffee/react/beatmapset-page/no-video-toggle.coffee
```diff
@@ -0,0 +1,29 @@
...
+# osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+{div, span} = React.DOM
+
+class BeatmapsetPage.NoVideoToggle extends React.Component
```
nanaya: check osu-checkbox

resources/assets/coffee/react/beatmapset-page/scoreboard-first.coffee
```diff
@@ -0,0 +1,59 @@
...
+              classModifiers: ['scoreboard']
+        div className: 'beatmapset-scoreboard-first__avatar-container',
+          div
+            className: 'beatmapset-scoreboard-first__avatar avatar avatar--beatmapset-scoreboard',
+            style:
+              backgroundImage: "url(#{@props.score.user.data.avatarUrl})"
+      div className: 'beatmapset-scoreboard-first__item beatmapset-scoreboard-first__item--right',
+        stats.map (m) =>
+          dt = Lang.get "beatmaps.beatmapset.show.extra.scoreboard.first.#{m}"
+          switch m
```
nanaya:
```
dd = @props.score[m]

dd =
  if m == 'accuracy'
    "#{Math.round ...}"
  else
    dd.toLocaleString()
```

resources/assets/coffee/_classes/url.coffee
```diff
@@ -19,14 +19,26 @@ class @Url
...
+  @beatmapset: (id) => "/beatmapsets/#{id}"
+
+  @beatmap: (id) -> "/beatmaps/#{id}"
+  @beatmapScores: (id) -> "/beatmaps/#{id}/scores"
 
   @beatmapDiscussionVote: (discussionId) => "/beatmap-discussions/#{discussionId}/vote"
   @beatmapDiscussionPost: (id) => "/beatmap-discussion-posts/#{id}"
   @beatmapDiscussionPosts: '/beatmap-discussion-posts'
 
+  @support: '/home/support'
```
nanaya: ?

resources/assets/coffee/react/_components/icon.coffee
```diff
@@ -17,4 +17,4 @@
 ###
 class @Icon extends React.Component
   render: =>
-    React.DOM.span className: "fa fa-#{@props.name}"
+    React.DOM.span className: "fa fa-#{@props.name} #{@props.className}"
```
nanaya: can't whatever uses this element style it at a parent instead? Most of the attributes in `fa` inherits the parent.

resources/assets/coffee/react/beatmapset-page.coffee
```diff
@@ -0,0 +1,40 @@
...
+#
+# osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+
+beatmaps = []
```
nanaya: this should be inside the propsFunction. (though now I look again it can also be a state in Main)

resources/assets/coffee/react/beatmapset-page/scoreboard-item.coffee
```diff
@@ -0,0 +1,48 @@
...
+        contents =
+          switch m
+            when 'rank'
+              "##{@props.position}"
+            when 'flag'
+              if @props.score.user.data.country
+                el FlagCountry,
+                  country:
+                    code: @props.score.user.data.country
+                    name: @props.countries[@props.score.user.data.country].name
```
nanaya: the country object has both `code` and `name`, why don't just pass it?

resources/assets/coffee/_classes/url.coffee
LiquidPL: used to link to supporter page if user hasn't got osu!direct

LiquidPL:
done for now

app/Models/BeatmapSet.php
```diff
@@ -684,4 +686,18 @@ public function approver()
     {
         return $this->belongsTo("App\Models\User", 'user_id', 'approvedby_id');
     }
+
+    public function description()
+    {
+        $topic = Topic::find($this->thread_id);
+        $post = Post::find($topic->topic_first_post_id);
+
+        $split = preg_split('[-{15}]', $post->bodyRaw);
```
nanaya: this should be `post_time`, `bodyRaw` is something different.

app/Http/routes.php
```diff
@@ -47,23 +47,19 @@
...
-    return Redirect::to('https://osu.ppy.sh/b/'.$id);
-}]);
+// Route::get('/beatmaps/packs', ['as' => 'packs', 'uses' => 'BeatmapController@getPacks']);
+// Route::get('/beatmaps/charts/{id?}', ['as' => 'charts', 'uses' => 'BeatmapController@getCharts']);
+
+Route::resource('/beatmaps', 'BeatmapController', ['only' => ['show']]);
+Route::get('/beatmaps/{id}/scores', ['as' => 'beatmap.scores', 'uses' => 'BeatmapController@getScores']);
+
+Route::resource('/beatmapsets', 'BeatmapsetsController', ['only' => ['index', 'show']]);
+Route::get('/beatmapset-search/{filters?}', ['as' => 'beatmaps.search', 'uses' => 'BeatmapsetsController@search']);
```
nanaya: should be just `beatmapsets/search`. Also the `as` part is wrong.

app/Http/Controllers/BeatmapController.php
```diff
@@ -21,106 +21,59 @@
...
     {
-        $fractal = new Manager();
-        $languages = Language::listing();
-        $genres = Genre::listing();
-        $beatmaps = fractal_collection_array(
-            BeatmapSet::listing(),
-            new BeatmapSetTransformer,
-            'beatmaps'
-        );
+        $set = Beatmap::find($id)->beatmapSet;
```
nanaya: `findOrFail`?

app/Http/Controllers/BeatmapController.php
```diff
@@ -21,106 +21,59 @@
...
+                $scores = $scores
+                    ->whereHas('user', function ($query) use (&$user) {
+                        $query->where('country_acronym', $user->country_acronym);
+                    });
+                break;
+            case 'friend':
+                $tableName = $scores->getModel()->getTable();
+
+                $scores = $scores
+                    ->join('phpbb_zebra', $tableName.'.user_id', '=', 'phpbb_zebra.zebra_id')
```
nanaya: can't this be just a `whereIn` of `$user`'s friends?

app/Transformers/BeatmapSetTransformer.php
```diff
@@ -39,15 +40,29 @@ public function transform(BeatmapSet $beatmap = null)
             'beatmapset_id' => $beatmap->beatmapset_id,
             'title' => $beatmap->title,
             'artist' => $beatmap->artist,
+            'description' => $beatmap->description(),
```
nanaya: probably should not be part of default json since it incurs two additional queries?

app/Models/BeatmapFailtimes.php
```diff
@@ -0,0 +1,69 @@
...
+
+    public function beatmap()
+    {
+        return $this->belongsTo(Beatmap::class);
+    }
+
+    public function delete()
+    {
+        // only because laravel can't seem to support composite primary keys
+        DB::table($this->table)
```
nanaya: `static::where->` is slightly shorter.

app/Models/BeatmapSet.php
```diff
@@ -684,4 +686,18 @@ public function approver()
...
+
+        // Return empty description if the pattern was not found
+        // (mostly older beatmapsets)
+        $description = count($split) > 1 ? $split[1] : '';
```
nanaya: `$split[1] ?? ''`

resources/assets/coffee/_classes/beatmapset-page-hash.coffee
```diff
@@ -0,0 +1,38 @@
...
+  @parse: (hash) ->
+    hash = hash[1..]
+    if @noMode hash
+      page: hash
+    else
+      split = hash.split '/'
+      mode: split[0]
+      page: split[1] || 'main'
+
+  @generate: (options) ->
```
nanaya: `{page, mode}` so it can be used with just `page` and `mode` inside the function.

resources/assets/coffee/_classes/url.coffee
```diff
@@ -19,14 +19,26 @@ class @Url
   @updateProfileAccount: '/account/update-profile'
   @pageAccount: '/account/page'
 
+  @beatmapDownload: (id, video = true) ->
+    url = "https://osu.ppy.sh/d/#{id}"
+    url += 'n' if not video
```
nanaya: `!` to `not`

app/Http/routes.php
LiquidPL: Wouldn't it get routed to `BeatmapsetsController@show` instead of `search` in this case,

nanaya: put it before the resource declaration.

LiquidPL:
done

also, the tests are apparently broken because of [this](https://github.com/LiquidPL/osu-web/blob/beatmap-page/tests/Controllers/BeatmapControllerTest.php).

if those tests are combined into a single method, everything seems to work.

nanaya:
add this line to the `travis.yml`:

`script: ./vendor/bin/phpunit`

nanaya:
tickbox images should be removed?

app/Http/Controllers/BeatmapController.php
```diff
@@ -21,106 +21,58 @@
...
+            case 'country':
+                $scores = $scores
+                    ->whereHas('user', function ($query) use (&$user) {
+                        $query->where('country_acronym', $user->country_acronym);
+                    });
+                break;
+            case 'friend':
+                $friends = array_map(function ($relation) {
+                    return $relation->zebra_id;
+                }, $user->friends->all());
```
nanaya: `lists('zebra_id')`

nanaya:
you need to update beatmaps menu links in `app/helpers` line 241.

LiquidPL:
found it at the same time

app/helpers.php
```diff
@@ -239,7 +238,7 @@ function nav_links()
         ];
     } else {
         $links['beatmaps'] = [
-            'getListing' => route('beatmaps'),
+            'getListing' => route('beatmapsets.index'),
```
nanaya: the key should also be updated.

resources/assets/less/bem/beatmapset-scoreboard.less
```diff
@@ -0,0 +1,139 @@
...
+    display: flex;
+    align-items: center;
+
+    height: 30px;
+
+    padding: 0 @spacing;
+
+    &--score {
+      &:hover {
+        background-color: #d1d2d3;
```
nanaya: Try the closest color variable currently available?

resources/assets/less/bem/beatmapset-success-rate.less
```diff
@@ -0,0 +1,80 @@
...
+*    osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+*    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+*    See the GNU Affero General Public License for more details.
+*
+*    You should have received a copy of the GNU Affero General Public License
+*    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+*
+*/
+
+@colour-retry: #fcde54;
```
nanaya: use variable for this as well. Also need to be scoped.

app/Http/Controllers/BeatmapController.php
```diff
@@ -21,106 +21,54 @@
...
-        }
-
-        $filters = ['data' => compact('modes', 'statuses', 'genres', 'languages', 'extras', 'ranks')];
-
-        return view('beatmaps.index', compact('filters', 'beatmaps'));
+        return ujs_redirect(route('beatmapsets.show', ['id' => $set->beatmapset_id]).'#'.$id);
     }
 
-    public function search()
+    public function getScores($id)
```
nanaya: just call it `scores` since it's not using the implicit controller routing.

app/Models/BeatmapFailtimes.php
```diff
@@ -0,0 +1,67 @@
...
+
+class BeatmapFailtimes extends Model
+{
+    protected $table = 'osu_beatmap_failtimes';
+
+    public $timestamps = false;
+
+    protected $casts = [
+        'beatmap_id' => 'integer',
+        'type' => 'string',
```
nanaya: string doesn't need casting

resources/assets/coffee/_classes/bar-chart.coffee
```diff
@@ -0,0 +1,97 @@
...
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+
+class @BarChart
+  transition: 1000
+
+  margins: {
```
nanaya: the brackets are not needed

app/Transformers/BeatmapSetTransformer.php
```diff
@@ -41,13 +43,35 @@ public function transform(BeatmapSet $beatmap = null)
             'artist' => $beatmap->artist,
             'play_count' => $beatmap->play_count,
             'favourite_count' => $beatmap->favourite_count,
+            'submitted' => $beatmap->submit_date->toIso8601String(),
+            'ranked' => $beatmap->approved_date->toIso8601String(),
```
nanaya: not all beatmapsets have `approved_date`.

app/Transformers/BeatmapSetTransformer.php
```diff
@@ -41,13 +43,35 @@ public function transform(BeatmapSet $beatmap = null)
             'artist' => $beatmap->artist,
             'play_count' => $beatmap->play_count,
             'favourite_count' => $beatmap->favourite_count,
+            'submitted' => $beatmap->submit_date->toIso8601String(),
```
nanaya: probably should be called `submitted_at` and `ranked_at` (or `_date`), especially the latter might be confused with a boolean if just `ranked`.

database/seeds/ModelSeeders/BeatmapSeeder.php
```diff
@@ -86,6 +87,8 @@ public function run()
                     $set->play_count = $set_playcount;
                     $set->favourite_count = $the_beatmap->favourite_count;
                     $set->user_id = array_rand_val($users)['user_id'];
+                    $set->approved_date = Carbon::now();
```
nanaya: may be better to randomly generate unapproved beatmapsets.

resources/assets/coffee/_classes/bar-chart.coffee
```diff
@@ -0,0 +1,96 @@
...
+  transition: 1000
+
+  margins:
+    top: 0
+    right: 0
+    bottom: 0
+    left: 0
+
+  constructor: (area, @options = {}) ->
+    @options.scales.x ||= d3.scale.linear()
```
nanaya: Looking again, this is wrong since it requires `scales` to be passed at all.

Something like this should work better (need check if actually works).

```coffeescript
options:
  scales:
    x: ...
    y: ...

constructor: (..., options = {}) ->
  _.merge @options, options
  ...
```

resources/assets/coffee/react/_components/beatmap-icon.coffee
```diff
@@ -23,19 +23,17 @@ el = React.createElement
...
     beatmap = @props.beatmap
 
-    difficultyRating = switch
-      when @props.overrideVersion? then @props.overrideVersion
-      when beatmap.difficulty_rating < 1.5 then 'easy'
-      when beatmap.difficulty_rating < 2.25 then 'normal'
-      when beatmap.difficulty_rating < 3.75 then 'hard'
-      when beatmap.difficulty_rating < 5.25 then 'insane'
-      else 'expert'
+    difficultyRating = if @props.overrideVersion? then @props.overrideVersion else DifficultyRating.get beatmap.difficulty_rating
```
nanaya: split it into multiple lines

resources/assets/coffee/react/_components/beatmap-icon.coffee
```diff
@@ -23,19 +23,17 @@ el = React.createElement
...
+
+    showTitle = @props.showTitle and !@props.overrideVersion?
```
nanaya: prefer `&&` and `||` to `and` and `or`

resources/assets/coffee/osu_common.coffee
```diff
@@ -18,6 +18,7 @@ along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
 @osu =
   isIos: /iPad|iPhone|iPod/.test(navigator.platform)
 
+  modes: ['osu', 'taiko', 'fruits', 'mania']
```
nanaya: create a `BeatmapHelper` class which contains this and the `DifficultyRating`?

resources/views/beatmapsets/show.blade.php
```diff
@@ -0,0 +1,40 @@
...
+        because the only library for it, which is experimental, requires PHP extension
+        which isn't supported by hhvm (v8js).
+    --}}
+@endsection
+
+@section("script")
+    @parent
+
+    <script data-turbolinks-eval="always">
+        var set = {!! json_encode($set['data']) !!};
```
nanaya: I noticed the previous method of embedding as json is better since it doesn't add random variable to global state. Check `resources/views/beatmaps/index.blade.php`

resources/assets/coffee/react/beatmapset-page/success-rate.coffee
```diff
@@ -0,0 +1,78 @@
...
+        scales: scales
+        className: 'beatmapset-success-rate'
+
+      @_successRateChart = new BarChart @refs.chartArea, options
+
+      $(window).on 'throttled-resize.beatmapSetPageSuccessRate', @_successRateChart.resize
+
+    @_successRateChart.loadData data
+
+  legend = ['retry', 'fail']
```
nanaya: why is it randomly here

LiquidPL:
should be all

resources/assets/coffee/_classes/bar-chart.coffee
```diff
@@ -0,0 +1,100 @@
...
+      bars
+        .enter()
+        .append 'rect'
+        .classed "#{@options.className}__bar #{@options.className}__bar--#{i}", true
+
+      bars
+        .transition @transition
+        .attr 'x', (d, i) => @options.scales.x i
+        .attr 'y', (d) => @options.scales.y d
+        .attr 'height', (d) => @height - @options.scales.y d
```
nanaya: height and y is flipped. Also for some reason this sometimes goes negative.

resources/assets/coffee/_classes/bar-chart.coffee
```diff
@@ -0,0 +1,100 @@
...
+      bars = @svgWrapper
+        .selectAll ".#{@options.className}__bar--#{i}"
+        .data data
+
+      bars
+        .enter()
+        .append 'rect'
+        .classed "#{@options.className}__bar #{@options.className}__bar--#{i}", true
+
+      bars
```
nanaya: the bar needs to be stacked on each other.

arflyte:
http://puu.sh/oDAHX/e00c8bfd39.png

Design updated, added success rate bar.

resources/assets/coffee/_classes/bar-chart.coffee
```diff
@@ -0,0 +1,115 @@
...
+class @BarChart
+  transition: 1000
+
+  margins:
+    top: 0
+    right: 0
+    bottom: 0
+    left: 0
+
+  options:
```
nanaya: as I learned today, apparently this shouldn't be done [1].

[1] (scroll down to warning section) https://coffeescript-cookbook.github.io/chapters/classes_and_objects/class-variables-and-instance-variables

LiquidPL:
done

resources/assets/coffee/_classes/bar.coffee
```diff
@@ -0,0 +1,40 @@
...
+#
+# osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+
+class @Bar
```
nanaya: I don't think this warrants using d3 ಠ_ಠ

resources/assets/coffee/react/beatmapset-page/success-rate.coffee
```diff
@@ -0,0 +1,104 @@
...
+
+    max = d3.max _.map data, (m) -> m[0] + m[1]
+
+    unless @_successRateChart
+      scales =
+        x: d3.scale.linear()
+        y: d3.scale.linear()
+
+      options =
+        domain: [0, max]
```
nanaya: this needs to be done on every props change otherwise the scale will stay at the initial data's `max` value.

resources/assets/coffee/_classes/bar.coffee
LiquidPL: Have anything else that can get fancy transitions like the rest of the graphs? The success rate bat looks quite dull when it just changes state while other graphs around animate.

nanaya: css transition

app/Transformers/BeatmapSetTransformer.php
```diff
@@ -41,13 +43,35 @@ public function transform(BeatmapSet $beatmap = null)
...
+    public function includeDescription(BeatmapSet $beatmapSet)
+    {
+        return $this->item($beatmapSet, function ($beatmapSet) {
+            return [
+                'description' => $beatmapSet->description(),
+            ];
+        });
+    }
+
+    public function includeAuthor(BeatmapSet $beatmapSet)
```
nanaya: see if whatever is using `user` include can be replaced with the `UserCompactTransformer` (also it's weird having `author` and `user` includes which points to more or less same thing).

app/Http/routes.php
```diff
@@ -47,23 +47,19 @@
...
+
+Route::resource('/beatmaps', 'BeatmapController', ['only' => ['show']]);
+Route::get('/beatmaps/{id}/scores', ['as' => 'beatmap.scores', 'uses' => 'BeatmapController@scores']);
+
+Route::get('/beatmapsets/search/{filters?}', ['as' => 'beatmapsets.search', 'uses' => 'BeatmapsetsController@search']);
+Route::resource('/beatmapsets', 'BeatmapsetsController', ['only' => ['index', 'show']]);
+
+Route::get('/b/{id}', 'BeatmapController@show');
 
 Route::get('/s/{id}', ['as' => 'beatmap-sets.show', function ($id) {
```
nanaya: looking again, this probably should be just be used for actual page url. See [this section](/ppy/osu-web/pull/489/files#diff-d7930c8d8e76bf6cae7b681d1958ae6bR125).

resources/assets/coffee/react/beatmapset-page/stats.coffee
```diff
@@ -0,0 +1,63 @@
...
+
+  render: ->
+    div className: 'page-contents__content',
+      div className: 'page-contents__row',
+        datas.map (m) =>
+          dt = Lang.get "beatmaps.beatmapset.show.stats.#{m}"
+
+          switch m
+            when 'cs'
+              dd = "#{@props.beatmap.cs.toLocaleString()}/10"
```
nanaya: apparently the `/10` part isn't needed.

resources/assets/coffee/react/beatmapset-page/stats.coffee
```diff
@@ -0,0 +1,63 @@
...
+            when 'hp'
+              dd = "#{@props.beatmap.drain.toLocaleString()}/10"
+            when 'od'
+              dd = "#{@props.beatmap.accuracy.toLocaleString()}/10"
+            when 'ar'
+              dd = "#{@props.beatmap.ar.toLocaleString()}/10"
+            when 'stars'
+              dd = "#{@props.beatmap.difficulty_rating.toLocaleString()}/10"
```
nanaya: change it to only show two digits after decimal point (`toFIxed(2)`).

app/Http/routes.php
LiquidPL: as in with or without keeping the `/beatmapsets/{id}` route?

nanaya: without keeping.

app/Transformers/BeatmapSetTransformer.php
LiquidPL: apparently the only place it is used is `BeatmapsetsController@discussion`, and it's not even used in beatmap discussions coffeescript anyway

LiquidPL:
Apparently the way we've been setting transitions in d3 doesn't work, and everything (at least the profile page chart) is using the default 250ms. I've resorted to using that in CSS too for now. Should I keep it that way or go with 1000ms anyway?

nanaya:
[maybe if you pass the variable to the function...?](/ppy/osu-web/blob/54e8e18b489c3566312faf51c64e5d858b9a7b88/resources/assets/coffee/_classes/line-chart.coffee#L170)

LiquidPL:
[actually, i was doing so](https://github.com/LiquidPL/osu-web/blob/cd0424110c204be01818bb2624771cf750608f38/resources/assets/coffee/_classes/bar-chart.coffee#L96), just removed that for now

d3js with `transition(1000)` and CSS with `transition: width 1s`: http://puu.sh/oFZzG.ogv

LiquidPL:
Apparently to set the transition duration you need to call `selection.transition().duration 1000` instead of `selection.transition 1000`. However I feel that 1 second might be slightly [too long](http://puu.sh/oGda8.ogv). Any thoughts on that?

nanaya:
I see. The default is fine, I guess.

resources/assets/coffee/react/_mixins/switchable-mode-page.coffee
```diff
@@ -0,0 +1,79 @@
...
+      @setCurrentPage null, page
+      return
+
+    target = $(".js-switchable-mode-page--page[data-page-id='#{page}']")
+
+    return unless target.length
+    # Don't bother scanning the current position.
+    # The result will be wrong when target page is too short anyway.
+    @scrolling = true
+    clearTimeout @timeouts.scrolling
```
nanaya: just call it `@modeScrollTimeout` and add `componentWillUnmount` to clear the timeout in this mixin so it's self-contained.

resources/assets/coffee/react/beatmapset-page/contents.coffee
```diff
@@ -0,0 +1,63 @@
...
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+{div, p} = React.DOM
+el = React.createElement
+
+class BeatmapsetPage.Contents extends React.Component
+  @modeSwitch: (e) =>
```
nanaya: not used?

app/Http/Controllers/BeatmapsetsController.php
```diff
@@ -19,15 +19,130 @@
...
+        }
+
+        $filters = ['data' => compact('modes', 'statuses', 'genres', 'languages', 'extras', 'ranks')];
+
+        return view('beatmaps.index', compact('filters', 'beatmaps'));
+    }
+
+    public function show($id)
+    {
+        $beatmapSet = BeatmapSet::find($id);
```
nanaya: `findOrFail`

resources/assets/coffee/react/_mixins/switchable-mode-page.coffee
LiquidPL: but then I still need to call `removeListeners` in components `componentWillUnmount` (at least in those that are currently using the mixin). Wouldn't those methods conflict in this case?

nanaya: mixin handles it.

resources/assets/coffee/react/beatmapset-page/contents-tab.coffee
```diff
@@ -0,0 +1,42 @@
...
+# along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+{a} = React.DOM
+
+class BeatmapsetPage.ContentsTab extends React.Component
+  onClick: (e) =>
+    return if @props.disabled
+    e.preventDefault()
+    $.publish 'beatmapset:playmode:set', @props.playmode
+    $.publish 'beatmapset:mode:set', @props.newMode
```
nanaya: this should say `beatmapset:beatmap:set`, and since it already infers the playmode, the `playmode` event above isn't needed. And fix variable name (something like `targetBeatmapId`).

resources/assets/coffee/react/_mixins/switchable-mode-page.coffee
```diff
@@ -0,0 +1,79 @@
...
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+
+pages = document.getElementsByClassName("js-switchable-mode-page--scrollspy")
+pagesOffset = document.getElementsByClassName("js-switchable-mode-page--scrollspy-offset")
+
+@SwitchableModeMixin =
```
nanaya: the only `mode` about this mixin is `setCurrentMode`. The rest don't really have anything to do with mode.

resources/assets/coffee/react/beatmapset-page/success-rate.coffee
```diff
@@ -0,0 +1,86 @@
...
+    @_renderChart()
+
+  componentWillUnmount: ->
+    $(window).off '.beatmapSetPageSuccessRate'
+
+  _renderChart: ->
+    return unless @props.failtimes.length > 0
+
+    failtimes = _.map @props.failtimes, (m) -> m.data
+    data = _.zip failtimes[0], failtimes[1]
```
nanaya: nothing in here indicates which one is `fail` and which one is `retry`?

resources/assets/coffee/react/beatmapset-page/success-rate.coffee
```diff
@@ -0,0 +1,86 @@
...
+        className: 'beatmapset-success-rate'
+
+      @_successRateChart = new BarChart @refs.chart, options
+
+      $(window).on 'throttled-resize.beatmapSetPageSuccessRate', @_successRateChart.resize
+
+    @_successRateChart.loadData data
+
+  render: ->
+    percentage = _.round ((@props.beatmap.playcount - @props.beatmap.passcount) / @props.beatmap.playcount) * 100
```
nanaya: slightly shorter: `((1 - @props.beatmap.passcount / @props.beatmap.playcount) * 100).toFixed()`

nanaya: huh, the number is `failcount`, not `passcount`. Success percentage is just `passcount/playcount*100`

resources/assets/coffee/react/profile-page/info.coffee
```diff
@@ -35,64 +35,61 @@ class ProfilePage.Info extends React.Component
 
     return keys
 
+  datas = ['twitter', 'skype', 'lastfm', 'playstyles']
```
nanaya: why is it located here and not inside `render`? Or even directly used for the map instead through a variable `['a', ...].map`.

resources/assets/coffee/react/profile-page/info.coffee
```diff
@@ -35,64 +35,61 @@ class ProfilePage.Info extends React.Component
...
+          when 'skype'
+            return unless @props.user.skype
+            dt = 'Skype'
+            dd = el 'a', href: "skype:#{@props.user.skype}?chat", @props.user.skype
+          when 'lastfm'
+            return unless @props.user.lastfm
+            dt = 'Last.fm'
+            dd = el 'a', href: "https://last.fm/user/#{@props.user.lastfm}", @props.user.lastfm
+          when 'playstyles'
+            return unless @props.user.playstyle.length
```
nanaya: updating the getter method in the model to return `null` when it's empty will simplify check for this part (`return if !@props.user[m]?`

resources/assets/coffee/react/profile-page/recent-achievements.coffee
```diff
@@ -30,23 +30,25 @@ class ProfilePage.RecentAchievements extends React.Component
...
+    div className: 'page-contents__content profile-achievements text-center',
+      div className: 'page-contents__row page-contents__row--top',
+        div className: 'profile-badge profile-badge--achievements',
+          span className: 'profile-badge__number',
             @props.achievementsCounts.current
 
         div className: 'profile-exp-bar',
           div
-            className: 'profile-exp-bar-fill'
+            className: 'profile-exp-bar--fill'
```
nanaya: `__fill`

resources/assets/less/bem/beatmapset-details.less
```diff
@@ -0,0 +1,111 @@
...
+
+.beatmapset-details {
+  &__avatar-container {
+    width: @beatmapset-avatar-container-size;
+    height: @beatmapset-avatar-container-size;
+
+    margin-left: auto;
+    margin-right: auto;
+
+    border-radius: 5px;
```
nanaya: not exactly 5px but there's `.default-border-radius()`.

resources/assets/less/bem/beatmapset-scoreboard-first.less
```diff
@@ -0,0 +1,141 @@
...
+    padding-bottom: @spacing * 0.5;
+
+    &--rank {
+      font-size: 24px;
+      color: @yellow-dark;
+    }
+
+    &--username {
+      font-size: 20px;
+      color: #1188aa;
```
nanaya: `@purple-darker`?

resources/assets/less/bem/beatmapset-panel.less
```diff
@@ -1,23 +1,160 @@
...
+    border: none;
+
+    width: 100%;
+    height: 100px;
+
+    .at2x('/images/layout/beatmaps/default-bg.png');
+    background-position: center !important;
+    background-size: cover !important;
+
+    text-shadow: 0px 1px 2px fade(#000, 100%);
```
nanaya: `.default-text-shadow()` is close enough?

resources/assets/less/bem/beatmapset-panel.less
```diff
@@ -1,23 +1,160 @@
...
+    border-bottom-left-radius: 0px;
+    border-bottom-right-radius: 0px;
+
+    border: none;
+
+    width: 100%;
+    height: 100px;
+
+    .at2x('/images/layout/beatmaps/default-bg.png');
+    background-position: center !important;
```
nanaya: what's `!important` doing in here and various other places.

nanaya: oh it's like this originally (ノﾟ ヮﾟ)ノ︵ ┻━┻

LiquidPL: Yea, I was just moving them around

resources/assets/coffee/react/beatmapset-page/success-rate.coffee
LiquidPL: technically the transfomer sorts them so that `fail` always comes first, should I actually grab them by type in coffeescript or just note that in comments or something?

resources/assets/less/bem/beatmapset-panel.less
LiquidPL: should i actually clean them up, or leave them for now?

nanaya: leaving them is fine

resources/assets/coffee/react/beatmapset-page/success-rate.coffee
nanaya: with just two rows, I think it's better to sort them here. Even better if you can somehow infer the class names using `fail` and `retry` instead of `0` and `1`.

LiquidPL:
this would be all then

app/Models/User.php
```diff
@@ -321,6 +321,10 @@ public function getOsuPlaystyleAttribute($value)
             }
         }
 
+        if (empty($array)) {
```
nanaya: looks like wrong variable name

resources/assets/coffee/react/beatmapset-page/success-rate.coffee
```diff
@@ -0,0 +1,97 @@
...
+
+  componentWillUnmount: ->
+    $(window).off '.beatmapSetPageSuccessRate'
+
+  _renderChart: ->
+    return unless @props.failtimes.length > 0
+
+    # sorting the failtimes array
+    # so that fail always comes before exit (database values)
+    failtimes = _.sortBy @props.failtimes, (m) -> m.type
```
nanaya: why not `keyBy`?

resources/assets/less/colours.less
```diff
@@ -49,6 +49,12 @@
 @green-dark: #668800;
 @green-darker: #445500;
 
+@diff-easy: #aadd00;
```
nanaya: why not use `@green`, `@yellow`, `@pink`, and `@purple`...?

nanaya: also, this belongs in `variables.less`.

resources/assets/coffee/react/beatmapset-page/main.coffee
```diff
@@ -0,0 +1,119 @@
...
+  mixins: [ScrollingPageMixin]
+
+  getInitialState: ->
+    optionsHash = BeatmapsetPageHash.parse location.hash
+    @initialPage = optionsHash.page
+
+    currentMode = if optionsHash.mode then parseInt optionsHash.mode, 10 else @props.displayedBeatmap
+
+    currentMode: currentMode
+    currentPlaymode: @props.beatmaps[currentMode].mode
```
nanaya: missing check to ensure `currentMode` is valid.

(and why is it called `currentMode` when it actually is `currentBeatmapId`?)

resources/assets/coffee/react/beatmapset-page.coffee
```diff
@@ -0,0 +1,30 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+
+set = JSON.parse(document.getElementById('json-beatmapset').text).data
+countries = JSON.parse(document.getElementById('json-countries').text).data
+
+propsFunction = =>
+  set: set
+  beatmaps: _.keyBy set.beatmaps.data, (o) -> o.id
```
nanaya: shouldn't most of props here be just part of state in `Main`?

resources/assets/coffee/react/beatmapset-page.coffee
```diff
@@ -0,0 +1,30 @@
...
+#
+# osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+
+set = JSON.parse(document.getElementById('json-beatmapset').text).data
```
nanaya: this and following line should be inside `propsFunction`

resources/assets/coffee/react/beatmapset-page.coffee
```diff
@@ -0,0 +1,30 @@
...
+countries = JSON.parse(document.getElementById('json-countries').text).data
+
+propsFunction = =>
+  set: set
+  beatmaps: _.keyBy set.beatmaps.data, (o) -> o.id
+  beatmapsByMode: _.groupBy set.beatmaps.data, (o) -> o.mode
+  beatmapCount: _.countBy set.beatmaps.data, (o) -> o.mode
```
nanaya: Can't it just use `beatmapsByMode[mode].length`?

nanaya: or `beatmapsByMode[mode]?` since it'll be undefined if no relevant mode.

resources/assets/coffee/react/_mixins/scrolling-page.coffee
```diff
@@ -0,0 +1,82 @@
...
+###
+
+pages = document.getElementsByClassName("js-switchable-mode-page--scrollspy")
+pagesOffset = document.getElementsByClassName("js-switchable-mode-page--scrollspy-offset")
+
+@ScrollingPageMixin =
+  componentWillUnmount: ->
+    clearTimeout @modeScrollTimeout
+
+  setCurrentMode: (_e, mode) ->
```
nanaya: this should be removed from this mixin. Current `master` branch added additional stuff for this function in `ProfilePage.Main` as well, making abstracting it much less useful.

If anything, something like `SetOneStateMixin` mixin is more useful.

```coffeescript
# rough idea, has not been actually tested
@SetOneStateMixin =
  setOneState: (key, newValue, callback) ->
    return callback?() if @state[key] == newValue

    @setState "#{key}": newValue, callback
```

resources/assets/coffee/react/beatmapset-page/scoreboard-item.coffee
```diff
@@ -0,0 +1,48 @@
...
+                href: laroute.route 'users.show', users: @props.score.user.data.id
+                @props.score.user.data.username
+            when 'icons'
+              div className: "badge-rank badge-rank--#{@props.score.rank}"
+            when 'score'
+              @props.score.score.toLocaleString()
+            when 'accuracy'
+              "#{_.round @props.score.accuracy * 100, 2}%"
+
+        span className: className, key: m, contents
```
nanaya: should use `div` instead of `span`.

resources/assets/coffee/react/profile-page/main.coffee
```diff
@@ -142,19 +71,16 @@ class ProfilePage.Main extends React.Component
     @pageJump null, @initialPage
 
 
-  componentWillUnmount: =>
-    for own _name, timeout of @timeouts
-      clearTimeout timeout
-
+  componentWillUnmount: ->
     $(window).stop()
```
nanaya: this should also be moved to the mixin

resources/assets/coffee/react/beatmapset-page/details.coffee
```diff
@@ -0,0 +1,91 @@
...
+  render: ->
+    div className: 'page-contents__content beatmapset-details',
+      div className: 'beatmapset-details__avatar-container',
+        div
+          className: 'beatmapset-details__avatar avatar avatar--beatmapset'
+          style:
+            backgroundImage: "url(#{@props.set.author.data.avatarUrl})"
+
+      div className: 'beatmapset-details__mapper',
+        Lang.get 'beatmaps.beatmapset.show.details.made-by'
```
nanaya: username should be part of the translation. And the username should be a link. Check the one in `resources/assets/coffee/react/beatmap-discussions/system-post.coffee`.

app/Transformers/BeatmapSetTransformer.php
```diff
@@ -25,8 +25,9 @@
 class BeatmapSetTransformer extends Fractal\TransformerAbstract
 {
     protected $availableIncludes = [
+        'description',
+        'author',
```
nanaya: should follow the relation name (`user`)?

resources/assets/coffee/react/beatmapset-page/contents.coffee
```diff
@@ -0,0 +1,60 @@
...
+              beatmap: beatmap
+              currentPage: @props.currentPage
+              currentBeatmapId: @props.currentBeatmapId
+
+        p className: 'beatmapset-difficulties__name',
+          Lang.get("beatmaps.mode.#{@props.currentPlaymode}") + " #{beatmap.version}"
+
+      div className: 'page-contents',
+        el BeatmapsetPage.Details,
+          set: @props.set
```
nanaya: is it not confusing calling a variable `set`?

app/Models/BeatmapSet.php
```diff
@@ -684,4 +686,18 @@ public function approver()
     {
         return $this->belongsTo("App\Models\User", 'user_id', 'approvedby_id');
     }
+
+    public function description()
+    {
+        $topic = Topic::find($this->thread_id);
+        $post = Post::find($topic->topic_first_post_id);
+
+        $split = preg_split('[-{15}]', $post->post_text);
```
nanaya: just noticed, hopefully there's no description which contains another `---------------` otherwise anything after that will be gone (since they'll be part of `$split[2]` and up)...

nanaya:
looks like history navigation is broken when navigating between user profile and beatmapset page :confused:

resources/assets/coffee/react/profile-page/main.coffee
```diff
@@ -18,119 +18,48 @@
...
-      currentMode: @validMode(optionsHash.mode ? props.user.playmode)
-      user: props.user
-      userPage:
-        html: props.userPage.html
-        initialRaw: props.userPage.raw
-        raw: props.userPage.raw
-        editing: false
-        selection: [0, 0]
-      isCoverUpdating: false
+    currentMode: @validMode(optionsHash.mode ? props.user.playmode)
```
nanaya: apparently I forgot the `@` sign.

resources/assets/coffee/react/beatmapset-page/main.coffee
```diff
@@ -0,0 +1,127 @@
...
+      .fail (xhr) =>
+        osu.ajaxError xhr
+      .always =>
+        $.publish 'beatmapset:scoreboard:loading', false
+        @setState loading: false
+
+  setCurrentBeatmapId: (_e, beatmapId) ->
+    return if @state.currentBeatmapId == beatmapId
+
+    @setState
```
nanaya: missing call to `setHash`

resources/assets/coffee/react/beatmapset-page/main.coffee
```diff
@@ -0,0 +1,127 @@
...
+      currentScoreboard: 'global'
+      scores: @state.beatmaps[beatmapId].scoresBest.data
+
+  componentDidMount: ->
+    @removeListeners()
+
+    $.subscribe 'beatmapset:beatmap:set.beatmapSetPage', @setCurrentBeatmapId
+    $.subscribe 'beatmapset:page:jump.beatmapSetPage', @pageJump
+    $.subscribe 'beatmapset:scoreboard:set.beatmapSetPage', @setCurrentScoreboard
+    $(window).on 'throttled-scroll.beatmapSetPage', @pageScan
```
nanaya: should be part of mixin

nanaya: don't forget to remove the other one as well

LiquidPL:
on what browser does this happen? seems to be firefox only for me

nanaya:
on chrome, the hash from previous page carries over until `pageScan` is ran.

Logging within `osu.setHash` shows `replacing state... https://osunext.myconan.net/s/181142#osu/top_ranks`

nanaya:
looks like it's unfortunate timing during navigation

1. browser navigation is fired &rarr; url is changed
2. page scroll (from browser scroll history?) is triggered, triggers `pageScan`
3. at this point, the previous page is still there, `pageScan` from previous page is done, replacing hash for previous page (with new url)
4. actual new page is loaded

nanaya:
...and that's for chrome, no idea about firefox yet

nanaya:
looks like [it's bug for firefox for this version of turbolinks](/turbolinks/turbolinks-classic/issues/573). In the meantime, adding this to `ScrollingPageMixin` helps the chrome hash problem

```coffeescript
currentLocation = =>
  "#{document.location.pathname}#{document.location.search}"

@ScrollingPageMixin =
  componentDidMount: ->
    @modeScrollUrl = currentLocation()
    ...the rest

  pageScan: ->
    return if @modeScrollUrl != currentLocation()
    ...the rest
```

LiquidPL:
Actually what happens in firefox is slightly different.

Let's put it that way:
1. Open a link leading to a beatmapset which **already has a hash** (eg. /s/325307#759638).
2. Click the link leading to your own user profile.
3. Click back in the browser, the beatmapset page loads, all is good.
4. Click forward in the browser. At this point, the user profile history entry is duplicated, with one being named appropriately, and the other having the name of the beatmapset page.

Also, when pressing back in 3, it seems that the page loads without turbolinks, as opposed to other steps.

nanaya:
Another interesting one:

0. new tab
1. open beatmapset page without hash
2. click on user page link
3. go back
4. history is now swapped with back history contains profile page instead

LiquidPL:
dunno, maybe we should change the hashes in a way that does not manipulate history? I don't think the hashes should be stored in the actual history, just the browsers URL bar.

LiquidPL:
Actually, after some googling I have found a `location.replace("#hash")` function.

After changing `osu.setHash` to this:

    setHash: (newHash) ->
      return if newHash == location.hash
  
      location.replace newHash

the bugs seem to have stopped, and it actually keeps the behavior of having the hash in history.

nanaya:
yeah, seems fine

nanaya:
Apparently it's not completely fixed :D 

1. new tab
2. open beatmapset page without hash
3. click on user page link
4. go back
5. click on user page link
6. go back
7. still on user page instead

LiquidPL:
i feel that turbolinks might be pushing an additional history entry, because when only when this happens I see the loading overlay (i guess, maybe it just loads that fast)

nanaya:
I tried updating to latest turbolinks* and it seems to be fixed... except the part I have to revert the `setHash` back to using `replaceState` since `state` is emptied when using `location.replace`. The structure is changed anyway though so I'll have to update the function regardless later.

LiquidPL:
so what should I do now? revert `osu.setHash` to what is was previously?

app/Models/BeatmapSet.php
```diff
@@ -684,4 +686,21 @@ public function approver()
...
+
+    public function description()
+    {
+        $topic = Topic::find($this->thread_id);
+        $post = Post::find($topic->topic_first_post_id);
+
+        // Any description (after the first match) that matches
+        // '[-{15}]' within its body doesn't get split anymore,
+        // and gets stored in $split[1] anyways
+        $split = preg_split('[-{15}]', $post->post_text, 2);
```
nanaya: It's actually my first time seeing `[...]` used as regexp delimiter...

PR #500, opened 2016-05-11
User page fixes

nanaya:
Something was broken on both frontend and backend =D

And misc cleanups.


peppy:
travis looks dead; merging anyway.

PR #506, opened 2016-05-13
Improved US and Liberia flags

Rocker-r:
Made them look more like the actual flags


peppy:
Check out the new versions we've put up (@arflyte is kinda picky) and if you're happy this one should be good to close!

PR #508, opened 2016-05-13
Update some flags

nanaya:
More stars and... less tofu.

And one... I don't even.

Includes fixes #507 of adding Puerto Rico flag.


peppy:
oops i conflicted this by merging #507 first..

PR #523, opened 2016-05-13
Audio preview on beatmapset pages

### LINKED ISSUE #512: Clicking the header on a beatmap page should allow playback of audio
In a similar fashion to the old site. There should be a play and pause state icon displayed. A design can be provided but you're welcome to go ahead and add functionality and we can work the design in place after.

should be a bounty of 50 or so. maybe more if design is good.

LiquidPL:
Fixes #512.

Currently the design is simply a play/pause button shown on hover of the header.

### DIFF
--- a/app/Models/Beatmapset.php
+++ b/app/Models/Beatmapset.php
@@ -536,6 +536,11 @@ class Beatmapset extends Model
         $this->storage()->put($this->coverPath().$target_filename, file_get_contents($source_path));
     }
 
+    public function previewURL()
+    {
+        return 'http://b.ppy.sh/preview/'.$this->beatmapset_id.'.mp3';
+    }
+
     public function storage()
     {
         if ($this->_storage === null) {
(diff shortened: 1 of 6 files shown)


app/Models/BeatmapSet.php
```diff
@@ -536,6 +536,11 @@ public function storeCover($target_filename, $source_path)
         $this->storage()->put($this->coverPath().$target_filename, file_get_contents($source_path));
     }
 
+    public function previewURL()
+    {
+        return 'http://b.ppy.sh/preview/'.$this->beatmapset_id.'.mp3';
```
AndyR207: Would it not be better to use

    return '//b.ppy.sh/preview/'.$this->beatmapset_id.'.mp3';

As forcing the content to load as HTTP will throw a content security warning when browsing with HTTPS.

peppy:
The hit area for the play/pause button should probably be the majority of the header cover image (minus the parts with selectable text. At  the moment it's hard to even know that this feature exists.

peppy:
The beatmap details page design is changing so I'm going to merge this is "functionally complete".

PR #525, opened 2016-05-13
Fix achievements in recent activities not displaing dates

### LINKED ISSUE #524: Wrong skill medal image used in event feed
Currently, the medals for "*n* star pass" and "*n* star FC" are always shown in the greyed out version in the user feed.

It is probably better to show those medals in the colour they have when unlocked, i.e. in silver and gold.

LiquidPL:
well, generally displaying as not achieved ¯\\\_(ツ)_/¯
Fixes #524.


nanaya:
Can use a rebase.

PR #528, opened 2016-05-16
Team Pages

kschamplin19:
What is done: 
- [x] The page renders and passes a Team
- [x] established since shows correctly based on Laravel "created_at" (may need changing depending on preference)
- [x] Team members page allows for viewing of team members (duh)
- [x] If the user is in the admins panel, allow them to edit permissions and add members.
- [x] Prevent the admin from setting themselves to member (no admin-less groups)

What needs to be done:
- [x] Saving the changes. Hopefully done in the next couple days.
- [ ] Info, stats, achievements. Requires details on preferred model layout.
- [ ] At some point, index of teams that are "public-join" if wanted
- [x] Add user dialog. Need to take a closer look at the LESS to see what we already have.
- [ ] "Creator" role with destructive powers (nuke the group). Cannot be removed from power.

What needs to be FIXED (bugs):
- [x] The selector dropping is finicky, esp when putting it before the add user option.
- [x] I probably screwed up the BEM somehow and broke the standard, yet left it in the BEM folder :/
- [ ] Changing tabs causes the banner to disappear (maybe because its temporarily hardcoded)


peppy:
Did anyone ever review this? There's a huge amount of effort put in here but it looks like it has been largely ignored.

notbakaneko:
Has this been abandoned? Closing until there are updates

PR #540, opened 2016-05-23
Display mods on beatmap scoreboards

### LINKED ISSUE #510: Mods are missing from score displays on beatmap page.
They should be displayed similar to how they are on the profile listing.

bounty: 10-20 (make sure the code is shared with profile score displays where possible).

LiquidPL:
Fixes #510.


resources/assets/coffee/react/beatmapset-page/scoreboard-item.coffee
```diff
@@ -15,12 +15,12 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
 ###
-{div, a} = React.DOM
+{div, a, img} = React.DOM
```
nanaya: the img isn't used

PR #544, opened 2016-05-27
Fix filename case

nekodex:
`BeatmapSetTransformer` -> `BeatmapsetTransformer`


peppy:
:octopus:

nanaya:
Pretty sure there are more places using `BeatmapSetTransformer` and thus need renaming as well <_<

PR #547, opened 2016-05-27
Fix message ordering in API

nekodex:


nanaya:
might as well merge the two functions.

nekodex:
Yeah p. much, but want to get it in place for @peppy to work with over the weekend

nanaya:
Basic thing is there anyway so fix it here? `¯\_(ツ)_/¯`

PR #551, opened 2016-05-31
New authorization (and more)

nanaya:
- single entry point at `App\Libraries\Authorize`
- `ensureCan()` allows simple immediate request termination
- the helper can probably use a better name than `authz`
- automated integer/float casting relying on `mysqlnd` driver
- user variable assignment from oauth2 using Laravel built-in `Auth`
- use actual forum acl (partial)
- I have a feeling this will break some model seeds/factories


peppy:
seems solid :+1:

PR #553, opened 2016-05-31
Don't open beatmaps in new tab/window

### LINKED ISSUE #552: Beatmap links shouldn't open in a new window/tab
Right now they do, when clicked from the listing page.

LiquidPL:
Fixes #552.


peppy:
Pending on beatmap listing actually remembering a search when returning to the page.

PR #557, opened 2016-06-01
Beatmapset audio-preview overlay fixes and tweaks

nekodex:
Fixes audio-preview overlay on the beatmapsets header making the title/artist links unclickable
Also tweaks the visual appearance a bit to always be visible and rewinds the preview when stopped


LiquidPL:
if we're having the audio rewind to the beggining when stopped, then maybe let's change the pause icon to a stop icon?

PR #566, opened 2016-06-02
Laravel 5.2

nanaya:
And a bunch of other updates.


app/Http/Controllers/Controller.php
```diff
@@ -19,13 +19,15 @@
  */
 namespace App\Http\Controllers;
 
-use Illuminate\Foundation\Bus\DispatchesCommands;
+// use Illuminate\Foundation\Auth\Access\AuthorizesRequests;
+// use Illuminate\Foundation\Auth\Access\AuthorizesResources;
```
nekodex: 😕

composer.json
```diff
@@ -2,36 +2,43 @@
   "name": "ppy/osu-web",
   "description": "osu!web",
   "type": "project",
-  "homepage": "https://jizz.ppy.sh",
+  "homepage": "https://new.ppy.sh",
   "license": "AGPL-3.0+",
   "support": {
     "email": "jizz@ppy.sh",
```
nekodex: 👍

nanaya: shall I replace this with `noreply@ppy.sh`?

peppy: support@ppy.sh is fine

PR #568, opened 2016-06-03
Optimise on-hover performance on beatmapsets listing page

nekodex:
Minimise repaints and composite instead so we get tasty GPU acceleration.

### DIFF
--- a/resources/assets/coffee/react/_components/beatmapset-panel.coffee
+++ b/resources/assets/coffee/react/_components/beatmapset-panel.coffee
@@ -21,11 +21,6 @@
 el = React.createElement
 
 class @BeatmapsetPanel extends React.Component
-  mixins: [React.addons.PureRenderMixin]
-
-  shouldComponentUpdate: (_nextProps, _nextState) ->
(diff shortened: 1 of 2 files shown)


resources/assets/coffee/react/_components/beatmapset-panel.coffee
```diff
@@ -21,6 +21,11 @@
 el = React.createElement
 
 class @BeatmapsetPanel extends React.Component
+  mixins: [React.addons.PureRenderMixin]
```
nanaya: this doesn't do anything because
- you need to use `React.createClass`
- you overrode the only thing it does anyway (`shouldComponentUpdate`)

:expressionless:

resources/assets/less/bem/beatmapset-panel.less
```diff
@@ -17,26 +17,40 @@
 *
 */
 
+.beatmapset-panel-overlay {
```
nanaya: don't put this here (屮゜Д゜)屮

PR #572, opened 2016-06-04
Add default parameter

nanaya:
The syntax looks a bit weird ¯\\\_(ツ)_/¯


peppy:
that syntax is uh... am i supposed to just smile and nod and click merge

Issue #560, opened 2016-06-01
Live Stream Page, Redesigned.


arflyte:
![](http://puu.sh/pcBt1/337ac1c5d2.jpg)
Image snippet at 2x. Please ignore the old style logo and header.

![](http://puu.sh/pcBuq/0c5e4ca47b.jpg)
It would be great if the featured livestream is an actual video stream instead of a screenshot.

Twitch username, followed with stream title and viewer count. All the texts are on top of the viewer. Black gradient 0%-50% opacity background on stream details. Snippet at 1x.

![](http://puu.sh/pcAAq/d313ad6aa3.jpg)
Other livestreams. Only the twitch username and viewer count are displayed below their thumbnail. Sorted by viewer count. Big drop shadow on hover. Snippet at 1x.

peppy:
**[Bounty](https://goo.gl/nFdoyI) offered**: 50-150 depending on implementation quality.

Feel free to reply in this thread should you have any questions.

RBRat3:
Would it be a nice courtesy to add the mute parameter to true if twitches embed is going to be used?

odihominem:
Hi everyone,
I have a working draft of redesigned live stream page. I hope it'll meet your expectations. Background for the featured livestream is a video.

![osu_livestreams_redesigned](https://cloud.githubusercontent.com/assets/15053832/15807004/ee45f882-2b51-11e6-8918-17296e465879.png)
![osu_livestreams_drop_shadow](https://cloud.githubusercontent.com/assets/15053832/15807042/cffad7d4-2b52-11e6-8605-dcddb2b3ff02.jpg)

peppy:
looks pretty solid. send over a pull request and we'll check it out :+1:

PR #575, opened 2016-06-05
Live Page Redesigned

odihominem:
#560


RedbackThomson:
Stream name search in the header bar?

peppy:
@odihominem thanks for the PR! It turns out your design missed the fact that the original mock-up image with @2x (double the size of the actual dimensions) which made it pretty incorrect.

You can check out @nanaya's implementation which addresses a few other features at the same time here: https://github.com/ppy/osu-web/pull/580

I'm going to award your a token bounty of 10 for your effort.

PR #586, opened 2016-06-08
Misc fixes

nanaya:
Now works with Windows. Probably should update the readme.


app/Http/Controllers/API/Controller.php
```diff
@@ -24,13 +24,18 @@
...
 {
     use DispatchesJobs, ValidatesRequests;
 
     public function __construct()
     {
-        Auth::onceUsingId(Authorizer::getResourceOwnerId());
+        try {
+            Auth::onceUsingId(Authorizer::getResourceOwnerId());
+        } catch (NoActiveAccessTokenException $_e) {
+            //
```
nekodex: Maybe add to this comment explaining why this is needed 🍄

PR #588, opened 2016-06-08
Multiplayer match history

LiquidPL:
Things done:

* loads new events every 10 seconds
* displays only 100 latest events by default, can show more in chunks of 100 (can change if too big)

Things to be done:

* we need to generate a separate cover size for the game headers (1090x100 at @1x at full page width)
* performance seems meh while loading further chunks of events (400+)


nekodex:
Regarding the game headers, use the `cover` sized images and have it `object-fit: cover` (to fill the header) for now

app/Libraries/ModsFromDB.php
```diff
@@ -0,0 +1,84 @@
...
+ *    osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+ *    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *    See the GNU Affero General Public License for more details.
+ *
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ */
+namespace App\Libraries;
+
+class ModsFromDB
```
nekodex: Instead of a Library, I'd move this into a model (to eventually get it into the Database)

LiquidPL: if i understand correctly, a separate model for mods?

app/Http/routes.php
```diff
@@ -79,6 +79,9 @@
 Route::get('/community/slack', ['as' => 'slack', 'uses' => 'CommunityController@getSlack']);
 Route::post('/community/slack/agree', ['as' => 'slack.agree', 'uses' => 'CommunityController@postSlackAgree']);
 
+Route::get('/mp/{match_id}', ['as' => 'multiplayer.match', 'uses' => 'MultiplayerController@match']);
+Route::get('/mp/{match_id}/history', ['as' => 'multiplayer.match.history', 'uses' => 'MultiplayerController@matchHistory']);
```
nekodex: Maybe make these `/matches/...` and have `/mp/...` redirect there?

LiquidPL: i had something like that with beatmaps earlier (like `/beatmaps/{id}` and `/beatmapsets/{id}` redirecting into `/b/` and `/s/`), but eventually we just decided to go with the shorter ones, so i dunno

nekodex: Ideally we want the full URLs to be canonical and the shorter ones redirecting to them, for example:
`/b/{id}` redirecting to `/beatmaps/{id}`
`/s/{id}` redirecting to `/beatmapsets/{id}`
`/mp/{id}` redirecting to `/matches/{id}`

(I confirmed this with peppy just to be sure)

app/Libraries/ModsFromDB.php
nekodex: Yeah 👍

nekodex:
There's a... slight... issue in narrow windows 😉

nekodex:
Also, -69 more...

LiquidPL:
i was considering to remove the paddings around it and have it span whole page on low widths (<1000 px, because I don't think it will be possible to contain all that info in this case. any thoughts on that?

nekodex:
The layout looks ok when the window is < 840, it's just weird between 840 and 1080ish.

Regarding content in smaller windows, maybe add a minimum-width (around 700px maybe) and truncate text (i.e. long names)?

nekodex:
Also a reminder for @arflyte to replace header. 😉

nekodex:
Some feedback, roughly grouped into categories:

### General
- If a match has no games in it, the page breaks (might be worth testing for more edge cases like this)
- The game mode icon should be per-game not per-match (head-to-head, etc)
- Players that failed shouldn't contribute to team score (they should probably also be greyed out)
- Game mode, difficulty and scoring mode should probably be shown for the games (the scoring mode should probably affect the display, but that can be done later)

### Team Matches
For team matches, the teams don't seem to be handled correctly:
- The team field is 0 for non-team/neutral (e.g. head-to-head), 1 for blue team and 2 for red team
- `shapes-team-2.svg` appears to be missing (use blue for both team-1 and no-team for now)
- Scores don't appear to be calculated correctly? (in team matches at least)
- The scores should be ordered with the winning team on top with the players ordered by score descending in each team.

### Design
- Players that failed should probably be greyed out?
- 'Failed' text should probably be aligned
- @arflyte says to remove the slot number for now as he didn't realise you could pick random slots to be in, so something like this:


### JSON
- Try not to include unused data (e.g. all the undisplayed beatmap/beatmapset info)
- Try not to fetch data you're not using (e.g. all the events on page load), fetch what you need and more as required
- Try not to send duplicate data (e.g. user data is included for every event), take a look at what beatmap-discussions does
- The full list of countries is probably not needed most of the time (include the needed info as part of the user data?)
- Try to structure data nicer if possible and not always mirror the DB structure, especially for legacy tables (some.. uh, let's just say they don't exactly have the best structure. :sweat_smile:)
e.g. something like:
```json
{
  "event_id": 660339549,
  "event_type": "match_created",
  "timestamp": "2016-05-08T11:23:45+0000",
  "user_id": 8367324,
},
{
  "event_id": 660339551,
  "event_type": "host_changed",
  "timestamp": "2016-05-08T11:23:45+0000",
  "user_id": 8367324,
},
{
  "event_id": 660339741,
  "event_type": "player_joined",
  "timestamp": "2016-05-08T11:24:24+0000",
  "user_id": 8370623,
}
```

Also after some discussion with @peppy, we decided that the 'show more' behaviour should probably change:
- Only show the most recent 500 events by default (sorted oldest to newest), scrolling to the last event on page load if required.
- New events should load in at the bottom of the page for ongoing matches.
- The 'x more...' link should be replaced with a 'view all' link that shows all events instead of the most recent 500 (`/mp/{id}?full` or something)

Feel free to hit me up (here or discord) if you have any questions :+1:

app/Libraries/ModsFromDB.php
```diff
@@ -0,0 +1,84 @@
...
+        [20, 'Fade In', 'FI'],
+        [24, '9K', '9K'],
+    ];
+
+    public function __construct($mods)
+    {
+        $this->mods = $mods;
+    }
+
+    public function getEnabledMods()
```
nanaya: doesn't seem useful as an instance method since everything that uses it also caches the result anyway.

app/Libraries/ModsFromDB.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        if (!$this->enabledMods) {
```
nanaya: this will return true if no mods is enabled (`![] === true`)

app/Http/Controllers/MultiplayerController.php
```diff
@@ -0,0 +1,72 @@
...
+namespace App\Http\Controllers;
+
+use App\Models\Multiplayer\Match;
+use App\Models\Country;
+use App\Transformers\Multiplayer\MatchTransformer;
+use App\Transformers\Multiplayer\EventTransformer;
+use App\Transformers\CountryTransformer;
+use Request;
+
+class MultiplayerController extends Controller
```
nanaya: `MatchesController`?

app/Models/Multiplayer/Game.php
```diff
@@ -19,24 +19,55 @@
...
+
     protected $dates = [
         'start_time',
         'end_time',
     ];
+
     public $timestamps = false;
 
+    protected $casts = [
+        'mods' => 'integer',
```
nanaya: I believe this is ignored because there's a `getModsAttribute` function defined. Also the mysqlnd driver already casts integer.

database/migrations/2015_01_01_133337_multiplayer_base_tables.php
```diff
@@ -0,0 +1,101 @@
...
+            $table->timestamp('start_time')->default(DB::raw('CURRENT_TIMESTAMP'))->nullable();
+            $table->timestamp('end_time')->nullable();
+            $table->string('name', 255);
+
+            $table->index('end_time', 'time');
+        });
+
+        $connection = DB::connection('mysql-mp');
+
+        $connection->statement('ALTER TABLE `matches` ADD private BIT(1) NOT NULL');
```
nanaya: bit? really? (ノﾟ ヮﾟ)ノ︵ ┻━┻

resources/assets/less/bem/osu-page-header.less
```diff
@@ -60,5 +66,26 @@
...
     }
+
+    &--mp-history {
+      font-style: italic;
+    }
+  }
+
+  &__title-box {
+    position: absolute;
+    bottom: (@spacing * 2 - 10px);
```
nanaya: `10 * 2 - 10`?

resources/assets/less/bem/osu-layout.less
```diff
@@ -152,6 +152,15 @@
       }
     }
 
+    &--page-mp-history {
+      ._page;
+      padding: @spacing * 2 130px;
```
nanaya: math operation in less should be put in parenthesis

resources/assets/less/bem/osu-layout.less
```diff
@@ -152,6 +152,15 @@
...
+
+      @media @mobile {
```
nanaya: default to mobile styling, add `@media @desktop` instead

resources/assets/less/bem/osu-layout.less
nanaya: there's variable for that 130px

resources/views/multiplayer/match.blade.php
```diff
@@ -0,0 +1,40 @@
...
+    See the GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+--}}
+@extends("master")
+
+@section("content")
+    <div class="js-react--mp-history"></div>
+    {{--
```
nanaya: this comment isn't going to happen, just remove it

database/migrations/2015_01_01_133337_multiplayer_base_tables.php
LiquidPL: that's what was in the db schema, ask @peppy

resources/views/multiplayer/match.blade.php
LiquidPL: ok

resources/assets/coffee/react/mp-history/content.coffee
```diff
@@ -0,0 +1,88 @@
...
+
+      div className: 'osu-layout__row osu-layout__row--page-mp-history',
+        if @props.remainingEventsCount > 0
+          div className: 'mp-history-content__show-more-box',
+            a
+              className: 'mp-history-content__show-more'
+              onClick: @showMore
+              Lang.get 'multiplayer.match.more-events', count: @props.remainingEventsCount
+
+        for id, i in gameIds
```
nanaya: why not just loop the events? And include game data if the event is a game

LiquidPL: well, the events need to be wrapped in a separate box like in [here](http://puu.sh/plb6g.png), so I can't render them like that in a separate loop

nanaya: A combination of `:first-child` and `:last-child` probably works?

nanaya:
apparently the detail area should like this according to flyte (subject to further changes)



- mods should be part of the first row (and `row-reversed` direction)
- no need to align 300s and Combo
- less spacing between columns for second row
- not sure if it should be `Miss` or `Misses`

also, game mods icon (global one) seems a bit small.

resources/assets/less/bem/mp-history-event.less
```diff
@@ -0,0 +1,79 @@
+/**
+*    Copyright 2016 ppy Pty. Ltd.
```
nanaya: some of the license headers (like this one) are not properly aligned.

(don't bother fixing them all, just the ones you added here)

LiquidPL:
>players ordered by score descending in each team

should scores be sorted in non-team games too?

LiquidPL:
this is (almost) ready for review right now, as the only thing I'm trying to figure out right now is the adjacent selector thing

resources/assets/coffee/react/mp-history/content.coffee
```diff
@@ -60,22 +48,16 @@ class MPHistory.Content extends React.Component
...
-
-          div key: id,
-            if (id - prevId > 1)
-              events = @props.events[prevId + 1..id - 1]
-
-              if !_.isEmpty events
-                el MPHistory.EventsList, events: events, lookupUser: @props.lookupUser
-
+        for event, i in @props.events
+          if event.game?
```
vendethiel: this can be a `when`

resources/assets/coffee/react/mp-history/content.coffee
```diff
@@ -60,22 +48,16 @@ class MPHistory.Content extends React.Component
...
-
-        if !_.isEmpty lastEvents
-          el MPHistory.EventsList, events: lastEvents, lookupUser: @props.lookupUser
+              key: event.id
+          else
+            el MPHistory.Event,
+              event: event
+              lookupUser: @props.lookupUser
+              key: event.id
+              last: @props.events[i + 1]? && @props.events[i + 1].game?
```
vendethiel: `@props.events[i + 1]?.game?

app/Http/Controllers/MatchesController.php
```diff
@@ -0,0 +1,112 @@
...
+use App\Transformers\Multiplayer\MatchTransformer;
+use App\Transformers\Multiplayer\EventTransformer;
+use App\Transformers\UserCompactTransformer;
+use Request;
+
+class MatchesController extends Controller
+{
+    protected $section = 'multiplayer';
+
+    public function match($match_id)
```
nanaya: `show`? (and use `Route::resource`)

resources/assets/coffee/react/mp-history/event.coffee
```diff
@@ -0,0 +1,51 @@
...
+        span className: 'mp-history-event__time',
+          moment(@props.event.timestamp).format 'HH:mm:ss'
+      div className: "mp-history-event__type mp-history-event__type--#{@props.event.event_type}",
+        @icons[@props.event.event_type].map (m) ->
+          el Icon, name: m, key: m
+      div className: 'mp-history-event__info-box',
+        if @props.event.event_type != 'match-disbanded'
+          a
+            className: 'mp-history-event__text mp-history-event__text--username'
+            href: laroute.route 'users.show', users: user.id
```
nanaya: a bunch of tournament matches seem to have a null user_id (and thus failing here).

resources/assets/less/bem/mp-history-event.less
```diff
@@ -0,0 +1,98 @@
...
+
+  max-width: 640px;
+  margin-left: auto;
+  margin-right: auto;
+
+  background-color: @grey-e;
+
+  padding-bottom: @spacing * 0.5;
+
+  .mp-history-content__show-more-box + &, .mp-history-game + & {
```
nanaya: I was thinking something something like this: http://codepen.io/nanaya/pen/ZOrBzG

app/Models/Mod.php
```diff
@@ -0,0 +1,71 @@
...
+ *    as published by the Free Software Foundation.
+ *
+ *    osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+ *    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *    See the GNU Affero General Public License for more details.
+ *
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ */
+namespace App\Models;
```
nanaya: looks more like a library at the moment

app/Models/Multiplayer/Event.php
```diff
@@ -37,4 +48,23 @@ public function game()
...
+    }
+
+    public function scopeDefault($query)
+    {
+        return $query->orderBy('event_id', 'asc');
+    }
+
+    public function getTextAttribute($value)
+    {
+        if (array_key_exists($value, self::EVENT_TYPES)) {
```
nanaya:
```php
if (...) {
  return ['type' => static::EVENT_TYPES[...]];
} else {
  return ['type' => 'other', 'text' => $value];
}
```

(though probably should be under different attribute/function name (something like `detail`?)

app/Models/Multiplayer/Game.php
```diff
@@ -19,24 +19,75 @@
...
     }
+
+    public function beatmap()
+    {
+        return $this->belongsTo(Beatmap::class);
+    }
+
+    public function getModsAttribute($value)
+    {
+        if (!$this->_mods) {
```
nanaya: Looks like empty mod array will return true here?

app/Models/Multiplayer/Game.php
```diff
@@ -19,24 +19,75 @@
...
+
     protected $dates = [
         'start_time',
         'end_time',
     ];
+
     public $timestamps = false;
 
+    const SCORING_TYPES = [
+        0 => 'score',
```
nanaya: as this way of lookup probably only happens when reading the attribute, better flip the array? so one can for example do `->where('scoring_type', static::SCORING_TYPES['accuracy'])`. The same for `TEAM_TYPES` below.

app/Models/Multiplayer/Score.php
```diff
@@ -19,14 +19,40 @@
  */
 namespace App\Models\Multiplayer;
 
-class Score extends Model
+use App\Models\Score\Model as BaseModel;
```
nanaya: is this just to get the score calculation functions? Perhaps better to move the functions to a trait.

nanaya: or two traits (enabledMods?)

app/Transformers/Multiplayer/GameTransformer.php
```diff
@@ -0,0 +1,59 @@
...
+        'scores',
+    ];
+
+    public function transform(Game $game)
+    {
+        return [
+            'id' => $game->game_id,
+            'start_time' => $game->start_time->toIso8601String(),
+            'end_time' => $game->end_time ? $game->end_time->toIso8601String() : null,
+            'mode' => Beatmap::modeStr($game->play_mode),
```
nanaya: do conversion in model?

resources/assets/coffee/react/mp-history/main.coffee
```diff
@@ -0,0 +1,87 @@
...
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+{div} = React.DOM
+el = React.createElement
+
+class MPHistory.Main extends React.Component
+  timeBetweenRefresh: 10000
```
nanaya: `refreshInterval`? (or `refreshTimeout`)

resources/assets/coffee/react/mp-history/header.coffee
```diff
@@ -0,0 +1,28 @@
...
+# osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+{div, h1, h2} = React.DOM
+
+class MPHistory.Header extends React.Component
```
nanaya: I don't think it's used anywhere else yet, but for this kind of component, [stateless function](https://facebook.github.io/react/docs/reusable-components.html#stateless-functions) can be used.

nanaya:
as you're defining new section (`multiplayer`), it needs to be defined as well at least in `resources/assets/less/utilities.less` for `u-section-gradient-down` and probably other locations.

app/Http/routes.php
```diff
@@ -79,6 +87,14 @@
 Route::get('/community/slack', ['as' => 'slack', 'uses' => 'CommunityController@getSlack']);
 Route::post('/community/slack/agree', ['as' => 'slack.agree', 'uses' => 'CommunityController@postSlackAgree']);
 
+// Route::get('/matches/{matches}', ['as' => 'multiplayer.match', 'uses' => 'MatchesController@match']);
```
nanaya: remove?

app/Http/Controllers/MatchesController.php
```diff
@@ -0,0 +1,112 @@
...
+
+        $events = $events->get();
+
+        if (!$full) {
+            $events = $events->reverse();
+        }
+
+        $userIds = [];
+
+        foreach ($events as $event) {
```
nanaya: shouldn't this be in transformer?

LiquidPL: not sure about this; i don't really have a transformer that would fit in this case, what i have here is basically a list of events, and a list of users.

nanaya:
Sorry, I forgot about this one.

LiquidPL:
relax, there's a third one in the works :dancer:

LiquidPL:
strange, github said that there is a conflict, yet merge went in with no problems

app/Models/Multiplayer/Score.php
```diff
@@ -19,14 +19,42 @@
...
 
+use App\Models\Beatmap;
+use App\Traits\Scoreable;
+
 class Score extends Model
 {
+    use Scoreable;
+
     protected $table = 'game_scores';
+    protected $connection = 'mysql-mp';
```
nanaya: not needed?

app/Transformers/BeatmapsetCompactTransformer.php
```diff
@@ -0,0 +1,36 @@
...
+
+use App\Models\Beatmapset;
+use League\Fractal;
+
+class BeatmapsetCompactTransformer extends Fractal\TransformerAbstract
+{
+    public function transform(Beatmapset $beatmapset)
+    {
+        return [
+            'beatmapset_id' => $beatmapset->beatmapset_id,
```
nanaya: just id like other transformers

app/Http/routes.php
```diff
@@ -38,10 +38,19 @@
 
 // beatmapsets
 Route::get('/beatmaps/{beatmaps}/scores', ['as' => 'beatmaps.scores', 'uses' => 'BeatmapsController@scores']);
-Route::get('/b/{beatmaps}', ['as' => 'beatmaps.show', 'uses' => 'BeatmapsController@show']);
+Route::get('/beatmaps/{beatmaps}', ['as' => 'beatmaps.show', 'uses' => 'BeatmapsController@show']);
```
nanaya: `::resource`

app/Http/Controllers/MatchesController.php
```diff
@@ -0,0 +1,112 @@
...
+            ->with([
+                'game.beatmap.beatmapset',
+                'game.scores' => function ($query) {
+                    $query->with('game')->default();
+                },
+            ])
+            ->where('event_id', '>', $since);
+
+        if ($full) {
+            $events = $events->default();
```
nanaya: assignment not needed (also for some other lines)

app/Http/Controllers/MatchesController.php
```diff
@@ -0,0 +1,112 @@
...
+            $match,
+            new MatchTransformer
+        );
+
+        $full = Request::input('full', false);
+
+        return view('multiplayer.match', compact('match', 'full'));
+    }
+
+    public function matchHistory($match_id)
```
nanaya: just call it history since it's already in `MatchesController`?

app/Models/Multiplayer/Game.php
```diff
@@ -19,24 +19,80 @@
...
+    }
+
+    public function getModeAttribute()
+    {
+        return Beatmap::modeStr($this->play_mode);
+    }
+
+    public function getScoringTypeAttribute($value)
+    {
+        return array_search($value, self::SCORING_TYPES, true);
```
nanaya: `array_search_null` is preferred because it's slightly more useful.

resources/assets/coffee/react/mp-history/game.coffee
```diff
@@ -0,0 +1,64 @@
...
+
+    winningTeam = if @props.teamScores.blue > @props.teamScores.red then 'blue' else 'red'
+    difference = Math.abs @props.teamScores.blue - @props.teamScores.red
+
+    scores = game.scores.data.map (m) ->
+      if m.team == winningTeam
+        m.winning = true
+      m
+
+    scores = _.orderBy scores, ['winning', 'score'], ['asc', 'desc']
```
nanaya: `false` is actually higher than `true`. This one works because the other value isn't `false` but 'undefined`

LiquidPL: wait, so should i set `winning` to `false` when needed and change sorting appropriately?

nanaya: Perhaps something like 'teamRank' instead? Kind of stupid with just two teams but at least the order is obvious.

nanaya: Or just sort it twice with first one using function `@props.teamScores[s.team]`

resources/assets/coffee/react/mp-history/header.coffee
```diff
@@ -0,0 +1,27 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+{div, h1, h2} = React.DOM
+
+MPHistory.Header = (props) ->
+  div className: 'osu-layout__row osu-layout__row--page-compact',
+    div className: 'osu-page-header osu-page-header--mp-history',
+      div className: 'osu-page-header__title-box osu-page-header__title-box--mp-history',
```
nanaya: block modifier can modify element.

LiquidPL:
should be done now

app/Http/routes.php
```diff
@@ -38,10 +38,19 @@
...
+Route::resource('beatmaps', 'BeatmapsController', ['only' => ['show']]);
+
+// redirects to beatmapset anyways so there's no point
+// in having an another redirect on top of that
+Route::get('/b/{beatmaps}', ['uses' => 'BeatmapsController@show']);
+
 Route::get('/beatmapsets/search/{filters?}', ['as' => 'beatmapsets.search', 'uses' => 'BeatmapsetsController@search']);
 Route::resource('/beatmapsets', 'BeatmapsetsController', ['only' => ['index']]);
-Route::get('/s/{beatmapsets}', ['as' => 'beatmapsets.show', 'uses' => 'BeatmapsetsController@show']);
+Route::get('/beatmapsets/{beatmapsets}', ['as' => 'beatmapsets.show', 'uses' => 'BeatmapsetsController@show']);
```
nanaya: same as above if you haven't noticed

app/Models/Score/Model.php
```diff
@@ -23,20 +23,22 @@
...
 abstract class Model extends BaseModel
 {
+    use Scoreable;
+
     protected $primaryKey = 'score_id';
 
     protected $casts = [
         'perfect' => 'boolean',
         'replay' => 'boolean',
+        'enabled_mods' => 'integer',
```
nanaya: ?

app/Transformers/BeatmapsetCompactTransformer.php
nanaya: still not id :(

resources/assets/coffee/react/mp-history/_base.coffee
```diff
@@ -0,0 +1 @@
+@MPHistory = {}
```
nanaya: just remembered, this should be in `react-namespaces.coffee`.

resources/assets/less/bem/mp-history-player-score.less
```diff
@@ -0,0 +1,166 @@
...
+
+    @media @narrow {
+      display: none;
+    }
+  }
+
+  &__main {
+    flex-grow: 1;
+
+    padding: @spacing * 0.5 @spacing * 2 0 @spacing * 1.5;
```
nanaya: can you actually read this o_O

resources/assets/less/bem/osu-page-header.less
```diff
@@ -42,6 +42,35 @@
     .at2x-simple("/images/headers/tournament.jpg");
   }
 
+  &--mp-history {
+    .at2x-simple("/images/headers/mp-history.png");
+    background-position: center;
+    height: 125px;
+  }
+
+  &--contests {
```
nanaya: why is it here?

resources/assets/less/bem/osu-page-header.less
```diff
@@ -69,4 +98,25 @@
       font-size: @font-size--header-description;
     }
   }
+
+  &--mp-history &__title {
```
nanaya: the alternative is inside `&__title`:

```less
&__title {
  .@{_top}--mp-history & {
    /* stuff */
  }
}
```

resources/assets/less/functions.less
```diff
@@ -25,6 +25,15 @@
   height: @diameter;
 }
 
+.triangle(@length) {
```
nanaya: ?_?

resources/assets/less/bem/osu-layout.less
```diff
@@ -189,6 +189,15 @@
       }
     }
 
+    &--page-mp-history {
+      ._page;
+      padding: @spacing * 2;
+
+      @media (min-width: 1200px) {
```
nanaya: `@screen-lg-min` is close enough.

resources/assets/less/bem/mp-history-player-score.less
```diff
@@ -0,0 +1,166 @@
...
+  @wide: ~"(min-width: 1200px)";
+  @narrow: ~"(max-width: 1200px)";
+
+  display: flex;
+
+  &__shapes {
+    width: 40px;
+    height: 60px;
+
+    background-size: 40px 60px;
```
nanaya: another `contain`

resources/assets/less/bem/mp-history-player-score.less
```diff
@@ -0,0 +1,166 @@
...
+ *    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *    See the GNU Affero General Public License for more details.
+ *
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+.mp-history-player-score {
+  @wide: ~"(min-width: 1200px)";
```
nanaya: another `@screen-lg-min`? Also I tend to avoid `max-width` media selector. Default being the smaller size and change behavior using `min-width` instead.

resources/assets/less/bem-index.less
```diff
@@ -1,21 +1,21 @@
...
- *
- *    osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
- *    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
- *    See the GNU Affero General Public License for more details.
- *
- *    You should have received a copy of the GNU Affero General Public License
- *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
- *
- */
+*    Copyright 2015 ppy Pty. Ltd.
```
nanaya: nooo

resources/assets/less/bem/osu-page-header.less
LiquidPL: actually i have no idea, something must have messed up during the merge i guess, as this [is still attributed to nekodex](https://puu.sh/qFG3j.png)

app/Http/routes.php
```diff
@@ -38,10 +38,19 @@
...
+
+// redirects to beatmapset anyways so there's no point
+// in having an another redirect on top of that
+Route::get('/b/{beatmaps}', ['uses' => 'BeatmapsController@show']);
+
 Route::get('/beatmapsets/search/{filters?}', ['as' => 'beatmapsets.search', 'uses' => 'BeatmapsetsController@search']);
-Route::resource('/beatmapsets', 'BeatmapsetsController', ['only' => ['index']]);
-Route::get('/s/{beatmapsets}', ['as' => 'beatmapsets.show', 'uses' => 'BeatmapsetsController@show']);
+Route::resource('/beatmapsets', 'BeatmapsetsController', ['only' => ['index', 'show']]);
+Route::get('/beatmapsets/{beatmapsets}', ['as' => 'beatmapsets.show', 'uses' => 'BeatmapsetsController@show']);
```
nanaya: I think you forgot to remove the line.

app/Transformers/BeatmapsetCompactTransformer.php
nanaya: not yet, sir ¯\( ° o°)/¯

resources/assets/coffee/react/mp-history/event.coffee
```diff
@@ -0,0 +1,54 @@
...
+    event_type += '-no-user' if !user?
+
+    className = 'mp-history-events__event mp-history-event'
+
+    div className: className,
+      div className: 'mp-history-event__time-box',
+        span className: 'mp-history-event__time',
+          moment(@props.event.timestamp).format 'HH:mm:ss'
+      div className: "mp-history-event__type mp-history-event__type--#{event_type}",
+        @icons[event_type].map (m) ->
```
nanaya: this fails when user is null because of no matching `<event>-no-user` (appended at line 34).

resources/assets/less/bem/mp-history-events.less
```diff
@@ -0,0 +1,59 @@
...
+    background-color: @grey-e;
+    max-width: 640px;
+
+    margin-left: auto;
+    margin-right: auto;
+  }
+
+  .event-end() {
+    border-radius: 0 0 @border-radius-base @border-radius-base;
+    padding-bottom: @spacing * 2;
```
nanaya: since the rows already have padding-bottom of `(@spacing / 2)`, this should be `(@spacing * 2 - @spacing / 2)`.

resources/assets/coffee/react/mp-history/event.coffee
```diff
@@ -51,4 +50,4 @@ class MPHistory.Event extends React.Component
               user.username
 
         span className: 'mp-history-event__text',
-          Lang.get "multiplayer.match.events.#{event_type}"
+          Lang.get "multiplayer.match.events.#{event_type}#{if !user? then '-no-user' else ''}"
```
vendethiel: could also be `#{[user or '-no-user']}`

LiquidPL: but wouldn't this [interpolate the user onto the string](https://puu.sh/qGIXy.png) when we don't want it to?

nanaya:
need to handle cases where beatmap has been removed.

resources/assets/coffee/react/mp-history/event.coffee
vendethiel: oh right. `if !` or `unless` (as edited) work fine then.

resources/assets/less/bem/mp-history-game.less
```diff
@@ -0,0 +1,216 @@
...
+    background-color: @grey-e;
+
+    &--no-teams {
+      border-radius: 0 0 @border-radius-base @border-radius-base;
+    }
+  }
+
+  &__player-score {
+    width: 100%;
+    height: 60px;
```
nanaya: Add padding bottom instead? (why are there so many fixed heights here)

resources/assets/less/bem/mp-history-game.less
```diff
@@ -0,0 +1,216 @@
...
+    width: 100%;
+    height: 100%;
+
+    background-color: #000000;
+    opacity: 0.5;
+
+    border-radius: @border-radius-base @border-radius-base 0 0;
+  }
+
+  &__thumb {
```
nanaya: combine with `header` element?

resources/assets/coffee/react/mp-history/event.coffee
```diff
@@ -0,0 +1,53 @@
...
+      div className: 'mp-history-event__time-box',
+        span className: 'mp-history-event__time',
+          moment(@props.event.timestamp).format 'HH:mm:ss'
+      div className: "mp-history-event__type mp-history-event__type--#{event_type}",
+        @icons[event_type].map (m) ->
+          el Icon, name: m, key: m
+      div className: 'mp-history-event__info-box',
+        if event_type != 'match-disbanded'
+          if user?
+            a
```
nanaya: shouldn't this be part of translation?

LiquidPL: um, which part in here is untranslated?

nanaya: isn't it a sentence like "<user> has done something"? Or is it just two separate thing?

LiquidPL: yea, i render the user and the message separately, so in the case that a user doesn't exist somehow, i can not display the user and show a different message which doesn't suggest a particular user, like 'an user has joined the match'

LiquidPL: also the user is a link and it would be painful to stuff that into the translation i guess

nanaya: it's not difficult, though probably a bit ugly (using `osu.link`).

resources/assets/less/bem/mp-history-game.less
```diff
@@ -0,0 +1,212 @@
...
+  &__team-type {
+    position: absolute;
+
+    right: 10px;
+    bottom: 10px;
+
+    width: 40px;
+    height: 40px;
+
+    background-size: 40px 40px;
```
nanaya: another contain

resources/assets/less/bem/mp-history-player-score.less
```diff
@@ -0,0 +1,171 @@
...
+
+  &__shapes {
+    width: 40px;
+    height: 60px;
+
+    background-size: contain;
+
+    border-radius: @border-radius-base 0 0 @border-radius-base;
+
+    display: none;
```
nanaya: this probably isn't the best idea... (also the min-width is too wide)

app/Models/Multiplayer/Event.php
```diff
@@ -28,6 +30,15 @@ class Event extends Model
     ];
     public $timestamps = false;
 
+    const EVENT_TYPES = [
+        'player-left' => 'PART',
+        'player-joined' => 'JOIN',
+        'player-kicked' => 'KICK',
```
nanaya: translation for this event is missing

resources/assets/coffee/react/mp-history/game.coffee
```diff
@@ -38,8 +38,8 @@ class MPHistory.Game extends React.Component
 
     div className: 'mp-history-events__game mp-history-game',
       el MPHistory.BeatmapHeader,
-        beatmap: game.beatmap.data
-        beatmapset: game.beatmap.data.beatmapset.data
+        beatmap: if game.beatmap? then game.beatmap.data else @deletedBeatmap
```
vendethiel: `game.beatmap ? @deletedBeatmap`

resources/assets/less/bem/mp-history-player-score.less
LiquidPL: actually, i don't understand why

nanaya: something important is missing in previous screenshot.

resources/assets/coffee/react/mp-history/game.coffee
```diff
@@ -0,0 +1,74 @@
...
+    className += ' mp-history-game__player-scores--no-teams' if !showTeams
+
+    winningTeam = if @props.teamScores.blue > @props.teamScores.red then 'blue' else 'red'
+    difference = Math.abs @props.teamScores.blue - @props.teamScores.red
+
+    scores = game.scores.data.map (m) ->
+      m.teamRank = if m.team == winningTeam then 1 else 2
+      m
+
+    scores = _.orderBy scores, ['teamRank', 'score'], ['desc', 'desc']
```
nanaya: you mean 'asc'

resources/assets/coffee/react/mp-history/beatmap-header.coffee
```diff
@@ -0,0 +1,65 @@
...
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+{div, a, span, h1, h2} = React.DOM
+el = React.createElement
+
+class MPHistory.BeatmapHeader extends React.Component
```
nanaya: this is actually GameHeader, isn't it?

LiquidPL: fair point

resources/assets/less/bem/mp-history-game.less
LiquidPL: i don't get where are you going with this, these elements are fixed height because they had those dimensions in the designs.
also how am i supposed to replace height with padding-bottom?

nanaya: remove this here and add padding in `&__main`? (also kind of fixes mobile)

resources/assets/less/bem/mp-history-events.less
```diff
@@ -0,0 +1,59 @@
...
+  &__event {
+    .event;
+
+    &:first-child, :not(&) + & {
+      padding-top: @spacing * 2;
+      border-radius: @border-radius-base @border-radius-base 0 0;
+    }
+
+    &:last-child {
+      .event-end;
```
nanaya: this has wrong padding instead (15px).

resources/assets/coffee/react/mp-history/game-header.coffee
```diff
@@ -0,0 +1,65 @@
...
+    if @props.game.end_time
+      timeString += "- #{timeEnd}"
+    else
+      timeString += Lang.get 'multiplayer.match.in-progress'
+
+    a
+      className: 'mp-history-game__header'
+      href: (laroute.route 'beatmaps.show', beatmaps: @props.beatmap.id) if @props.beatmap.id
+      style:
+        backgroundImage: "url(#{@props.beatmapset.covers.cover})" if @props.beatmapset.covers.cover
```
nanaya: the condition is always true. A different check is needed (`cover_updated_at` being not null).

vendethiel:
👍

PR #595, opened 2016-06-10
Topic move

nanaya:
Apparently I already did most of the model stuff a long time ago. Now with actual interface to move the thing.


peppy:
Great job. I especially love the design. Maybe we don't need a designer after all.

nekodex:
nanayadesign™ is best 👍

Issue #582, opened 2016-06-07
[Modding] Crucial changes before going live with the new Modding Panel


Loctav:
Before we can make all new submitted mapsets use the new Modding Panel, there are still missing core features that need to be addressed before moving forward. 

I am not sure if we keep using the old beatmap listing while we use the new modding panel for now (and up until the new beatmap listing for individual beatmaps is done entirely). If we use the new beatmap listing, we require the beatmap admin options there, too. (like changing Online Offset or Online Tags)
## Permissions & Moderation
- [x] give QAT and GMT mod rights (/g/7 & /g/4)
- [x] add an option to soft-delete posts (hide them for all but moderators)
- [x] Silences/Restrictions should prevent users from posting issues or issue comments
- ~~add option to lock single issue threads, discussions per difficulty or discussion as a whole~~
- ~~add option to nuke/delete mapsets~~
## Missing Features
- [x] make links in issues and issue comments clickable 
- [ ] add option to generate a link to a specific issue/issue thread (like linking to a specific forum post) (related to https://github.com/ppy/osu-web/issues/458)
- [x] add option to filter issues by the user who posted them (see https://github.com/ppy/osu-web/issues/450)
- [ ] add option to change issue type manually as poster and moderator (related to https://github.com/ppy/osu-web/issues/448 ?)
- ~~add markdown support (see https://github.com/ppy/osu-web/issues/451)~~
- [ ]  add general section for each difficulty (see https://github.com/ppy/osu-web/issues/447)
#### Nomination, Qualification, Disqualification
- [x] add `Nominate` and `Qualify` button for Nominators & QAT (/g/28 & /g/7) 
  - Maps for Ranked require 1 `Nominate` and 1 `Qualify` (Drain time of all difficulties < 5:00 min )
  - Map for Approval require 2 `Nominate` and 1 `Qualify` (Drain time of all difficulties >= 5:00 min)
  - `Nominate` and `Qualify` apply to the whole mapset at once.
  - all `Nominate` are revoked when the map gets updated by the mapper or when a Nominator/QAT "pops it" (probably needs an own button, too)
- [x] add `Disqualify`button for QAT (/g/7) (revokes all `Nominate` and `Qualify`), returning it from Qualified to Pending
- [x] add a Nomination/Disqualification log (when did who place a `Nominate`, `Qualify`, `Disqualify` or `Pop`)
## Interface adjustments
[...]

Issue #450, opened 2016-04-12
[Modding] Add ability to filter by specific conditions


peppy:
Here's a couple of filters we need. Probably a dropdown or similar?
- Issues closed by mapper (if a mapper closes an issue themselves, it may not have been checked correctly, so this will be a commonly used one).
- Issues posted by current user (so you can narrow down to only your own posts).

possibly others

nanaya:
I think this has been implemented?

PR #596, opened 2016-06-10
Add filter to beatmap discussions for threads created by the current user

nekodex:
Addresses some of the issues mentioned in #582 and #450


resources/assets/coffee/react/beatmap-discussions/discussions.coffee
```diff
@@ -117,6 +117,7 @@ BeatmapDiscussions.Discussions = React.createClass
 
   filtered: (discussion) ->
     switch @props.currentFilter
+      when 'mine' then discussion.user_id != @props.currentUser.id
```
nanaya: as you mentioned this function looks quite dumb :dancer:

nekodex: 👍

PR #598, opened 2016-06-11
Spanish Lang

Uniterno:
Almost everything translated (or at least a 90%)...

I didn't translate stuff related to Slack, and some stuff that I really don't have an idea on how to translate it and keep the idea of the word (you know, some languages just can't be translated, and have to be interpreted)

Probably this resource won't be used quite yet until a final Language Implementation, but here is it ¯\_(ツ)_/¯

### DIFF
--- a/resources/lang/es/admin.php
+++ b/resources/lang/es/admin.php
@@ -39,7 +39,7 @@ return [
 
                 'forum-name' => 'Foro #:id: :name',
 
-                'no-cover' => 'Fondo no definido',
+                'no-cover' => 'Portada no definida',
 
                 'submit' => [
                     'save' => 'Guardar',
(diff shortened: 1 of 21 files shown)


Uniterno:
I've copied, pasted and translated files from "en" to "es", so I don't know about StyleCI stuff... lol
As far as I know, my changes shouldn't break anything, hehe.

Shadoxfix:
Just wondering. What would happen in the case a new entry is made? Do non-English languages automatically use the English variant if the entry is not present? I would imagine it to be quite a pain to maintain if it wasn't the case because at some point someone will forget to update the other languages.

peppy:
It will always fallback to English.

Going forward we definitely need a better way to manage localisations. We may resort to google docs again like with other osu! resources.

PR #603, opened 2016-06-16
switch to proper many to many relationship

kschamplin19:
Should™ work.


nanaya:
doesn't seem to bring much benefit (or at all at the moment) and even adds burden of manually casting date?

kschamplin19:
its proper form. no need for a model for the pivot tables, plus you can just call $user->achievements() or $achievement->users().
of course if you don't like it that's fine, i think it's more clear this way.
also attaching an achievement is really simple, just $user->achievements()->attach($achievement_id)

nanaya:
I don't seem to be able to find out how to set the bridge's date...

kschamplin19:
pass an array as the second param when attaching.
$user->achievements()->attach($achievement_id, ["date" => "valid date"]);
updating is also possible
$user->achievements()->updateExistingPivot($achievement_id, $attributes);
[this is helpful](http://laraveldaily.com/pivot-tables-and-many-to-many-relationships/)
[official documentation](https://laravel.com/docs/5.1/eloquent-relationships#many-to-many)

nanaya:
Looks not necessary for now. And adding support for casting the pivot attribute seems even worse (if work at all) ( https://laracasts.com/discuss/channels/general-discussion/cast-pivot-table-value-to-date )

nanaya:
And the documentation for newPivot seems removed in 5.1...

kschamplin19:
oh god ._.

yeah this is a pretty situational tool

closing.

PR #606, opened 2016-06-16
Polish translation + small fixes

Xyloo:


nanaya:
the translation and the fixes should be separated in two PRs.

PR #607, opened 2016-06-16
Small fixes

Xyloo:
better wording/grammar

### DIFF
--- a/resources/lang/en/community.php
+++ b/resources/lang/en/community.php
@@ -126,4 +126,25 @@ return [
             'instructions' => 'click the heart button to proceed to the osu!store',
         ],
     ],
+    'slack' => [
+       'header' => [
+            'small' => 'osu!dev',
+            'large' => 'osu!public Slack Access',
+        ],
+
+        'disabled' => 'The public slack community is temporarily unavailable. If you wish to reach out, please create an issue on the appropriate <a href="https://github.com/ppy">github repository</a> or contact us at <a href="mailto::mail">:mail</a>.',
+
+        'guest-begin' => 'You need to be ',
+        'guest-middle' => 'logged in',
+        'guest-end' => ' to receive a slack invite!',
+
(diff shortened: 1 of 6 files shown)


resources/lang/en/community.php
```diff
@@ -126,25 +126,4 @@
             'instructions' => 'click the heart button to proceed to the osu!store',
         ],
     ],
-    'slack' => [
```
nekodex: Some of these translations are still in use, so we probably don't want to remove them all...

resources/lang/en/users.php
```diff
@@ -170,7 +170,7 @@
             'total_hits' => 'Total Hits',
             'total_score' => 'Total Score',
         ],
-        'title' => 'profile / :username',
+        'title' => ':username\'s profile',
```
nekodex: For interpolated strings, you can use double quotes instead of escaping, e.g.:
```php
'title' => ":username's profile"
```

resources/lang/en/community.php
Xyloo: As far as I know there will be a new page made for public discord anyways, but these changes are available to view and copy if we need them

resources/lang/en/users.php
Xyloo: I was worried about it messing with the actual :username

nekodex: It doesn't mess with it and behaves fine 😃

resources/lang/en/community.php
nekodex: Sure, but the page is still available right now (https://new.ppy.sh/community/slack). So if we merge these translation changes, that page will be full of missing translation errors... 😛 

We'll delete the translations when we remove/replace the page.

resources/lang/en/users.php
Xyloo: Will fix when I come back home

PR #617, opened 2016-06-23
Authorization updates

nanaya:
- post edit and delete doesn't check for reply
- shortcut for logged in and restricted/silenced check
- restricted means silenced


app/Libraries/OsuAuthorize.php
```diff
@@ -391,30 +359,47 @@ public function checkUserPageEdit($user, $pageOwner)
...
     }
+
+    public function ensureLoggedIn($user, $prefix = '')
+    {
+        if ($user === null) {
+            throw new AuthorizationException($prefix.'require_login');
+        }
+    }
+
+    public function ensureCleanRecord($user, $prefix = '')
```
peppy: cool function name :p

PR #622, opened 2016-06-30
Brazilian Portuguese Translation

UnusualToast:
All files translated to Brazilian Portuguese.

There might be some inconsistencies regarding osu!supporter/modding/accuracy, which are some hard words to translate. Should be mostly fine, though.

edit: grammar

### DIFF
--- a/resources/lang/pt-br/admin.php
+++ b/resources/lang/pt-br/admin.php
@@ -1,0 +1,85 @@
+<?php
+
+/**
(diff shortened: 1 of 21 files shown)


Xyloo:
well, you need to fix what StyleCI says. Easiest way is to create an account at StyleCI and it can automatically make a fixing pull request, so you don't have to change each one manually

PR #623, opened 2016-06-30
Dutch translation

flashwave:
Everyone seems to be doing it so why not Dutch?


ErinvanderVeen:
This translation is not fit for release, it contains many grammatical and spelling errors.

PR #635, opened 2016-07-06
Add missing forum acl migrations

### LINKED ISSUE #605: Missing phpbb tables?
Some of the phpbb tables seem to be missing from the migrations, giving you errors if you try to make new topics etc.

Tables that I've noticed missing: phpbb_acl_options, phpbb_acl_roles_data.

Do these need to be added to migrations and seeded at all? Or am I being retarded?

LiquidPL:
Fixes #605.

### DIFF
--- a/app/Models/Forum/AuthGroup.php
+++ b/app/Models/Forum/AuthGroup.php
@@ -17,31 +17,12 @@
  *    You should have received a copy of the GNU Affero General Public License
  *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
  */
-namespace App\Models;
+namespace App\Models\Forum;
 
 use Illuminate\Database\Eloquent\Model;
 
-class UserGroup extends Model
+class AuthGroup extends Model
 {
-    protected $table = 'phpbb_user_group';
-    protected $primaryKey = 'group_id';
+    protected $table = 'phpbb_acl_groups';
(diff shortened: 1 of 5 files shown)


nekodex:
Does this need seed data to go with it to work?

LiquidPL:
Alright, now seeders set up permissions for the seeded users and forums.

However, this is not the case for any user that is created manually, as in that case an entry in `phpbb_user_group`. Should this be noted in the readme?

peppy:
When creating a user they should be assigned to the default group, which gets some amount of ACL permissions, so I don't think it's necessary to note it in the readme for now.

peppy:
Awarding 40 bounty for this.

app/Models/Forum/AuthGroup.php
```diff
@@ -0,0 +1,28 @@
...
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ */
+namespace App\Models\Forum;
+
+use Illuminate\Database\Eloquent\Model;
+
+class AuthGroup extends Model
+{
+    protected $table = 'phpbb_acl_groups';
```
nanaya: um, this one is also defined for model in `app/Models/Forum/Authorize.php`

PR #640, opened 2016-07-06
Use the previously designed warning box

### LINKED ISSUE #638: Forum posting box shouldn't be visible when no permission to post or thread locked

nanaya:
Fixes #638 ┐(･_･)┌

The warning box only visible at the end of topic and reply buttons are hidden.


nanaya:
Note that on locked thread, it'll first show "Please login to reply." and then replaced with "Can not reply to a locked thread." after logging in :dancer:

PR #643, opened 2016-07-09
Spanish translation fixes

TheBITLINK:
I don't know if this is the right place, but there are some mistakes in the spanish translation.


Uniterno:
I was going to send a PR with fixes from https://new.ppy.sh/forum/t/474254 later
But now that you have already opened the PR, maybe you can find and fix these issues for me? That way I don't have to open an unnecessary PR.

TheBITLINK:
Done.

Respecting the dates, they could be easily localized too, as osu!web seems to use moment.js for date parsing.

Now, for the countries, it requires additional work , country names are stored in their own table on the database AFAIK.

PR #644, opened 2016-07-13
Misc fixes

nanaya:
- correctly show `[deleted user]` for beatmapsets which user is deleted
- prevent user posts count from going negative


peppy:
what the fuck travis.

PR #645, opened 2016-07-13
Turbolinks fixes

nanaya:
Just realized yesterday that version 5 has one important change which obsoletes some of previously used techniques.

Namely it saves document using `cloneNode` so elements data and events are cleared out upon navigation. Storing element state using `el.someData = value` doesn't quite work anymore.

Reference: https://github.com/turbolinks/turbolinks/issues/30

### DIFF
--- a/resources/assets/coffee/_classes/forum-cover.coffee
+++ b/resources/assets/coffee/_classes/forum-cover.coffee
@@ -40,7 +40,7 @@ class @ForumCover
 
 
   closeModal: (e) =>
-    return unless @hasCoverEditor() && @header[0]._open
+    return unless @hasCoverEditor() && @isModalOpen()
 
     if e
       return if $(e.target).closest('.js-forum-cover--open-modal').length
@@ -51,7 +51,7 @@ class @ForumCover
     Fade.out $('.blackout')[0]
     @header[0].classList.remove 'forum-category-header--cover-modal'
 
-    @header[0]._open = false
+    @isModalOpen(false)
 
 
   hasCover: =>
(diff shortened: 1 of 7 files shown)


resources/assets/coffee/_classes/forum-cover.coffee
```diff
@@ -62,24 +62,23 @@ class @ForumCover
...
-  toggleModal: (e) =>
-    e.preventDefault()
+  isModalOpen: (isModalOpen) =>
+    return false if !@hasCoverEditor()
 
-    if @header[0]._open
-      @closeModal()
-    else
-      @openModal()
+    if isModalOpen?
```
vendethiel: is "false" not an acceptable value for `isModalOpen`?

nanaya: `isModalOpen?` is true if the value is neither null nor undefined.

vendethiel: I think my question still stands – doesn't "false" make more sense as a "isXXX" value?

nanaya: ? Not sure what you're asking? If it's `isModelOpen == false` then the `data-modal-open` will be emptied (the `: void 0` added by coffeescript) and `@isModalOpen()` will return false?

rightfold: It's misleading to allow something named `isXXX` to have values other than `true` and `false`.

nanaya: the attribute doesn't support anything but string to begin with. Also I changed it to `0` instead of `undefined`.

PR #646, opened 2016-07-13
Navigate between forum posts using left and right keys

### LINKED ISSUE #639: Left and right keys should navigate between forum posts

TheBITLINK:
This should close #639

I don't know if this should be in its own class. If that's the case, tell me.

### DIFF
--- a/resources/assets/coffee/_classes/forum.coffee
+++ b/resources/assets/coffee/_classes/forum.coffee
@@ -44,6 +44,8 @@ class @Forum
     $(document).on 'click', '.js-post-url', @postUrlClick
     $(document).on 'submit', '.js-forum-posts-jump-to', @jumpToSubmit
 
+    $(document).on 'keyup', @keyboardNavigation
+
     $.subscribe 'stickyHeader', @stickHeader
 
 
@@ -136,6 +138,17 @@ class @Forum
       Turbolinks.visit @postUrlN(postN)
 
     true
+
+  keyboardNavigation: (e) =>
+    return if $(e.target).is('input') or not @_postsCounter.length
+
+    e.preventDefault()
+
+    n = switch e.which
+      when 37 then @currentPostPosition - 1
+      when 39 then @currentPostPosition + 1
+
+    try @jumpTo n
 
 
   scrollTo: (postId) =>
--- a/resources/assets/coffee/_classes/forum.coffee
+++ b/resources/assets/coffee/_classes/forum.coffee
@@ -140,7 +140,7 @@ class @Forum
     true
 
   keyboardNavigation: (e) =>
-    return if $(e.target).is('input, select, textarea') or e.target.isContentEditable or not @_postsCounter.length
+    return if osu.isInputElement(e.target) or not @_postsCounter.length
 
     e.preventDefault()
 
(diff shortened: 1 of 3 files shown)


nanaya:
shouldn't there be a check if it's inside an input box? 

<del>Also the function name can use a better name than just `keyup`</del> more like it doesn't belong in that class since it doesn't deal with the seekbar. I'd just put that in `Forum`.

nanaya:
oh and better use its own `js-` prefixed class name otherwise I may forget the class name is used by something from javascript. Or maybe just check if `@_postsCounter.length != 0` instead of delegate.

resources/assets/coffee/_classes/forum.coffee
```diff
@@ -137,6 +139,17 @@ class @Forum
 
     true
 
+  keyboardNavigation: (e) =>
+    return if $(e.target).is('input') or not @_postsCounter.length
```
nanaya: I think [this](https://github.com/ccampbell/mousetrap/blob/master/mousetrap.js#L986) is the complete list of inputs.

resources/assets/coffee/_classes/forum.coffee
```diff
@@ -137,6 +139,16 @@ class @Forum
 
     true
 
+  keyboardNavigation: (e) =>
+    return if $(e.target).is('input, select, textarea') or e.target.isContentEditable or not @_postsCounter.length
```
nanaya: Move the input check to `osu_common`? Also perhaps skip using jquery to avoid creating jquery object every keypress.

resources/assets/coffee/osu_common.coffee
```diff
@@ -59,6 +59,8 @@ along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
   parseJson: (id) ->
     JSON.parse document.getElementById(id).text
 
+  isInputElement: (el) ->
+    el.tagName == 'INPUT' or el.tagName == 'SELECT' or el.tagName == 'TEXTAREA' or el.isContentEditable;
```
nanaya: stray semicolon

resources/assets/coffee/osu_common.coffee
```diff
@@ -223,3 +225,4 @@ along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
                 o.find("[ref=#{k}]").html v
 
             area.append o
+
```
nanaya: um

TheBITLINK:
Whoops, sorry about that, i'm not used to coffee

resources/assets/coffee/osu_common.coffee
vendethiel: `el.tagName in ['INPUT', 'SELECT', 'TEXTAREA']` is also a nicer way to write it.
(The compiler unrolls it if the lhs is an array literal (and splatless))

nanaya: oh wow I didn't know it does that much.

nanaya:
Looks fine. I'll do final check sometime later and merge if there's no problem.

nanaya:
it sure works, except it feels weird due to how post scrolling logic works.

PR #647, opened 2016-07-17
Wrong variable name

nanaya:
Didn't pay enough attention.

### DIFF
--- a/app/Models/Forum/Topic.php
+++ b/app/Models/Forum/Topic.php
@@ -80,8 +80,8 @@ class Topic extends Model
             }
 
             if (($params['cover'] ?? null) !== null) {
-                $cover->topic()->associate($topic);
-                $cover->save();
+                $params['cover']->topic()->associate($topic);
+                $params['cover']->save();
             }
         });
 


app/Models/Forum/Topic.php
```diff
@@ -80,8 +80,8 @@ public static function createNew($forum, $params, $pollParams = null)
             }
 
             if (($params['cover'] ?? null) !== null) {
```
vendethiel: `isset($params['cover'])`?

nanaya: it might be set with value of null.

vendethiel: sure?

```
$ php -r '$a = [null]; var_dump(isset($a[0]));'
bool(false)
```

nanaya: what the hell. Also, no.

nanaya: I think I did that somewhere else but the fact I forgot it already means I shouldn't use it.

vendethiel: okay, that's a fair point :P.

nanaya: more like because it's not a function but a language construct?

vendethiel: yes. That's why for the longest time, `isset(literal)` was a parse error. It now has a better error message now, though.

nanaya: still parse error for null.
```
>>> isset(null)
PHP Parse error: Syntax error, unexpected ')', expecting '[' on line 1
```

nanaya: and of course :D 

```
>>> isset('1')
PHP Parse error: Syntax error, unexpected ')', expecting T_PAAMAYIM_NEKUDOTAYIM on line 1
```

vendethiel: what's your PHP version? as said, it's a very recent change

```
$ php -v
PHP 5.5.33 (cli) (built: Mar 25 2016 10:45:27)
Copyright (c) 1997-2015 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2015 Zend Technologies
$ php -r "isset(null);"
Fatal error: Cannot use isset() on the result of an expression (you can use "null !== expression" instead) in Command line code on line 1
```

nanaya: looks like the artisan tinker shell ate the message.

```
$ php --version
PHP 7.0.8 (cli) (built: Jul  7 2016 01:16:14) ( NTS )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.8, Copyright (c) 1999-2016, by Zend Technologies
$ php -r "isset(null);"
PHP Fatal error:  Cannot use isset() on the result of an expression (you can use "null !== expression" instead) in Command line code on line 1

Fatal error: Cannot use isset() on the result of an expression (you can use "null !== expression" instead) in Command line code on line 1
$ artisan tinker
Psy Shell v0.7.2 (PHP 7.0.8 — cli) by Justin Hileman
>>> isset(null)
PHP Parse error: Syntax error, unexpected ')', expecting '[' on line 1
>>>
```

PR #649, opened 2016-07-18
More Spanish transl. fixes

comentarinformal:
One or two flagrant mistakes (desargas, remover imagenes)
Still a lot to do and discuss (unconsistent case, anglicisms...), but now a bit less.

### DIFF
--- a/resources/lang/es/authorization.php
+++ b/resources/lang/es/authorization.php
@@ -32,13 +32,13 @@ return [
     'chat' => [
         'channel' => [
             'read' => [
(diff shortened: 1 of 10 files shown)


Uniterno:
Approve :+1: 

Some mistakes done by translating based on literal english lang, this fixes are more logical in the context they are used...
(I wouldn't take "remover imagenes" as a flagrant mistake, but "Eliminar" makes more sense.)

Yay!

TheBITLINK:
And here we go again...

PR #653, opened 2016-07-20
Load infos of profile page on demand instead of on page load

miterosan:
I am currently changing that profile pages stop loading things like allscores... in inital page load.
Instead the infos should be loaded seperate and on demand only. 

I do it over a GET html request.
- [x]  inital implementation
- [x]  limiting results
- [x]  adding offset
- [ ]  implement it into the current webpage
  - [x] history section
  - [ ] beatmap section 
  - [ ] recent section
  - [ ] ranks section
  - [ ] medals section(?)
  - [ ] remove the code that gives it all over the props func


miterosan:
soo, I won't update this pr in the next 3 months, because it's a pain to develop on a slow pc, and my free time got less.

PR #660, opened 2016-07-28
New header

nanaya:
User profile looks a bit desert.


peppy:
Gonna merge this, but I want to bring up in discussion with @arflyte whether the active page should be white instead of yellow in the dropdown list (or highlighted in some other way). I think this would add clarity.

PR #670, opened 2016-08-01
Debug env alternative to elasticsearch

LiquidPL:
It's not the best we could have, but well, it's not supposed to be used in production anyways :dancer:


app/Models/Beatmapset.php
```diff
@@ -359,7 +396,12 @@ public static function search(array $params = [])
 
         self::sanitizeSearchParams($params);
 
-        $beatmap_ids = self::searchES($params);
+        if (App::environment('local')) {
```
nanaya: check for elasticsearch host config instead? (and remove the default)

app/Models/Beatmapset.php
```diff
@@ -342,6 +343,42 @@ function ($e) {
...
+        $query = self::where('title', 'like', '%'.$query.'%');
+
+        if ($mode) {
+            $query = $query->whereHas('beatmaps', function ($query) use (&$mode) {
+                $query->where('playmode', '=', $mode);
+            });
+        }
+
+        if ($genre) {
+            $query = $query->where('genre_id', '=', $genre);
```
nanaya: the assignment isn't needed.

nanaya:
Offset and limit are missing :(

config/elasticsearch.php
```diff
@@ -2,5 +2,5 @@
 
 
 return [
-    'hosts' => [env('ES_HOST', 'localhost:9200')],
+    'hosts' => [env('ES_HOST')],
```
nanaya: `(array) env(...)` here and check for `empty(config(...))`?

app/Models/Beatmapset.php
```diff
@@ -19,6 +19,7 @@
  */
 namespace App\Models;
 
+use App;
```
nanaya: This isn't used.

(I thought styleci caught this already but apparently it didn't)

PR #673, opened 2016-08-01
Allow logging in by email

nanaya:
To match what input box says =D

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -75,18 +75,21 @@ class UsersController extends Controller
         if (LoginAttempt::isLocked($ip)) {
             return error_popup('your IP address is locked. Please wait a few minutes.');
         } else {
-            $username = Request::input('username');
+            $usernameOrEmail = Request::input('username');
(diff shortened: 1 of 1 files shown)


peppy:
:+1:

app/Http/Controllers/UsersController.php
```diff
@@ -75,18 +75,21 @@ public function login()
...
+            }
 
             if (Auth::check()) {
                 return Auth::user()->defaultJson();
             } else {
-                LoginAttempt::failedAttempt($ip, $username);
+                LoginAttempt::failedAttempt($ip, $usernameOrEmail);
 
-                return error_popup('wrong password or username');
+                return error_popup('wrong password or email');
```
lunafreya: I think it needs to say "Wrong password or username/email", not "wrong password or email".

peppy:
It does not. We are phasing out username login, so this is correct.

avail:
no more username logins? what's the reason behind that :o

peppy:
it's better security and also makes sure the user keeps the email up-to-date.

PR #683, opened 2016-08-09
Beatmap info page redesign

### LINKED ISSUE #559: Beatmap Details Page, Redesigned
New design for the beatmap page. Designs not final but ready for implementation. All of these image snippets are at 2x, unless mentioned.

## Cover and Main Information

Cover is beatmap's background. 50%-70% gradient black overlay on the cover for visibility. Will be fine tuned in the future. All the text except in the beatmap details box, has drop shadows. Snippet at 1x.

Song preview when cover is clicked. Song preview function will be finalised in the future.

### Mode and Difficulty Selector

Available modes are available in white, 50% opacity if not available. Selected mode has a 50% black background.

Underneath the modes are the list of available difficulties for selected mode. Line break at 10. Selected difficulty has a 50% black background.

Difficulty title, play count and favourites are displayed under the difficulty list. By default it shows the selected difficulty; it will display other difficulty's info when hovered.

Star Difficulty only appears on hover.

### Main Information

Song title (Semi Bold) followed by song artist (Medium).

### Mapper Info

Mapper avatar, and name; followed by map submission date and ranked date. Mapper name italicised, date is bolded.

### Download buttons

Buttons to download the maps. A dialog box will appear if a non-supporter clicked osu!direct, promoting supporter functions.

## Beatmap Details

3 boxes, black 50% opacity. Seperated by 2px. All of the texts inside this area has no drop shadows.

### Basic Stats

Length, Tempo, Node count and Slider count.

### Advanced Stats

Extra information showing the Circle Size, HP Drain, Accuracy, Approach Rate and Star Difficulty. The order is not finalised and will be updated in the future.

All the bars are white 50% opacity. Filled with white, except Star Difficulty which is orange. Resize transitions when different difficulty is selected.

The values on the right are centered.

### User Rating

Identical to the current website, displaying the rating spread of the map. Rating spread chart is solid blue for every value.

## Extra Information

Image snippet at 1x. Displaying additional info. Box height is fixed. The description is scrollable inside the box if it's taller than the box. Tags are clickable. Hide extra tags if it's taller the height limit to discourage mappers spamming tags.

Success rate function is identical to the current site. Point of Failure graph is attached to the bottom of the extra info box.

### Ranking Chart Filter

User will be able to change the list to show global results, player's country results or results between their friends.

By default all the mod filters are enabled. Toggled by clicking it. If enabled, list will only show games with at least one of the mods used. Results without any mods will be displayed at all time.

If you think there's a better way for the filters, do suggest.

The list will show the top 50. Grey highlight on hover (in the future, a quick info panel will appear http://puu.sh/pcxCx/76bd9eea21.jpg)

### Best Result / Player Best

Best result or player's best appears as a box, within the list.
Snippet at 1x. Gold border for #1 result, none for player best.

Best result is shown on top of the list. If the player best is within 50, result is shown within the list (eg: between #7 and #9 with the player best is #8). If the player's best is beyond 50, it's shown at the bottom of the list at #50.

#### Player

Rank number #1 should be in gold, vertically center aligned between the blue guide lines. Player avatar with usual drop shadow. Italicised bold player name on top of player's country flag and team flag.

#### Score Details

Mods used are displayed. All the scores details are italicised, additional bold for score.

X/Y/Z is a placeholder for 300/100/50. Will be updated when I can come up with the better way to represent them.

Relative time when the score was achieved is displayed under the scores.

LiquidPL:
Resolves #559.

~~React/CS ready for review, only thing left to do is mod filtering.~~ everything pretty much done

Also I recall talking with nanaya about handling the converts and how he'll escalate this to @arflyte. Anything on that?


LiquidPL:
alright, everything should be done here for now

app/Http/Controllers/BeatmapsController.php
```diff
@@ -52,29 +53,55 @@ public function scores($id)
 
         $beatmap = Beatmap::findOrFail($id);
         $mode = Request::input('mode', Beatmap::modeStr($beatmap->playmode));
+        $enabled_mods = Request::input('enabledMods');
```
nanaya: pass in array instead and convert to bitset by extending function (convert to library?) used for the other direction.

app/Http/Controllers/BeatmapsController.php
```diff
@@ -52,29 +53,55 @@ public function scores($id)
...
         } catch (\InvalidArgumentException $ex) {
             return error_popup($ex->getMessage());
         }
 
+        if ($enabled_mods) {
+            if (!$user->isSupporter()) {
+                return error_popup(trans('errors.supporter_only'));
+            }
+
+            $query->whereIn('enabled_mods', [$enabled_mods, 0]);
```
nanaya: what's with the `0`?

app/Http/Controllers/BeatmapsController.php
```diff
@@ -52,29 +53,55 @@ public function scores($id)
...
-                $scores = $scores
+                $query
                     ->whereIn('user_id', model_pluck($user->friends(), 'zebra_id'));
                 break;
         }
 
-        return fractal_collection_array($scores->get(), new ScoreTransformer, 'user');
+        $scores = fractal_collection_array($query->get(), new ScoreTransformer, 'user');
+        $userScore = null;
+        $userScorePosition = -1;
```
nanaya: initialize with null?

app/Http/Controllers/BeatmapsController.php
```diff
@@ -52,29 +53,55 @@ public function scores($id)
...
+
+        if ($user) {
+            $score = $beatmap->scoresBest()->where('user_id', $user->user_id)->with('user')->first();
+
+            if ($score && ($score->enabled_mods === $enabled_mods || $score->enabled_mods === 0)) {
```
nanaya: `enabled_mods` returns array

app/Http/Controllers/BeatmapsController.php
```diff
@@ -52,29 +53,55 @@ public function scores($id)
...
+                $userScore = fractal_item_array($score, new ScoreTransformer, 'user');
+                $userScorePosition = $query->limit(null)->count() + 1;
```
nanaya: this is just wrong.

app/Libraries/BBCodeFromDB.php
```diff
@@ -280,7 +280,7 @@ public function parseYoutube($text)
         return $text;
     }
 
-    public function toHTML()
+    public function toText()
```
nanaya: allow modifier parameter and extend there instead?

resources/assets/coffee/_classes/hits.coffee
```diff
@@ -15,22 +15,23 @@
...
 ###
-{a} = React.DOM
-el = React.createElement
 
-class BeatmapsetPage.ExtraTab extends React.Component
-  pageSwitch: (e) =>
-    e.preventDefault()
+class @Hits
+  @generate: ({score, playmode}) ->
+    elements = ['count300', 'count100', 'count50']
```
nanaya:
```
elements =
  if playmode == 'mania'
    [...]
  else
    [...]
```

resources/assets/coffee/_classes/hits.coffee
```diff
@@ -15,22 +15,23 @@
...
+      elements = ['countgeki', 'count300',  'countkatu', 'count100', 'count50']
 
-    $.publish 'beatmapset:page:jump', @props.page
+    header = ''
+    values = ''
 
-  render: ->
-    className = 'link link--white link--no-underline page-extra-tabs__item'
-    className += ' page-extra-tabs__item--active' if @props.page is @props.currentPage
+    for elem, i in elements
```
nanaya: map and join?

resources/assets/coffee/_classes/stacked-bar-chart.coffee
```diff
@@ -37,12 +37,19 @@ class @BarChart
     @svgWrapper = @svg.append 'g'
 
   loadData: (data) ->
+    data = _.map data, (d) ->
```
nanaya: can be simplified using one big loop

nanaya:
```coffeescript
    @data = []
    for d, i in data
      for v, j in d.values
        @data[j] ?= []
        @data[j].push
          type: d.type
          value: v
          height: if i == 0 then 0 else @data[j][i - 1].value + @data[j][i - 1].height
```

resources/assets/coffee/react/_components/mods.coffee
```diff
@@ -19,12 +19,23 @@
...
   render: ->
-    div className: 'mods',
+    modsClassName = 'mods'
+    imageClassName = 'mods__mod-image'
+
+    classModifiers = @props.classModifiers || []
+
+    for mod in classModifiers
+      modsClassName += " mods--#{mod}"
+      imageClassName += " mods__mod-image--#{mod}"
```
nanaya: modifier at block level can modify element.

resources/assets/coffee/react/beatmapset-page/header.coffee
```diff
@@ -28,34 +28,123 @@ class BeatmapsetPage.Header extends React.Component
...
+            playmode: mode
+            currentBeatmapId: @props.currentBeatmap.id
+            newBeatmapId: _.last @props.beatmapList[mode]
+            currentPlaymode: @props.currentBeatmap.mode
+
       div
-        className: 'beatmapset-header',
-        onClick: @onClick,
+        className: 'beatmapset-header'
+        onClick: @togglePreview
```
nanaya: create a separate layer for this so the `preventPropagation` thing isn't needed (also you missed some links).

nanaya: also the header feels weird with hovering highlights the play button but doesn't always play the thing when clicked.

nanaya: I just remembered there's this `isClickable` thing. Check its usage in `resources/assets/coffee/shared.coffee`

resources/assets/coffee/react/beatmapset-page/header.coffee
```diff
@@ -28,34 +28,123 @@ class BeatmapsetPage.Header extends React.Component
...
+                  continue if !@props.beatmapset.video
+                  link = Url.beatmapDownload @props.beatmapset.beatmapset_id, true
+                when 'no-video'
+                  link = Url.beatmapDownload @props.beatmapset.beatmapset_id, false
+                when 'direct'
+                  firstRow = elem
+                  icon = 'angle-double-down'
+                  link = if currentUser.isSupporter then Url.beatmapDownloadDirect @props.beatmapset.beatmapset_id else laroute.route 'support-the-game'
+
+              a
```
nanaya: make this a function instead?

app/Http/Controllers/BeatmapsController.php
LiquidPL: `0` is no mods enabled

resources/assets/coffee/react/beatmapset-page/header.coffee
```diff
@@ -28,34 +28,123 @@ class BeatmapsetPage.Header extends React.Component
...
+                key: elem
+                className: 'beatmapset-header__button btn-osu-big'
```
nanaya: extend `btn-osu-big`.

app/Http/Controllers/BeatmapsController.php
nanaya: wouldn't the result then includes both with and without mods?

LiquidPL: that's how it's supposed to work; scores with enabled mods + scores with nomod

resources/assets/coffee/react/beatmapset-page/main.coffee
```diff
@@ -49,36 +49,47 @@ BeatmapsetPage.Main = React.createClass
...
-    return if scoreboard != 'global' && !currentUser.isSupporter
+    enabledMods = @state.enabledMods ^ enabledMod
 
     @scoresCache ?= {}
-    cacheKey = "#{@state.currentBeatmapId}-#{@state.currentPlaymode}-#{scoreboard}"
+    cacheKey = "#{@state.currentBeatmapId}-#{@state.currentPlaymode}-#{enabledMods}-#{scoreboardType}"
 
     loadScore = =>
-      @setState scores: @scoresCache[cacheKey]
+      @setState
```
nanaya: should `abort()` previous xhr otherwise wrong scores may show up at wrong beatmap.

app/Http/Controllers/BeatmapsController.php
nanaya: that doesn't sound right...?

Also I just realized in current form,

score1: mod1, mod2, mod3
score2: mod1
score3: -

filtering by mod1 would return score2 and score3 while it should be score1 and score2?

resources/assets/coffee/react/beatmapset-page/score-big.coffee
```diff
@@ -0,0 +1,81 @@
...
+
+BeatmapsetPage.ScoreBig = (props) ->
+  className = 'beatmapset-score-big'
+  classNamePosition = 'beatmapset-score-big__position'
+
+  if props.position == 1
+    className += ' beatmapset-score-big--first-score'
+    classNamePosition += ' beatmapset-score-big__position--first-score'
+
+  classNamePosition += ' beatmapset-score-big__position--not-top-score' if props.position > 99
```
nanaya: Either the 99 here or the 50 in comment for this in less file is wrong.

resources/assets/coffee/react/beatmapset-page/scoreboard.coffee
```diff
@@ -15,100 +15,81 @@
...
-                scoreboard: m
-                currentScoreboard: @props.currentScoreboard
-
-          div className: "#{bn}__line"
-
-      div className: className,
+    userScoreFound = false
+
+    mods = ['EZ', 'NF', 'HT', 'HR', 'SD', 'PF', 'DT', 'NC', 'HD', 'FL']
+    if @props.playmode == 'mania' && @props.convert == true
```
nanaya: pretty sure both converts and non-converts can use same set of mods (by selecting mania mode explicitly).

app/Http/Controllers/BeatmapsController.php
LiquidPL: First of all, [this](https://puu.sh/qFzgY.png).

Also, I'm not sure if we can show scores with a *particular* mod enabled in any other way than iterating over the scores and checking the bitset on each one.

resources/assets/coffee/react/beatmapset-page/header.coffee
```diff
@@ -28,34 +28,123 @@ class BeatmapsetPage.Header extends React.Component
...
+                span className: 'beatmapset-header__value-icon', el Icon, name: 'heart'
+                span className: 'beatmapset-header__value-name', @props.beatmapset.favourite_count.toLocaleString()
+
+            span
+              className: 'beatmapset-header__star-difficulty'
+              style:
+                visibility: 'hidden' if !@props.hoveredBeatmap
+              "#{osu.trans 'beatmaps.beatmapset.show.stats.stars'} #{if @props.hoveredBeatmap then @props.hoveredBeatmap.difficulty_rating.toFixed 2 else ''}"
+
+          div className: 'beatmapset-header__details-text beatmapset-header__details-text--title',
```
nanaya: why not just use `a` here directly?

app/Http/Controllers/BeatmapsController.php
nanaya: looks more like exclude operation? Something like `enabled_mods mod 2048 = 0`? Not sure how it'll perform though.

LiquidPL: i'm assuming this is being done on php level, right?

resources/assets/coffee/react/beatmapset-page/scoreboard-tab.coffee
```diff
@@ -15,17 +15,14 @@
...
 # along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
 ###
-{span} = React.DOM
+{div} = React.DOM
 
-class BeatmapsetPage.ScoreboardTab extends React.Component
-  _scoreboardSwitch: =>
-    $.publish 'beatmapset:scoreboard:set', scoreboard: @props.scoreboard
+BeatmapsetPage.ScoreboardTab = (props) ->
+  className = 'beatmapset-scoreboard__tab page-tabs__tab'
```
nanaya: extend the block instead of mixing?

app/Http/Controllers/BeatmapsController.php
nanaya: yes

resources/assets/less/bem/beatmapset-score-big.less
```diff
@@ -0,0 +1,170 @@
...
+    &--bottom {
+      display: flex;
+      align-items: center;
+      justify-content: flex-end;
+
+      height: 25px;
+      align-self: flex-end;
+      background-color: @grey-e;
+
+      z-index: 0;
```
nanaya: personally I'd use something different than this. (see store product page)

resources/assets/coffee/_classes/stacked-bar-chart.coffee
nanaya: just a warning, code above contains a subtle bug

resources/assets/coffee/react/beatmapset-page/info.coffee
```diff
@@ -0,0 +1,101 @@
...
+                  a
+                    key: tag
+                    className: 'beatmapset-info__text beatmapset-info__text--tag'
+                    href: laroute.route 'beatmapsets.index', q: tag
+                    tag
+
+      div className: 'beatmapset-info__success-rate beatmapset-success-rate',
+        div className: 'beatmapset-success-rate__label beatmapset-success-rate__label--main', osu.trans 'beatmaps.beatmapset.show.info.success-rate'
+
+        div className: 'beatmapset-success-rate__bar',
```
nanaya: extend the bar class instead.

resources/assets/less/bem/beatmapset-info.less
```diff
@@ -0,0 +1,90 @@
...
+  &__text-box {
+    width: 250px;
+    padding-bottom: @spacing * 0.5;
+
+    &--source {
+      height: 35px;
+    }
+
+    &--tags {
+      height: 80px;
```
nanaya: why is it fixed height?

resources/assets/less/bem/beatmapset-stats.less
```diff
@@ -1,49 +1,138 @@
...
+    align-items: center;
   }
 
-  &__line {
-    .default-line;
+  &__icon {
+    width: 15px;
+    height: 15px;
+
+    background-size: 15px 15px;
```
nanaya: `contain`?

resources/assets/less/bem/beatmapset-stats.less
```diff
@@ -1,49 +1,138 @@
...
+    background-color: rgba(0, 0, 0, 0.5);
+
+    margin-bottom: 1px;
 
-    &--full {
-      display: block;
+    &--basic {
+      display: flex;
+      padding-left: 24px;
+      height: 35px;
```
nanaya: padding instead of fixed height?

resources/assets/less/bem/beatmapset-stats.less
```diff
@@ -1,49 +1,138 @@
...
+  &__row {
+    background-color: rgba(0, 0, 0, 0.5);
+
+    margin-bottom: 1px;
 
-    &--full {
-      display: block;
+    &--basic {
+      display: flex;
+      padding-left: 24px;
```
nanaya: where did the number come from?

nanaya: a plain center alignment seems to work?

nanaya:
when the description box contains image, it pushes out other thing :(

nanaya:
The hover styling seems wrong (should be semi transparent box just like selected one)

resources/assets/coffee/react/beatmapset-page/scoreboard-tab.coffee
LiquidPL: i think that less somehow can't extend from the nested selectors (like `&__tab` in this case)

resources/assets/less/bem/beatmapset-score-big.less
LiquidPL: what exactly?

LiquidPL:
should be pretty much done

LiquidPL:
done, still need @arflyte to provide the final nomod filter button

nekodex:
A few things I noticed:
- New game mode selector hasn't been implemented? Page still uses the old tabs:
  
  vs
  

- Hover performance in some places is... really bad. e.g.:
  Beatmap selector
    
  Mod filter
    
Maybe try offloading more to CSS instead of javascript/react?

- The right-margin on the mod selectors is causing all the icons to flash when moving the mouse between icons
  

- Scores should only disappear after new ones are fetched, i.e. no "Loading scores..."
  Page scroll position gets messed up when the page height keeps changing

- Clickable area on the audio preview is way too large (side-effect from previous design's behaviour probably)
Flyte missed this in his intial spec, but is going to think about it and work it in. For now just make the clickable area the same size as the play button itself.

nekodex:
A few more things:
- The mod filter seems to be behaving as `and` instead of `or`
- Selecting 'no fail' toggles the filter but doesn't seem to change the scores being displayed
- The selected mod filter doesn't seem to toggle its state (i.e. become bright) until after the response from the server is received

LiquidPL:
I moved the mod filter hovering stuff to CSS, but I don't think this is going to be possible with the map picker, as it is supposed to dynamically change beatmap name/stars in the html, so there's pretty much nothing that I can actually offload to CSS in here.

LiquidPL:
> The selected mod filter doesn't seem to toggle its state (i.e. become bright) until after the response from the server is received

i guess this makes sense if I am to keep displaying old scores until new load

nekodex:
new nomod icons from @arflyte 

mod_nomod
mod_nomod 2x

app/Transformers/BeatmapsetTransformer.php
```diff
@@ -59,6 +59,7 @@ public function transform(Beatmapset $beatmapset = null)
             'tags' => $beatmapset->tags,
             'video' => $beatmapset->video,
             'status' => $beatmapset->status(),
+            'ratings' => $beatmapset->ratingsCount(),
```
nekodex: Also, you don't want this here - it will slow down every other page that uses the BeatmapsetTransformer... especially ones that display multiple Beatmapsets (e.g. user profile page)

I'd move it to an include and only use it when required instead.

app/Models/Beatmapset.php
```diff
@@ -814,6 +814,26 @@ public function approver()
         return $this->belongsTo("App\Models\User", 'user_id', 'approvedby_id');
     }
 
+    public function userRatings()
+    {
+        return $this->hasMany(UserBeatmapsetRating::class);
+    }
+
+    public function ratingsCount()
```
nekodex: Also, should probably use a groupBy so it doesn't have to load the entire rating dataset into memory...

resources/assets/coffee/react/beatmapset-page/scoreboard-tab.coffee
nanaya: what

resources/assets/coffee/react/beatmapset-page/scoreboard-tab.coffee
```diff
@@ -15,17 +15,14 @@
...
-    className += ' beatmapset-scoreboard__tab--active' if @props.scoreboard == @props.currentScoreboard
-
-    span
-      className: className
-      onClick: @_scoreboardSwitch
-      osu.trans "beatmaps.beatmapset.show.extra.scoreboard.#{@props.scoreboard}"
+  div
+    className: className
+    onClick: ->
+      $.publish 'beatmapset:scoreboard:set', scoreboardType: props.type
```
nanaya: this neither set `resetMods` nor `enabledMod`, resulting in `null` `enabledMods` and breaks stuff. Switch type, select a filter, and then unselect it - `enabledMods` then contains something like this:

resources/assets/coffee/react/beatmapset-page/info.coffee
```diff
@@ -0,0 +1,101 @@
...
+
+      @_failurePointsChart = new StackedBarChart @refs.chartArea, options
+
+    @_failurePointsChart.loadData data
+
+  render: ->
+    percentage = _.round (@props.beatmap.passcount / (@props.beatmap.playcount + @props.beatmap.passcount)) * 100
+
+    div className: 'osu-layout__row osu-layout__row--page-beatmapset beatmapset-info',
+      div className: 'beatmapset-info__details',
```
nanaya: probably better to show three rows instead of keeping as columns on mobile?

resources/assets/coffee/react/beatmapset-page/score.coffee
```diff
@@ -0,0 +1,51 @@
...
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+{div, a} = React.DOM
+el = React.createElement
+
+BeatmapsetPage.Score = (props) ->
```
nanaya: apparently a `PureRenderMixin` is much faster for this one.

resources/assets/coffee/react/mp-history/score.coffee
```diff
@@ -47,7 +47,9 @@ class MPHistory.Score extends React.Component
         div className: 'mp-history-player-score__info-box mp-history-player-score__info-box--stats',
           div className: 'mp-history-player-score__stat-row mp-history-player-score__stat-row--first',
             div className: 'mp-history-player-score__mods-box',
-              el Mods, mods: @props.score.mods, reversed: true
+              el Mods,
```
nanaya: something should be done with this in mobile. At least the username need to be stay visible.

resources/assets/less/bem/beatmapset-scoreboard.less
```diff
@@ -18,110 +18,61 @@
...
+      &:hover .@{_top}__mod-box {
+        opacity: 0.5;
+      }
 
-    &--score {
-      &:hover {
-        background-color: @grey-d;
-        border-radius: 5px;
+      & .@{_top}__mod-box:hover {
+        opacity: 1;
```
nanaya: no hover effect if any mod is selected (no `--initial` modifier)?

nanaya:
oh lol I see =D

resources/assets/less/bem/beatmapset-scoreboard.less
LiquidPL: i think the hover effects should be visible only when no mods are selected

resources/assets/coffee/react/beatmapset-page/main.coffee
```diff
@@ -49,36 +49,50 @@ BeatmapsetPage.Main = React.createClass
...
-  setHash: ->
+  setHash: =>
     osu.setHash BeatmapsetPageHash.generate
-      page: @state.currentPage
       beatmapId: @state.currentBeatmapId
       playmode: @state.currentPlaymode
 
-  setCurrentScoreboard: (_e, {scoreboard, forceReload = false}) ->
+  setCurrentScoreboard: (_e, {scoreboardType = @state.currentScoreboardType, enabledMod = null, forceReload = false, resetMods = false}) =>
     return if @state.loading
```
LiquidPL: i mean right now this is supposed not to let you start a new download if there is another in progress, so that something like this shouldn't actually happen

nanaya: but in this case won't it show wrong beatmap scores after the current process complete if user changes the beatmap when it's still loading?

select beatmap A &rarr; load score for A &rarr; select beatmap B &rarr; score for A still loading &rarr; load finished, shows score for A

nanaya: by beatmap I mean difficulty.

LiquidPL:
done for now

nanaya:
now with stray scrollbar?



The wrapping is just weird. Perhaps use non-fixed height instead?

LiquidPL:
Well you'd probably have to ask @arflyte what does he think about it

LiquidPL:
Also I assume that non-fixed height should be only used on mobile I guess (as a scrollbar on a touchscreen is gonna be quite problematic)

LiquidPL:
is this any good?

arflyte:
I totally missed that description section. I'll think of something on how to fix it. Leave it as it is for now?

LiquidPL:
i think so

resources/assets/coffee/react/beatmapset-page/scoreboard.coffee
```diff
@@ -15,100 +15,90 @@
...
+        div className: 'beatmapset-scoreboard__mods-wrapper',
+          div className: modsClassName,
+            for mod in mods
+              el BeatmapsetPage.ScoreboardMod,
+                key: mod
+                mod: mod
+                enabled: _.includes @props.enabledMods, mod
 
       div className: className,
         if @props.scores.length > 0
```
nanaya: since scores is not reset in `setCurrentScoreboard`, this will always be true?

resources/assets/less/bem/beatmapset-stats.less
```diff
@@ -1,49 +1,136 @@
...
+    &--basic {
+      display: flex;
+      padding: 10px 0;
+    }
+
+    &--advanced {
+      padding: 15px 15px 0 15px;
+    }
+
+    &:last-child {
```
nanaya: if it's `margin-top` instead, this isn't needed?

resources/assets/less/bem/beatmapset-success-rate.less
```diff
@@ -1,83 +1,69 @@
...
-    font-style: italic;
+    font-size: 12px;
+    color: @grey-5;
+
+    &--main {
+      text-align: center;
+    }
+
+    &--percentage {
+      margin-left: -50%;
```
nanaya: looks a bit wrong in ie

resources/assets/less/bem/mods.less
```diff
@@ -58,7 +73,12 @@
     width: @_mod-width;
     height: @_mod-height;
 
-    &--large {
+    .@{_top}--small & {
+      width: @_mod-width-small;
+      height: @_mod-height-small;
+    }
+
+    .@{_top}--large {
```
nanaya: missing `&`

nanaya:
I have this diff here which may be useful. https://0paste.com/9266-667b3890.txt

LiquidPL:
>um :running_man: (probably there will be more but this one is a bit big so I'm putting this out earlier)

The rank on `.score-big` was supposed to be centered within it's bounding box, and it [looks](http://i.imgur.com/p1EJnN9.png) sorta meh on higher ranks when aligned at a certain point. Can try moving the avatar and username slightly to the right.

resources/assets/coffee/react/beatmapset-page/score.coffee
```diff
@@ -18,16 +18,18 @@
...
 el = React.createElement
 
-class BeatmapsetPage.ScoreboardItem extends React.Component
+BeatmapsetPage.Score = React.createClass
+  mixins: [React.addons.PureRenderMixin]
+
   render: ->
-    elements = ['rank', 'flag', 'player', 'mods', 'icons', 'score', 'accuracy']
+    div className: 'beatmapset-score',
+      for elem in ['position', 'flag', 'player', 'mods', 'rank', 'score', 'accuracy', 'hits']
```
LiquidPL: so i assume that you want me to just type out each element manually without any loops?

resources/assets/less/bem/beatmapset-score-big.less
```diff
@@ -0,0 +1,174 @@
...
+  height: 90px;
+  width: calc(~"100% - 6px");
+
+  .default-border-radius;
+  .default-box-shadow;
+
+  margin: @spacing * 2 auto;
+
+  &--first-score {
+    border: 3px solid @yellow;
```
nanaya: I'd rather use `box-shadow: 0 0 0 3px @yellow, 0 4px @box-shadow-spread @box-shadow-color;` than this and all that 1.7s.

resources/assets/less/bem/beatmapset-score-big.less
```diff
@@ -0,0 +1,174 @@
...
+
+    align-self: center;
+
+    &--first-score {
+      color: @yellow-dark;
+    }
+
+    &--not-top-score {
+      // aka past the top 50 scores
+      padding: 0 @spacing;
```
nanaya: not relevant anymore?

resources/assets/less/bem/beatmapset-score-big.less
nanaya: Anyway, this one doesn't seem to be needed.

resources/assets/less/bem/beatmapset-score.less
```diff
@@ -0,0 +1,83 @@
...
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+.beatmapset-score {
+  display: flex;
+  align-items: center;
+
+  padding: @spacing * 0.5 @spacing * 2;
```
nanaya: here, have some parentheses `(((((((((())))))))))`

resources/assets/coffee/react/beatmapset-page/score.coffee
nanaya: yeah, unless you think this loop actually doing anything useful?

resources/assets/less/bem/beatmapset-score.less
```diff
@@ -0,0 +1,83 @@
...
+      padding-right: @spacing;
+    }
+
+    &--rank {
+      height: 18px;
+      width: 18px;
+      margin-right: @spacing;
+    }
+
+    &--score, &--accuracy, &--hits {
```
nanaya: a dedicated modifier (or element) instead of this?

resources/assets/less/bem/beatmapset-stats.less
```diff
@@ -1,49 +1,163 @@
...
-    margin-bottom: @grid-gutter-width / 8;
+  width: 275px;
+
+  &__row {
+    background-color: rgba(0, 0, 0, 0.5);
+
+    margin-top: 1px;
+
+    &--preview {
+      display: flex;
```
nanaya: thanks to this, the click target is a bit small.

LiquidPL: How about making the whole box containing that play button clickable?

resources/assets/less/bem/beatmapset-score-big.less
LiquidPL: >`@box-shadow-spread @box-shadow-color`

actually those appear to be undefined

resources/assets/less/bem/beatmapset-score-big.less
```diff
@@ -0,0 +1,164 @@
...
+
+  height: 90px;
+
+  .default-border-radius;
+  .default-box-shadow;
+
+  margin: @spacing * 2 0;
+
+  &--first-score {
+    box-shadow: 0 0 0 3px @yellow, 0 4px 3px fade(@grey-0, 25%);
```
nanaya: move the default box shadow (spread, color) to variable

resources/assets/less/functions.less
```diff
@@ -34,7 +34,11 @@
 }
 
 .default-box-shadow() {
-  box-shadow: 0px 1px 3px fade(#000, 25%);
+  box-shadow: 0px 1px 3px @box-shadow-spread @box-shadow-color;
```
nanaya: I said wrong thing and somehow propagated this far.

resources/assets/less/variables.less
```diff
@@ -39,6 +39,9 @@
 
 @z-index--beatmapset-header: 1;
 
+@box-shadow-spread: 3px;
+@box-shadow-color: fade(@gray-0, 25%);
```
nanaya: doesn't compile. (this also needs fixing)

resources/assets/less/functions.less
LiquidPL: what is exactly wrong here (aside from the fact that there are 5 parameters instead of 4)

resources/assets/less/bem/avatar.less
```diff
@@ -91,6 +91,16 @@
     box-shadow: none;
   }
 
+  &--beatmapset {
```
nanaya: there's same modifier a bit on the top (also, try sorting by name).

PR #685, opened 2016-08-09
Make it possible to search for beatmaps with rank status Any

### LINKED ISSUE #667: Beatmap Listing any rank status is broken
Currently when you try to search for the Rank Status "Any", this will behave like "Ranked & Approved". Reason being the line https://github.com/ppy/osu-web/blob/fa2d88f79653755190b644efcdee3a6c6611de54/app/Http/Controllers/BeatmapsetsController.php#L55 where `id=null` behaves like an absent id, which will then again default to 0 or "Ranked & Approved" on the server.
A simple fix would be changing id to `-1` as it is in the current live version.

<sub><sup>The reason I don't include a pull-request is that I think the submission would be too minor to justify me showing up as a contributor on its own and I don't have setup the env to properly test changes locally. When I find something bigger, I'll properly submit a PR.

WorldSEnder:
Fixes #667


nekodex:
Sorry this pull request has been overlooked for so long. If you can bring this up to date, we can get it merged.

nekodex:
Just as a side note, the beatmap search functionality was written as a working prototype and needs to be revisited. So while this pull request fixes the stated bug, it's more a bandaid on code that needs to be rewritten.

PR #692, opened 2016-08-15
User preferences page

### LINKED ISSUE #550: User Preferences
This is an open-ended task to add user preferences to the system. We have a design for this (and I think it was partly implemented) so ask if you are wanting to take this one. It will likely require consultation directly with us on an on-going basis, so let us know if you are interested in helping here and we'll talk it through.

This will involve tasks like:
- Change avatar
- Change profile details
- Change email or password
- Change signature

**[Bounty](https://goo.gl/nFdoyI) offered**: 500-1500 depending on implementation quality.

Feel free to reply in this thread should you have any questions.

LiquidPL:
Joint effort of me and @Risers.

Resolves #550.

Known problems:
- I don't think there is a way to grab a notifications count (or a fully functional notification system fwiw), or private messages count (according to @peppy the PM system will be revamped so dunno).
- I didn''t know how to tap into the current avatar storage (as in http://a.ppy.sh), so the avatars are uploaded to the same s3 bucket as profile covers. We might want to leave it be I guess, but we'll probably have to migrate the avatars to new storage or add a fallback of some sort.

~~Things left to do:~~
- ~~Online users graph.~~


LiquidPL:
this one should be done too (apart from properly doing the avatars)

nanaya:
- update profile is too complex (my mistake but it's adding even more complexity here)
- has some security problem (allowing assigning value to any columns)
- it's not that difficult to reimplement current avatar storage (it's actually just a single bucket containing files named `user_id` on top level)
- looks like $user->avatarUrl() will incur query every time it's called (caused by `profileCustomization()->firstOrNew([])`)
- the tablet profile doesn't seem to be needed yet
- UserPrefs class doesn't work with turbolinks
- not sure if worth using react just for upload box (the upload element itself isn't even react)
- element id should be unique in html

I think the graph thingy can still be used though the class names can use a bit of fixes.

LiquidPL:
So am I supposed to close that one given the circumstances? Or should I rewrite it?

PR #705, opened 2016-08-18
Display XH & SH ranks in user profile

### LINKED ISSUE #134: Display XH & SH ranks in user profile
From this thread: https://osu.ppy.sh/forum/t/94891

TPGPL:
Resolves #134


TPGPL:
Ok, I think it's done now.

nanaya:
there are no such columns in the table...

cl8n:
For this stat to go live, you'd need to add the logic to score submissions as well as re-calculate every user's rank counts with their submitted scores.

The concept is nice, but I wonder if we'll ever see it...

PR #713, opened 2016-08-22
Update header image with higher resolution/more tasteful version.

peppy:


nanaya:
the css needs update as well...

PR #720, opened 2016-08-26
More contest fixes

nekodex:
A bunch of under-the-hood fixes and cleanup. Also adds support for showing the results of completed contests (well, only music ones initially) .


peppy:
looks good, let's announce some results!

PR #722, opened 2016-08-29
Match history fixes

LiquidPL:
* games in progress now properly update their data once the game ends
* games that have no end time are not shown (save for the most recent one)
* team games that are in progress don't display the team scoring


resources/assets/coffee/react/mp-history/main.coffee
```diff
@@ -52,8 +53,17 @@ class MPHistory.Main extends React.Component
     .done (data) =>
       return if _.isEmpty data.events.data
 
-      newEvents = _.concat @state.events, data.events.data
-      lastEvent = _.last(newEvents)
+      lastIndex = _.findLastIndex @state.events, (o) => o.id == @state.since + 1
```
nanaya: why is it needed?

resources/assets/coffee/react/mp-history/content.coffee
```diff
@@ -51,6 +51,8 @@ class MPHistory.Content extends React.Component
         div className: 'mp-history-events',
           for event, i in @props.events
             if event.detail.type == 'other'
+              continue if !event.game.data.end_time && event.id != @props.lastGameId
```
nanaya: perhaps add `event.game?` check as well while at it.

resources/assets/coffee/react/mp-history/main.coffee
LiquidPL: so that we can remove the event with a match in progress (+ any other event that was created after it) and replace it with ones we got in an update

nanaya: \<something about the next event id is not necessarily going to be `previous id + 1`\>

```coffeescript
  ...
  data:
    since: @minNextEventId() # wtf is even "since"
  ...

  previousEvents = _.filter @state.events, (e) =>
    e.id < data.events.data[0].id

  newEvents = _.concat previousEvents, data.events.data
...

minNextEventId: =>
  lastOngoingGame = ._findLast @state.events, (e) =>
    e.game? && !e.game.date.end_time?

  if lastOngoingGame?
    lastOngoingGame?.id - 1
  else
    _.last(@state.events)?.id ? 0
```

resources/assets/coffee/react/mp-history/main.coffee
```diff
@@ -64,7 +74,8 @@ class MPHistory.Main extends React.Component
         events: newEvents
         allEventsCount: data.all_events_count
         users: newUsers
-        since: _.last(newEvents).id
+        since: if !lastGameEvent? || lastGameEvent.game.data.end_time then lastEvent.id else lastGameEvent.id - 1
+        lastGameId: if lastGameEvent? then lastGameEvent.id else 0
         disbanded: lastEvent.event_type == 'match-disbanded'
```
nanaya: this can be removed and let first line of `@loadHistory` be 

```coffeescript
return if _.last(@state.events)?.detail.type == 'match-disbanded'
```
instead.

(and I guess the current one has some kind of timing problem because `setState` is async.)

resources/assets/coffee/react/mp-history/main.coffee
```diff
@@ -64,7 +74,8 @@ class MPHistory.Main extends React.Component
...
 
     .always =>
```
nanaya: and the timeout should be stored in a variable so it can be cleared during unmount to avoid running it after unmounted.

resources/assets/coffee/react/mp-history/content.coffee
```diff
@@ -51,6 +51,8 @@ class MPHistory.Content extends React.Component
         div className: 'mp-history-events',
           for event, i in @props.events
             if event.detail.type == 'other'
+              continue if event.game? && !event.game.data.end_time && event.id != @props.lastGameId
```
nanaya: this will render the `MPHIstory.Game` is `event.game` is `null` or `undefined`.

resources/assets/coffee/react/mp-history/main.coffee
```diff
@@ -64,12 +71,19 @@ class MPHistory.Main extends React.Component
         events: newEvents
         allEventsCount: data.all_events_count
         users: newUsers
-        since: _.last(newEvents).id
-        disbanded: lastEvent.event_type == 'match-disbanded'
+        since: @minNextEventId
```
nanaya: passing a function? (and why even keep as state)

nanaya:
rebase and one missing optional `?` somewhere.

resources/assets/coffee/react/mp-history/main.coffee
```diff
@@ -64,12 +70,18 @@ class MPHistory.Main extends React.Component
         events: newEvents
         allEventsCount: data.all_events_count
         users: newUsers
-        since: _.last(newEvents).id
-        disbanded: lastEvent.event_type == 'match-disbanded'
+        lastGameId: _.findLast(newEvents, (e) -> e.game?)?.id ? 0
```
nanaya: is the default 0 needed?

resources/assets/coffee/react/mp-history/main.coffee
```diff
@@ -42,18 +43,23 @@ class MPHistory.Main extends React.Component
...
       data:
-        since: @state.since
+        since: @minNextEventId()
 
     .done (data) =>
       return if _.isEmpty data.events.data
 
-      newEvents = _.concat @state.events, data.events.data
-      lastEvent = _.last(newEvents)
+      lastIndex = _.findLastIndex @state.events, (e) -> e.id < data.events.data[0].id
```
nanaya: this doesn't seem to be much faster than what I initially had since it still loops through the whole array ┐(･_･ )┌

nanaya: oh right. I got it inverted.

resources/assets/coffee/react/mp-history/game.coffee
```diff
@@ -50,7 +50,7 @@ class MPHistory.Game extends React.Component
             lookupUser: @props.lookupUser
             key: m.slot
 
-      if showTeams
+      if showTeams && @props.event.game.data.end_time
```
nanaya: this is the other optional missing `?`, btw. (though realistically speaking it should never be 0 - hence optional)

PR #728, opened 2016-09-06
Soft deleting forum posts

### LINKED ISSUE #627: Ability to soft delete forum posts
Rather than implementing "split thread", we should have a moderator function to toggle hiding posts from the public in a thread.

LiquidPL:
Resolves #627.


app/Http/routes.php
```diff
@@ -151,6 +151,7 @@
     Route::get('p/{posts}', ['as' => 'forum.posts.show', 'uses' => 'PostsController@show']);
     Route::get('posts/{posts}/raw', ['as' => 'forum.posts.raw', 'uses' => 'PostsController@raw']);
     Route::resource('posts', 'PostsController', ['only' => ['destroy', 'update', 'edit']]);
+    Route::post('posts/{posts}/hide', ['as' => 'forum.posts.hide', 'uses' => 'PostsController@hide']);
```
peppy: why not add this to the `PostsController` definition in the previous line?

app/Libraries/OsuAuthorize.php
```diff
@@ -414,6 +414,16 @@ public function checkForumTopicVote($user, $topic)
         return 'ok';
     }
 
+    public function checkHiddenForumPostView($user)
+    {
+        $this->ensureLoggedIn($user);
+
+        // not sure if this is enough :/
+        if ($user->isGMT()) {
```
peppy: can't use existing function like `checkForumTopicModerate`?

app/Models/Forum/Topic.php
```diff
@@ -343,7 +343,13 @@ public function pollEnd()
     public function postsCount()
     {
         if ($this->postsCount === null) {
-            $this->postsCount = $this->posts()->count();
+            $query = $this->posts();
+
+            if (priv_check('HiddenForumPostView')->can()) {
+                $query->withTrashed();
```
peppy: not sure this should be included in the post count.

LiquidPL: without this `count()` would only see the visible posts, breaking stuff for moderators who can see hidden posts

app/Http/routes.php
LiquidPL: mostly because the resource controller doesn't provide a sensible route for something like *hiding* a post (i mean i could theoretically add it to destroy as we want to shift away from actually deleting posts, but that just doesn't sound right)

LiquidPL:
should the post author be able to hide their own posts (as it is right now with deleting)?

peppy:
as long as there's no option to unhide. it should be visible as *delete* the the end user (and even admins). unhide should probably be database only, or using a special interface.

peppy:
let's leave it visible for admins, but the wording should be "delete" and "undelete" and there should be no actual *hard* delete.

users should only see delete.

LiquidPL:
what happens if all posts are 'deleted'? should the topic be hidden too?

peppy:
definitely

peppy:
need to resolve conflicts with master here

LiquidPL:
done

app/Models/Forum/Topic.php
```diff
@@ -60,6 +64,19 @@ class Topic extends Model
         'topic_approved' => 'boolean',
     ];
 
+    protected static function boot()
+    {
+        parent::boot();
+
+        if (!priv_check('ForumTopicModerate')->can()) {
```
nanaya: see previous comment on current user and model

resources/assets/coffee/forum/topic-ajax.coffee
```diff
@@ -19,23 +19,44 @@ $(document).on 'ajax:success', '.delete-post-link', (_event, data) ->
...
-      osu.pageChange()
+  if currentUser.isAdmin || currentUser.isGMT
+    $post = $el.find '.forum-post'
+    toggle = _event.target
+
+    switch toggle.id
+      when 'delete'
+        $post.addClass 'forum-post--hidden'
+        toggle.id = 'undelete'
+        $(toggle).children()
```
nanaya: just get the controller action to return the new button instead.

resources/assets/coffee/forum/topic-ajax.coffee
```diff
@@ -19,23 +19,44 @@ $(document).on 'ajax:success', '.delete-post-link', (_event, data) ->
...
+      when 'undelete'
+        $post.removeClass 'forum-post--hidden'
+        toggle.id = 'delete'
+        $(toggle).children()
+          .removeClass 'fa-undo'
+          .addClass 'fa-trash'
+
+    $(toggle).prop 'title', osu.trans "forum.post.actions.#{toggle.id}"
+    $(toggle).data 'confirm', osu.trans "forum.post.confirm_#{toggle.id}"
+  else
```
nanaya: `currentHeight` should be inside this since it's not needed in the other one

database/migrations/2016_09_06_124728_add_soft_deletes_to_posts_topics_tables.php
```diff
@@ -0,0 +1,39 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('phpbb_posts', function (Blueprint $table) {
+            $table->softDeletes();
```
nanaya: that would be confusing. Something will be done for this... I think. Just add the index for now.

resources/assets/coffee/forum/topic-ajax.coffee
```diff
@@ -19,23 +19,44 @@ $(document).on 'ajax:success', '.delete-post-link', (_event, data) ->
...
-        post.setAttribute 'data-post-position', originalPosition - 1
-
-      osu.pageChange()
+  if currentUser.isAdmin || currentUser.isGMT
+    $post = $el.find '.forum-post'
+    toggle = _event.target
+
+    switch toggle.id
+      when 'delete'
+        $post.addClass 'forum-post--hidden'
```
LiquidPL: i probably shouldn't be renaming a whole [bem class](https://github.com/ppy/osu-web/blob/master/resources/assets/less/bem/forum-post.less)

also this isn't really js-related, just applied in js

app/Models/Forum/Post.php
```diff
@@ -43,6 +47,19 @@ class Post extends Model
 
     private $normalizedUsers = [];
 
+    protected static function boot()
+    {
+        parent::boot();
+
+        if (!priv_check('ForumTopicModerate')->can()) {
```
LiquidPL: can I do it so that every query that originates from the controller will be unscoped? otherwise I'll probably have to conditionally add `withTrashed()` to every single relevant query 😕

nanaya: Last I checked it's only needed in `TopicController#show`, `PostsController#destroy`, and the post position thingy (Topics's `nthPost`, `postPosition` and `postsPosition`).

The rest (cache update, topic move, etc) seem to require explicit with/without deleted post regardless of current user.

LiquidPL:
should be done~

app/Models/Forum/Topic.php
```diff
@@ -142,6 +145,31 @@ public function removePost($post, $user = null)
         return true;
     }
 
+    public function restorePost($post, $user = null)
+    {
+        DB::transaction(function () use ($post, $user) {
```
nanaya: `$user` not used? (though maybe should be used to log the action)

app/Models/Forum/Topic.php
```diff
@@ -313,12 +341,24 @@ public function scopeRecent($query, $params = null)
...
+
+        return $posts->skip(intval($n) - 1)->first();
     }
 
     public function postPosition($postId)
     {
-        return $this->posts()->where('post_id', '<=', $postId)->count();
+        $posts = $this->posts();
+
+        if (priv_check('ForumTopicModerate')->can()) {
```
nanaya: same as above.

app/Models/Forum/Topic.php
```diff
@@ -357,7 +397,13 @@ public function pollEnd()
     public function postsCount()
     {
         if ($this->postsCount === null) {
-            $this->postsCount = $this->posts()->count();
+            $posts = $this->posts();
+
+            if (priv_check('ForumTopicModerate')->can()) {
```
nanaya: same as above.

app/Models/Forum/Topic.php
```diff
@@ -313,12 +341,24 @@ public function scopeRecent($query, $params = null)
...
+        $posts = $this->posts();
+
+        if (priv_check('ForumTopicModerate')->can()) {
+            $posts->withTrashed();
+        }
+
+        return $posts->where('post_id', '<=', $postId)->count();
     }
 
     public function postsPosition($sortedPosts)
```
nanaya: this probably can be removed to save passing around the `$withTrashed` variable (just increment the first post's `postPosition` in the view).

LiquidPL:
done

app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -228,11 +237,17 @@ public function show($id)
...
+        // position of the first post, incremented in the view
+        // to generate positions of further posts
+        $postPosition = $topic->postPosition($topic->posts->first()->post_id);
 
         $pollSummary = PollOption::summary($topic, Auth::user());
 
-        Event::fire(new TopicWasViewed($topic, $posts->last(), Auth::user()));
+        Event::fire(new TopicWasViewed(
+            $topic,
+            $topic->posts->last(),
```
nanaya: should be `$posts->last()`

app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -167,21 +167,30 @@ public function show($id)
             ::with([
                 'forum.cover',
                 'pollOptions.votes',
-            ])
-            ->findOrFail($id);
+            ]);
+
+        if ($canModerate = priv_check('ForumTopicModerate')->can()) {
```
nanaya: `$withTrashed`?

app/Libraries/OsuAuthorize.php
```diff
@@ -354,7 +355,9 @@ public function checkForumPostDelete($user, $post)
         }
 
         $position = $post->postPosition;
-        $topicPostsCount = $post->topic->postsCount();
+        $topicPostsCount = $post->topic->postsCount(
+            $this->doCheckUser($user, 'ForumTopicModerate', $post->topic)->can()
```
nanaya: assuming without trashed here should be safe.

resources/views/forum/_search.blade.php
```diff
@@ -66,7 +66,12 @@ class="forum-search-logo js-forum-search-button u-forum--bg-link"
                                 <i class="fa fa-angle-right"></i>
                             </button>
                         </form>
-                        {{ trans("forum.search.total_posts", ["posts_count" => $topic->postsCount()]) }}
+                        {{
+                            trans(
+                                "forum.search.total_posts",
+                                ["posts_count" => $topic->postsCount(priv_check('ForumTopicModerate')->can())]
```
nanaya: just create and pass `$postsCount`

app/Models/Forum/Topic.php
```diff
@@ -455,7 +484,7 @@ public function setFirstPostCache()
 
     public function setLastPostCache()
     {
-        $lastPost = $this->posts()->last()->first();
+        $lastPost = $this->posts()->orderBy('post_id', 'desc')->first();
```
LiquidPL: last time I checked the original version didn't work and just returned the first element anyways

nanaya:
looks like it's easier to just ignore deleted post from all the post position/counter thingy?

resources/views/forum/_search.blade.php
LiquidPL: is this blade actually used anywhere? can't seem to find anything that might reference it

nanaya: I don't think so. The only functionality it has has been replaced by the always visible footer bar.

LiquidPL:
seems like that, but only thing that would have to be changed is loading more posts, as right now it tries to do so when the loaded posts count is *different* than the total post count, which breaks the page for moderators where the it will usually be larger

LiquidPL:
and also the thing that some topics will display like 9/7 for moderators

nanaya:
that shouldn't happen if ignoring the deleted posts in counter.

LiquidPL:
i think that this behavior actually makes some sense, as it allows mods to see if there are any deleted posts in the topic (i suppose)

LiquidPL:
just a small reminder that this is ready to be looked through

app/Http/Controllers/Forum/ForumsController.php
```diff
@@ -62,8 +62,17 @@ public function show($id)
...
         );
 
-        $pinnedTopics = $forum->topics()->pinned()->orderBy('topic_type', 'desc')->recent()->get();
-        $topics = $forum->topics()->normal()->recent(compact('sort', 'withReplies'))->paginate(15);
+        $pinnedTopics = $forum->topics();
+        $topics = $forum->topics();
+
+        if (priv_check('ForumTopicModerate')->can()) {
+            $pinnedTopics->withTrashed();
+            $topics->withTrashed();
```
nanaya: indicator for trashed topics in index?

LiquidPL:
should be ready to be looked through again 👌

resources/views/forum/topics/replace_delete_button.blade.php
```diff
@@ -0,0 +1,64 @@
...
+    See the GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+--}}
+Timeout.set(0, function () {
+    $el = $(".js-forum-post[data-post-id={{ $post->post_id }}]")
+
+    $toggle = $el.find(".delete-post-link");
+    action = $toggle.attr('data-action');
```
nanaya: or just use the actual php variable.

resources/views/forum/topics/replace_delete_button.blade.php
```diff
@@ -0,0 +1,64 @@
...
+
+    if (currentUser.isAdmin || currentUser.isGMT) {
```
nanaya: here as well, etc

app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -233,11 +234,19 @@ public function show($id)
             abort($skipLayout ? 204 : 404);
         }
 
-        $postsPosition = $topic->postsPosition($posts);
+        $firstPostId = $posts->first()->post_id;
```
nanaya: is supposed to be this first post of the current thread or just of the currently loaded posts?

app/Models/Forum/Post.php
```diff
@@ -169,6 +172,15 @@ public function getBodyRawAttribute()
...
+        return $query->orderBy('post_id', 'desc')->limit(1);
+    }
+
+    public function scopeShowDeleted($query, $showDeleted)
+    {
+        if ($showDeleted) {
+            $query->withTrashed();
+        }
+
+        return $query;
```
nanaya: pretty sure the return isn't needed.

app/Http/Controllers/Forum/TopicsController.php
LiquidPL: thread, I'll fix it

app/Http/Controllers/Forum/PostsController.php
```diff
@@ -32,35 +32,39 @@ class PostsController extends Controller
...
 
-        priv_check('ForumPostDelete', $post)->ensureCan();
+        $post = Post::query()->showDeleted($showDeleted)->findOrFail($id);
+        $topic = $post->topic()->showDeleted($showDeleted)->first();
 
-        $deletedPostPosition = $post->topic->postPosition($post->post_id);
+        $action = Request::input('action');
 
-        $post->topic->removePost($post, Auth::user());
+        priv_check('ForumPostDelete', $post)->ensureCan();
```
nanaya: this is a bit confusing because the position here means restore action also checks for `Delete`.

LiquidPL: so should i make a separate permission for restoring posts?

nanaya: might as well split the action?

app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -233,11 +234,22 @@ public function show($id)
...
 
-        $postsPosition = $topic->postsPosition($posts);
+        $firstPostId = $topic->posts()
+            ->showDeleted($showDeleted)
+            ->orderBy('post_id', 'asc')
+            ->first()->post_id;
+
+        // position of the first post, incremented in the view
+        // to generate positions of further posts
+        $firstPostPosition = $topic->postPosition($firstPostId);
```
nanaya: this probably should be position of the first loaded position?

(also there seems to be something wrong in "position of first post of the topic in the topic". Something with "first")

app/Models/Forum/Topic.php
```diff
@@ -253,6 +284,15 @@ public function scopeNormal($query)
         return $query->where('topic_type', 0);
     }
 
+    public function scopeShowDeleted($query, $showDeleted)
+    {
+        if ($showDeleted) {
+            $query->withTrashed();
+        }
+
+        return $query;
```
nanaya: not needed as well

app/Models/Forum/Topic.php
```diff
@@ -315,33 +355,16 @@ public function scopeRecent($query, $params = null)
 
     public function nthPost($n)
     {
-        return $this->posts()->skip(intval($n) - 1)->first();
+        $posts = $this->posts();
```
nanaya: back to initial version?

app/Models/Forum/Topic.php
```diff
@@ -369,12 +392,21 @@ public function pollEnd()
     public function postsCount()
     {
         if ($this->postsCount === null) {
-            $this->postsCount = $this->posts()->count();
+            $this->postsCount = $this->posts->count();
```
nanaya: shouldn't this ignore the deleted posts?

nanaya: looking again, what's the change for?

resources/assets/less/bem/forum-post.less
```diff
@@ -40,6 +40,10 @@
     margin-bottom: 0;
   }
 
+  .js-forum-post-hidden& {
```
nanaya: normally written the other way around (`&.and-another-class`)

app/Http/Controllers/Forum/PostsController.php
LiquidPL: dunno, this feels like duplicating all the code retrieving models and rendering the view

nanaya: Apparently there doesn't seem to be much duplication, though?

```php
public function destroy($id)
{
    $post = Post::findOrFail($id);

    priv_check('ForumPostDelete', $post)->ensureCan();

    $deletedPostPosition = $post->topic->postPosition($post->post_id);
    $post->topic->removePost($post, Auth::user());

    if ($post->topic->isTrashed()) {
        $redirect = route('forum.forums.show', $post->forum);
        return ujs_redirect($redirect);
    }

    return js_view('forum.topics.destroy', compact('post', 'deletedPostPosition'));
}

public function restore($id)
{
    priv_check('ForumTopicModerate')->ensureCan();

    $post = Post::withTrashed()->findOrFail($id);
    $topic = $post->topic()->withTrashed()->first();
    $topic->restorePost($post, Auth::user());

    return js_view('forum.topics.restore', compact('post'));
}
```

app/Models/Forum/Topic.php
LiquidPL: strange, on my computer it was like this earlier:
```php
$posts = $this->posts()

$this->postsCount = $posts->count()
```

app/Http/Controllers/Forum/TopicsController.php
LiquidPL: sorry, I don't really understand what you mean there

LiquidPL:
did I miss something there?

app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -146,12 +146,12 @@ public function reply(HttpRequest $request, $id)
...
         if ($post->post_id !== null) {
             $posts = collect([$post]);
-            $postsPosition = $topic->postsPosition($posts);
+            $postPosition = $topic->postPosition($post->post_id);
 
             Event::fire(new TopicWasReplied($topic, $post, Auth::user()));
             Event::fire(new TopicWasViewed($topic, $post, Auth::user()));
 
-            return view('forum.topics._posts', compact('posts', 'postsPosition', 'topic'));
+            return view('forum.topics._posts', compact('posts', 'postPosition', 'topic'));
```
nanaya: should be `firstPostPosition`

resources/views/forum/topics/replace_delete_button.blade.php
```diff
@@ -0,0 +1,56 @@
...
+
+    window.forum.setTotalPosts(window.forum.totalPosts() + countDifference);
+    window.forum.setDeletedPosts(window.forum.deletedPosts() - countDifference);
+
+    for (i = window.forum.posts.length - 1; i >= 0; i--) {
+        post = window.forum.posts[i];
+
+        originalPosition = forum.postPosition(post);
+
+        if (originalPosition < {{ $post->topic->postPosition($post->post_id) }}) {
```
nanaya: the counter is incorrect. When doing delete, the post above is decremented as well. Meanwhile on restore, restored post doesn't get incremented.

resources/views/forum/topics/replace_delete_button.blade.php
```diff
@@ -0,0 +1,56 @@
...
+            return $el.remove();
+        });
+    @endif
+
+    countDifference = {{ Route::currentRouteName() === 'forum.posts.delete' ? -1 : 1 }};
+
+    window.forum.setTotalPosts(window.forum.totalPosts() + countDifference);
+    window.forum.setDeletedPosts(window.forum.deletedPosts() - countDifference);
+
+    for (i = window.forum.posts.length - 1; i >= 0; i--) {
```
nanaya: looking again, it's probably easier to do this first before removing/restoring the post and break when post id matches.

LiquidPL:
whew, should be done

resources/views/forum/topics/delete.blade.php
```diff
@@ -0,0 +1,26 @@
...
+    You should have received a copy of the GNU Affero General Public License
+    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+--}}
+@extends('forum.topics.replace_delete_button')
+
+@section('moderatorAction')
+    $el.addClass("js-forum-post--hidden");
+@endsection
+
+@section('countDifference')
```
nanaya: plain php variable?

resources/views/forum/topics/replace_delete_button.blade.php
```diff
@@ -0,0 +1,54 @@
...
+    @yield('countDifference')
+
+    for (i = window.forum.posts.length - 1; i >= 0; i--) {
+        var post = window.forum.posts[i];
+
+        var position = forum.postPosition(post);
+
+        post.setAttribute("data-post-position", position + countDifference);
+
+        if (post.getAttribute('data-post-id') == {{ $post->post_id }}) {
```
nanaya: `forum.postId`? And comparison should be strict

resources/views/forum/topics/replace_delete_button.blade.php
```diff
@@ -0,0 +1,54 @@
...
+    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+    See the GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+--}}
+Timeout.set(0, function () {
+    @yield('countDifference')
+
+    for (i = window.forum.posts.length - 1; i >= 0; i--) {
```
nanaya: `var i`

app/Http/Controllers/Forum/ForumsController.php
```diff
@@ -62,8 +62,14 @@ public function show($id)
...
         );
 
-        $pinnedTopics = $forum->topics()->pinned()->orderBy('topic_type', 'desc')->recent()->get();
-        $topics = $forum->topics()->normal()->recent(compact('sort', 'withReplies'))->paginate(15);
+        $pinnedTopics = $forum->topics();
+        $topics = $forum->topics();
+
+        $showDeleted = priv_check('ForumTopicModerate')->can();
+
+        $pinnedTopics = $pinnedTopics->pinned()->showDeleted($showDeleted)->orderBy('topic_type', 'desc')->recent()->get();
```
nanaya: just use `$forum->topics()` directly?

app/Http/Controllers/Forum/PostsController.php
```diff
@@ -41,26 +41,37 @@ public function __construct()
 
     public function destroy($id)
     {
-        $post = Post::findOrFail($id);
+        $post = Post::query()
```
nanaya: `Post::showDeleted` works

app/Http/Controllers/Forum/PostsController.php
```diff
@@ -41,26 +41,37 @@ public function __construct()
...
             $redirect = route('forum.forums.show', $post->forum);
 
             return ujs_redirect($redirect);
         }
 
-        return [
-            'postId' => $post->post_id,
-            'postPosition' => $deletedPostPosition,
-        ];
+        $countDifference = -1;
```
nanaya: why not just during extend in the main view.

app/Http/Controllers/Forum/PostsController.php
```diff
@@ -32,7 +32,7 @@ class PostsController extends Controller
     public function __construct()
     {
         $this->middleware('auth', ['only' => [
-            'destroy',
+            'changeVisibility',
```
nanaya: wrong action

app/Http/Controllers/Forum/PostsController.php
```diff
@@ -85,9 +96,9 @@ public function update($id)
 
         $posts = collect([$post->fresh()]);
         $topic = $post->topic;
-        $postsPosition = $topic->postsPosition($posts);
+        $postPosition = $topic->postPosition($post->post_id);
 
-        return view('forum.topics._posts', compact('posts', 'postsPosition', 'topic'));
+        return view('forum.topics._posts', compact('posts', 'postPosition', 'topic'));
```
nanaya: should be `firstPostPosition`

app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -233,11 +234,24 @@ public function show($id)
             abort($skipLayout ? 204 : 404);
         }
 
-        $postsPosition = $topic->postsPosition($posts);
+        $firstPostId = $topic->posts()
+            ->showDeleted($showDeleted)
+            ->orderBy('post_id', 'asc')
```
nanaya: `->min('post_id')` probably works as well.

app/Http/Kernel.php
```diff
@@ -38,7 +38,7 @@ class Kernel extends HttpKernel
         Middleware\VerifyCsrfToken::class,
         Middleware\SetLocale::class,
         Middleware\AutologinFromLegacyCookie::class,
-        Middleware\VerifyPrivilegedUser::class,
+        // Middleware\VerifyPrivilegedUser::class,
```
nanaya: still

app/Http/Controllers/Forum/PostsController.php
LiquidPL: wait, so where exactly, because i'm not sure what do you mean here.

nanaya: @extends('some.view', ['some' => 'variable'])

resources/views/forum/topics/replace_delete_button.blade.php
```diff
@@ -0,0 +1,54 @@
...
+
+    osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+    See the GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+--}}
+Timeout.set(0, function () {
+    var countDifference = {{ $countDifference }};
```
nanaya: also I mean instead of this just inline the php variable.

resources/views/forum/topics/_header_counter.blade.php
```diff
@@ -34,3 +34,22 @@
...
 </div>
+@if(priv_check('ForumTopicModerate', $topic)->can())
+    <div class="counter-box counter-box--forum">
+        <div class="counter-box__content">
+            <div class="counter-box__title">
+                {{ trans('forum.topics.show.deleted-posts') }}
+            </div>
+
+            <div class="counter-box__count js-forum__deleted-count">
+                @if ($newTopic)
```
nanaya: not sure why I used separate span for the total. Can you change it as well? Also inlining it (`{{ $newTopic ? 0 : $... }}`) probably isn't too long

resources/views/forum/topics/_header_counter.blade.php
```diff
@@ -15,22 +15,31 @@
...
     along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
 --}}
-<div class="counter-box">
+<div class="counter-box counter-box--forum">
     <div class="counter-box__content">
         <div class="counter-box__title">
             {{ trans('forum.topics.show.total_posts') }}
         </div>
 
         <div class="counter-box__count">
```
nanaya: the js class should be moved here.

resources/assets/less/bem/forum-category-header.less
```diff
@@ -111,6 +111,7 @@
     margin-left: 10px;
     display: flex;
     align-items: flex-end;
+    flex-direction: column;
```
nanaya: this probably shouldn't be changed. Also the `titles` element probably need an `align-items: flex-end`

resources/views/forum/topics/_post_hide_action.blade.php
```diff
@@ -0,0 +1,39 @@
+{{--
+    Copyright 2016 ppy Pty. Ltd.
```
nanaya: need update

PR #731, opened 2016-09-16
Optimise Voting

nekodex:
Let's not fetch all the contest's votes just to count the number of votes you've made 👍


app/Models/Contest.php
```diff
@@ -56,7 +56,7 @@ public function vote(User $user, ContestEntry $entry)
             $vote->delete();
         } else {
             // there's probably a race-condition here, but abusing this just results in the user diluting their vote... so *shrug*
-            if ($this->votes->where('user_id', $user->user_id)->count() < $this->max_votes) {
+            if ($this->votes()->where('user_id', $user->user_id)->count() < $this->max_votes) {
```
peppy: help

PR #735, opened 2016-09-20
Prepare framework for contest submission page.

peppy:
This doesn't actually add upload logic, but it does locally split the contest system in a way that allows for entry (submission) and voting (plus results).

Should be a clean merge from which we can add more.

### DIFF
--- a/app/Http/Controllers/ContestsController.php
+++ b/app/Http/Controllers/ContestsController.php
@@ -37,9 +37,13 @@ class ContestsController extends Controller
             abort(404);
         }
 
-        return view("contests.{$contest->type}")
-                ->with('contest', $contest)
-                ->with('contestJson', $contest->defaultJson($user));
+        if ($contest->isShowingEntries()) {
+            return view("contests.voting.{$contest->type}")
+                    ->with('contest', $contest)
+                    ->with('contestJson', $contest->defaultJson($user));
+        } else {
+            return view('contests.enter')->with('contest', $contest);
+        }
     }
 
     public function vote($id)
@@ -53,5 +57,16 @@ class ContestsController extends Controller
         $contest->vote($user, $entry);
 
         return $contest->defaultJson($user);
+    }
+
+    public function submit($id)
+    {
+        $user = Auth::user();
+        $contest = Contest::findOrFail($id);
+
+        priv_check('ContestEnter', $contest)->ensureCan();
+
+        //todo: upload accept logic
+        abort(501);
     }
 }
--- a/app/Http/Controllers/ContestsController.php
+++ b/app/Http/Controllers/ContestsController.php
@@ -37,7 +37,7 @@ class ContestsController extends Controller
             abort(404);
         }
 
-        if ($contest->isShowingEntries()) {
+        if ($contest->isVotingStarted()) {
             return view("contests.voting.{$contest->type}")
                     ->with('contest', $contest)
                     ->with('contestJson', $contest->defaultJson($user));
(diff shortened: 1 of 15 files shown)


app/Http/Controllers/ContestsController.php
```diff
@@ -37,9 +37,13 @@ public function show($id)
             abort(404);
         }
 
-        return view("contests.{$contest->type}")
-                ->with('contest', $contest)
-                ->with('contestJson', $contest->defaultJson($user));
+        if ($contest->isShowingEntries()) {
```
nekodex: Hrm the name `isShowingEntries` feels kinda ambiguous in this context

app/Transformers/ContestTransformer.php
```diff
@@ -37,7 +37,7 @@ public function transform(Contest $contest)
             'type' => $contest->type,
             'header_url' => $contest->header_url,
             'max_votes' => $contest->max_votes,
-            'ends_at' => $contest->ends_at->toIso8601String(),
+            'voting_ends_at' => $contest->voting_ends_at->toIso8601String(),
```
nekodex: I think you're missing the migration for this?

app/Libraries/OsuAuthorize.php
```diff
@@ -209,12 +209,24 @@ public function checkChatChannelRead($user, $channel)
         return $prefix.'no_access';
     }
 
+    public function checkContestEnter($user, $contest)
+    {
+        $this->ensureLoggedIn($user);
+        $this->ensureCleanRecord($user);
+
+        if ($contest->isSubmissionOpen()) {
+            return 'contest.voting_over';
```
nekodex: `contest.voting.over`?

app/Libraries/OsuAuthorize.php
```diff
@@ -209,12 +209,24 @@ public function checkChatChannelRead($user, $channel)
         return $prefix.'no_access';
     }
 
+    public function checkContestEnter($user, $contest)
+    {
+        $this->ensureLoggedIn($user);
+        $this->ensureCleanRecord($user);
+
+        if ($contest->isSubmissionOpen()) {
```
nekodex: Is this supposed to be `!$contest->isSubmissionOpen()`?

resources/views/contests/voting.blade.php
```diff
@@ -0,0 +1,36 @@
...
+    You should have received a copy of the GNU Affero General Public License
+    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+--}}
+@extends('contests.base')
+
+@section('contest-content')
+    <div class="contest__description">{!! Markdown::convertToHtml($contest->description_voting) !!}</div>
+
+    @if ($contest->voting_ends_at->isPast())
+        <div class='contest__voting-ended'>{{trans('contest.over')}}</div>
```
nekodex: `contest.voting.over`?

app/Http/Controllers/ContestsController.php
peppy: got a suggestion for a better name? it's a hard one since it encapsulates voting, post-voting and results

app/Transformers/ContestTransformer.php
peppy: oops

nekodex: Also if you change this, you'll probably need to update the coffees...

app/Http/Controllers/ContestsController.php
nekodex: Hrm maybe `submissionsClosed` or something? Not sure.

peppy: that could mean pre-submission too :(

nekodex:
Now that you've made those columns nullable, might want to add null checks to where you're using the dates (e.g. `entry_starts_at` in `enter.blade.php`)?

Causes the contest page to explode ATM due to the migration adding those columns to existing contests and then the page attempting to `->isPast` on null

PR #737, opened 2016-09-20
Logout and code reissuing on verification

### LINKED ISSUE #733: No way to logout or re-request verification email
image

Should at very least be able to logout and try again.

NeoAdonis:
Resolves #733

### DIFF
--- a/app/Libraries/UserVerification.php
+++ b/app/Libraries/UserVerification.php
@@ -134,6 +134,17 @@ class UserVerification
         return response([], 200);
     }
 
+    public function reissue()
+    {
+        if ($this->isDone()) {
+            return $this->verified();
+        }
+
+        $email = $this->user->user_email;
+        $this->issue();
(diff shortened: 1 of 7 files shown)


app/Libraries/UserVerification.php
```diff
@@ -134,6 +134,17 @@ public function verified()
         return response([], 200);
     }
 
+    public function reissue()
+    {
+        if ($this->isDone()) {
+            return $this->verified();
+        }
+
+        $email = $this->user->user_email;
```
peppy: not sure what this is doing here, but probably not what you intended.

NeoAdonis: Sorry. Saw line 43 and thought you needed to store email for issuing (now I see it's to show to the user). FIxing along with code styling.

peppy:
I'm merging this because it is pissing me off (i can't get into my own account at the moment).

@nanaya you may want to look through these changes when you're next available.

PR #744, opened 2016-10-04
Ranking boards [WIP]

NeoAdonis:
osuweb-rankings

I've implemented what it could be called skeletons for the overall, country and mappers ranking boards. Opened pull request and labeled as a Work In Progress because I need a bit of guidance regarding page layouts, but hopefully these additions will serve as a good base.

The layouts are initially based on the current ranking boards. I assume the osu!team has already sketches of the expected layouts; if so, I ask if they can provide them so I can complete the layout, otherwise I encourage you to give me suggestions about what should be added or modified.

Regarding the filtering by friends functionality, I've implemented it on backend, but not yet on frontend. What I need is to trigger an `onChange` event that triggers the page reload with the filter toogled; easily done by using the `onchange` tag of the checkbox, but I assume this is not the best way of doing it when there's React and Elixir, so any suggestion regarding this issue is appreciated!

Pending tasks:
- [ ] Adjust layouts.
- [ ] Modify strings (many of these are just placeholders).
- [ ] Checkbox to allow supporters filter by friends.
- [ ] Add descriptions below headers where necessary.
- [ ] Add and assign unique backgrounds for headers (currently using the 'tournament' one).
- [ ] Modify nav menu links to redirect to these boards.

Also I didn't implemented ranking chart scoreboards as I couldn't find the Models or tables on DB corresponding such data.


lunafreya:
You should have to let @arflyte deal with the whole design, what you did was deviate from his original vision.

NeoAdonis:
Yeah, sorry. Actually, I assume there's a already designed page layout for these pages, one of the reasons I created this PR: ask for those designs if they are done. I don't think the scoreboards per se can deviate too much from what is planned (those are based on the beatmap scoreboards), neither the backend, so whatever changes are needed, I probably could work from these.

nanaya:
Done by @nekodex.

PR #746, opened 2016-10-04
many translations to pt-br

HWSKY:
hi, i decided to delete my fork and create another one with just the translations files to avoid another pull constraints. 
all files included was translated according to /lang/en/ folder


HWSKY:
Let's see if now will be possible to send a successful commit, my past fork has many translations mixed to controls php files, if this one be passed, i will input the others too.
I'm so excited.

PR #753, opened 2016-10-07
Full French translation

leonekmi:
Hello,

_Micro-edit : I have fixed the StyleCI check issue_

I have made a full translation of osu!web in French.
_(Bonjour, je suis France ^^)_

Cordially,

leonekmi


peppy:
Nice job 👍

PR #760, opened 2016-10-12
Forum username clipping

### LINKED ISSUE #28: Long nicknames consisting of one word get cut
As Sulker@Slack noticed, nicks bigger than the current limit will get cut off on jizz forums, looking quite bad.

Short-term solution is easy: change the infobox from 120px to 140px. I don't know if there's people with bigger nicks than me, and if it would work in them, though.

nanaya:
Fixes #28.

Also included a bunch of refactors and fix for mobile.


peppy:
👍

PR #774, opened 2016-10-15
Small fixes 2

UnusualToast:
Also team-types (lines 24-29) ideally should be translated, but it's hard to translate those terms. Might be a good idea to leave it as is?


HWSKY:
good revision, i recommend your changes

resources/lang/pt-br/multiplayer.php
```diff
@@ -28,18 +28,18 @@
...
         ],
         'events' => [
-            'player-left' => ':user Saiu da partida',
-            'player-joined' => ':user Juntou-se a partida',
+            'player-left' => ':user saiu da partida',
+            'player-joined' => ':user juntou-se à partida',
             'player-kicked' => ':user foi expulso da partida',
             'match-created' => ':user criou a partida',
-            'match-disbanded' => ':the partida foi dissolvida',
+            'match-disbanded' => ':the partida foi cancelada',
```
HWSKY: sounds like much better

PR #777, opened 2016-10-15
Beatmapsets fixes

LiquidPL:
* decrease preview volume
* add favouriting beatmapsets (soon)


resources/assets/coffee/react/beatmapset-page/main.coffee
```diff
@@ -166,6 +166,7 @@ class BeatmapsetPage.Main extends React.Component
     @setCurrentScoreboard null, scoreboardType: 'global', resetMods: true
 
     @audioPreview = document.getElementsByClassName('js-beatmapset-page--audio-preview')[0]
+    @audioPreview.volume = 0.45
```
nanaya: probably better as attribute in the element itself? Or does it not work?

LiquidPL: apparently you can't set volume in html, so yes

PR #799, opened 2016-10-27
Ｌａｒａｖｅｌ　５．３

nanaya:
Missing:

- migration to `passport`
- upgrade Elixir to 6.x


public/index.php
```diff
@@ -3,7 +3,7 @@
 /**
  * Laravel - A PHP Framework For Web Artisans.
  *
- * @author   Taylor Otwell <taylorotwell@gmail.com>
+ * @author   Taylor Otwell <taylor@laravel.com>
```
peppy: 👍

PR #808, opened 2016-10-28
[WIP] Add Laravel Passport

ameliaikeda:
Simple migration; to create a new password-based grant client, just `php artisan passport:client --password` and edit it to use the same secret as what's currently in use.

Alternatively, it could be cached in redis before the tables are dropped.

(Note: still needs actual testing against the osu! client)


nanaya:
Haven't gotten around finishing this either but I think it works with this diff applied:

https://0paste.com/9467-27df01f2.txt

ameliaikeda:
I'll get around to this tonight or tomorrow, but that `Passport::routes()`
call will change the oauth endpoint to /oauth/token.

nanaya:
I think that's fine considering it's not in use yet anyway. The alternative I can think of is

- create its own route file to avoid App namespacing
- copy the controller as well
- ?

ameliaikeda:
Oh, well if it's not in use then go right for the default passport routes
declaration.

PR #818, opened 2016-10-31
Fix yarn installation failing because of missing permissions

LiquidPL:


LiquidPL:
wait no I actually realized that it needs to be called in `build.sh` for production deployment ;_;

how about just install it locally or something?

build.sh
```diff
@@ -26,6 +26,5 @@ php artisan migrate --force
 php artisan lang:js resources/assets/js/messages.js
 php artisan laroute:generate
 
-npm install -g yarn
```
nanaya: just replace with this instead `command -v yarn || npm install -g yarn`.

nanaya:
also, can you replace `APP_KEY` value with a valid string in `.env.example` while at it?

PR #828, opened 2016-11-04
add topic watch migration

### LINKED ISSUE #823: phpbb_topics_watch migration missing
used in `app/Models/Forum/TopicWatch.php`

kschamplin19:
fixes #823 

if anything I messed up the index things I forgot how to database.


nekodex:
Hey, thanks for the contribution. 😃

However, legacy tables (`phpbb_*` and `osu_*`) shouldn't be added as 'normal' migrations - they'll fail when run during our production deploy (because the tables already exist in production).

Instead they should be added to the base table migrations
(see: `database/migrations/2015_01_01_133337_base_tables.php`)

nekodex:
Columns shouldn't be nullable, also make sure to set the charset/collation as well as the row format too (see the other migrations in that base table file for reference).

database/migrations/2015_01_01_133337_base_tables.php
```diff
@@ -1352,6 +1352,15 @@ public function up()
         });
         $this->comment('osu_mod_queue', 'Data table for BanchoBot mod queue ~mm201');
         $this->setRowFormat('osu_mod_queue', 'DYNAMIC');
+
+        Schema::create('phpbb_topics_watch', function (Blueprint $table) {
+            $table->unsignedMediumInteger('user_id')->nullable();
+            $table->unsignedMediumInteger('topic_id')->nullable();
+            $table->unsignedTinyInteger('notify_status');
+            $table->index('topic_id');
```
LiquidPL: you might want to add index names too

PR #830, opened 2016-11-04
Use MySQL 5.7 on Travis-CI

nekodex:
We're using MySQL 5.7, so Travis probably should too...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,6 @@ cache:
     - node_modules
     - $HOME/.yarn-cache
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -4,8 +4,10 @@ cache:
     - node_modules
     - $HOME/.yarn-cache
 
-addons:
-  mariadb: 10.0
+services:
```
nanaya: might as well remove this if installing mysql manually?

PR #765, opened 2016-10-14
more content and translation to forum.php

HWSKY:
this file was originally translated according to /lang/en/forum.php root file
all translations and dependencies are fully included in this file since the last change, commit: 625f425 
sorry for the mess on pull requests, i will try send the single commits in single pulls

PR #832, opened 2016-11-04
Fun with pt-br

nanaya:
There are so many pull requests for this which are wrong at some level so I figured might as well fix them all at once. Someone who speaks the language may need to check it again. Or just merge it and let people complain =D

Closes #765.
Closes #767.
Closes #822.
Closes #813.
Closes #812.


Kuron-kun:
They all seem properly translated and the updated ones would just to translate some therms that would be more appropriate in portuguese. The only issue might be on the file name and some codes (copy-paste issues, maybe?), since I don't really know how it works... D:

Sadly we don't have many portuguese translators here, I guess.

PR #853, opened 2016-11-22
Beatmap stats in discussion

nanaya:
Recycling \o/

### DIFF
--- a/resources/assets/coffee/react/_components/beatmap-basic-stats.coffee
+++ b/resources/assets/coffee/react/_components/beatmap-basic-stats.coffee
@@ -1,0 +1,46 @@
+###
+# Copyright 2015-2016 ppy Pty. Ltd.
+#
+# This file is part of osu!web. osu!web is distributed with the hope of
+# attracting more community contributions to the core ecosystem of osu!.
+#
+# osu!web is free software: you can redistribute it and/or modify
+# it under the terms of the Affero GNU General Public License version 3
+# as published by the Free Software Foundation.
+#
+# osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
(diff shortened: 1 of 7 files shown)


resources/assets/coffee/react/_components/beatmap-basic-stats.coffee
```diff
@@ -0,0 +1,46 @@
...
+# along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+{div, span} = React.DOM
+
+bn = 'beatmap-basic-stats'
+
+@BeatmapBasicStats = ({beatmapset, beatmap}) ->
+  div
+    className: bn
+    for stat in ['total_length', 'bpm', 'count_circles', 'count_sliders']
```
HWSKY: Using enum will be more efficient?
`Count_circles` and` count_sliders` are not implemented yet. The following stats are standardized like the others included? I can implement this minor addition.

LiquidPL: >`count_circles` and `count_sliders` are not implemented yet

it's not that they are not implemented, these just don't need to be altered in any way before being displayed

also, [this](https://github.com/ppy/osu-web/blob/master/app/Transformers/BeatmapTransformer.php)

resources/assets/coffee/react/_components/beatmap-basic-stats.coffee
```diff
@@ -0,0 +1,46 @@
...
+      value =
+        if stat == 'bpm'
+          beatmapset.bpm
+        else
+          beatmap[stat]
+
+      value =
+        if stat == 'total_length'
+          moment(0).seconds(value).format 'm:ss'
```
HWSKY: If you're using the momentjs library, you could also use `'mm:ss'` for beatmaps longer than 9:59 or use `'h:mm:ss'` for longer marathons, use if case will be welcomed.

PR #858, opened 2016-11-24
Contest tweaks

nekodex:
Makes links in the contest description box less blue.

Also collapses all sections within nested contests by default. Contests having lots of entries resulted in the other sections being pushed off-screen. This led to confusing UX as users may not see the other headings and realise they're expandable/collapsable...

### DIFF
--- a/resources/assets/less/bem/contest.less
+++ b/resources/assets/less/bem/contest.less
@@ -36,7 +36,7 @@
     > h1, h2, h3, h4, h5, h6 {
       color: white;
     }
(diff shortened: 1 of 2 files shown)


resources/assets/less/bem/contest.less
```diff
@@ -36,6 +36,9 @@
     > h1, h2, h3, h4, h5, h6 {
       color: white;
     }
+    & a {
```
nanaya: the `&` isn't needed

PR #868, opened 2016-12-02
Delete discussion post

nanaya:
Limited functionality:
- ~~~only for replies~~~ now can delete discussion as well 
- only poster or admin can delete
- only admin can restore
- whoever did the last deletion is recorded (but not who restored it)
- event (resolve/unresolve) can't be undone

┐(ﾟ ヮﾟ)┌

### DIFF
--- a/app/Libraries/OsuAuthorize.php
+++ b/app/Libraries/OsuAuthorize.php
@@ -59,6 +59,24 @@ class OsuAuthorize
         return $this->cache[$cacheKey];
     }
 
+    public function checkBeatmapDiscussionDestroy($user, $discussion)
+    {
+        $prefix = 'beatmap_discussion.destroy.';
+
+        $this->ensureLoggedIn($user);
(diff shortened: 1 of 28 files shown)


app/Libraries/OsuAuthorize.php
```diff
@@ -59,6 +59,24 @@ public function doCheckUser($user, $ability, $object)
...
+        $prefix = 'beatmap_discussion.destroy.';
+
+        $this->ensureLoggedIn($user);
+        $this->ensureCleanRecord($user);
+
+        if ($user->user_id !== $discussion->user_id) {
+            return;
+        }
+
+        if ($discussion->beatmapDiscussionPosts()->withoutDeleted()->count() > 1) {
```
nekodex: exists instead of count > 1?

nanaya: different thing, sir

PR #873, opened 2016-12-03
Make french locale available and configured

leonekmi:
I just forget to make language avivable in pull #753 , my bad.

## _Changes :_
- config/app.php (Add fr to config for Laravel Translation service)
- app/Libraries/LocaleMeta.php (reordering and added French(Français) to the file)


nanaya:
- wrong target branch; should be `master` instead of `release`.
- need to add language information to `app/Libraries/LocaleMeta.php`.

leonekmi:
Oh, thanks for the precision, i don't know so much Laravel Framework,

and yes i have changed the target branch

nanaya:
Language name should be in its native language (for example 日本語 for Japanese). I don't know about French but pretty sure it's not called "French" in French.

Also would be nice if it's located in the same order as in `app.php`.

nanaya:
Also, what's "avivable"? It doesn't seem to be an English word. Looking for it in French doesn't seem to have relevant meaning either.

leonekmi:
Oh, i want say **available**, sorry and yes i pushed the fixes

PR #878, opened 2016-12-07
Small beatmapset fixes

### LINKED ISSUE #870: Star difficulty metre is not capped on beatmap listing pages
It should be, otherwise [this happens](http://puu.sh/sBBUJ/d987d7a390.png).

[Example page](https://new.ppy.sh/beatmapsets/192827#osu/839855)

cl8n:
- Prevent stat bars from going over values of 10 (fixes #870)
- Specify approval type (instead of "ranked on ...", show "qualified on ...", etc.)
- Show when a map was last updated, if it isn't approved


resources/assets/coffee/react/_components/beatmapset-mapping.coffee
```diff
@@ -20,7 +20,7 @@ el = React.createElement
 
 bn = 'beatmapset-mapping'
 
-@BeatmapsetMapping = ({user, beatmapset}) ->
+@BeatmapsetMapping = ({user, beatmapset, beatmap}) ->
```
nanaya: any reason to not use data from `beatmapset` itself?

cl8n: some beatmaps have different statuses than others within the set (e.g. sets with both Loved and Graveyarded maps)

nanaya: apparently this should show the beatmapset state anyway (said the guy responsible for this).

nanaya: what I'm saying is @peppy said this particular react component should show the `beatmapset` approved status, not `beatmap`.

cl8n: oh, sorry I misunderstood what you said. that seems a bit strange to use the set's approved status (since it might differ from the actual map), but I can change it back if it's intentional

cl8n:
This should be ready to go now. I left some things in the Model and Transformers that I didn't end up using, but since they only complete them more, I figure they wouldn't be harmful additions.

nanaya:
Thanks!

PR #887, opened 2016-12-13
IE fixes

nanaya:
Slightly less broken now.


nanaya:
if you're using vs code, it's the `files.eol` option (unix is `\n`).  Other editors should have similar setting as well.

PR #891, opened 2016-12-15
Add beatmapset availability info

### LINKED ISSUE #881: Beatmap page is missing copyright takedown notices
https://gist.github.com/peppy/b077bcef445c2312b7973c52f67b0ed0

The new beatmap pages for the songs in that notice don't have the copyright notice (at least not that I can see). I think it would be confusing for a user to download a song from the new beatmap page, then try to play it and get no sound in the client.

nekodex:
Also hides download buttons when a beatmapset is no longer available (fixes #881)

### DIFF
--- a/resources/assets/less/bem/beatmapset-header.less
+++ b/resources/assets/less/bem/beatmapset-header.less
@@ -70,7 +70,7 @@
   }
 
   &__availability-info {
-    font-size: smaller;
+    font-size: @font-size--title-small;
     padding: @spacing;
     margin-top: @spacing;
     background: rgba(0, 0, 0, 0.5);
@@ -79,7 +79,7 @@
 
   &__availability-link {
     padding-top: @spacing;
-    font-size: x-small;
+    font-size: @font-size--small;
   }
 
   &__details-text {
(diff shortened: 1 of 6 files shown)


resources/assets/less/bem/beatmapset-header.less
```diff
@@ -69,6 +69,19 @@
...
+    font-size: smaller;
+    padding: @spacing;
+    margin-top: @spacing;
+    background: rgba(0, 0, 0, 0.5);
+    color: @yellow-dark;
+  }
+
+  &__availability-link {
+    padding-top: @spacing;
+    font-size: x-small;
```
nanaya: use pixel instead of user-defined default?

resources/assets/less/bem/beatmapset-header.less
```diff
@@ -69,6 +69,19 @@
     min-height: @header-height;
   }
 
+  &__availability-info {
+    font-size: smaller;
```
nanaya: this one as well.... or not. Not even mdn can tell how much smaller is `smaller` though.

resources/assets/less/bem/beatmapset-header.less
nanaya: Or one of those `@font-size--<arbitrary-size>`

PR #893, opened 2016-12-17
Add strings and more localisation

leonekmi:
**/home/download and /forum is now better located**
adding more translations
fixes
_Sorry for mass editing, my new translation manager doesn't love empty lines, etc..._


leonekmi:
Conflicts will be resolved into this week-end

nanaya:
This changes too many unwanted things as well (removing comments, hardcoding configs, usage of single quotes in string containing single quotes, etc).

Also if you're suggesting usage of some kind of tool better suggest the tool itself instead of making pull request of its output.

PR #898, opened 2016-12-22
added accounts to pt-br

HWSKY:
This file was originally translated according to lang/en/accounts.php root file.
All translations and dependencies are fully included in this file since the last change, commit: a4ca4a8


resources/lang/pt-br/accounts.php
```diff
@@ -0,0 +1,43 @@
...
+ *    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *    See the GNU Affero General Public License for more details.
+ *
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+return [
+    'edit' => [
+        'title' => 'Configurações de Conta',
```
Tarkoviski: Configurações d**a** conta

resources/lang/pt-br/accounts.php
```diff
@@ -0,0 +1,43 @@
...
+        'title_compact' => 'configurações',
+
+        'avatar' => [
+            'title' => 'Editar Avatar',
```
Tarkoviski: We agreed to use capitalized letters only at the start of new sentences. I think it would be better to use "Editar avatar" instead of "Editar Avatar".

resources/lang/pt-br/accounts.php
```diff
@@ -0,0 +1,43 @@
...
+        ],
+
+        'profile' => [
+            'title' => 'Editar Perfil',
```
Tarkoviski: Same here.

resources/lang/pt-br/accounts.php
HWSKY: i've been translating as said the push title

PR #899, opened 2016-12-27
Wiki?

nanaya:
☜(ﾟヮﾟ☜)

- no table of contents
- no special markdown parsing
- ~~no persistent language navigation (coming soon™?)~~ kind of works
- ~~no cache busting (currently expiration set to 1 hour)~~ cached with shiny refresh button
- images are proxied as well (except none of them are linked properly from what I can tell)
- ~~also, `jp` isn't language code for Japanese. `ja` is.~~ not my problem 💃 
- ~~there's no proper error page as well (coming soon™)~~
- ~~and no github link~~


Repflez:
>also, jp isn't language code for Japanese. ja is

Already fixed in ppy/osu-wiki#76.

peppy:
looks good so far 👍 . are you okay with this being deployed in its current state, or shall i wait for the last few things in your list to be finished off?

nanaya:
I think it's fine as it currently is

PR #915, opened 2017-01-17
Contest fixes and stuff

nekodex:
Reworked contest pages to fix some layout issues. Also adds some (hacky) preliminary support for features needed for 'best of' voting (unmasked entry names, filtering by played beatmaps, etc)

PR #916, opened 2017-01-18
More contest fixes and stuff

nekodex:
General cleanup and fixes (including the remainder from #915). Also optimises the number of db queries used.


app/Models/Contest.php
```diff
@@ -54,6 +55,11 @@ public function cachedVoteAggregates()
         });
     }
 
+    public function isBestOf()
+    {
+        return isset($this->extra_options['best_of']);
```
nanaya: shouldn't it be like actual content check or something instead of just isset

app/Models/Contest.php
```diff
@@ -181,19 +187,54 @@ public function vote(User $user, ContestEntry $entry)
...
+                        ->from('osu_beatmaps')
+                        ->where('osu_beatmaps.playmode', '=', $playmode)
+                        ->whereIn('beatmap_id', function ($query) use ($user) {
+                            $query->select('beatmap_id')
+                                ->from('osu_user_beatmap_playcount')
+                                ->where('user_id', '=', $user->user_id);
+                        });
+                });
+        } else {
+            $entries = $this->entries()->with('contest');
```
nanaya: the `with('contest')` can be moved to the return part?

app/Models/Contest.php
```diff
@@ -221,7 +262,7 @@ public function defaultJson($user = null)
 
         return json_encode([
```
nanaya: on unrelated note, apparently none of other `defaultJson` actually return json. Also none of other json_encode explicitly set the options.

resources/assets/less/bem/contest-voting-list.less
```diff
@@ -93,16 +93,18 @@
...
-      flex-direction: column;
-      justify-content: space-around;
     }
   }
 
+  &__votes-bar {
+    position: absolute;
+    height: 100%;
+    background-color: #3F4E51;
+    margin: -2px -@spacing;
```
nanaya: `top: 0; left: 0` also works (and less math)

resources/assets/coffee/react/contest/voting/entry.coffee
```diff
@@ -33,10 +33,11 @@ class Contest.Voting.Entry extends React.Component
           el TrackPreview, track: @props.entry
       if @props.options.showLink && @props.entry.preview
         div className: 'contest-voting-list__icon contest-voting-list__icon--bg',
-          a className: 'tracklist__link', href: (if @props.contest.best_of then laroute.route('beatmapsets.show', beatmapset: @props.entry.preview) else @props.entry.preview),
-            i className: "fa fa-fw fa-lg #{@props.contest.link_icon ? 'fa-cloud-download'}"
+          a className: 'tracklist__link', href: @props.entry.preview,
+            i className: "fa fa-fw fa-lg #{@props.contest.link_icon}"
```
nanaya: just noticed, if `el Icon` helper is used, the `fa-` prefix in icon name isn't needed.

PR #920, opened 2017-01-20
Ensure absolute url for images in wiki

nanaya:
Fix broken image on certain cases.


peppy:
got some examples of where this is being used? not sure we want to allow external links in the first place..

nanaya:
The problem is on turbolinks navigation for local images.

peppy:
oooh i see.

PR #921, opened 2017-01-20
Allow .osz files (up to 20MB) for beatmap contests (for aspire)

nekodex:
(also includes minor style/layout fixes)


resources/assets/less/bem/contest-userentry-list.less
```diff
@@ -21,7 +21,6 @@
   flex-wrap: wrap;
   background-color: #1A2223;
   color: white;
-  padding: 20px;
-  padding-right: 0;
+  padding: 10px 10px 0px 10px;
```
nanaya: 0px should be just 0 =D

Issue #942, opened 2017-01-31
Language choose


kj415j45:
During I translate osu-web,I found a BUG(maybe it's feature XD)

If I haven't login,I can't change language :(
But if I login,I can change language :)

Maybe it'll be difficult for someone who first enter osu-web and doesn't understand english because he haven't have an account and login to change language(maybe)

kj415j45:
And I think we should remove italic for chinese,because:
![image](https://cloud.githubusercontent.com/assets/18349191/22477778/6080d952-e822-11e6-91af-4d98200f05da.png)
it's so stupid XD

benpigchu:
【the Chinese font should be a common sans-serif font rather than “宋体”】

nanaya:
seems fixed

PR #944, opened 2017-02-01
Lowercase email for lookup

nanaya:
There are currently 10-ish people who have their email containing uppercase letter in the database.


nanaya:
Apparently the username part is actually case sensitive in the spec so this is probably a bit wrong ┐(ツ)┌

kj415j45:
I think alter the database is better.Though it won't take too much time to strtolower()


By the way, please hava a look: issue #942 :)

PR #952, opened 2017-02-02
:amount already includes the unit

nanaya:
As reported by @kj415j45.


kj415j45:
I think it's better to change  _:amount_ because "amount" seem to be a number

PR #960, opened 2017-02-08
Add a better font

kj415j45:
黑体 is better than Arial Unicode MS
Arial Unicode MS:
image
黑体:
image


kj415j45:
More screenshot of 黑体:
image
image

PR #965, opened 2017-02-10
Support mobile meta tags for toolbar color and fullscreen

MuhunKim:
I've Added more meta tags for style of mobile platform 
### toolbar color meta tags
  * [IOS docs](https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/MetaTags.html)
  * [Windows docs](https://msdn.microsoft.com/en-us/library/gg491732(v=vs.85).aspx#msapplication-navbutton-color) : Windows phones also apply.
### home screen mode
 * [Android Chrome docs](https://developer.chrome.com/multidevice/android/installtohomescreen)
### Splash Screen
* [Chrome Web docs](https://developers.google.com/web/updates/2015/10/splashscreen)


peppy:
file has wrong filename
is in wrong location
etc.

nanaya:
Doesn't seem useful. IE toolbar thing requires it to be pinned first (and weird) and so is iOS one (doesn't seem useful, can't even copy current url when in fullscreen mode). Also, wrong as pointed above.

PR #970, opened 2017-02-10
Also escape in forum view

### LINKED ISSUE #969: Joined <strong>January 2014</strong>
image

The &lt;strong&gt; tag is escaped.

nanaya:
Will do a better fix later

Fixes #969.


peppy:
🐟

PR #975, opened 2017-02-13
Add vote weighting support for the 'best of...' contests

nekodex:
Adds weighted vote support for best_of contests, also includes some bonus (display) tweaks and optimisations for the contest result pages too.


app/Models/Contest.php
```diff
@@ -189,33 +182,45 @@ public function vote(User $user, ContestEntry $entry)
...
+                        ->selectRaw('*')
+                        ->selectRaw('(SELECT FLOOR(SUM(`weight`)) FROM `contest_votes` WHERE `contest_entries`.`id` = `contest_votes`.`contest_entry_id`) AS votes_count')
+                        ->limit(50); // best of contests tend to have a _lot_ of entries...
+                } else {
+                    $entries = $entries->withCount('votes');
+                }
+
+                return $entries->orderBy('votes_count', 'desc')->get();
+            });
+        } else {
```
nanaya: not needed cuz the block returns the thing.

app/Transformers/ContestEntryTransformer.php
```diff
@@ -42,13 +42,11 @@ public function transform(ContestEntry $entry)
...
     {
         return $this->item($entry, function ($entry) {
-            $voteCounts = $entry->contest->cachedVoteAggregates()->where('contest_entry_id', $entry->id)->first();
-
             return [
                 'actual_name' => $entry->name,
                 'user_id' => $entry->user_id,
                 'username' => ($entry->user ?? (new \App\Models\DeletedUser))->username,
-                'votes' => $voteCounts ? $voteCounts->votes : 0,
+                'votes' => $entry->votes_count ? (int) $entry->votes_count : 0,
```
nanaya: p.sure falsy value already casts into 0.

app/Models/Contest.php
```diff
@@ -48,13 +48,6 @@ public function voteAggregates()
         return $this->hasMany(ContestVoteAggregate::class);
     }
 
-    public function cachedVoteAggregates()
```
nanaya: remove the `voteAggregates` method and the model as well?

resources/assets/coffee/react/contest/voting/entry.coffee
```diff
@@ -24,12 +24,20 @@ class Contest.Voting.Entry extends React.Component
     if @props.contest.show_votes
       votePercentage = _.round((@props.entry.results.votes / @props.totalVotes)*100, 2)
       relativeVotePercentage = _.round((@props.entry.results.votes / @props.winnerVotes)*100, 2)
+      rank = @props.rank + 1
```
nanaya: `+1` when passing the props instead?

resources/assets/coffee/react/contest/voting/entry.coffee
```diff
@@ -24,12 +24,20 @@ class Contest.Voting.Entry extends React.Component
...
 
     selected = _.includes @props.selected, @props.entry.id
 
     div className: "contest-voting-list__row#{if selected && !@props.contest.show_votes then ' contest-voting-list__row--selected' else ''}",
+      if @props.contest.show_votes
+        div className: 'contest-voting-list__rank',
+          if rank < 4
+            div className: "contest-voting-list__trophy contest-voting-list__trophy--#{rank}",
```
nanaya: There's new `parentClass` props for `Icon` component which kind of simplifies this thing.

resources/assets/coffee/react/contest/voting/entry.coffee
```diff
@@ -39,15 +47,19 @@ class Contest.Voting.Entry extends React.Component
...
-          "#{@props.entry.results.votes} votes"
-          if not isNaN(votePercentage)
-            " (#{votePercentage}%)"
+        if @props.contest.best_of
+          div className:'contest__vote-count contest__vote-count--no-percentages',
+            "#{@props.entry.results.votes} points"
+        else
+          div className:'contest__vote-count',
+            "#{@props.entry.results.votes} votes"
+            if not isNaN(votePercentage)
```
nanaya: looks like `isFinite` a better test.

app/Models/Contest.php
```diff
@@ -189,33 +182,45 @@ public function vote(User $user, ContestEntry $entry)
...
-            if ($this->show_votes) {
+        if ($this->show_votes) {
+            return Cache::remember("contest_entries_with_votes_{$this->id}", 5, function () use ($entries) {
                 $entries = $entries->with('user');
+
+                if ($this->isBestOf()) {
+                    $entries = $entries
+                        ->selectRaw('*')
+                        ->selectRaw('(SELECT FLOOR(SUM(`weight`)) FROM `contest_votes` WHERE `contest_entries`.`id` = `contest_votes`.`contest_entry_id`) AS votes_count')
+                        ->limit(50); // best of contests tend to have a _lot_ of entries...
```
nanaya: how about applying this to non bestof as well? and perhaps also the sum weight?

nanaya: as in skip the conditional check and just use this for both cases.

nekodex: 'cuz one of the queries is like 4x slower? ┐(´д｀)┌

nekodex: also re: limit; the best_of contests will have the full results released separately, whereas not so much for the other contests

app/Models/Contest.php
nekodex: Hrm looking at this again, I think leaving it in helps the readability here?

PR #929, opened 2017-01-26
Topic reply live preview

LiquidPL:
Resolves #30.

PR #1044, opened 2017-03-15
Add translate section 

kj415j45:
#929 add a new section but only to **en**, so i add them to all
and translate **zh**


kj415j45:
will add #1054 in one or two day(if this PR is not necessary, please close it)

PR #1055, opened 2017-03-29
'Fix' timezone for Beatmapset datetime fields

nekodex:
Currently they're stored as GMT+8, thanks to legacy stuff


nanaya:
WHERE IS THE METAPROGRAMMERING

nekodex:
SORRY NANAYA, METAPROGRAMMERING IS IN ANOTHER CASTLE

PR #1063, opened 2017-04-03
Break long words/links in wiki pages

### LINKED ISSUE #1060: wiki page
image
image
we can see those words and images are out of the container

benpigchu:
This fixes #1060


nanaya:
- should be in `bem/wiki-page.less`
- use `min-width: 0` instead of `max-width: 100%` because in desktop mode, the maximum width isn't 100% (it's `100% - <sidebar-width>`).

PR #1069, opened 2017-04-05
Add endpoint for legacy systems to trigger beatmapset cover regeneration

nekodex:
'cuz oauth is hard and effort for systems that will be replaced or something

NOTE: update `.env` and change `SHARED_COOKIE_SECRET` => `SHARED_INTEROP_SECRET` before deploying


app/Http/Middleware/LegacyInterOpAuth.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        if (!$request->hasHeader('X-LIO-Signature') || !presence($request->query('timestamp'))) {
```
nanaya: you mean `!present`

app/Http/Middleware/LegacyInterOpAuth.php
```diff
@@ -0,0 +1,35 @@
...
+            abort(403);
+        }
+
+        $timestamp = $request->query('timestamp');
```
nanaya: might as well assign this first before checking for presence.

app/Http/Middleware/LegacyInterOpAuth.php
```diff
@@ -0,0 +1,35 @@
...
+        $signature = $request->header('X-LIO-Signature');
+        $expected = hash_hmac('sha1', $request->fullUrl(), config('osu.legacy.shared_interop_secret'));
+
+        $diff = Carbon::createFromTimestamp($timestamp)->diffInSeconds();
```
nanaya: slightly simpler(?) with `->addMinutes(5)->isPast()`?

app/Http/Middleware/LegacyInterOpAuth.php
```diff
@@ -0,0 +1,35 @@
...
+
+        if ($diff > 300 || $signature !== $expected) {
```
nanaya: something timing attack something `hash_equals`. (also in various other places)

app/Http/Controllers/LegacyInterOpController.php
```diff
@@ -0,0 +1,41 @@
...
+namespace App\Http\Controllers;
+
+use App\Jobs\RegenerateBeatmapsetCover;
+use App\Models\Beatmapset;
+use Illuminate\Foundation\Bus\DispatchesJobs;
+use Illuminate\Routing\Controller as BaseController;
+
+class LegacyInterOpController extends BaseController
+{
+    use DispatchesJobs;
```
nanaya: update app base controller to not explode on missing `$session` and use that instead.

nanaya: I can't type.

nanaya: wait what why is it ended up as another review what why

nanaya: I can't type.

nanaya:
also styleci exploded again?

app/Http/Middleware/LegacyInterOpAuth.php
nekodex: This is deliberately allowing for 5mins on either side due to system clock drift still being a thing (apparently :v)

nanaya: who had the idea to default absolute value it (ノ･_･)ノ︵ ┻━┻

nekodex: Cuz difference is absolute? >.> (e.g a 5 degree difference can mean either +5 or -5)

app/Http/Controllers/LegacyInterOpController.php
nekodex: Won't that add csrf checks and stuff?

app/Http/Middleware/LegacyInterOpAuth.php
nanaya: that sounds more like tolerance or margin of error =|

app/Http/Controllers/LegacyInterOpController.php
nanaya: but that's not where it's defined?

nanaya: I think it's always active unless you put it in the exception or update the middleware definitions.

nanaya: Also the impact of failing verification is just it resets the session.

app/Http/Controllers/LegacyInterOpController.php
```diff
@@ -0,0 +1,40 @@
...
+
+    public function regenerateBeatmapsetCovers($id)
+    {
+        $beatmapset = Beatmapset::findOrFail($id);
+
+        $job = (new RegenerateBeatmapsetCover($beatmapset))->onQueue('beatmap_processor');
+        $this->dispatch($job);
+
+        return response(['success' => true]);
```
nanaya: just return array

app/Http/Middleware/LegacyInterOpAuth.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
```
nanaya: missing license header.

app/Http/Middleware/LegacyInterOpAuth.php
```diff
@@ -0,0 +1,30 @@
...
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        $timestamp = $request->query('timestamp');
+        $diff = Carbon::createFromTimestamp($timestamp)->diffInSeconds();
+        $signature = $request->header('X-LIO-Signature');
+        $expected = hash_hmac('sha1', $request->fullUrl(), config('osu.legacy.shared_interop_secret'));
+
+        if (!present($signature) || !present($timestamp) || $diff > 300 || !hash_equals($expected, $signature)) {
```
nanaya: inverse the check to save some characters =D (`if(present($timestamp) && $diff < 300 && ...) { return $next...; } abort...;

peppy:
💯

PR #1070, opened 2017-04-05
Send message to both addresses

nanaya:
Not sure if there's better way short of using email confirmation link thingy. Also can use a better wording.

### DIFF
--- a/app/Http/Controllers/AccountController.php
+++ b/app/Http/Controllers/AccountController.php
@@ -138,9 +138,11 @@ class AccountController extends Controller
             ->fill(Request::input('user_email'));
 
         if ($userEmail->save() === true) {
+            $mailTo = [$user->user_email];
             if (present($previousEmail)) {
-                Mail::to($previousEmail)->send(new UserEmailUpdated($user));
+                $mailTo[] = $previousEmail;
             }
+            Mail::to($mailTo)->send(new UserEmailUpdated($user));
 
             return ['message' => trans('accounts.update_email.updated')];
         } else {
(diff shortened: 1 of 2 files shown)


resources/views/emails/user_email_updated-en.blade.php
```diff
@@ -18,6 +18,8 @@
 Hi {{ $user->username }},
 
 This is just a confirmation that your osu! email address has been changed to "{{ $user->user_email }}".
+You should receive this message to both of your new and old email addresses.
+Please make sure that you received this email at your new address otherwise you may get locked out of your account.
```
nekodex:
```
This is a confirmation email to inform you that your osu! email address has been changed to: "{{ $user->user_email }}"
For security reasons, this email has been sent both to your new and old email address.
Please ensure that you received this email at your new address to prevent losing access your osu! account in the future.
```

app/Http/Controllers/AccountController.php
```diff
@@ -138,8 +138,12 @@ public function updateEmail()
             ->fill(Request::input('user_email'));
 
         if ($userEmail->save() === true) {
+            $addresses = [$user->user_email];
             if (present($previousEmail)) {
-                Mail::to($previousEmail)->send(new UserEmailUpdated($user));
+                $addresses[] = $previousEmail;
+            }
+            foreach ($addresses as $address) {
+                Mail::to($address)->send(new UserEmailUpdated($user));
```
nekodex: Apparently you can send to multiple addresses by passing an object/collection to `Mail::to` (see https://laravel.com/docs/5.4/mail#sending-mail)

nanaya: this is for consistency of not showing previous email address

PR #1072, opened 2017-04-07
Fix user ranking graphs with missing data

cl8n:
Tiny fix for ranking graphs that don't have all data points set

See https://osu.ppy.sh/u/3666350 vs https://new.ppy.sh/u/3666350; the latter graph shows incorrectly shifted dates. This change fills the gap with a straight line


nanaya:
Thanks!

PR #1073, opened 2017-04-07
Also set approval for beatmaps in the set

### LINKED ISSUE #1066: Qualification via moddingv2 is not setting approved status properly - beatmap discussions
https://new.ppy.sh/beatmapsets/527051/discussion

Currently, viewing this map from beatmap discussions suggests it is qualified.

Viewing it from a database perspective, the **beatmapset** is set to ```approved = 3```, whilst all the difficulties (**beatmaps**) contained within are set to ```approved = -2```.

This causes the game to display "No records set!" in-game. 

Attempting to set scores results in "the map's ranking state has changed or it is no longer available" error, which is probably expected behavior given that the difficulties are all ```approved = -2```.

nanaya:
Should fix #1066. I'm not sure if it's the correct behavior ┐(ﾟ ヮﾟ)┌

### DIFF
--- a/app/Models/Beatmap.php
+++ b/app/Models/Beatmap.php
@@ -26,6 +26,7 @@ class Beatmap extends Model
 {
     protected $table = 'osu_beatmaps';
     protected $primaryKey = 'beatmap_id';
+    protected $guarded = [];
 
     protected $casts = [
         'orphaned' => 'boolean',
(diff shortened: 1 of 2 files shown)


cl8n:
approved_date should be reset on disqualification

cl8n:
also, approvedby_id should be set to the final nominator's user_id (and reset on disqualification)

PR #1076, opened 2017-04-10
Sort by plays

### LINKED ISSUE #1075: Popular Beatmap's Order
https://my.mixtape.moe/cnyzzg.png
The popular beatmaps is not sorted by amount of plays

nanaya:
It wasn't sorted before. Fixes #1075.

Also cleaned up some css and sql.

### DIFF
--- a/resources/assets/less/bem/user-home-beatmapset.less
+++ b/resources/assets/less/bem/user-home-beatmapset.less
@@ -16,26 +16,32 @@
  *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-.user-home-beatmap-list {
-  @top: user-home-beatmap-list;
+.user-home-beatmapset {
+  .link-plain();
+  .own-layer();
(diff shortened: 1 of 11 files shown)


resources/assets/less/bem/user-home-beatmapset.less
```diff
@@ -16,26 +16,32 @@
...
+  .own-layer();
+  display: flex;
+  margin-bottom: 10px;
+  width: 100%;
 
-  &__beatmap {
-    display: flex;
-    margin-bottom: 10px;
-    width: 100%;
+  border-radius: (@border-radius-base + 1px);
```
nekodex: this feels a bit... silly? just bump it to `@border-radius-large` imo

nanaya: isn't it just to avoid bleeding(?) from cover's rounded corners? Hence `+1px`.

PR #1089, opened 2017-04-19
Misc API fixes

### LINKED ISSUE #1086: API v2 issues
- [ ] `User` from `/users/<id>` lookup returns a `Country` object, while the `User` in scores lookup return s a string for the same variable name. We need to avoid such clashes (either return same thing in both or rename the string one to `CountryCode` or something).
- [ ] `combo` should probably be `maxCombo`.
- [ ] move all `count` values into a `statistics` array.

nekodex:
Changes:
- Return standard (compact)user objects in messages
- Convert API responses to camel-case for messages and channels (will do beatmaps later)
- Add missing 'perfect' attribute to score response
- Move/group hit stats into a subarray on scores

(fixes #1086)


app/Transformers/API/Chat/ChannelTransformer.php
```diff
@@ -28,10 +28,10 @@ class ChannelTransformer extends Fractal\TransformerAbstract
     public function transform(Channel $channel)
     {
         return [
-            'channel_id' => $channel->channel_id,
+            'channelId' => $channel->channel_id,
```
nanaya: I meant snake case whoops 🐍

nekodex: (ノ・_・)ﾉ︵ ┻━┻

PR #1095, opened 2017-04-24
CSS Fixes

nanaya:
Previous change of adding margin-bottom broke quite a lot of stuff so it's reverted. And some more other minor cleanups.


resources/assets/less/bem/forum-pippi.less
```diff
@@ -0,0 +1,39 @@
...
+ *    See the GNU Affero General Public License for more details.
+ *
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+.forum-pippi {
+  @extra-margin: (-@grid-gutter-width + 5px);
+
+  .at2x-simple('/images/layout/forum-top-pippy.png');
```
nekodex: pippy? pippi.

PR #1101, opened 2017-04-25
Now with local video

nanaya:
crf30 preset:veryslow tune:animation profile:high level:4.2 fps:source cfr "web optimized"


nekodex:
Probably want the video file to be on assets instead of the web server 😉

PR #1105, opened 2017-04-28
osu!wiki - add TOC scrollbars when overflowing vertically

MegaApplePi:
Hmm, the only problem with this is that scrollbars aren't really styleable.

fix for #1111


nekodex:
After discussion with @arflyte, we've decided to fix this issue in a different way - see #1120

PR #1106, opened 2017-04-28
fix locale_name

kj415j45:
add missing locale_name()                     

before:


nanaya:
The design from flyte uses short names. Also you missed one place.

kj415j45:
short names?~~(pt-br may not)~~

where? I only find this...

PR #1108, opened 2017-04-29
Fix mobile view for contest entry names and headers

### LINKED ISSUE #797: [Contest Index Page] Subtitle get cropped on mobile

MegaApplePi:
Styling changes to fix #797.
- applies when `@mobile`
  - ~~reset text-overflow, overflow, and white-space rules for contest entry name~~
  - increase entry item height to allow more text
  - (increase image height to match above)
  - decrease legend font-size
  - decrease contest list padding to give more space for text, image, border, etc.
  - reset heading top margin to 0
  - use new defined font-size for mobile header v2
  - issue: 3 lines max for both contest entry name and headers (or they will overflow when `@mobile`)
  - `justify-content: center;` for heading v2

_Left: this branch; Right: current_
image
image

### DIFF
--- a/resources/assets/less/bem/contest-list-item.less
+++ b/resources/assets/less/bem/contest-list-item.less
@@ -30,10 +30,10 @@
 
   border-right: 5px solid #323E41;
 
-  @media @mobile{
+  @media @mobile {
     height: 90px;
   }
- 
+
   &:active, &:focus, &:hover {
     color: white;
     text-decoration: none;
@@ -92,17 +92,14 @@
   }
 
   &__name {
-    font-size: 17px;
-    text-overflow: ellipsis;
-    overflow: hidden;
-    white-space: nowrap;
-    @media @mobile {
-      font-size: 14px;
-      text-overflow: initial;
-      overflow: initial;
-      white-space: initial;
-    }
+    font-size: 14px;
     padding-bottom: 5px;
+    @media @desktop {
+      font-size: 17px;
+      text-overflow: ellipsis;
+      overflow: hidden;
+      white-space: nowrap;
+    }
   }
 
   &__date {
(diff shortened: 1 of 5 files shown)


resources/assets/less/bem/contest-list-item.less
```diff
@@ -30,6 +30,10 @@
 
   border-right: 5px solid #323E41;
 
+  @media @mobile{
+    height: 90px;
+  }
+
```
nekodex: trailing whitespace

resources/assets/less/bem/contest-list-item.less
```diff
@@ -94,6 +98,9 @@
     white-space: nowrap;
     @media @mobile {
       font-size: 14px;
+      text-overflow: initial;
+      overflow: initial;
+      white-space: initial;
```
nekodex: probably cleaner to move the overflow stuff into a `@media @desktop {` block instead of setting them on the base and then resetting them here

resources/assets/less/bem/osu-page-header-v2.less
```diff
@@ -67,6 +67,10 @@
     font-size: @font-size--new-header-subtitle;
     margin-bottom: 10px;
     margin-top: 5px;
+
```
nekodex: trailing whitespace

resources/assets/less/bem/osu-page-header-v2.less
```diff
@@ -24,7 +24,7 @@
   padding: 20px;
   display: flex;
   flex-direction: column;
-  justify-content: flex-start;
+  justify-content: center;
```
nekodex: this pushes the header down further than it should be on desktop, I'd put it in a `@media @mobile` block instead

PR #1109, opened 2017-04-29
osu!wiki - add text-indent

MegaApplePi:
Indent the text by 2em, requested by @kj415j45 for Chinese

- add text-indent
- fix spacings for headings, lists, and images

PR #1116, opened 2017-05-02
osu!wiki - increase margin-bottom for paragraph

MegaApplePi:
_(Continued from #1109.)_ Increase `margin-bottom` for `<p/>` tags inside `.wiki-content` to `1em` (was `10px`).


resources/assets/less/bem/wiki-content.less
```diff
@@ -65,6 +65,10 @@
     }
   }
 
+  & p {
```
nanaya: the `&` is pointless, and move this to the top before `&__header`.

PR #1122, opened 2017-05-04
Make table of contents scroll with the page when content is too long in the wiki pages

### LINKED ISSUE #1120: Wiki sidebar (table of contents) should scroll with the page when content is too long
The sidebar on long pages ([for example](https://new.ppy.sh/wiki/osu!wiki_contribution_guide/)) should scroll with the rest of the page instead of being pinned to the top of the window.

For pages where the sidebar content is shorter than the window height, it should still behave as it currently does and be pinned.

This issue supersedes #1111.

benpigchu:
Fix #1120


nanaya:
The floating feels a bit weird. I've made a different one with no floating at all for long toc.

nanaya:
Done in #1124.

PR #1123, opened 2017-05-08
Password reset

nanaya:
Seems working.


peppy:
looks good from a pass on the code 🥇 

will deploy later today when you're around, just in case

PR #1124, opened 2017-05-09
Don't float wiki page table of contents if too long

### LINKED ISSUE #1120: Wiki sidebar (table of contents) should scroll with the page when content is too long

nanaya:
And one other fix.

Fixes #1120.


kj415j45:
#1122？

PR #1125, opened 2017-05-09
osu!wiki - add image captions

MegaApplePi:
- ~~add `figure` and `figcaption` styles~~
- ~~add function `addImageCaption`~~
- add `&__figure` and `&__figure-caption` styles
- attach a document ready (jQuery) event to make `<figure/>`

Notes:
- The caption is ~~stolen~~ taken from the title attribute.
- Only 1 element (the image element) must be present for this to trigger
  - _If other elements are present, assume the image is inline and stop_
- The title attribute must be present for this to trigger
- The parent element has to be the `<p/>` tag (`&__paragraph`)
- The pop-up hover text will appear and behave as normal.


MegaApplePi:
@RaikireHiuduo, a design check is needed, but the pop-up hover text will appear and behave as normal.

MegaApplePi:
(Weird hard reset thingy, but updated to resolve issues with base branch.)

nanaya:
This should extend WikiProcessor instead of doing it in javascript.

nanaya:
Already done by extending markdown parser.

PR #1129, opened 2017-05-12
Store build propagation history in database

LiquidPL:
Adds an env variable `TRACKED_STREAMS` which determines the builds which history is being updated.


app/Models/Build.php
```diff
@@ -51,6 +51,12 @@ public function scopeLatestByStream($query, $streamIds)
         $query->whereIn('build_id', $latestBuildIds);
     }
 
+    public function scopePropagationHistory($query)
+    {
+        return $query->where('allow_bancho', '>', 0)
```
nanaya: pretty sure it's a boolean

app/Models/Build.php
```diff
@@ -51,6 +51,12 @@ public function scopeLatestByStream($query, $streamIds)
...
+            ->where('test_build', 0);
```
nanaya: this as well

app/Models/BuildPropagationHistory.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Model;
+
+class BuildPropagationHistory extends Model
+{
+    protected $table = 'osu_build_propagation_history';
```
nanaya: if only the table is called `build_propagation_histories`...

app/Models/BuildPropagationHistory.php
```diff
@@ -0,0 +1,13 @@
...
+    const UPDATED_AT = null;
```
nanaya: not sure if it's a supported construct. Perhaps better just set `$timestamps` to `false`.

database/migrations/2017_05_12_111922_add_build_propagation_history_table.php
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('osu_build_propagation_history', function ($table) {
+            $table->increments('id');
+            $table->integer('stream_id')->unsigned();
```
nanaya: record build version as well?

LiquidPL: this stores data from all builds being online at any given moment, so I don't think it would make sense to associate a build version to it.

app/Models/BuildPropagationHistory.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Model;
+
+class BuildPropagationHistory extends Model
+{
+    protected $table = 'osu_build_propagation_histories';
```
nanaya: if only the table is called `build_propagation_histories`, this line isn't even needed.

database/migrations/2017_05_12_111922_add_build_propagation_history_table.php
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('osu_build_propagation_histories', function ($table) {
+            $table->increments('id');
+            $table->integer('stream_id')->unsigned();
```
nanaya: Just checked and indeed it should record per-build, not per-stream.

nanaya: and per-stream record isn't exactly exciting.

LiquidPL: so it should be only build version, instead of stream id?

nanaya: yes

database/migrations/2017_05_12_111922_add_build_propagation_history_table.php
```diff
@@ -0,0 +1,32 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\Schema;
+
+class AddBuildPropagationHistoryTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void p
```
nanaya: return what

app/Models/BuildPropagationHistory.php
```diff
@@ -0,0 +1,15 @@
+<?php
```
nanaya: license

app/Console/Commands/BuildsUpdatePropagationHistory.php
```diff
@@ -0,0 +1,53 @@
+<?php
```
nanaya: license

database/migrations/2017_05_12_111922_add_build_propagation_history_table.php
LiquidPL: *that never happened*

nanaya:
can use style fix.

PR #1131, opened 2017-05-14
Update home.php [PT-BR]

Kuron-kun:
"simulador de ritmo gratuito" looks much better than "simulador de círculos grátis para jogar". The second option would be the english equivalent for "circle-clicking free game" and that's not suitable at all for reading, in my opinion. Plus, changed "beatmaps aprovados" into "Beatmaps Aprovados" for better capitalization.

### DIFF
--- a/resources/lang/pt-br/home.php
+++ b/resources/lang/pt-br/home.php
@@ -32,7 +32,7 @@ return [
         ],
 
         'slogan' => [
-            'main' => 'simulador de círculos grátis para jogar',
+            'main' => 'simulador de ritmo gratuito',
             'sub' => 'o ritmo está a um clique de distância',
         ],
     ],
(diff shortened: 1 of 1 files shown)


resources/lang/pt-br/home.php
```diff
@@ -51,7 +51,7 @@
             ],
         ],
         'beatmaps' => [
-            'new' => 'Novos beapmaps aprovados',
+            'new' => 'Novos Beatmaps Aprovados',
```
Tarkoviski: Capital letters in titles are rarer in Portuguese.

Kuron-kun: Yep, seems you're correct. Changed the file. Thank you :)

Kuron-kun:
Closed because file was merged into #1179

PR #1138, opened 2017-05-16
Feature/flex grid/store

notbakaneko:
Replacing bootstrap css from store-related pages with a flexbox-derived grid.


resources/assets/less/store.less
```diff
@@ -18,5 +18,6 @@
 
 @import "app";
 
+@import "bem/flex-grid";
```
nanaya: `bem-index.less`.

resources/assets/less/bem/flex-grid.less
```diff
@@ -0,0 +1,53 @@
+.grid {
```
nanaya: license copypasta mising. Also one block per file, named with the block name.

resources/assets/less/bem/flex-grid.less
```diff
@@ -0,0 +1,53 @@
...
+    flex-wrap: wrap;
+  }
+
+  // Use this if we don't want the grid to collapse into a stack.
+  // FIXME: better name
+  &__no-block {
+    display: flex;
+  }
+
+  &--gutters {
```
nanaya: modifiers should be declared before elements

resources/assets/less/bem/flex-grid.less
```diff
@@ -0,0 +1,53 @@
...
+  // size to let other cells expand and fill the gap.
+  &--squash {
+    flex: 0 1 0%;
+  }
+
+  &--fill {
+    flex: 1 0 100%;
+  }
+
+  // compiler doesn't like it if it starts with a number
```
nanaya: hmmmm

resources/assets/less/bem/flex-grid.less
```diff
@@ -0,0 +1,53 @@
+.grid {
+  // 840px seems a big large for this to trigger?
+  @media @desktop {
+    display: flex;
+    flex-wrap: wrap;
+  }
+
+  // Use this if we don't want the grid to collapse into a stack.
+  // FIXME: better name
+  &__no-block {
```
nanaya: wait is this modifier or element D:

notbakaneko:
Hang on, I think I trolled myself with the naming :trollface:

resources/assets/less/bem/grid.less
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  // Use this if we don't want the grid to collapse into a stack.
+  // FIXME: better name
+  &--xs {
+    display: flex;
+  }
+
+  &--gutters {
+    margin-left: -@flex-grid--gutter-width;
```
nanaya: shouldn't the gutters also apply to top and bottom? In case of wrapped cells.

nanaya: Or I guess one can just add margin bottom...

notbakaneko: Top and bottom spacing in grid column layouts tend to be different from the column gutters, so we don't really want to apply them together.

resources/assets/less/bem/grid-cell.less
```diff
@@ -0,0 +1,50 @@
...
+
+.grid-cell {
+  flex: 1;
+  .grid--gutters > & {
+    padding-left: @flex-grid--gutter-width;
+    padding-right: @flex-grid--gutter-width;
+  }
+  // squashes the flex content down to the smallest possible
+  // size to let other cells expand and fill the gap.
+  &--squash {
```
nanaya: can use sort by name

PR #1144, opened 2017-05-17
osu!wiki - redirects

MegaApplePi:
add redirects


MegaApplePi:
#1160

MegaApplePi:
#1160

peppy:
did you mean to close this though? the linked issue seems mildly unrelated

MegaApplePi:
Not really, I had thought that a search feature for the wiki would somehow do more than what redirects could. Though, if the search feature is somewhat unrelated to redirects, I guess I'll keep this going.

MegaApplePi:
(Weird hard reset thingy, but updated to resolve conflicts with base branch.)

nanaya:
make a `/wiki/redirect.txt` in osu-wiki which contains path and its redirect target and parse that instead.

MegaApplePi:
alrighty, I think I got it working now

app/Models/Wiki/Redirect.php
```diff
@@ -0,0 +1,46 @@
...
+
+    public function __construct($path, $locale)
+    {
+        $this->path = $this->cleanPath($path);
+        $this->locale = $locale;
+    }
+
+    public function target()
+    {
+        $redirectList = explode(PHP_EOL, static::fetchContent('redirect.txt'));
```
nanaya: should be cached. And may want to cache full data while at it, stored in a list with redirect source as the key.

```
[
    'pp' => 'Performance_Points',
    'asg' => 'Article_Style_Guide',
]
```

Also may want to remove `/wiki/` prefix.

app/Models/Wiki/Redirect.php
```diff
@@ -0,0 +1,46 @@
...
+namespace App\Models\Wiki;
+
+class Redirect extends Page
+{
+    public $path;
+    public $locale;
+
+    public function __construct($path, $locale)
+    {
+        $this->path = $this->cleanPath($path);
```
nanaya: also make it lowercase and replace space with underscore

app/Http/Controllers/WikiController.php
```diff
@@ -46,7 +46,13 @@ public function show($path)
         $subtitle = array_pop($titles);
 
         if ($page->page() === null) {
-            $status = 404;
+            $redirect = new Wiki\Redirect($path, $this->locale());
+            if ($redirect->target() !== null) {
+                header('Location: '.$redirect->target());
```
nanaya: use `return ujs_redirect(...)`

app/Models/Wiki/Redirect.php
```diff
@@ -0,0 +1,46 @@
...
+        $this->locale = $locale;
```
nanaya: not used

app/Models/Wiki/Redirect.php
MegaApplePi: What do you mean?

kj415j45: `$path` use lowercase without space(replace by underscore)

maybe some method can do that(?)

MegaApplePi: oh, I see, force lowercase, and clean the path. 👌 thanks!

app/Models/Wiki/Redirect.php
```diff
@@ -0,0 +1,63 @@
...
+
+class Redirect extends Page
+{
+    public $path;
+
+    private $cache = [];
+
+    public function __construct($path)
+    {
+        $this->path = $path/*str_replace(' ', '_', strtolower($this->path))*/;
```
nanaya: ??

app/Http/Controllers/WikiController.php
```diff
@@ -46,7 +46,12 @@ public function show($path)
         $subtitle = array_pop($titles);
 
         if ($page->page() === null) {
-            $status = 404;
+            $redirect = new Wiki\Redirect($path);
+            if ($redirect->target() !== null) {
+                return ujs_redirect('wiki/'.$redirect->target());
```
nanaya: `route('wiki.show', $...)`

app/Models/Wiki/Redirect.php
```diff
@@ -0,0 +1,63 @@
...
+                    }
+
+                    if (present($redirect)) {
+                        return $redirect;
+                    }
+                }
+            );
+        }
+
+        if (array_key_exists($this->path, $this->cache['redirect'])) {
```
nanaya: not needed. Just append the return with `?? null`.

app/Models/Wiki/Redirect.php
MegaApplePi: Whoops, that was me trying to see how to get the passed argument `$path` from `Page.php` onto `Redirect.php` :s.

app/Http/Controllers/WikiController.php
MegaApplePi: use `return ujs_redirect(route('wiki.show', $redirect->target()));`?

app/Models/Wiki/Redirect.php
```diff
@@ -0,0 +1,59 @@
...
+
+    public function target()
+    {
+        if (!array_key_exists('redirect', $this->cache) or true) {
+            $this->cache['redirect'] = Cache::remember(
+                $this->cacheKeyPage(),
+                static::CACHE_DURATION,
+                function () {
+                    try {
+                        $redirect = json_decode(OsuWiki::fetchContent('wiki/redirect.json'), true);
```
nanaya: a yaml file is probably easier to write

(looking again, a custom format might be indeed better avoided)

app/Models/Wiki/Redirect.php
```diff
@@ -0,0 +1,59 @@
...
+                    } catch (GitHubNotFoundException $_e) {
+                        return;
+                    }
+
+                    if (present($redirect)) {
```
nanaya: just return immediately on decode `return json_decode(...)` (or `return Yaml::parse(...)`)

app/Models/Wiki/Redirect.php
```diff
@@ -0,0 +1,55 @@
...
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace App\Models\Wiki;
+
+use App\Libraries\OsuWiki;
+use Cache;
+
+class Redirect extends Page
```
nanaya: Just noticed, why does it extend Page?

MegaApplePi: ~~I have not a clue (it just works for me with it there ¯\\\_(ツ)\_\/¯)~~
I guess it is to use one of the cache functions (`cacheKeyPage()`) from Page, otherwise I will need to define it or something?

Also, I left an `or true` expression on line 39... whoops

nanaya: just put `wiki:redirect` for the cache key.

MegaApplePi: what should the `static::CACHE_DURATION` be then, 300?

nanaya: might be better a bit shorter like an hour.

app/Models/Wiki/Redirect.php
```diff
@@ -0,0 +1,55 @@
...
+
+    public function target()
+    {
+        if (!array_key_exists('redirect', $this->cache)) {
+            $this->cache['redirect'] = Cache::remember(
+                'wiki:redirect',
+                60,
+                function () {
+                    try {
+                        return json_decode(OsuWiki::fetchContent('wiki/redirect.json'), true);
```
nanaya: in case you missed it, yaml format is much simpler (just `key: value`) but it's ok if you're fine with json.

MegaApplePi: I'm not familiar with yaml, is it literally `key: value`?

MegaApplePi: the yaml docs I read made it look like it was more than that :s

nanaya: see the header of [news page](https://raw.githubusercontent.com/ppy/osu-wiki/master/news/2015-05-30-osuweekly-12.md) for example.

MegaApplePi: huh, the more I know... ok, I'll switch it over to yaml

PR #1163, opened 2017-05-18
Switch/update to Laravel Mix/Webpack

notbakaneko:
Done with as few changes to existing code as possible, so there's quite a few derps in here :D

Empirical observations are it builds slightly slower than gulp for the initial build, but it's much faster on the rebuild when watching; the webpack headers make the files fatter.

Also has about a 780MB RSS (620MB if called directly without yarn) on my machine when doing a production build vs ~500MB for gulp :|


notbakaneko:
Trying to push all js/coffee files to compile into entry files as `import` statements.
Ideally, we would want to stop messing with the build config file and have the list of files to compile handled via a module/components imports.

The blocking issue for this is dealing with imports in coffeescript while using imports-loader

notbakaneko:
Switching to es6-style imports is a significantly larger and more intrusive change, so everything related to that will be done as separate task.

The scope of the changes to use es6 imports and webpack's module resolution can be viewed here (WIP):
https://github.com/notbakaneko/osu-web/pull/1

webpack.mix.js
```diff
@@ -0,0 +1,149 @@
...
+      },
+      {
+        test: /\.coffee$/,
+        use: ['imports-loader?this=>window', 'coffee-loader']
+      }
+    ]
+  }
+};
+
+// use polling if watcher is bugged.
```
notbakaneko: specifically an issue of chokidar with kqueue

peppy:
why is this being worked on? we have much higher other priorities

PR #1177, opened 2017-05-19
Store exactly the same time for all build histories

LiquidPL:
Will probably be better for eventual grouping.


nanaya:
Add database transaction while at it.

PR #1191, opened 2017-05-24
Add score mode for rankings

nekodex:
...plus some cleanups here and there.


resources/assets/coffee/react/ranking/page.coffee
```diff
@@ -43,19 +42,134 @@ class Ranking.Page extends React.Component
       pages: props.paging.pages
       loading: false
       mode: props.mode
-      rankingType: 'performance' # hard coded until other types are implemented
+      rankingType: props.ranking_type
+
+
+  columnSettings: =>
+    {
```
nanaya: what's with all the braces

PR #1201, opened 2017-05-28
Server-Side ACTION message handling

walle303:
Implement support for action messages in the server, this method does not require any modifications to the database.

I'll rewrite it if IRC-style encoded messages aren't wanted in the database, but from the looks of the system now, the database is probably already full of these.

Changes:
+ Strip all ASCII control characters from incoming and outgoing messages
+ Process ACTION messages and tag them with an is_action flag
+ Drop the maximum message length from 1024 characters to 1000 characters, This is done to leave room for the ACTION message tagging
+ Process incoming messages with is_action into ACTION messages

This pairs with ppy/osu#860

### DIFF
--- a/app/Http/Controllers/API/ChatController.php
+++ b/app/Http/Controllers/API/ChatController.php
@@ -105,10 +105,15 @@ class ChatController extends Controller
 
     public function postMessage()
     {
-        if (mb_strlen(Request::input('message'), 'UTF-8') >= 1024) {
+        $messagecontent = preg_replace('/[\x00-\x1f\x7F]/', '', Request::input('message'));
+
+        if (mb_strlen(Request::input('message'), 'UTF-8') >= 1000) {
             abort(422);
         }
 
+        if (Request::input('is_action') === 'True') {
+            $messagecontent = "\001ACTION ".$messagecontent."\001";
+        }
         switch (Request::input('target_type')) {
             case 'channel':
                 $target = Channel::findOrFail(Request::input('target_id'));
@@ -136,7 +141,7 @@ class ChatController extends Controller
             return error_popup(trans('api.error.chat.limit_exceeded'), 429);
         }
 
-        $message = $target->receiveMessage(Auth::user(), Request::input('message'));
+        $message = $target->receiveMessage(Auth::user(), $messagecontent);
 
         return json_item(
             $message,
--- a/app/Transformers/API/Chat/MessageTransformer.php
+++ b/app/Transformers/API/Chat/MessageTransformer.php
@@ -38,7 +38,8 @@ class MessageTransformer extends Fractal\TransformerAbstract
             'target_type' => $message->target_type,
             'target_id' => $message->target_id,
             'timestamp' => json_time($message->timestamp),
-            'content' => $message->content,
+            'content' => preg_replace('/\x01ACTION |[\x00-\x1f\x7F]/', '', $message->content),
+            'is_action' => preg_match('/\x01ACTION\s.*\x01/', $message->content),
         ];
     }
 


app/Http/Controllers/API/ChatController.php
```diff
@@ -105,10 +105,15 @@ public function privateMessages()
 
     public function postMessage()
     {
-        if (mb_strlen(Request::input('message'), 'UTF-8') >= 1024) {
+        $messagecontent = preg_replace('/[\x00-\x1f\x7F]/', '', Request::input('message'));
```
nanaya: should be in model (see below). And `1F` for consistency.

app/Http/Controllers/API/ChatController.php
```diff
@@ -105,10 +105,15 @@ public function privateMessages()
...
+
+        if (mb_strlen(Request::input('message'), 'UTF-8') >= 1000) {
             abort(422);
         }
 
+        if (Request::input('is_action') === 'True') {
```
nanaya: use `get_bool` (update the helper if needed)

app/Http/Controllers/API/ChatController.php
```diff
@@ -105,10 +105,15 @@ public function privateMessages()
...
+            $messagecontent = "\001ACTION ".$messagecontent."\001";
```
nanaya: make a mutator in model which accepts array of `content` and `is_action` which then clean up and set content accordingly. And `\x01` for consistency.

(put in trait for now until someone writes better abstraction between two Message models)

app/Transformers/API/Chat/MessageTransformer.php
```diff
@@ -38,7 +38,8 @@ public function transform($message)
             'target_type' => $message->target_type,
             'target_id' => $message->target_id,
             'timestamp' => json_time($message->timestamp),
-            'content' => $message->content,
+            'content' => preg_replace('/\x01ACTION |[\x00-\x1f\x7F]/', '', $message->content),
```
nanaya: accessor in model. `ACTION` matcher should start with `^`

app/Transformers/API/Chat/MessageTransformer.php
```diff
@@ -38,7 +38,8 @@ public function transform($message)
...
+            'is_action' => preg_match('/\x01ACTION\s.*\x01/', $message->content),
```
nanaya: also in model. And using `starts_with` should be enough instead of full regexp (which also missing `^` and `$`).

app/Http/Controllers/API/ChatController.php
```diff
@@ -136,7 +141,7 @@ public function postMessage()
             return error_popup(trans('api.error.chat.limit_exceeded'), 429);
         }
 
-        $message = $target->receiveMessage(Auth::user(), Request::input('message'));
+        $message = $target->receiveMessage(Auth::user(), $messagecontent);
```
nanaya: call the variable `$body` (matches what the function definition calls it)

peppy:
After further discussion, we'll change the database structure for this now (and add an `is_action` column to `messages` and `messages_private`). Did you want to look at adding a database migration for this in osu-web and removing the custom handling of ACTION control characters from this PR?

peppy:
Need a migration added to match
`ADD COLUMN is_action TINYINT NOT NULL DEFAULT '0' AFTER content`
on `osu_chat.messages` and `osu_chat.messages_prviate`

walle303:
Ah, I was going to do that in the first place, but I don't know much about Laravel, I'll look into it though.
Is the code that parses messages from Bancho and stores them in the database available anywhere, that would need to be updated to handle action messages as well.

PR #1210, opened 2017-05-31
Do not break line in download button and the span under the button

### LINKED ISSUE #1208: Unexpected break line
English Version:(It should not break line between "operating" and "systems")
image
Chinese Version:(It should not break line between "Windows" and "版"   &&   "osu!" and "即将到来")
image

benpigchu:
fix #1208


nanaya:
It'll be fixed in different way together with design update for mobile.

PR #1222, opened 2017-06-02
return empty array if images is null.

notbakaneko:
So iterators don't burst into flames

### DIFF
--- a/app/Models/Store/Product.php
+++ b/app/Models/Store/Product.php
@@ -115,7 +115,7 @@ class Product extends Model
                 $this->images = json_decode($this->images_json, true);
             }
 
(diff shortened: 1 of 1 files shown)


app/Models/Store/Product.php
```diff
@@ -115,7 +115,7 @@ public function images()
                 $this->images = json_decode($this->images_json, true);
             }
 
-            return $this->images;
+            return $this->images ?: [];
```
nanaya: `??` is better because `true` can be anything in php.

notbakaneko: Oh right, php7 actually has a `??` operator now >_>

PR #1227, opened 2017-06-05
Country rankings and filtering by country

nekodex:
Uh so this ended up being bigger a bigger change than planned...:
- adds ability to view country rankings (now per game mode!)
- adds ability to view scores filtered by country
- remove react based score listing and replaces with blade one (relying on turbolinks for speed)
- adds new db table and cron tasks to calculate per-game-mode country rankings
- looks like buttes on mobile, will fix... later™

The ranking blades are a bit... duplicated. Works for now, but it would be nice to come up with a way to generate/share the tables or something later.


app/Http/Controllers/RankingController.php
```diff
@@ -30,50 +33,71 @@ class RankingController extends Controller
...
+
+            $stats = CountryStatistics::where('display', 1)
+                ->with('country')
+                ->where('mode', $modeInt)
+                ->orderBy('performance', 'desc')
+                ->limit(static::PAGE_SIZE)
+                ->offset(static::PAGE_SIZE * ($page - 1));
+
+            $scores = json_collection($stats->get(), 'CountryStatistics', ['country']);
+        } else { // if $type == 'performance' || $type == 'score'
```
nanaya: can use a fallback when `$type` is invalid to avoid rendering random things.

resources/assets/less/bem/ranking-page-table.less
```diff
@@ -35,7 +73,18 @@
...
 
   &__rank-column {
-    display: flex;
-    justify-content: flex-end;
+    width: 50px;
+    text-align: center;
+    opacity: 1;
+  }
+
+  &__rounded-left {
```
nanaya: or just `:first-child` and `:last-child`

resources/assets/less/bem/ranking-page-table.less
```diff
@@ -17,11 +17,49 @@
...
 
 .ranking-page-table {
-  background-color: #333;
-  padding: 15px 20px 0;
-  margin-bottom: 10px;
-  color: #fff;
-  font-size: @font-size--normal;
+  min-width: 650px;
+  width: 100%;
+  text-align: center;
```
nanaya: right alignment for numbers seems better

resources/views/rankings/country.blade.php
```diff
@@ -0,0 +1,87 @@
...
+                    <td class="ranking-page-table__rounded-left ranking-page-table__rank-column">
+                        #{{ $scores->firstItem() + $index }}
+                    </td>
+                    <td>
+                        <a class="ranking-page-table__user-link" href="{{route('ranking', [
+                            'mode' => $mode,
+                            'type' => 'performance',
+                            'country' => $score['code'],
+                        ])}}">
+                            @include('objects._country-flag', [
```
nanaya: snake case for view filename?

resources/assets/less/bem/ranking-page-table.less
```diff
@@ -17,11 +17,49 @@
...
+  &__column {
+    &--dimmed {
+      opacity: 0.6;
+    }
+  }
+
+  &__row {
+    background: #444;
+    border-bottom: 3px solid #333;
+    height: 34px;
```
nanaya: not really needed

resources/assets/less/bem/ranking-page-table.less
```diff
@@ -17,11 +17,49 @@
...
+
+  &__column {
+    &--dimmed {
+      opacity: 0.6;
+    }
+  }
+
+  &__row {
+    background: #444;
+    border-bottom: 3px solid #333;
```
nanaya: apparently a bit broken in chrome. Perhaps just use table's spacing.

resources/assets/less/bem/ranking-page-header.less
```diff
@@ -37,4 +39,27 @@
...
     text-transform: capitalize;
   }
+
+  &__flag {
+    .link-plain();
+    &:hover {
+      .flag-country {
+        opacity: 0.5;
+      }
+      .@{top}__flag-overlay {
```
nanaya: put in `__flag-overlay`?

resources/assets/less/bem/ranking-page-header.less
```diff
@@ -37,4 +39,27 @@
...
+    }
+  }
+
+  &__flag-overlay {
+    display: none;
+    pointer-events: none;
+    font-size: 20px;
+    position: absolute;
+    left: 2px;
+    top: 10px;
```
nanaya: this just so weird. Make `__flag` an inline flex and turn this into actual overlay (with background color etc).

resources/views/objects/_mode-selector.blade.php
```diff
@@ -0,0 +1,28 @@
...
+    osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+    See the GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+--}}
+
+<ul class="page-mode">
+    @foreach (['osu', 'taiko', 'fruits', 'mania'] as $tab)
```
nanaya: can use `MODES` array

app/helpers.php
```diff
@@ -834,3 +834,30 @@ function clamp($number, $min, $max)
 {
     return min($max, max($min, $number));
 }
+
+// e.g. 100634983048665 -> 100.63 trillion
+function suffixed_number_format($number)
+{
+    $suffixes = ['', 'k', 'millon', 'billion', 'trillion']; // TODO: localize
```
nanaya: this may actually involve more than just changing suffixes (think Japanese 万 億 etc) 💃

resources/assets/less/bem/ranking-page.less
```diff
@@ -0,0 +1,26 @@
...
+ *    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *    See the GNU Affero General Public License for more details.
+ *
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+.ranking-page {
+  background-color: #333;
+  padding: 15px 20px 0;
```
nanaya: the padding needs to be on the parent/`osu-page`

app/Http/Controllers/RankingController.php
```diff
@@ -30,50 +33,76 @@ class RankingController extends Controller
...
-            ->offset(static::PAGE_SIZE * ($page - 1));
+        $country = null;
+        $modeInt = Beatmap::modeInt($mode);
+
+        if ($type === 'country') {
+            $maxResults = CountryStatistics::where('display', 1)
+                ->where('mode', $modeInt)
+                ->count();
+
+            $maxPages = ceil($maxResults / static::PAGE_SIZE);
```
nanaya: can be moved out as well

PR #1241, opened 2017-06-09
Fix missing ']' in store-supporter-tag querySelector.

notbakaneko:
Apparently, Safari was the only browser that cared about the missing `]`


nanaya:
thanks jquery =D

PR #1249, opened 2017-06-14
Beatmapsets search refactor

### LINKED ISSUE #552: Beatmap links shouldn't open in a new window/tab

### LINKED ISSUE #908: Beatmap Listing Filter issue [ "Minor" ]
Hi.

While i was wandering on the website, i found a problem with the links in the beatmap listing.

**Explanation**:

When you arrive on the beatmap listing page, you are on this link **https://new.ppy.sh/beatmapsets**
 Then when you change the filter for modes (example: Mania) the link update to **.../beatmapsets/?m=3**
    ~~~
![po1](https://cloud.githubusercontent.com/assets/25019842/21788896/9951b61e-d6d1-11e6-981e-edab1a84f74b.PNG)

Obviously **m=3** for Mania / **m=2** Catch / **m=1** Taiko and **m=0** for Osu! (STD).

**Here is the problem** : Since **m=0** is a blank value, it disappear when you change to Osu!. 
    ~~~
po2

But, the actual link for "Any mode" filter is **m=none**.
    ~~~
![po3](https://cloud.githubusercontent.com/assets/25019842/21788918/b9948d02-d6d1-11e6-87b5-95391387bc68.PNG)


So actually, we can't switch from Osu! to Any (and the other way around) since they are considered the same and won't update/resfresh the list.

**P.S**: You can forcefully update from Any to Osu! by adding m=0 (like this https://new.ppy.sh/beatmapsets/?m=0 ) to the link then refesh the page and same goes for Osu! to Any (with https://new.ppy.sh/beatmapsets/? ) but most people wouldn't even know this.

### LINKED ISSUE #1142: Beatmap listing shows all maps when only osu!std is chosen by default
When you first arrive at the beatmap listing, you can see "osu!" as the highlighted gamemode, so it theoretically should list only osu!std beatmaps, but instead it shows all beatmaps, as if the "Any" option was chosen. 

unknown-1

If you click "osu!" manually, it properly lists osu!std beatmaps only.

I'm fairly certain it's related to #908.

nanaya:
This thing has gotten a bit larger than I expected 8D

- actual search on main page load
- don't unmount react but save state instead on navigation; fixes #552, closes #553
- added total to `Beatmapset::search()` result
- less confusing castings which also fixes #908 and fixes #1142
- fixed bugs
- added bugs

### DIFF
--- a/resources/assets/coffee/react/beatmaps/main.coffee
+++ b/resources/assets/coffee/react/beatmaps/main.coffee
@@ -203,10 +203,6 @@ class Beatmaps.Main extends React.PureComponent
 
       state[key] = value
 
-      if key == 'query'
-        # ;_;
-        @initialInput = value
-
     filters: state
     isExpanded: expand
 
--- a/resources/assets/coffee/react/beatmaps/main.coffee
+++ b/resources/assets/coffee/react/beatmaps/main.coffee
@@ -171,7 +171,7 @@ class Beatmaps.Main extends React.PureComponent
           page: 1
           url: @state.paging.url
           loading: false
-          more: data.length > 10
+          more: data.length > 0
         loading: false
 
       @setState newState, ->
--- a/resources/assets/coffee/react/beatmaps/main.coffee
+++ b/resources/assets/coffee/react/beatmaps/main.coffee
@@ -25,7 +25,9 @@ class Beatmaps.Main extends React.PureComponent
 
     @xhr = {}
 
-    @state = JSON.parse(props.container.dataset.reactState ? null)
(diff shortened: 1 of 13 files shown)


resources/assets/coffee/react/beatmaps/main.coffee
```diff
@@ -19,258 +19,202 @@
...
+      dataType: 'json'
+      data: params
+    .done (data) =>
+      newState =
+        beatmaps: data
+        paging:
+          page: 1
+          url: @state.paging.url
+          loading: false
+          more: data.length > 10
```
nekodex: guess this should become `> 0` too for consistency?

PR #1250, opened 2017-06-14
Add check for supporter duration against amount donated.

notbakaneko:
This is to make the store refuse to add an invalid donation into the cart, in additional to the transaction validation.


app/Models/Store/Order.php
```diff
@@ -283,10 +284,14 @@ private function newOrderItem(array $params)
...
         if ($product->custom_class === 'supporter-tag') {
-            $targetUsername = $params['extraData']['username'];
-            $user = User::where('username', $targetUsername)->first();
-            // TODO: burst into flames if user is null
-            $params['extraData']['target_id'] = $user->user_id;
+            $targetId = $params['extraData']['target_id'];
+            $user = User::where('user_id', $targetId)->firstOrFail();
+            $params['extraData']['username'] = $user->username;
+
+            $duration = $params['extraData']['duration'];
```
nanaya: why not just `= ...:getDuration($params['cost'])`?

notbakaneko: I recall being trolled by that, but it might make more sense...

app/Models/Store/Order.php
```diff
@@ -283,10 +284,11 @@ private function newOrderItem(array $params)
 
         // FIXME: custom class stuff should probably not go in Order...
         if ($product->custom_class === 'supporter-tag') {
-            $targetUsername = $params['extraData']['username'];
-            $user = User::where('username', $targetUsername)->first();
-            // TODO: burst into flames if user is null
-            $params['extraData']['target_id'] = $user->user_id;
+            $targetId = $params['extraData']['target_id'];
+            $user = User::where('user_id', $targetId)->firstOrFail();
```
nekodex: This might be a bit unrelated, but are the target users checked to see if they're restricted/banned before allowing supporter tags to be gifted to them? I feel like that should probably be a thing?

notbakaneko: We should probably consider that as a separate issue

PR #1253, opened 2017-06-15
Always show accuracy in play-details

### LINKED ISSUE #1146: Top ranks on profile are missing accuracy display
I think we need to get this back in the design.

image

notbakaneko:
resolves #1146 

There's already logic to show the accuracy if there's no `score.weight` or `score.pp`; this just makes it show all the time.
It might look kind of unbalanced, though?

image

### DIFF
--- a/resources/assets/coffee/react/_components/play-detail.coffee
+++ b/resources/assets/coffee/react/_components/play-detail.coffee
@@ -74,7 +74,7 @@ el = React.createElement
                 else
                   score.score.toLocaleString()
             div
-              className: 'detail-row__detail-row detail-row__score-details'
+              className: 'detail-row__score-details'
               if score.weight
                 div
                   className: 'detail-row__text-score'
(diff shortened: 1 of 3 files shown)


arflyte:
Increase the padding between pp and weighted, and reduce the line height between weighted and accuracy

resources/assets/coffee/react/_components/play-detail.coffee
```diff
@@ -73,17 +73,15 @@ el = React.createElement
                   osu.trans('users.show.extra.top_ranks.pp', amount: Math.round(score.pp))
                 else
                   score.score.toLocaleString()
-            if score.weight
-              div
-                className: 'detail-row__detail-row detail-row__detail-row--bottom'
-                span
+            div
+              className: 'detail-row__detail-row detail-row__score-details'
```
nanaya: more like `detail-row__detail-row--score-details`

notbakaneko:
now with 100% less bad renderer
image

notbakaneko:
I've just noticed the mod icons don't show on Safari? ಠ_ಠ

notbakaneko:
Safari doesn't consider the background-image as part of its content sizing, so the height of the mod icons block in Safari is 0 🎉

notbakaneko:
separate issue for missing mod icons https://github.com/ppy/osu-web/issues/1256

PR #1257, opened 2017-06-16
fixes #1256

### LINKED ISSUE #1256: Mod icons don't show in Safari
Chrome:
Chrome

Safari:
Safari

(ignore the text position differences >_>)

notbakaneko:
Sets `min-height` so Safari doesn't consider it a 0 height block.

### DIFF
--- a/resources/assets/less/bem/mod.less
+++ b/resources/assets/less/bem/mod.less
@@ -17,7 +17,7 @@
  */
 
 .mod {
-  flex: none;
+  position: absolute;
   width: 100%;
   height: 100%;
   background-size: contain;


resources/assets/less/bem/mod.less
```diff
@@ -17,6 +17,18 @@
  */
 
 .mod {
+  .mods & {
```
nanaya: using `.full-size()` seems to work as well (and remove that `flex`/`width`/`height`).

notbakaneko: `position: absolute` by itself seems to fix it; I'm not sure on the need for `left`, `top` or `content`

notbakaneko: don't seem to need to others, so just going to set `position: absolute` for now ┐(￣ー￣)┌

nanaya:
Might as well squash and rebase.

PR #1270, opened 2017-06-19
More search fixes

nanaya:
See commit log =D

### DIFF
--- a/resources/views/home/_search_results.blade.php
+++ b/resources/views/home/_search_results.blade.php
@@ -22,7 +22,8 @@
                 @lang("home.search.{$mode}.title")
             </h2>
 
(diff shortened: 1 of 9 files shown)


resources/views/home/_search_results.blade.php
```diff
@@ -22,7 +22,8 @@
                 @lang("home.search.{$mode}.title")
             </h2>
 
-            @if (empty($result['data']))
+            {{-- `empty(collect())` is false :D --}}
```
nanaya: masterpiece 🥇

nekodex: nanaya pls

Issue #220, opened 2015-11-18
Make page descriptions more descriptive


peppy:
At the moment it's pretty generic.

![](https://puu.sh/lpMqF/04751e041d.png)

Old site for reference:

![](https://puu.sh/lpMtb/9c4ff9d9b0.png)

Lemmmy:
Adding onto this, opengraph should be used where possible. For example, forum threads could have the covers as an `og:image`, profiles the profile picture etc.

Also, for beatmaps:
`og:title` should be `Artist - Title [Difficulty] (mapped by Mapper)`
`og:description` should contain some information about the map like its length, BPM, stats
`og:image` the map thumbnail
`og:type` could be `music.song` (to include duration)

TPGPL:
Is this one done or what else needs to be done here?

peppy:
I think the main culprits have been resolved. Will close for now.

PR #1273, opened 2017-06-20
Set page description meta on forum

notbakaneko:
Sets the `description` meta tag according to the forum page being viewed, instead of showing the default.

ref: #220


notbakaneko:
Might be better to serialize the models as opengraph objects and extract the necessary info from that...

app/Models/Forum/Forum.php
```diff
@@ -216,4 +216,16 @@ public function isOpen()
     {
         return $this->forum_type === 1;
     }
+
+    public function toMetaDescription()
+    {
+        $stack = [__('forum.title')];
```
nekodex: I don't think we use `__` anywhere else in the codebase, probably better to remain consistent and use `trans`.

PR #1276, opened 2017-06-21
Set page description on beatmapset page

notbakaneko:
Sets the description meta tag when viewing a beatmapset

ref: #220


notbakaneko:
I was originally going to add opengraph data in for this and other pages as well, but I think there needs to be more discussion on that to standardize the data, formatting and implementation to be used, instead of just shoving `pageDescription` around.

nekodex:
I think these page descriptions should also have `osu!` in them, not just `beatmaps / whatever`

PR #1282, opened 2017-06-25
Translations in CoffeeScript

Nekonyx:
Will correct translations in several places.
image
image
image


nanaya:
there's originally bug in Lang.js which doesn't do fallback at all and thus `osu.trans` was made. If it now works then it'd be better to revert back to using original `Lang.get` and remove the `osu.trans` (and other) functions.

Nekonyx:
Yep, fallbacks work. You can check demo here: http://nneko.ru/beatmapsets or
image

PR #1292, opened 2017-07-04
Always cache fast_imagesize result

nanaya:
Because `null` isn't cached ( ･_･)


app/helpers.php
```diff
@@ -602,8 +602,18 @@ function fast_imagesize($url)
         $data = curl_exec($curl);
         curl_close($curl);
 
-        return read_image_properties_from_string($data);
+        $result = read_image_properties_from_string($data);
```
notbakaneko: and then I discovered I left the comment as pending :D

nanaya: returning null allows null-coalescing stuff in case default is available `myfunc() ?? ['default']`

notbakaneko: speaking of null-coalescing... https://github.com/ppy/osu-web/blob/e5cf1f2/app/Models/Wiki/Image.php#L57 O_o

Issue #1280, opened 2017-06-25
Make a second column for language selection


leonekmi:
![Screenshot 1](http://i.imgur.com/VeNAOcK.png)

Too much languages is supported, a second column will be necessary or a modal for language choice.

kj415j45:
nanaya used to edited less to increse its height, but after `ru` added, it short again(

second column may be a good choice...

benpigchu:
【 Make the list scrollable is also an option 】
【 Maybe in the future, the secound column is also not enough 】

kj415j45:
also, you can't select language on mobile phone :(

kj415j45:
this issue may happen again if new language add.

MegaApplePi:
Maybe a button that says `Other...` then display a dialogue to select the other locales that aren't listed?

notbakaneko:
@arflyte ↑

Maybe make the height of the menu dependent on the list being shown?
The main drop down actuallly seems like an odd place to put the language switcher.

kj415j45:
i think `<select/>` can be a good choice

leonekmi:
https://github.com/ppy/osu-web/commit/40916801d98135383145d6caed19e590d099a292 fixes this issue.
![image](https://user-images.githubusercontent.com/20224422/38686467-ab31edac-3e74-11e8-8495-10aa1c0f0dfd.png)

PR #1294, opened 2017-07-05
Even taller menu

nanaya:
Whee 💃 

(avoid language menu overflow)


kj415j45:
any other better solution? 
ref: #1280

Issue #1308, opened 2017-07-12
Approval maps should only require 2 nominations


Loctav:
*All* maps only need **2** nominations these days - this has been changed recently in the Ranking Criteria. Right now, Approval maps (>5 minutes drain time) still need 3 nominations on the modding v2 discussion panel.

PR #1310, opened 2017-07-12
Approval maps only require 2 nominations

turbedi:
Should fix issue #1308 (?)


peppy:
A candidate for an .env variable, but as this will likely change once the v2 system goes live (to be dynamic based on map length, user reputation etc.) it's not worth adding for now.

PR #1314, opened 2017-07-13
[PT-BR] Update Translations

gustarrini:


kj415j45:
you have to know, all words after `:` means it is parameter, never translate it

like `:count` should never be translate

but something like 'username: xxxx', you can translate everything(username,xxxx).

also, some singular-plural issues you need to notice, here's refer question: http://stackoverflow.com/questions/31775626/why-does-laravels-trans-choice-always-show-the-singular-case

and if you have other singular-plural issues, you can pm me on discord, i can tell you how to make it works correctly.

kj415j45:
Another thing:

I create Style CI fix PR for you, it will save your time, please check then merge it

resources/lang/pt-br/supporter_tag.php
```diff
@@ -0,0 +1,30 @@
...
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+return [
+    'user_search' => [
+        'searching' => 'procurando...',
+        'not_found' => 'Esse usuário não existe!',
+    ],
+    'duration' => [
+        'months' => ':lenght mês|:lenght meses|{0}',
```
nanaya: as mentioned, this and the next line need to be fixed.

gustarrini: Will fix It asap.

PR #1319, opened 2017-07-14
Corrections for Russian Translation #1

Nekonyx:
Can be merged but do not please. I expect other changes that may occur.


Nekonyx:
Ready :tada:

PR #1321, opened 2017-07-14
Even more fixes

### LINKED ISSUE #1320: Friends listing; Impossible to remove friends
It says you need to log in, even if you are already.

nanaya:
Fixes #1320 and another kind of related bugs.


nanaya:
Now with rank check because turned out it's easy enough.

PR #1333, opened 2017-07-18
Usercard search

nanaya:
Now with card instead of just avatar.

Also fix alignment stuff.

### DIFF
--- a/resources/views/objects/_usercard.blade.php
+++ b/resources/views/objects/_usercard.blade.php
@@ -15,9 +15,22 @@
     You should have received a copy of the GNU Affero General Public License
     along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
 --}}
-
+@php
+    $blockClass = 'usercard';
+    if (isset($popup) && $popup) {
+        $blockClass += ' usercard--popup';
+    }
+    if (count($_modifiers ?? null) > 0) {
+        foreach ($_modifiers as $modifier) {
+            $blockClass .= ' usercard--'.$modifier;
+        }
+    }
+@endphp
 @if (isset($user) || isset($loading))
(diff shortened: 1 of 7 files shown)


resources/views/objects/_usercard.blade.php
```diff
@@ -34,22 +47,26 @@
                     <div class="usercard__username">{{isset($user) ? $user->username : 'Loading...'}}</div>
                     <div class="usercard__flags">
                         @if (isset($loading))
-                            @include('objects._country_flag', ['country_code' => 'XX'])
+                            <div class="usercard__flag">
```
nekodex: move this div to outside of `@if`?

resources/views/objects/_usercard.blade.php
```diff
@@ -34,22 +47,26 @@
...
-                                <div class="js-react--friendButton" data-target="{{$user->user_id}}"></div>
+                            <div class="usercard__flag">
+                                @if (isset($user->country))
+                                    @include('objects._country_flag', [
+                                        'country_code' => $user->country->acronym,
+                                        'country_name' => $user->country->name,
+                                    ])
+                                @endif
+                            </div>
+                            <div class="usercard__flag">
```
nekodex: wait, wut ಠ_ಠ

nekodex:
icon alignment in the friend button seems off now (uncentered?)

before:

after:

nanaya:
But I didn't change the button styling at all ಠ_ಠ

nanaya:
Apparently because it was just accidentally centered, not properly centered.

nekodex:
the icon alignment during hover is still misaligned..

nanaya:
which icon?

nekodex:
friend button icon

PR #1334, opened 2017-07-18
Guest friend button

nanaya:
Don't render the button if current user is guest.


nanaya:
I need this to fix something else in #1333.

PR #1337, opened 2017-07-18
Js fixes

nanaya:
More reactTurbolinks thing, and saving state of beatmap page. The latter is still kind of buggy.


Nekonyx:
IE11? 😮

Issue #962, opened 2017-02-09
No longer displaying three beatmap panels wide on listing


peppy:
Somehow became 2 wide 😦

nekodex:
Dependent on #1339

PR #1338, opened 2017-07-18
Three beatmap panels on listing

Nekonyx:
Fixes: #962


nekodex:
Thanks for the contribution, however we need to do some design work (to increase page width) before doing this otherwise it ends up looking bad. Gonna put this on hold for now.

notbakaneko:
Closing this until there are updates on the design since it's likely other changes will be needed.

PR #1347, opened 2017-07-20
More sentry fixes

nekodex:
Add user locale to error metadata and fix sourcemap generation

### DIFF
--- a/resources/views/layout/metadata.blade.php
+++ b/resources/views/layout/metadata.blade.php
@@ -46,14 +46,14 @@
 <script src="{{ mix("js/vendor.js") }}" data-turbolinks-track="reload"></script>
 @if(config('services.sentry.public_dsn') !== '')
     <script src="//cdn.ravenjs.com/3.17.0/raven.min.js" crossorigin="anonymous"></script>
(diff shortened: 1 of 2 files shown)


resources/views/layout/metadata.blade.php
```diff
@@ -46,13 +46,14 @@
...
-    <script>Raven.config('{{ config('services.sentry.public_dsn') }}', {
-        release: '{{ config('osu.git-sha') }}'
-    }).install();</script>
+    <script>Raven.config('{{ config('services.sentry.public_dsn') }}', {release: '{{ config('osu.git-sha') }}'}).install();</script>
 @endif
 <script src="{{ mix("js/app-deps.js") }}" data-turbolinks-track="reload"></script>
 <script src="{{ mix("js/app.js") }}" data-turbolinks-track="reload"></script>
 <script src="/vendor/js/timeago-locales/jquery.timeago.{{ locale_for_timeago(Lang::getLocale()) }}.js" data-turbolinks-track="reload"></script>
+@if(config('services.sentry.public_dsn') !== '')
+    <script>Raven.setUserContext({lang: Lang.locale});</script>
```
nanaya: why is it not right after `Raven.config` thingy above?

nanaya: (current locale is available from `currentLocale`)

PR #1350, opened 2017-07-21
Even more js fixes

nanaya:
Race condition, overloading class (not quite fixed), wrong class, and ujs/turbolinks stuff.


resources/assets/coffee/_classes/account-edit.coffee
```diff
@@ -28,6 +28,8 @@ class @AccountEdit
   initializeUpdate: (e) =>
     form = e.currentTarget
 
+    return if form.dataset.accountEditAuto != '1'
```
nekodex: auto what?

nanaya: all the autos =D

PR #1355, opened 2017-07-24
Position qtip with clientRect instead of bounding box

### LINKED ISSUE #1248: tooltip is centred if link text is overflowing
If the link's text is overflowing into a new line, the hover text is centred.

notbakaneko:
closes #1248

with a fork of qtip2
https://github.com/notbakaneko/qTip2/compare/master...position-with-client-rect


nanaya:
das coo

notbakaneko:
you saw nothing

Issue #1346, opened 2017-07-20
Group listing page


peppy:
We need the [group/team listing page](https://osu.ppy.sh/g/7) to be re-implemented. Should likely use the exact same logic as the friend listing, including online/offline display.

lunafreya:
You should have to consider this: #528 is STILL WIP right now. But about the group, are you intending to migrate old osu!web groups into new osu!web teams?

nekodex:
Teams will be a seperate (new) system, this is different.

Also, done.

PR #1356, opened 2017-07-24
Groups page

nekodex:
Adds group listing pages for the various osu! team groups.

Implements #1346


resources/views/groups/show.blade.php
```diff
@@ -0,0 +1,43 @@
...
+            <h1 class="osu-page-header__title">
+                {{ $group->group_name }}
+            </h1>
+            <div class="osu-page-header__title osu-page-header__title--small">
+                {{ $group->group_desc }}
+            </div>
+        </div>
+    </div>
+    <div class="osu-page osu-page--generic osu-page--dark-bg">
+        @foreach (['online', 'offline'] as $status)
```
nanaya: I've been wondering why is it not just plain object `['online' => [...], 'offline' => [...]]` and loop that instead (`$status => $users`).

app/helpers.php
```diff
@@ -926,3 +926,24 @@ function format_percentage($number, $precision = 2)
 {
     return sprintf("%.{$precision}f%%", round($number, $precision));
 }
+
+function group_users_by_online_state(&$users)
```
nanaya: no need to pass by reference?

app/Http/Controllers/GroupsController.php
```diff
@@ -0,0 +1,51 @@
...
+        $users = $group->users()
+            ->with([
+                'country',
+                'userProfileCustomization',
+            ])
+            ->default()
+            ->orderBy('username', 'asc')
+            ->get();
+
+        $grouped = group_users_by_online_state($users);
```
nanaya: Maybe also update `Friends#index` to use this.

app/helpers.php
```diff
@@ -926,3 +926,24 @@ function format_percentage($number, $precision = 2)
...
+
+    foreach ($users as $user) {
+        if ($user->isOnline()) {
+            $online[] = $user;
+        } else {
+            $offline[] = $user;
+        }
+    }
+
+    $online = collect($online);
```
nanaya: not sure using collect is worth it just to do `$stuff->count()` instead of `count($stuff)`.

resources/assets/less/bem/usercard.less
```diff
@@ -60,6 +60,10 @@
     .@{top}--popup & {
       opacity: 0;
     }
+
+    // to prevent annoying white-pixel artefacts appearing due to
+    // browsers not handling 'overflow: hidden' properly
+    border-radius: 4px;
```
nanaya: default border radius?

app/helpers.php
nekodex: TIL pass-by-reference in PHP is actually _slower_... 🤦‍♂️

PR #1358, opened 2017-07-25
Make BeatmapSet cover generation more robust

nekodex:
Now scans all Beatmaps in the sets to try and find a valid image for the BeatmapSet (instead of just trying the first Beatmap and giving up). Also caught and silenced some exceptions.


app/Models/Beatmapset.php
```diff
@@ -601,29 +601,34 @@ public function regenerateCovers()
...
-            $beatmap = $this->beatmaps()->first();
-            $beatmapFilename = $beatmap->filename;
-            $bgFilename = self::scanBMForBG("$workingFolder/$beatmapFilename");
+            // scan through all the beatmaps in this set and find the first one with a valid background image
+            foreach ($this->beatmaps as $beatmap) {
+                $bgFilename = self::scanBMForBG("{$workingFolder}/{$beatmap->filename}");
 
-            if (!$bgFilename) {
-                $this->update(['cover_updated_at' => $this->freshTimestamp()]);
+                if (!$bgFilename) {
```
nanaya: just noticed this will fail if filename is `0` =D

nekodex: pretty sure `0` is not a valid image filename

app/Models/Beatmapset.php
```diff
@@ -601,29 +601,34 @@ public function regenerateCovers()
...
-            $beatmap = $this->beatmaps()->first();
-            $beatmapFilename = $beatmap->filename;
-            $bgFilename = self::scanBMForBG("$workingFolder/$beatmapFilename");
+            // scan through all the beatmaps in this set and find the first one with a valid background image
+            foreach ($this->beatmaps as $beatmap) {
+                $bgFilename = self::scanBMForBG("{$workingFolder}/{$beatmap->filename}");
 
-            if (!$bgFilename) {
-                $this->update(['cover_updated_at' => $this->freshTimestamp()]);
+                if ($bgFilename === false) {
```
nanaya: and then there might be case when the beatmap file exists but no background defined?

nekodex: Then it's irrelevant?

nekodex: I mean, it will just loop through the beatmaps, not find any background and not generate any cover? As you would expect?

nanaya: hrm, the previous one also fail on `''` though?

nanaya: oh =S

nanaya:
Oh whoops :D

PR #1362, opened 2017-07-25
Fix beatmapset friend ranking.

nekodex:
Missed a reference when refactoring stuff, whoops.


nanaya:
And there's another thing forgotten :dancer: #1361.

PR #1368, opened 2017-07-26
Helper method for printing login required message and link

notbakaneko:
Might be less confusing to use compared to self-assembly ┐(´д｀)┌


app/helpers.php
```diff
@@ -196,6 +196,15 @@ function read_image_properties_from_string($string)
     }
 }
 
+function require_login($text_key, $link_text_key)
```
notbakaneko: Could make this accept a single parameter and do the `_` and `link_text` lookups implicitly, but that might not be obvious when adding translations...

nanaya: or just to make the link so `text_key` can have another parameters in addition to `link`?

PR #1373, opened 2017-07-27
User color

nanaya:
No more prepending it with '#'.

A bit worse than I hoped because some other places caching it =[


peppy:
Will merge coinciding with next lazer release (currently hard crashes).

PR #1376, opened 2017-07-27
Docker-based provisioning method

LiquidPL:
Got tired of Vagrant, so I made this.

Install docker and docker-compose, then run `docker-compose up` in the main directory.

If you have a non-standard uid/gid (i.e. not `1000`), run `docker-compose build` first, while supplying the ids as env variables, like this: `_UID=1001 _GID=1001 docker-compose build`.

Currently only tested on my own Linux machine.
EDIT: Also works on Windows 10 (with official Docker for Windows)


docker/nginx/nginx-osu-next
```diff
@@ -0,0 +1,52 @@
...
+        listen [::]:80 default_server ipv6only=on;
+
+        server_name next.ppy.sh beta.osu.hiroto.eu chi.ppy.sh localhost;
+
+        # error_log /vagrant/logs/nginx.log;
+
+        root /data/osu!web/public;
+        index index.php index.hh index.html;
+
+        # Cloudflare
```
notbakaneko: Cloudflare IPs shouldn't be applicable to the development environment?

docker/php/build.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
```
LiquidPL: Not sure really; I can't really merge this with the `build.sh` in the root dir, as it needed to be split between the composer and node containers.

README.md
```diff
@@ -26,6 +26,20 @@ There are a few different options to get started:
 - Create a fresh Ubuntu environment and run `sudo bootstrap.sh`. Note that this script is intended to be run on a *sandboxed environment*; do not run it on a shared development system without first understanding what it does.
 - Use the php built-in server (run `php -S 127.0.0.1:8080` from the `public` folder). You will still need a database backend.
 - Use vagrant (`cd vagrant; vagrant up`). Note that this is not actively supported or tested, and it is known to be quite slow on some platforms.
+- Use Docker:
+  - First, [install](https://docs.docker.com/compose/install/) Docker and Docker Compose, then run `docker-compose up` in the main directory.
```
nanaya: link should be in Docker and Docker Compose, to their own install page

README.md
```diff
@@ -26,6 +26,20 @@ There are a few different options to get started:
...
+  - Due to the nature of Docker (a container is killed when the command running in it finishes), the Yarn container will be run in watch mode.
+  - Note that if you use a non-standard user/group id (for example when your user wasn't the first one created on the system), you need to run the `docker-compose build` command first, supplying your ids as env variables:
+
+  ```bash
+  _UID=1001 _GID=1001 docker-compose build
```
nanaya: or just `_UID=$(id -u) _GID=$(id -g)`. Also `up` runs `build` if the image doesn't already exist anyway. Running `up` with correct env results in same thing.

README.md
```diff
@@ -26,6 +26,20 @@ There are a few different options to get started:
...
+  ```
+  This shouldn't be necessary during any further runs, unless you decide to rebuild your containers.
+
+  - Because of some bad database migrations, this command will currently fail. They are completely safe to ignore, as they nullify each other, so just simply delete those two files locally:
```
nanaya: update the migration instead (`MIN(contest_id) as contest_id`). And update the other one to just `(new OldMigration)->down()` and (`up()`).

docker/php/setup.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
nanaya: this file should be shared between all services

docker/php/setup.sh
```diff
@@ -0,0 +1,7 @@
...
+
+groupadd -g $_gid osuweb
+useradd -u $_uid -g $_gid osuweb
+
+mkdir /home/osuweb
```
nanaya: `useradd -m`

docker/php/build.sh
nanaya: fake the yarn command? (`echo true > /usr/local/bin/yarn && chmod +x /usr/local/bin/yarn`)

docker-compose.yml
```diff
@@ -0,0 +1,63 @@
+version: '3'
+
+services:
+    php:
```
nanaya: two spaces indentation for yaml files

PR #1383, opened 2017-07-29
Replace direct calls to Lang.get with osu.trans

cl8n:
This was causing some issues with locales that aren't fully translated. Some examples currently are multiplayer matches in Polish, and the supporter store page in Italian


notbakaneko:
We're waiting an upstream library to be updated with a fix, fyi

cl8n:
oh alright, that's good. this should still be fixed for now though, since it's broken as-is

PR #1387, opened 2017-08-01
Custom error message (and some cleanups)

nanaya:
Something like this? ┐(･_･)┌


peppy:
I dunno, I still feel like this is over-engineering for what it is. The method used here means there's no way to easily show the "current builds" header, right? I'd check if that's required, because I think if @arflyte was to actually make this page look half-decent the build header would have a place in the design (so users can actually get to useful places instantly, rather than requiring a second click to go "back to index" first)

nanaya:
Will be revisited after new changelog merged in.

PR #1389, opened 2017-08-01
Discussion logs

nanaya:
No filter yet because I'm not sure how useful it's going to be and how the filter should even work. Or whether it's even needed. Also all logs are visible to everyone except the usernames which are only visible to admin.

Also the logging part looks terrible =[
And missing ip address. And I wonder if it should be part of `Log` model instead of its own model. But then again lookup will be difficult because it'll be searching within... serialized php array. Or something.

(I kind of ran out of idea)


resources/assets/less/bem/beatmapset-event.less
```diff
@@ -0,0 +1,57 @@
...
+    width: 8px;
+    align-self: stretch;
+
+    .type(@name, @bg-color) {
+      &--@{name} {
+        background-color: @bg-color;
+      }
+    }
+
+    .type(approve, @blue);
```
nanaya: the colors are assigned by yours truly.

nanaya:
Also "beatmap updated by user" event is missing.

PR #1398, opened 2017-08-02
Only allow resolve state change for issues

nanaya:
┐(･_･)┌


nekodex:
Maybe also add a test in the opposite direction to ensure resolving a praise fails? ┐(･_･)┌

PR #1403, opened 2017-08-04
Misc discussion fixes

nekodex:
Two little things:
- Fixes 'Last edited by <original poster>' always being shown after a deleted post is restored (even if it hadn't been edited)
- Hide the edit post interface when hitting 'save' with no changes (instead of just doing nothing)

### DIFF
--- a/resources/assets/coffee/react/beatmap-discussions/post.coffee
+++ b/resources/assets/coffee/react/beatmap-discussions/post.coffee
@@ -125,7 +125,9 @@ class BeatmapDiscussions.Post extends React.PureComponent
 
 
   updatePost: =>
-    return if @state.message == @props.post.message
+    if @state.message == @props.post.message
+      @setState editing: false
+      return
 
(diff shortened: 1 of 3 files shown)


resources/assets/coffee/react/beatmap-discussions/post.coffee
```diff
@@ -187,7 +189,7 @@ class BeatmapDiscussions.Post extends React.PureComponent
                   classNames: ["#{bn}__info-user"]
                 delete_time: osu.timeago @props.post.deleted_at
 
-        if @props.post.updated_at != @props.post.created_at && @props.post.updated_at != @props.post.deleted_at
+        if @props.post.updated_at != @props.post.created_at && @props.post.updated_at != @props.post.deleted_at && @props.lastEditor?.id
```
nanaya: `lastEdited.id` may actually be null if the editor account has been deleted :(

PR #1417, opened 2017-08-10
ValidationErrors->isAny() seems to be flipped

notbakaneko:
?_?


nanaya:
good point ?_?

Also maybe just return the inverse of the other one.

notbakaneko:
but the extra function call :trollface:

PR #1426, opened 2017-08-14
Move featured artist name to top when width gets narrow

notbakaneko:
Discussed with @arflyte to have a quick fix for this problem:

image

and move the name above the image.
There's still the potential issue of long names overflowing behind the image.


nanaya:
apparently it's originally already doing that (`margin-top: 0`) but then broken by #1108 (`justify-content: center`).

peppy:
What's the outcome of this?

notbakaneko:
just going to override `justify-content` and centre it

PR #1432, opened 2017-08-15
Monkey patch Lang.get

notbakaneko:
upstream fixes haven't been released yet, either 🚒


nanaya:
a while ago I wrote this

```coffeescript
Lang.origGet = Lang.get
Lang.get = (key, replacements, locale) ->
  try
    Lang.origGet key, replacements, locale
  catch
    Lang.setLocale fallbackLocale
    message = Lang.origGet key, replacements, locale
    Lang.setLocale currentLocale

    message
```

notbakaneko:
passing in `locale` in the current version of lang.js is broken when it has to do a fallback :D

nanaya:
yeah the thing above "fixes" it.

notbakaneko:
We should probably monkey patch all the calls then ᕕ( ᐛ )ᕗ

resources/assets/coffee/lang-js.coffee
```diff
@@ -0,0 +1,29 @@
...
+#    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#    See the GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+
+# Lang.js monkey patch
+Lang.origGet = Lang.get
+Lang.get = (key, replacements, locale) ->
```
nanaya: Looking again the arguments can probably be removed and call the original method with `.apply null, arguments` instead.

The workaround in `osu.trans` can also be removed.

notbakaneko:
Should probably remove the calls to `osu.trans` too, in a separate pull request

resources/assets/coffee/lang-js.coffee
```diff
@@ -0,0 +1,29 @@
...
+###
+
+# Lang.js monkey patch
+Lang.origGet = Lang.get
+Lang.get = ->
+  try
+    Lang.origGet.apply @, arguments
+  catch
+    Lang.setLocale fallbackLocale
+    message = Lang.origGet.apply @, arguments
```
nanaya: apparently this should pass `fallbackLocale` (or nothing at all) instead :D

resources/assets/coffee/osu_common.coffee
```diff
@@ -242,11 +235,6 @@
   transChoice: (key, count, replacements) ->
     message = Lang.choice key, count, replacements, currentLocale
```
nanaya: can be _simplified_.

notbakaneko: That's the original, there's no point changing it since we can just get rid of it

peppy:
good to merge this for now?

PR #1436, opened 2017-08-17
User profile optimisation (stage 1)

nekodex:
First phase reduces the amount of (unused) beatmapset json being sent initially to speed up downloading and parsing. (should reduce page size by ~50-60%)

In addition, caching has been added for user follower counts to ease the load from more popular players.

Also, the max displayed beatmaps in each section has been temporarily reduced to 50 for the time being.

More optimisations to come...

### DIFF
--- a/app/Jobs/UpdateUserFollowerCountCache.php
+++ b/app/Jobs/UpdateUserFollowerCountCache.php
@@ -21,8 +21,6 @@
 namespace App\Jobs;
 
 use App\Models\User;
-use Cache;
-use Carbon\Carbon;
 use Illuminate\Bus\Queueable;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Queue\InteractsWithQueue;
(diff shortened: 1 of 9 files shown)


app/Jobs/UpdateUserFollowerCountCache.php
```diff
@@ -0,0 +1,61 @@
...
+     */
+    public function handle()
+    {
+        $key = User::CACHING['follower_count']['key'];
+        $duration = User::CACHING['follower_count']['duration'];
+
+        Cache::put(
+            "{$key}:{$this->user_id}",
+            User::find($this->user_id)->uncachedFollowerCount(),
+            Carbon::now()->addDay($duration)
```
nanaya: is it addDay or addHour? Also may want to merge both as one function?

app/Models/User.php
```diff
@@ -802,6 +810,25 @@ public function friends()
...
     }
 
+    public function uncachedFollowerCount()
+    {
+        return UserRelation::where('zebra_id', $this->user_id)->where('friend', 1)->count();
+    }
+
+    public function followerCount()
+    {
+        $key = self::CACHING['follower_count']['key'];
```
nanaya: move the one in jobs to function here and use it like this?
```
return cache::get() ?? this->cachecount
```

app/Models/User.php
```diff
@@ -802,6 +810,30 @@ public function friends()
...
+            $count,
+            Carbon::now()->addHours($duration)
+        );
+
+        return $count;
+    }
+
+    public function followerCount()
+    {
+        return Cache::get(self::CACHING['follower_count']['key'].":{$this->user_id}") ?? $this->cacheFollowerCount();
```
nanaya: PICK ONE SIR

nekodex: DIVERSITY SIR

PR #1437, opened 2017-08-18
Add redirect for /g/ to /home/groups/

nekodex:


nanaya:
should start moving all closures to controller so the routes can be cached

https://laravel.com/docs/5.4/controllers#route-caching

PR #1445, opened 2017-08-18
Faster user rank

nanaya:
Forgot to make pull request of this ಠ_ಠ

Added slave reading thingy while at it. Not sure if actually works. Setting `DB_HOST_READONLY` will set it otherwise it'll connect to default. Or I think that's what it's supposed to do and nothing else ┐(ﾟ ヮﾟ)┌

### DIFF
--- a/app/Models/Score/Best/Model.php
+++ b/app/Models/Score/Best/Model.php
@@ -145,7 +145,8 @@ abstract class Model extends BaseModel
     {
         $alwaysAccurate = false;
 
-        $query = static::where('beatmap_id', '=', $this->beatmap_id)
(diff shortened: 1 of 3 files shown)


nanaya:
There's also option to set different write and read host for each connection though probably not that good of idea.

app/Models/Score/Best/Model.php
```diff
@@ -156,13 +158,22 @@ public function userRank($options)
 
         if (isset($options['type'])) {
             $query->withType($options['type'], ['user' => $this->user]);
+
+            if ($options['type'] === 'country') {
```
peppy: any reason this is specifically for `country` type?

PR #1447, opened 2017-08-21
Remove closures and clean up some routes

### LINKED ISSUE #1438: Use route-caching
Requires removing the closures that are being used

https://laravel.com/docs/5.4/controllers#route-caching

cl8n:
Route caching works now that all closures are gone (fixes #1438)

The redirect thing is a bit of a hacky solution, but I think it's actually quite nice since it leaves the target routes easily readable in the routing file


routes/web.php
```diff
@@ -43,9 +43,7 @@
 
         Route::resource('orders.items', 'OrderItemsController', ['only' => ['update']]);
 
-        Route::get('/', function () {
-            return ujs_redirect(route('admin.store.orders.index'));
-        });
+        Route::get('/', '\App\Http\Controllers\RedirectController')->name('redirect.admin.store.orders.index');
```
nanaya: might as well make a function which does this

```php
function route_redirect($path, $target)
{
    // using `:` because not used anywhere else for route name
    return Route::get($path, '\App\Http\Controllers\RedirectController')->name("redirect:{$target}");
}
```

routes/web.php
```diff
@@ -165,42 +161,25 @@
...
-Route::get('users/register', function () {
-    return Redirect::to('https://osu.ppy.sh/p/register');
-})->name('users.register');
 Route::get('users/{id}/card', 'UsersController@card')->name('users.card');
 Route::resource('users', 'UsersController', ['only' => ['show']]);
 
 Route::group(['prefix' => 'help'], function () {
     // help section
-    Route::get('wiki/{page}', 'WikiController@show')->name('wiki.show')->where('page', '.+');
+    Route::get('wiki/{page?}', 'WikiController@show')->name('wiki.show');
```
nanaya: the `->where` can't be removed.

app/helpers.php
```diff
@@ -337,10 +337,7 @@ function timeago($date)
 
 function current_action()
 {
-    $currentAction = \Route::currentRouteAction();
-    if ($currentAction !== null) {
-        return explode('@', $currentAction, 2)[1];
-    }
+    return explode('@', \Route::currentRouteAction(), 2)[1] ?? null;
```
nanaya: the `\` isn't actually needed here.

app/Http/Controllers/RedirectController.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+namespace App\Http\Controllers;
+
+class RedirectController extends Controller
+{
+    public function __invoke(...$args)
+    {
+        // Redirect routes should be named 'redirect.<target>'
+        return ujs_redirect(route(explode('redirect.', \Route::currentRouteName(), 2)[1], ...$args));
```
nanaya: this should just be `$args`. Or remove the `...$args` above and use `func_get_args()` instead.

routes/web.php
cl8n: why is that? the case where `page` is empty is handled

routes/web.php
cl8n: will do, but colons (and hyphens, from what I tested) get internally converted to dots anyway so I'll keep the current prefix

routes/web.php
nanaya: to handle `Some/Page/Name/With/Slashes`

routes/web.php
nanaya: how did you test it?

```
Route::get('test', 'HomeController@index')->name('some:test-route.with^characters');

| | GET|HEAD | test | some:test-route.with^characters | App\Http\Controllers\HomeController@index | |
```

cl8n: `Route::currentRouteName()` returns a version where all punctuation are dots; the action of `RedirectController` broke with the colon because `explode()` didn't do anything

nanaya: weird, I remember having to change the explode pattern because I changed the name to colon yesterday...

routes/web.php
cl8n: woops, didn't think to test that case

I probably should have asked about that *before* deleting it :^)

routes/web.php
nanaya: it does indeed keep the colon here?

cl8n: seems that the only place it actually changes is with `currentRouteName()`. strange. is there really any benefit to using other punctuation though? these routes are basically hidden from the developer and won't ever be manually accessed

nanaya: The test I did above is with `currentRouteName`. Just to avoid weird care if there's another route with `redirect.` in its name.

cl8n: well... shoot, I know this isn't supposed to happen, but "on my machine" it's just always dots. might be an issue on my end, let me try it again real quick

cl8n: yeaaaaaah so I guess I just forgot to clear the route cache from when I first added it. I brought this upon myself, didn't I

fixed now :p

nanaya:
thanks!

PR #1458, opened 2017-08-25
Improve wiki search

nanaya:
- title field has highest priority
- (cleaned up) path field has higher priority (requires reindexing)
- limit search in current locale

Incoming complains from people using other locales can't search english wiki pages.

### DIFF
--- a/app/Models/Wiki/Page.php
+++ b/app/Models/Wiki/Page.php
@@ -62,36 +62,45 @@ class Page
             'from' => ($params['page'] - 1) * $params['limit'],
         ]);
 
-        if ($params['locale'] !== null) {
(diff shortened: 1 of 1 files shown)


peppy:
Hmm, may be a good idea to always search english as well (just give it lower priority maybe)

nanaya:
I looked at it but it's beyond me at the moment. Need to read more.

peppy:
💯

peppy:
well, :99: (let's fix the english fallback at some point)

notbakaneko:
I'd consider having a separate fields for each locale instead a `locale` field, then we can do a multi_match on the locale's field and english

PR #1462, opened 2017-08-25
Always include English pages for wiki search

nanaya:
Turns out it's pretty simple ☜(ﾟヮﾟ☜)

### DIFF
--- a/app/Models/Wiki/Page.php
+++ b/app/Models/Wiki/Page.php
@@ -61,31 +61,50 @@ class Page
             'size' => $params['limit'],
             'from' => ($params['page'] - 1) * $params['limit'],
         ]);
+        $searchParams['body']['query']['bool']['must'] = [];
 
-        $matchParams[] = [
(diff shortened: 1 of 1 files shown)


app/Models/Wiki/Page.php
```diff
@@ -61,31 +61,50 @@ public static function search($rawParams, $locale = null)
...
+        $searchParams['body']['query']['bool']['must'] = [];
+
+        $searchParams['body']['query']['bool']['must'][] = [
+            'bool' => [
+                'minimum_should_match' => 1,
+                'should' => [
+                    ['match' => [
+                        'locale' => [
+                            'query' => $params['locale'] ?? App::getLocale(),
+                            'boost' => 1000,
```
notbakaneko: 三三三ヽ( ﾟ ヮﾟ)ﾉ

nanaya: BOOST

PR #1475, opened 2017-08-30
Update to latest oauth library

nanaya:
No changelog found, hopefully just updating compatibility with backend oauth library ( ･_･)a


nanaya:
hold on updating one more version ಠ_ಠ

PR #1477, opened 2017-08-31
Add instrumentation to Elasticsearch queries

nekodex:


app/Models/Beatmapset.php
```diff
@@ -485,6 +486,7 @@ public static function searchDB(array $params = [])
 
     public static function search(array $params = [])
     {
+        $startTime = microtime(true);
```
nanaya: perhaps just update the beatmapset search to use `Search` as well? Otherwise this search will be recorded twice when run on global search page.

app/Libraries/Search.php
```diff
@@ -102,7 +103,14 @@ public function search($mode)
         $key = __FUNCTION__.':'.$mode;
 
         if (!array_key_exists($key, $this->cache)) {
+            $startTime = microtime(true);
+
             $this->cache[$key] = $class::search($this->params);
+
+            if (config('datadog-helper.enabled', false)) {
```
nanaya: Isn't the default already set in the config file?

app/Models/Beatmapset.php
nekodex: Yeah that looks non-trivial relative to the quick instrumentation I'm adding, feel free to refactor it later...

nanaya:
oh whoops.

PR #1478, opened 2017-08-31
Maybe fix user searching on empty string

notbakaneko:
reading `Search::urlParams()` triggers search which ends up calling `User::search()` with an empty string query

ᕕ( ᐛ )ᕗ


nanaya:
Or maybe just skip searching inside urlParams. The function is half-done at the moment anyway.

PR #1480, opened 2017-09-01
Fix missing parameter in search instrumentation

nekodex:
whoops


nanaya:
So I just learned that `timing` accepts microsecond while `microtiming` accepts second. 👍

PR #1492, opened 2017-09-05
Beatmapsets by score fix

nanaya:
More missed update to beatmapset_id column removal.


notbakaneko:
the migration isn't there ?_?

notbakaneko:
`Beatmapset::scopeWithHasCompleted` needs to be updated to remove the reference to scores `beatmapset_id`as well

PR #1508, opened 2017-09-06
Enable database persistent connections by default

peppy:


peppy:
doesn't look like travis can recover from this

PR #1542, opened 2017-09-13
fix negative item cost

notbakaneko:
Die on negative price; ignore submitted value.

### DIFF
--- a/app/Models/Store/Order.php
+++ b/app/Models/Store/Order.php
@@ -280,6 +280,10 @@ class Order extends Model
 
     private function newOrderItem(array $params)
     {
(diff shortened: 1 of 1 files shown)


app/Models/Store/Order.php
```diff
@@ -280,15 +280,25 @@ private function removeOrderItem(array $params)
 
     private function newOrderItem(array $params)
     {
+        if ($params['cost'] < 0) {
```
nanaya: shouldn't this after the override at username-change?

notbakaneko: nope

PR #1544, opened 2017-09-13
Combine user profile endpoints (api and web)

nekodex:


app/Http/Controllers/UsersController.php
```diff
@@ -218,21 +188,63 @@ public function show($id, $mode = null)
...
-            'jsonChunks'
-        ));
+        $rankHistoryData = $user->rankHistories()
+            ->where('mode', Beatmap::modeInt($currentMode))
+            ->first();
+
+        $rankHistory = $rankHistoryData ? json_item($rankHistoryData, 'RankHistory') : [];
+
+        if (Request::is('api/*')) {
+            $userArray['statistics'] = $statistics;
```
nanaya: what's with this subtle difference between api and non-api ಠ_ಠ

routes/web.php
```diff
@@ -236,13 +236,20 @@
         Route::resource('beatmapsets', '\App\Http\Controllers\BeatmapsetsController', ['only' => ['show']]);
 
         //  GET /api/v2/me
-        Route::get('me', 'UsersController@me');
+        Route::get('me', '\App\Http\Controllers\UsersController@me');
```
nanaya: at this rate might as well remove the `API` prefix from route group? <_<

app/Http/Controllers/UsersController.php
nekodex: Because the web side will change to that once the rest of the data is moved to async loading

routes/web.php
nekodex: soon™

app/Http/Controllers/UsersController.php
nanaya: change to... what?

nekodex: the lesser data in the api response.

PR #1546, opened 2017-09-13
Validate order item on saving

notbakaneko:
Why is model validation on save not a standard thing?  ಠ_ಠ

This also always forces `OrderItem` to always have a `cost` value when adding or updating the cart because right now, it doesn't.


notbakaneko:
should probably look at moving the `Order` validation too 🤔

PR #1562, opened 2017-09-21
Add August spotlight medal assets

Ephemeralis:


peppy:
Colours and alignment are slightly off, but merging to get this pushed out for now.

PR #1570, opened 2017-09-26
Refactor user email/password validation

nanaya:
Backend part of user registration. Mainly just refactor of user email/password change validation.


nekodex:
So... uh, where are the limits and other protections to prevent people just hammering the user create endpoint to mass make fake users?

PR #1594, opened 2017-10-10
Fix 'show more' links on user profiles not disappearing after loading the last batch of data

LiquidPL:
This happened specifically when the length of last fetched data was equal to the amount of elements loaded with each fetch.


nekodex:
The queries required to get those counts aren't really optimal and don't really work at scale (which is the main reason why I didn't do this initially)... probably a better idea to fetch n+1 results each time (while only displaying n results) and check that way

peppy:
@nekodex what was the consensus on this? i remember you looking into an alternative solution at some point

nekodex:
@peppy Read the comment directly above yours?

peppy:
Ok, just checking no work was done elsewhere to resolve this one since my memory was vague.

LiquidPL:
Reworked this to do the n+1 fetch for scores, as we seem to have accurate counts for everything else.

app/Transformers/UserTransformer.php
```diff
@@ -137,6 +139,20 @@ public function includeRecentActivities(User $user)
         );
     }
 
+    public function includeMostPlayedBeatmapsetCount(User $user)
```
nekodex: I don't think this is still used? Probably want to remove it because it's still being included and running the queries...

nanaya:
This probably a bit simpler (not quite tested) https://0paste.com/18191.txt

LiquidPL:
What commit did you base this patch on? I can't really seem to apply it to anything.

nanaya:
master.

app/Http/Controllers/UsersController.php
```diff
@@ -334,7 +331,7 @@ private function recentKudosu($user, $perPage = 5, $offset = 0)
         );
     }
 
-    private function mostPlayedBeatmapsets($user, $perPage = 5, $offset = 0)
+    private function mostPlayedBeatmapsets($user, $perPage = 6, $offset = 0)
```
LiquidPL: not sure if it would make sense for those to be `original_count + 1`, for some more code readability

nanaya:
Another one unifying per page setting and more per page graveyards (maybe). https://0paste.com/19224.txt

app/Http/Controllers/UsersController.php
```diff
@@ -250,20 +246,34 @@ public function show($id, $mode = null)
...
+                'recentlyReceivedKudosu' => 5,
+            ];
+
             $beatmapsets = [
-                'most_played' => $this->mostPlayedBeatmapsets($user),
-                'ranked_and_approved' => $this->rankedAndApprovedBeatmapsets($user),
-                'favourite' => $this->favouriteBeatmapsets($user),
-                'unranked' => $this->unrankedBeatmapsets($user),
-                'graveyard' => $this->graveyardBeatmapsets($user),
+                'most_played' => $this->mostPlayedBeatmapsets($user, $perPage['beatmapPlaycounts'] + 1),
```
peppy: probably worth adding a comment on why the `+ 1` is necessary, for future explorers of this code.

LiquidPL:
btw should I change pagination here to load 20 more like in #2043?

LiquidPL:
nvm

Issue #1593, opened 2017-10-10
Pippi's legs bleed through the online users graph in the home page.


Nochys:
![this is very distracting](https://puu.sh/xUDvl.jpg)
![this is very very distracting](https://puu.sh/xUDFK.jpg)

CrazyHatish:
I works fine on Chrome, but has this problem on Firefox. Both Pippi and the graph seem to be properly positioned in the div with `bottom: 0`. Only solution I found was to change Pippi's position to `bottom: 1px`.

CrazyHatish:
Setting `.landing-hero__pippi { line-height: 1; }` seems to work

PR #1600, opened 2017-10-12
Fix Pippi overflowing the landing-hero div

CrazyHatish:
Solves issue #1593.
The `line-height: 1.25` property from the osu-layout__row div (inherited from html) was interfering with the positioning of the Pippi image directly below on some browsers.
Changing the line-height to 1 solved the issue, but made the header slightly shorter. Another option would be reducing Pippi's height by a few px.


nanaya:
No this doesn't fix it. Also this changes so many other unrelated things.

PR #1613, opened 2017-10-18
Include final hitobjects in valid timestamp range

### LINKED ISSUE #1612: moddingv2 timeline does not support map's last objects
timeline posts including objects in the last second of a map result in "Failed saving post" error. (end of a map determined by final circle, slider start, or spinner start location)

for example, if a map's final object is at 01:30:353, the latest timestamp that can be posted is 01:29:999

visual example: https://www.youtube.com/watch?v=3SbChAiGiLc

cl8n:
closes #1612 

lame solution for now; just allows a timestamp at 1 second past the beatmap length to include any objects in that final second


peppy:
This has been fixed at a database level! The only change required is to adjust comparisons to be inclusive <=.

PR #1614, opened 2017-10-19
Fix wrong max-width for mobile.

nanaya:
Honestly I'd just nuke `@media @mobile` ಠ_ಠ

Mobile first stuff.

Though considering the original design is only desktop maybe should go the other way with desktop-first ?_?

### DIFF
--- a/resources/assets/less/variables.less
+++ b/resources/assets/less/variables.less
@@ -18,8 +18,9 @@
 
 @import "app";
 
+@screen-mobile-max: (@screen-sm-min - 1px);
 @desktop: ~"(min-width: @{screen-sm-min})";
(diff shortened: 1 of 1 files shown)


resources/assets/less/variables.less
```diff
@@ -18,8 +18,9 @@
 
 @import "app";
 
+@screen-mobile-max: @screen-sm-min - 1px;
 @desktop: ~"(min-width: @{screen-sm-min})";
-@mobile: ~"(max-width: @{screen-sm-min})";
+@mobile: ~"(max-width: @{screen-mobile-max})";
```
notbakaneko: `@media (max-width: 840px - 1px) {`
wat

need `(@screen-sm-min - 1px);` to get 839px

Issue #1609, opened 2017-10-17
moddingv2 outside opinion topic unresolving


Ephemeralis:
Currently, once a discussion point is marked as "resolved" either by a moderator or the map owner, the point cannot be re-raised for discussion by any means outside of posting a duplicate issue.

While the onus on issue resolution resides with the mapper, providing them with the explicit impetus to unilaterally declare what is and is not an issue does not exactly fit the spirit of modding, especially if moddingv2 is to become the primary platform for the discussion of such issues.

Allowing upvoted comments within a resolved thread to automatically unresolve the issue would be the most intuitive approach to solving this problem. Allowing beatmap nominators to unresolve issues a la the QAT would also work here to some degree.

nanaya:
The thread starter can already reopen it. Nominator can be added so they can unresolve issue if needed.

Not sure what you mean with "upvoted comment within a resolved thread" because there's no comment voting?

Ephemeralis:
Allowing BNG to resolve/unresolve seems an elegant enough solution here.

PR #1619, opened 2017-10-23
Move thread resolve priv check to backend

nanaya:
And allow BNG to resolve as well.

Related to #1609? ┐(' ヮ' )┌


app/Models/Model.php
```diff
@@ -59,4 +59,12 @@ public function saveOrExplode($options = [])
 
         return $result;
     }
+
+    public function save(array $options = [])
+    {
+        $result = parent::save($options);
+        priv_clear_cache();
```
nekodex: I feel like this shouldn't be on the base model? Should just call it as needed? A lot of unnecessary clearing is gonna happen... (also why is this even needed?)

nanaya: Changes to model may change authorization.

peppy: which models? shouldn't it only be done where it is necessary?

nanaya: there's currently only manual way to figure out whether it is necessary.

1. priv_check on a model
2. update the model
3. priv_check again for view

If `2` changes result of priv check, the cache must be cleared at one point before `3` is done. Placing it in controller can also work but it means either clearing the cache everywhere before rendering the view (incl. transformer) or checking whether priv check is used by the view, already cached, and affected by the just updated model.

nekodex:
Admins still seem to bypass the `$discussion->canBeResolved()` check, allowing them to 'resolve' praises and stuff still?

nanaya:
no

nekodex:
oh wait branch wasn't up to date, nvm derp

PR #1625, opened 2017-10-26
Bring tournament pages into 2017

nekodex:
Same (general) functionality, new paint job.

(as a bonus, adds support for tournament header banners, basic localisation and cleanups)

listing:

registration:


peppy:
👍

PR #1635, opened 2017-10-27
Rename inactive users on username change

notbakaneko:


app/Models/User.php
```diff
@@ -130,13 +131,33 @@ public function changeUsername($newUsername, $type = 'support')
...
+        $name = $try > 0 ? "{$newUsername}_{$try}" : $newUsername;
+
+        try {
+            return $this->updateUsername($name, $oldUsername, $type);
+        } catch (QueryException $ex) {
+            if (!is_sql_unique_exception($ex) || $try > 9) {
+                throw $ex;
+            }
+
+            return $this->tryUpdateUsername($try + 1, $newUsername, $oldUsername, $type);
```
nanaya: `retr`--- oh wait

notbakaneko: (ﾉ ・ω・)ﾉ ミ┻━┻

PR #1663, opened 2017-11-01
Change forum moderation icon colours so they stand out

nekodex:


nanaya:
Need correct activated color?

PR #1664, opened 2017-11-01
Beatmapset watches

nanaya:
Seems working ┐(･_･)┌


nekodex:
I think you forgot to add an extra header column on the watchlist/index page for the cover

nekodex:
Wording in the email is no longer consistent with the new terminology

PR #1665, opened 2017-11-01
Unchecked out carts shouldn't have invoices

notbakaneko:
because that would be silly :trollface:

### DIFF
--- a/app/Models/Store/Order.php
+++ b/app/Models/Store/Order.php
@@ -187,8 +187,12 @@ class Order extends Model
                 return 'Awaiting Payment';
             case 'incart':
                 return '';
-            default:
+            case 'paid':
+            case 'shipped':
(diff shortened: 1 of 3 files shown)


app/Models/Store/Order.php
```diff
@@ -178,6 +178,20 @@ private function getShipping()
...
+    public function getStatusText()
+    {
+        switch ($this->status) {
+            case 'cancelled':
+                return 'Cancelled';
+            case 'checkout':
+                return 'Awaiting Payment';
+            case 'incart':
+                return '';
+            default:
```
notbakaneko: default as `Paid` seems silly, though....

nekodex: yeah nah, add an unknown/invalid status if anything?

PR #1668, opened 2017-11-02
Catch exceptions when sending notifications.

notbakaneko:
Errors caused by notification send failures causing weird things to happen
:|

Notification connection failures should have caused IPN failures, but for some reason Paypal picked it up as successful and didn't retry ?_?


peppy:
Did you figure a way to limit the slack timeout?

notbakaneko:
Superseded by #1679

PR #1679, opened 2017-11-06
Make Store notifications actually queueable.

notbakaneko:
Makes store notification serializable.
Try not to dump entire nested contents of `Order` object into queue.
Set a timeout when sending to Slack.
Also fix a wrong `addError` method call. 

The queue worker should be called with something like `--queue=store-notifications --delay=60 --tries=10` to prevent it from spamming retries forever.

### DIFF
--- a/config/store.php
+++ b/config/store.php
@@ -4,4 +4,7 @@ return [
     'order' => [
         'prefix' => presence(env('STORE_ORDER_PREFIX'), 'store'),
     ],
+    'queue' => [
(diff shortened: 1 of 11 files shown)


config/store.php
```diff
@@ -4,4 +4,7 @@
     'order' => [
         'prefix' => presence(env('STORE_ORDER_PREFIX'), 'store'),
     ],
+    'queue' => [
+        'notifications' => presence(env('STORE_NOTIFICATIONS_QUEUE'), 'store-notifications'),
```
nanaya: add to `.example`?

nanaya:
Need to remember actually setting the parameters.

PR #1683, opened 2017-11-06
Link support osu! to new store page

### LINKED ISSUE #1634: [/home/support] Support osu! links are still redirecting to the old page

notbakaneko:
fixes #1634 

`STORE_SUPPORTER_TAG_ID=` needs to be set in `.env` since it's variable.


.env.example
```diff
@@ -101,6 +101,7 @@ PUSHER_SECRET=
 PAYMENT_SANDBOX=true
 
 STORE_NOTIFICATION_CHANNEL=test
+STORE_SUPPORTER_TAG_ID=10 # Your id will vary.
```
nanaya: `_PRODUCT_ID`?

notbakaneko:
I'm thinking it may be worth having matching routes for products with a `custom_class`?

nanaya:
probably

routes/web.php
```diff
@@ -212,6 +211,12 @@
             'CheckoutController',
             ['only' => ['index', 'store']]
         );
+
+        Route::resource(
+            'product',
```
nanaya: plural?

notbakaneko: the current route is singular :(

nanaya: replace it? :D

app/helpers.php
```diff
@@ -985,3 +985,8 @@ function group_users_by_online_state($users)
         'offline' => $offline,
     ];
 }
+
+function product_url($product)
+{
+    return route('store.product.show', ['product' => $product]);
```
nanaya: `route('store.products.show', $products)` _should_ work so this helper probably isn't needed.

nanaya:
may want to search for `/store/product/`

PR #1687, opened 2017-11-07
Payment processor tweaks

notbakaneko:


app/Libraries/Payments/CentiliPaymentProcessor.php
```diff
@@ -63,6 +63,11 @@ public function getNotificationType()
         return $mapping[$this['status']] ?? "unknown__{$this['status']}";
```
nanaya: maybe update this to use the `TypeRaw`?

PR #1721, opened 2017-11-14
Forum mobile edit box size

### LINKED ISSUE #1718: Forum post editor is claustophobically small
Should be resizeable. Currently isn't (tested on safari)

nanaya:
Nuke out zoom button and make the default as large as desktop's. Fixes #1718.

Also resizable textarea in ios does not exist natively.


peppy:
This doesn't resolve not being able to make the edit window bigger on desktop. Working with long replies is nigh impossible to the point I have to use an external editor to do so.

If resizing textboxes can't be done, maybe we should have a "toggle fullscreen edit" mode or something"

nanaya:
There's that resize bar handle on desktop...?

nanaya:
Oh safari doesn't have one :D

nanaya:
Just remembered it was disabled because it doesn't quite work with the floating box.

PR #1730, opened 2017-11-15
Include the one week delay for ranking maps

nanaya:
Not sure if the logic is correct ┐(°～° )┌

(also not sure why there's a `startOfDay`)

### DIFF
--- a/app/Models/Beatmapset.php
+++ b/app/Models/Beatmapset.php
@@ -83,7 +83,7 @@ class Beatmapset extends Model
     ];
 
     const NOMINATIONS_PER_DAY = 3;
-    const QUALIFICATIONS_PER_DAY = 6;
+    const RANKED_PER_DAY = 6;
     const MINIMUM_DAYS_FOR_RANKING = 7;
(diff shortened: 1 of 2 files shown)


app/Models/Beatmapset.php
```diff
@@ -84,6 +84,7 @@ class Beatmapset extends Model
 
     const NOMINATIONS_PER_DAY = 3;
     const QUALIFICATIONS_PER_DAY = 6;
```
Xinnoh: 8 maps are moved from qualified -> ranked per day

nekodex:
I used `startOfDay` because it looked weird to have an "estimate" show a specific time. Maybe just get rid of timeago and replace with just the estimated date to be ranked?

app/Models/Beatmapset.php
nekodex: Also these variables are named... wrong (my bad), `NOMINATIONS_PER_DAY` should actually be `QUALIFICATIONS_PER_DAY` and `QUALIFICATIONS_PER_DAY` should be `RANKED_PER_DAY`

PR #1732, opened 2017-11-16
Only count non-converted beatmaps

notbakaneko:


resources/assets/coffee/react/_components/playmode-tabs.coffee
```diff
@@ -40,7 +40,10 @@ class @PlaymodeTabs extends React.Component
             'data-disabled': disabled
             osu.trans "beatmaps.mode.#{mode}"
             if @props.showCounts
-              count = _.size(@props.beatmaps[mode])
+              filtered = _.filter @props.beatmaps[mode], (beatmap) ->
```
nanaya: you mean `_.reduce`

notbakaneko: how about `no`

nanaya: how about `sumBy`

notbakaneko: `yes`

PR #1755, opened 2017-11-17
Apparently Paypal doesn't include the invoice_id sometimes...

notbakaneko:
but sandbox has it ಠ_ಠ

### DIFF
--- a/app/Models/Store/Order.php
+++ b/app/Models/Store/Order.php
@@ -82,7 +82,7 @@ class Order extends Model
         ]);
     }
 
-    public static function scopeWherePaymentTransactionId($query, $transactionId)
+    public static function scopeWherePaymentTransactionId($query, $transactionId, $provider)
     {
         $ordersTable = (new static)->getTable();
         $paymentsTable = (new Payment)->getTable();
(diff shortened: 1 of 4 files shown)


app/Models/Store/Order.php
```diff
@@ -82,6 +82,17 @@ public function scopeWhereOrderNumber($query, $orderNumber)
         ]);
     }
 
+    public static function scopeWherePaymentTransactionId($query, $transactionId)
+    {
+        $ordersTable = (new static)->getTable();
+        $paymentsTable = (new Payment)->getTable();
+
+        return $query
+            ->join($paymentsTable, "{$paymentsTable}.order_id", '=', "{$ordersTable}.order_id")
```
nanaya: `whereIn('order_id', Payment::select(...)->where()...)` for less table name hacks?

app/Models/Store/Order.php
```diff
@@ -82,6 +82,18 @@ public function scopeWhereOrderNumber($query, $orderNumber)
         ]);
     }
 
+    public static function scopeWherePaymentTransactionId($query, $transactionId, $provider)
+    {
+        $ordersTable = (new static)->getTable();
```
nanaya: why is this thing still here (ノﾟ ヮﾟ)ノ︵ ┻━┻

notbakaneko: (ノﾟ ヮﾟ)ノ︵ ┻━┻

PR #1760, opened 2017-11-17
Update coffeescript to reflect updated permissions

nekodex:
Should have been part of #1758


resources/assets/coffee/react/beatmap-discussions/discussion.coffee
```diff
@@ -166,8 +168,8 @@ class BeatmapDiscussions.Discussion extends React.PureComponent
       user: @props.users[post.user_id]
       lastEditor: @props.users[post.last_editor_id]
       canBeEdited: @props.currentUser.isAdmin || @isOwner(post)
```
nanaya: seems to need update as well

nekodex: whoops, they shouldn't be able to edit posts

PR #1766, opened 2017-11-17
Fix buttons in ie

nanaya:
Less icon crushing by preventing shrink.


peppy:
wonder if this will fix mobilesafari too, same issue I believe

PR #1775, opened 2017-11-19
Fix user lookups for numeric usernames

### LINKED ISSUE #1605: The /users/ route behave differently from the /u/ route
Found by @lslqtz 
https://osu.ppy.sh/u/2420195176 returns the profile of user whose username is 2420195176 .
https://osu.ppy.sh/users/2420195176 returns 404 .
Is this expected behavior? Or is this a bug?

cl8n:
closes #1605 
fixes 404s when looking up numeric usernames (ex. "630" - [old site](https://osu.ppy.sh/u/630), [new site](https://osu.ppy.sh/users/630))

not sure if this is the best way to go about it

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -363,13 +363,10 @@ class User extends Model implements AuthenticatableContract, Messageable
 
             default:
                 if (is_numeric($username_or_id)) {
-                    $user = self::lookup($username_or_id, 'id', $find_all);
-                    if ($user !== null) {
-                        return $user;
-                    }
+                    $user = static::lookup($username_or_id, 'id', $find_all);
                 }
-                $user = self::where('username', $username_or_id)->orWhere('username_clean', '=', $username_or_id);
-                break;
+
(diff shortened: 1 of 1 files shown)


app/Models/User.php
```diff
@@ -363,10 +363,12 @@ public static function lookup($username_or_id, $lookup_type = null, $find_all =
...
                 if (is_numeric($username_or_id)) {
-                    $user = self::where('user_id', $username_or_id);
-                } else {
-                    $user = self::where('username', $username_or_id)->orWhere('username_clean', '=', $username_or_id);
+                    $user = self::lookup($username_or_id, 'id', $find_all);
+                    if ($user !== null) {
+                        return $user;
+                    }
                 }
+                $user = self::where('username', $username_or_id)->orWhere('username_clean', '=', $username_or_id);
```
nanaya:
```
default:
  if (is_numeric) {
    $user = ...;
  }
  return $user ?? static::lookup(...);
```

cl8n: aha I knew there had to be a better way. thanks

PR #1776, opened 2017-11-20
Ability to remove userpages

cl8n:
... by soft-deleting the post and breaking the relationship to it

this way, users can actually get rid of the "me!" tab instead of just reducing it to a tiny bar

### DIFF
--- a/app/Http/Controllers/AccountController.php
+++ b/app/Http/Controllers/AccountController.php
@@ -176,6 +176,7 @@ class AccountController extends Controller
 
         if (Request::isMethod('delete')) {
             $user->destroyPage();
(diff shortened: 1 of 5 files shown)


cl8n:
there's a small issue with styling where the icons won't stay visible when hovering over the right half of their box. I'm not sure how to fix ;_;

nanaya:
Looks like this will prevent people with previous supporter tag (but not anymore) from creating user page after deleting it.

nanaya:
There's already code to hide the userpage when it's empty but it needs updating.

cl8n:
added in #1777

PR #1779, opened 2017-11-20
Fix user lookups for numeric usernames II: PHP Strikes Back

cl8n:


app/Models/User.php
```diff
@@ -362,7 +362,7 @@ public static function lookup($username_or_id, $lookup_type = null, $find_all =
                 break;
 
             default:
-                if (is_numeric($username_or_id)) {
+                if (ctype_digit($username_or_id)) {
```
nanaya: `ctype_digit(1)` is false

cl8n: it's a string though

nanaya: but is it (it's called in quite a few places)

cl8n: yeah, every call (with default lookup) is a string. should I cast it just to be safe?

nanaya: yes, before someone gets confused why `lookup(48)` works but returns wrong user.

Issue #1773, opened 2017-11-19
Kudosu history can be spammed by clicking the 👍 button a lot


Xinnoh:
![screen shot 2017-11-19 at 1 49 20 pm](https://user-images.githubusercontent.com/30382015/32987819-95ddb59a-cd30-11e7-996f-14f50a858cf1.png)

PR #1780, opened 2017-11-20
Limit voting per discussion

nanaya:
Three per hour. I think.

### DIFF
--- a/app/Models/BeatmapDiscussion.php
+++ b/app/Models/BeatmapDiscussion.php
@@ -20,6 +20,7 @@
 
 namespace App\Models;
 
+use Cache;
 use Carbon\Carbon;
 use DB;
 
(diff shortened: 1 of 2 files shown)


app/Models/BeatmapDiscussion.php
```diff
@@ -283,6 +285,19 @@ public function isDeleted()
         return $this->deleted_at !== null;
     }
 
+    public function userRecentVotesCount($user, $increment = false)
+    {
+        $key = "beatmapDiscussion:{$this->getKey()}:votes:{$user->getKey()}";
+
+        if ($increment) {
+            Cache::add($key, 0, 60);
```
nanaya: probably better use `RateLimiter` or something smarter.

peppy:
Closes https://github.com/ppy/osu-web/issues/1773.

PR #1787, opened 2017-11-21
Special post type for mapper

### LINKED ISSUE #1739: Mappers should not be able to receive kudosu from their own map
https://puu.sh/ynfgv/634064bfe4.png
If a mapper creates a discussion on their set, others can upvote it and the mapper gets kudosu. Map owners can't receive kudosu in mv1, and shouldn't be the case for mv2 either.

nanaya:
Something like this? ┐(-｡ｰ;)┌

Can use input on icon and color and everything else.

Fixes #1739.

### DIFF
--- a/app/Models/BeatmapDiscussion.php
+++ b/app/Models/BeatmapDiscussion.php
@@ -40,7 +40,7 @@ class BeatmapDiscussion extends Model
         'praise' => 0,
         'suggestion' => 1,
         'problem' => 2,
-        'mapper' => 3,
+        'mapper_note' => 3,
     ];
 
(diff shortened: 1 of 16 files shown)


nekodex:
Is there a reason to use 'mapper' for all the variables/etc when all the visible displays are 'notes'?

nanaya:
because it's initially called question and need to be renamed. Then I figured it's easier to read when it explicitly says mapper (in `hasValidMessageType` for example).

nekodex:
Probably also want to add a data migration to convert existing issues/etc by map creators into mapper notes

Issue #1767, opened 2017-11-17
Top Navigation Header opens too easily


dorothy3242:
I've been running into this a _ton_ where any time I switch tabs on my browser and try to move my mouse to click something near the top of a page, I can't because the navigation hover menu opened from my quickly moving over its area. You can simply move the mouse down more and up again... but it gets irritating to have to regularly do this after a while. 
The menu is getting in the way a lot of the time.
This is especially punishing when you try to move your cursor to the top to change mode tabs and accidentally go too high, opening it again, and requiring you to move off of the menu again to get it to close. 

Video of condensed footage of frustrations:
https://youtu.be/IALcAS4EkL0

nekodex:
Resolved by #1795. Please reopen this issue if you find the fix insufficient.

PR #1795, opened 2017-11-22
Add slight delay for showing header menu

nanaya:
Need a bit more testing 💃 

Hopefully good enough to solve #1767.

### DIFF
--- a/resources/assets/coffee/_classes/nav.coffee
+++ b/resources/assets/coffee/_classes/nav.coffee
@@ -52,6 +52,11 @@ class @Nav
     (force || !document.body.classList.contains('modal-open')) && @popup[0]?
 
 
+  clearTimeouts: =>
+    Timeout.clear @hideTimeout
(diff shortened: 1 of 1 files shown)


resources/assets/coffee/_classes/nav.coffee
```diff
@@ -64,6 +64,21 @@ class @Nav
...
+
+    Timeout.clear @hideTimeout
+    Timeout.clear @showTimeout
+    @hideTimeout = Timeout.set 250, @hidePopup
+
+
+  delayedShowPopup: =>
+    Timeout.clear @hideTimeout
+    Timeout.clear @showTimeout
+    @showTimeout = Timeout.set 250, @showPopup
```
nanaya: Looks like 100 is a bit too low to allow passing header vertically without showing the popup.

nekodex:
Can probably pull those Timeout.clear pairs out into another function seeing how many times they're used together

PR #1797, opened 2017-11-22
Fix wrong field being updated on username change

notbakaneko:
there_was_an_attempt.jpg


app/Models/User.php
```diff
@@ -179,11 +179,10 @@ private function updateUsername($newUsername, $oldUsername, $type)
             Forum\Forum::where('forum_last_poster_id', $this->user_id)->update(['forum_last_poster_name' => $newUsername]);
             // DB::table('phpbb_moderator_cache')->where('user_id', $this->user_id)->update(['username' => $newUsername]);
             Forum\Post::where('poster_id', $this->user_id)->update(['post_username' => $newUsername]);
+            Forum\Topic::where('topic_poster', $this->user_id)
+                ->update(['topic_first_poster_name' => $newUsername]);
             Forum\Topic::where('topic_last_poster_id', $this->user_id)
```
nanaya: I think you need to look up a bit at line 179 =D

PR #1799, opened 2017-11-22
Add command to refresh topic cache to fix username change updating wrong field

notbakaneko:
turns out that getting the ids and then using `array_chunk` and selecting just the chunks is much faster than using Eloquent's `::chunk`


app/Console/Commands/FixUsernameChangeTopicCache.php
```diff
@@ -0,0 +1,130 @@
...
+        // chunk() freezes every 1000 records as it queries and offsets.
+        $chunks = array_chunk($array, 1000);
+        foreach ($chunks as $chunk) {
+            $this->processChunk($chunk, $bar);
+        }
+    }
+
+    private function processChunk($chunk, $bar)
+    {
+        $topics = Topic::withTrashed()->whereIn('topic_id', $chunk)->get();
```
nanaya: `with('users')`?

PR #1816, opened 2017-11-26
[zh] Chinese patch-9.2

kj415j45:
Update for Modding v2

And minor updates


resources/lang/zh/beatmaps.php
```diff
@@ -34,10 +34,11 @@
     'discussions' => [
         'allow_kudosu' => '给予 kudosu',
         'delete' => '删除',
-        'deleted' => '被 :editor 于 :delete_time 删除',
+        'deleted' => '被 :editor 于 :delete_time 删除。',
         'deny_kudosu' => '收回 kudosu',
         'edit' => '编辑',
-        'edited' => '最后由 :editor 编辑于 :update_time',
+        'edited' => '最后由 :editor 编辑于 :update_time 。',
```
peppy: extra whitespaces intentional?

kj415j45: ehhh, i check some context, remove the full stop(`。`) would be better.

will fix it later today

PR #1820, opened 2017-11-27
Add reply and resolve button to beatmapset discussion

### LINKED ISSUE #1746: Resolved checkbox is easy to overlook when replying to discussions
As mentioned in https://twitter.com/ascendanceosu/status/930770066706231296, it looks like users are regularly posting a comment and forgetting to hit the resolved checkbox.

I think we should remove the checkbox and instead of two buttons, something like "Reply" and "Reply & Resolve".

notbakaneko:
fixes #1746
It looks odd to have 2 buttons both with the reply arrow icon, though...

Left the checkbox in, since it still has an effect when using enter to submit.

### DIFF
--- a/resources/assets/coffee/react/beatmap-discussions/new-reply.coffee
+++ b/resources/assets/coffee/react/beatmap-discussions/new-reply.coffee
@@ -194,7 +194,7 @@ class BeatmapDiscussions.NewReply extends React.PureComponent
     return if e.keyCode != 13
 
     e.preventDefault()
-    @throttledPost()
+    @throttledPost(e)
 
 
   toggleResolveDiscussion: (e) =>
--- a/resources/assets/coffee/react/beatmap-discussions/new-reply.coffee
+++ b/resources/assets/coffee/react/beatmap-discussions/new-reply.coffee
@@ -75,7 +75,7 @@ class BeatmapDiscussions.NewReply extends React.PureComponent
         className: "#{bn}__footer"
         div className: "#{bn}__actions",
           div className: "#{bn}__actions-group",
-            if @props.discussion.can_be_resolved && @props.discussion.current_user_attributes.can_resolve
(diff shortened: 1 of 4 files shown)


nanaya:
Not sure about submitting on enter with two buttons (and having checkbox which changes one of the buttons function)...

Maybe just change the button text depending on resolve checkbox state? And have "Reply and keep open"/"... keep resolved" (or some better wordings) when no change.

peppy:
can we just remove the checkbox and have two buttons? maybe i'm just biased from using github for too long, but that feels the most intuitive

nanaya:
that works as well though the non-primary button should be colored differently.

...I don't remember we have such color though 🤔

notbakaneko:
I think there shouldn't be a checkbox but an indicator somewhere when its resolved

notbakaneko:
also `cmd+enter` will do a normal comment on github

notbakaneko:
image

image

could do with different button colours, though 🤔

nanaya:
Press enter to what

notbakaneko:
Existing Korean and Chinese translations for `beatmaps.discussions.reply_notice` need to be updated.

resources/assets/coffee/react/beatmap-discussions/new-reply.coffee
```diff
@@ -127,26 +137,32 @@ class BeatmapDiscussions.NewReply extends React.PureComponent
...
     return if !@validPost()
     LoadingOverlay.show()
 
     @postXhr?.abort()
     @setState posting: true
 
+    resolved = switch event.currentTarget.dataset.action
+               when 'resolve' then true
+               when 'reopen' then false
+               else @state.resolveDiscussion
```
nanaya: apparently this state can be removed and use the props directly instead.

PR #1821, opened 2017-11-27
Beatmapset page fix styling

nanaya:
Current

Fixed


notbakaneko:
Is this the only page with a multi-colour body background?

nanaya:
I think this is all ┐(･_･)┌

PR #1824, opened 2017-11-27
Add wiki title search

### LINKED ISSUE #1817: Make wiki links case-insensitive
When visiting a wiki link with incorrect casing, the user should be redirected to the correctly cased link, rather than 404'd.

nanaya:
Try searching by title using elasticsearch if no exact match or redirect.

Fixes #1817.


peppy:
Did you happen to figure out the performance issues we were seeing with this yesterday?

PR #1828, opened 2017-11-28
Add way to put outdated notice box in wiki page

### LINKED ISSUE #1827: Add a way to mark wiki entries as "outdated"
Generally on wikipedia this is done by attaching an outdated header. We need something basic to display and tag pages as being outdated. The main purpose is to allow english base pages to be updated without deleting every other translation for the page (which is what some people have been doing until now).

nanaya:
Just like news, put `outdated: true` in page metadata area

```
---
outdated: true
---
<content goes here>
```

Fixes #1827.

### DIFF
--- a/resources/lang/en/wiki.php
+++ b/resources/lang/en/wiki.php
@@ -31,5 +31,10 @@ return [
             'link' => 'Show on GitHub',
             'refresh' => 'Refresh',
         ],
+
+        'outdated' => [
+            '_' => 'This page is outdated. Check :default for latest update and perhaps help translating it!',
+            'default' => 'English version',
+        ],
     ],
 ];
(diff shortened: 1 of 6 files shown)


resources/lang/en/wiki.php
```diff
@@ -31,5 +31,10 @@
             'link' => 'Show on GitHub',
             'refresh' => 'Refresh',
         ],
+
+        'outdated' => [
+            '_' => 'This page is outdated. Check :default for latest update and perhaps help translating it!',
```
peppy: `This page contains an outdated translation of the original content. Please check the :default for the most accurate information (and consider updating the translation if you are able to help out)!`

PR #1829, opened 2017-11-28
Allow user to resolve notes

### LINKED ISSUE #1825: Mapper notes should display under the "Pending" filter
As discussed on discord, when a mapper makes a note they want it to be seen by others (they either have an issue or something important to say). As most people are filtering to pending, there's consensus that they should be displayed here.

nanaya:
So it can be put into pending filter. Fixes #1825.

### DIFF
--- a/app/Models/BeatmapDiscussion.php
+++ b/app/Models/BeatmapDiscussion.php
@@ -44,7 +44,7 @@ class BeatmapDiscussion extends Model
         'mapper_note' => 3,
     ];
 
-    const RESOLVABLE_TYPES = [1, 2];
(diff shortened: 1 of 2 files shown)


nekodex:
This also allows mappers to "resolve" notes, which we don't want? We just want notes to always be displayed (p. much ignoring most of the filters)

Xinnoh:
Notes can be used to ask questions, when the question is answered it should be resolved.

screen shot 2017-11-29 at 9 52 08 am

peppy:
We're adding back suggestions/problems for use for asking things which can be resolved.

PR #1830, opened 2017-11-28
Make user_sig column utf8mb4

### LINKED ISSUE #1771: SQL error when trying to set unicode characters as a signature
Bug image

Tested with a few other (special) unicode characters and didn't work with them either.

notbakaneko:
fixes #1771


nanaya:
This looks like will take a quite bit.

Also migration file names are snake cased.

peppy:
Should be okay to run as a migration, from experience.

PR #1850, opened 2017-11-30
 Increase max friends for supporters to 500

### LINKED ISSUE #1708: Increase friend limit
Opening as an issue first to promote discussion on github
Currently one of the most popular [feature requests](https://osu.ppy.sh/forum/t/176843) is to increase maximum friends from 250 to higher.
The limit has been set to 250 for almost 5 years, and is one of the most outdated rules on the site considering how much the game has grown since.
Increasing the limit only takes a couple minutes [on this page](https://github.com/ppy/osu-web/blob/5420da692f42a7371c9c92c049f0e8cceb4b9165/config/osu.php) and would be accepted as a welcome improvement across the entire community.

There's also an option to make this a supporter exclusive feature, which would encourage more people to buy the tag.

nekodex:
resolves #1708

### DIFF
--- a/app/Http/Controllers/FriendsController.php
+++ b/app/Http/Controllers/FriendsController.php
@@ -64,27 +64,31 @@ class FriendsController extends Controller
     public function store()
     {
         $currentUser = Auth::user();
-        $friends = $currentUser->friends()->get();
+        $friends = $currentUser->friends(); // don't fetch (avoids potentially instantiating 500+ friend objects)
 
         if ($friends->count() >= $currentUser->maxFriends()) {
             return error_popup(trans('friends.too_many'));
         }
 
-        $target_id = get_int(Request::input('target'));
-        $targetUser = User::lookup($target_id, 'id');
+        $targetId = get_int(Request::input('target'));
+        $targetUser = User::lookup($targetId, 'id');
+
+        if (!$targetUser) {
+            abort(404);
+        }
 
         $alreadyFriends = $friends
-            ->where('user_id', $target_id)
+            ->where('user_id', $targetId)
             ->first();
 
         if (!$alreadyFriends) {
             UserRelation::create([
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/FriendsController.php
```diff
@@ -63,21 +63,23 @@ public function index()
 
     public function store()
     {
-        if (Auth::user()->friends()->count() >= config('osu.user.max_friends')) {
+        $currentUser = Auth::user();
+        $friends = $currentUser->friends()->get();
```
nanaya: two `select count(*)` might be faster than one `select *` (+object creations)?

app/Http/Controllers/FriendsController.php
```diff
@@ -63,21 +63,23 @@ public function index()
...
+
+        if ($friends->count() >= $currentUser->maxFriends()) {
             return error_popup(trans('friends.too_many'));
         }
 
         $target_id = get_int(Request::input('target'));
-        $user = User::find($target_id)->firstOrFail();
+        $targetUser = User::lookup($target_id, 'id');
```
nanaya: not used? Or is it supposed to fail when not exist? (except it doesn't anymore)

app/Http/Controllers/FriendsController.php
```diff
@@ -63,21 +63,23 @@ public function index()
...
     {
-        if (Auth::user()->friends()->count() >= config('osu.user.max_friends')) {
+        $currentUser = Auth::user();
+        $friends = $currentUser->friends()->get();
+
+        if ($friends->count() >= $currentUser->maxFriends()) {
             return error_popup(trans('friends.too_many'));
         }
 
         $target_id = get_int(Request::input('target'));
```
nanaya: can use some camelCase.

PR #1857, opened 2017-11-30
Better checkbox and account playstyle edit

nanaya:
Noticed there's problem with opacity settings between checkbox's box and its tick which needed refactor.

Also updated playstyle change handling so it only make one request at a time and abort any previous requests.

Oh and `element.dataset.someStuff = '1'` doesn't trigger redraw in ie =D


nekodex:
If you're gonna batch updates, maybe increase the debounce delay so it will actually catch more than one click?

PR #1862, opened 2017-11-30
Fix check for allow kudosu button

### LINKED ISSUE #1861: Kudosu denial cannot be reverted
Currently on moddingv2, if a bn/qat denies a post of kudosu, other bns cannot undeny the post should they deem it kudosu-worthy. There are many times where posts will be incorrectly denied and another bn needs to step in, so the lack of this feature is problematic.

nanaya:
Accidentally broken. Fixes #1861.


peppy:
👍

PR #1870, opened 2017-12-01
Bubble popping

nekodex:
Reset nominations when a BN/QAT posts a new problem on a beatmapset that already has nominations (but is not yet qualified)

### DIFF
--- a/app/Http/Controllers/BeatmapsetsController.php
+++ b/app/Http/Controllers/BeatmapsetsController.php
@@ -144,22 +144,13 @@ class BeatmapsetsController extends Controller
             $lastDiscussionUpdate = $beatmapset->lastDiscussionTime();
             $lastEventUpdate = $beatmapset->events()->max('updated_at');
 
-            $hasUpdates = false;
-            if ($lastDiscussionUpdate !== null || $lastEventUpdate !== null) {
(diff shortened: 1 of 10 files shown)


app/Http/Controllers/BeatmapsetsController.php
```diff
@@ -139,13 +139,29 @@ public function discussion($id)
...
-        $lastUpdated = $beatmapset->lastDiscussionTime();
 
-        if ($returnJson && (
-            $lastUpdated === null ||
-            ($requestLastUpdated !== null && $requestLastUpdated >= $lastUpdated->timestamp)
-        )) {
-            return response([], 304);
+        if ($returnJson) {
+            $lastDiscussionUpdate = $beatmapset->lastDiscussionTime();
+            $lastEventUpdate = $beatmapset->events()->max('updated_at');
```
nanaya:
```php
if ($lastEventUpdate !== null) {
  $lastEventUpdate = Carbon::parse($lastEventUpdate);
}

$latestUpdate = max($lastDiscussionUpdate, $lastEventUpdate);

if ($latestUpdate === null || $requestLastUpdate < $latestUpdate) {
  return response([], 304)
}
```

PR #1887, opened 2017-12-04
Set correct filter when jumping to discussion

### LINKED ISSUE #1877: When clicking on timestamps, the page should swap to show all mods
When viewing one of the mv2 comment categories (eg resolved, pending, mine), if you click on a timestamp, it will scroll to the timestamp but there is no comment to show unless the user scrolls to the top to change views.

eg screenshot shows pending mods only
screen shot 2017-12-03 at 4 09 01 pm

nanaya:
Probably should look into using promise thingy already.

Fixes #1877.

### DIFF
--- a/resources/assets/coffee/react/beatmap-discussions/main.coffee
+++ b/resources/assets/coffee/react/beatmap-discussions/main.coffee
@@ -254,22 +254,31 @@ class BeatmapDiscussions.Main extends React.PureComponent
       else
         'generalAll'
 
-    @setMode null,
-      mode: mode
-      callback: =>
-        @setCurrentBeatmapId null,
-          id: discussion.beatmap_id
-          callback: =>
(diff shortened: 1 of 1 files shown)


resources/assets/coffee/react/beatmap-discussions/main.coffee
```diff
@@ -254,22 +254,32 @@ class BeatmapDiscussions.Main extends React.PureComponent
...
+        'total'
+
+    setFilter = =>
+      @setFilter null, filter: filter, callback: setMode
+    setMode = =>
+      @setMode null, mode: mode, callback: setCurrentBeatmapId
+    setCurrentBeatmapId = =>
+      @setCurrentBeatmapId null, id: discussion.beatmap_id, callback: jump
+    jump = =>
+      console.log 'jumping'
```
peppy: intentional?

PR #1911, opened 2017-12-06
Replace close button in forum reply box

### LINKED ISSUE #1624: Make show/hide reply box more intuitive
Right now it is possible to hide and show your reply box when browsing any thread in the forum.
The button for show is placed here
grafik
and the button for hide is placed here, only visible whie hovering the reply box area:
grafik

How this could be improved:
1. Add a mouse-over to the button for hiding
2. Change the `x` icon to a `-`. The current icon suggests that your text is getting deleted rather than that the reply box is hidden
3. While the reply box is opened, the `show` button on the bottom right could change to a redundant hide button as it really does nothing in the meantime either. 

The latter would especially help with the following issue:
Once you put your first characters into the reply box, it will no longer auto-hide when clicking the area above it.
If you are at the very bottom of the page the `hide` button no longer appears upon hovering the reply box area.
This means that the only way to hide your reply box and reread some post in full height is to start scrolling up which isn't logically apparent. It may also be considered that many people would move their cursor to the area above the reply box because on many non-osu! websites scrolling up wouldn't work if a potentially scrollable textbox like the reply box is selected so they wouldn't even see the `hide` button reappear upon scrolling up which makes you feel stuck.

nanaya:
Using similar logic as new discussion. Fixes #1624.

Now it's just:
- pin toggle at corner
- focus text box to pin
- clicking anywhere else doesn't unpin anymore
- RIP close button


nekodex:
Probably needs some feedback when you hit the 'show reply box' button when it's already pinned... right now it just does nothing which feels weird. Maybe flash the reply box border to draw attention to it or something?

PR #1921, opened 2017-12-07
Add direct support for elasticsearch indexing

notbakaneko:
Adds support for re-indexing into the Elasticsearch indices.
This does not add index update on model change yet; that will be a separate PR on top of this one.

This should replace the need to run the deprecated elasticsearch river plugin and let us look at upgrading Elasticsearch. Re-running the indexing on the same index will remove soft-deleted records from the index, so, it's not necessary to do the whole creating a new index and dropping the old one unless there's a hard-delete or the schema changes.

The `es:index-documents` command supports hot-indexing and cold-indexing; 
when cold-indexing it will write to the existing index;
when hot-indexing it will create a new index and update the alias mapping when completed.

If the field mapping of existing fields change, a new index should be created to update the schema.

The first time the indexing is run, it will need to be run as 
`php artisan es:index-documents --hot --cleanup`, since the field mappings of boolean columns in the model are now mapped to `boolean` instead of `long`. `1/0` are interchangeable with `true/false` when querying a `boolean` field, so the existing queries should work fine.


notbakaneko:
uh, I should probably add bulk updating first >_>

app/Traits/EsIndexable.php
```diff
@@ -0,0 +1,164 @@
...
+            return $next;
+        }, $baseQuery, $batchSize, $fromId);
+
+        $duration = time() - $startTime;
+        Log::info("Indexed {$count} records in {$duration} s.");
+    }
+
+    /**
+     * Paginates and indexes the recordsets using key-set pagination instead of
+     *  the offset pagination used by chunk().
```
nanaya: apparently there's this `chunkById`

notbakaneko:
will need some change to the beatmap search to use the correct field for sorting

notbakaneko:
Just remembered we can just alias multiple indices to the same name, so there's no reason not to split the indices up first.

app/Libraries/Elasticsearch/Indexing.php
```diff
@@ -0,0 +1,98 @@
...
+{
+    public static function deleteIndex(string $name)
+    {
+        return Es::indices()->delete([
+            'index' => $name,
+            'client' => ['ignore' => 404],
+        ]);
+    }
+
+    public static function getOldIndices(string $alias)
```
nanaya: it's just one `OldIndex`, isn't it?

nanaya: wut it's not ಠ_ಠ

notbakaneko: alias can have many indices and index can have many aliases :D

PR #1980, opened 2017-12-13
Fix tabbing when in news article previews

blissfulyoshi:
Prevent tabbing to the image to prevent the user from being able to tab both the image and the title text of the news article.


notbakaneko:
What's the use-case for this?

Tabbing through is not being considered at this time.

blissfulyoshi:
The use case is tabbing between news articles on the homepage, which currently lets you tab to 2 links to the same article, the image and the title text.

As to whether or not you want to fix it, that is up to you, but unless you're planning to redesign the news section, I don't see any harm in fixing a minor bug there.

peppy:
I'm not against this being added, but there are a lot of other tab fixes which should also be applied, including either showing the hidden menu or not tabbing through invisible links.

PR #1981, opened 2017-12-13
Configurable required hype

nanaya:
Starting from the easiest =D


config/osu.php
```diff
@@ -21,6 +21,9 @@
         'thumbnailer' => env('BM_PROCESSOR_THUMBNAILER', 'http://localhost:4001'),
         'sentry' => env('BM_PROCESSOR_SENTRY'),
     ],
+    'beatmapset' => [
+        'required_hype' => get_int(env('BEATMAPSET_REQUIRED_HYPE')) ?? 5,
```
nekodex: why not just `get_int(env('BEATMAPSET_REQUIRED_HYPE', 5))`?

nanaya: because `BEATMAPSET_REQUIRED_HYPE=` will turn it into 0.

nekodex: good thing we used that pattern everywhere else in this file then 👍

PR #1983, opened 2017-12-13
Convert store columns to fixed precision

notbakaneko:


nanaya:
And then it becomes string when accessed through laravel...

notbakaneko:
so it does ಠ_ಠ

nanaya:
bcmath time...? ಠ_ಠ

peppy:
are we ok to merge this in its current state?

notbakaneko:
should be safe ™️

PR #1985, opened 2017-12-13
Hype post

nanaya:
Seems kind of somehow working.


nekodex:
Probably want to update this text:

PR #1989, opened 2017-12-13
Add admin ability to remove beatmapset covers

nekodex:
as it says on the tin

### DIFF
--- a/app/Http/Controllers/Admin/BeatmapsetsController.php
+++ b/app/Http/Controllers/Admin/BeatmapsetsController.php
@@ -26,13 +26,22 @@ use Request;
 
 class BeatmapsetsController extends Controller
 {
-    protected $section = 'admin.beatmapsets';
+    protected $section = 'admin';
+    protected $actionPrefix = 'beatmapsets-';
 
     public function covers($id)
     {
         $beatmapset = Beatmapset::findOrFail($id);
 
         return view('admin.beatmapsets.cover', compact('beatmapset'));
+    }
+
+    public function removeCovers($id)
+    {
+        $beatmapset = Beatmapset::findOrFail($id);
+        $beatmapset->removeCovers();
+
+        return back();
     }
 
     public function regenerateCovers($id)
--- a/app/Http/Controllers/Admin/BeatmapsetsController.php
+++ b/app/Http/Controllers/Admin/BeatmapsetsController.php
@@ -41,7 +41,7 @@ class BeatmapsetsController extends Controller
         $beatmapset = Beatmapset::findOrFail($id);
         $beatmapset->removeCovers();
 
-        return back();
+        return response([], 204);
     }
 
     public function regenerateCovers($id)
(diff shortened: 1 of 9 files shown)


resources/lang/en/layout.php
```diff
@@ -146,9 +146,9 @@
             'logs-index' => 'log',
             'beatmapsets' => [
                 '_' => 'beatmapsets',
-                'covers' => 'covers',
                 'show' => 'detail',
             ],
+            'beatmapsets-covers' => 'beatmapset covers',
```
nanaya: keys with string values are before arrays.

routes/web.php
```diff
@@ -20,6 +20,7 @@
 Route::group(['as' => 'admin.', 'prefix' => 'admin', 'namespace' => 'Admin'], function () {
     Route::get('/beatmapsets/{beatmapset}/covers', 'BeatmapsetsController@covers')->name('beatmapsets.covers');
     Route::post('/beatmapsets/{beatmapset}/covers/regenerate', 'BeatmapsetsController@regenerateCovers')->name('beatmapsets.covers.regenerate');
+    Route::post('/beatmapsets/{beatmapset}/covers/remove', 'BeatmapsetsController@removeCovers')->name('beatmapsets.covers.remove');
```
nanaya: `Route::delete`?

resources/views/admin/beatmapsets/cover.blade.php
```diff
@@ -18,20 +18,29 @@
...
+            <h2>{{$beatmapset->title}} - {{$beatmapset->artist}}</h2>
             <br>
-        @endforeach
+            {!! Form::open([
+                'route' => ['admin.beatmapsets.covers.regenerate', $beatmapset->beatmapset_id],
+                'method' => 'POST'
+            ]) !!}
+                <button>{{trans('admin.beatmapsets.covers.regenerate')}}</button>
+            {!! Form::close() !!}
+            {!! Form::open([
```
nanaya: There's this `data-remote` thingy which can have `data-reload-on-success`.

[Example](/ppy/osu-web/blob/aa817fd71a1457055eec711e11af2ea6b80d8148/resources/views/beatmapset_watches/index.blade.php#L86).

routes/web.php
nekodex: idk, delete feels wrong in this case as it's not a singular resource that's being deleted?

app/Http/Controllers/Admin/BeatmapsetsController.php
```diff
@@ -35,6 +36,14 @@ public function covers($id)
         return view('admin.beatmapsets.cover', compact('beatmapset'));
     }
 
+    public function removeCovers($id)
+    {
+        $beatmapset = Beatmapset::findOrFail($id);
+        $beatmapset->removeCovers();
+
+        return back();
```
nanaya: Causes one extra page request =D

PR #1990, opened 2017-12-13
Vote log

nanaya:
No interface or anything, just logging who did the vote and record the current votes if the vote causes kudosu grant/revoke.

### DIFF
--- a/app/Models/BeatmapDiscussion.php
+++ b/app/Models/BeatmapDiscussion.php
@@ -152,8 +152,7 @@ class BeatmapDiscussion extends Model
                 }
 
                 $eventExtraData['votes'] = $this
-                    ->beatmapDiscussionVotes()
-                    ->get()
+                    ->beatmapDiscussionVotes
                     ->map
                     ->forEvent();
             } elseif ($event === 'recalculate') {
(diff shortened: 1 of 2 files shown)


app/Models/BeatmapDiscussion.php
```diff
@@ -143,19 +143,25 @@ public function refreshKudosu($event)
...
             if ($event === 'vote') {
                 if ($change > 0) {
                     $beatmapsetEventType = BeatmapsetEvent::KUDOSU_GAIN;
                 } else {
                     $beatmapsetEventType = BeatmapsetEvent::KUDOSU_LOST;
                 }
+
+                $eventExtraData['votes'] = $this
+                    ->beatmapDiscussionVotes()
+                    ->get()
```
notbakaneko: `->beatmapDiscussionVotes`  is already a Collection so `()->get()` seems unnecessary?

nanaya: apparently not.

```
Undefined property: Illuminate\Database\Eloquent\Relations\HasMany::$map in /home/edho/osu/osu-web/app/Models/BeatmapDiscussion.php:156
```

┐(･_･)┌

notbakaneko:
>_>

nanaya:
>_<

PR #1995, opened 2017-12-15
Add links to legacy forum threads on beatmapset pages that don't use moddingv2

### LINKED ISSUE #1946: Provide a link to beatmap discussion on sets with v1
https://osu.ppy.sh/beatmapsets/570924/
It's not possible to access the [v1 thread of a map](https://osu.ppy.sh/forum/t/556231) with the new forum. This makes it impossible to reach discussions that happened on the old forum.
Yes, mv2 is replacing v1, but users are unable to access v1 threads on sets ranked prior to mv2.

nekodex:
fixes #1946

### DIFF
--- a/config/osu.php
+++ b/config/osu.php
@@ -81,6 +81,7 @@ return [
         'base' => 'https://osu.ppy.sh',
         'dev' => 'https://discord.gg/ppy',
         'installer' => 'https://m1.ppy.sh/r/osu!install.exe',
+        'legacy-forum-thread-prefix' => '/forum/t/',
         'smilies' => '/forum/images/smilies',
         'support-the-game' => '/p/support#transactionarea',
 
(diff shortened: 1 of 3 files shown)


resources/assets/coffee/react/beatmapset-page/header.coffee
```diff
@@ -195,6 +195,14 @@ class BeatmapsetPage.Header extends React.Component
                 icon: 'comments-o'
                 props:
                   href: laroute.route 'beatmapsets.discussion', beatmapset: @props.beatmapset.id
+            else if @props.beatmapset.legacy_thread_url
```
nanaya: the check is always true because of prefix?

config/osu.php
```diff
@@ -81,6 +81,7 @@
         'base' => 'https://osu.ppy.sh',
         'dev' => 'https://discord.gg/ppy',
         'installer' => 'https://m1.ppy.sh/r/osu!install.exe',
+        'legacy-forum-thread-prefix' => '/forum/t/',
```
nanaya: apparently most of the config keys are snake cased.

nanaya: or maybe not for routes? Why did they use dashes to begin with ┐(°～° )┌

Xinnoh:
Doesn't appear to be working

nekodex:
It hasn't been deployed yet.

Issue #1944, opened 2017-12-11
Allow users to choose the order of favourite beatmaps on their profile


Xinnoh:
On the userpage, beatmaps are sorted by id. The oldest six maps in favourites/ranked are listed for other players see.
The user should be able to choose which 6 maps other players initially see when browsing favourite/ranked beatmaps.

peppy:
as a quick-fix, it should probably show the most recently favourited maps.

nanaya:
Other sort options can be done through beatmap search :thinking:

PR #1999, opened 2017-12-15
Show favourite beatmaps on user profile in reverse favourited order

notbakaneko:
sort of related to #1944

tried adding an index to `dateadded` for sorting but mysql didn't seem to want to use it

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -731,7 +731,13 @@ class User extends Model implements AuthenticatableContract, Messageable
 
     public function favouriteBeatmapsets()
     {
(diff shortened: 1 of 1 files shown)


peppy:
What's the reason for dropping back to raw sql?

notbakaneko:
`hasManyThrough` doesn't work on composite keys and the ordering needs to be done on the intermediary table; we can't just select ids and select `Beatmapset` since that won't return the correct order.

peppy:
lame

notbakaneko:
apparently this is breaking search :|

PR #2020, opened 2017-12-18
Allow beatmap discussion reply editor to be closed

notbakaneko:
with cancel button or `esc` key


resources/assets/coffee/react/beatmap-discussions/new-reply.coffee
```diff
@@ -74,6 +76,8 @@ class BeatmapDiscussions.NewReply extends React.PureComponent
         className: "#{bn}__footer"
         div className: "#{bn}__actions",
           div className: "#{bn}__actions-group",
+            @renderCancelButton()
```
nanaya: dis coo.

(probably should be removed)

notbakaneko: 😑

PR #2023, opened 2017-12-18
Fix beatmap discussion reply's cancel button style

notbakaneko:
looks silly in safari >_>
and everything not chrome

also remove the button that isn't supposed to be there :trollface:

### DIFF
--- a/resources/assets/less/bem/beatmap-discussion-post.less
+++ b/resources/assets/less/bem/beatmap-discussion-post.less
@@ -75,9 +75,11 @@
     &--cancel {
       .reset-input();
       position: absolute;
-      top: 10px;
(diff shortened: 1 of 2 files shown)


resources/assets/less/bem/beatmap-discussion-post.less
```diff
@@ -73,9 +73,11 @@
     }
 
     &--cancel {
+      .reset-input();
       position: absolute;
       top: 10px;
       right: 10px;
+      padding: 3px 13px; // how big the safari button would have been.
```
nanaya: probably better set width to `@beatmap-discussion--reply-padding` and add left/right padding accordingly?

PR #2031, opened 2017-12-19
User beatmapset admin

### LINKED ISSUE #1969: Add (admin level?) page to view a user’s usage of the modding discussion system
Moderators need some visibility of a user's posts/actions/votes-received on the system. Doesn't need to be pretty (or public-facing) but just functional.

nanaya:
More like crude sql interface. The views aren't quite done ( ･_･)a

Entry point(?): `/users/{id}/beatmapset-activities`

Kinda fixes #1969.


nekodex:
I'm taking this over to add some styling to get these pages into a usable state while nanaya is off.

resources/views/beatmap_discussion_posts/_item.blade.php
```diff
@@ -0,0 +1,65 @@
...
+                </div>
+            </div>
+        </div>
+        <div class="beatmap-discussion__discussion">
+            <div class="beatmap-discussion__top">
+                <div class="beatmap-discussion-post beatmap-discussion-post--discussion">
+                    <div class="beatmap-discussion-post__content">
+                        <div class="beatmap-discussion-post__user-container">
+                            <div class="beatmap-discussion-post__avatar">
+                                <a href="{{ route('users.beatmapset-activities', $post->beatmapDiscussion->user) }}">
```
nanaya: should be `$post->user`? (also for avatar and links etc)

nanaya:
THE BACKEND SEEMS RATHER FINE-ISH.

(no review cuz my own pull request)

nekodex:
Should probably mention that this was rushed to get _something_ out for moderators to use and as such is rather _scrappy_ quality (and shouldn't be indicative of target code quality.)

Will be reworked when fixed up for public use.

PR #2032, opened 2017-12-19
Don't explode if beatmapset is deleted

notbakaneko:
Quick fix for exploding user pages
https://osu.ppy.sh/users/6969988


app/Models/BeatmapDiscussion.php
```diff
@@ -292,7 +292,7 @@ public function vote($params)
     public function title()
     {
         if ($this->beatmap_id === null) {
-            return $this->beatmapset->title;
+            return $this->beatmapset ? $this->beatmapset->title : '[deleted beatmap]';
```
nanaya: 🤔 

```php
// beatmap-specific discussion
if ($this->beatmap !== null) {
    return "{$this->beatmapset->title} [{$this->beatmap->version}]";
}

// generalAll discussion
if ($this->beatmap_id === null && $this->beatmapset !== null) {
    return $this->beatmapset->title;
}

return '[deleted beatmap]';
```

nanaya: or a bit more detailed with something like

```
if beatmap_id == null
  if beatmapset
    beatmapset title
  else
    deleted beatmap
else
  if beatmap
    beatmapset [beatmap]
  else
    beatmapset [deleted difficulty]
```

PR #2034, opened 2017-12-19
Skip validation on updating user's last visit

nanaya:
( ﾟ◡ﾟ)


nanaya:
Actually might be better check if anything is actually changed first...?

PR #2043, opened 2017-12-20
Show 20 extra results per profile page pagination, rather than 5

### LINKED ISSUE #2010: Increase pagination per “show more” on profile
One of the largest remaining frustrations reported on the new site is that the "show more" links to reveal more scores/betamaps on profile pages do not show enough items each time they are clicked.

Let's change them to show 20 or 50 at a time.

peppy:
The initial display is unchanged – only pagination is affected.

Closes #2010.


nanaya:
A bit easier with #1594.

peppy:
Closing in favour of #1594

PR #2045, opened 2017-12-20
Update README.md

kj415j45:


kj415j45:
on hold

nanaya:
Maybe if it's in the file itself instead of an external link.

kj415j45:
the short guide in README is not very clear. so should i create another md? @nanaya

Issue #1873, opened 2017-12-01
Slow queries which need to be fixed


peppy:
```SELECT `user_id`, `beatmap_id`, `score`, `rank` FROM `osu_scores_high` WHERE `user_id` = 3172980```

peppy:
```select * from `osu_user_stats_fruits` where exists (select * from `phpbb_users` where `osu_user_stats_fruits`.`user_id` = `phpbb_users`.`user_id` and (`user_warnings` = 0 and `user_type` = 0)) and `country_acronym` = 'TR' order by `ranked_score` desc limit 50 offset 500```

peppy:
```select * from `osu_user_stats_mania` where exists (select * from `phpbb_users` where `osu_user_stats_mania`.`user_id` = `phpbb_users`.`user_id` and (`user_warnings` = 0 and `user_type` = 0)) and `country_acronym` = 'VE' order by `rank_score` desc limit 50 offset 9150```

peppy:
```select min(`post_id`) as aggregate from `phpbb_posts` where `phpbb_posts`.`topic_id` = 134389 and `phpbb_posts`.`topic_id` is not null and `phpbb_posts`.`deleted_at` is null```

peppy:
```SELECT count(*) AS AGGREGATE FROM `phpbb_zebra` WHERE `zebra_id` = 18983 AND `friend` = 1``` (index on `(zebra_id, friend)` will help a bit)

peppy:
```select * from `osu_user_stats_fruits` where exists (select * from `phpbb_users` where `osu_user_stats_fruits`.`user_id` = `phpbb_users`.`user_id` and (`user_warnings` = 0 and `user_type` = 0)) and `country_acronym` = 'VN' order by `ranked_score` desc limit 50 offset 9700``` many queries hitting the non-optimised end of ranking pages (googlebot?)

peppy:
```select * from `osu_user_stats_fruits` where exists (select * from `phpbb_users` where `osu_user_stats_fruits`.`user_id` = `phpbb_users`.`user_id` and (`user_warnings` = 0 and `user_type` = 0)) and `country_acronym` = 'NO' order by `ranked_score` desc limit 50 offset 550```

peppy:
```select count(*) as aggregate from `osu_scores_high` where `user_id` = 5287410 and `pp` > (select `pp` from `osu_scores_high` where `score_id` = 2416055216)```

peppy:
[...]

PR #2047, opened 2017-12-20
Include score in osu_scores_high index

notbakaneko:
so the index covers the query in https://github.com/ppy/osu-web/issues/1873#issue-278523250

Adds a new index and drops the old one.
Creating the index will take a while 🔥


peppy:
What is the score being used for in this query?

notbakaneko:
some calculation thing

peppy:
Need more specific details. This is not an arbitrary index change, and will take 3-4 days to update and increase the overall database size by a non-trivial amount.

notbakaneko:
closing in favour of storing the missing counts

PR #2051, opened 2017-12-20
Add migration for new rank cache columns

peppy:


database/migrations/2017_12_20_090905_add_new_rank_cache_columns.php
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('osu_user_stats', function (Blueprint $table) {
```
notbakaneko: `use Illuminate\Database\Schema\Blueprint;` or remove `Blueprint`, same for others

PR #2056, opened 2017-12-20
Limit loaded forum posts to 250

nanaya:
Seems to work ┐(°～° )┌


nanaya:
Current one is already bouncing anyway 💃

Issue #2012, opened 2017-12-17
Search still has major issues


peppy:
As reported [here](https://twitter.com/Ted_the_red/status/942286132871647232), search still has a long way to go.

- User search:

Searching for "Toy" (user Toy is nowhere to be seen). Recently active users and exact matches should be prioritised

<img src='https://shipusercontent.com/0bc77910500c145f28e70942001219db/Pasted%20Image%201.png' title='Pasted Image 1.png' width=555>

- WIki search:

Searching for "osu world cup" (page titled "osu! world cup" is nowhere to be seen). Recently updated and... matches relevant to search terms(?) should be displayed.

<img src='https://shipusercontent.com/1eba837df29e5b4ca0f1376d98e6d918/Pasted%20Image%202.png' title='Pasted Image 2.png' width=954>

peppy:
Beatmap search also reported to be less reliable: https://twitter.com/vinxis1/status/942348957203578882

nanaya:
Not sure how he did the beatmap search because I'm getting different result.

![](https://s.myconan.net/2017-12/firefox_2017-12-19_19-18-15.jpg)

TedTheRed:
It was probably the generic search. 
Generic search directed to beatmap page
![generic search](https://i.imgur.com/CQ8KHgz.jpg)
Search directly from the beatmap page
![beatmap serach](https://i.imgur.com/Jn3ct0n.jpg)

peppy:
well spotted!

TedTheRed:
Ah the generic search is bad probably because its default is prioritizing updated first and then relevancy. If your making a generic search request, relevancy should take priority first and then sort by updated. 

Should probably also check out the behavior of switching to pending or graveyard. If there's a query it defaults to updated instead of relevancy

peppy:
Wiki and beatmap have been resolved.

Better user search is a WIP.

PR #2061, opened 2017-12-21
Slightly smarter beatmap search default

nanaya:
Can use some more testing.


peppy:
Possibly related to solving the beatmap portion of #2012?

nanaya:
yes

PR #2062, opened 2017-12-21
Ensure that all suggestions/problems are resolved before allowing nomination

### LINKED ISSUE #1940: Nomination should only be possible when all suggestions/problems are in a resolved state
After further discussion on discord, it is currently hard to know which problems were pointed out after a nomination occurred (in the case they reset nomination status).

To resolve this, let's make it only possible to nominate a map after all suggestions and problems are resolved. This means that any pending items when a map is in a nominated state are guaranteed to have been posted after the nomination.

nekodex:
Fixes #1940


resources/lang/en/beatmaps.php
```diff
@@ -134,6 +134,7 @@
         'reset-confirm' => 'Are you sure? Posting a new problem will reset nominations.',
         'required-text' => 'Nominations: :current/:required',
         'title' => 'Nomination Status',
+        'unresolved-issues' => 'There are still unresolved issues that must be addressed first.',
```
nanaya: snakes 🐍 

(yes the others also need to be fixed soon™)

resources/assets/coffee/react/beatmap-discussions/nominations.coffee
```diff
@@ -176,13 +185,12 @@ class BeatmapDiscussions.Nominations extends React.PureComponent
...
-                    el BigButton,
-                      modifiers: ['full']
-                      text: osu.trans 'beatmaps.nominations.nominate'
-                      icon: 'thumbs-up'
-                      props:
-                        disabled: @props.beatmapset.nominations.nominated
-                        onClick: @nominate
+                    if @props.currentDiscussions.unresolvedIssues > 0
+                      # wrapper 'cuz putting a title/tooltip on a disabled button is no worky...
+                      div title: osu.trans('beatmaps.nominations.unresolved-issues'),
```
nanaya: 🐍

PR #2063, opened 2017-12-21
Show correct star difficulty for converted beatmaps

### LINKED ISSUE #2008: Converted beatmap do not show correct star difficulty
As reported [here](https://twitter.com/Chickenbible/status/942286705922641920).

notbakaneko:
fixes #2008

also updates BeatmapSeeder to work if there's existing data.


database/migrations/2017_12_19_053800_create_beatmap_difficulty.php
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        Schema::create('osu_beatmap_difficulty', function (Blueprint $table) {
+            $table->charset = 'utf8mb4';
+
+            $table->unsignedInteger('beatmap_id');
+            $table->tinyInteger('mode')->default(0);
+            $table->unsignedInteger('mods');
+            $table->float('diff_unified', null, null); // creates a double instead of float.
+            $table->timestamp('last_update');
```
nekodex: probably want `->useCurrent()` here

peppy:
should this migration be part of the initial tables migration?

notbakaneko:
It shouldn't, otherwise the table won't be added in dev unless the migrations are reset

peppy:
I guess we can start doing that. The tradition has just been updating the base ones and reinitiating db for old-system-existing tables.

app/Http/Controllers/BeatmapsetsController.php
```diff
@@ -86,6 +86,7 @@ public function show($id)
     {
         $beatmapset = Beatmapset
             ::with('beatmaps.failtimes', 'user')
+            ->with('beatmaps.difficulty')
```
nanaya: why is it not `difficulties` ಠ_ಠ

notbakaneko: itisamystery👻

database/seeds/ModelSeeders/BeatmapSeeder.php
```diff
@@ -1,21 +1,32 @@
...
      *
      * @return void
      */
     public function run()
     {
-        $beatmaps_array = [];
-        $beatmapset_array = [];
-        $overbeatmaps = [];
-        $overbeatmapsets = [];
+        if ($this->command) {
```
nanaya: what's with this check? Is it not always available or something?

notbakaneko: `command` is optional for `Seeder`s

nekodex:
Sorry, I didn't realise that was a legacy table - in that case the migration should go under base_tables to remain consistent with our current standard.

The reasoning for this was (iirc) to explicitly segregate legacy tables from the the new stuff we're building so people coming into the project looking at the migrations for reference won't be led astray by potentially questionable table structure choices. Also, it prevented the migrations from running on production during deployment.

We originally decided that dev db data was disposable and resetting the DB was acceptable when base_tables changed. Whether we still think that's acceptable? 🤷‍♂️

If we're going to change the standard, we should at least discuss and agree on a direction going forward instead of just changing it randomly in an unrelated pull request?

PR #2069, opened 2017-12-22
Use cached score ranks for display

notbakaneko:
blocked on waiting for new column values to be calculated

fixes https://github.com/ppy/osu-web/issues/1873#issue-278523250


notbakaneko:
soon ™️

PR #2072, opened 2017-12-22
Sort on correct play count field

notbakaneko:
was sorting on play counts on beatmaps instead of the beatmap set


nanaya:
das coo

PR #2101, opened 2017-12-26
Correct Beatmap search artist and title sort order

notbakaneko:
Depends on data being reindexed with new schema first.


notbakaneko:
closing this as I remembered an older branch with proper fixes

PR #2106, opened 2017-12-26
Fix bad English in the Store localization file

### LINKED ISSUE #1932: Bad english on store out-of-stock message
`This type is currently out of stock :(. Try other type or check back soon.`

should probably read something like

`Unfortunately this item is out of stock. Use the dropdown to choose a different type or check back later!`

TPGPL:
Closes #1932


resources/lang/en/store.php
```diff
@@ -47,8 +47,8 @@
         'name' => 'Name',
 
         'stock' => [
-            'out' => 'Currently out of stock :(. Check back soon.',
-            'out_with_alternative' => 'This type is currently out of stock :(. Try other type or check back soon.',
+            'out' => 'This item is currently out of stock. Check back later!',
+            'out_with_alternative' => 'Unfortunately this item is out of stock. Use the dropdown to choose a different type or check back later!',
```
Aergwyn: `choose a different type`? This still reads sketchy. (May be my ignorance because of missing knowledge of the specifics of the page though)

TPGPL: probably different type of merchandise

peppy:
Please stop removing other languages when fixing english.

PR #2154, opened 2018-01-03
More deterministic beatmap discussion post ordering

### LINKED ISSUE #2131: Sort by timeline not always showing in order
example: see hard on this set https://osu.ppy.sh/beatmapsets/706150/discussion

both screenshots were taken with sort mode set to timeline.
image

image

My post, (Noffy) , as shown in the first screenshot, the most recently made post, got slotted between two other points on the timeline that it shouldn't be. Changing the sort mode to date and back to timeline doesn't fix it either. 

Edit:
It changed again after I made another post, changing it from the screenshotted order, to a different out-of-order order. I have no idea.
image

Points will display in the correct order on the top timeline, but not in the discussions thread itself.

notbakaneko:
fixes #2131 

┐(ツ)┌


nanaya:
Might as well remove the initial sorting in `main.coffee`

PR #2171, opened 2018-01-04
Disable deleting first post of beatmapset thread

### LINKED ISSUE #2104: On the new site, It's possible to delete the original map post for sets with v1
delete option does not show on the old site
screen shot 2017-12-26 at 5 49 06 pm

lmao
screen shot 2017-12-26 at 5 52 13 pm

Still on the bss https://osu.ppy.sh/s/691505

nanaya:
┐(ಠ_ಠ)┌

Fixes #2104.


app/Models/Forum/Topic.php
```diff
@@ -126,25 +129,35 @@ public function addPost($poster, $body)
 
     public function removePost($post, $user = null)
```
notbakaneko: waitaminute, how long has `$user` been unused for? ಠ_ಠ
`restorePost()` too ಠ_ಠ

app/Models/Forum/Topic.php
```diff
@@ -126,25 +129,35 @@ public function addPost($poster, $body)
...
 
-            if ($this->posts()->exists() === true) {
-                $this->refreshCache();
-            } else {
-                $this->delete();
-            }
+        try {
+            return DB::transaction(function () use ($post, $user) {
+                if ($post->delete() === false) {
+                    $this->validationErrors()->addTranslated('post', $post->validationErrors()->toSentence());
```
notbakaneko: optionally can use `validationErrors()->merge(..)`

app/Models/Forum/Topic.php
```diff
@@ -126,25 +129,35 @@ public function addPost($poster, $body)
...
+                    $this->forum->refreshCache();
+                }
+
+                if ($post->user !== null) {
+                    $post->user->refreshForumCache($this->forum, -1);
+                }
+
+                return true;
+            });
+        } catch (ModelNotSavedException $_e) {
```
notbakaneko: may be better to just let the exception propagate and let the caller handle it instead of
`if false throw catch return false if false throw catch return false` party

app/Models/Forum/Topic.php
nanaya: but one is errors for post and the other is for topic...

PR #2182, opened 2018-01-05
Add support for model after commit hooks

notbakaneko:
Primitive `after_commit` support to be used by elasticsearch updates on model save when transaction completes.


nanaya:
Can use an example on how to use this ಠ_ಠ

notbakaneko:
AfterCommitTest :trollface:

app/Models/Model.php
```diff
@@ -93,4 +112,23 @@ public function saveOrExplode($options = [])
...
+
+            return;
+        }
+
+        if ($this instanceof AfterCommit) {
+            $transaction->addCommittable($this);
+        }
+
+        if ($this instanceof AfterRollback) {
+            $this->addRollbackable($this);
```
nanaya: `$transaction`?

app/Providers/TransactionStateServiceProvider.php
```diff
@@ -0,0 +1,64 @@
...
+     * @return void
+     */
+    public function register()
+    {
+        // Connection after commit tracker.
+
+        // Laravel uses a singleton DatabaseManager that reuses connections
+        // by name, so a singleton for tracking should be good enough.
+        // The alternative is to override Connection and its subclasses.
+        $this->app->singleton('TransactionState', function ($app) {
```
nanaya: there's `TransactionState` class and then there's `TransactionState` singleton which is actually `TransactionStateManager` ಠ_ಠ

app/Models/Model.php
```diff
@@ -21,6 +21,9 @@
 namespace App\Models;
 
 use App\Exceptions\ModelNotSavedException;
+use App\Libraries\Transactions\AfterCommit;
```
nanaya: I'd just add an empty method in base model and let the actual models override it instead of having to do the ceremonies...

notbakaneko: Don't want to be tracking models that don't do anything for after commit

nanaya:
still randomly failing =[

notbakaneko:
:rage3:

notbakaneko:
all the reverts :trollface:

PR #2183, opened 2018-01-05
Use updated elasticsearch index aliases

notbakaneko:
Use the new per-type `beatmaps`/`posts` aliases instead of everything in one alias.


app/Models/Beatmapset.php
```diff
@@ -415,7 +415,7 @@ public static function searchParams(array $params = [])
     public static function searchES(array $params = [])
     {
         $searchParams = [
-            'index' => config('osu.elasticsearch.index'),
+            'index' => config('osu.elasticsearch.index.beatmaps'),
```
nanaya: might as well use `static::esIndexName()` instead of this roundabout way? ಠ_ಠ

(though on unrelated note, a prefix/suffix/something for current env would be nice so tests don't clobber dev index)

PR #123, opened 2015-10-09
Fix event parser, etc

nanaya:
Fixes #121

PR #2196, opened 2018-01-10
Better linkify for beatmapset discussions

### LINKED ISSUE #2132: It's not possible to link to other modding discussion posts
image

due to the link not registering after the #, it is impossible to link to other discussion posts for any purpose such as reference in a format that it is clickable.

This:
- discourages discussion of related points at different times (can't cross-link two seperate posts)
- ^ the above can lead to if a modder suggests a fix at one time, and someone else wants to suggest something similar for a different time, they can't just simply link to the previous helpful post.
- can make it more difficult to say something like "if you apply a General post from a different difficulty, apply changes to this difficulty too for spread."

nekodex:
Fixes url regex missing some chars, which broke links to other discussions.

Also compacts the display of discussion urls to be either `#123` when on the same beatmapset or `1234#456` for discussions on another beatmapset.

Resolves #2132


resources/assets/coffee/react/beatmap-discussions/post.coffee
```diff
@@ -125,12 +125,40 @@ class BeatmapDiscussions.Post extends React.PureComponent
     @setState editing: true, =>
       @textarea.focus()
 
+  discussionLinkify: (text) =>
+    route = laroute.route('beatmapsets.discussion', {'beatmapset': "([0-9]+)"}) # what could possibly go wrong?
+    routeRegex = new RegExp(route.replace('/' , '\/'), 'i')
```
nanaya: `_.escapeRegExp`?

nekodex: except not ( ﾟ◡ﾟ)

peppy: is anyone supposed to be able to understand these kinda of comments? because i for one cannot.

nekodex: `_.escapeRegexp` doesn't work in this context because only the forward slashes are to be escaped. nanaya realised this and exclaimed "oh right" in person when he realised (after posting the comment). Thus I added a note so to remember it had been addressed.

I was also expecting this to not have taken over a day to get merged so I didn't bother spending 5mins to write up explaining what had happened.

PR #2200, opened 2018-01-10
dispatch elasticsearch asynchronously when refreshing hypes

### LINKED ISSUE #2195: ES updates should be async and not block
Noticed this while posting in a beatmapset discussion - it has the potential to break the ability to post if elasticsearch goes down or network issues occur. Also adds (albeit probably minuscule) latency to posting.

Can probably just shove on laravel's job queue and call it a day.

notbakaneko:
Seems to work

fixes #2195 

as a side note, I don't think `refreshCache()` should be automagically indexing the document; `save()`, yes, but not `refreshCache()`

### DIFF
--- a/app/Jobs/EsDeleteDocument.php
+++ b/app/Jobs/EsDeleteDocument.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace App\Jobs;
+
+use Illuminate\Bus\Queueable;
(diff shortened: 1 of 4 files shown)


nanaya:
`actual code : ceremonies` ratio here sure is great. Also, missing license.

PR #2203, opened 2018-01-11
Add hype and nomination count to beatmapset search results

### LINKED ISSUE #1950: Display a beatmap's hype on compact beatmap listing
I can check what sets a user is trying to rank by checking on their profile if there are sets with star priority. (Eg. this has 4 maps for rank https://puu.sh/yDWFh/38a8dcba62.png)
A map's hype level can be used in place to indicate what sets are for rank. This would be a helpful feature for BNs browsing rankable maps in the pending section (and actually gives a purpose to the hype train!!)
screen shot 2017-12-11 at 4 31 44 pm 3
screen shot 2017-12-11 at 4 31 44 pm 2

nekodex:
fixes #1950, also adds nomination counts

update: now also indexes nominations, so sorting by hype will also show beatmapsets that have nominations on top


resources/lang/en/beatmaps.php
```diff
@@ -123,6 +123,7 @@
         'explanation_guest' => 'Login and hype this beatmap to make it more visible for nomination and ranking!',
         'new_time' => "You'll get another hype at :new_time.",
         'remaining' => 'You have :remaining hype left.',
+        'required-text' => 'Hype: :current/:required',
```
nanaya: 🐍

app/Models/Beatmapset.php
```diff
@@ -878,6 +891,7 @@ public function disqualify(User $user, $comment)
             ]);
 
             $this->setApproved('pending', $user);
+            $this->refreshCache();
```
nanaya: part of `BeatmapsetEvent#save` instead?

app/Models/Beatmapset.php
```diff
@@ -590,6 +591,12 @@ public static function searchSortParamsES(array $params)
             'title' => 'title.raw',
         ];
 
+        // subsorts
+        $subsort = [];
+        if ($params['sort_field'] === 'hype') {
```
notbakaneko: directly adding to `$sortFields` is probably better if using `if` for comparison.

PR #2205, opened 2018-01-12
Show discussion for deleted maps

### LINKED ISSUE #1814: Discussions of deleted difficulties are inaccessible
If a mapper receives mods on a beatmap, and then deletes that specific difficulty and updates the mapset (by adding another difficulty, if that makes a difference), the discussion instance of the previous difficulty will disappear. In the kudosu history of the modders of that map, it will display the map as *[deleted beatmap]*. This makes sense for regular users, but for moderating, this is becomes a problem due to bns and qat being unable to see or take action on anything that happened in that discussion, which opens up room for exploit.

For example take a look at [this user's kudosu history](https://osu.ppy.sh/users/2124783#kudosu):
image
All of them link to https://osu.ppy.sh/beatmapsets/695465/discussion, along with the id for the respective discussion posts, but the posts themselves aren't there.

The discussion instances are obviously still there in the database, but they can't be viewed. Having some kind of soft-delete for discussions might be useful for this. So once the beatmap is deleted, the discussion would still available as a deleted beatmap tab (without download of course), but locked for both posting new things and voting. Since the beatmapsetid is stored for discussions, this seems like it'd be possible retroactively as well.

nanaya:
And prevent replying/posting on them.

Fixes #1814.


nekodex:
I'm thinking deleted beatmaps probably need a different styling.. Right now they blend in a bit too much with the other difficulties... maybe 50% opacity or something?

resources/lang/en/beatmaps.php
```diff
@@ -40,6 +40,7 @@
         'edited' => 'Last edited by :editor :update_time.',
         'kudosu_denied' => 'Denied from obtaining kudosu.',
         'message_placeholder' => 'Type here to post',
+        'message_placeholder_deleted_beatmap' => 'Posting on deleted difficulty is not allowed.',
```
nekodex: `"_xx_ is not allowed"` sounds a bit... blunt? Maybe something like `"This difficulty has been deleted so it may no longer be discussed."`

PR #2206, opened 2018-01-12
Latest laravels

nanaya:
And packages (just php). Needs more testing but at least the tests are green 🥗 

On unrelated note, it seems the tumblr package isn't used anymore.


notbakaneko:
seems safe so far ┐(ツ)┌

peppy:
let's do it

PR #2207, opened 2018-01-12
Fix beatmapset nomination sorting

nanaya:
Sort requires an array. Basic sorts seem to work. No idea about the nomination (apart of small set of data) :D


nanaya:
NOT NEEDED

PR #2216, opened 2018-01-15
Hide scores for deleted beatmapsets

nanaya:
Not sure if this is the correct place for this ಠ_ಠ


peppy:
I've fixed this at a submission and ranking level, so i don't think we need to do filtering at this level.

PR #2220, opened 2018-01-16
change datadog prefix config key to get old tag prefixes back.

notbakaneko:
laravel-datadog-helper prefixes all tags by default now.


peppy:
We may want to consider making the prefix universal to all datadog metrics (cc/ @nekodex) but this works as a fix for now.

PR #2218, opened 2018-01-16
composer require symfony/yaml

notbakaneko:
Required to use `Symfony\Component\Yaml\Yaml`;
Used to be included implicitly by phpunit, not anymore.

PR #2225, opened 2018-01-17
Increase minimum php version to reflect updated symfony package requirements

nekodex:
Looks like updated symfony package uses new syntax introduced in PHP 7.1

Probably due to #2218?


nanaya:
And then the bootstrap doesn't work anymore in ubuntu 16.04 because it doesn't have php7.1 :dancer:

PR #2226, opened 2018-01-17
Container travis

nanaya:
Now with 100% less sudo.


nanaya:
10 seconds is totally enough for elasticsearch to be up =]

notbakaneko:
can `wget -q --waitretry=1 --retry-connrefused -T 10 -O - http://127.0.0.1:9200` until elasticsearch is up :trollface:

PR #2228, opened 2018-01-17
Update store stock automatically on checkout

notbakaneko:
Seems to work; store side really needs cleaning/streamlining ಠ_ಠ


app/Http/Controllers/Store/Controller.php
```diff
@@ -34,4 +34,19 @@ protected function userCart()
...
+
+    protected function hasPendingCheckout()
+    {
+        $cart = $this->userCart();
+        if ($cart) {
+            return $cart->isProcessing();
+        }
+    }
+
+    protected function redirectIfPendingCheckout()
```
nanaya: not used? (probably something along the line "it doesn't work because actually using it looks stupid")

app/Models/Store/Order.php
```diff
@@ -215,11 +239,32 @@ public function getTotal()
...
+    {
+        return in_array($this->status, ['incart', 'processing'], true);
+    }
+
     public function isEmpty()
     {
         return !$this->items()->exists();
     }
 
+    public function isAwaitingPayment()
```
nanaya: not used?

app/Http/Controllers/Store/CartController.php
```diff
@@ -0,0 +1,63 @@
...
+            'store',
+        ]]);
+
+        return parent::__construct();
+    }
+
+    public function index()
+    {
+        $order = $this->userCart();
+        if ($order->isModifiable() === false) {
```
nanaya: use `hasPendingCheckout`?

app/Libraries/OrderCheckout.php
```diff
@@ -67,21 +93,44 @@ public function getCentiliPaymentLink()
...
         DB::connection('mysql-store')->transaction(function () {
             $order = $this->order->lockSelf();
 
             // cart should only be in:
-            // incart -> if user hits the callback first.
+            // processing -> if user hits the callback first.
             // paid -> if payment provider hits the callback first.
             // any other state should be considered invalid.
-            if ($order->status === 'incart') {
+            if ($order->status === 'processing') {
```
nanaya: `->isProcessing()`? Also in at least one other place.

app/Libraries/OrderCheckout.php
```diff
@@ -140,6 +211,26 @@ public static function complete($orderNumber)
...
+    /**
+     * Helper method for failing checkout with just the order number.
+     *
+     * @param string $orderNumber
+     * @return Order
+     */
+    public static function fail($orderNumber)
+    {
+        // select for update will lock the table if the row doesn't exist,
+        // so do a double select.
```
nanaya: But the second select isn't done here...? The comment can be confusing?

app/Libraries/OrderCheckout.php
```diff
@@ -140,6 +211,26 @@ public static function complete($orderNumber)
         return $order;
     }
 
+    /**
+     * Helper method for failing checkout with just the order number.
```
nanaya: Would something like `::for($orderNumber)` a bit more useful? `OrderCheckout::for('blah')->failCheckout()` etc.

app/Libraries/OrderCheckout.php
```diff
@@ -28,18 +28,41 @@
...
+     */
     private $order;
 
-    public function __construct(Order $order)
+    /**
+     * @var string|null
+     */
+    private $provider;
+
+    public function __construct(Order $order, string $provider = null)
```
nanaya: can use a default or `failCheckout` will put in `-failed` as transaction id if it's empty...

app/Libraries/OrderCheckout.php
notbakaneko: comment is out of date 三三三ヽ( ﾟ ヮﾟ)ﾉ

app/Http/Controllers/Store/CheckoutController.php
```diff
@@ -52,11 +54,21 @@ public function __construct()
         return parent::__construct();
     }
 
+    public function destroy($id = null)
+    {
+        // ignore dummy parameter for now, just release the current cart.
```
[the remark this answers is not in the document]
notbakaneko: Still waiting for the mythical laravel singular resource 🦄 
Could use `laravel-singular-resource-routes` or add separate `show`/`delete` routes 🤷‍♀️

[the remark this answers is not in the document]
nanaya: either sounds good for now 🦄

app/Libraries/OrderCheckout.php
notbakaneko: that was the idea

nanaya: looks weird as opposed to something like `unknown-failed` ┐(ツ)┌

PR #2229, opened 2018-01-17
ES updates for searches

nanaya:
Just remembered beatmapset has try/catch because the query may be invalid (unbalanced quotes etc).


app/helpers.php
```diff
@@ -116,6 +116,35 @@ function es_records($results, $class)
...
+
+    Log::debug($error);
+
+    // default return on failure
+    return [
+        'hits' => [
+            'hits' => [],
+            'total' => 0,
+        ],
+        'error' => $error ?? null,
```
notbakaneko: putting an exception object in a field that Elasticsearch uses to put a string message in seems like asking to be trolled

app/helpers.php
```diff
@@ -116,6 +116,38 @@ function es_records($results, $class)
     return $records;
 }
 
+function es_search($params)
+{
+    try {
+        return Es::search(array_merge_recursive([
+            'client' => [
+                'timeout' => 10,
+                'connect_timeout' => 5,
```
nekodex: are these seconds? if so, these are probably way too high

nekodex:
As a side note, I wonder if there's query builders or anything for Elasticsearch so we don't have to keep manually build these queries in every other model that needs search capability

app/helpers.php
notbakaneko: maybe `connect_timeout => 0.5`? I don't think we want `timeout` itself to be too low since that's the timeout for the entire operation whether it's completed or not.

nekodex: If Elasticsearch is taking _seconds_ to return results, there's probably something seriously wrong...?

Looking at our search metrics, nothing has even come close to even 1s (except user search which isn't Elasticsearch iirc) so maybe 1-2s is fine for timeout?

notbakaneko: and then it lags in dev, so this should probably be a config setting :v

PR #2241, opened 2018-01-18
Create userpage forum when seeding

LiquidPL:
While seeding forums, also adds a forum for userpages at the defauld id in config. Useful for when someone might want to test userpages, i guess.


nanaya:
The forum id is configurable through `USER_PAGE_FORUM_ID`.

PR #2242, opened 2018-01-19
Fix missing error message when editing userpage as a silenced user

### LINKED ISSUE #2186: When trying to edit profile while silenced, a blank text box appears
Silenced users on the new site can still not post, but when trying to edit their profile, a blank text box appears.
How would one go about fixing this issue? or is there a better place for posting this?

blanktextbox

LiquidPL:
~~Added an actual error message, so something will be shown instead of an empty box.~~ Disable middleware silence check so `OsuAuthorize` one with a proper error message can actually go through.

Fixes* #2186.

*partially

### DIFF
--- a/resources/lang/en/authorization.php
+++ b/resources/lang/en/authorization.php
@@ -148,8 +148,6 @@ return [
                 'locked' => 'User page is locked.',
                 'not_owner' => 'Can only edit own user page.',
                 'require_supporter_tag' => 'Supporter tag is required.',
-                'silenced' => "Can't edit your userpage while silenced.",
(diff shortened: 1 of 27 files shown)


nanaya:
This only partially fix the problem.

resources/lang/en/authorization.php
```diff
@@ -148,6 +148,8 @@
                 'locked' => 'User page is locked.',
                 'not_owner' => 'Can only edit own user page.',
                 'require_supporter_tag' => 'Supporter tag is required.',
+                'silenced' => "Can't edit your userpage while silenced.",
```
nanaya: looks like the default one is good enough?

PR #2248, opened 2018-01-19
Paginate beatmap pack list and use the same sort on all types.

### LINKED ISSUE #2234: Beatmap Packs should be listed reverse chronologically
and paginated

Currently theme/artist aren't chronological but it makes more sense to make everything use the same sort.

notbakaneko:
fixes #2234

### DIFF
--- a/resources/views/packs/index.blade.php
+++ b/resources/views/packs/index.blade.php
@@ -66,5 +66,6 @@
                 @endforeach
             </div>
         </div>
+        @include('objects._pagination', ['object' => $packs, 'modifier' => 'type'])
     </div>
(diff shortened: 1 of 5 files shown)


resources/views/packs/index.blade.php
```diff
@@ -66,5 +66,6 @@
                 @endforeach
             </div>
         </div>
+        @include('objects._pagination', ['object' => $packs, 'modifier' => 'type'])
```
nekodex: Pagination should probably be inside the above container instead of between container and the footer (pagination is always within the container elsewhere on the site)

PR #2251, opened 2018-01-19
add a type filter to es:index-documents

notbakaneko:
`es:index-users` removed.

Added `--types=` option to `es:index-documents`; use a comma separated list to specify multiple types:
`--types=users,beatmapsets,posts`

Unknown types are ignored.
Putting no types will default to `beatmapsets,posts,users`

Can cron different types on different schedules ( ﾟ ヮﾟ)


nanaya:
...merge back the `EsIndexCommand`? And maybe rename to `es:index` or something...

app/Console/Commands/EsIndexDocuments.php
```diff
@@ -40,5 +48,128 @@ class EsIndexDocuments extends EsIndexCommand
...
+        }
+    }
+
+    protected function starterMessage(array $oldIndices)
+    {
+        if ($this->inplace) {
+            $this->warn('Running in-place reindex.');
+            $confirmMessage = 'This will reindex in-place (schemas must match)';
+        } else {
+            $this->warn('Running index transfer.');
```
nanaya: index...what ಠ_ಠ

PR #2256, opened 2018-01-19
Add removing of avatars

### LINKED ISSUE #1992: Avatars could be resettable
A lot applications let you unset your own avatar if you wish to do so and just slap some default or no avatar on you, sounds like a relatively low effort thing if you have a thing that sets avatars already so why not?

Could be under user preferences right next to your avatar settings as its own control-thing

LiquidPL:
Fixes #1992.

### DIFF
--- a/app/Traits/UserAvatar.php
+++ b/app/Traits/UserAvatar.php
@@ -71,6 +71,6 @@ trait UserAvatar
             }
         }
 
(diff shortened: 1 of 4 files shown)


nanaya:
"No." -@arflyte

(just upload a blank image or something)

PR #2262, opened 2018-01-22
Elasticsearch 6 support

notbakaneko:
updates to use Elasticsearch 6:

- `type: string` is now `type: text`
- `index: not_analyzed` fields are now `type: keyword`
- `laravel-elasticsearch` package very unrequired

blocked until we actually setup elasticsearch 6 <_<


config/elasticsearch.php
```diff
@@ -3,7 +3,10 @@
 
 return [
     'hosts' => (array) env('ES_HOST', 'localhost:9200'),
-
-    'search_timeout' => get_float(env('ES_SEARCH_TIMEOUT')) ?? 5,
-    'search_connect_timeout' => get_float(env('ES_SEARCH_CONNECT_TIMEOUT')) ?? 0.5,
+    'connectionParams' => [
+        'client' => [
+            'timeout' => get_float(env('ES_SEARCH_TIMEOUT')) ?? 5,
+            'connect_timeout' => get_float(env('ES_SEARCH_CONNECT_TIMEOUT')) ?? 0.5,
```
nanaya: the original idea was to allow longer timeouts for indexing =S

nanaya: also the env name can use an update

notbakaneko: elasticsearch doesn't wait for indexing to run before responding; responses get returned immediately unless `?refresh=wait_for` is set

notbakaneko: although it still needs to wait for the parser and errors to output 🤔

nanaya: and can still use name update because it's initially named for search function (`ES_SEARCH_...`)

notbakaneko:
soon™️

PR #2270, opened 2018-01-23
Discussion navigation history

### LINKED ISSUE #1596: Navigation in modding system doesn't reflect in URL
- Clicking an issue in history then going back doesn't take you back to history.
- Selecting a tab doesn't update the URL (probably should).
- Selecting a filter mode doesn't update the URL (probably should).

nanaya:
Here be dragons.

May or may not work reliably. Fixes #1596.


nekodex:
u wot m8 :unamused:

nekodex:
Is there any reason why this isn't using prettyurls™?

nekodex:
Also, when visiting a url with a discussion hash in it (e.g. `discussion#/9`), the hash part gets removed from the current url when the page finishes loading

nekodex:
Looks like this breaks the compact display of discussion links:


(that link is to the same beatmapset, so should just be displayed as `#233`)

nekodex:
If you make a post and navigate away (e.g. to a linked discussion) and hit back... the page state seems to revert to the initial page load state or something (i.e. your new post appears to vanish).

When the ajax update poll kicks in, your post will reappear... feels kinda bad.

nekodex:
Back/forward don't seem to work properly when navigating between the `General` and `Timeline` tabs... url/history is rewritten but back/forward do nothing?

Issue #2283, opened 2018-01-26
Entering an invalid email address throws exception


nekodex:
Trying to change your email address (in account settings) to an invalid email address throws an exception.

The user gets a `Address in mailbox given [email] does not comply with RFC 2822, 3.6.2.` or similar error message, which is 'fine' (wording could use work); however the exception isn't caught and ends up in sentry every time this happens.

nekodex:
Uh, looking into this a bit more: it looks like the email address is persisted to DB (even when invalid). It then tries to send an email (to that invalid email address) informing the user it has been updated (which triggers the exception).

...SOOOOO, while it looks to the user as if the email failed to be changed to the invalid input... it actually has already persisted it to the DB. :1st_place_medal:

nanaya:
Fixed*

PR #2286, opened 2018-01-26
Actual email validator

nanaya:
Apparently the one used by laravel.


peppy:
Does this close #2283?

PR #2294, opened 2018-01-30
Nomination reset message

### LINKED ISSUE #2282: Disqualified .... ago ([object Object]). (Message when first nomination is reset)
image
image

https://osu.ppy.sh/beatmapsets/717058/discussion
this happened after a map with 1 nomination was reset to 0, triggered by a BN posting a problem. The set in question was never qualified to begin with.

nanaya:
Fixes #2282.


nekodex:
Probably shouldn't say "Disqualified by ..." when nominations are reset. Also, the nomination reset notice should probably be in addition to the disqualification message (instead of replacing it)

nekodex:
Apparently previous disqualification reasons are still relevant even after being renominated (so others can confirm stuff was fixed properly and whatnot idk)

...and if there's no previous disqualification, then don't show the disqualification message? >.> (i.e. just show reset one?)

nanaya:
All previous reasons are still visible from event history page

resources/lang/en/beatmaps.php
```diff
@@ -130,18 +130,19 @@
...
-        'nominated-by' => 'nominated by :users',
-        'nominate-confirm' => 'Nominate this beatmap?',
+        'nominate_confirm' => 'Nominate this beatmap?',
+        'nominated_by' => 'nominated by :users',
         'qualified' => 'Estimated to be ranked :date, if no issues are found.',
-        'qualified-soon' => 'Estimated to be ranked soon, if no issues are found.',
-        'reset-confirm' => 'Are you sure? Posting a new problem will reset nominations.',
+        'qualified_soon' => 'Estimated to be ranked soon, if no issues are found.',
         'required_text' => 'Nominations: :current/:required',
+        'reset_at' => 'Reset :time_ago by discussion :discussion.',
```
nekodex: probably worth being more explicit here and saying `Nominations reset :time_ago by...`

resources/assets/coffee/react/beatmap-discussions/nominations.coffee
```diff
@@ -102,6 +103,12 @@ class BeatmapDiscussions.Nominations extends React.PureComponent
       else if event.type == 'nominate'
         nominators.push(@props.users[event.user_id])
 
+    if reset? && reset.type == 'nomination_reset'
```
nekodex: Do type check in the transformer instead of here?

PR #2296, opened 2018-01-30
Ignore kudosu on deleted beatmaps

nanaya:
...maybe?


nekodex:
we probably need something to trigger recalculation of kudosu when beatmaps are deleted to make this useful?

nanaya:
I'm thinking somewhere in legacy interop controller thingy as part of refresh beatmapset cache.

Xinnoh:
Don't think this should be added, people can still make valid suggestions while a map is uploaded. Just because it's deleted doesn't mean they put in effort to mod it

Mirruuu:
I agree with Sinnoh, even if the difficulty is deleted the input was still valuable. It shouldn't be rendered moot if the difference is deleted.

nekodex:
After some discussion, we've decided this is probably not the best approach to take. Gonna close this in favour of #2308 to handle kudosu on deleted beatmaps for the time being.

PR #2300, opened 2018-01-30
topic_last_post_time should be set to null instead of 0

notbakaneko:
due to change from #2280


app/Models/Forum/Forum.php
```diff
@@ -38,6 +38,16 @@ class Forum extends Model
         'enable_sigs' => 'boolean',
     ];
 
+    public function getForumLastPostTimeAttribute($value)
```
nanaya: maybe grouped with other set/get attributes?

PR #2302, opened 2018-01-31
Fix scoreboard sorting

### LINKED ISSUE #2082: Beatmap leaderboards seem to display tied scores in random order
image
image)  

The old website and the ingame leaderboard displays tied scores in the order they were achieved, which makes sense. On the new website they don't seem to be listed in any particular order as far as I can tell.

nekodex:
fixes #2082


nanaya:
Dem foxes 🦊

nekodex:
Apparently using the `hidden` column is not a good idea until index is added/updated, so removing for now.

PR #2309, opened 2018-02-01
Better vote styling

nanaya:
Limit maximum length and add vertical scroll when in mobile.


notbakaneko:
This seems odd with mobile layout

image

nanaya:
As intended™

PR #2310, opened 2018-02-01
Hide unnecessary stats from bot userpages

### LINKED ISSUE #2268: Bot user pages should not show any stats
Similar to the [old site](https://osu.ppy.sh/u/3), bots (group_id 29) should have any stats sections should be omitted.

LiquidPL:
w00t
image

Fixes #2268.


app/Models/User.php
```diff
@@ -639,6 +639,11 @@ public function isRegistered()
         return $this->isGroup(UserGroup::GROUPS['default']);
     }
 
+    public function isBot()
+    {
+        return $this->isGroup(UserGroup::GROUPS['bot']);
```
nanaya: Because some users in bot group are actual players as well, change this to check `$this->group_id` instead.

resources/assets/coffee/react/profile-page/header-extra.coffee
```diff
@@ -210,6 +210,8 @@ class ProfilePage.HeaderExtra extends React.Component
 
 
   rankChartHover: (_e, {data} = {}) =>
+    return if @props.user.isBot
```
nanaya: not needed?

PR #2311, opened 2018-02-01
Show exact join date on hover

### LINKED ISSUE #2146: Hovering over user join date doesn't give exact date they joined at
It does on the old site, but nothing happens on the new site

LiquidPL:
Fixes #2146.


resources/assets/coffee/osu_common.coffee
```diff
@@ -134,11 +134,13 @@
...
 
-  timeago: (time) ->
+  timeago: (time, text = null) ->
+    text = time if text == null
+
     el = document.createElement('time')
     el.classList.add 'timeago'
     el.setAttribute 'datetime', time
-    el.textContent = time
+    el.textContent = text
```
nanaya: this doesn't do anything because `$.timeago` replaces it.

resources/assets/coffee/react/profile-page/header-extra.coffee
```diff
@@ -21,8 +21,8 @@ el = React.createElement
 
 bn = 'profile-header-extra'
 
-rowValue = (value) ->
-  "<strong>#{value}</strong>"
+rowValue = (value, className = '', title = '') ->
+  "<strong class='#{className}' title='#{title}'>#{value}</strong>"
```
nekodex: this is gonna result in `class='' title=''` when `class` and `title` aren't given

nekodex:
Also I noticed this is using a date+time popup, but these are join _dates_ (i.e. no time)... so you get:

LiquidPL:
I'd argue that the old website displays both date and time, also when displaying only the date it looks kinda... repetitive? 🤔 

image

LiquidPL:
(actually it displayed no date for `since the beginning` accounts, but that's a separate topic i think)

nekodex:
The alternative is to change `UserTransformer` to not convert from datetime to date if we want to display the time as well... although not sure how useful that extra detail is? :man_shrugging:

resources/assets/coffee/react/profile-page/header-extra.coffee
```diff
@@ -97,17 +97,18 @@ class ProfilePage.HeaderExtra extends React.Component
 
           div className: "#{bn}__rows",
             if moment(@props.user.join_date).isBefore moment('2008-01-01')
-              div
-                className: "#{bn}__row"
-                title: moment(@props.user.join_date).format(osu.trans('common.datetime.year_month.moment')),
+              time
```
nekodex: This `time` should be inside the `#{bn}__row` `div` instead of replacing it, otherwise the alignment of the tooltip ends up being off:

nanaya:
not sure if worth all this changes instead of just changing moment format and adding right title.

nekodex:
Hey man, sorry to do this, but we're gonna close this PR...

While reviewing, we noticed most of the work for this was in styling date/time tooltips differently from the default tooltips and after some discussion with @arflyte (about why they were even styled differently), we decided to instead unify the tooltip styles.

This unfortunately makes most of this pull request redundant (i.e. after the tooltip styling is fixed/unified, a `title: [date]` is all that will be required)

Kind of a crappy outcome for what should have been a simple fix, but this was our bad. Sorry!

PR #2312, opened 2018-02-01
Basic locale check command

nanaya:
Stuff ヽ( ﾟ ヮﾟ)ﾉ (also in travis)

```
artisan locale:check
```

### DIFF
--- a/app/Console/Commands/LocaleCheck.php
+++ b/app/Console/Commands/LocaleCheck.php
@@ -87,7 +87,7 @@ class LocaleCheck extends Command
 
     public function basePath($locale)
     {
-        return resource_path().'/lang/'.$locale;
+        return resource_path("lang/{$locale}");
     }
 
     public function checkExtraHeads($locale, $baseHeads, $targetHeads)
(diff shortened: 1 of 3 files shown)


app/Console/Commands/LocaleCheck.php
```diff
@@ -0,0 +1,198 @@
...
+            $this->info('');
+
+            $this->info("End checking locale {$locale}");
+            $this->info('======');
+        }
+    }
+
+    public function basePath($locale)
+    {
+        return resource_path().'/lang/'.$locale;
```
notbakaneko: `resource_path()` accepts subpath as parameter

app/Console/Commands/LocaleCheck.php
```diff
@@ -0,0 +1,198 @@
...
+    public function heads($locale, $files)
+    {
+        $basePath = $this->basePath($locale);
+        $trimStart = strlen($basePath) + 1; // +1 is trailing slash
+        $trimEnd = -strlen('.php');
+
+        $heads = [];
+
+        foreach ($files as $file) {
+            $heads[] = str_replace('/', '.', substr($file, $trimStart, $trimEnd));
```
notbakaneko: doesn't this mean keys in subdirectories won't get picked up ?_?

PR #2313, opened 2018-02-01
Allow newline between various tags

### LINKED ISSUE #268: BBCode not always working on new website
687474703a2f2f692e696d6775722e636f6d2f65337876395a342e706e67

Things like [notice] and [color] work most of the time, but not always when posting from new website.

HoutarouOreki:
Should fix #268 ~~*for color tags*~~.

~~I'll allow myself a question. How do I put regular expressions in code like this?~~

```PHP
$text = str_replace("[i:{$this->uid}]", '<em>', $text);
```
Never mind, I think I figured it out myself.

Still I'm not really sure what I'm doing, just tried following the patterns. Added a function parseHeading because headings were passed in parseInlineSimple, which I made to pass multiline things, and headings shouldn't be multiline.

Also shouldn't I have renamed the "inline" function, since it now parses multiline? 🙅‍♂️🔫


HoutarouOreki:
```
1) BBCodeForDBTest::testAll
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'[color=white:1]Colored text! (name)[/color:1]
+'[color=white:1]Colored text! (name)[):1]
 
-[color=#ffffff:1]Colored text! (code)[/color:1]
+[color=#ffffff:1]Colored text! (code)[):1]
 
-[color=#FFFFFF:1]Colored text! (uppercased hex)[/color:1]
+[color=#FFFFFF:1]Colored text! (uppercased hex)[):1]
/home/travis/build/ppy/osu-web/tests/Libraries/BBCodeForDBTest.php:37
/home/travis/build/ppy/osu-web/vendor/phpunit/phpunit/phpunit:53
FAILURES!
Tests: 114, Assertions: 275, Failures: 1.
```
Uhh, I guess for some reason there's `)` instead of `/color`, I'm looking into this.

app/Libraries/BBCodeForDB.php
```diff
@@ -155,6 +155,17 @@ public function parseInlineSimple($text)
         return $text;
     }
 
+    public function parseHeading($text)
```
notbakaneko: https://github.com/ppy/osu-web/pull/2313/files#diff-8f78753fe51dad48708f01cc0c0723efL147 
these changes seem unnecessary.

HoutarouOreki: Oops, I forgot to change the heading function to be single-line. That's why I made it - because headings aren't supposed to be spanned across multiple lines. Fixing.

As for the multi-line modifier, I wanted to do that but I don't see where I should put it, I don't see where are the forward slashes at. Normally I have like `/something/modifiers` kind of patterns in JS, but here it's `#something#` and it throws me off. Care to hint me?

notbakaneko: flags goes in the same place, after the pattern delimiter:
```
/pattern/s
#pattern#s
~pattern~s
```
are equivalent.

`/` is not being used here because of `[/tag`, otherwise the `/` needs to be escaped.

HoutarouOreki:
Okay, I guess it's fine now.

app/Libraries/BBCodeForDB.php
HoutarouOreki: Okay, I guess it's all fine now?

notbakaneko:
you don't have to merge `master` every other commit ಠ_ಠ

HoutarouOreki:
🤔 

Well GitHub throws a triangle at me with an exclamation mark every time it's "out-of-date" and a big button to update it, so I thought that was the right ting to do.

PR #2317, opened 2018-02-02
Cleanups

nanaya:
Need a bit more check.


nekodex:
Looks like this changes some of the JSON responses (User, etc), so probably should check with @peppy before merging because this will probably break lazer's online functionality...

peppy:
Let's merge this; I'll push out a fix for lazer today.

PR #2334, opened 2018-02-07
Fix elements in post preview being uninteractable

### LINKED ISSUE #2333: Unable to reveal box/spoilerbox's content on forum post preview
they just dont work

LiquidPL:
Fixes #2333.

Not sure if it should be this way, but it should work.


nanaya:
It was originally because some elements aren't generated with correct stuff (thread link for example). Probably can be removed if that link is replaced with dummy link and remove all overlay related things as well.

resources/views/forum/topics/_post.blade.php
```diff
@@ -43,7 +43,7 @@ class="forum-post__stripe"
 
         <div class="forum-post__body">
             <div class="forum-post__content forum-post__content--header">
-                <a class="js-post-url link link--gray" href="{{ route('forum.posts.show', $post->post_id) }}">
+                <a class="js-post-url link link--gray" href="{{ $options['preview'] ? '#' : route('forum.posts.show', $post->post_id) }}">
```
nanaya: checking `$post->exists` is probably better.

PR #2344, opened 2018-02-08
Add a link to avatar edit on the avatar in user profile

### LINKED ISSUE #2337: clicking on profile picture doesn't redirect to change profile picture
on the old site, you could go to your user profile and click, and it would redirect to change your profile picture. this feature is absent on the new site.

LiquidPL:
Fixes #2337.

### DIFF
--- a/resources/assets/coffee/react/_components/user-avatar.coffee
+++ b/resources/assets/coffee/react/_components/user-avatar.coffee
@@ -16,7 +16,7 @@
 #    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
 ###
 
-{a, div} = ReactDOMFactories
+{div} = ReactDOMFactories
 
 bn = 'avatar'
 
@@ -28,19 +28,10 @@ bn = 'avatar'
 
   className = "#{bn} #{modifiers}"
 
-  avatar =
+  if props.user.id?
     div
       className: className
       style:
         backgroundImage: "url('#{props.user.avatar_url}')"
-
-  if props.user.id?
-    if props.user.id == currentUser.id && _.includes modifiers, 'profile'
-      a
-        href: laroute.route('account.edit') + '#avatar-edit'
-        title: osu.trans 'users.show.change_avatar'
-        avatar
-    else
-      avatar
   else
     div className: "#{className} #{bn}--guest"
--- a/resources/assets/coffee/react/_components/user-avatar.coffee
+++ b/resources/assets/coffee/react/_components/user-avatar.coffee
@@ -16,7 +16,7 @@
 #    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
 ###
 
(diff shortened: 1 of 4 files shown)


resources/assets/coffee/react/_components/user-avatar.coffee
```diff
@@ -28,10 +28,19 @@ bn = 'avatar'
...
-  if props.user.id?
+  avatar =
     div
       className: className
       style:
         backgroundImage: "url('#{props.user.avatar_url}')"
+
+  if props.user.id?
+    if props.user.id == currentUser.id && _.includes modifiers, 'profile'
+      a
```
nanaya: no need to wrap the avatar, just replace its element.

resources/assets/coffee/react/_components/user-avatar.coffee
```diff
@@ -28,10 +28,19 @@ bn = 'avatar'
...
+        href: laroute.route('account.edit') + '#avatar-edit'
```
nanaya: `#avatar`. The url already says `/edit`. Also use string interpolation.

resources/assets/coffee/react/_components/user-avatar.coffee
```diff
@@ -28,10 +28,19 @@ bn = 'avatar'
...
 
-  if props.user.id?
+  avatar =
     div
       className: className
       style:
         backgroundImage: "url('#{props.user.avatar_url}')"
+
+  if props.user.id?
+    if props.user.id == currentUser.id && _.includes modifiers, 'profile'
```
nanaya: `props.modifiers`. `modifiers` is plain string (still works but not quite as intended).

resources/assets/coffee/react/_components/user-avatar.coffee
```diff
@@ -28,10 +28,19 @@ bn = 'avatar'
 
   className = "#{bn} #{modifiers}"
 
-  if props.user.id?
+  avatar =
```
nanaya: ~~this should be inside `if props.user.id?`~~

Looking again, this whole thing can (probably) be rewritten into

```coffeescript
## might as well replace with PureComponent while at it.
modifiers = ...

element = div
props = className: ...

if @props.user.id?
  props.style = backgroundImage: "..."

  if ... currentUser ...
    element = a
    props.title = ...
    props.href = ...
else
  props.className += ...

element props
```

resources/assets/coffee/react/_components/user-avatar.coffee
nekodex: I'd recommend wrapping the avatar in `ProfilePage.HeaderInfo` instead of modifying `UserAvatar` like this

nanaya: Or wrap in headerinfo as suggested elsewhere.

PR #2345, opened 2018-02-09
Scoreboard friend/self highlighting

### LINKED ISSUE #2321: Highlight friends on map leaderboards
The old website highlights users you have friended when viewing a map page's leaderboard section, drawing attention to them.
image

The new website does not
image

nekodex:
Adds highlighting of your own and your friends' scores on beatmap scoreboards.

(Also prevents double headers happening if you're at the top of a scoreboard)

fixes #2321

### DIFF
--- a/resources/assets/coffee/osu_common.coffee
+++ b/resources/assets/coffee/osu_common.coffee
@@ -30,7 +30,7 @@
 
 
   currentUserIsFriendsWith: (user_id) ->
-    currentUser && _.find currentUser.friends, (o) -> o.target_id == user_id
+    _.find currentUser.friends, target_id: user_id
 
 
   executeAction: (element) =>
--- a/resources/assets/coffee/osu_common.coffee
+++ b/resources/assets/coffee/osu_common.coffee
@@ -29,6 +29,10 @@
     (body.scrollHeight - body.scrollTop) - body.clientHeight
 
 
+  currentUserIsFriendsWith: (user_id) ->
+    currentUser && _.find currentUser.friends, (o) -> o.target_id == user_id
+
+
   executeAction: (element) =>
     if !element?
       osu.reloadPage()
--- a/resources/assets/less/bem/beatmap-scoreboard-table.less
+++ b/resources/assets/less/bem/beatmap-scoreboard-table.less
@@ -114,6 +114,18 @@
         background-color: #ccc;
       }
     }
+
+    &--friend {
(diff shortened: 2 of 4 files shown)


resources/assets/coffee/osu_common.coffee
```diff
@@ -29,6 +29,10 @@
     (body.scrollHeight - body.scrollTop) - body.clientHeight
 
 
+  currentUserIsFriendsWith: (user_id) ->
+    currentUser && _.find currentUser.friends, (o) -> o.target_id == user_id
```
nanaya: `currentUser` is always defined, I believe? Empty object for anonymous. Also can pass `target_id: user_id` instead of function.

resources/assets/less/bem/beatmap-scoreboard-table.less
```diff
@@ -20,10 +20,27 @@
   @_top: beatmap-scoreboard-table;
   @row-height: 1.8em;
 
-
   width: 100%;
   overflow-x: auto;
 
+  .row-colour(@colour) {
+    td {
```
nanaya: not sure how much it helps but `> td` is probably slightly faster (and doesn't randomly recolor the nonexistent child tables).

resources/assets/less/bem/beatmap-scoreboard-table.less
```diff
@@ -20,10 +20,27 @@
...
+      background-color: lighten(@colour, 20%);
+    }
+
+    &:nth-child(odd) {
+      td {
+        background-color: lighten(@colour, 20%);
```
nanaya: different color in case a bunch of friends ranked consecutively? (in which case this function can be used for default color as well)

nekodex: It looks weird if there are different colours and stuff isn't consecutive - you get stuff like this:


which just looks wrong imo

resources/assets/less/bem/beatmap-scoreboard-table.less
```diff
@@ -109,11 +119,26 @@
       padding-right: 10px;
     }
 
+    // styles applied to td instead of tr 'cuz firefox is a poop (can't do rounded corners on tr)
+    &:nth-child(odd) {
+      td {
```
nanaya: more specific here as well...? (and the remaining ones)

PR #2346, opened 2018-02-09
Display nominations in order of nomination

### LINKED ISSUE #2084: Nominators are listed in wrong order
a
^Order in which the map was nominated according to history

b
v2 discussion page shows them in the opposite order which doesn't seem intended, listing the first nominator first would be a better approach, as already mentioned in discord.

This happens for all maps that get nominated in v2.

nekodex:
fixes #2084

### DIFF
--- a/resources/assets/coffee/react/beatmap-discussions/nominations.coffee
+++ b/resources/assets/coffee/react/beatmap-discussions/nominations.coffee
@@ -233,7 +233,7 @@ class BeatmapDiscussions.Nominations extends React.PureComponent
                 className: "#{bn}__note #{bn}__note--nominators"
                 dangerouslySetInnerHTML:
                   __html: osu.trans 'beatmaps.nominations.nominated_by',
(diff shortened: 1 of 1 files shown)


resources/assets/coffee/react/beatmap-discussions/nominations.coffee
```diff
@@ -233,7 +233,7 @@ class BeatmapDiscussions.Nominations extends React.PureComponent
                 className: "#{bn}__note #{bn}__note--nominators"
                 dangerouslySetInnerHTML:
                   __html: osu.trans 'beatmaps.nominations.nominated_by',
-                    users: nominators.map (user) ->
+                    users: nominators.reverse().map (user) ->
```
nanaya: I wonder if it's better to unshift instead of push when generating the array. A bit clearer about the order or something.

nanaya:
also more like "display nominators"?

PR #2355, opened 2018-02-12
Docker updates

LiquidPL:
* don't cache config and routes when in debug env (`APP_DEBUG` is not `false`)
* ~~upgrade PHP to 7.2~~ not yet, pending library updates
* switch yarn to a self-installed setup (as `npm install yarn` is apparently deprecated)
* add redis and elasticsearch

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,12 +28,13 @@ There are a few different options to get started:
 - Use Docker:
   - First, install [Docker](https://www.docker.com/community-edition) and [Docker Compose](https://docs.docker.com/compose/install/), then run `docker-compose up` in the main directory.
   - Due to the nature of Docker (a container is killed when the command running in it finishes), the Yarn container will be run in watch mode.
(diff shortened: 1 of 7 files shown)


docker/yarn/setup.sh
```diff
@@ -1,8 +1,7 @@
 #!/bin/bash
 
-userdel node
+# workaround for macos default groupid (20)
+groupmod -g 24 dialout
```
nanaya: maybe just change to 4 digit number instead of trying to pinpoint in between numbers (or call `groupadd` with `-o` option).

README.md
```diff
@@ -8,7 +8,7 @@ The future face of osu!.
 Requirements
 ------------
 
-- A PHP 7.1+ (or equivalent) environment
+- A PHP 7.2 (or equivalent) environment
```
nanaya: What I'm saying is it should be `7.1` instead of `7.1+` because on of the libraries is broken in 7.2 ಠ_ಠ

PR #2379, opened 2018-02-14
Add 'show more' for recent activity on user profiles

### LINKED ISSUE #2365: "Recent" section of the profile is missing the "Show me more" option
image
old profile displays 4 by default with a show me more option, and 15 when you click the "show me more" link.

image
 new profiles display 5 by default, but does not have any "show me more" option. 5 is a pretty limited number of items if there is no way to see more than that :/

nekodex:
fixes #2365

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -268,7 +268,7 @@ class UsersController extends Controller
                 'user_achievements',
                 'follower_count',
                 'page',
-                'recent_activities',
+                'recent_activity',
                 'ranked_and_approved_beatmapset_count',
                 'unranked_beatmapset_count',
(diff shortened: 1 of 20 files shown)


app/Http/Controllers/UsersController.php
```diff
@@ -54,7 +54,7 @@ public function __construct()
 
             return $next($request);
         }, [
-            'only' => ['scores', 'beatmapsets', 'kudosu'],
+            'only' => ['scores', 'beatmapsets', 'kudosu', 'recentActivity'],
```
nanaya: `recentActivities`?

nekodex: actually I think it should probably be `activity` instead of `activities` everywhere 😔

PR #2389, opened 2018-02-14
Limit voting based on current beatmapset state

### LINKED ISSUE #2350: Disable upvoting on ranked beatmapsets
Currently this can be used to give out kds to modders without even the mapper knowing after a mapset has been ranked. 
It would be unreasonable to expect moderators to check back on ranked sets to make sure that upvoting is not being misused on these sets, when currently the expectation is such moderation (if any) would be done prior to / during qualification.

nanaya:
Fixes #2350.

Backend only. Frontend side will be done later... maybe...

### DIFF
--- a/resources/lang/en/authorization.php
+++ b/resources/lang/en/authorization.php
@@ -34,6 +34,7 @@ return [
         'vote' => [
             'limit_exceeded' => 'Please wait a while before casting more votes',
             'owner' => 'Can not vote own discussion!',
+            'wrong_beatmapset_state' => 'Can only vote discussions of pending beatmaps.',
         ],
     ],
 
--- a/resources/lang/en/authorization.php
+++ b/resources/lang/en/authorization.php
@@ -33,8 +33,8 @@ return [
 
         'vote' => [
             'limit_exceeded' => 'Please wait a while before casting more votes',
-            'owner' => 'Can not vote own discussion!',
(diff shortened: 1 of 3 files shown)


app/Libraries/BeatmapsetDiscussionVoteService.php
```diff
@@ -0,0 +1,94 @@
...
+
+    public function isValid()
+    {
+        $this->validationErrors()->reset();
+
+        if ($this->discussion->isLocked()) {
+            $this->validationErrors()->add('base', '.locked');
+        }
+
+        if (!in_array($this->discussion->beatmapset->status(), static::VOTABLE_BEATMAPSET_STATES, true)) {
```
nekodex: Apparently graveyard shouldn't be a votable state unless the voter is BN or above 🤷‍♂️

resources/lang/en/authorization.php
```diff
@@ -34,6 +34,7 @@
         'vote' => [
             'limit_exceeded' => 'Please wait a while before casting more votes',
             'owner' => 'Can not vote own discussion!',
+            'wrong_beatmapset_state' => 'Can only vote discussions of pending beatmaps.',
```
nekodex: nice english >.>
```php
'owner' => "Can't vote on your own discussion.",
'wrong_beatmapset_state' => 'Can only vote on discussions of pending beatmaps.',
```

PR #2406, opened 2018-02-15
Differentiate hover and active/focus state buttons

### LINKED ISSUE #2403: Download osu! button buggy
The download button hover animation doesn't refresh after clicking it. Clicking anywhere after resets it. It also is nice to add the push and glow effect like the one in the home page.
https://streamable.com/uwgcp

nanaya:
Fixes #2403.


nekodex:
This might be a browser thing but it doesn't seem fixed in chrome/firefox? It just seems to stay in the focus/active state and subsequent hovers do nothing... 😕

PR #2417, opened 2018-02-15
Adjust the year in the licence

TPGPL:


nanaya:
We're (still) discussing this internally and the commit needs to be updated anyway so I'm closing this for now.

PR #2442, opened 2018-02-16
Add recommended difficulty and converted beatmap filters to beatmap search panel

### LINKED ISSUE #2366: Recommended star / converted beatmaps toggles missing
These are both toggles which filter maps further, missing on the new listing.

notbakaneko:
fixes #2366

doesn't handle the difficulty of converted maps yet, that will be done separately.


resources/assets/coffee/_classes/beatmapset-filter.coffee
```diff
@@ -31,6 +31,7 @@ class @BeatmapsetFilter
 
 
   @charToKey:
+    c: 'general'
```
notbakaneko: g was taken ( ﾟ ヮﾟ)

app/Models/Beatmapset.php
```diff
@@ -524,8 +531,28 @@ public static function searchES(array $params = [])
                 break;
         }
 
+        // recommended difficulty
+        if ($params['recommended'] && $params['user'] !== null) {
+            // TODO: index convert difficulties and handle them.
+            $mode = Beatmap::modeStr($param['mode'] ?? Beatmap::MODES['osu']);
```
nanaya: more like `$params`. Also is it converting back and forth between string and int ಠ_ಠ

notbakaneko: sure is :trollface: 
Either that or something should be done about `User::statistics`

PR #2447, opened 2018-02-16
Rename the "news" tab to "dashboard"

### LINKED ISSUE #2393: Call the "news" tab on the homepage "main" instead.
image

there's more on this page than just the news, and it can be confusing when there's another totally different page for browsing news posts. It took me a while to realize that this "news" tab and the "news" page that you can get to through the top drop down menu are _two different pages_

TPGPL:
Resolves #2393


peppy:
travis not happy

PR #2501, opened 2018-02-20
Remove unnecessary "Edit" word on settings

Joehuu:
Settings are a place to edit stuff anyway. First PR. Correct me if wrong. Fixed coding with apostrophes instead of quotations. Hope that doesn't break. I know that Google Translate is not accurate. Check translations if wrong.


nekodex:
As kj415j45 said, unless you actually speak all those languages, their translation strings should be removed instead of guessing/machine translating.

That way, the translators can easily find the missing translation keys to update.

Joehuu:
I'll just leave it as is.

peppy:
That seems like a counter-productive outcome..

PR #2505, opened 2018-02-20
Users controller refactor

### LINKED ISSUE #2365: "Recent" section of the profile is missing the "Show me more" option

nanaya:
Should fix #2365.

And removes `recent_activity` from user json. It's recently renamed so I assume it's not used api-side.


app/Http/Controllers/UsersController.php
```diff
@@ -213,32 +193,49 @@ public function store()
...
-                return $this->mostPlayedBeatmapsets($this->user, $this->perPage, $this->offset);
+        static $mapping = [
+            'most_played' => 'beatmapPlaycounts',
+            'favourite' => 'favouriteBeatmapsets',
+            'ranked_and_approved' => 'rankedAndApprovedBeatmapsets',
+            'unranked' => 'unrankedBeatmapsets',
+            'graveyard' => 'graveyardBeatmapsets',
+        ];
+
+        $page = $mapping[$type] ?? abort(404);
```
nanaya: not sure if best idea or worst :D

app/Http/Controllers/UsersController.php
```diff
@@ -400,130 +385,82 @@ private function sanitizedLimitParam()
...
-        $scores = $user->scoresFirst($mode, true)
-            ->orderBy('score_id', 'desc')
-            ->with('beatmap', 'beatmap.beatmapset')
-            ->limit($perPage)
-            ->offset($offset)
-            ->get();
-
-        ScoreBestModel::fillInPosition($scores);
+        switch ($page) {
+            case 'recentActivity':
```
nekodex: maybe worth sorting alphabetically or something?

nekodex: also not sure how I feel about this all being in a giant case statement 😕

PR #2526, opened 2018-02-22
Add Mappers Guild Pack I web badge assets

Ephemeralis:


peppy:
I'm not sure we want to do this? My policy has always been that medals should only be made for things which are achievable by all users.

peppy:
Oh, this is a beatmap pack. Should probably be titled "Mapper's Guild Pack 1" or something more descriptive.

Xinnoh:
Anyone will be able to join the mappers' guild in future as long as they show interest in featured artists and are ok at mapping. The group is only limited in size for now to ensure thing go smoothly.

PR #2531, opened 2018-02-22
Improved page title

### LINKED ISSUE #2108: Change the displayed message on the tab header
firefox_2017-12-26_18-02-13

The left one (old forum) is better than the right one (new forum) imo. it allows people to have an idea about what's inside. this is crucial especially on smaller tab width

nanaya:
Flipped stuff around. Additionally added "| osu!" at the end. Inputs welcome.

Fixes #2108.

### DIFF
--- a/resources/lang/pl/layout.php
+++ b/resources/lang/pl/layout.php
@@ -34,7 +34,6 @@ return [
             'getDownload' => 'pobierz',
             'getIcons' => 'ikony',
             'groups-show' => 'grupy',
-            'index' => 'osu!',
             'legal-show' => 'informacje',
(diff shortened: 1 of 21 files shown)


resources/lang/pl/layout.php
```diff
@@ -34,7 +34,6 @@
             'getDownload' => 'pobierz',
             'getIcons' => 'ikony',
             'groups-show' => 'grupy',
-            'index' => 'osu!',
```
TPGPL: 'index' => 'przegląd',

peppy:
Let's go with this and see what people think. Looks good to me.

PR #2557, opened 2018-02-26
Proper delete validation

### LINKED ISSUE #2550: Deleting a modding post causes an error message
When clicking delete, the post is successfully deleted, but this error message shows up and the post remains displayed locally. Refreshing the page causes it to disappear as intended.

nanaya:
Implement actual validation instead of using `null` check. And do some refactors on handling `ModelNotSaved`.

Can be simplified further by doing it at global exception handler for general cases (all of them at the moment). Maybe later (and not sure if good idea yet).

Fixes #2550.

### DIFF
--- a/app/Http/Controllers/Forum/TopicsController.php
+++ b/app/Http/Controllers/Forum/TopicsController.php
@@ -335,8 +335,8 @@ class TopicsController extends Controller
 
         try {
             $topic = Topic::createNew($forum, $params, $poll ?? null);
-        } catch (ModelNotSavedException $_e) {
-            abort(422);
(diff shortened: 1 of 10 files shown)


app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -335,8 +335,8 @@ public function store(HttpRequest $request)
 
         try {
             $topic = Topic::createNew($forum, $params, $poll ?? null);
```
notbakaneko: `Topic::addPost` called through `Topic::createNew` doesn't explode when failing to save, so maybe should do something about it while changing all this?

PR #2567, opened 2018-02-27
Recommended Difficulty Calculation

Xinnoh:
Will this even work who knows
Hint: it didn't

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -1259,23 +1259,22 @@ class User extends Model implements AuthenticatableContract, Messageable
      * @return float
      */
     public function recommendedStarDifficulty(string $mode)
-    {
-        $stats = $this->statistics($mode);
-        if ($stats) {
-            if ($mode == osu){
-                return pow($stats->rank_score, 0.4) * 0.195;
-            }elseif ($mode == taiko){
-                return pow($stats->rank_score, 0.4) * 0.165;
-            }elseif ($mode == fruits){
-                return pow($stats->rank_score, 0.4) * 0.155;
-            }elseif ($mode == mania){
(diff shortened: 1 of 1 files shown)


nanaya:
Missing quotes aside, where did the numbers come from?

Xinnoh:
I asked a few users from various ranks in other modes whether they felt the recommended difficulty was too high, and what they'd prefer to be recommended. Averaged their suggestions to find what the multiplier should be changed to.

nanaya:
Can you post the data here? At least for me the suggested osu!mania difficulty rating is too low. Lowering it further wouldn't help...

(and meanwhile the same applies to osu! mode. It's suggesting 1.79 while I usually play 2.7-3)

Xinnoh:
I collected some [quantitative data here](https://docs.google.com/spreadsheets/d/1OqzMExjhZJNPkY019fSs39cITVFl1eAvOsiSnKHYBf8/edit?usp=sharing) to get a clearer vision on this.
For players with 5000+ pp, the formula is pretty accurate, but the suggested difficulty becomes way too low for lower rank players.
There should be a check that applies a different multiplier based on the player's ability, which I don't know how to do soooo...

nanaya:
Missing the mode...

Xinnoh:
I was working on this for a while, but I need to close it for now due to the uncertainty of how performance points are going to change. I'll reopen it once things have settled down.

peppy:
I'm not sure this is... related?

Xinnoh:
Data I've collected is based off 
1. How much pp each player has
2. The average of their top 5 plays
Suggested difficulty is 100% based off how much pp a player has. If this data is going to change then the suggested difficulty is going to change. I don't know to what extent it will change so I'd prefer to be safe

PR #2569, opened 2018-02-27
Show keys for osu!mania maps

nanaya:
More convenient when searching.


nanaya:
That said, an actual keys filter would be more useful =D

Issue #2059, opened 2017-12-20
Forum post count and post history do not exist


E223:
On the old userpages there is a post count which also links to a user's post history.
![](https://death.s-ul.eu/YemQYpkM.png)

peppy:
@arflyte need to decide if we still want to show this and if so, need a design for it.

arflyte:
![2b0550cb01](https://user-images.githubusercontent.com/3632750/36190487-06be1b4c-119c-11e8-867b-1cd91af54556.png)

Feel free to replace the words with something better; I can't think of anything better than 'post contribution' for now.

peppy:
This needs to link to a search showing a users' posts, for what it's worth.

PR #2578, opened 2018-02-28
Show post count on user profile page

notbakaneko:
#2059

no link to post history yet, that will be separate

### DIFF
--- a/resources/lang/en/users.php
+++ b/resources/lang/en/users.php
@@ -94,6 +94,7 @@ return [
         'origin_country_age' => ':age from :country',
         'origin_country' => 'From :country',
         'page_description' => 'osu! - Everything you ever wanted to know about :username!',
+        'post_count' => 'Contributed :count forum posts',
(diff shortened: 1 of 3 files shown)


resources/lang/en/users.php
```diff
@@ -94,6 +94,10 @@
         'origin_country_age' => ':age from :country',
         'origin_country' => 'From :country',
         'page_description' => 'osu! - Everything you ever wanted to know about :username!',
+        'post_count' => [
+            '_' => 'Contributed :count',
+            'count' => ':count forum posts',
```
nanaya: singular variation as well?

resources/lang/en/users.php
```diff
@@ -94,6 +94,10 @@
         'origin_country_age' => ':age from :country',
         'origin_country' => 'From :country',
         'page_description' => 'osu! - Everything you ever wanted to know about :username!',
+        'post_count' => [
```
nanaya: arrays goes to the arrays area.

PR #2584, opened 2018-02-28
Added username as possible entry for login

Endaris:
It is now consistent with the error message for invalid credentials and I confirmed that it is indeed possible to login with username instead of email address.


LiquidPL:
I think we actually might want to phase out username logins in the near future.

peppy:
Thanks for the submission, but we are moving forward to login via email specifically. This is intentionally omitted.

PR #2591, opened 2018-03-01
Refactor beatmapset cover generation

### LINKED ISSUE #2533: Old beatmap thumbnails don't show
The very first beatmaps with at least a thumbnail don't show it. Also on their beatmap page.
New Site: https://osu.ppy.sh/beatmapsets?sort=ranked_asc

Old Site: https://osu.ppy.sh/p/beatmaplist?l=1&r=0&q=&g=0&la=0&ra=&s=4&o=0&r=0&q=

nekodex:
Gets rid of the horrible manual temp-file management mess with the working-directory-paths-being-passed-around nonsense, hopefully this will result in less zombie files being left around.

Also adds some fallbacks in for finding a beatmapset's background image. Should result in less missing cover images.

Fixes #2533


nanaya:
Meanwhile, I've been adding wrapper to some php functions to return null instead of false on failure :dancer:

PR #2604, opened 2018-03-02
from country -> in country

Xinnoh:
screen shot 2018-03-02 at 9 37 26 am

Living in a country doesn't mean you're from there. Eg I'm Canadian, but it says I'm from Singapore.
Some users would prefer to not be identified with their country that they live in


peppy:
i disagree with this change

PR #2605, opened 2018-03-02
Replace missed instances of "log" with "sign"

Joehuu:
~Don't know how to link tournaments to sign in, so I edited without it. I might miss other instances.~


peppy:
Probably want to add invalidation comments on other localisations

TPGPL:
Polish is good

kj415j45:
⚠ Request Change: Remove notes in zh locale files since nothing changed.

kj415j45:
Besides, there's a change in `en/Tournaments.php`: `:login` -> `sign in`. I don't think this is correct.

(I don't create Line comment because my Chrome on mobile crash when loading so much changes :v)

resources/lang/en/tournament.php
```diff
@@ -33,7 +33,7 @@
             'cancel' => 'Cancel Registration',
         ],
         'entered' => 'Your are registered for this tournament.<br><br>Note that this does not mean you have been assigned to a team.<br><br>Further instructions will be sent to you via email closer to the tournament date, so please ensure your osu! account\'s email address is valid!',
-        'login_to_register' => 'Please :login to view registration details!',
+        'login_to_register' => 'Please sign in to view registration details!',
```
nanaya: no need to change this one. The text uses whatever `users.anonymous.login_link` is.

Don't forget to remove comment for other locales.

PR #2606, opened 2018-03-02
Format user profile post count

### LINKED ISSUE #2598: Post count missing number format
Should probably have commas and what not.

### LINKED ISSUE #2601: [Polish translation, Profile page] Contributed number is undefined
actually it's a probably with Polish translation <sup>again</sup>, but i think it still shouldnt happen.

notbakaneko:
fixes #2598

also fixes #2601

### DIFF
--- a/resources/assets/coffee/react/profile-page/header-extra.coffee
+++ b/resources/assets/coffee/react/profile-page/header-extra.coffee
@@ -69,7 +69,7 @@ class ProfilePage.HeaderExtra extends React.Component
     joinDate = moment(@props.user.join_date)
     joinDateTitle = joinDate.format('LL')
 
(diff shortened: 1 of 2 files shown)


resources/assets/coffee/react/profile-page/header-extra.coffee
```diff
@@ -69,7 +69,7 @@ class ProfilePage.HeaderExtra extends React.Component
     joinDate = moment(@props.user.join_date)
     joinDateTitle = joinDate.format('LL')
 
-    postCount = osu.transChoice 'users.show.post_count.count', @props.user.post_count
+    postCount = osu.transChoice 'users.show.post_count.count', Number(@props.user.post_count).toLocaleString()
```
peppy: it the `Number()` part required? doesn't look like it is in other usages of `toLocaleString`

PR #2639, opened 2018-03-07
API beatmapset lookup endpoint

### LINKED ISSUE #2625: Retrieving beatmapset with just a beatmap's ID
The response from the API's `/beatmaps/:beatmap_id` endpoint currently returns a beatmap's set incompletely.

image

It's missing the list of the beatmaps in the set for starters. This is currently a roadblock in looking up beatmaps by ID in the client (from chat links etc.), since beatmaps need to be displayed as part of a set. Should sets be looked up by a second request using the `beatmapset_id` of the response or is the entire set supposed to be included in this response?

nekodex:
Similar to beatmap lookup, currently only supports looking up by `beatmap_id`.

e.g.: `/api/v2/beatmapsets/lookup?beatmap_id=660345`

Ideally this functionality will eventually be merged into search, along with the ability to search other beatmap/set metadata, etc.

fixes #2625

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -281,6 +281,8 @@ Route::group(['as' => 'api.', 'prefix' => 'api', 'namespace' => 'API', 'middlewa
         Route::get('beatmapsets/search/{filters?}', '\App\Http\Controllers\BeatmapsetsController@search');
 
         // Beatmapsets
+        //   GET /api/v2/beatmapsets/lookup
+        Route::get('beatmapsets/lookup', 'BeatmapsetsController@lookup');
         //   GET /api/v2/beatmapsets/:beatmapset/download
         Route::get('beatmapsets/{beatmapset}/download', ['uses' => '\App\Http\Controllers\BeatmapsetsController@download']);
         //   GET /api/v2/beatmapsets/:beatmapset_id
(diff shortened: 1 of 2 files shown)


routes/web.php
```diff
@@ -255,34 +255,36 @@
...
             // POST /api/v2/chat/messages/new
-            Route::post('messages', ['uses' => 'ChatController@postMessage']);
+            Route::post('messages', 'ChatController@postMessage');
         });
 
         Route::resource('rooms', 'RoomsController', ['only' => ['show']]);
 
         Route::group(['prefix' => 'beatmapsets'], function () {
-            Route::get('favourites', ['uses' => 'BeatmapsetsController@favourites']);     //  GET /api/v2/beatmapsets/favourites
+            Route::get('favourites', 'BeatmapsetsController@favourites');     //  GET /api/v2/beatmapsets/favourites
```
nanaya: might as well merge with the `// Beatmapsets` group below while at it?

nekodex: hrm I wonder if that endpoint should be nested under `users` or `me` instead?

nanaya: hrm right :confused:

nekodex: Actually that endpoint probably isn't even needed anymore, I'll make a new PR removing that stuff.

PR #2650, opened 2018-03-08
Less misleading modding discussions tabs

### LINKED ISSUE #2641: General (All Difficulties) and General tab names are misleading and confusing.
Anyone I know not already familiar with it can and will get these two confused. What's the difference? Huh? And would accidentally post points for the mapset in the "General" tab.

Suggestion:
General (All Difficulties) -> General (Mapset)
or
General (All Difficulties) -> Mapset

and
General -> General (DifficultyName)

Where DifficultyName would be automatically replaced with Easy, Insane, Marathon, etc, whatever the current difficulty you are on is called.  (It would need to be truncated for extra long difficulty names of course)

I think this would make it much clearer what the difference is between the two, and prevent accidental misuse of the two tabs.

shortthu:
closes #2641


nekodex:
Yeah, so this PR doesn't work:


This wasn't a straight-forward translation fix, so I did it myself. See: #2657

shortthu:
👌

PR #2670, opened 2018-03-11
Add unweighted accuracy to score leaderboard

### LINKED ISSUE #2665: Overall Accuracy on Ranked Score Leaderboard Page
In the current site, overall accuracy is used on the score leaderboard page (an accuracy that I believe takes into account every single score, not weighted). However on the new site, weighted (pp) accuracy is displayed instead.  I think it makes more sense for the accuracy that accounts for every score set to be there, rather than the accuracy based on pp-weight . 

As of right now, I don't think the overall accuracy can be seen anywhere on the new site.

cl8n:
resolves #2665 

chrome_2018-03-10_22-36-17

need confirmation that `accuracy` is the correct field for the acc shown at https://osu.ppy.sh/p/playerranking

### DIFF
--- a/resources/lang/en/rankings.php
+++ b/resources/lang/en/rankings.php
@@ -28,6 +28,7 @@ return [
     ],
     'stat' => [
         'accuracy' => 'Accuracy',
+        'unweighted_accuracy' => 'Unweighted Acc.',
         'active_users' => 'Active Users',
         'country' => 'Country',
         'play_count' => 'Play Count',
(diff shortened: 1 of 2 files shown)


peppy:
I'm not sure this is something we even want to display (not 100% it's even being updated correctly).

LiquidPL:
I'm assuming that it could be displayed on hover regular accuracy when necessary (I think I saw a post once showing an S play with 100% acc, rounded up from 99.995 or something :thinking: ).

peppy:
For now I don’t think this is the correct path forward; I’m all for deprecating this stat permanently.

cl8n:
aw, alright. I thought the stat was nice for completionists to get a sense of their accuracy across all plays, but I guess it doesn't make a lot of sense to be comparing it to other players' on the leaderboard

PR #2677, opened 2018-03-12
Tooltips and flag links

### LINKED ISSUE #2669: Link avatar and flags on beatmap rankings
Also the color change of the arrow on tooltips blends in with white backgrounds which looks bad.

nanaya:
Fixes #2669. Cleaned up some styles while at it.


notbakaneko:
guess the borders don't need to be changed? :trollface:

nanaya:
Nuked out and rebased :boom:

PR #2684, opened 2018-03-13
Add default system font for Korea macOS

DingGGu:
In Korea's macOS, Apple recommends using "Apple SD Gothic Neo"

It looks better than before.

Reference: https://support.apple.com/en-us/HT206872

Before:

After:

### DIFF
--- a/resources/assets/less/bootstrap-variables.less
+++ b/resources/assets/less/bootstrap-variables.less
@@ -29,7 +29,7 @@
 @link-color: @blue-darker;
 @link-hover-color: @blue;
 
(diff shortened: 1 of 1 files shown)


DingGGu:
@notbakaneko How about now?

Osu website using Apple Gothic now, and I feel Apple Gothic is lagging aesthetically.

Mac applications or iOS applications do not use Apple Gothic.
It is a font that can be seen in programs such as Java applets which are rarely used at present.

notbakaneko:
@DingGGu Thanks for your contribution; we're rolling up this change with other font changes #2735

PR #2685, opened 2018-03-13
Null rank history if blank

nanaya:
Either that or fake it correctly and not just empty array? :man_shrugging: 

The only react component using it already does the null check.

The client doesn't like empty array :dancer:


peppy:
Resolves remaining part of ppy/osu#2148

PR #2690, opened 2018-03-14
Fix client beatmap discussion failing to parse url with empty path components

notbakaneko:
Fixes parse failing when url has a trailing `/` or one of the components is empty.
Also does whitelisting of `mode` and `filter` when parsing.


resources/assets/coffee/_classes/beatmap-discussion-helper.coffee
```diff
@@ -21,6 +21,9 @@ class @BeatmapDiscussionHelper
   @DEFAULT_MODE: 'timeline'
   @DEFAULT_FILTER: 'total'
 
+  @MODES = ['timeline', 'general', 'generalAll']
```
nanaya: `event`?

PR #2692, opened 2018-03-14
Check beatmap cover images on qualification

nekodex:
...and queue a regeneration if any are missing

### DIFF
--- a/app/Jobs/CheckBeatmapsetCovers.php
+++ b/app/Jobs/CheckBeatmapsetCovers.php
@@ -1,7 +1,7 @@
 <?php
 
 /**
- *    Copyright 2015-2017 ppy Pty. Ltd.
+ *    Copyright 2015-2018 ppy Pty. Ltd.
  *
  *    This file is part of osu!web. osu!web is distributed with the hope of
  *    attracting more community contributions to the core ecosystem of osu!.
--- a/app/Jobs/CheckBeatmapsetCovers.php
+++ b/app/Jobs/CheckBeatmapsetCovers.php
@@ -1,0 +1,57 @@
+<?php
+
+/**
+ *    Copyright 2015-2017 ppy Pty. Ltd.
+ *
+ *    This file is part of osu!web. osu!web is distributed with the hope of
+ *    attracting more community contributions to the core ecosystem of osu!.
+ *
+ *    osu!web is free software: you can redistribute it and/or modify
+ *    it under the terms of the Affero GNU General Public License version 3
+ *    as published by the Free Software Foundation.
+ *
+ *    osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+ *    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *    See the GNU Affero General Public License for more details.
+ *
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace App\Jobs;
+
+use App\Models\Beatmapset;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Queue\SerializesModels;
+
+class CheckBeatmapsetCovers implements ShouldQueue
+{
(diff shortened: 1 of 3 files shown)


app/Models/Beatmapset.php
```diff
@@ -830,6 +831,32 @@ public function regenerateCovers(array $sizesToRegenerate = null)
...
+                curl_exec($ch);
+                $responseCode = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+                curl_close($ch);
+
+                if ($responseCode !== 200) {
+                    $allGood = false;
+                    break;
+                }
+            }
+        } catch (\Exception $e) {
```
nanaya: what is this catching?

app/Models/Beatmapset.php
```diff
@@ -830,6 +831,32 @@ public function regenerateCovers(array $sizesToRegenerate = null)
         $this->update(['cover_updated_at' => $this->freshTimestamp()]);
     }
 
+    public function allCoverImagesPresent()
+    {
+        $allGood = true;
+        try {
+            foreach ($this->allCoverURLs() as $size => $url) {
+                $ch = curl_init($url);
+                curl_setopt($ch, CURLOPT_HEADER, true);
```
nanaya: there's this `curl_setopt_array`.

app/Jobs/CheckBeatmapsetCovers.php
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        // trigger a cover regeneration if any cover images are missing
+        if (!$this->beatmapset->allCoverImagesPresent()) {
+            $job = (new RegenerateBeatmapsetCover($this->beatmapset))->onQueue('beatmap_high');
```
nanaya: Apparently there's `Dispatchable` trait which may simplify things.

app/Models/Beatmapset.php
```diff
@@ -830,6 +831,32 @@ public function regenerateCovers(array $sizesToRegenerate = null)
         $this->update(['cover_updated_at' => $this->freshTimestamp()]);
     }
 
+    public function allCoverImagesPresent()
+    {
+        $allGood = true;
+        try {
+            foreach ($this->allCoverURLs() as $size => $url) {
+                $ch = curl_init($url);
```
nanaya: better off as helper function?

app/Models/Beatmapset.php
```diff
@@ -830,6 +831,32 @@ public function regenerateCovers(array $sizesToRegenerate = null)
         $this->update(['cover_updated_at' => $this->freshTimestamp()]);
     }
 
+    public function allCoverImagesPresent()
+    {
+        $allGood = true;
```
nanaya: a bit shorter...

```php
... function ...()
{
    foreach (... as ...) {
        if (!url_check(...)) {
            return false;
        }
    }

    return true;
}
```

?

app/Jobs/CheckBeatmapsetCovers.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+/**
+ *    Copyright 2015-2017 ppy Pty. Ltd.
```
nanaya: 2018 =D

app/Jobs/CheckBeatmapsetCovers.php
nekodex: Yeah I don't think it really helps... if anything, I'd argue it makes it less readable not more readable?

```php
$job = (new RegenerateBeatmapsetCover($this->beatmapset))->onQueue('beatmap_high');
dispatch($job);
```
vs
```php
(new RegenerateBeatmapsetCover($this->beatmapset))->dispatch()->onQueue('beatmap_high');
```

notbakaneko: I don't think `Dispatchable` helps when the job is a different class.

nanaya: more like

```
RegenerateBeatmapsetCover::dispatch($this->beatmapset))->onQueue('beatmap_high');
```

nekodex: My readability complaint remains, `dispatch()->onQueue()` makes no sense to me, I expect dispatch to... dispatch, not to be chained onto to enqueue elsewhere

PR #2699, opened 2018-03-15
Bump React to v16.2.0

nekodex:
For performance improvements and whatnot.

I haven't noticed anything breaking, but double-check to be sure.


notbakaneko:
ᕕ( ᐛ )ᕗ

Issue #2696, opened 2018-03-15
History navigation buttons don't work as expected


peppy:
- Go to a beatmap page
- Scroll down to scores a click a profile
- Scroll down in profile
- Hit back button
- *page scrolls to top of page unexpectedly*
- *page does not scroll to where you were scrolled to*
- Hit forward button
- *page scrolls randomly*
- *page does not scroll to where you were scrolled to*

![Pasted Image 1.gif](https://shipusercontent.com/63ed03d1372bbe946dd7f4c1e970d90d/Pasted%20Image%201.gif)

This happens in basically all navigation site-wide and needs to be fixed as highest priority. In my eyes it makes the site frustratingly unusable.

nekodex:
"fixed"  - peppy

PR #2710, opened 2018-03-16
Turbolinks, react, and navigation

nanaya:
Should fix part of #2696. Need more checks. There are some components not yet updated to `registerPersistent` thingy.

- unmount after replacing the page
- save state on `before-cache` to where the component is attached to
- disable anchor existence check in turbolinks because some of the anchors used don't exist which causes page fetch on history navigation

### DIFF
--- a/resources/assets/coffee/react/beatmapset-page/main.coffee
+++ b/resources/assets/coffee/react/beatmapset-page/main.coffee
@@ -23,40 +23,39 @@ class BeatmapsetPage.Main extends React.Component
   constructor: (props) ->
     super props
 
-    optionsHash = BeatmapsetPageHash.parse location.hash
-    @initialPage = optionsHash.page
-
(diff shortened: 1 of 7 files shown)


resources/assets/coffee/react/beatmapset-page/main.coffee
```diff
@@ -23,40 +23,39 @@ class BeatmapsetPage.Main extends React.Component
...
+    @scoreboardXhr = null
+    @favouriteXhr = null
 
-    currentBeatmap = BeatmapHelper.find
-      group: beatmaps
-      id: optionsHash.beatmapId
-      mode: optionsHash.playmode
+    @state = JSON.parse(@props.container.dataset.state ? 'null')
+    @restoredState = @state?
+    console.log 'restored state?', @restoredState
```
nekodex: neat

nekodex:
If you navigate back/forwards while scores are being loaded on the beatmap page, the page gets stuck in the 'loading' state.

Issue #2590, opened 2018-03-01
Search implementation cleaning/refactoring


notbakaneko:
Currently, calling search and then handling the results and pagination is a bit too much of a parameter hot-potato juggling grab-bag.

Some of the changes are already backported to #2332, so this is continuing on that and will lead into making a search mode for #2059, and fixing up displaying posts within topic searches.

PR #2711, opened 2018-03-16
Add link to to user post history

### LINKED ISSUE #2059: Forum post count and post history do not exist

notbakaneko:
Adds a link to the user's post history on their profile page.
fixes #2059 

also pulls in more changes for/from #2590


resources/views/objects/search/_forum_options.blade.php
```diff
@@ -15,41 +15,62 @@
...
+            <input
+                name="{{ $fields['user'] }}"
+                value="{{ request($fields['user']) }}"
+                class="search-forum-options__input search-forum-options__input--text"
+            >
+        </label>
+    @endif
+
+    {{-- FIXME: remove querystring check? --}}
+    @if ($fields['topicId'] !== false && present(request($fields['topicId'])))
```
nanaya: wtf is all this `!== false` check ಠ_ಠ

nanaya: I see ಠ_ಠ but why not just null

notbakaneko: `null` or `!isset` meant to use the existing value, until I changed it to always force the defaults, so I guess it has no meaning now 💃

resources/lang/en/users.php
```diff
@@ -66,6 +66,12 @@
 
         'here' => 'here', // this is substituted in when generating a link above. change it to suit the language.
     ],
+
+    'posts' => [
+        'title' => ':username\'s posts',
+        'page_description' => 'osu! - Everything you ever wanted to know about :username!',
```
nanaya: wat

PR #2735, opened 2018-03-19
Update font-family for non-latin character sets.

notbakaneko:
includes https://github.com/ppy/osu-web/pull/2684

`Hiragino Kaku Gothic ProN` - Japanese and Chinese text on OSX and iOS.
`Apple SD Gothic Neo` - Hangul on OSX
`Meiryo` - Japanese on Vista+
`Microsoft YaHei` - Chinese on Vista+
`sans-serif` should fallback to some form of `Hiragino Sans` or `Hiragino Sans GB` on OSX or `SimHei`/`SimSun`? on WIndows

Also removes the mythical `Arial Grande`


shortthu:
Can we have something for Vietnamese too? They are not fully supported in Exo 2
Found on Google Fonts that Exo (1) have Vietnamese support

nanaya:
@shortthu I don't think we're adding more web fonts. Otherwise feel free suggesting better font(s) for your preferred languages.

Issue #2798, opened 2018-03-24
An <a> tag with a wrong output in username changing page.


ExerciseBook:
![test](https://user-images.githubusercontent.com/6327311/37864392-f5302430-2fa8-11e8-9daa-3567b6b90ded.png)

AndyThePie:
I have no idea what I'm doing I'm so sorry

Looking at inspect element, it appears that the **<** and **>** are instead `&lt;` and `&gt;.`
Replacing them with the proper brackets seems to fix it.

christopher-dG:
Looks like it's the same as #2793.

notbakaneko:
fixed by https://github.com/ppy/osu-web/pull/2805

PR #2805, opened 2018-03-26
Allow change username server-side error messages to be html

### LINKED ISSUE #2793: HTML <strong> formatting isn't applied to name change message
https://osu.ppy.sh/store/products/32

capture

notbakaneko:
fixes #2793 and #2798

Also move the messages to translations.


app/Models/User.php
```diff
@@ -300,13 +300,13 @@ public static function validateUsername($username, $previousUsername = null)
...
             if ($remaining->days > 365 * 2) {
                 //no need to mention the inactivity period of the account is actively in use.
-                return ['Username is already in use!'];
+                return [trans('model_validation.user.username_in_use')];
             } elseif ($remaining->days > 0) {
-                return ["This username will be available for use in <strong>{$remaining->days}</strong> days."];
+                return [trans_choice('model_validation.user.username_available_in_days', $remaining->days)];
             } elseif ($remaining->h > 0) {
-                return ["This username will be available for use in <strong>{$remaining->h}</strong> hours."];
+                return [trans_choice('model_validation.user.username_available_in_hours', $remaining->hours)];
```
nanaya: The property name used between here and the check doesn't match ಠ_ಠ

Or more like `->hours` doesn't exist.

resources/lang/en/model_validation.php
```diff
@@ -88,11 +88,28 @@
         'invalid_email' => "Doesn't seem to be a valid email address.",
         'too_short' => 'New password is too short.',
         'unknown_duplicate' => 'Username or email address already used.',
+        'username_available_in_days' => 'This username will be available for use in <strong>:count</strong> day.|This username will be available for use in <strong>:count</strong> days.',
+        'username_available_in_hours' => 'This username will be available for use in <strong>:count</strong> hour.|This username will be available for use in <strong>:count</strong> hours.',
```
nanaya: The `<strong>` part can probably be extended to include its unit as well (or just removed because current design has `<strong>` wrapping the whole thing). And then the hours and days version can be merged or something.

resources/lang/en/model_validation.php
```diff
@@ -88,11 +88,28 @@
         'invalid_email' => "Doesn't seem to be a valid email address.",
         'too_short' => 'New password is too short.',
         'unknown_duplicate' => 'Username or email address already used.',
+        'username_available_in_days' => '{1} This username will be available for use in :count day.|[2,*] This username will be available for use in :count days.',
+        'username_available_in_hours' => '{1} This username will be available for use in :count hour.|[2,*] This username will be available for use in :count hours.',
```
nanaya: does it actually do anything? Looks the same as without explicit numbers.

notbakaneko: it makes the `pl` translation use the `en` rules when untranslated

nanaya: looks more like symfony/laravel bug ಠ_ಠ also missing 0.

notbakaneko: we won't need 0 yet :trollface:

nanaya: I'd rather "fix" the `trans_choice` function so we don't need to do this for everything.

PR #2826, opened 2018-03-27
User badge

nanaya:
Show the badges. Upscaled and stuff.


nekodex:
In IE11 the marquee animation doesn't play and the popup doesn't resize correctly:


Safari (desktop and mobile) doesn't seem to mask properly:

peppy:
- On first appearance, the marquee effect should decelerate probably over two or three circulations of all badges (if more than one is present). Currently the effect is pretty hard to see.
- May want to change alignment to left aligned when expanded, looks a bit weird with centered rows of unequal length.

PR #2827, opened 2018-03-27
Elasticsearch settings update

notbakaneko:
- configurable number of shards when creating index (doesn't affect existing indices)
- increment error counter on Datadog on exception


app/helpers.php
```diff
@@ -141,7 +141,13 @@ function es_search($params)
         $error = $e;
     }
 
-    Log::debug($error);
+    if (isset($error) && config('datadog-helper.enabled')) {
```
nanaya: `isset` check rather useless because it's always assigned with exception? And not sure why there's a null coalesce at the end =D

app/Libraries/Elasticsearch/Search.php
```diff
@@ -152,6 +153,14 @@ private function fetch()
             $this->error = $e;
         }
 
+        if (isset($this->error) && config('datadog-helper.enabled')) {
```
nanaya: See other comment about `isset` check.

app/helpers.php
notbakaneko: oh wait, forgot about the return up there :best:

PR #2828, opened 2018-03-27
Cache beatmap listing search result

notbakaneko:
Quick fix to cache some of the results of the beatmap listing to reduce the queue size on elasticsearch.

### DIFF
--- a/app/Models/Beatmapset.php
+++ b/app/Models/Beatmapset.php
@@ -425,11 +425,6 @@ class Beatmapset extends Model implements AfterCommit
         return $params;
     }
 
-    private static function shouldCacheSearch(array $params)
-    {
-        return !(present($params['query']) || $params['recommended']);
-    }
-
     public static function searchES(array $params = [])
     {
         $searchParams = [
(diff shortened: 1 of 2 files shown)


app/Models/Beatmapset.php
```diff
@@ -1319,4 +1326,17 @@ public static function removeMetadataText($text)
...
+    {
+        ksort($params);
+        unset($params['user']);
+
+        return 'beatmapset-search:'.serialize($params);
+    }
+
+    private static function shouldCacheSearch(array $params)
+    {
+        return !(present($params['query']) || $params['recommended']);
```
nanaya: also for `status` of 2 (favorites) and 6 (own maps) and `rank` (own ranked maps).

app/Models/Beatmapset.php
```diff
@@ -1319,4 +1326,21 @@ public static function removeMetadataText($text)
...
+        unset($params['user']);
+
+        return 'beatmapset-search:'.serialize($params);
+    }
+
+    private static function shouldCacheSearch(array $params)
+    {
+        return !(
+            present($params['query'])
+            || in_array($params['status'], [2, 6], true) // favourites, ranked.
```
nanaya: More like 6 is own maps. Ranked is another parameter :dancer: 

Accidentally commented in somewhere useless :ok_hand:

PR #2834, opened 2018-03-28
Limit height of user page section on user profiles

### LINKED ISSUE #2815: Userpage (me! section) should be collapsible
As evidenced by [vaxei](https://osu.ppy.sh/users/vaxei)'s profile.

(also collapses by default, like old website)

nekodex:
...'cuz some people like to play silly buggers

fixes #2815


resources/assets/less/bem/page-extra.less
```diff
@@ -80,6 +85,12 @@
     width: 100%;
   }
 
+  &__content-overflow-wrapper {
+    max-height: 400px;
```
shortthu: Why use 400px since the max height of _most_ old userpage is 192px?

Xinnoh: From what I see, the userpage doesn't have the option to expand and is locked at 400.
400 is approx the size of an expanded userpage on old web

resources/assets/less/bem/page-extra.less
```diff
@@ -35,6 +35,11 @@
     padding: 0;
   }
 
+  &--me-section {
+    padding-right: 0px;
```
nanaya: hmmm

nanaya: I think most of scrollbars like does follow the padding (as in doesn't stick to the edge).

resources/assets/less/bem/page-extra.less
```diff
@@ -97,5 +119,9 @@
     right: 0;
     top: 0;
     cursor: move;
+
+    .@{top}--userpage & {
+      margin-right: 30px;
```
nanaya: Apparently the margin is different between mobile and desktop. Also both have their own variables.

PR #2844, opened 2018-03-29
Creation resources/lang/de

sharpManu:
Added new language folder 'de' with /de/accounts.php


nanaya:
There's already German translation project at #2240. As this one has fewer translated files at the moment, please join the other one instead.

PR #2847, opened 2018-03-29
Update mods order

### LINKED ISSUE #2835: Mods on scoreboards are out of order
https://osu.ppy.sh/beatmapsets/700945#osu/1483942
The order here is DT, FL, HD, HR, NC for some reason.
The format should be HD, HR, DT, FL else lots of people will be annoyed :x

Skipped the other mods for simplicity, but they're also out of order.

nanaya:
I think?

Fixes #2835.


app/Libraries/ModsHelper.php
```diff
@@ -28,21 +28,25 @@ class ModsHelper
         [3, 'HD'],
         [4, 'HR'],
         [5, 'SD'],
```
Xinnoh: My bad, SD/PF come after FL

This needs to be double checked by someone

PR #2850, opened 2018-03-30
Don't lowercase the username line in mobile nav

MegaApplePi:
whoops


resources/assets/less/bem/navbar-mobile-item.less
```diff
@@ -17,7 +17,10 @@
  */
 
 .navbar-mobile-item {
-  text-transform: lowercase;
+  &:not(.navbar-mobile-item--user){
+    text-transform: lowercase;
```
nanaya: set correct transform in user modifier below instead.

PR #2901, opened 2018-04-04
Check user is trying to checkout with current cart

notbakaneko:
Check that the user is trying to checkout with the current cart instead of an old one when beginning checkout. This resolves an issue where a user tries to checkout in one tab but deletes and creates a new cart in a different tab.

~~A side-effect of this is; trying to do a free checkout with a non-free order will now force the page to reload instead of showing the loading overlay forever (still won't checkout).~~
That was a bug :D


resources/assets/lib/store-checkout.coffee
```diff
@@ -37,13 +37,14 @@ export class StoreCheckout
 
     $(@CHECKOUT_SELECTOR).on 'click.checkout', (event) =>
       provider = event.target.dataset.provider
+      orderId = event.target.dataset.orderId
```
nanaya: looks like some already have `orderNumber`, maybe use that instead?

notbakaneko: `orderNumber` is for external reference

PR #2906, opened 2018-04-04
New header

nanaya:
RIP stuff (mainly quick search and current page breadcrumbs).

I think there are still quite a few leftover from old stuff (translations, etc).

Probably will be cleaned up a bit more later ┐(°～° )┌


resources/assets/less/bem/nav2.less
```diff
@@ -0,0 +1,201 @@
...
+    &::after {
+      content: '';
+      position: absolute;
+
+      border-radius: 10px;
+      top: (@_link-padding-vertical + @_link-highlight-margin-vertical);
+      right: 100%;
+      width: 3px;
+      height: calc(100% ~'-' (@_link-padding-vertical * 2 + @_link-highlight-margin-vertical * 2));
+      background-color: @yellow;
```
notbakaneko: I see a trailing space 🤔

app/Http/Controllers/HomeController.php
```diff
@@ -145,12 +145,13 @@ public function search()
         ]);
 
         $search = new Search($params);
+        $isSearchPage = true;
```
notbakaneko: seems like search should be its own section at this point ┐(°～° )┌

nanaya: There's a whole new search page design coming in, so  ┐(°～° )┌

PR #2911, opened 2018-04-05
Touchdevice mod display

### LINKED ISSUE #2876: Add a mod icon for touchscreen
+TD should also show up on leaderboards

nanaya:
~~Also sort the filter bar icons according to display order.~~ So it's based on in-game order.

Fixes #2876.


cl8n:
not sure about the mod order, I think it's most intuitive without any changes (because it matches the in-game mod selection). also looks nicer that way since the mod colors are grouped

PR #2953, opened 2018-04-10
Add missing tables and test for model table existence

notbakaneko:


tests/TableTest.php
```diff
@@ -0,0 +1,75 @@
...
+            $class::first();
+        } catch (Exception $e) {
+            return $class;
+        }
+    }
+
+    private function checkDir(string $basePath, string $namespace = '')
+    {
+        $errors = [];
+        $entries = scandir($basePath);
```
nanaya: There's this `File::allFiles()`.

tests/TableTest.php
```diff
@@ -0,0 +1,75 @@
...
+
+class TableTest extends TestCase
+{
+    public function testTableExistence()
+    {
+        $this->modelsPath = app_path().'/Models';
+        $errors = $this->checkDir($this->modelsPath);
+
+        // prints a diff with the classes that errored.
+        $this->assertEquals([], $errors);
```
nanaya: Probably should use `assertSame`.

tests/TableTest.php
```diff
@@ -0,0 +1,69 @@
...
+
+class TableTest extends TestCase
+{
+    private $modelsPath;
+
+    public function testTableExistence()
+    {
+        $errors = [];
+
+        $files = (new Finder)->in(app_path().'/Models')->files();
```
nanaya: Why don't just use the `File` thingy from laravel? ?_?

notbakaneko: It's a wrapper around `Finder` anyway ┐(ツ)┌

composer.json
```diff
@@ -37,6 +37,7 @@
     "paypal/rest-api-sdk-php": "*",
     "predis/predis": "~1.0",
     "sentry/sentry-laravel": "*",
+    "symfony/finder": "*",
```
nanaya: this probably isn't needed. If laravel changes the return, it'll explode in different way anyway and this wouldn't help ┐(°～° )┌

PR #2957, opened 2018-04-10
Use Laravel's `File` thingy

nanaya:
TIL ┐(' ヮ' )┌


app/Console/Commands/LocaleCheck.php
```diff
@@ -149,33 +150,12 @@ public function everything($locale, $heads)
...
-            }
-
-            if (is_dir($entryPath)) {
-                $files = array_merge($files, $this->files($locale, $entryPath));
-            }
-
-            if (is_file($entryPath) && ends_with($entryPath, '.php')) {
-                $files[] = $entryPath;
-            }
+        foreach (File::allFiles($this->basePath($locale)) as $entry) {
```
notbakaneko: can just return `File::allFiles` directly and `->getPathname()` in `heads()` or some of those other `SplFileInfo` methods

PR #2965, opened 2018-04-11
Wait to acquire row lock before checking if record can be deleted.

notbakaneko:
It should be waiting for `status` to finish committing, otherwise it's doing the check with an old value and then queueing the delete anyway

### DIFF
--- a/app/Models/Store/Order.php
+++ b/app/Models/Store/Order.php
@@ -340,7 +340,7 @@ class Order extends Model
 
     public function delete()
     {
-        DB::connection($this->connection)->transaction(function () {
+        $this->getConnection()->transaction(function () {
             // wait for any pending status update to be committed, otherwise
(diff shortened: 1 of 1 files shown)


app/Models/Store/Order.php
```diff
@@ -340,11 +340,15 @@ public function cancel()
 
     public function delete()
     {
-        if ($this->isModifiable() === false) {
-            // in most cases this would return a null key because the lookup for the cart
-            // would return a new cart anyway?
-            throw new Exception("Delete not allowed on Order ({$this->getKey()}).");
-        }
+        DB::connection($this->connection)->transaction(function () {
```
nanaya: `$this->getConnection()->transaction(...)`  looks like the same thing.

notbakaneko: making a separate PR for the existing ones

PR #2966, opened 2018-04-11
Add a birthday field to account settings

### LINKED ISSUE #2175: Birthday cannot be set on settings page
The old website's edit profile panel has an option to set your birthday, but this is absent on the new website.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/53389956-birthday-cannot-be-set-on-settings-page?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

LiquidPL:
Fixes #2175.

image

Uses a native HTML5 date field, with a regular text field with a `yyyy-mm-dd` placeholder for IE.


app/Models/User.php
```diff
@@ -502,6 +502,22 @@ public function getUserBirthdayAttribute($value)
...
+            $this->attributes['user_birthday'] = '';
+
+            return;
+        }
+
+        try {
+            $date = new Carbon($value);
+            $this->attributes['user_birthday'] = $date->format('d-m-Y');
+        } catch (Exception $e) {
+            throw new Exception(trans('errors.invalid-date-format'));
```
nanaya: This isn't how validation is done ಠ_ಠ

resources/views/accounts/_edit_entry_date.blade.php
```diff
@@ -0,0 +1,36 @@
...
+<div class="account-edit-entry js-account-edit" data-account-edit-auto-submit="1">
+    <input
+        class="account-edit-entry__input js-account-edit__input"
+        type="date"
+        min="1900-01-01"
+        max="{{ Carbon\Carbon::now()->toDateString() }}"
+        name="user[{{ $field }}]"
+        maxlength="10"
+        placeholder="yyyy-mm-dd"
+        data-last-value="{{ Auth::user()->$field->toDateString() }}"
```
nanaya: fail on `null` birthday (also the following line).

peppy:
After internal discussion we will not be storing or displaying birthdays going forward. All existing data will be purged in the very near future.

PR #2998, opened 2018-04-13
Define React and friends as externals in webpack

notbakaneko:
2 copies of React and friends were being served ( ﾟ ヮﾟ)


nanaya:
Is it from importing `TextareaAutosize`? Just noticed, shouldn't some of the `import-shims` be part of `vendor`? ಠ_ಠ

notbakaneko:
`vendor`only contains pre-built modules that can be concat'ed together.

PR #3001, opened 2018-04-13
Virtualize beatmap listing

### LINKED ISSUE #2117: Beatmap listings should be unloaded at some point
Same issue as #2048 but on the beatmap listing

imo having 40+ maps per page like the old system would work better because there's no correlation between maps, whereas forum posts directly affect each other.

notbakaneko:
sort of fixes #2117

Virtualizes the beatmap listing to keep it fast after scrolling through many pages. 

I gave up on using `react-virtualized` and opted to fork`react-virtual-list` to make it work with our flexbox layout on the page instead, since:
- it required way too many changes to work
- didn't do a very good job of restoring the scroll position when navigating back in history.
- this option turned out to be massively simpler ( ﾟ ヮﾟ)

It doesn't eject the already loaded data yet

### DIFF
--- a/resources/assets/less/bootstrap-variables.less
+++ b/resources/assets/less/bootstrap-variables.less
@@ -53,7 +53,7 @@
 @input-color: @gray-dark;
 @input-border-focus: @pink;
 
-@screen-sm-min: 900px; // sync with osu.isDesktop
+@screen-sm-min: 900px; // sync with osu.isDesktop, react/beatmaps/main.coffee
 @screen-lg-min: 1300px;
 @screen-md-min: @screen-sm-min;
 
--- a/resources/assets/less/bootstrap-variables.less
+++ b/resources/assets/less/bootstrap-variables.less
@@ -53,7 +53,7 @@
 @input-color: @gray-dark;
 @input-border-focus: @pink;
 
-@screen-sm-min: 900px; // sync with osu.isDesktop, react/beatmaps/main.coffee
+@screen-sm-min: 900px; // sync with osu.isDesktop
 @screen-lg-min: 1300px;
 @screen-md-min: @screen-sm-min;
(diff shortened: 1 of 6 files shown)


nanaya:
This seems to work...?

https://0paste.com/20671-fae0f485.txt

notbakaneko:
kind of wanted to avoid the whole row wrapper element thing.

Also pushing the column count through as `props` to the virtual list causes it to render the correct height when the list is at the end and the window resizes wider; otherwise we'd have to store the `chunk`ed list somewhere or keep track of something to tell it to recalculate at that point

notbakaneko:
keeping columnCount in state should allow it to rerender, though, hmm...

nanaya:
Not sure if forking the thing is better idea though. The project itself doesn't seem to be completely dead either.

notbakaneko:
ok, apparently just changing `items` only isn't enough to make it update the height of the scroll area ಠ_ಠ

notbakaneko:
rebased and sorta squashed whee

resources/assets/less/bootstrap-variables.less
```diff
@@ -53,7 +53,7 @@
 @input-color: @gray-dark;
 @input-border-focus: @pink;
 
-@screen-sm-min: 900px; // sync with osu.isDesktop
+@screen-sm-min: 900px; // sync with osu.isDesktop, react/beatmaps/main.coffee
```
notbakaneko: should probably figure out a way to import this into scripts

nanaya: why not use the `isDesktop` function? ಠ_ಠ

notbakaneko: owait :D

PR #3022, opened 2018-04-16
Show total first place rank count on profile; display pp instead of score 

### LINKED ISSUE #2941: First Place Ranks count
Would it be too troublesome to add a small metric that shows how many #1s you have since the new website only expands by 20 instead of 50 now.

https://i.imgur.com/AFPqPOX.png

Something like "First Place Ranks (362)" or something similar.

___________________________________________________________________________________________________________
Also I'm not sure why it still shows the map score instead of PP.

nanaya:
Fixes #2941.


nanaya:
I tried changing the score to show pp but then it means loved maps will be showing 0pp... ಠ_ಠ

TPGPL:
maybe show pp only when the map is ranked then? 🤔

PR #3036, opened 2018-04-18
Add check on new player for posting on forum

nanaya:
No link because there's currently no way to embed link to error message. It doesn't sound like a good idea anyway for main message; perhaps add some kind of extended info to exceptions (and then a way to pass it from authorization).

Can use better wording for posting too many on help forums.


resources/lang/en/authorization.php
```diff
@@ -84,6 +84,11 @@
                 'not_owner' => 'Only poster can edit the post.',
                 'topic_locked' => 'Can not edit post of a locked topic.',
             ],
+
+            'store' => [
+                'play_more' => 'Try playing the game before posting on the forums, please! If you have a problem with playing, please post to the Help and Support forum.',
+                'too_many_posts' => 'Please wait for responses of current posts.',
```
peppy: incorrect key, and not sure this makes sense (as getting responses to posts doesn't increase your post limit)

peppy:
Oh, that's an actual test failing

resources/lang/en/authorization.php
```diff
@@ -84,6 +84,11 @@
                 'not_owner' => 'Only poster can edit the post.',
                 'topic_locked' => 'Can not edit post of a locked topic.',
             ],
+
+            'store' => [
+                'play_more' => 'Try playing the game before posting on the forums, please! If you have a problem with playing, please post to the Help and Support forum.',
+                'too_many_help_posts' => 'Try playing the game before posting further. If you still have trouble playing, check help pages.',
```
nekodex: Maybe something along the lines of "You need to play the game more before you can make additional posts."?

PR #3056, opened 2018-04-21
Add missing log files to .gitignore

imncls:
If an error occurred with using the yarn, log files (e.g. `yarn-error.log`) were not ignored.
So, I updated `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,7 +20,6 @@ desktop.ini
 
 # Bower and node directories
 /node_modules
-/npm-debug.log
 
 # Mix manifest
 /public/mix-manifest.json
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -37,6 +36,13 @@ desktop.ini
 # server logs
 /logs/osu/*
 
+# Logs
+logs
+*.log
```
nanaya: Not sure if we want to do this. (also the line above)

peppy: yeah, these should probably be for specific paths rather than this broad.

notbakaneko:
closed in favour of #3077

PR #3074, opened 2018-04-24
Switch beatmap search rank filter to filter on beatmap_id

notbakaneko:
Filter on beatmap_id instead, removing the extraneous database query.


app/Libraries/Search/BeatmapsetSearch.php
```diff
@@ -236,6 +218,10 @@ private function getPlayedBeatmapIds()
                 ->forUser($this->params->user)
                 ->select('beatmap_id');
 
+            if ($rank !== null) {
```
notbakaneko: may be better as `!empty($rank)` with a default of `[]`? I dunno

PR #3082, opened 2018-04-25
Add button on user profile linking to change username store page

### LINKED ISSUE #2421: Change player name when hovering near it on user page

nekodex:
After internal discussion we decided it made more sense to put the link on the user settings page amongst the rest of the changeable user profile fields.

fixes #2421


Lyawi:
Sorry if it's not "allowed" to comment here after merging, but why not putting the button nearly the name instead of having a button looking a bit out of place?

PR #3146, opened 2018-05-08
Fix modding history not applying user filter

### LINKED ISSUE #3145: User Modding History is broken and showing ..collective history?
If you go to any users modding history page, you will currently see the same thing on every page: what appears to be activity for ALL users. 

The only sections currently working correctly and being different depending on which user's page you're seeing is "Most upvoted by" and "Most upvoted"

notbakaneko:
fixes #3145 

`request()->all()`/`request()->toArray()` also does not have the same values as `request()`

### DIFF
--- a/app/Models/BeatmapDiscussion.php
+++ b/app/Models/BeatmapDiscussion.php
@@ -105,7 +105,7 @@ class BeatmapDiscussion extends Model
             $query->withoutDeleted();
         }
 
-        if (!($params['is_moderator'] ?? false)) {
+        if (!($rawParams['is_moderator'] ?? false)) {
             $query->whereHas('user', function ($userQuery) {
                 $userQuery->default();
             });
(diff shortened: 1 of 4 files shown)


nanaya:
wtf is this "checks" tab ヽ( ﾟ ヮﾟ)ﾉ

app/Models/BeatmapDiscussion.php
```diff
@@ -105,7 +105,7 @@ public static function search($rawParams = [])
             $query->withoutDeleted();
         }
 
-        if (!($params['is_moderator'] ?? false)) {
+        if (!($rawParams['is_moderator'] ?? false)) {
```
nanaya: ( ¬‿¬)

notbakaneko: good thing defaults to `false` ( ¬‿¬)

peppy: can we stick to using sentences to comment on code rather than faces? it's very hard to follow along.

PR #3151, opened 2018-05-09
Update spoilerbox design

### LINKED ISSUE #2460: Lack of outline makes it difficult to tell what is and isn't in a spoilerbox
Once a spoiler box is opened, there is no outline or any form of distinction to show where the box's content ends. This can make it difficult to tell when it is open what does and does not belong in the opened box.  On the old website this was done by a slight variation in background color for the inside of the box.

MegaApplePi:
fixes #2460


resources/assets/less/bem/bbcode-spoilerbox.less
```diff
@@ -34,16 +34,21 @@
 
     .@{top}.js-spoilerbox--open & {
       display: block;
+      background: #DDD;
```
notbakaneko: lowercase

resources/assets/less/bem/bbcode-spoilerbox.less
```diff
@@ -34,16 +34,21 @@
 
     .@{top}.js-spoilerbox--open & {
       display: block;
+      background: #ddd;
+      border-radius: 4px;
```
nanaya: more like `.default-border-radius()`.

PR #3152, opened 2018-05-09
Only show discussion starters in user filter

### LINKED ISSUE #3148: Users that only replied got to "filter by user" list
example: https://osu.ppy.sh/beatmapsets/680812/discussion/1439436/timeline?user=2202645
Myxomatosis replied here https://osu.ppy.sh/beatmapsets/680812/discussion/1439436/timeline#/146730
while they got to the filter by user list 
image

result of it, everything is showed as 0
image

notbakaneko:
fixes #3148

### DIFF
--- a/resources/assets/coffee/react/beatmap-discussions/header.coffee
+++ b/resources/assets/coffee/react/beatmap-discussions/header.coffee
@@ -106,7 +106,7 @@ class BeatmapDiscussions.Header extends React.PureComponent
             className: "#{bn}__filter-group #{bn}__filter-group--stats"
             el BeatmapDiscussions.UserFilter,
               selectedUser: if @props.selectedUserId? then @props.users[@props.selectedUserId] else null
-              users: @discussionStarters()
+              users: @props.discussionStarters
 
             div
               className: "#{bn}__stats"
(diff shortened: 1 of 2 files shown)


resources/assets/coffee/react/beatmap-discussions/header.coffee
```diff
@@ -121,6 +121,11 @@ class BeatmapDiscussions.Header extends React.PureComponent
               beatmap: @props.currentBeatmap
 
 
+  discussionUsers: =>
```
nekodex: How about `discussionStarters` or something to be more descriptive?

Also not sure if it's better to do this once-off when discussions are loaded as opposed to each render?

resources/assets/coffee/react/beatmap-discussions/header.coffee
```diff
@@ -121,6 +121,11 @@ class BeatmapDiscussions.Header extends React.PureComponent
...
+    userIds = _.uniq(_.map(_.values(@props.discussions), 'user_id'))
+    _.pickBy(@props.users, (value) -> _.includes(userIds, value.id))
```
nanaya: Here have a magic thingy chained written in javascript.

    _($r.props.discussions).map('user_id').uniq().map(function(k){return $r.props.users[k]}).sortBy('username').value()

PR #3154, opened 2018-05-09
Sort qualified beatmap listing according to qualified queue date

### LINKED ISSUE #3072: Qualified listing needs a sorting option reflecting the qualified queue
Beatmap Qualified Listing order no longer reflects the queue, because it is not directly related to the date it was most recently qualified now, due to the changes to how qualified works for v2 maps

This can be extremely confusing as it will make it look like a map has a week to rank or the same as those around it, even if it has fewer days left.
There is now nothing indicating on the listing how close a qualified map is to being ranked.

If it is added, this should be the default sorting for the qualified section.

notbakaneko:
fixes #3072 

Maps that are not in the queue will be ordered below those that are, regardless of sort direction.

index mapping should be updated first so that the query doesn't die while the field doesn't exist
`
curl -XPUT -H "Content-Type: application/json" http://ES_HOST/beatmaps/_mapping/beatmaps -d '{ "properties": { "queued_at": { "type": "date" } } }'
`


app/Libraries/Search/BeatmapsetSearchRequestParams.php
```diff
@@ -68,15 +68,15 @@ public function __construct(Request $request, ?User $user = null)
         }
 
         $sort = explode('_', $request['sort']);
-        $this->sort = static::normalizeSort(
+        $this->sort = $this->normalizeSort(
```
nanaya: Not related with this pull request but this assignment and related functions seem a bit... interesting.

- first it create sort with given params
- and then the created sort is thrown out, replaced with another pretty much the same sort just the `sort[0]` mapped to correct field
- it's put in array
- the `normalizeSort` function loops the array which is always single element.
- (kind of related to this pull request) goes to great length using `array_unshift` instead of just doing it before appending with provided sort
  - it prepends the resulting sort, potentially overriding any previous sorts (`['artist', 'approved_date']` &rarr; `['queued_at', 'artist', 'approved_date']`)
  - except it doesn't matter because there's always only one sort

notbakaneko: probably should have added a comment there >_>

PR #3166, opened 2018-05-10
Retina-ify beatmapset-panel

nekodex:
This had been bothering me, so I fixed it.

(also includes a bit of cleanup, removing an unneeded container class or two)

### DIFF
--- a/resources/assets/coffee/react/_components/beatmapset-panel.coffee
+++ b/resources/assets/coffee/react/_components/beatmapset-panel.coffee
@@ -16,7 +16,7 @@
 #    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
 ###
 
-{div,a,i,span} = ReactDOMFactories
+{div,a,i,img,span} = ReactDOMFactories
 el = React.createElement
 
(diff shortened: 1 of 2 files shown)


resources/assets/coffee/react/_components/beatmapset-panel.coffee
```diff
@@ -16,7 +16,7 @@
 #    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
 ###
 
-{div,a,i,span} = ReactDOMFactories
+{div,a,i,img,span} = ReactDOMFactories
```
notbakaneko: looks unused

resources/assets/coffee/react/_components/beatmapset-panel.coffee
```diff
@@ -43,6 +43,11 @@ class @BeatmapsetPanel extends React.PureComponent
     $(document).off ".#{@eventId}"
 
 
+  hideImage: (e) ->
+    # hides img elements that have errored (hides native browser broken-image icons)
+    e.currentTarget.style.display = 'none';
```
notbakaneko: is that a semicolon

PR #3170, opened 2018-05-11
Filter blank discussions and change to case-insensitive username sort

### LINKED ISSUE #3167: Mysterious [deleted user] in user filter
image
https://osu.ppy.sh/beatmapsets/675086/discussion/-/generalAll

image
https://osu.ppy.sh/beatmapsets/738656/discussion

No clue what this means. Trying to click it doesn't do anything at all. 
Initially I thought it may have to do with restricted users, but that doesn't appear to be the case.

notbakaneko:
fixes #3167 and makes username sorting case-insensitive

### DIFF
--- a/resources/assets/coffee/react/beatmap-discussions/main.coffee
+++ b/resources/assets/coffee/react/beatmap-discussions/main.coffee
@@ -276,7 +276,9 @@ class BeatmapDiscussions.Main extends React.PureComponent
 
 
   discussions: =>
-    @cache.discussions ?= _.keyBy @state.beatmapset.discussions, 'id'
+    @cache.discussions ?= _ @state.beatmapset.discussions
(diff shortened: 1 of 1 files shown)


resources/assets/coffee/react/beatmap-discussions/main.coffee
```diff
@@ -276,16 +276,17 @@ class BeatmapDiscussions.Main extends React.PureComponent
 
 
   discussions: =>
-    @cache.discussions ?= _.keyBy @state.beatmapset.discussions, 'id'
+    @cache.discussions ?= _ @state.beatmapset.discussions
+                            .filter (d) -> !_.isEmpty(d)
```
nanaya: Or `reject ... -> _.isEmpty ...`. And looks like `currentDiscussions` can use this function as well (and remove similar check in it).

PR #3174, opened 2018-05-11
Extract replay file handler to an object

nanaya:
Sanity check before continuing to score deletion thingy =)

The object is created so file deletion can be queued separately from the score after the score itself is deleted.

Not sure how to better handle the storage config ಠ_ಠ

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -59,7 +59,7 @@ SLACK_ENDPOINT=https://myconan.net/null/
 # AVATAR_CACHE_PURGE_PREFIX=
 
 # Either "s3" or "local".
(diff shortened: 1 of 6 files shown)


.env.example
```diff
@@ -58,6 +58,9 @@ SLACK_ENDPOINT=https://myconan.net/null/
 # AVATAR_STORAGE=local-avatar
 # AVATAR_CACHE_PURGE_PREFIX=
 
+# Either "s3" or "local".
+# SCORE_REPLAYS_STORAGE_NAMESPACE=local
```
nanaya: whoops wrong variable name.

nanaya:
Superseded by #3207.

PR #3200, opened 2018-05-15
Don't overflow userpage extra section if dragging outside viewport

### LINKED ISSUE #2618: [Profile page] Disable ability to break the width of the page by changing block position

MegaApplePi:
fixes #2618 
This also prevents the user from dragging the scroll area downwards.


MegaApplePi:
Alright, I'm unable to get osu-web to load certain pages to check if I had set the argument correctly. So I'll close this for someone else to fix it.

PR #3203, opened 2018-05-15
Pre-check topic create auth

### LINKED ISSUE #3180: Don't show "Post new topic" button in subforums where user doesn't have permission to create a topic
Because it's confusing.

nanaya:
Fixes #3180.


nekodex:
I'd probably just make the not-logged-in behaviour be the same as the has-no-permission behaviour (i.e. disabled/greyed out saying 'please login' or whatever)?

Feels weird, especially in the case where you go from clicking new topic -> logging in -> button greyed out because you don't have permission to post

PR #3207, opened 2018-05-15
Delete scores on qualification

### LINKED ISSUE #2787: Qualifying should always remove leaderboards
previously disqualified maps do not delete the leaderboards upon requalification.

[this map is an example](https://osu.ppy.sh/beatmapsets/702111#osu/1544643): the top score on it was set before this one qualified 

it nukes upon hitting ranked anyways but it can cause a ton of broken replays in case stuff is remapped and constellations where the top play of a newly qualified map is months old and doesnt make any sense potentially

nanaya:
Fixes #2787.

### DIFF
--- a/app/Jobs/RemoveBeatmapsetBestScores.php
+++ b/app/Jobs/RemoveBeatmapsetBestScores.php
@@ -65,7 +65,7 @@ class RemoveBeatmapsetBestScores implements ShouldQueue
         foreach (Beatmap::MODES as $mode => $_modeInt) {
             static::scoreClass($mode)::whereIn('beatmap_id', $beatmapIds)
                 ->orderBy('score_id')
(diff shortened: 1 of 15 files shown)


app/Jobs/RemoveBeatmapsetBestScores.php
```diff
@@ -0,0 +1,74 @@
...
+     * @return void
+     */
+    public function handle()
+    {
+        $beatmapIds = model_pluck($this->beatmapset->beatmaps(), 'beatmap_id');
+
+        foreach (Beatmap::MODES as $mode => $_modeInt) {
+            static::scoreClass($mode)::whereIn('beatmap_id', $beatmapIds)
+                ->orderBy('score_id')
+                ->where('score_id', '<', $this->maxScoreIds[$mode])
```
nekodex: Shouldn't this be `<=`?

PR #3209, opened 2018-05-16
Show past tournaments on the tournament listing

### LINKED ISSUE #3165: Tournament listing should show all active tournaments
Previously only tournaments in the sign-up stage were shown. Now that we have useful content on the tournament pages, let's show all active tournaments (in any state) on https://osu.ppy.sh/community/tournaments.

nekodex:
Will probably need to add pagination later, but should be ok for the time being.

fixes #3165


app/Models/Tournament.php
```diff
@@ -28,14 +28,19 @@ class Tournament extends Model
...
-            ->where('signup_close', '>', Carbon::now())
-            ->orderBy('play_mode', 'desc')
+        $tournaments = self::query()
             ->orderBy('tournament_id', 'desc')
             ->get();
+
+        $current = $tournaments->where('end_date', '>', Carbon::now());
+        $previous = $tournaments->where('end_date', '<=', Carbon::now());
+
+        return [
```
nanaya: Can be either `compact()` or just do the `where` right here instead of assigning it first.

app/Models/Tournament.php
```diff
@@ -28,14 +28,19 @@ class Tournament extends Model
...
     protected $dates = ['signup_open', 'signup_close', 'start_date', 'end_date'];
 
-    public static function getRegistrationStage()
+    public static function getGroupedListing()
     {
-        return self::query()
-            ->where('signup_open', '<', Carbon::now())
-            ->where('signup_close', '>', Carbon::now())
-            ->orderBy('play_mode', 'desc')
+        $tournaments = self::query()
```
nanaya: `static`? :zap:

app/Models/Tournament.php
```diff
@@ -28,14 +28,19 @@ class Tournament extends Model
...
             ->orderBy('tournament_id', 'desc')
             ->get();
+
+        $current = $tournaments->where('end_date', '>', Carbon::now());
+        $previous = $tournaments->where('end_date', '<=', Carbon::now());
```
nanaya: Probably doesn't matter but this `Carbon::now()` is different than the one before.

And that said, the entire thing has similar "problem" :dancer:

PR #3210, opened 2018-05-16
fallback keys should still be translated in case they are supposed to be empty.

### LINKED ISSUE #3208: Broken strings on the osu!supporter page
https://osu.ppy.sh/home/support

notbakaneko:
fixes #3208


nanaya:
More like those empty translations should be removed. Also Crowdin refuses to translate empty strings.

notbakaneko:
if only the loop on the support page didn't literally rely on the translated string being empty

PR #3212, opened 2018-05-16
Don't release stock if at 0.

notbakaneko:


app/Models/Store/Product.php
```diff
@@ -169,6 +169,12 @@ public function release($quantity)
             return true;
         }
 
+        // stock may have been directly updated to 0.
+        // TODO: should count reservations and available stock separately or something.
+        if ($this->stock <= 0) {
+            return false;
```
nanaya: wtf does the return do ಠ_ಠ (also in the one above)

nanaya: *return value

notbakaneko: SUCCE-..wait the last one doesn't return 😕

nanaya: looks like the original function returned the save return and then nuked when changed to use increment.

Issue #2866, opened 2018-03-31
We need a better font-family


kj415j45:
![unknown-2](https://user-images.githubusercontent.com/18349191/38159975-310f44f2-34e7-11e8-83e1-f9440a217257.png)
This is current font-family's performance in zh locale, It seems that the Chinese '直' is displaying Japanese one.

Also, from members of osu-translate-zh's report, different browser shows different things:(Left is FireFox Nightly, Right is Chrome Latest)
![unknown-1](https://user-images.githubusercontent.com/18349191/38160005-fcc354a8-34e7-11e8-80ce-1cfa1b2352ab.png)

We suggest that different locales should use different css so that we can get the best view to users(at least CJK)

kj415j45:
![28932b024e2cb3e9](https://user-images.githubusercontent.com/18349191/38184481-9808c982-367a-11e8-8f84-f6bea034adab.png)
Addition ScreenShot, include an error ID

Cyanoxygen:
There is a problem that `DengXian` has **copyrights** . Maybe it is dangerous to use it. 
`system-ui` works but some or old browsers. Picture is here:
![effect](https://user-images.githubusercontent.com/29174191/40917532-3593cfde-6836-11e8-9d36-44061616f31b.png)     
So I think the best choice is per-language font selections, as @kj415j45 said. I have let that guy to use `Exo\ 2,Helvetica Neue, Tahoma, sans-serif` (Delete CJK fonts priority) but **nothing changed.**
In summary, `Exo\ 2` should be followed with `system-ui`, then set other font-family.

PR #3215, opened 2018-05-16
Allow overriding default font

### LINKED ISSUE #3169: Broken texts in Vietnamese localisation
Yes because Exo 2 doesn't support Vietnamese.

I was told to find a new one to use for the Vietnamese localisation but not sure how to choose so I'll leave that for the designer 😛 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58227872-broken-texts-in-vietnamese-localisation?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nanaya:
With the magic of CSS variable. Doesn't work in IE.

The whole site looks kind of weird though...

Fixes #3169.


nekodex:
Not sure this is the correct approach? This won't fix Vietnamese text being broken for anyone not using the Vietnamese locale... e.g. on the beatmap pages, etc?

nanaya:
The alternative is finding Vietnamese version of exo 2 because it shares alphabet with Latin.

shortthu:
Can't find any Vietnamese version of Exo 2 with Google (using Vietnamese keywords) so there is no hope finding it unless someone do it.
Only Exo 1 have a Vietnamese version though.

shortthu:
Well I think the fix is enough because not many people uses Vietnamese in the beatmap page or forum anyway.

kj415j45:
It seems that this is a good solution for #2866 too. We'll try to find the best order and font for zh.

PR #3223, opened 2018-05-17
Minor tense mistake

Phippe:


TPGPL:
We now have a dedicated localisation project available for those interested. As a result, we will no longer be accepting localisation input via pull requests.

To contribute you can visit https://crowdin.com/project/osu-web and register a free account to get started. You can help by voting on existing translations or suggesting alternatives in cases they are sub-optimal.

PR #3225, opened 2018-05-18
Add toggle to block private messages from non-friends

nekodex:
This will only affect forum PMs initially, but going forward this will affect all chat (in-game and out).


resources/assets/less/bem/account-edit-entry.less
```diff
@@ -183,4 +200,18 @@
       opacity: 1;
     }
   }
+
+  &__status-left {
+    display: flex;
+    align-items: center;
+
+    @media @desktop {
+      width: @label-width + 25;
```
nanaya: looks like the unit is missing

PR #3231, opened 2018-05-20
Normalize wiki redirect keys

MegaApplePi:
Since they should be lowercase letters for comparison.


nanaya:
Is it needed?

MegaApplePi:
Hmm, I don't think it's really needed. Someone will just need to keep an eye out in the redirect.yaml file for letter-casing mistakes.

nanaya:
Any new redirects should be tested anyway I think.

PR #3262, opened 2018-05-24
Shipping address updates for paypal checkout

notbakaneko:
- remove shipping address when not required
- force shipping address when required


app/Libraries/Payments/PaypalCreatePayment.php
```diff
@@ -110,6 +126,20 @@ private function getRedirectUrls()
             ->setCancelUrl(route('payments.paypal.declined', ['order_id' => $this->order->order_id]));
     }
 
+    private function getShippingAddress()
+    {
+        $address = $this->order->address;
+
+        return (new ShippingAddress())
```
nanaya: A quick check shows that passing array works as well instead of calling all the functions (also applies to all other paypal objects?).

.env.example
```diff
@@ -109,6 +109,7 @@ PAYPAL_URL=https://www.sandbox.paypal.com/cgi-bin/webscr
 PAYPAL_MERCHANT_ID=
 PAYPAL_CLIENT_ID=
 PAYPAL_CLIENT_SECRET=
+PAYPAL_NO_SHIPPING_EXPERIENCE_PROFILE=
```
nanaya: `..._ID`...?

app/Libraries/Payments/PaypalCreatePayment.php
```diff
@@ -92,7 +102,7 @@ private function getAmount()
 
     private function getItemList()
     {
-        return (new ItemList())
+        $itemList = (new ItemList())
```
nanaya: Check `requiresShipping` in `getShippingAddress` (returning `null` accordingly) instead so it can still be just `return new ItemList(...)`?

app/Libraries/Payments/PaypalCreatePayment.php
notbakaneko: and we're not going to be using any of it

nanaya: ?_?

I mean something like `return new ShippingAddress(['city' => ...])`.

PR #3263, opened 2018-05-24
Move elasticsearch schemas to json

notbakaneko:
To better align with directly dumping/loading the schema with elasticsearch. It's a bit easier to do development for new indices if we can just deal with json.

Also disable dynamic field creation.

can just create an index with something like
`curl -XPUT -H 'Content-Type: application/json' "http://localhost:9200/i_r_new_index" -d @config/schemas/schema.json`


nanaya:
Maybe a better example than creating index? Because it's currently just `Model::esCreateIndex('indexname')` which seems easier than `curl ...`

Also json can't have comment :disappointed:

notbakaneko:
When constantly changing index settings under development, needing only to work with json and directly with elasticsearch is a lot simpler that doing it through php. 
It also means not having to rely on Laravel for loading larger datasets like scores, so we can find faster ways to populate large indices.

Elasticsearch supports comments in json; if only `json_decode` also did ☹️ 

`Model::esCreateIndex` isn't going away either

PR #3266, opened 2018-05-25
Limit locale menu height

nanaya:
Less mysterious overflow.


notbakaneko:
it can be non-obvious it's scrollable when it cuts off now, though 🙁 
image

nanaya:
Only in osx with scrollbar default hidden though.

PR #3271, opened 2018-05-25
Allow Project Loved members to add/remove maps to the Loved category

Ephemeralis:
This is a somewhat hacky tack-on to the Beatmapset admin panel which does a number of things:

- Introduces an admin controller type that allows admins, QAT and Project Loved usergroup members to access the functions contained within (previous one allows only g5 admins)
- Alters the existing Beatmapset admin panel to use this new controller
- Alters the view for said panel to hide functions not applicable to the other two groups
- Adds a toggle that swaps maps that aren't ranked or approved to Loved, the inverse returning a map to WIP status instead without removing scores

It probably could be done better, but I'm not very good at Laravel and it works well enough from what I've tested. Also a good springboard to add other admin features for the QAT later on (such as BN usergroup addition, so on, so forth).


nekodex:
As peppy mentioned on discord, we don't want to be hacking stuff into place on the new site. Furthermore, adding non-admin stuff to the admin section is also something we don't want to do.

The best way to implement this is as an extension to the current beatmap modding system (by adding a new button that behaves similarly to nominate).

Gonna close this PR for now and [open up an issue](https://github.com/ppy/osu-web/issues/3344) for this to see if we can get this in soonish™

PR #3285, opened 2018-05-30
Return json on error for a json request

nanaya:
`ajax` only covers requests with `X-Requested-With=XMLHttpRequest`.
And change `expectsJson` to `wantsJson`. The former also includes `ajax` requests.

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -108,7 +108,7 @@ class Handler extends ExceptionHandler
         } else {
             $message = $this->exceptionMessage($e);
 
-            if ($request->ajax() || $request->wantsJson()) {
+            if ($request->expectsJson()) {
                 $response = response(['error' => $message]);
             } else {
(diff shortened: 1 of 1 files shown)


notbakaneko:
Actually, why doesn't the html version of `unauthenticated` return `401` also?

app/Exceptions/Handler.php
```diff
@@ -120,7 +120,7 @@ public function render($request, Exception $e)
 
     protected function unauthenticated($request, AuthenticationException $exception)
     {
-        if ($request->expectsJson()) {
+        if ($request->wantsJson()) {
```
nekodex: seems fine to use `expectsJson` here?



https://github.com/laravel/framework/blob/2906b572aa83ff46b8ad57093dd0e859c0ff783f/src/Illuminate/Http/Request.php#L704

nanaya:
I vaguely remember browsers showing native auth dialog if returning 401.

PR #3290, opened 2018-05-31
Add ability to block users and see blocked users

nekodex:
Will eventually block more stuff going forward, initially will only affect the following:
- user profiles masking of blocked users
- prevents popup usercards of blocked users from being shown
- prevents forum PMs from blocked users (old site only, not 100% if it actually obeys this 😅)
- hides interaction buttons for blocked users (on user profiles, usercards, etc)
- possibly some more stuff on the old site


app/Models/User.php
```diff
@@ -1104,6 +1111,13 @@ public function setPlaymodeAttribute($value)
         $this->osu_playmode = Beatmap::modeInt($value);
     }
 
+    public function hasBlocked(self $user)
```
nanaya: `static`? :zap:

resources/assets/less/bem/user-list.less
```diff
@@ -0,0 +1,32 @@
...
+  align-self: flex-start;
+  flex-grow: 1;
+
+  &__content {
+    max-width: 250px;
+    overflow: hidden;
+    height: auto;
+
+    &--folded {
+      height: 0px;
```
nanaya: Why not just `display: none` so the `overflow: hidden` isn't needed? And pretty sure `height: auto` isn't needed in any case.

Also maybe want to check jquery's `slideToggle` if that's the original intention.

resources/assets/coffee/react/_components/block-button.coffee
```diff
@@ -0,0 +1,108 @@
...
+      else
+        #blocking
+        @xhr = $.ajax
+          type: 'POST'
+          url: laroute.route 'blocks.store', target: @props.user_id
+
+      @xhr
+      .done @updateBlocks
+      .fail osu.emitAjaxError(@button)
+      .always @requestDone
```
nanaya: looks like this is sometimes run a bit too late after unmounted, causing this warning

resources/assets/coffee/react/_components/block-button.coffee
```diff
@@ -0,0 +1,108 @@
...
+
+
+  updateBlocks: (data) =>
+    @setState block: _.find(data, (o) => o.target_id == @props.user_id), ->
+      currentUser.blocks = _.filter data, relation_type: 'block'
+      currentUser.friends = _.filter data, relation_type: 'friend'
+      $.publish 'user:update', currentUser
+      $.publish 'blockButton:refresh'
+      $.publish 'friendButton:refresh'
+      $.publish 'user:page:update'
```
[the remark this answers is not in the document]
nekodex: triggers a user page update (so the profile masking happens)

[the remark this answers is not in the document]
nanaya: is it not triggered by `user:update`?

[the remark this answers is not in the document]
nekodex: apparently not 🤷‍♂️

app/Models/User.php
nekodex: can't use `static` as a param type in php, go figure

resources/views/objects/_usercard.blade.php
```diff
@@ -77,7 +77,7 @@
                             @endif
                             <div class="usercard__icon js-react--friendButton" data-target="{{$user->user_id}}"></div>
 
-                            @if ($user->exists)
+                            @if ($user->exists && !Auth::user()->hasBlocked($user))
```
nanaya: `Auth::user()` may be null.

resources/assets/coffee/_classes/account-edit-blocklist.coffee
```diff
@@ -0,0 +1,46 @@
...
+  constructor: ->
+    $(document).on 'click', @jsClass, @toggle
+    $.subscribe 'user:update', @updateBlockCount
+    @visible = false
+
+
+  updateBlockCount: =>
+    $(@jsClass)
+      .parents('.account-edit-entry')
+      .find('.account-edit-entry__label')
```
nanaya: why not just assign this thing its own js class.

resources/assets/coffee/_classes/account-edit-blocklist.coffee
```diff
@@ -0,0 +1,46 @@
...
+#    osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+#    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#    See the GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+
+class @AccountEditBlocklist
+  bn: 'user-list__content'
```
nanaya: `bn` means `block name` (or at least that's what I mean) but this is element, not block. Also I'd just use `hidden` for toggling the thing.

resources/assets/coffee/_classes/account-edit-blocklist.coffee
```diff
@@ -0,0 +1,46 @@
...
+  jsClass: '.js-account-edit-blocklist'
+
+  constructor: ->
+    $(document).on 'click', @jsClass, @toggle
+    $.subscribe 'user:update', @updateBlockCount
+    @visible = false
```
nanaya: may be wrong when visiting the page again through a link instead of browser back/forward button.

1. Open page
2. toggle (show &rarr; hide)
3. open page again through header (avatar &rarr; settings)
4. toggle again (show &rarr; show)

PR #3317, opened 2018-06-06
Boost scores of beatmap search results if they contain all terms

### LINKED ISSUE #3272: Beatmap Search Function isn't as accurate as in Old Website
Currently, if you search for a beatmap on the new website, it isn't showing as first element the strongest relation with the string given. What I mean is that it looks like matches with tags are as important or even more important than title, artist, source or mapper's name. Let me explain myself better with an example:

If for example you search for 'One Piece', the old website first shows all the beatmaps related to 'One Piece'. However, on the new website, the results are a complete a mess, showing results with no relation to that (tags having 'one' or 'piece' in it). Basicaly it should show results with direct relation to the search string, be it source, title, artist or mapper, after that should be a comparison with tags.

https://osu.ppy.sh/beatmapsets?q=one%20piece vs https://osu.ppy.sh/p/beatmaplist?q=one%20piece

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58851328-beatmap-search-function-isn-t-as-accurate-as-in-old-website?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

notbakaneko:
Increases the scores of result hits when they contain all the search terms (+1 per match) over those that don't.

fixes #3272


nanaya:
looks like if there are more tags and not part of title/source, the result is still lower even though it has all the keywords?



2 has word "to" in source and tags and title
3 has source in japanese (no match), no match for title either, but all the keywords are in tags

Similar thing here

notbakaneko:
Added some dodgy scaling to the partial match scores.
The better option is to normalize the components of the bool query to be of equal weights but that's not currently supported.

PR #3319, opened 2018-06-07
Add build generator to attach unlinked changelogs

nanaya:
Create build on new tag. Repository data (in `repositories` table) is needed for mapping between repository and stream.


nanaya:
Seems working-ish.

PR #3327, opened 2018-06-09
Update scoreboard colouring when hovering over rows

### LINKED ISSUE #3297: Link colour on rows is not readable
image

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/59124538-link-colour-on-rows-is-not-readable?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

MegaApplePi:
resolves #3297


resources/assets/less/bem/beatmap-scoreboard-table.less
```diff
@@ -127,8 +127,14 @@
     }
 
     &:hover {
+      color: #fff;
+
       > td {
-        background-color: #ccc;
+        background-color: #555;
+
+        > a {
```
nanaya: should be inside `__user-link`

resources/assets/less/bem/beatmap-scoreboard-table.less
```diff
@@ -164,6 +170,11 @@
 
   &__user-link {
     color: #555;
+
+    &:hover {
+      font-weight: bold;
```
nanaya: this moves the columns when hovering on the longer name in the table when the table is at minimum size.

resources/assets/less/bem/beatmap-scoreboard-table.less
MegaApplePi: Moving this to the `&__user-link:hover` block means that link colour will always be grey, even when the row is hovered, until the user moves their cursor over the link itself.

nanaya: Inside `.@{_top}__body-row:hover &`.

resources/assets/less/bem/beatmap-scoreboard-table.less
```diff
@@ -127,8 +131,10 @@
     }
 
     &:hover {
+      color: #fff;
```
nanaya: also need different color for `__perfect` (`@green-lighter`).

PR #3336, opened 2018-06-12
Unjustify changelog change text.

### LINKED ISSUE #3323: Changelog text is justified
Happens when the description is long and has multiple lines.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/59398933-changelog-text-is-justified?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

notbakaneko:
fixes #3323 

Also align items to baseline which has the side-effect of aligning the text on mobile to the left instead of centre.

### DIFF
--- a/resources/assets/less/bem/changelog-change.less
+++ b/resources/assets/less/bem/changelog-change.less
@@ -20,7 +20,7 @@
   display: flex;
   flex-direction: row;
   flex-grow: 1;
-  align-items: center;
+  align-items: baseline;
 
   padding: 5px 0;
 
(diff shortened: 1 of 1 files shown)


resources/assets/less/bem/changelog-change.less
```diff
@@ -57,7 +57,6 @@
 
   &__right {
     font-size: @font-size--normal-2;
-    text-align: justify;
 
     @media @mobile {
       text-align: center;
```
nanaya: this probably should be removed as well.

PR #3350, opened 2018-06-15
Alpine image rewrite proposal

### LINKED ISSUE #3349: Use Alpine official Docker Library PHP image
There is a [official Docker library](https://hub.docker.com/_/php/) available on Docker. Recommend we use that for the following reasons:

- Alpine images are smaller than the current used image. This would trim the size of the osu-web image smaller than it would be normally.

- Almost all PHP runtimes are in the official image, tagged. This would function as intended to the current use case of `osu-web`.

I'll be experimenting on this before I do a PR to propose my change.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/59711286-use-alpine-official-docker-library-php-image?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sr229:
This resolves #3349 

Rewrites the dev container for osu-web to use the official Docker PHP image


notbakaneko:
https://github.com/sr229/osu-web/blob/329e5e2f2a6c7a348b5826a7dad9e465817614d6/docker-compose.yml#L47
Do we actually need to set this? This seems to be the host user which may not exist

sr229:
@notbakaneko probably a small oversight and from the old code since the new image defaults to UID 1000 indefinitely. Will remove

nanaya:
Is it actually tested? Also unless something changed, php and node processes inside docker need to be run under same uid as host user and it's not always 1000.

sr229:
@nanaya Based on how other people wrote their PHP stacks, this was the most optimal way based on some of the repositories that uses PHP as well.

LiquidPL:
I'm 99% positive that this will break for people with nonstandard uids.

sr229:
@LiquidPL Your container if has a supplied user, will use that user given ou set the ``USER`` stage. I'm pretty certain even if you run this at macOS it would work well since it's standard OCI container

notbakaneko:
You don't have to keep merging master in if you're not adding any new commits yet, fyi

sr229:
@notbakaneko Kindly elaborate which containers, I might break stuff unintentionally, and I'd need a concentrated list.

nanaya:
For starter, it can't even complete `docker-compose build`.

```
groupadd: invalid group ID 'osuweb'
useradd: invalid user ID '-g'
ERROR: Service 'yarn' failed to build: The command '/bin/sh -c /setup.sh' returned a non-zero code: 3
```

Fixing that, `osuweb-composer` is failing on `up`

```
osuweb-composer  | /usr/local/bin/docker-php-entrypoint: exec: line 9: bash: not found
osuweb-composer exited with code 127
```

Adding `bash`, another command failed because alphine's busybox `cp` doesn't support `-n`.

```
osuweb-composer  | cp: unrecognized option: n
```

Then `osuweb-yarn` and `osuweb-composer` runs everything as root. May want to fix that as well.

```
edho@ubuntu:~/git/osu-web$ ls -ld vendor node_modules public/*/
drwxr-xr-x 957 root root  37k Jun 29 19:04 node_modules/
drwxr-xr-x   2 root root 4.1k Jun 29 19:04 public/css//
drwxrwxr-x   4 edho edho 4.1k Jun 29 19:04 public/fonts//
drwxrwxr-x  12 edho edho 4.1k Jun 22 19:34 public/images//
drwxr-xr-x   3 root root 4.1k Jun 29 19:04 public/js//
drwxr-xr-x   5 root root 4.1k Jun 29 19:04 public/vendor//
drwxr-xr-x   8 root root 4.1k Jun 29 19:12 vendor/
```

What exactly did you test?

sr229:
Alpine uses `ash`, so I was not aware `bash` was a hard requirement. Will add later.

Also I might trim the userID and GIDs too as well

sr229:
Possible workaround on yarn containers is to use `node:alpine` as image.

sr229:
@nanaya can you try again? I tested it on [Docker Playground](https://labs.play-with-docker.com/) and it seems to work fine now.

sr229:
@notbakaneko I already fixed that, somehow, seems to be an edge case, will retry again

PR #3352, opened 2018-06-15
Pagination template cleanup

nanaya:
Initially to be used in changelog page but seems small enough to be pushed out earlier.

v0 because there's another one already without version number but this one predated the other one.

Also attached random fix quotes commit just because.


resources/views/store/index.blade.php
```diff
@@ -18,7 +18,7 @@
 @extends('master')
 
 @section('content')
-    @include("store.header")
+    @include('store.header')
```
notbakaneko: I'd rather we not make these kind of changes in pull requests that have nothing to do with them

nanaya: there's just no place to do this because I don't think I'll do full cleanup on this thing and having pr with just this is dumb.

PR #3354, opened 2018-06-15
Prompt for login when clicking on forum thread reply button

### LINKED ISSUE #2890: Clicking 'reply' in a thread while not logged in (or without permission) should be more... helpful
Two related problems here:
 - When scrolled to the end of a thread, the reply button doesn't appear to do anything (except change colour)
 - When clicking 'reply' mid-thread, a "Please sign in to reply." box appears with no obvious way to login.
 
Maybe disable the reply button in cases where the user is logged in but cannot post due to permissions (with tooltip explaining)... otherwise, show the login box?

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/56499043-clicking-reply-in-a-thread-while-not-logged-in-or-without-permission-should-be-more-helpful?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

notbakaneko:
- Shows sign-in popup if not signed in
- flashes the reply box when clicking reply while at the bottom of a thread.

fixes #2890

### DIFF
--- a/resources/views/forum/topics/show.blade.php
+++ b/resources/views/forum/topics/show.blade.php
@@ -346,9 +346,9 @@
                         </span>
                     </button>
                 @else
-                    <a
-                        href="{{ route('forum.topics.show', $topic) }}"
-                        class="btn-osu-big btn-osu-big--forum-reply js-forum-topic-reply--stick js-login-required--click"
+                    <button
+                        type="button"
+                        class="btn-osu-big btn-osu-big--forum-reply js-user-link"
                     >
                         <span class="btn-osu-big__content">
                             <span class="btn-osu-big__icon">
--- a/resources/views/forum/topics/show.blade.php
+++ b/resources/views/forum/topics/show.blade.php
@@ -361,7 +361,7 @@
                                 </span>
                             </span>
                         </span>
-                    </a>
+                    </button>
                 @endif
             </div>
         </div>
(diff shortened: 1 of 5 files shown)


resources/views/forum/topics/show.blade.php
```diff
@@ -328,22 +328,41 @@ class="btn-circle btn-circle--topic-nav"
...
+                                <span class="btn-osu-big__text-top">
+                                    {{ trans('forum.topics.actions.reply') }}
+                                </span>
+                            </span>
                         </span>
+                    </button>
+                @else
+                    <a
+                        href="{{ route('forum.topics.show', $topic) }}"
+                        class="btn-osu-big btn-osu-big--forum-reply js-forum-topic-reply--stick js-login-required--click"
```
nanaya: probably should just use `js-user-link` (and type button) without the topic reply js at all (it'll still reload once logged in). Also use `fa-sign-in-alt` icon like everywhere else.

resources/views/forum/topics/show.blade.php
```diff
@@ -328,22 +328,41 @@ class="btn-circle btn-circle--topic-nav"
...
-                                {{ trans('forum.topics.actions.reply') }}
+                            <span class="btn-osu-big__left">
+                                <span class="btn-osu-big__text-top">
+                                    {{ trans('forum.topics.actions.login_reply') }}
+                                </span>
                             </span>
                         </span>
-                    </span>
-                </button>
+                    </a>
```
nanaya: wrong closing tag.

PR #3362, opened 2018-06-18
Update placeholder text to show where the user will be posting to.

### LINKED ISSUE #2900: When there is a timestamp at the start of a post in General (All difficulties), it should warn the player where they are posting
Almost no discussions that belong in General (All difficulties) start with a timestamp.
I accidentally posted a whole mod in the "General (all diffs) section again when I should have been posting in the timeline. This has happened on a number of occasions because there is no warning.

If there is a timestamp at the start of a post and the user isn't posting in the timeline panel, it should warn them that timestamps generally belong in the timeline, similar to the "someone else posted an issue at this time" thing
This would be really helpful to avoid posting mods in the wrong tab

Mockup:
screen shot 2018-04-04 at 10 33 52 am

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/56597766-when-there-is-a-timestamp-at-the-start-of-a-post-in-general-all-difficulties-it-should-warn-the-player-where-they-are-posting?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

notbakaneko:
fixes #2900

Decided not to make the textbox state per-mode unless it actually becomes problematic having the text remain while switching tabs.


resources/lang/en/beatmaps.php
```diff
@@ -64,6 +63,12 @@
             'in_timeline' => 'To mod multiple timestamps, post multiple times (one post per timestamp).',
         ],
 
+        'message_placeholder' => [
+            'general' => 'Type here to post to General (:version)',
+            'generalAll' => 'Type here to post to General (All difficulties) ',
+            'timeline' => 'Type here to Timeline (:version)'
```
nanaya: how to timeline ( ･_･)a

(looks like missing some words and a trailing comma)

resources/lang/en/beatmaps.php
```diff
@@ -64,6 +63,12 @@
             'in_timeline' => 'To mod multiple timestamps, post multiple times (one post per timestamp).',
         ],
 
+        'message_placeholder' => [
+            'general' => 'Type here to post to General (:version)',
+            'generalAll' => 'Type here to post to General (All difficulties) ',
```
nanaya: There's a random trailing space in the string.

resources/lang/en/beatmaps.php
notbakaneko: I got distracted 🔪

PR #3325, opened 2018-06-08
Retrieve user best performance scores from elasticsearch

notbakaneko:
Uses elasticsearch aggregations to collate the beatmaps to be displayed.
Indexing of scores will be handled by a https://github.com/ppy/osu-elastic-indexer due to indexing performance via Laravel not being very good.

WIP until the new indices are finished deploying.

fixes #3015

PR #3363, opened 2018-06-18
Retrieve user best performance scores from elasticsearch

### LINKED ISSUE #3015: Some profiles are *very* slow to load
Not all, just some.

Example: https://osu.ppy.sh/users/4937439 and https://osu.ppy.sh/users/3172980 takes around 4-6 seconds compared to <1 second for most.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/57119680-some-profiles-are-very-slow-to-load?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

notbakaneko:
Uses elasticsearch aggregations to collate the beatmaps to be displayed.
Indexing of scores will be handled by a https://github.com/ppy/osu-elastic-indexer due to indexing performance via Laravel not being very good.

fixes #3015

This is #3325 without the updates to the index command since they're not really necessary for this.

Running aggregations sorted by pp for the 1000 users with the most scores in parallel (8 per index) at random, basically grinding the case were the heap gets constantly churned:
```
high_scores_osu: searches: 1000, total: 71036.0 ms, avg: 71.036 ms
high_scores_fruits: searches: 1000, total: 25225.0 ms, avg: 25.225 ms
high_scores_mania: searches: 1000, total: 29515.0 ms, avg: 29.515 ms
high_scores_taiko: searches: 1000, total: 20322.0 ms, avg: 20.322 ms
runtime: 37.400363 seconds
```

without the heap churning, results look more like this:
```
high_scores_osu: searches: 1000, total: 2344.0 ms, avg: 2.344 ms
high_scores_fruits: searches: 1000, total: 104.0 ms, avg: 0.104 ms
high_scores_mania: searches: 1000, total: 283.0 ms, avg: 0.283 ms
high_scores_taiko: searches: 1000, total: 97.0 ms, avg: 0.097 ms
```
---

Improves performance of profile loading.


app/Libraries/Elasticsearch/Search.php
```diff
@@ -201,9 +228,9 @@ private function fetch()
         try {
             return datadog_timing(
                 function () {
-                    return new SearchResponse(Es::search($this->toArray()));
+                    return new SearchResponse($this->client()->search($this->toArray()));
                 },
-                config('datadog-helper.prefix_web').'.search.fetch',
+                config('datadog-helper.prefix_web').".{$this->statTag}.fetch",
```
nekodex: I'd use the tag array below (`[type' => ...`) instead of changing the metric name. Probably better to add another tag for score type in addition to `type` instead of replacing `type`.

app/Libraries/Elasticsearch/Search.php
```diff
@@ -219,7 +246,7 @@ function () {
 
         if (config('datadog-helper.enabled')) {
             Datadog::increment(
-                config('datadog-helper.prefix_web').'.search.errors',
+                config('datadog-helper.prefix_web').".{$this->statTag}.errors",
```
nekodex: As above re: metric name

PR #3365, opened 2018-06-18
Faster topic cache update

nanaya:
- update `DB::transaction` to `$this->getConnection()` in some models
- directly increment/decrement cache counts
- queue heavy updates
- still need to be checked further

### DIFF
--- a/app/Models/Forum/Topic.php
+++ b/app/Models/Forum/Topic.php
@@ -21,8 +21,8 @@
 namespace App\Models\Forum;
 
 use App\Exceptions\ModelNotSavedException;
-use App\Jobs\EsDeleteDocument;
 use App\Jobs\EsIndexDocument;
+use App\Jobs\UpdateUserForumCache;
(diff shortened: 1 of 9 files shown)


app/Models/Forum/Topic.php
```diff
@@ -199,33 +188,31 @@ public function moveTo($destinationForum)
...
             $originForum = $this->forum;
             $this->forum()->associate($destinationForum);
             $this->save();
 
-            $posts = $this->posts()->withTrashed()->with(['forum', 'topic'])->get();
-            $posts->each(function ($post) use ($destinationForum) {
-                $post->forum()->associate($destinationForum);
-                $post->save();
-            });
+            $this->posts()->withTrashed()->update(['forum_id' => $this->forum_id]);
```
notbakaneko: need to disable timestamps otherwise laravel tries to update an empty column with the time

nanaya: das coo

PR #3366, opened 2018-06-18
Fix marking read and poll render of deleted topic

### LINKED ISSUE #3358: Some of deleted threads display on new web.
I assume they are deleted by OP on old web or something similar. They cannot be viewed.
This is visible to moderators only.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/59779964-some-of-deleted-threads-display-on-new-web?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nanaya:
Fixes #3358.


app/Models/Forum/PollOption.php
```diff
@@ -36,6 +36,7 @@ public function post()
     {
         return $this
             ->belongsTo(Post::class, 'topic_id', 'topic_id')
+            ->withTrashed()
```
notbakaneko: having some associations include deleted records and others not without any labeling seems confusing

PR #3372, opened 2018-06-19
Suffix build name with stream name

nanaya:
Forgot about this until I think about how to view build changelog when there are multiple builds from same date.

### DIFF
--- a/app/Models/Build.php
+++ b/app/Models/Build.php
@@ -46,9 +46,11 @@ class Build extends Model
             return;
         }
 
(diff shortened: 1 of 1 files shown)


peppy:
Do the build suffixes actually show up anywhere in the interface? We probably don't want them to (at which point, do we even need them?)

nanaya:
it’s in the url. Displayed version on page doesn’t have it.

nanaya:
And then it's gone :dancer:

PR #3374, opened 2018-06-19
Require stream name to view build

nanaya:
Redirect old urls to first build found.


resources/views/changelog/show.blade.php
```diff
@@ -48,7 +48,7 @@ class="changelog__build-link"
                     @if ($activeBuild->versionNext() !== null)
                         <a
                             class="changelog__build-link"
-                            href="{{ route('changelog.show', ['build' => $activeBuild->versionNext()->version]) }}"
+                            href="{{ build_url($activeBuild) }}"
```
peppy: should this be `->versionNext()`?

PR #3377, opened 2018-06-19
Fix osu! brand naming

Joehuu:


resources/lang/en/community.php
```diff
@@ -23,13 +23,13 @@
         'header' => [
             // size in font-size
             'big_description' => 'Love osu!?<br/>
-                                Support the osu! dev :D',
+                                Support the osu!dev :D',
```
peppy: should probably just be "Support the development of osu! :D" or something similar.

resources/lang/en/layout.php
```diff
@@ -180,7 +180,7 @@
         ],
         'support' => [
```
peppy: is this footer only shown when logged out? if so we can probably just remove this section.

Joehuu:
I have a question. Shouldn't the title of the forums be `osu! forums` not `osu! community`.

peppy:
Yeah, it may be best to rename it now that "community" is a top level header.

Joehuu:
The three other pages under community are not consistent in terms of wording. I think all of them should have `osu!` in front of them, and remove `community` altogether.

I'm also confused about the capitalisation after osu!. Should it be `osu! forums` or `osu! Forums`? Check the others listed.



`osu! Community Contests` -> `osu! Contests`?



`Community Tournaments` -> `osu! Tournaments`?



`Live Streams` -> `osu! Live Streams`?

peppy:
I don't think we need the 'osu!' prefix before "Community Contests"

PR #3379, opened 2018-06-20
Allow limiting of excessive beatmapset downloads

nekodex:
add these to `.env`:
```
BEATMAPSET_USER_DOWNLOAD_LIMIT_HOURLY=x
BEATMAPSET_USER_DOWNLOAD_LIMIT_HOURLY_SUPPORTER=y
```


app/Models/User.php
```diff
@@ -1022,6 +1022,11 @@ public function maxFriends()
         return $this->isSupporter() ? config('osu.user.max_friends_supporter') : config('osu.user.max_friends');
     }
 
+    public function beatmapsetDownloadAllowance()
+    {
+        return $this->isSupporter() ? config('osu.beatmapset.download_limit') : config('osu.beatmapset.download_limit_supporter');
```
peppy: i think this is back to front

nekodex: and this is why we have code reviews!... I need caffeine >.>

PR #3431, opened 2018-06-27
Consistent mode order in beatmap panel

### LINKED ISSUE #2783: Beatmap listing difficulty icons should be ordered by mode
https://osu.ppy.sh/users/4236057
Referencing Shooting Stars, Hydra, and Synchromanica in the image. It seems that the order of mode icons is based on the order of which mode was uploaded first. 
Eg for shooting stars, I uploaded in order of catch, taiko, mania, std.

The order should be std, taiko, catch, mania since 99% of the site does it that way

screen shot 2018-03-23 at 9 37 35 pm

nanaya:
While waiting for the improved listing. Fixes #2783.

---


nanaya:
If only javascript object order can be relied on...

notbakaneko:
Maybe it's time for Map 🤔 

...inb4 doesn't work on IE

PR #3437, opened 2018-06-29
Corrected origin_country

Hickacou:
"De" is more correct than "Depuis" in that case.

Add any details pertaining to developers above the break.

- [ ] Depends on #PR
- Closes #ISSUE

---

Add a sentence or two describing this change in plain english. This will be displayed on the [changelog](https://osu.ppy.sh/home/changelog). A single screenshot or short gif is also welcomed.


nanaya:
Hi, thanks for the correction but this should to through [crowdin](https://crowdin.com/project/osu-web).

PR #3439, opened 2018-06-29
Fix awkward word ending in korean changelog page

Alex4386:
In Korean Translation, changelog page has awkward word ending in the link,
This PR solves the issue that link was awkward like `this`
image
to `this`
image

---

Add a sentence or two describing this change in plain english. This will be displayed on the [changelog](https://osu.ppy.sh/home/changelog). A single screenshot or short gif is also welcomed.


notbakaneko:
translations are done on crowdin; this is probably the one you want https://crowdin.com/translate/osu-web/21/en-ko#466

PR #3463, opened 2018-07-06
Update to latest Twitch API

nanaya:
Now with more objects, abstract game name, ~~and shady regexp~~.

Also fixed featured stream video playback.

---


app/Models/LivestreamCollection.php
```diff
@@ -36,17 +36,50 @@ public static function promote($id)
...
-    public function download()
+    public function downloadStreams()
+    {
+        return $this->download('streams?first=40&game_id=21465');
+    }
+
+    public function downloadUsers($userIds)
+    {
+        if (count($userIds) === 0) {
+        }
```
nekodex: Is there supposed to be something inside this if block? 🤔

app/Models/LivestreamCollection.php
```diff
@@ -36,17 +36,50 @@ public static function promote($id)
...
+
+        $paramString = '';
+
+        foreach ($userIds as $id) {
+            $paramString .= "&id={$id}";
+        }
+
+        return $this->download("users?{$paramString}");
```
nekodex: Won't this result in `users?&id=...`?

app/Models/Twitch/Stream.php
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    public function url()
+    {
+        return "https://twitch.tv/{$this->user['login']}";
+    }
+
+    public function preview($width, $height)
+    {
+        return str_replace(
```
nekodex: Maybe `strtr` instead to save nested `str_replace`?

app/Models/LivestreamCollection.php
nanaya: Yes it will. No, it's not a problem.

app/Models/LivestreamCollection.php
nanaya: That's a good question. And I can't remember what's this for ?_?

app/Models/LivestreamCollection.php
nekodex: Why not just something like `return $this->download('users?id='.join($userIds, '&id='));`?

PR #3465, opened 2018-07-06
Add tooltip for registered players count

### LINKED ISSUE #3461: Add tooltip to "players" icon on tournament
Kinda self explained, since the icon is bit ambiguous (can be translated as number of players who plays or number of players who registered on tournament)
screenshot

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/60532035-add-tooltip-to-players-icon-on-tournament?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nanaya:
Fixes #3461, not sure about the wordings.

---

### DIFF
--- a/resources/lang/en/tournament.php
+++ b/resources/lang/en/tournament.php
@@ -20,12 +20,17 @@
 
 return [
     'index' => [
+        'none_running' => 'There are no tournaments running at the moment, please check back later!',
+        'registration_period' => 'Registration: :start to :end',
+
         'header' => [
             'subtitle' => 'A listing of active, officially-recognised tournaments',
             'title' => 'Community Tournaments',
         ],
-        'none_running' => 'There are no tournaments running at the moment, please check back later!',
-        'registration_period' => 'Registration: :start to :end',
+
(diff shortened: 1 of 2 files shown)


notbakaneko:
`:count players registered` maybe?

nanaya:
The number would be rather redundant? Also other icon description tooltips in user page and beatmapset page don't seem have it either.

notbakaneko:
It's not a call to action nor does clicking on the icon do or show anything specific to the tooltip's description, so, it should really include the number imo; otherwise, the tooltip should be limited specifically to the icon portion only.

nanaya:
So this should be changed to only appear on the icon as well, I guess?

notbakaneko:
yes, it should

PR #3474, opened 2018-07-10
Show different text for users that were never seen

### LINKED ISSUE #3357: "Last seen 48 years ago"
I apologize if this wastes your time in any way.

I was randomly looking at user pages when I randomly stumbled across a user who was "last seen 48 years ago." (specifically, [rlagmlrjjs0's](https://osu.ppy.sh/users/2123086) page.)

It's an empty user page, so it probably doesn't matter much, but I thought I'd point it out. 

Thanks for your time.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/59768874-last-seen-48-years-ago?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

notbakaneko:
fixes #3357

- removes the tooltip is `user_lastvisit` is `0`
- shows `Never` if time is start of epoch

image

---

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -626,6 +626,11 @@ class User extends Model implements AuthenticatableContract, Messageable
         return $this->user_lastvisit > Carbon::now()->subMinutes(config('osu.user.online_window'));
     }
 
+    public function isNeverOnline()
+    {
+        return Carbon::createFromTimestampUTC(0)->diffInSeconds($this->user_lastvisit) === 0;
+    }
+
     public function isPrivileged()
     {
         return $this->isAdmin()
--- a/resources/assets/coffee/osu_common.coffee
+++ b/resources/assets/coffee/osu_common.coffee
@@ -158,9 +158,14 @@
 
   timeago: (time) ->
     el = document.createElement('time')
-    el.classList.add 'timeago'
-    el.setAttribute 'datetime', time
-    el.textContent = time
+
+    if moment(0).isSame(time)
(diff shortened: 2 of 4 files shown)


app/Models/User.php
```diff
@@ -626,6 +626,11 @@ public function isOnline()
         return $this->user_lastvisit > Carbon::now()->subMinutes(config('osu.user.online_window'));
     }
 
+    public function isNeverOnline()
+    {
+        return Carbon::createFromTimestampUTC(0)->diffInSeconds($this->user_lastvisit) === 0;
```
nanaya: `user_lastvisit` should be updated to use `get_time_or_null` helper instead.

resources/assets/coffee/osu_common.coffee
```diff
@@ -158,9 +158,14 @@
 
   timeago: (time) ->
     el = document.createElement('time')
-    el.classList.add 'timeago'
-    el.setAttribute 'datetime', time
-    el.textContent = time
+
+    if moment(0).isSame(time)
```
nanaya: then check if null instead. Not sure if should be here or before this function is called though.

(also wondering why not just `time.getTime() == 0` / `$date->getTimestamp() === 0`)

notbakaneko:
fixed in database

PR #3475, opened 2018-07-10
Changelog index filter

nanaya:
- No UI to do the filtering
- Accepted parameters are `from` (build name), `to` (build name), and `stream` (`name`, not `display_name`)
- The only hint when the page has some kind of filtering is the page (html) title

---


resources/assets/coffee/react/changelog-index/main.coffee
```diff
@@ -16,12 +16,37 @@
...
+    # Example: 2018-07-06T05:43:21+00:00
+    build.created_at.substr(0, 10)
+
+
 class ChangelogIndex.Main extends React.PureComponent
+  constructor: (props) ->
+    super props
+
+    builds = props.data.builds
+    hasMore = builds.length == props.data.search.limit
```
notbakaneko: if the total range has the same number of builds as limit, the pager still shows but clicking it triggers `ujs_redirect` on the controller and the rest of `showMore` bursts into flames (which also makes `always` not get called)

PR #3488, opened 2018-07-13
Change header to show streams instead of builds

### LINKED ISSUE #2700: Changelog page has incorrect user counts
Currently only the latest build from each stream is shown. This means that the user counts don't add up correctly.

old:

new:

Around 3,000 user are missing (which are on the previous version of stable). We should still have all the counts on this page add up to the total users online.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/56165483-changelog-page-has-incorrect-user-counts?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nanaya:
The design is the same but the backing model is changed from build to stream.
Mainly so the users count reflects overall count of that stream instead of one specific build.

Fixes #2700.

Adds quite a bit of queries (user count and latest build) as well. I don't know how to preload them nicely.

---


nekodex:
@arflyte is going to redesign that section header to make it less confusing

PR #3491, opened 2018-07-16
Fix home page dying on news cache retrieval failure

### LINKED ISSUE #3193: The last sane cached content should be used when github is inaccessible
Right now the site stops loading on any pages requesting from github, even if there is potentially usable stale data.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58336755-the-last-sane-cached-content-should-be-used-when-github-is-inaccessible?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

peppy:
Closes #3193.
Stops site falling over when github is down.

Untested, but PRing as a proposal to get this resolved ASAP because it's been sitting unfixed for months but has caused the site to stop working at least twice.

---

### DIFF
--- a/app/Models/News/Index.php
+++ b/app/Models/News/Index.php
@@ -84,14 +84,40 @@ class Index
         return 'news:index:'.static::VERSION;
     }
 
+    public static function cacheKeyFallback()
+    {
+        return 'news:index:'.static::VERSION."-fallback";
+    }
+
     public static function index()
     {
-        return Cache::remember(
-            static::cacheKey(),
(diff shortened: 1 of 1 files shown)


peppy:
See #3493.

PR #3492, opened 2018-07-16
Force index when removing beatmapset high scores

notbakaneko:
force index beatmap_score_lookup

---

### DIFF
--- a/app/Jobs/RemoveBeatmapsetBestScores.php
+++ b/app/Jobs/RemoveBeatmapsetBestScores.php
@@ -69,7 +69,7 @@ class RemoveBeatmapsetBestScores implements ShouldQueue
             $class::whereIn('beatmap_id', $beatmapIds)
                 ->orderBy('score_id')
                 ->where('score_id', '<=', $this->maxScoreIds[$mode])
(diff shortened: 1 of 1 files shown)


app/Jobs/RemoveBeatmapsetBestScores.php
```diff
@@ -63,9 +64,12 @@ public function handle()
         $beatmapIds = model_pluck($this->beatmapset->beatmaps(), 'beatmap_id');
 
         foreach (Beatmap::MODES as $mode => $_modeInt) {
-            static::scoreClass($mode)::whereIn('beatmap_id', $beatmapIds)
+            $class = static::scoreClass($mode);
+            $table = (new $class)->getTable();
+            $class::whereIn('beatmap_id', $beatmapIds)
                 ->orderBy('score_id')
                 ->where('score_id', '<=', $this->maxScoreIds[$mode])
+                ->from(DB::raw("{$table} FORCE INDEX (beatmap_score_lookup)"))
```
peppy: probably best to leave the old version commented out and mention that it can be reverted once we switch all database servers to MySQL 5.7 (is an issue only in 5.6).

PR #3493, opened 2018-07-16
Cache extra copy of news index with longer ttl

### LINKED ISSUE #3193: The last sane cached content should be used when github is inaccessible

notbakaneko:
Thanks github

fixes #3193

---

### DIFF
--- a/app/Models/News/Index.php
+++ b/app/Models/News/Index.php
@@ -95,12 +95,10 @@ class Index
                 function () {
                     $content = array_reverse(OsuWiki::fetch('news'));
 
-                    Cache::remember(
+                    Cache::put(
                         static::cacheKey().'-long',
-                        static::CACHE_DURATION_LONG,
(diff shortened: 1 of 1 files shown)


notbakaneko:
of course `put` and `remember` have different param orders, because that makes sense /s

PR #3497, opened 2018-07-17
Add hover effect on the clickable username greeting

### LINKED ISSUE #3473: No hover effect on clickable username greeting
Maybe add an underline when hovering.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/60685084-no-hover-effect-on-clickable-username-greeting?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TPGPL:
- Closes #3473 

*might look a bit weird though*

---


notbakaneko:
Yeah, we're not doing this because it does look weird

notbakaneko:
no wait, that was something else

resources/views/home/user.blade.php
```diff
@@ -22,7 +22,7 @@
         'title' => trans('home.user.header.welcome', ['username' => link_to(
             route('users.show', Auth::user()),
             Auth::user()->username,
-            ['class' => 'link link--white link--no-underline']
```
nekodex: `.link--underline` is not a valid class, you probably want to just remove it. `.link--no-underline` is what disables the underline on hover, so removing this class will do what you want.

TPGPL:
Applied

PR #3507, opened 2018-07-20
Switch to using forked laravel-datadog-helper with url tagging disabled

nekodex:
Fixes issue with Datadog not liking metrics that have large numbers of unique tags on them - each request was getting tagged with the full uri requested... which resulted in lots of tags.

(also includes composer dependency updates)

---

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,10 +12,14 @@
     {
       "type": "vcs",
       "url": "https://github.com/nanaya/slack-php"
+    },
+    {
+      "type": "vcs",
+      "url": "https://github.com/nekodex/laravel-datadog-helper"
     }
   ],
   "require": {
-    "chaseconey/laravel-datadog-helper": "*",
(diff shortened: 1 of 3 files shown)


nanaya:
Now I think again, we probably should

- put forks under `ppy`
- have the changes reviewed
- open issue/pull request to merge back upstream?

nekodex:
Not sure about the forking under ppy instead of personal account thing, feels like that would just  result in random libraries (that we potentially touch only once) cluttering up the organisation?

I already created a pull request to upstream, just pointing at my own branch until (when/if) changes are merged upstream. You can review the changes there if you really want.

This feels like a slippery slope though, are we going to scrutinise changes to any third-party library we touch by our own code standards? Surely this is the code maintainer's choice not ours?

nanaya:
Looks like we already have some forks here?

Having someone else to check would be probably slightly more useful than nothing at all? I vaguely remember finding a bug in one of changes of a fork.

PR #3522, opened 2018-07-26
Split apart forum topic moderating permissions

cl8n:
uses `m_lock` for locking permission and `f_sticky` for pinning (I'm not sure if these are the right permissions, I just found them on phpbb's wiki)

some groups have permission to lock/pin topics on specific forums, but none of those acl permissions were being checked. this change makes it so groups other than qat/gmt can lock/pin in their forums. this was a problem because members of the project loved group couldn't lock/pin in the project loved forum

---


nekodex:
Hey, first of all, thanks for the contribution! 😃

So the reasoning behind not checking the ACL is because we're trying to (eventually) move away from using the phpBB ACL altogether in an effort to simplify permissions across the system (i.e. no more arbitrary users with unique special permissions). Per-forum moderation wasn't really used, so we missed that while implementing the new forum.

After some internal discussion, the direction we want to go with this is to simplify things to just grant specific groups the ability to moderate a subforum (instead of checking permissions per-user/per-action). e.g. by adding a column to `phpbb_forums` with a serialized list of `group_id`s that can moderate the forum and checking that in `OsuAuthorize` instead of using ACL `m_lock`/`f_stick`/etc checks.

If you would like to rework your PR to go in this direction it would be greatly appreciated - if not that's fine, let us know and one of the team can take over instead.

cl8n:
yeah I'll give it a shot

will there be any way (/should I add a way) for admins to set which groups can moderate specific forums?

peppy:
Just via database is fine; it's very rare we make such changes.

cl8n:
alright, got rid of the acl stuff and added a column in phpbb_forums like you suggested. I kept the override for GMT and QAT so they don't have to be added to every forum manually

app/Models/Forum/Forum.php
```diff
@@ -139,6 +139,18 @@ public function scopeDisplayList($query)
...
+    public function getModeratorGroupsAttribute($value)
+    {
+        if (present($value)) {
+            return unserialize($value);
+        }
+    }
+
+    public function setModeratorGroupsAttribute($value)
+    {
+        $this->attributes['moderator_groups'] = ($value === null || count($value) === 0) ? '' : serialize($value);
```
nanaya: Should just use laravel's `$casts`. And the column type should be `text`.

nekodex:
Oh, probably should also edit `checkForumTopicStore` and `checkForumTopicReply` in `OsuAuthorize` to allow moderators to post in the relevant forums too.

nanaya:
Just noticed delete button is missing because of missing new check in javascript side.

cl8n:
@nanaya not sure if this was the best way to handle it. I wanted to use the ForumModerate check in js without re-writing the whole thing (since it's a bit more complicated now)

database/migrations/2018_07_26_211352_add_moderating_groups_to_forums.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('phpbb_forums', function (Blueprint $table) {
+            $column = $table->text('moderator_groups')->nullable();
```
nanaya: no need to assign.

Issue #2592, opened 2018-03-01
Add a notice to pages stating there is additional functionality if you're logged in.


nekodex:
This supersedes/fixes #2376 and #2217.

Assigned to @arflyte for design.

arflyte:
![photoshop_2018-03-02_16-54-02](https://user-images.githubusercontent.com/3632750/36889006-62060bb2-1e3b-11e8-952e-f6681de53a3b.png)

Floating panel that sticks to the top of the screen. Someone need to rewrite the stuff in the box and describe stuff.

![photoshop_2018-03-02_17-00-24](https://user-images.githubusercontent.com/3632750/36889056-8cc49bde-1e3b-11e8-80e4-3967998e6907.png)

The yellow bar trails back to the avatar.

peppy:
Is this design not current? What was the reason for not implementing this?

peppy:
I thought it was going to appear when something was clicked that required login. If it is always there that's a bit obnoxious, agreed.

arflyte:
It was designed to only appear when something was clicked. If you want something permanent, then I'll have to redesign it to make it less obnoxious, but it won't grab much attention telling user to login.

So, you guys want a permanent notice or non permanent notice? I prefer the latter.

notbakaneko:
Currently, clicking on things that require login will show the actual sign-in form ready for input, which is probably better than a message and having to make another click to bring up the sign-in form.

The issue is when the functionality isn't originally visible when not logged in, like for the beatmap download button where there isn't anything to even click on in the first place.

arflyte:
![photoshop_180809-192657](https://user-images.githubusercontent.com/3632750/43893530-3f51b1f0-9c0a-11e8-885f-86097508fcc4.png)

Permanent bar design.

LiquidPL:
Can I get that lock symbol in full quality (or maybe just the psd, if that's easier)?

peppy:
@LiquidPL if you’re looking to implement this, just use a generic fontawesome icon for the time being. @arflyte please provide logo if you see this.

PR #3528, opened 2018-07-26
Feature/search require login placeholder

### LINKED ISSUE #3279: Beatmap listing should make it obvious that login is required to search
Currently there's nothing on the page mentioning that you need to login to search:

At very least, there should be a sub-heading saying "login to search". #2592 related, although further discussion is required if we use that implementation.

notbakaneko:
This was discussed a while ago with @arflyte when implementing one of the previous items for the beatmap search page, that we'd put a placeholder search box for the time being until #2592 is ready.

closes #3279

---

### DIFF
--- a/resources/assets/coffee/react/beatmaps/search-panel.coffee
+++ b/resources/assets/coffee/react/beatmaps/search-panel.coffee
@@ -52,14 +52,19 @@ class Beatmaps.SearchPanel extends React.PureComponent
 
   renderGuest: =>
     div
-      className: 'osu-page-header osu-page-header--beatmapsets-header-guest'
+      className: 'beatmapsets-search'
       div
         className: 'osu-page-header__background'
         style:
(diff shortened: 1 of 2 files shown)


resources/assets/coffee/react/beatmaps/search-panel.coffee
```diff
@@ -52,14 +52,18 @@ class Beatmaps.SearchPanel extends React.PureComponent
...
+      className: 'beatmapsets-search'
       div
         className: 'osu-page-header__background'
         style:
           backgroundImage: "url(#{@props.background})"
-      h1
-        className: 'osu-page-header__title'
-        'Beatmaps'
+      div className: 'fancy-search fancy-search--beatmapsets js-user-link',
+        input
```
nekodex: You can tab into the box and type while not triggering the popup. Maybe disable the input box or trigger the popup on enter or something?

PR #3529, opened 2018-07-27
Allows beatmaps to be set as 'loved'

### LINKED ISSUE #3344: Allow beatmaps to be set as 'loved'
Should be added as an extension to the beatmap modding system via a new button that behaves similarly to 'Nominate'. (i.e. confirmation, log action to history, etc)

Access should be limited to GMT or higher in addition to the 'Project Loved' group.

This supersedes PR #3271

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/59610726-allow-beatmaps-to-be-set-as-loved?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

notbakaneko:
closes #3344

---

### DIFF
--- a/app/Http/Controllers/BeatmapsetsController.php
+++ b/app/Http/Controllers/BeatmapsetsController.php
@@ -259,7 +259,7 @@ class BeatmapsetsController extends Controller
     {
         $beatmapset = Beatmapset::findOrFail($id);
 
-        priv_check('BeatmapsetNominate', $beatmapset)->ensureCan();
+        priv_check('BeatmapsetLove', $beatmapset)->ensureCan();
 
         $nomination = $beatmapset->love(Auth::user());
         if (!$nomination['result']) {
--- a/app/Http/Controllers/BeatmapsetsController.php
+++ b/app/Http/Controllers/BeatmapsetsController.php
@@ -255,6 +255,26 @@ class BeatmapsetsController extends Controller
         return $beatmapset->defaultDiscussionJson();
     }
 
+    public function love($id)
+    {
+        $beatmapset = Beatmapset::findOrFail($id);
+
+        priv_check('BeatmapsetNominate', $beatmapset)->ensureCan();
+
+        $nomination = $beatmapset->love(Auth::user());
+        if (!$nomination['result']) {
+            return error_popup($nomination['message']);
+        }
+
+        BeatmapsetWatch::markRead($beatmapset, Auth::user());
+        (new NotifyBeatmapsetUpdate([
+            'user' => Auth::user(),
+            'beatmapset' => $beatmapset,
+        ]))->delayedDispatch();
+
+        return $beatmapset->defaultDiscussionJson();
+    }
+
     public function update($id)
     {
         $beatmapset = Beatmapset::findOrFail($id);
--- a/app/Libraries/OsuAuthorize.php
+++ b/app/Libraries/OsuAuthorize.php
@@ -333,7 +333,7 @@ class OsuAuthorize
 
         static $prefix = 'beatmap_discussion.nominate.';
 
-        if (!$user->isGMT() && !$user->isQAT() && !$user->isGroup(UserGroup::GROUPS['loved'])) {
+        if (!$beatmapset->canBeLovedBy($user)) {
             return 'unauthorized';
         }
 
--- a/app/Libraries/OsuAuthorize.php
+++ b/app/Libraries/OsuAuthorize.php
@@ -23,11 +23,11 @@ namespace App\Libraries;
 use App\Exceptions\AuthorizationException;
 use App\Models\Beatmapset;
 use App\Models\BeatmapsetEvent;
-use App\Models\UserGroup;
 use App\Models\Chat\Channel as ChatChannel;
 use App\Models\Forum\Authorize as ForumAuthorize;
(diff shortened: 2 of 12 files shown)


app/Http/Controllers/BeatmapsetsController.php
```diff
@@ -255,6 +255,26 @@ public function nominate($id)
...
+
+        priv_check('BeatmapsetLove', $beatmapset)->ensureCan();
+
+        $nomination = $beatmapset->love(Auth::user());
+        if (!$nomination['result']) {
+            return error_popup($nomination['message']);
+        }
+
+        BeatmapsetWatch::markRead($beatmapset, Auth::user());
+        (new NotifyBeatmapsetUpdate([
```
cl8n: mark read &notify aren't necessary

app/Transformers/BeatmapsetTransformer.php
```diff
@@ -121,6 +121,7 @@ public function includeCurrentUserAttributes(Beatmapset $beatmapset)
         $ret = [
             'can_hype' => $hypeValidation['result'],
             'can_hype_reason' => $hypeValidation['message'] ?? null,
+            'can_love' => $beatmapset->isLoveable() && ($currentUser->isAdmin() || $currentUser->isGMT() || $currentUser->isQAT()),
```
cl8n: missing check for loved user group

app/Models/Beatmapset.php
```diff
@@ -544,6 +566,26 @@ public function nominate(User $user)
...
+            return [
+                'result' => false,
+                'message' => trans('beatmaps.nominations.incorrect_state'),
+            ];
+        }
+
+        $this->getConnection()->transaction(function () use ($user) {
+            $this->events()->create(['type' => BeatmapsetEvent::LOVE, 'user_id' => $user->user_id]);
+            $this->update(['approved' => static::STATES['loved']]);
+            $this->beatmaps()->update(['approved' => static::STATES['loved']]);
```
cl8n: this and the line above it should be replaced with `$this->setApproved('loved', $user)` (it also sets the approver id)

app/Models/Beatmapset.php
```diff
@@ -544,6 +566,26 @@ public function nominate(User $user)
...
+    public function love(User $user)
+    {
+        if (!$this->isLoveable()) {
+            return [
+                'result' => false,
+                'message' => trans('beatmaps.nominations.incorrect_state'),
+            ];
+        }
+
+        $this->getConnection()->transaction(function () use ($user) {
```
cl8n: I think this transaction should include the same tasks as qualifying a beatmap? (delete user ratings, check covers, delete scores, make a global event)

notbakaneko: loving skips most of that apparently

cl8n: well, it skips all of it atm because maps are only loved manually. seems like it should do all of that though?

app/Http/Controllers/BeatmapsetsController.php
nekodex: If you're watching a beatmapset, I'd argue that you'd probably want to be notified if it becomes loved?

app/Models/Event.php
```diff
@@ -29,7 +29,7 @@ class Event extends Model
     public $patterns = [
         'achievement' => "!^(?:<b>)+<a href='(?<userUrl>.+?)'>(?<userName>.+?)</a>(?:</b>)+ unlocked the \"<b>(?<achievementName>.+?)</b>\" achievement\!$!",
         'beatmapPlaycount' => "!^<a href='(?<beatmapUrl>.+?)'>(?<beatmapTitle>.+?)</a> has been played (?<count>[\d,]+) times\!$!",
-        'beatmapsetApprove' => "!^<a href='(?<beatmapsetUrl>.+?)'>(?<beatmapsetTitle>.+?)</a> by <b><a href='(?<userUrl>.+?)'>(?<userName>.+?)</a></b> has just been (?<approval>ranked|approved|qualified)\!$!",
+        'beatmapsetApprove' => "!^<a href='(?<beatmapsetUrl>.+?)'>(?<beatmapsetTitle>.+?)</a> by <b><a href='(?<userUrl>.+?)'>(?<userName>.+?)</a></b> has just been (?<approval>ranked|approved|qualified|loved)\!$!",
```
nekodex: Will this render correctly on the old site?

app/Libraries/OsuAuthorize.php
```diff
@@ -326,6 +326,27 @@ public function checkBeatmapDiscussionPostStore($user, $post)
...
+    {
+        $this->ensureLoggedIn($user);
+
+        static $prefix = 'beatmap_discussion.nominate.';
+
+        if (!$beatmapset->canBeLovedBy($user)) {
+            return 'unauthorized';
+        }
+
+        if (!$beatmapset->isLoveable()) {
```
nekodex: I'd move the beatmapset state check out and then combine `Beatmapset::canBeLovedBy` with this method, otherwise admins will be able to bypass the state check...

PR #3532, opened 2018-07-27
Alias index after index group is finished

### LINKED ISSUE #3515: Incorrect forum posts index alias creation timing
When creating new posts index, the alias pointing to new index is created too early - after indexing topics - instead of after all indexing (posts) is done.

Currently it goes like this:

1. create new index
2. index topics
3. update alias to new index &larr; except there's no posts yet
4. index posts

While it should be:

1. create new index
2. index topics
3. index posts
4. update alias to new index

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/61261879-incorrect-forum-posts-index-alias-creation-timing?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

notbakaneko:
- Workaround to update the index alias after indexing the group instead of after the first type.
- Checks if all types in the group are supposed to be in the same index.

fixes #3515

---


app/Console/Commands/EsIndexDocuments.php
```diff
@@ -98,55 +100,59 @@ protected function finish(array $indices, array $oldIndices)
...
+                $type::esReindexAll(static::BATCH_SIZE, 0, $options, function ($progress) use ($bar) {
                     $bar->setProgress($progress);
                 });
             }
 
-            if ($i === 0) {
-                $indices[] = $type::esIndexName();
+            $bar->finish();
+
+            if (!$this->inplace
```
nanaya: or just outside the loop? (and maybe revert back using `=== 0` check for the first one)

PR #3541, opened 2018-07-30
Require separator to show changelog

nanaya:


tests/Models/ChangelogEntryTest.php
```diff
@@ -0,0 +1,52 @@
...
+        $entry->message = "Hidden\n---";
+        $this->assertSame(null, $entry->messageHTML());
+
+        $entry->message = "Hidden\n\n---";
+        $this->assertSame(null, $entry->messageHTML());
+
+        $entry->message = "Hidden\n\n---Still hidden";
+        $this->assertSame(null, $entry->messageHTML());
+
+        $entry->message = "Hidden\n---\n\nStill hidden";
```
peppy: any reason for not allowing this one? i've seen some people do this occasionally

nanaya: Because it's markdown for heading.

peppy: was referring to the `\n\n`, but resolved IRL (this test is testing the preceding part only).

PR #3548, opened 2018-08-01
Allow users to hide their online status

### LINKED ISSUE #3544: Ability to hide "last online" timestamp
I've had multiple requests for this. Some people don't want others to know when they last visited the site. Understandable.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/61494939-ability-to-hide-last-online-timestamp?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cl8n:
resolves #3544 

still updates last visit behind-the-scenes, but doesn't show it while appearing offline

---


cl8n:
yeah I agree that makes more sense. I'll do it like how `user_allow_pm` is flipped into `pm_friends_only`

app/Models/User.php
```diff
@@ -443,12 +444,27 @@ public function setPmFriendsOnlyAttribute($value)
         $this->user_allow_pm = !$value;
     }
 
+    public function getHidePresenceAttribute($value)
```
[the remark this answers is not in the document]
nekodex: btw accessors/getters should be using `$value` instead of `$this->...`

nekodex:
BTW the renamed attribute in `UserTransformer` will also modify API responses that include `User` objects. Whoever merges this, make sure to poke @peppy first to have him update the game client.

cl8n:
I think it makes more sense to have a new accessor for the displayed date, because if you have to keep calling `$attributes['user_lastvisit']` you'd need to run it through a Carbon helper every time... kinda defeating the point of accessors

easier to read the code too, fwiw

resources/views/objects/_usercard.blade.php
```diff
@@ -94,7 +94,7 @@ class="user-action-button user-action-button--message"
             </div>
             <div class="usercard__status-bar usercard__status-bar--{{!isset($loading) && $user->isOnline() ? 'online' : 'offline'}}">
                 <span class="far fa-fw fa-circle usercard__status-icon"></span>
-                <span class="usercard__status-message" title="{{isset($loading) || $user->isOnline() ? '' : ($user->user_lastvisit ? trans('users.show.lastvisit', ['date' => $user->user_lastvisit->diffForHumans()]) : '')}}">
+                <span class="usercard__status-message" title="{{isset($loading) || $user->isOnline() || $user->hide_presence ? '' : trans('users.show.lastvisit', ['date' => $user->user_lastvisit->diffForHumans()])}}">
```
notbakaneko: probably should extract the check logic for the entire usercard block into a variable like `$showOnline = !isset($loading) && $user->isOnline();` and use that instead.

resources/views/objects/_usercard.blade.php
```diff
@@ -92,10 +94,10 @@ class="user-action-button user-action-button--message"
                     </div>
                 </div>
             </div>
-            <div class="usercard__status-bar usercard__status-bar--{{!isset($loading) && $user->isOnline() ? 'online' : 'offline'}}">
+            <div class="usercard__status-bar usercard__status-bar--{{ $showOnline ? 'online' : 'offline' }}">
                 <span class="far fa-fw fa-circle usercard__status-icon"></span>
-                <span class="usercard__status-message" title="{{isset($loading) || $user->isOnline() || $user->hide_presence ? '' : trans('users.show.lastvisit', ['date' => $user->user_lastvisit->diffForHumans()])}}">
-                    {{!isset($loading) && $user->isOnline() ? trans('users.status.online') : trans('users.status.offline')}}
+                <span class="usercard__status-message" title="{{ isset($loading) || is_null($user->displayed_last_visit) ? '' : trans('users.show.lastvisit', ['date' => $user->user_lastvisit->diffForHumans()]) }}">
```
notbakaneko: `$showOnline ? blah : ''`
Then it's uniform with the other parts of the usercard ?_?

cl8n: if they're online, it isn't supposed to show anything

Aergwyn:
> Whoever merges this, make sure to poke peppy first to have him update the game client.

Is this still relevant?

notbakaneko:
there should be a fix ready to go

PR #3550, opened 2018-08-01
Prevent csrf and session cookies from being sent in API responses.

nekodex:
as per title

---

### DIFF
--- a/app/Http/Middleware/VerifyCsrfToken.php
+++ b/app/Http/Middleware/VerifyCsrfToken.php
@@ -36,12 +36,6 @@ class VerifyCsrfToken extends BaseVerifier
     public function handle($request, Closure $next)
     {
         try {
-            foreach ($this->except as $except) {
-                if ($request->is($except)) {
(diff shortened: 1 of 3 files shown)


app/Http/Middleware/VerifyCsrfToken.php
```diff
@@ -28,13 +28,20 @@
...
     protected $except = [
+        'api/*',
         'oauth/authorize',
         'oauth/access_token',
     ];
 
     public function handle($request, Closure $next)
     {
         try {
+            foreach ($this->except as $except) {
```
nanaya: already done in `parent::handle`?

nekodex: the parent will skip the csrf check, but still sets the csrf token cookie...

https://github.com/laravel/framework/blob/ab62f037fe7d2b611722801f73f2541a1c7553ec/src/Illuminate/Foundation/Http/Middleware/VerifyCsrfToken.php#L67

PR #3554, opened 2018-08-02
Force image dimensions to be rechecked when contests are made live

nekodex:
Resolves an issue where images (that were not yet uploaded prior to the contest page being accessed for the first time) were being cached as a 404, which resulted in having no image dimension metadata (which broke the javascript image viewer relying on this metadata).

---


nanaya:
I wonder if it's better to add option for `fast_imagesize` to not cache on error instead.

peppy:
Have to consider that this fix would handle more than just 404s (ie. if the image was updated/fixed but the filename didn't have a timestamp update). I think it's nice to have that.

nanaya:
Note this only fix image dimension. Actual url isn't prefixed so cloudflare cache will still need to be purged.

PR #3564, opened 2018-08-06
Let users move their own maps to loved

cl8n:
removes leftover check from nominating a beatmap

---


notbakaneko:
I'm not sure if loving a beatmap should be excepted from owner checks

peppy:
Should be okay; there are few members which are able to use this action and they are generally trusted with knowing what they are doing.

cl8n:
maps are only moved to loved when they pass through a whole system of nomination and community voting anyway, it shouldn't matter if the map happens to be owned by a member of the group

Issue #2583, opened 2018-02-28
Can't report plays on beatmap top rankings


Joehuu:
You can't report plays also.
![report](https://user-images.githubusercontent.com/35318437/36804939-e75977f8-1c70-11e8-9fb6-84baf32dc5d2.png)

arflyte:
![324a346eb5](https://user-images.githubusercontent.com/3632750/36827289-036c7f04-1d56-11e8-8ad2-d6ef83f87eed.png)

The design is there, but not implemented yet

cl8n:
where/how are score reports stored?

peppy:
````
CREATE TABLE `osu_user_reports` (
  `report_id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `user_id` int(11) NOT NULL,
  `score_id` int(11) unsigned NOT NULL DEFAULT '0',
  `mode` tinyint(11) NOT NULL DEFAULT '0',
  `reason` enum('Insults','Spam','Cheating','UnwantedContent','Nonsense','Other') NOT NULL DEFAULT 'Cheating',
  `reporter_id` int(11) NOT NULL,
  `comments` text,
  `timestamp` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
  PRIMARY KEY (`report_id`),
  UNIQUE KEY `unique` (`reporter_id`,`user_id`,`score_id`),
  KEY `timestamp` (`timestamp`),
  KEY `user_lookup` (`user_id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8 ROW_FORMAT=DYNAMIC;
```

notbakaneko:
@cl8n is working on this but github is not giving the option to set the assignment to him

peppy:
@cl8n are you still looking into this one?

cl8n:
working on it right now o/

I will open a PR very soon

Issue #2580, opened 2018-02-28
Cannot report users


ThePooN:
There's currently no way that I'm aware of to report an user from their profile. A quick form like on the older website would be a welcomed addition, as the only other way to report someone is in-game.

peppy:
@arflyte 

previously looked like this:

![Pasted Image 1.png](https://shipusercontent.com/bf94684695ab327a088c9e5c9cff83f5/Pasted%20Image%201.png)

![Pasted Image 3.png](https://shipusercontent.com/6b77650325f993fcac6c7c37ab1b9ca5/Pasted%20Image%203.png)

![Pasted Image 4.png](https://shipusercontent.com/221de1b606307f6c872c5b259df260f9/Pasted%20Image%204.png)

peppy:
This is top priority and needs to be completed before anything else. Is anyone already working on this?

peppy:
@arflyte is a design complete for this even? if not, drop *everything* and make it happen today.

arflyte:
Odd, I thought someone already implemented it. My bad, missed this. Here's the design.

![photoshop_180806-135547](https://user-images.githubusercontent.com/3632750/43697680-c3373e04-9980-11e8-9c2a-c8d13bf50643.png)
![photoshop_180806-135626](https://user-images.githubusercontent.com/3632750/43697685-c5bed916-9980-11e8-9f9b-34e56be933c8.png)

PR #3567, opened 2018-08-07
Add base table for osu_user_reports

notbakaneko:
Base table and model for #2580 and #2583

---

### DIFF
--- a/app/Models/UserReport.php
+++ b/app/Models/UserReport.php
@@ -1,0 +1,44 @@
+<?php
+
+/**
+ *    Copyright 2015-2018 ppy Pty. Ltd.
+ *
+ *    This file is part of osu!web. osu!web is distributed with the hope of
+ *    attracting more community contributions to the core ecosystem of osu!.
+ *
+ *    osu!web is free software: you can redistribute it and/or modify
+ *    it under the terms of the Affero GNU General Public License version 3
+ *    as published by the Free Software Foundation.
+ *
+ *    osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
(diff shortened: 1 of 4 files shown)


app/Models/UserReport.php
```diff
@@ -0,0 +1,51 @@
...
+    public $timestamps = false;
+
+    public function reporter()
+    {
+        return $this->belongsTo(User::class, 'reporter_id');
+    }
+
+    public function score()
+    {
+        return $this->belongsTo(BestModel::getClass($this->mode), 'score_id');
```
nanaya: as I recently learned, using `$this` in association definition doesn't work for `with()` preloads.

notbakaneko: could use `morphTo` except the `int` mode breaks things especially when it's `0`

app/Models/UserReport.php
```diff
@@ -0,0 +1,51 @@
...
+
+    protected $table = 'osu_user_reports';
+    protected $primaryKey = 'report_id';
+    protected $guarded = [];
+
+    protected $dates = ['timestamp'];
+
+    public $timestamps = false;
+
+    public function reporter()
```
cl8n: I don't think any of these relationships will be used

notbakaneko: Defining the existing associations now instead of later.
Also makes it easier to check for things when using `tinker`

app/Models/UserReport.php
notbakaneko: I don't think it'll be an issue, yet; to use `morphTo` we'll probably have to change the data type, otherwise we might have to make a separate association for each `mode` or something :|

PR #3569, opened 2018-08-08
Update supporter page artwork with new commission

peppy:


resources/assets/less/bem/stg-header.less
```diff
@@ -18,7 +18,7 @@
 
 .stg-header {
   .default-box-shadow();
-  background-image: url('/images/headers/community-support.png');
```
nanaya: looks like whoever wrote this initially forgot to use `.at2x-simple()` instead of this.

nanaya:
There's another problem with positioning on mobile but it's been a problem anyway so it shall be fixed separately.

PR #3570, opened 2018-08-08
Add ability to report users

### LINKED ISSUE #2580: Cannot report users

notbakaneko:
Just creates reports, doesn't do anything fancy with them

closes #2580

depends on #3567

---


app/Libraries/OsuAuthorize.php
```diff
@@ -891,6 +891,13 @@ public function checkUserFavouriteRemove($user)
         return 'ok';
     }
 
+    public function checkUserReport($user)
+    {
+        $this->ensureCleanRecord($user);
```
cl8n: why is a clean record required for this?

peppy: don't think this is required

app/Http/Controllers/UsersController.php
```diff
@@ -178,6 +180,34 @@ public function recentActivity($_userId)
...
     }
 
+    public function report($id)
+    {
+        $user = User::lookup($id);
+        if ($user === null || !priv_check('UserShow', $user)->can()) {
+            return response()->json([], 404);
+        }
+
+        // ignore reports from users without a clean record.
```
cl8n: well, this comment doesn't apply anymore

nitpicky review lol

PR #3581, opened 2018-08-16
Add store order history and checkout improvements

notbakaneko:
Based on a really old design.

- show user their store order history,
- no longer a single cart checkout system; this is largely to handle people starting the checkout process but not completing it, and the delays in the triggering of the callback by payment processors.
- explicitly mark if items have reserved stock instead of trying to guess from status.

Cancellation of old uncompleted orders will be added later.

---

Users can now view their own order history.


routes/web.php
```diff
@@ -232,8 +232,9 @@
         Route::resource('cart', 'CartController', ['only' => ['store']]);
 
         Route::delete('checkout', 'CheckoutController@destroy')->name('checkout.destroy');
-        Route::get('checkout', 'CheckoutController@show')->name('checkout.show');
-        Route::resource('checkout', 'CheckoutController', ['only' => ['store']]);
+        Route::resource('checkout', 'CheckoutController', ['only' => ['show', 'store']]);
+
+        Route::resource('orders', 'OrdersController', ['only' => ['index', 'show']]);
```
nanaya: no show action in controller?

resources/assets/less/bem/store-order.less
```diff
@@ -0,0 +1,103 @@
...
+    position: absolute;
+    top: 0;
+    left: 0;
+    height: 100%;
+    width: @border-radius-base;
+    border-radius: @border-radius-base 0 0 @border-radius-base;
+  }
+
+  &__amount {
+    font-size: 14px;
```
nanaya: there's this fancy `@font-size--<size>` variables (also for other font sizes in this file)

app/Models/Store/Order.php
```diff
@@ -262,6 +262,11 @@ public function canCheckout()
         return in_array($this->status, ['incart', 'processing'], true);
     }
 
+    public function hasInvoice()
```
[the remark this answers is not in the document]
notbakaneko: no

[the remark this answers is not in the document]
nanaya: but it currently is?

resources/views/store/checkout.blade.php
```diff
@@ -37,21 +37,10 @@
...
-                            'data-confirm' => trans('common.confirmation'),
-                            'data-remote' => '1',
-                        ]
-                    );
-                @endphp
                 <div class="alert alert-danger">
                     <p>
                         {{ trans('store.checkout.pending_checkout.line_1') }}<br>
-                        {!! trans('store.checkout.pending_checkout.line_2', ['link' => $cancelLink]) !!}
+                        {{ trans('store.checkout.pending_checkout.line_2') }}
```
nanaya: no way to cancel anymore? And then there's nothing routing to `Checkout#destroy` (which probably needs update so it can cancel any processing orders).

app/Http/Controllers/Store/CheckoutController.php
```diff
@@ -63,10 +63,10 @@ public function destroy()
         return ujs_redirect(route('store.products.index'));
     }
 
-    public function show()
+    public function show($id)
```
nanaya: a bit weird going to `Checkout#show` just to do `Checkout#store`. Sounds like the order is wrong. Maybe `Order#checkout` and `Order#process` instead? Or `update` instead of `store`. Though I think `/order/<number>/checkout` makes a bit more sense than `/checkout/<number>`; be it part of `Order` controller or pseudo-nested route.

resources/views/store/checkout.blade.php
notbakaneko: The current method of the user cancelling an inprogress checkout is no longer supported since being unable to handle multiple checkouts separately was one of the reasons it was added. Some kind of stale cart cleaner is being planned to replace it, but the details of that are tbd.

A side-effect is`failCheckout()` is only triggered by Paypal checkouts at the moment until I figure out where to put it for Xsolla checkouts.

app/Http/Controllers/Store/CheckoutController.php
notbakaneko: not changing any of the routes in this

notbakaneko: route cleaning is a separate thing

resources/views/store/checkout.blade.php
nanaya: > Some kind of stale cart cleaner is being planned to replace it, but the details of that are tbd.

More like stale order cleaner because it's not cart anymore.

PR #3585, opened 2018-08-17
Add a disclaimer when viewing translations of legal info.

cl8n:
uses the "legal" property added by https://github.com/ppy/osu-wiki/pull/1593

this PR renames some of the locale strings, not sure how to handle that with crowdin so I only changed the English ones

---


resources/lang/en/wiki.php
```diff
@@ -33,8 +33,10 @@
             'refresh' => 'Refresh',
         ],
 
-        'outdated' => [
-            '_' => 'This page contains an outdated translation of the original content. Please check the :default for the most accurate information (and consider updating the translation if you are able to help out)!',
+        'translation' => [
+            'legal' => 'This is not legally binding bla bla please read the :default.',
```
nekodex: maybe something like
```This translation is provided for convenience only. The original English text shall be the sole legally binding version.```

PR #3588, opened 2018-08-19
Show a warning when replying to a very old topic

### LINKED ISSUE #3311: Display a notice on bumping dead threads
Forum needs some sort of notice around the posting field to indicate a thread has been dead for lets say 6 months+. People really can't seem to grasp the idea of checking the timestamps of the last post, bumping threads that have been dead for years.

> This thread has been inactive for a while. Only post here if its really necessary, otherwise please consider starting a new thread instead.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/59284658-display-a-notice-on-bumping-dead-threads?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cl8n:
closes #3311 

looks a little strange without changing the warning design, since the avatar box takes up so much space. two versions of the text depending on user's permission to create a new topic in that forum

chrome_2018-08-19_01-06-32

chrome_2018-08-19_01-08-52

---


resources/views/forum/topics/show.blade.php
```diff
@@ -97,6 +97,23 @@
         ]) !!}
             @if (priv_check('ForumTopicReply', $topic)->can())
                 <div class="osu-page osu-page--small-desktop">
+                    @if ($topic->topic_last_post_time->copy()->addMonths(config('osu.forum.necropost_months'))->isPast())
```
notbakaneko: move this to a getter on the model

resources/views/forum/topics/show.blade.php
```diff
@@ -97,6 +97,23 @@
...
+                        <div class="warning-box">
+                            <div class="warning-box__icon">
+                                <i class="fas fa-exclamation-triangle"></i>
+                            </div>
+
+                            @if (priv_check('ForumTopicStore', $topic->forum)->can())
+                                <span>
+                                    {!! trans('forum.topic.create.necropost.new_topic._', [
+                                        'create' => '<a href="'.e(route('forum.topics.create', ['forum_id' => $topic->forum])).'">'.e(trans('forum.topic.create.necropost.new_topic.create')).'</a>',
```
notbakaneko: use `link_to($url, $title)` instead of string building

resources/lang/en/forum.php
```diff
@@ -86,6 +86,15 @@
             'preview_hide' => 'Write',
             'submit' => 'Post',
 
+            'necropost' => [
+                '_' => 'This topic has been inactive for a while. Only post here if you have a specific reason to do so.',
```
notbakaneko: This key isn't being used together with a related key in the same translation, so name it  something besides `_`

app/Models/Forum/Topic.php
```diff
@@ -567,6 +567,15 @@ public function isLocked()
         return $this->topic_status !== static::STATUS_UNLOCKED;
     }
 
+    public function isActive()
+    {
+        return $this
+            ->topic_last_post_time
+            ->copy()
```
notbakaneko: I should point out that this will die if `topic_last_post_time` returns `null` if for whatever reason the db value is 0

notbakaneko: probably not worth handling it at the moment, since the parts that can actually make it `null` have other issues which actually prevent it from happening

PR #3589, opened 2018-08-20
Add a checker for user best scores

notbakaneko:
Helper class to check for user best scores that shouldn't be in the elasticsearch index.

---


app/Libraries/UserBestScoresCheck.php
```diff
@@ -0,0 +1,89 @@
...
+
+namespace App\Libraries;
+
+use App\Libraries\Elasticsearch\Es;
+use App\Models\Score\Best;
+use App\Models\User;
+
+/**
+ * Used to check the best scores for a user if it exists in elasticsearch but not longer
+ * exists in the database. A mistach tends to break the current pager, so this can be run to
```
nekodex: I think you made a mistach here?

nekodex:
Might be worth adding an artisan command/wrapper to this instead of having to run via tinker each time?

PR #3590, opened 2018-08-21
Combine shared database config sections

notbakaneko:
Most settings are the same and the list is going to get longer.

---


notbakaneko:
which CI didn't like 🔥

PR #3595, opened 2018-08-23
Link wiki page titles to their pages

### LINKED ISSUE #3112: Make the leading page of the wiki title clickable
So you won't have to add `Main page: {page}`, `Click to return to {page name} page.`, or `Part of the {page} series.`(i.e. for the `Best Of series` page) on the top of the pages. Without these redirects, it is impossible to go back to the leading page like on the team page.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/57718164-make-the-leading-page-of-the-wiki-title-clickable?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cl8n:
closes #3112 

---


resources/views/wiki/show.blade.php
```diff
@@ -28,12 +28,16 @@
...
-                        {{ $page->subtitle() }}
+                        <a class="osu-page-header__link" href="{{ wiki_url($page->parentPath(), $page->requestedLocale) }}">
+                            {{ $page->subtitle() }}
+                        </a>
                     </h2>
                 @endif
 
                 <h1 class="osu-page-header__title osu-page-header__title--main">
-                    {{ $page->title() }}
+                    <a class="osu-page-header__link" href="{{ wiki_url($page->path, $page->requestedLocale) }}">
```
notbakaneko: does the page need to link back to itself in the title?

cl8n: idk, I just added it to be consistent with the other title. it provides an easy rightclick->copy for sharing the page too

notbakaneko:
@cl8n any progress on handling the pages without direct parents?

Joehuu:
I don't think we need it, since every article should have a direct parent (index). The wiki is just not completed.

https://osu.ppy.sh/help/wiki/Article_Styling_Criteria#index-pages

cl8n:
I haven't worked on it. will this be needed after the wiki has those index pages (as Joehuu mentioned)?

cl8n:
If there isn't a direct parent, it won't link the subtitle. this shouldn't ever happen once the wiki is complete

did this instead of linking the next parent up because I thought it would be strange if the subtitle didn't match where the link was pointing

app/Models/Wiki/Page.php
```diff
@@ -245,6 +245,19 @@ public function pagePath()
         return $this->path.'/'.$this->locale.'.md';
     }
 
+    public function hasParent()
+    {
+        return $this->parentPath() !== null
+            && (new self($this->parentPath(), $this->requestedLocale))->page() !== null;
```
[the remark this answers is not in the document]
notbakaneko: `static` should always be preferred over `self` unless there's a specific reason to call only that implementation which itself probably points to other problems with the code

resources/views/wiki/show.blade.php
notbakaneko: Title either shouldn't be linked or have the hover effect removed (basically, shouldn't look like a link).

PR #3597, opened 2018-08-24
API for new messaging system

nekodex:
The new messaging system is ending up pretty big, so I've extracted out just the API-side changes so we can start reviewing those first.

There's still some cleaning up to be done, but I'm making this pull request early so people can start reviewing it.

I've also started on documenting the API (which can be used as a reference for reviewing this PR), the WIP version of which is viewable here: https://nekodex.github.io/slate/ (undecided whether we want to use slate going forward or not)

(Also I should probably mention that this will break chat within lazer, so client updates will be required.)

---


app/Models/Chat/Channel.php
```diff
@@ -20,40 +20,142 @@
...
+        'pm' => 'PM',
+        'group' => 'GROUP',
+    ];
+
     public function messages()
     {
-        $this->hasMany(Message::class, 'channel_id');
+        $messages = $this->hasMany(Message::class, 'channel_id');
+
+        if ($this->type === self::TYPES['public']) {
```
notbakaneko: This doesn't work with `::with()`, so 
```
Channel::where('type', 'PUBLIC')->with('messages')->get()
```
doesn't apply the where clause, so I'm beginning to think we shouldn't be baking in extra behaviour into association methods.

app/Models/Chat/Channel.php
```diff
@@ -20,40 +20,142 @@
...
+        }
+
         $message = new Message();
         $message->user_id = $sender->user_id;
-        $message->content = $body;
+        $message->content = $content;
         $message->is_action = $isAction;
         $message->channel()->associate($this);
         $message->save();
+        $message = $message->fresh();
```
notbakaneko: can just `refresh()`; also neither is needed if it's just for updating the `message_id`

app/Models/Chat/Channel.php
```diff
@@ -20,40 +20,142 @@
...
 
-        return $message->fresh();
+        $userChannel = UserChannel::where(['channel_id' => $this->channel_id, 'user_id' => $sender->user_id]);
+        if ($userChannel) {
```
notbakaneko: Builder is never falsy

app/Http/Controllers/Chat/ChannelsController.php
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        if (!$channel->hasUser(Auth::user())) {
+            $channel->addUser(Auth::user());
+        }
+
+        return response([], 204);
+    }
+
+    public function part($channel_id, $user_id)
```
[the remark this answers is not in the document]
nekodex: Currently yeah, the intention was to eventually allow admins/etc to add/remove users from channels via these

resources/assets/coffee/react/beatmaps/search-panel.coffee
```diff
@@ -56,7 +56,7 @@ class Beatmaps.SearchPanel extends React.PureComponent
       div
         className: 'osu-page-header__background'
         style:
-          backgroundImage: "url(#{@props.background})"
+          backgroundImage: @props.background ? "url(#{@props.background})" : ""
```
notbakaneko: I think too many things might have been added to this commit >_>

app/Models/Chat/Channel.php
```diff
@@ -109,18 +109,18 @@ public function receiveMessage(User $sender, string $content, bool $isAction = f
...
         $message->user_id = $sender->user_id;
         $message->content = $content;
         $message->is_action = $isAction;
+        $message->timestamp = Carbon::now();
         $message->channel()->associate($this);
         $message->save();
-        $message = $message->fresh();
 
-        $userChannel = UserChannel::where(['channel_id' => $this->channel_id, 'user_id' => $sender->user_id]);
+        $userChannel = UserChannel::where(['channel_id' => $this->channel_id, 'user_id' => $sender->user_id])->first();
```
notbakaneko: could just call `update` directly after `where` if it's not going to throw an exception

app/Http/Controllers/Chat/ChatController.php
```diff
@@ -0,0 +1,185 @@
...
+                    $presence['users'] = $channelMembers;
+                }
+
+                if ($channel->type === Channel::TYPES['pm']) {
+                    // if this is a pm
+                    $users = array_filter($channelMembers, function ($k) {
+                        // remove current user, leaving only the other party
+                        return $k !== Auth::user()->user_id;
+                    });
+                    $targetUser = User::find(array_shift($users));
```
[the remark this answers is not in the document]
notbakaneko: Occurs to me that we should probably be batch selecting the all the users first and then assigning them, instead of one user per loop

PR #3602, opened 2018-08-27
Inverse colours for the profile extra information

### LINKED ISSUE #3562: Design: Clarify personal information more on profile
chrome_2018-08-05_14-17-23
If you ask me, it's not that you directly understand the post count is clickable (except you hover over it) or getting information when hovering over joined/last seen. Also the fact links usually are in blue color makes it a bit confusing

TPGPL:
Closes #3562 

---

### DIFF
--- a/resources/assets/coffee/react/profile-page/header-extra.coffee
+++ b/resources/assets/coffee/react/profile-page/header-extra.coffee
@@ -142,15 +142,18 @@ class ProfilePage.HeaderExtra extends React.Component
 
         div className: "#{bn}__column #{bn}__column--text #{bn}__column--shrink",
           div className: "#{bn}__rows",
-            @fancyLink
+            div
+              className: "#{bn}__row"
               key: 'location'
               icon: 'fas fa-map-marker-alt'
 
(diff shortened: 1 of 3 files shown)


resources/assets/coffee/react/profile-page/header-extra.coffee
```diff
@@ -142,15 +142,18 @@ class ProfilePage.HeaderExtra extends React.Component
 
         div className: "#{bn}__column #{bn}__column--text #{bn}__column--shrink",
           div className: "#{bn}__rows",
-            @fancyLink
```
notbakaneko: the text was being set by `fancyLink`
image

I'd update `fancyLink` to set a different css class if it's not a link

TPGPL: would using `:not` selector be fine for that?

notbakaneko: except the twitter link is in a `span` so it doesn't apply the correct colour.

Also missing the colour for the forum posts link

Issue #3607, opened 2018-08-28
Option to hide deleted comments/posts for moderators || Give deleted issues transparency in the timeline


Mirruuu:
It's very annoying having to scroll down so much for threads that have mostly deleted posts such as https://osu.ppy.sh/beatmapsets/836819/discussion/-/generalAll#/506715 so having the option to remove them would be nice and convenient.

Additionally there is no distinction between removed or active issues on the modding thread:
![image](https://user-images.githubusercontent.com/25852637/44748658-960c1200-ab10-11e8-8f8d-31ccb7c4ee83.png)
The suggestion is actually deleted but it looks like an open issue:
![image](https://user-images.githubusercontent.com/25852637/44748703-aae8a580-ab10-11e8-85f8-dc4b94bdad9b.png)
Could fix this as an individual problem by giving deleted issues more transparency in the timeline than open ones, but if an option to hide all deleted posts is added that would also partially solve that problem.

PR #3608, opened 2018-08-29
Make deleted discussion visibly less opaque.

notbakaneko:
Because 0.75 opacity is like nothing

partially addresses #3607

---


Mirruuu:
It looks the same still, was it correctly merged? 
image
image

notbakaneko:
oh, the _other_ timeline

Issue #3370, opened 2018-06-19
You should be able to bring up header+search filters when scrolled down in the beatmaps listing


dorothy3242:
Being lost in an endless scrolling abyss with no easy way to escape is not fun and extremely inconvenient. If someone many maps deep into scrolling down on the beatmap listing page, it is not user-friendly to have to scroll the entire way back up just to change the search term or filters, or go to a different page on the website.

There should either be a way to bring the search bar and filters up after scrolling down, or a jump to top button. Or both. Preferably both.

arflyte:
Adapting beatmap search to the new mini navigation. Back to top button at bottom right.

![e46226ea95](https://user-images.githubusercontent.com/3632750/41641154-497d40fa-749f-11e8-9037-fdc99257c8c5.png)

TPGPL:
should be closed

peppy:
Resolved via #3809.

PR #3617, opened 2018-08-31
Add back to top button on beatmap listing

notbakaneko:
partly addresses #3370

Sticky search bar is going to be done separately since I expect implementing that to be much messier than it looks.

- should only scroll if the current scroll position is below the top of search panel.
- clicking the button should preserve the previous scroll position and clicking on it again should restore the scroll position.
- scrolling after clicking "back to top" will clear the saved scroll position.
- new search or changing the sort order should cause a reset.

---

Allow users to scroll the page back up to the search input on the beatmap listing page by clicking a button.


resources/assets/lib/back-to-top.coffee
```diff
@@ -0,0 +1,76 @@
...
+      @setState lastScrollY: null
+    else
+      scrollY = if @props.anchor? then $(@props.anchor.current).offset().top else 0
+      if window.scrollY > scrollY
+        @setState lastScrollY: window.scrollY
+
+        @destroyTooltip()
+        window.scrollTo(window.scrollX, scrollY)
+
+        Timeout.set 0, () =>
```
nanaya: sometimes triggered when scrolling to top/anchor on firefox :confused:

resources/assets/lib/back-to-top.coffee
```diff
@@ -0,0 +1,76 @@
...
+
+  onClick: (_e) =>
+    if @state.lastScrollY?
+      @destroyTooltip()
+      window.scrollTo(window.scrollX, @state.lastScrollY)
+
+      @setState lastScrollY: null
+    else
+      scrollY = if @props.anchor? then $(@props.anchor.current).offset().top else 0
+      if window.scrollY > scrollY
```
nanaya: `window.pageYOffset` for better compatibility.

nanaya:
Just remembered about this but do you think the button should be hidden when around the top and no saved position?

resources/assets/lib/back-to-top.coffee
```diff
@@ -0,0 +1,68 @@
...
+    document.removeEventListener 'scroll', @onScroll
+
+
+  render: =>
+    button
+      className: 'back-to-top'
+      'data-tooltip-float': 'fixed'
+      onClick: @onClick
+      title: if @state.lastScrollY? then osu.trans('common.buttons.back_to_previous') else osu.trans('common.buttons.back_to_top')
+      ref: @container
```
nanaya: not used anymore

notbakaneko:
Now with stupid workaround for firefox

notbakaneko:
I thought about hiding the button if the viewport is above the anchor point but couldn't be bothered adding more things to watch for its position, although now that it's using an `IntersectionObserver`, it'd be simpler to handle

PR #3624, opened 2018-09-03
First new topic button in forum is too wide

notbakaneko:
---

### DIFF
--- a/resources/assets/less/bem/forum-topics-spacer.less
+++ b/resources/assets/less/bem/forum-topics-spacer.less
@@ -19,7 +19,7 @@
 .forum-topics-spacer {
   display: flex;
   align-items: center;
-  margin: 0px 40px;
+  margin: 30px 40px;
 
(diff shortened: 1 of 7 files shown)


resources/assets/less/bem/forum-topics-spacer.less
```diff
@@ -16,23 +16,45 @@
...
     }
+
+    // needs to:
+    // - expand to fit text.
+    // - not affect pager layout.
+    // - be the same width in blocks with no pager.
+    // - not expand to fill up space.
+    // - have minimum width.
+    // won't work if text is way too long.
+    .forum-topics-spacer--pager & {
```
nanaya: or maybe just `[[----flex: 1----][flex:none][----flex: 1----]]`? Paired with min-width set in `btn-osu-big--forum-new-topic` instead.

nanaya:
rip travis

PR #3637, opened 2018-09-07
Add ability to report scores

### LINKED ISSUE #2583: Can't report plays on beatmap top rankings

cl8n:
closes #2583 

the colors don't follow flyte's design because his version has a dark theme

---

### DIFF
--- a/app/Http/Controllers/BeatmapsController.php
+++ b/app/Http/Controllers/BeatmapsController.php
@@ -67,16 +67,13 @@ class BeatmapsController extends Controller
         $query->withMods($mods);
         $query->withType($type, compact('user'));
 
-        $includes = ['user', 'user.country'];
-        if ($user !== null) {
-            $includes[] = "userReportPresence:user_id($user->user_id)";
-        }
-
         $results = [
-            'scores' => json_collection($query->forListing(), 'Score', $includes),
+            'scores' => json_collection($query->forListing(), 'Score', ['user', 'user.country']),
         ];
 
         if ($user !== null) {
+            $results['reportedScores'] = (clone $query)->withReportBy($user)->pluck('score_id');
+
             $score = (clone $query)->where('user_id', $user->user_id)->first();
 
             if ($score !== null) {
--- a/app/Http/Controllers/BeatmapsController.php
+++ b/app/Http/Controllers/BeatmapsController.php
@@ -67,8 +67,13 @@ class BeatmapsController extends Controller
         $query->withMods($mods);
         $query->withType($type, compact('user'));
 
+        $includes = ['user', 'user.country'];
+        if ($user !== null) {
+            $includes[] = "userReportPresence:user_id($user->user_id)";
+        }
+
         $results = [
-            'scores' => json_collection($query->forListing(), 'Score', ['user', 'user.country']),
+            'scores' => json_collection($query->forListing(), 'Score', $includes),
         ];
 
         if ($user !== null) {
--- a/app/Models/Score/Best/Model.php
+++ b/app/Models/Score/Best/Model.php
@@ -267,6 +267,13 @@ abstract class Model extends BaseModel
         }
     }
 
+    public function scopeWithReportBy($query, $user)
+    {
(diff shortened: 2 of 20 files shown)


resources/assets/coffee/react/beatmapset-page/scoreboard-table.coffee
```diff
@@ -16,11 +16,37 @@
...
 
-{a, div, table, tr, td, th, thead, tbody} = ReactDOMFactories
+{a, button, i, span, table, tr, td, th, thead, tbody} = ReactDOMFactories
 el = React.createElement
 bn = 'beatmap-scoreboard-table'
 
 BeatmapsetPage.ScoreboardTable = (props) ->
+  props.scores = props.scores.map (score) ->
+    score.reportButton =
+      if score.userReportPresence.reported
```
notbakaneko: userReportPresence will be undefined if not logged in.

Fetching the presence information is doing an an extra query per score adding up to a lot of extra queries.

resources/assets/lib/report-form.coffee
```diff
@@ -31,9 +31,7 @@ options = [
 ]
 
 export class ReportForm extends PureComponent
-  constructor: (props) ->
```
notbakaneko: pass customization options in through `props` and compose specialized variations of `ReportForm` instead of making it differentiate between score and user

https://reactjs.org/docs/composition-vs-inheritance.html#specialization

resources/assets/coffee/react/beatmapset-page/main.coffee
```diff
@@ -143,6 +143,18 @@ class BeatmapsetPage.Main extends React.Component
...
+  setScoreReportPresence: (_e, scoreId) =>
+    scores = @state.scores
+    scoreIndex = scores.findIndex((s) => s.id == scoreId)
+
+    newScore = scores[scoreIndex]
+    newScore.userReportPresence.reported = true
+
+    scores.splice(scoreIndex, 1, newScore)
+
+    @setState scores: scores
```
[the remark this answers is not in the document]
cl8n: that makes sense, thanks for the explanation

I have this working now by making a shallow copy in setState's updater, is this good practice for changing properties of state objects?

resources/assets/coffee/react/beatmapset-page/scoreboard-table.coffee
```diff
@@ -16,11 +16,38 @@
 #    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
 ###
 
-{a, div, table, tr, td, th, thead, tbody} = ReactDOMFactories
+{a, button, div, i, span, table, tr, td, th, thead, tbody} = ReactDOMFactories
 el = React.createElement
 bn = 'beatmap-scoreboard-table'
 
 BeatmapsetPage.ScoreboardTable = (props) ->
+  props.scores = props.scores.map (score) ->
```
notbakaneko: Don't modify `props` in a component, they're supposed to be immutable within a component.

Also consider making a separate component for the report button that takes the `score` and `mode` to render

app/Http/Controllers/BeatmapsController.php
```diff
@@ -67,8 +67,13 @@ public function scores($id)
         $query->withMods($mods);
         $query->withType($type, compact('user'));
 
+        $includes = ['user', 'user.country'];
+        if ($user !== null) {
+            $includes[] = "userReportPresence:user_id($user->user_id)";
```
notbakaneko: Probably better to return a list of reported score ids for the current beatmap as a top level property in `$results` instead of trying to inline the state.

app/Models/Score/Best/Model.php
```diff
@@ -267,6 +267,13 @@ public function scopeWithType($query, $type, $options)
         }
     }
 
+    public function scopeWithReportBy($query, $user)
```
notbakaneko: `scopeWithReportedBy`

cl8n:
superseded by #3855

PR #3643, opened 2018-09-09
Error on topic replies that only contain quotes

### LINKED ISSUE #2858: Refuse ability to post if content is only a long quote
This has been an issue when it isn't supposed to be, for some reason occurent in the skinning subforum. Random posts only quoting OP and nothing else. Got this in mind, maybe something else to tackle this issue?

cl8n:
closes #2858 

---


app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -160,6 +160,13 @@ public function reply(HttpRequest $request, $id)
             'body' => 'required',
         ]);
 
+        $trimmedBody = trim(Request::input('body'));
+
+        if (preg_match('/^\[quote(?:=".+?")?\].+?\[\/quote\]/s', $trimmedBody, $matches)
```
nanaya: should be done in model validation (the body validation above is outdated and shouldn't be there either).

cl8n:
now checks for posts containing only any number of quotes

> [5:50 PM] Death: Sometimes there are like 5 - 7 OP quotes in one post lol

because apparently that happens too

app/Models/Forum/Post.php
```diff
@@ -233,6 +233,14 @@ public function isValid()
             }
         }
 
+        $trimmedBody = trim($this->body_raw);
+        while (preg_match('/^\[quote(?:=".+?")?\].+?\[\/quote\]/s', $trimmedBody, $matches)) {
```
notbakaneko: This doesn't quote match nested quotes right:

```[quote="i"][quote="am"][quote="a"]teapot[/quote][/quote][/quote]```
matches
```[quote="i"][quote="am"][quote="a"]teapot[/quote]```

Comment (nanaya) [reply app/Models/Forum/Post.php]: or maybe just use `BBCodeFromDB::removeBlockQuotes`?

Comment (cl8n) [reply app/Models/Forum/Post.php]: yeah I should be just using that. but now upon checking, it doesn't seem to work either in all cases

File: app/Libraries/BBCodeFromDB.php [hunk 3]
```diff
@@ -382,21 +382,26 @@ public static function removeBlockQuotes($text)
     {
         $level = 0;
         $marker = 0;
+        $quoteBeginning = 0;
 
-        while ($marker >= 0 && $marker < mb_strlen($text) && $level >= 0) {
+        while ($marker >= 0 && $level >= 0) {
```
notbakaneko: This loop should be implemented as forward-only to avoid infinite loops, which are really undesirable for this method as it's being used by an unattended job.

app/Libraries/BBCodeFromDB.php
```diff
@@ -382,21 +382,26 @@ public static function removeBlockQuotes($text)
...
+                if ($level === 0) {
+                    $quoteBeginning = $match['start'][1];
+                }
                 $level++;
             } elseif (present($match['end'][0])) {
                 $level--;
                 $marker = $match['end'][1] + mb_strlen($match['end'][0]);
                 if ($level === 0) {
-                    $text = mb_substr($text, $marker, mb_strlen($text) - $marker);
+                    $text = mb_substr($text, 0, $quoteBeginning).mb_substr($text, $marker);
```
notbakaneko: turns out `preg_match` returns non-multibyte offsets, so we have to use the non `mb_` versions of the string functions or use some combination of the `mb_ereg` functions to get the offsets.

You can test using any string with multibyte characters like
```
ねねぽお誕誕ふお
[quote:1337]abcdef8あうっっとどあえうフォそそcっっ。「」22？？？？[/quote:1337]
茶々ちゃ＼＼・・・＼／as2feff
```

PR #3645, opened 2018-09-10
Add missing variable for topic lock view

nanaya:
Another bug from recent forum moderation update.

### DIFF
--- a/app/Http/Controllers/Forum/TopicsController.php
+++ b/app/Http/Controllers/Forum/TopicsController.php
@@ -102,14 +102,17 @@ class TopicsController extends Controller
     {
         $topic = Topic::withTrashed()->findOrFail($id);
 
-        priv_check('ForumModerate', $topic->forum)->ensureCan();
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -102,14 +102,17 @@ public function lock($id)
     {
         $topic = Topic::withTrashed()->findOrFail($id);
 
-        priv_check('ForumModerate', $topic->forum)->ensureCan();
+        $moderationPriv = priv_check('ForumModerate', $topic->forum);
+
+        $moderationPriv->ensureCan();
+        $userCanModerate = $moderationPriv->can();
```
notbakaneko: wouldn't this be always `true` by this point?

nanaya: yes, it was either add one line assigning true or this... thing. Maybe should just return true for that `ensureCan` thing.

PR #3650, opened 2018-09-11
Better display name for unmatched users

nanaya:
Looks like using username isn't a good idea.


notbakaneko:
does this fix some `osu-totallyarealusername` thing showing in the comments?

nanaya:
That one shouldn't happen in live site.

PR #3662, opened 2018-09-12
Paginate comments

nanaya:
Seems working...?

### DIFF
--- a/resources/assets/coffee/react/_components/comments.coffee
+++ b/resources/assets/coffee/react/_components/comments.coffee
@@ -44,9 +44,18 @@ class @Comments extends React.PureComponent
 
 
   render: =>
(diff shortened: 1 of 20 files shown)


resources/assets/coffee/react/_components/comments.coffee
```diff
@@ -85,19 +94,16 @@ class @Comments extends React.PureComponent
 
   appendBundle: (_event, {comments}) =>
     @setState
+      # remove old objects included in new bundle by relying on uniqBy keeping first item
       comments:
-        _(@state.comments)
-          .concat(comments.comments)
-          .reverse()
+        _(comments.comments)
+          .concat(@state.comments)
```
notbakaneko: I've not actually seen a case where it's been an issue yet, but if we're `concat`ing to existing state, we should probably be using the functional `(prevState, prevProps) ->` version of `setState`

PR #3664, opened 2018-09-13
Add option to download user's top plays

### LINKED ISSUE #2797: You can't download replays from the userpage's top plays
This star icon allows you to download an .osr file of that play
screen shot 2018-03-24 at 1 19 46 pm

It's only for plays that are in the top 1000 of the leaderboard, but if its stored somewhere then might as well link it for all

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/59105072-you-can-t-download-replays-from-the-userpage-s-top-plays?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F21853994&utm_medium=issues&utm_source=github).
</bountysource-plugin>

notbakaneko:
option on the (very bare) side menu on hover if available

image

closes #2797

---


resources/assets/lib/play-detail-menu.coffee
```diff
@@ -0,0 +1,80 @@
...
+  componentWillUnmount: =>
+    $(document).off ".#{@uuid}", @hide
+
+
+  hide: (e) =>
+    event = e.originalEvent
+    if event.keyCode == 27 || (event.button == 0 && !(@menu.current in event.composedPath()))
+      @setState active: false
+
+    @props.onHide?()
```
nanaya: this probably should be part of the conditional above otherwise right clicking menu button results in weird state.

notbakaneko:
Will need to do some fancy pants design merging after https://github.com/ppy/osu-web/pull/3637 is done

app/Http/Controllers/Users/ReplaysController.php
```diff
@@ -0,0 +1,68 @@
...
+namespace App\Http\Controllers\Users;
+
+use App\Http\Controllers\Controller;
+use App\Models\Score\Best;
+use App\Models\User;
+use Illuminate\Contracts\Filesystem\FileNotFoundException;
+
+class ReplaysController extends Controller
+{
+    public function show($userId, $beatmapId, $mode)
```
nanaya: shouldn't this be limited to logged in users?

resources/assets/lib/play-detail-menu.coffee
```diff
@@ -0,0 +1,82 @@
...
+  componentDidMount: =>
+    $(document).on "click.#{@uuid}", @hide
+    $(document).on "keyup.#{@uuid}", @hide
+
+
+  componentWillUnmount: =>
+    $(document).off ".#{@uuid}", @hide
+
+
+  hide: (e) =>
```
nanaya: should add `@state.active` check otherwise all other instances will run `@props.onHide` when clicking inside a menu.

resources/assets/lib/play-detail-menu.coffee
```diff
@@ -0,0 +1,82 @@
...
+    event = e.originalEvent
+    if event.keyCode == 27 || (event.button == 0 && !(@menu.current in event.composedPath()))
```
nanaya: maybe add check for button as well otherwise the button doesn't toggle. Or move `@menu` ref instead.

resources/assets/lib/play-detail-menu.coffee
notbakaneko: previously every instance had its own listener without jquery, so this wasn't a problem. Then I decided to have the event handling multiplexed into a single listener through jquery, but seeing as it triggers every handler, I'm not sure it's such a good idea anymore :s

notbakaneko: NEVERMIND :best:

PR #3672, opened 2018-09-14
Validate presence of topic title

nanaya:
Add presence validation at model level and remove the one in controller (which is missing in update action).


tests/Controllers/ForumTopicsControllerTest.php
```diff
@@ -133,6 +129,43 @@ public function testStore()
         $this->assertSame($initialTopicCount + 1, Forum\Topic::count());
     }
 
+    public function testUpdate()
```
notbakaneko: separate test cases for expected failure and expected success cases

PR #3674, opened 2018-09-15
Change the Source Code link to redirect to the osu-web repository

TPGPL:
I think it makes sense to me that it *should* link to the osu-web repository instead of the organization.

---


LiquidPL:
I'm pretty sure the link is related to the whole osu! codebase, not just web. Also there are other repositories in there that form parts of the new web/bancho replacement backend.

notbakaneko:
Like said above, it's source for everything, not just the site.

PR #3676, opened 2018-09-15
Increase top margin for comment replies

cl8n:
20px makes it consistent with the spacing between the rest of the comments. not sure why this was set differently, it looks pretty strange

---


notbakaneko:
Replies are supposed to be closer to each other than regular comments so that there is visual distinction between comments and reply blocks

PR #3683, opened 2018-09-16
Allow GMT/QAT to edit discussions

### LINKED ISSUE #3642: Ability for GMT/QAT to edit discussions and replies
There is currently no way to edit posts by other users in modding discussions for moderators, meaning posts with unnecessary remarks or insults need to either be left alone as is or be outright deleted, removing any kind of useful information or thoughts the post might otherwise have provided to the discussion, see:

image

Naxesss:
should resolve #3642


notbakaneko:
We want there to be an actual audit trail and history before supporting editing of discussion posts; until then, abusive posts should be deleted

PR #3689, opened 2018-09-18
Update hashers for Laravel 5.6+

notbakaneko:
Laravel 5.6 onwards has an additional method on the Hasher interface
...which is just `password_get_info($hashedValue)` on the default implementations

...which is just there to complete the PHP password hashing api?

Basically we are going need this to update laravel beyond 5.5

---


notbakaneko:
We'll come back to this on the next LTS

PR #3698, opened 2018-09-20
Update localisations and crowdin configuration

peppy:
Probably need to to a vague check this time, it's been piling up (had some issues with the crowdin dumping configuration)


resources/lang/fi/store.php
```diff
@@ -105,7 +105,7 @@
     'supporter_tag' => [
         'gift' => 'lahjoita pelaajalle',
         'require_login' => [
-            '_' => 'Ollaksesi Tukija sinun täytyy olla :link!',
+            '_' => '',
```
notbakaneko: Some of the translations got removed like this one, but I can't seem to find any history of the change

TPGPL: Yeah this one was available to translate on crowdin for no reason

nanaya: original translation has been changed so looks like crowdin assumes it's a different thing.

TPGPL: Oh right i think it was after every `supporter tag` was changed into `osu!supporter tag`

PR #3717, opened 2018-09-27
Add padding to Settings section title on mobile

TPGPL:
before
image 

after
image

---


nanaya:
It's already fixed in latest version (which hasn't been deployed).

PR #3720, opened 2018-09-28
Change user session management to use sets instead of scan

nekodex:
Yeah so Redis' SCAN command isn't so scalable.

This replaces the need to perform the SCAN by maintaining an additional SET (per user) alongside the normal session keys that contains a list of all their sessions.

---

### DIFF
--- a/app/Libraries/Session/Store.php
+++ b/app/Libraries/Session/Store.php
@@ -35,7 +35,7 @@ class Store extends \Illuminate\Session\Store
      *
      * @return string
      */
-    protected function keyPrefix(int $userId = null)
+    protected static function keyPrefix(int $userId = null)
     {
         $userId = $userId ?? 'guest';
 
@@ -47,9 +47,9 @@ class Store extends \Illuminate\Session\Store
      *
      * @return string
      */
-    protected function sessionListKey(int $userId = null)
+    protected static function sessionListKey(int $userId = null)
     {
-        return config('cache.prefix').':'.$this->keyPrefix($userId);
+        return config('cache.prefix').':'.static::keyPrefix($userId);
     }
 
     /**
@@ -73,10 +73,10 @@ class Store extends \Illuminate\Session\Store
     {
         if (Auth::check()) {
             $userId = Auth::user()->user_id;
-            $fullSessionId = $this->keyPrefix($userId).":{$sessionId}";
+            $fullSessionId = static::keyPrefix($userId).':'.$sessionId;
             $this->handler->destroy($fullSessionId);
 
-            Redis::srem($this->sessionListKey($userId), config('cache.prefix').':'.$fullSessionId);
+            Redis::srem(static::sessionListKey($userId), config('cache.prefix').':'.$fullSessionId);
 
             return true;
         }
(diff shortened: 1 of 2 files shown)


app/Libraries/Session/Store.php
```diff
@@ -145,12 +213,31 @@ public function stripPrefix($sessionId)
     public function migrate($destroy = false, int $userId = null)
     {
         // Overriden to allow passing through $userId to namespace session ids
+
         if ($destroy) {
+            if (!$this->isGuestSession()) {
+                Redis::srem($this->sessionListKey($userId), $this->getId());
```
nanaya: missing cache prefix in `$this->getId()`?

app/Libraries/Session/Store.php
```diff
@@ -145,12 +213,31 @@ public function stripPrefix($sessionId)
...
         $this->setExists(false);
         $this->setId($this->generateSessionId($userId));
 
         return true;
     }
+
+    /**
+     * Save the session data to storage.
+     *
+     * @return bool
```
nanaya: no actual return in the function?

app/Libraries/Session/Store.php
```diff
@@ -30,24 +30,81 @@ class Store extends \Illuminate\Session\Store
...
+     * Returns whether the current session is a guest session based on the key prefix of the session id.
+     *
+     * @return bool
+     */
+    public function isGuestSession()
+    {
+        $sessionId = $this->getId();
+        $prefix = substr($sessionId, 0, strlen($sessionId) - static::SESSION_ID_LENGTH - 1);
+
+        return $prefix === $this->keyPrefix(null);
```
nanaya: `starts_with($this->getId(), $this->keyPrefix())`?

app/Libraries/Session/Store.php
```diff
@@ -30,24 +30,81 @@ class Store extends \Illuminate\Session\Store
...
+     */
     public function destroyUserSession($sessionId)
     {
         if (Auth::check()) {
             $userId = Auth::user()->user_id;
-            $this->handler->destroy($this->keyPrefix($userId).$sessionId);
+            $fullSessionId = $this->keyPrefix($userId).":{$sessionId}";
+            $this->handler->destroy($fullSessionId);
+
+            Redis::srem($this->sessionListKey($userId), config('cache.prefix').$fullSessionId);
```
nanaya: missing `:` after cache prefix?

app/Libraries/Session/Store.php
```diff
@@ -30,24 +30,81 @@ class Store extends \Illuminate\Session\Store
...
+     * Destroys a session owned by the current user that is identified by the given id.
+     *
+     * @return bool
+     */
     public function destroyUserSession($sessionId)
     {
         if (Auth::check()) {
             $userId = Auth::user()->user_id;
-            $this->handler->destroy($this->keyPrefix($userId).$sessionId);
+            $fullSessionId = $this->keyPrefix($userId).":{$sessionId}";
```
nanaya: Not sure if mixing string interpolation and concatenation in single line a good thing.

PR #3721, opened 2018-09-28
Disallow replying to deleted comment

nanaya:
Replying to reply of a deleted comment is still allowed though.

Fixes #3713.

PR #3722, opened 2018-09-28
Don't fail validating comment with missing commentable

nanaya:
Laravel isn't even trying

```
Illuminate/Database/QueryException with message 'SQLSTATE[42S22]: 
Column not found: 1054 Unknown column 'comments.' in 'where clause' 
(SQL: select exists(select * from `comments` where `comments`.`` is null) as `exists`)'
```


nanaya:
NEVER MIND THIS IS REQUIRED FOR #3721.

PR #3727, opened 2018-09-28
Add Forum indicator in Subscriptions page

TPGPL:
not sure about the classes for links

---

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -558,6 +558,25 @@ function link_to_user($user_id, $user_name = null, $user_color = null)
     }
 }
 
+function link_to_forum($forum_id, $forum_name)
+{
+    if ($forum_id instanceof App\Models\Forum\Topic) {
+        $forum_id = $forum_id->getKey();
+    }
+
+    if ($forum_id instanceof App\Models\Forum\Forum) {
+        $forum_name = $forum_name->getKey();
+    }
+
+    if ($forum_id) {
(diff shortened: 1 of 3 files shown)


resources/views/forum/forums/_topic.blade.php
```diff
@@ -67,6 +67,14 @@ class="
                 {{ $topic->titleNormalized() }}
             </a>
 
+            <div>
+                {!! trans("forum.topic.location", [
+                    "forum" => link_to_forum(
+                        $topic->forum_id,
+                        $topic->forum,
```
notbakaneko: trailing comma in parameter list

app/helpers.php
```diff
@@ -563,6 +563,25 @@ function link_to_user($user_id, $user_name = null, $user_color = null)
...
+    }
+
+    if ($forum_id instanceof App\Models\Forum\Forum) {
+        $forum_name = $forum_name->getKey();
+    }
+
+    if ($forum_id) {
+        $forum_url = e(route('forum.forums.show', $forum_id));
+
+        return "<a class='forums__forum-topic-link' href='{$forum_url}' style='{$style}'>{$forum_name}</a>";
```
notbakaneko: `$style` is not defined

notbakaneko: It's using the `forum_id` or the entire `forum` object as the link text? That doesn't seem right in any way.

PR #3728, opened 2018-09-28
Add disclaimer if there are no store orders

TPGPL:


resources/lang/en/store.php
```diff
@@ -89,6 +89,7 @@
         'paid_on' => 'Order placed :date',
 
         'invoice' => 'View Invoice',
+        'no_orders' => 'No orders to view.',
```
nanaya: I don't think `to view` part is needed.

TPGPL: i believe in most pages i visited, it was `No <item> to view.`

nanaya:
You should first check if your changes actually work.

TPGPL:
I would if docker actually worked

nanaya:
Report problems on docker instead of blindly making changes and pull requests.

PR #3738, opened 2018-10-01
Search classes refactoring

notbakaneko:
- parse parameters and set defaults earlier.
- changes `HasSearch` from trait to class.
- limit use of `page` to request params.
- include `_id` as part of beatmap search sort order, so that there's a deterministic order if the sort values would otherwise have ben the same.
- excluding `_source` from results should be using `false` instead of fields that don't exist in `_source` (`_id` is always included in the elasticsearch response).

This is a dependency of some incoming changes but this is PR'ed first since it's not dependent on the other changes, so we can check for breakage on this part separately.

---

### DIFF
--- a/app/Libraries/Elasticsearch/HasSearch.php
+++ b/app/Libraries/Elasticsearch/HasSearch.php
@@ -25,6 +25,7 @@ abstract class HasSearch
     protected $highlight;
     protected $params;
     protected $query;
+    protected $source;
     protected $type;
 
     public function __construct(SearchParams $params)
@@ -84,7 +85,7 @@ abstract class HasSearch
      */
     public function source($fields)
     {
-        $this->params->source = $fields;
+        $this->source = $fields;
 
         return $this;
     }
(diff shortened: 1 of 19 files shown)


app/Libraries/Search/PostSearchRequestParams.php
```diff
@@ -27,8 +27,10 @@ class PostSearchRequestParams extends PostSearchParams
 {
     public function __construct(Request $request, User $user)
     {
+        $this->__construct();
```
nanaya: more like `parent::__construct()`.

app/Libraries/Elasticsearch/HasSearch.php
```diff
@@ -20,50 +20,35 @@
...
-        $this->size = $size;
+        $this->params->from = $from;
 
         return $this;
     }
 
     /**
      * @return $this
      */
-    public function page(?int $page)
```
nanaya: `/users/:user/posts` stops working because this is being called in the controller.

app/Libraries/Search/PostSearchRequestParams.php
notbakaneko: wtf 🥇

PR #3743, opened 2018-10-02
Add explanations to beatmap discussion buttons

### LINKED ISSUE #2112: Add tooltips for modding edit pin button and upvote/downvote
It is unclear what these actually do just by looking at them, leading their functionality of keeping the edit box always on top of the page / awarding kudosu to be found by either trial and error or spread through word of mouth. These functions should really be able to learned within the interface itself.

Additionally, for the pin button, it can make it unclear whether the colors mean the box is currently pinned or unpinned since there is no text, like for the forum's subscription button, to tell you so.

TPGPL:
Fixes part of #2112


resources/lang/en/beatmaps.php
```diff
@@ -124,6 +126,10 @@
             'wip' => 'Note: This beatmap is marked as a work-in-progress by the creator.',
         ],
 
+        'votes' => [
+            'up' => 'Upvote this discussion',
```
nanaya: I don't think this helps because the problem is it isn't clear that voting up/down affects kudosu.

resources/assets/coffee/react/beatmap-discussions/discussion.coffee
```diff
@@ -127,6 +127,7 @@ class BeatmapDiscussions.Discussion extends React.PureComponent
       i className: "fas fa-#{icon}"
       span className: "#{vbn}__count",
         @props.discussion.votes[type]
+      title: osu.trans('beatmaps.discussions.votes.#{type}')
```
nanaya: This doesn't work. Should be part of `button` options, not its children.

resources/assets/coffee/react/beatmap-discussions/new-discussion.coffee
```diff
@@ -91,6 +91,11 @@ class BeatmapDiscussions.NewDiscussion extends React.PureComponent
               onClick: @toggleSticky
               span className: 'btn-circle__content',
                 i className: 'fas fa-thumbtack'
+              title:
```
nanaya: Same as above.

PR #3746, opened 2018-10-04
Add post count to user forum info box

### LINKED ISSUE #2742: Missing details under user on forums
Missing post count, ~online status, and supporter status.~

### LINKED ISSUE #3573: Quicklink to private messages on new forums
I didn't find this anywhere as an issue so:
old page contained links to quickly pm someone about a forum post they made - at least i've been using this feature from time to time instead of going to someone's profile to click "send private message"

example

cl8n:
closes #2742 
I guess also closes #3573 ..?

doesn't follow exactly the design since the new one has yet to be implemented chrome_2018-10-03_17-16-29

---


resources/views/forum/topics/_post_info.blade.php
```diff
@@ -65,6 +65,11 @@ class="forum__user-flag forum__user-flag--country"
         </div>
 
         <div class="forum-post__info-panel-extra-bottom">
+            <div class="forum-post__info-row">
+                <a class="forum-post__info-link" href="{{ route("users.posts", $user) }}">
+                    {{ trans_choice('forum.post.info.post_count', number_format($user->user_posts)) }}
```
nanaya: `number_format` shouldn't be used here because the rule might be checked incorrectly (e.g. numbers between 1000 and 1999 will use singular form for English).

cl8n: should I just leave it un-formatted for now then? also, this pattern is used in a few other places

nanaya: yes, I'll fix it later.

resources/views/forum/topics/_post_info.blade.php
```diff
@@ -65,6 +65,11 @@ class="forum__user-flag forum__user-flag--country"
         </div>
 
         <div class="forum-post__info-panel-extra-bottom">
+            <div class="forum-post__info-row">
+                <a class="forum-post__info-link" href="{{ route("users.posts", $user) }}">
+                    {{ trans_choice('forum.post.info.post_count', $user->user_posts) }}
```
nanaya: As it turned out, this errors out on reply since the value is raw sql statement and can't be converted to integer.

cl8n: should I change `db_unsigned_increment` so that it handles all of the database logic too? I'm not sure how to fix this

notbakaneko:
This PR is getting kind of stale, should we close it?

PR #3747, opened 2018-10-04
Fix usercard broken when loaded but user is null

notbakaneko:
Because the 2 conditions are not strictly exclusive if a user is deleted or restricted.

---


nanaya:
Looks like the whole block should be moved inside instead. Also the `$loading` check should be just `$loading` because `isset($loading)` is still true for any non-null value including `false`...

PR #3755, opened 2018-10-05
Rename isDeleted/withoutDeleted to trashed/withoutTrashed in model

nanaya:
Similar to the one in `SoftDeletes` trait.


app/Models/BeatmapDiscussion.php
```diff
@@ -519,11 +519,6 @@ public function denyKudosu($deniedBy)
         });
     }
 
-    public function isDeleted()
```
nanaya: There's `trashed` function already, doing exactly same thing :ok_hand:

PR #3758, opened 2018-10-06
Add more tooltips to beatmapset panels

### LINKED ISSUE #2854: Favourites don't have a tooltip on beatmap listing
https://streamable.com/i8tgx

TPGPL:
Closes #2854 

---


resources/lang/en/beatmaps.php
```diff
@@ -298,4 +298,8 @@
         'C' => 'C',
         'D' => 'D',
     ],
+    'panel' => [
+        'playcount' => 'Playcount: :count',
```
LiquidPL: not sure if there's a need to include the counts in the tooltip when they're already on the panel

PR #3786, opened 2018-10-11
Ensure mysql has been started before running php tasks (docker)

nanaya:
The other common(?) problem when using docker for first time.


notbakaneko:
conflictsss

PR #3796, opened 2018-10-12
Adjust .gitignore files modes

LiquidPL:
Each time the dev enviroment is brought up, modes of some .gitignore files get changed. Not sure if just commiting those changes is a good idea, so I'm welcome to any other suggestions.


notbakaneko:
let's not mark `.gitignore` files with `+x`...

LiquidPL:
I understand, but for some reason it just marks them like that each time i bring up my dev server, and I have no idea why.

LiquidPL:
In fact, it seems that all of `storage/` is at 777...
image

PR #3800, opened 2018-10-14
Use implied mods when converting to a bit set

### LINKED ISSUE #3799: Wrong speed in replay playback with NC mod when downloading from new site
In the new site, when downloading a NCHD replay of my Koigokoro [Hard] (97pp) play (from my top scores list - username is the_fbobis https://osu.ppy.sh/users/11736686), osu!client is playing back the replay in normal speed (and no pitch changes) rather than in Double Time (1.5x), though it shows the mods NC HD upon playing back the replay. 

However on the old site, when I download the same replay, it works fine, and plays in Double Time (1.5x) with the pitch change up +7 semitones. 

I have also discovered that if you change speed of replay downloaded from new site, at 2x it will be playing at its supposed speed (“Nightcore”), while at 0.5x you get a "Daycore". 

osu! version: 20180811.2 
osu!web version: 2018.1012.0

cl8n:
fixes #3799 ... probably. can't really test because I don't have the data

### DIFF
--- a/app/Libraries/ModsHelper.php
+++ b/app/Libraries/ModsHelper.php
@@ -84,6 +84,13 @@ class ModsHelper
         foreach (static::AVAILABLE_MODS as $availableMod) {
             if (in_array($availableMod[1], $mods, true)) {
                 $bitset ^= (1 << $availableMod[0]);
+
+                $impliedIds = $availableMod[2] ?? null;
+                if ($impliedIds !== null) {
+                    foreach ($impliedIds as $id) {
+                        $bitset ^= (1 << $id);
+                    }
+                }
             }
         }
 


app/Libraries/ModsHelper.php
```diff
@@ -84,6 +84,13 @@ public static function toBitset($mods)
         foreach (static::AVAILABLE_MODS as $availableMod) {
             if (in_array($availableMod[1], $mods, true)) {
                 $bitset ^= (1 << $availableMod[0]);
+
+                $impliedIds = $availableMod[2] ?? null;
+                if ($impliedIds !== null) {
+                    foreach ($impliedIds as $id) {
```
nanaya: or just `foreach ($availableMod[2] ?? [] as $id) {`

cl8n:
superseded by #3811 , didn't realize that this was already a converted attribute

PR #3803, opened 2018-10-15
Markdown updates

### LINKED ISSUE #3744: Make links in comments clickable
example taken from here https://osu.ppy.sh/beatmapsets/749150#osu/1578009

### LINKED ISSUE #3790: [beatmap comments] single newline does not work
you need two newlines for it to display one new line
example

image
image

nanaya:
- visible soft breaks (closes #3790)
- autolink (closes #3744)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,11 +25,12 @@
     "guzzlehttp/guzzle": "*",
     "itsgoingd/clockwork": ">=3.0",
     "jenssegers/agent": "^2.6",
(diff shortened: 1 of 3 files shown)


MegaApplePi:
Would this also affect how softbreaks are handled in the wiki? Not that it matters much since most stuff there shouldn't be using soft breaks anyways.

nanaya:
Confusing as it is, it doesn't affect wiki.

PR #3809, opened 2018-10-16
Pinned header for all pages

notbakaneko:
- Header is now always in a fixed position on all pages.
- Anchors should now always anchor to the top end of the page instead of trying to centre.
- Elements that stick on scroll like modding discussions tabs and new discussion text box now use `position: sticky` to stick.
- Forum floating header is replaced by a placeholder container attached to the main header
- Removes a bunch of `js-sync-height--reference` `js-sync-height--target`
  - elements that need to do sticky stacking should calculate their position directly

A mini search bar is added below the header on the beatmap listing page when scrolled down:
image

The back to top offset is currently wrong for the beatmap listing, but I'm leaving it for now since it ends up in a convenient position.

---


resources/assets/coffee/_classes/sticky-header.coffee
```diff
@@ -16,27 +16,60 @@
...
     @stickMarker = document.getElementsByClassName('js-sticky-header')
+    @visible = false
 
+    $(window).on 'throttled-scroll', @applyCss
     $(window).on 'throttled-scroll throttled-resize', @stickOrUnstick
     $(document).on 'turbolinks:load osu:page:change', @stickOrUnstick
 
 
+  applyCss: ->
+    header = document.getElementById('js-pinned-header')
```
nanaya: my guess says initializing/storing `document.getElementsByClassName` and using it like `@header[0]` is slightly faster.

resources/assets/less/bem/beatmapsets-search.less
```diff
@@ -60,22 +67,31 @@
...
+    display: flex;
+    justify-content: space-between;
+  }
+
   &__icon {
     position: absolute;
     right: 0;
     top: 0;
     height: 100%;
-    width: (@_icon-width + 20px);
```
nanaya: If you're wondering, this was originally intended so one day `math: strict` [option](http://lesscss.org/usage/#less-options-math) can be turned on.

(though apparently it's also needed for assigning variable which I haven't done)

notbakaneko:
variable position anchors are fun ಠ_ಠ

resources/views/layout/header.blade.php
```diff
@@ -55,3 +65,4 @@
         ])
     </div>
 @endif
+
```
nanaya: extra newline?

resources/assets/less/bem/forum-topic-floating-header.less
```diff
@@ -1,5 +1,5 @@
 /**
- *    Copyright 2015-2018 ppy Pty. Ltd.
+ *    Copyright 2015-2017 ppy Pty. Ltd.
```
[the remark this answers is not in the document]
notbakaneko: I saw this in the diff but couldn't find the change ಠ_ಠ

resources/assets/less/bem/beatmap-discussion-new-float.less
```diff
@@ -34,6 +34,13 @@
   margin-bottom: 10px;
   pointer-events: none;
 
+  &--pinned {
+    position: sticky;
+    position: -webkit-sticky;
```
nanaya: Looks like autoprefixer already adds this though in different order?

resources/assets/less/bem/page-extra-tabs.less
```diff
@@ -17,32 +17,16 @@
  */
 
 .page-extra-tabs {
-  @top: page-extra-tabs;
-  @bg: #333;
+  position: sticky;
+  position: -webkit-sticky;
```
nanaya: another one here

resources/assets/less/functions.less
```diff
@@ -29,6 +29,21 @@
   transition: @target @duration @type;
 }
 
+.fragment-target-fn(@offset: 0) {
+  &[id]::before {
```
nanaya: This blocks the content it's covering.

PR #3810, opened 2018-10-16
Switch beatmap listing to cursor-based pagination

### LINKED ISSUE #3368: Beatmaps Listing suddenly started to show the same songs
I was looking for new osu! songs. I downloaded a few of them, then i switched browser tab to the other site. But when i came back, beatmaps listing was looking like this (scrolling over and over will load the same beatmaps):
image

While writing this i just found that forcing scroll shows that other beatmaps are loaded, but they are overwritten just a while after:
https://img.wieku.me/18/2018-06-18%2018-28-06.mp4

notbakaneko:
closes #3368

React doesn't like duplicate keys.

The `/beatmapsets/search` endpoint now returns an object in the form of 
```
{ 
  "beatmapsets": [], 
  "cursor": {}
}
```
instead of an array, so anything using it will need to update to get data out of the `beatmaps` key.
`cursor` contains what should be sent to the endpoint for the next page.

---


app/Libraries/Elasticsearch/Search.php
```diff
@@ -141,6 +156,16 @@ public function response() : SearchResponse
         return $this->response;
     }
 
+    /**
+     * @return $this
+     */
+    public function searchAfter(?array $searchAfter)
```
nanaya: not used?

app/Libraries/Search/BeatmapsetSearchRequestParams.php
```diff
@@ -37,6 +37,10 @@ public function __construct(Request $request, ?User $user = null)
         $this->user = $user;
         $this->from = $this->pageAsFrom(get_int($request['page']));
 
+        if ($request['cursor'] !== null) {
+            $this->searchAfter = array_values($request['cursor']);
```
nanaya: may want to add `is_array` check.

resources/assets/coffee/react/beatmaps/main.coffee
```diff
@@ -179,19 +179,17 @@ class Beatmaps.Main extends React.PureComponent
   fetchNewState: (newQuery = false) =>
     @fetchResults(newQuery)
     .then (data) =>
-      more = data.length > 0
-
-      beatmaps: if newQuery then data else [].concat(@state.beatmaps, data)
+      beatmaps: if newQuery then data.beatmapsets else [].concat(@state.beatmaps, data.beatmapsets)
```
nanaya: as it turned out `concat` doesn't modify any of the arrays :dancer:

app/Libraries/Elasticsearch/Search.php
notbakaneko: for direct use in command line

nanaya: shouldn't this set `$this->params->searchAfter` then?

PR #3811, opened 2018-10-16
Use raw enabled_mods for replay header

### LINKED ISSUE #3799: Wrong speed in replay playback with NC mod when downloading from new site

notbakaneko:
Instead of encoding the display values

closes #3799

---

### DIFF
--- a/tests/Libraries/ReplayFileTest.php
+++ b/tests/Libraries/ReplayFileTest.php
@@ -41,7 +41,7 @@ class ReplayFileTest extends TestCase
     public function testHeaderChunk()
     {
         // known good header.
-        $known = '0056ef33010b2039323536623735613064663334356262623237396533353438303732396631340b08492e522e5265616c0b2033393336326435393239313738633162626265373465323130646639316232399e000600000016000500000059911f00460101180400000b008014b73dec8dd508';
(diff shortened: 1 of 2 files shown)


tests/Libraries/ReplayFileTest.php
```diff
@@ -41,7 +41,7 @@ public function testEndChunk()
     public function testHeaderChunk()
     {
         // known good header.
-        $known = '0056ef33010b2039323536623735613064663334356262623237396533353438303732396631340b08492e522e5265616c0b2033393336326435393239313738633162626265373465323130646639316232399e000600000016000500000059911f00460101180400000b008014b73dec8dd508';
+        $known = '0056ef33010b2039323536623735613064663334356262623237396533353438303732396631340b08492e522e5265616c0b2033393336326435393239313738633162626265373465323130646639316232399e000600000016000500000059911f00460101184200000b008014b73dec8dd508';
```
notbakaneko: We really need some kind of external decoder implementation that we can include for tests

PR #3825, opened 2018-10-20
Create country seeder with all countries

### LINKED ISSUE #3824: Full seeding for countries
Would it be okay if someone from the main team mysqldump the `osu_countries` table from the live database? Just the `acronym`, `name`, `display` and `shipping_rate` columns are enough if all columns are too much to ask. Of course this is assuming all countries exist in prod and not just the ones with sizeable userbases.

If mysqldump is not viable, can you tell me which standard you use in `osu_countries`.`acronym` column so I can create a valid seeder for it?

This would be better for initial setup on local because of 2 things:

1. Some issues could be related to countries, and having all of them in database is better for debug purposes.
2. When the current user's `country_acronym` field doesn't exist in `osu_countries` table it throws an exception, which is annoying for first time setups when the developer doesn't know what's going on (e.g. myself)

taylankasap:
- Closes #3824

---

Database will have all countries instead of 50 fake countries when you run the seeders.


notbakaneko:
@taylankasap you'll need to enable `Allow edits from maintainers.` on the PR before we can merge this

taylankasap:
@notbakaneko okay I enabled it

PR #3834, opened 2018-10-25
Add ability to sort comments

nanaya:
- check if anything is broken
- fix sort option links so they look clickable
- sanity check
  - ~~params parser is getting dumb~~
  - query building for the cursor probably belongs somewhere else
  - ?


nanaya:
I think it's fine-ish now °～° hi @notbakaneko

resources/assets/coffee/react/_components/comment-show-more.coffee
```diff
@@ -30,7 +30,14 @@ class @CommentShowMore extends React.PureComponent
       loading: false
 
 
+  componentWillUnmount: =>
+    @xhr?.abort()
+
+
   render: =>
+    return null if @props.comments.length >= @props.total
+    return null unless (@props.moreComments[@props.parent?.id ? null] ? true)
```
notbakaneko: probably needs check for `@props.moreComments` or set `defaultProps` else it might explode when viewing permalinked comments with deep nested replies.

We might want to consider setting `defaultProps` on the other components too

resources/assets/coffee/react/_components/comments-manager.coffee
```diff
@@ -56,34 +61,53 @@ class @CommentsManager extends React.PureComponent
...
+      commentableMeta: _.concat @state.commentableMeta, commentBundle.commentable_meta
+      moreComments: moreComments
 
 
   update: (_event, {comment}) =>
     @setState
-      comments: osu.updateCollection @state.comments, [comment]
-      users: osu.updateCollection @state.users, [comment.user, comment.editor]
-      commentableMeta: _.concat @state.commentableMeta, [comment.commentable_meta]
+      comments: @mergeCollection @state.comments, [comment]
```
notbakaneko: does this mean `osu.updateCollection` isn't used anymore? 🤔

resources/assets/coffee/react/_components/comment-show-more.coffee
notbakaneko: time to start using `propTypes` :trollface:

PR #3838, opened 2018-10-26
Adjust beatmapset panel transition times

### LINKED ISSUE #3754: Odd Transition on Beatmap Card Hover
https://streamable.com/62fti

The animation looked janky. Please set the transition period the same for both items. Use slightly longer animation period.

LiquidPL:
Apparently the rank status pill outright ignored the transitions.

Resolves #3754.

---

Fix wonky transitions of the play button and ranked/video status on beatmapset panels.
animations


resources/assets/less/bem/beatmapset-panel.less
```diff
@@ -166,11 +167,24 @@
     display: flex;
   }
 
+  &__status {
```
notbakaneko: there isn't an element `__status`?

PR #3841, opened 2018-10-27
Improve new forum topic preview

### LINKED ISSUE #3558: New thread preview is not obvious that it's a preview
image
there's nothing really indicating that it is a preview, gave me a scare that I had accidentally posted the thread already without meaning to. This needs to be more clearly labelled.

LiquidPL:
Resolves #3558.

---

Moved topic preview so it's in the same section as the topic editor, in order to avoid confusion.
image


resources/views/forum/topics/create.blade.php
```diff
@@ -50,6 +40,15 @@ class="forum-post__stripe"
                 @include("forum.topics._post_info", ["user" => Auth::user(), "options" => ["large" => true]])
 
                 <div class="forum-post__body">
+                    <div class="forum-post__preview js-post-preview--box hidden">
+                        <div class="forum-post__content forum-post__content--header">
+                            post preview
```
TPGPL: Localise

resources/assets/less/bem/forum-post.less
```diff
@@ -255,6 +259,10 @@
     }
   }
 
+  &__preview {
+    background-color: #cccccc;
```
notbakaneko: use the shorthand for this

resources/lang/en/forum.php
```diff
@@ -153,6 +153,8 @@
         'create' => [
             'create_poll' => 'Poll Creation',
 
+            'preview' => 'post preview',
```
notbakaneko: probably should be Title Case

PR #3854, opened 2018-11-02
New messaging/chat system

nekodex:
Okay, this one is gonna be a doozy.

This chat system was done as a prototype test for a possible future direction for osu-web.

In addition to the actual chat interface, it contains 3 new major core additions:
- The addition of TypeScript as a potential replacement for CoffeeScript
- A MobX-based centralised data-store for user/chat info (which will eventually be extended to cover other objects and be smart enough to auto-retrieve/cache as required)
- An event-driven/dispatcher architecture for systems that have complex UI interactions

Not sure if this new event-driven/dispatcher architecture is the direction we want to use for everything going forward?... It might be overkill for simpler things? I'm open to discussion.

The roll-out will be limited to osu! team members initially for usability and scalability testing/feedback before being rolled out to more users.

---


resources/assets/ts/actions/chat-actions.ts
```diff
@@ -0,0 +1,40 @@
...
+ *
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ */
+// tslint:disable:max-classes-per-file
+import Message from 'models/chat/message';
+import DispatcherAction from './dispatcher-action';
+
+export class ChatChannelSwitchAction implements DispatcherAction {
+  constructor(public channelId: number) {
```
notbakaneko: Is there a reason this takes `number` instead of `Channel` like how the other actions accept `Message`?

resources/assets/ts/models/chat/channel.ts
```diff
@@ -0,0 +1,133 @@
...
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { action, computed, observable, transaction} from 'mobx';
+import Message from './message';
+
+export interface ChannelJSON {
+  channel_id: number;
+  type: 'PUBLIC'|'PRIVATE'|'MULTIPLAYER'|'SPECTATOR'|'TEMPORARY'|'PM'|'GROUP';
```
notbakaneko: could do with a type alias (and also change `Channel.type` to it?)

app/Http/Controllers/ChatController.php
```diff
@@ -0,0 +1,57 @@
...
+    public function __construct()
+    {
+        $this->middleware('auth');
+
+        return parent::__construct();
+    }
+
+    public function index()
+    {
+        if (!Auth::user()->isPrivileged()) {
```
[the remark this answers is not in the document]
nekodex: pretty sure the auth middleware prevents `Auth::user` from being null here

resources/assets/ts/chat.ts
```diff
@@ -0,0 +1,37 @@
...
+import MainView from './chat/main-view';
+import OsuCore from './osu-core';
+
+declare global {
+  interface Window {
+    OsuCore: OsuCore;
+  }
+}
+
+const core: OsuCore = window.OsuCore = window.OsuCore || new OsuCore(window);
```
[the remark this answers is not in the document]
nekodex: Not sure. My reasoning for this is the case where `OsuCore` is eventually used in other components too, ensuring that it's only instantiated once (without making it a singleton?)

resources/assets/ts/dispatcher.ts
```diff
@@ -0,0 +1,38 @@
...
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import DispatcherAction from './actions/dispatcher-action';
+import DispatchListener from './dispatch-listener';
+
+export default class Dispatcher {
+  private callbacks: DispatchListener[] = [];
+  private trace: boolean = false;
```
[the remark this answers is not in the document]
nekodex: enables debug output, console.logs actions that are dispatched

resources/assets/ts/chat/main-view.tsx
```diff
@@ -0,0 +1,108 @@
...
+export default class MainView extends React.Component<PropsInterface, any> {
+  constructor(props: any) {
+    super(props);
+
+    if (!_.isEmpty(props.presence)) {
+      this.props.dataStore.channelStore.updatePresence(props.presence);
+    }
+  }
+
+  componentDidMount() {
```
notbakaneko: most (all?) of this seem like they should be in the constructor or set up before `MainView` instead of after it's rendered once

resources/assets/ts/chat/chat-state-store.ts
```diff
@@ -41,7 +41,10 @@ export default class ChatStateStore implements DispatchListener {
 
   handleDispatchAction(action: DispatcherAction) {
     if (action instanceof ChatChannelSwitchAction) {
-      this.lastReadId = this.root.channelStore.getOrCreate(action.channelId).lastReadId;
+      const lastReadId = this.root.channelStore.getOrCreate(action.channelId).lastReadId;
+      if (lastReadId) {
```
notbakaneko: what happens if it's `0`? can it even be `0`?

PR #3855, opened 2018-11-02
Add reporting and replay downloads to best scores

notbakaneko:
---


app/Http/Controllers/ScoresController.php
```diff
@@ -0,0 +1,53 @@
...
+namespace App\Http\Controllers;
+
+use App\Models\Beatmap;
+use App\Models\Score\Best\Model as ScoreBest;
+use Auth;
+use PDOException;
+
+class ScoresController extends Controller
+{
+    public function report($mode, $id)
```
nanaya: If score validates id and mode, can't this be merged to `Users#report`? Looks like it'll simplify some things like the form.

resources/assets/less/bem/beatmap-scoreboard-table.less
```diff
@@ -18,41 +18,62 @@
...
+    .@{passive-row}.@{name} & {
+      background-color: lighten(@color, 20%);
+    }
+
+    .@{passive-row}.@{name}:hover & {
+      color: #fff;
+      background-color: @color;
+
+      .@{_top}--menu-active & {
+        color: unset;
```
nanaya: this turns gray `0` to black. Also weird highlight is disabled but user card is still shown. The invisible menu burger button also still works. A `pointer-events: none` at top level might be simpler?

resources/assets/lib/play-detail-menu.coffee
```diff
@@ -17,45 +17,91 @@
...
+      @removePortal()
 
 
   componentDidUpdate: (_prevProps, prevState) =>
-    if prevState.active != @state.active
-      if @state.active then @props.onShow?() else @props.onHide?()
+    return if prevState.active == @state.active
+
+    if @state.active
+      if @portal?
```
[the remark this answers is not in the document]
notbakaneko: it also doesn't follow scroll when in an element independently scrollable from `body` (like the scoreboard table ヽ( ﾟ ヮﾟ)ﾉ)

app/Http/Controllers/ScoresController.php
notbakaneko: `UsersController#report` intentionally checks if it should 404 or not similarly to `#show`; the `ReportForm` component intentionally doesn't contain any network submission logic and the state and `onSubmit` parts of `ReportUser` and `ReportScore` are planned to be migrated to be `Provider`-based where they'd share the same implementation

PR #3856, opened 2018-11-02
fix(gitattributes): Files should be LF by default now

### LINKED ISSUE #3851: [docker] php exits with code 1
logs: https://pastebin.com/ScnJ2nfX

`standard_init_linux.go:190: exec user process caused "no such file or directory"` seems to make trouble, but changing line-ending auto conversion for git didnt solve the problem

sr229:
Tells Git to force files to be checked-out and normalize all files in index as LF by default.

- Closes #3851 

---

There are instances where files in the repository when checked out end up with CRLF. `gitattributes` has been modified to always check out and push as LF by default.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
 * text=auto
-*.sh eol=lf
+* eol=lf


cl8n:
```
* text=auto eol=lf
*.* -eol
*.sh eol=lf
```

sr229:
Should be applied now.

.gitattributes
```diff
@@ -1,2 +1,3 @@
-* text=auto
+* text=auto eol=lf
+*.* -eol
```
nanaya: This isn't needed and will cause files to be converted to platform-specific newlines.

sr229: This is kinda needed? We might end up with a repository with mixed endings.

sr229: The main rationale here was @TPGPL having issues with their PR, I triaged it to line endings for some reason. Keep in mind this is CRLF on a **Linux** Environment - this will cause issues.

nanaya: That's not what this line does.

cl8n: my bad for suggesting this change without an explanation---this PR title no longer has anything to do with the changes, we just wanted to fix https://github.com/ppy/osu-web/issues/3851. my suggested changes are meant to only force LF for files with no extension

cl8n:
not related to 3851, but you might as well force LF on tests/Libraries/bbcode_examples/* too, some bbcode tests fail otherwise

sr229:
@cl8n the glob matches all files however, if I'm not mistaken.

PR #3860, opened 2018-11-04
Fix comments in beatmapsets/events

### LINKED ISSUE #3686: Disqualifications do not show a preview in the events listing
Disqualifications currently just say `[no preview]` on the `/event` page.

ss 2018-09-16 at 07 39 40

This should be changed into actually showing the disqualification post's content:
ss 2018-09-16 at 07 42 13

ss 2018-09-16 at 07 43 31

milaneyia:
Attempt to fix it by adding a new field to the array passed, will fix the ones created from now on.

- Closes #3686

---

Disqualifications currently just say [no preview] on the /event page, from now on will display the actual message.

### DIFF
--- a/app/Models/BeatmapsetEvent.php
+++ b/app/Models/BeatmapsetEvent.php
@@ -54,6 +54,7 @@ class BeatmapsetEvent extends Model
             $discussionPostId = $object->getKey();
             $discussionId = $object->beatmap_discussion_id;
             $beatmapsetId = $object->beatmapDiscussion->beatmapset_id;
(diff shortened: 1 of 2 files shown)


peppy:
Is this still valid? Is there a reason this hasn't been reviewed in 2ish months?

nanaya:
Should just get the relevant posts and show them instead of storing extra copy of the message.

milaneyia:
iirc it was getting too slow when i tried that, but yes, should be that way
i'll close this since i don't have the time to actually do it now

PR #3863, opened 2018-11-07
Add API scopes

### LINKED ISSUE #3857: Add API scopes
at the very least I think it'd make sense to have:
- reading private data (e.g. friends list)
- writing data (e.g. sending chat messages)
- identifying (only `/api/v2/me`)

just so that I don't have to let the user blindly trust that my app doesn't do something it shouldn't (though fwiw, I think a more extensive set of scopes would be a lot better)

I can look into adding this myself if it's not a dev priority, but I figure someone else should decide which scopes exist

cl8n:
closes #3857 

~~adds five scopes: `beatmapset.download`, `chat.read`, `chat.write`, `user.identify` and `user.read`~~
https://github.com/ppy/osu-web/pull/3863#issuecomment-436904257

---

### DIFF
--- a/app/Http/Middleware/CheckScopesForAPI.php
+++ b/app/Http/Middleware/CheckScopesForAPI.php
@@ -18,17 +18,28 @@
  *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-return [
-    'error' => [
-        'chat' => [
-            'limit_exceeded' => 'You are sending messages too quickly, please wait a bit before trying again.',
-            'too_long' => 'The message you are trying to send is too long.',
-        ],
-    ],
+namespace App\Http\Middleware;
 
-    'scopes' => [
-        'identify' => 'Identify your osu! account',
-        'read' => 'View your private profile information',
-        'write' => 'Perform actions on your behalf',
-    ],
-];
+use Laravel\Passport\Http\Middleware\CheckScopes;
+
+class CheckScopesForAPI
+{
+    /**
+     * Handle an incoming request.
+     *
+     * @param \Illuminate\Http\Request $request
+     * @param \Closure                 $next
+     * @param  mixed                ...$scopes
+     * @return \Illuminate\Http\Response
+     * @throws \Illuminate\Auth\AuthenticationException
+     * @throws \Laravel\Passport\Exceptions\MissingScopeException
+     */
+    public function handle($request, $next, ...$scopes)
+    {
+        if (is_api_request()) {
+            return app(CheckScopes::class)->handle($request, $next, ...$scopes);
+        }
+
+        return $next($request);
+    }
+}
--- a/routes/web.php
+++ b/routes/web.php
@@ -287,9 +287,10 @@ Route::group(['as' => 'api.', 'prefix' => 'api', 'namespace' => 'API', 'middlewa
             Route::apiResource('channels', '\App\Http\Controllers\Chat\ChannelsController', ['only' => ['index']]);
         });
 
-        Route::resource('rooms', 'RoomsController', ['only' => ['show']]);
+        Route::apiResource('rooms', 'RoomsController', ['only' => ['show']]);
 
         Route::group(['prefix' => 'beatmapsets'], function () {
+            // TODO: Should be moved under /me, and probably call to UsersController instead?
             Route::get('favourites', 'BeatmapsetsController@favourites');     //  GET /api/v2/beatmapsets/favourites
         });
 
@@ -299,7 +300,7 @@ Route::group(['as' => 'api.', 'prefix' => 'api', 'namespace' => 'API', 'middlewa
         //   GET /api/v2/beatmaps/lookup
         Route::get('beatmaps/lookup', 'BeatmapsController@lookup');
         //   GET /api/v2/beatmaps/:beatmap_id
-        Route::resource('beatmaps', 'BeatmapsController', ['only' => ['show']]);
+        Route::apiResource('beatmaps', 'BeatmapsController', ['only' => ['show']]);
 
         // Beatmapsets
         //   GET /api/v2/beatmapsets/search/:filters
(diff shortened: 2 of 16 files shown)


cl8n:
that does seem simpler; how would I do that though since lots of these controllers/actions are not specific to the API?

nekodex:
There's `is_api_request()` in `helpers.php`

routes/web.php
```diff
@@ -274,65 +274,73 @@
...
-            });
-            Route::apiResource('channels', '\App\Http\Controllers\Chat\ChannelsController', ['only' => ['index']]);
-        });
+        // No scopes required
+        Route::apiResource('rooms', 'RoomsController', ['only' => ['show']]);
 
-        Route::resource('rooms', 'RoomsController', ['only' => ['show']]);
-
-        Route::group(['prefix' => 'beatmapsets'], function () {
-            Route::get('favourites', 'BeatmapsetsController@favourites');     //  GET /api/v2/beatmapsets/favourites
```
nekodex: Why is this being removed?

nekodex:
Also after a bit of internal discussion, we're thinking just three scopes initially is enough:
 - `identify`
 - `read`
 - `write`

`identify` and global `read` being available to third-party clients, but global `write` only available for lazer to begin with (AFAIK chat-related endpoints would be the only 'write' ones ATM, with things like account settings being added later).

Once the API is more finalised later on, we'll add more granularity to the read/write scopes.

routes/web.php
cl8n: you can get the same data from the users beatmapsets route, and it seems like that's what the osu! client does already

nekodex: IIRC Lazer is using that endpoint because it's only used for displaying user profiles of other users (i.e. it doesn't do anything with the current users' favourite beatmaps in any way yet).

I'd argue it's worth preserving the favourite beatmaps URL endpoint for convenience sake (maybe moving it to `/me/beatmapsets/favourites`) even if it ends up being an alias to `UsersController@beatmapsets` instead.

Having said that, removing/altering API endpoints should be done in a separate pull request instead of as part of this (unrelated) one.

app/Http/Middleware/CheckScopesForAPI.php
```diff
@@ -0,0 +1,45 @@
...
+     * Handle an incoming request.
+     *
+     * @param \Illuminate\Http\Request $request
+     * @param \Closure                 $next
+     * @param  mixed                ...$scopes
+     * @return \Illuminate\Http\Response
+     * @throws \Illuminate\Auth\AuthenticationException
+     * @throws \Laravel\Passport\Exceptions\MissingScopeException
+     */
+    public function handle($request, $next, ...$scopes)
```
cl8n: is this good practice? I don't know if middleware are supposed to call other middleware

Itsyuka: You might as well just use `Laravel\Passport\Http\Middleware\CheckScopes::class` instead of this file, since you'll only need to be defining the scopes on API routes anyways.

cl8n: they're defined on the controllers, not the routes (see conversation above), and most of these controllers are not specific to the API

Itsyuka: I see, if that's the route they are willing to take, then there isn't really a choice. So by their standard, it's fine.

cl8n:
good idea, added

cl8n:
why should an empty scope give no grant? the things that don't have scope requirements are just public info and read-only

notbakaneko:
The default permission is going to be deny, not public. Permission grants need to be explicit; oauth clients should not be performing actions where permission is not granted; an empty scope requests no permissions.

peppy:
While I don’t know best practices with scopes, it would make sense to match the behaviour of the other sites’ APIs you mention.

notbakaneko:
Regardless of of the implementation, the thing to remember is: permissions granted by the user should be explicit.

This means that when an OAuth authorization prompt appears, it needs to display the permissions that are going to be given to it. The default prompt isn't good enough:
screen shot 2018-11-19 at 16 08 32

Access your account to do what? View details? See who you're following? Build a monkey boat? This isn't an issue if the permissions are only allow/deny all, but when you consider that `identify` shows this, what does the previous one mean?
screen shot 2018-11-19 at 19 32 10

The existing prompt doesn't quite explain the extent of permissions being granted if public is being made the default.

Compare with Github's when no scopes are requested:
screen shot 2018-11-19 at 17 24 15

Endpoint responses can contain data that requires additional permissions, and accessing an endpoint with this token doesn't include that data.


# Implicit scoping
Requested scopes should be normalized into their explicit forms; this makes auditing, changes and review simpler.

e.g. If having no scope provided implies `public`, the scope should be normalized to `public` when it's issued.
If requesting `read` implies `public` permissions, then it should be normalized as `read public` and listed and granted as separate permissions.

This way if `read` gets more implicit permissions in the future, existing tokens won't automatically get permissions they weren't originally granted. If implicit permissions were removed, it's not an issue since they were initially granted on the existing tokens.


# 3rd party authorization
There needs to be a distinction between tokens  
- issued to allow a 3rd party to act on behalf of a user,
- issued directly to the user.


## Tokens issued to 3rd parties 
Token granted from authorization codes allow 3rd parties access to information and allow actions to be performed on your behalf. 

This type of token shouldn't be doing anything on your behalf that wasn't granted; so while it can access public endpoints, it shouldn't be accessing it _as_ you, if it wasn't granted the permission to do so, as some endpoints can and do contain additional data according to the user accessing them that are _not_ public.


## Personal tokens
In contrast to tokens for 3rd party use, this kind of token requested by the user themselves for their own use. They're basically equivalent to logging in with your username and password; getting these issued will typically require verifying with your login, so in these cases we can make the assumption that permission to access as the user is granted.

The expectation here is these tokens are private to the user and not shared with any 3rd party.


---

Deny by default makes it easier and quicker to review and audit changes.
The way the scope middleware is applied makes it too easy for actions that are not supposed to be in the public scope become public.


We want to get API authorization out so people can stop doing dodgy things like giving their logins to everyone, but until all the the endpoints are audited and have any fixes necessary to not include or allow actions they are not supposed to, the default implicit scope is doing to be deny all.

cl8n:
I wasn't aware that implicit scoping was done in these cases, your reasons make more sense now. thanks

for this PR, should I just add a "public" scope to the routes that don't have any scopes currently?

notbakaneko:
The existing routes with no scopes defined should be not be publicly accessible for now, until we can finish auditing them. They'll have the scopes required for access added to them later after they're reviewed properly.

notbakaneko:
we're going to go with more limited permissions in #3921 first and then roll out additional ones

PR #3874, opened 2018-11-13
Allow editing of forum polls

### LINKED ISSUE #1711: Polls cannot be edited in new forum view
Editing an OP that contains a poll using the new forum interface provides no means to alter details of the poll in any capacity whatsoever.

notbakaneko:
closes #1711

Allows polls to be edited during a certain time window after they're created.

The existing `TopicPoll` related stuff probably needs some cleaning up, but I ended up skipping that since it'd requiring changing and checking too many other things.

---


resources/views/forum/topics/_poll.blade.php
```diff
@@ -15,55 +15,79 @@
...
+                {{ priv_check('ForumTopicVote', $topic)->message() }}
+            @else
+                <button class="btn-osu-lite btn-osu-lite--default js-checkbox-validation--submit" disabled>
+                    {{ trans('forum.topics.show.poll.vote') }}
+                </button>
+            @endif
+        </div>
+    {!! Form::close() !!}
+
+    @if ($canEditPoll ?? false)
```
nanaya: probably should be passed as well from the poll edit post action.

PR #3886, opened 2018-11-15
Replace timeago with moment

### LINKED ISSUE #3762: Years should not round up at 0.5
https://osu.ppy.sh/beatmapsets/584128
This was updated 1 year and 7 months ago, it shows as graved two years ago on my profile as it rounds up at 0.5
Rounding up years should not be a thing

screen shot 2018-10-07 at 4 48 51 pm

notbakaneko:
Removes `timeago` library and uses `moment` for relative time display.
We're already using `moment` for the tooltips so `timeago` itself offers very little of use.
<em>updates automatically</em>

Also set relative times to round down; closes #3762

~~whoops, need more fixes first :D~~

---

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -79,7 +79,7 @@ return [
     'locale' => 'en',
 
     /*
-     * Make sure to check locale name in timeago, momentjs, and carbon.
+     * Make sure to check locale name in momentjs, and carbon.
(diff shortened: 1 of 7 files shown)


notbakaneko:
Going to hold on this while we look at some other alternatives

notbakaneko:
replacing `moment` and `timeago` with `date-fns` actually seems to be a better option

PR #3921, opened 2018-11-27
Apply OAuth permission scoping to API

notbakaneko:
- Makes the API endpoints require OAuth tokens to have the correct scopes.
- ~~Implicit scope normalization isn't included in this change, so for the time being, scopes will have to be explicitly requested.~~

The current checking of permissions and the ability to filter output based on permissions is kind of crap, so we're probably going to have to do something about that too.

Initially, the only scope available for third-party authorization is `identify` for `/me`; other endpoints and scopes will be made available as they're checked.

Expected behaviour should be:
- `require-scopes` with parameters should require tokens to have all the scopes listed
- `require-scopes` without parameters should deny access by default except for tokens with `*` scope, or more explicit scopes are required
- only `password` grant clients can request `*` scope

lazer needs to be update to include `"scope": "*"` in its token requests

---

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -28,6 +28,7 @@ use Illuminate\Auth\AuthenticationException;
 use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Foundation\Exceptions\Handler as ExceptionHandler;
 use Illuminate\Session\TokenMismatchException;
+use Laravel\Passport\Exceptions\MissingScopeException;
 use Sentry;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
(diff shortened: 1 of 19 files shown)


grumd:
Hey, sorry for commenting here, but I don't know where else to ask, and I didn't want to create an issue for a simple question. Is it possible to use api v2 from a third party server right now? I was able to create a client using `/oauth/clients`, then I was able to create a token using `/oauth/token` with `'grant_type': 'client_credentials'`, but the Bearer token I received doesn't let me use any v2 endpoints, they simply return `{ "authentication": "basic" }` even when I send an appropriate header `Authorization: Bearer <token>`

peppy:
The new API is not officially ready for use yet. It is also in a state of development. Unless you are happy to figure things out yourself and with things potentially breaking (specs are not final) then it is not suited for third-party use yet.

grumd:
@peppy Yeah I understand that it's still in development, and that I would have to deal with bugs and/or constant changes in the API. I'm fine with that. My question is simply whether it's *possible* to use it -- because right now I have troubles with authentication and I seemingly tried everything I could but it doesn't work. I can keep scraping htmls instead, but I'm willing to switch to an unstable API v2 if you guys would just briefly guide me what I need to do to successfully authenticate with API v2. I described above what I could accomplish in 3-4 hours of reading osu-web code and online docs, but I'm stuck right now. Pretty please! I would be a beta tester for API v2 and would try to fix any bugs I find with PRs here if I'm able to!

PR #3933, opened 2018-12-03
Fix anchor offsets on news pages

### LINKED ISSUE #3928: News anchor links have wrong offset
Pinned navigation blocking the header.

notbakaneko:
closes #3928 

We should probably look at supporting markdown extensions for custom header ids (which seems to have made no progress in commonmark) instead, like:

```
# Header {#anchor-here}
```

---

### DIFF
--- a/resources/assets/less/bem/osu-md.less
+++ b/resources/assets/less/bem/osu-md.less
@@ -66,6 +66,10 @@
       margin-top: 0;
     }
 
+    & > a {
+      .fragment-target-fn();
+    }
+
     &--2 {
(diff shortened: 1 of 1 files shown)


resources/assets/less/bem/osu-md.less
```diff
@@ -66,6 +66,10 @@
       margin-top: 0;
     }
 
+    & > a {
```
nanaya: `&` optional. And probably should look into setting `&[id], [id] { ... }` instead (not sure if better).

PR #3944, opened 2018-12-06
Add post and profile length limit

### LINKED ISSUE #3943: User page has no character limit
Just as the title says, editing user pages have no character limit.
I tried saving my user page with over 250,000 characters, an error occured but still the contents are saved. After that, my profile is totally inaccessible, just a blank page appears either in old and new. The only way to remove it is by using the moderation tools.

Adding a character limit (e.g. 60,000 like in the old site) will avoid this issue.

notbakaneko:
closes #3943

Client-side check to be handled separately.

Error message makes no sense imo, since the validation doesn't have any info on what context it's actually being used for.

---


app/Models/Forum/Post.php
```diff
@@ -223,6 +223,10 @@ public function isValid()
             $this->validationErrors()->add('post_text', 'required');
         }
 
+        if ($this->isDirty('post_text') && mb_strlen($this->post_text) > config('osu.forum.max_post_length')) {
```
nanaya: This should check `body_raw` length otherwise this will include length of the bbcode uid and other additional attributes.

PR #3993, opened 2018-12-11
Enable feature vote

### LINKED ISSUE #1173: Voting on feature requests
We need to display current star priority at a minimum

nanaya:
With very basic interface :dancer: 

Now with stars :stars: 

Resolves #1173.


nanaya:
adding stars

nanaya:
added stars

nekodex:
You may want to double-check the queries...

resources/views/forum/topics/_feature_vote.blade.php
```diff
@@ -0,0 +1,71 @@
...
+            {!! trans('forum.topics.show.feature_vote.info._', [
+                'feature_request' => '<strong>'.trans('forum.topics.show.feature_vote.info.feature_request').'</strong>',
+                'supporters' => link_to(route('support-the-game'), trans('forum.topics.show.feature_vote.info.supporters')),
+            ]) !!}
+        </div>
+
+        <div
+            class="
+                forum-topic-feature-vote__votes
+                {{ $topic->osu_starpriority > 1000 ? 'forum-topic-feature-vote__votes--small' : '' }}
```
nekodex: might be worth moving this to a config variable?

resources/lang/en/sort.php
```diff
@@ -16,25 +18,15 @@
...
+    'new' => 'Recent',
+    'old' => 'Old',
+    'top' => 'Top',
 
-  &__item-arrow {
-    margin-left: 5px;
-  }
-}
+    'forum_topics' => [
+        'new' => 'Last replied',
```
nekodex: "Last replied" kinda implies _my_ last reply, so maybe "Last reply"? (also might want to update the topic listing to match too)

resources/lang/en/forum.php
```diff
@@ -268,6 +269,12 @@
...
                 'do' => 'Promote this request',
 
+                'info' => [
+                    '_' => 'This is a :feature_request. Feature requests can be voted up by :supporters.',
+                    'feature_request' => 'feature request',
+                    'supporters' => 'supporters',
+                ],
+
                 'user' => [
                     'count' => '{0} no vote|{1} :count vote|[2,*] :count votes',
```
nekodex: Noticed this while testing... not due to PR changes but probably should still be fixed -- `no vote` should be `no votes`...



(may need to doublecheck where else the string is used)

PR #3996, opened 2018-12-11
Remove exists dependent subquery when validating previous users of a username

notbakaneko:
Bad query when there are no matches.
Also move query to readonly database.

---

### DIFF
--- a/app/Libraries/UsernameValidation.php
+++ b/app/Libraries/UsernameValidation.php
@@ -124,8 +124,8 @@ class UsernameValidation
 
     public static function usersOfUsername(string $username) : Collection
     {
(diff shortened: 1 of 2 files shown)


app/Libraries/UsernameValidation.php
```diff
@@ -123,10 +124,8 @@ public static function validateUsersOfUsername(string $username) : ValidationErr
 
     public static function usersOfUsername(string $username) : Collection
     {
-        $users = User::whereHas('usernameChangeHistory', function ($query) use ($username) {
-            $query->where('username_last', $username);
-        })->get();
-
+        $user_ids = UsernameChangeHistory::on('mysql-readonly')->where('username_last', $username)->pluck('user_id');
```
nanaya: can use some camelcase variable.

PR #3999, opened 2018-12-12
Change user page requirement

### LINKED ISSUE #3995: Can't edit "me!" section of user profile if supporter tag has expired
Users who don't have an active supporter tag can't edit the "me!" section of their profile if they leave it blank, even if they've already had a userpage in the past. 

Users who leave it blank see this message: "You have to be an osu!supporter to unlock this feature."

This seems like unwanted behaviour since users are able to keep editing their userpage / me! after their supporter tag expires, except if they leave it blank.

nanaya:
Changed from having active supporter tag to ever having supporter tag. Fixes #3995.


Nath789:
Team members can't edit user page of other users. We should be able to edit them like in the old site.

PR #4010, opened 2018-12-13
Move typescript files into lib

notbakaneko:
Going to have the *script modules under the same directory going forwards.

---


nanaya:
It's a bit confusing calling the directory `lib` but can't move it to `js` at the moment - which I think is a better name - because there's already stuff in it which doesn't use module loader thingy.

Also laravel the framework removed assets directory and moved stylesheets (`sass` for some reason) and javascripts (`js`) to [the same directory as `views` and `lang`](https://github.com/laravel/laravel/tree/d1db3cd089d6e9f81a4c8a65d401939f85cbbd5c/resources).

Probably it's fine to merge it now and look at it again later.

PR #4020, opened 2018-12-18
Add support for reporting comments

### LINKED ISSUE #3918: Add an option to report users in beatmap comments
people writing inappropriate stuff / spam is fairly common, so i'd say this is needed

notbakaneko:
closes #3918

Defaults all comment reports to `Spam` for now, will need to allow more options in the future.
Changed `UserReport::save` to always throw if not valid when saving.

---


app/Models/UserReport.php
```diff
@@ -28,6 +30,14 @@ class UserReport extends Model
     use Validatable;
 
     const CREATED_AT = 'timestamp';
+    const REPORTABLES = [
+        'comment' => Comment::class,
+        'user' => User::class,
+        'score_osu' => Best\Osu::class,
```
nanaya: Might be confused with `Score` class

PR #4044, opened 2018-12-28
Fix changelog page display being broken on some browsers

peppy:
Already deployed, PRing for consistency with master.


notbakaneko:
shouldn't be needed if #4057 behaves as expected and nothing breaks

Issue #3729, opened 2018-09-28
A few glitched thumbnails of maps on the new website


Celektus:
The thumbnails of these maps are glitched on the new site 

- [https://osu.ppy.sh/beatmapsets/599552#osu/1278479](url) 
- [https://osu.ppy.sh/beatmapsets/644520#osu/1365964](url)
- [https://osu.ppy.sh/beatmapsets/599182#osu/1266380](url)

Same issue as https://github.com/ppy/osu-web/issues/2453 as far as I'm concerned. Would be great if you could fix these quickly

![dlynlcmw4aivmjo](https://user-images.githubusercontent.com/22661428/46224910-e963c100-c357-11e8-8680-08776dab7feb.jpg)

nekodex:
Triggering a regeneration of the covers seems to have fixed them.

We'll probably run a reprocess across all maps at some point but for the time being if anyone notices other maps with this issue, please post them in this issue and we'll manually regenerate them for now!

Gonna leave this issue open until we can run the bulk reprocess across all beatmaps.

Lyawi:
https://osu.ppy.sh/beatmapsets/751640#taiko/1582345
https://osu.ppy.sh/beatmapsets/707146#taiko/1553641

omkelderman:
list of beatmap set ids of which a cover.jpg does not exist:

- 43
- 123
- 132
- 1573
- 3756
- 48523
- 55226
- 89535
- 90930
- 114389
- 228460
- 241365
- 243338
- 252981
- 262053
- 286460
- 297018
- 304958
- 305286
- 330927
- 347556
- 360153
- 374550
- 388428
- 396591
- 412371
- 413610
- 437113
- 506213
- 618054
- 694649
- 707334

cl8n:
@omkelderman, some of these mapsets don't have backgrounds. they're not supposed to have a cover

omkelderman:
this list is generated out of a dataset of an application im running, I checked a few manually to see if they had a background and wrongly assumed all of them did, sorry about that. Manually checked all of them and indeed found a few, will edit my prev comment to take those away.

TheMoMan:
https://osu.ppy.sh/beatmapsets/463701
I want to say this one broke because the background is placed in a /sb directory.

omkelderman:
few more that have no background on new web, but do have an image on old web:
* https://osu.ppy.sh/beatmapsets/431309
* https://osu.ppy.sh/beatmapsets/438819

[...]

PR #4053, opened 2019-01-02
Fix beatmapset cover thumbnailer image presence check

nekodex:
Was incorrectly ignoring path when checking for the presence of background images.

---


peppy:
Fixed issue mentioned [here](https://github.com/ppy/osu-web/issues/3729#issuecomment-448876856).

PR #4079, opened 2019-01-09
Remove ability to purchase tournament banners from store after tournament ends

### LINKED ISSUE #3778: Tie tournament-related store purchases into supplied dates
So as to avoid situations where some idiot forgets to turn off profile banner purchases once a tournament has actually finished, a custom field should be added to the store db that prevents items from appearing for sale outside of a given date range.

This date range could be pulled from a linked tournament_id table, though an additional field for "final week" will likely need to be added to the tournaments table to support this particular use case.

notbakaneko:
closes #3778

Add generic `available_until` check on the product itself

---


nekodex:
We probably want to add another field to the tournaments table for the banner sale cut-off date instead of using the tournament end-date. The end-date of the tournament is too late (i.e. it's after the tournament is completely over) and there's usually some leeway for when the grand-finals are actually played so a x-days-before-end thing won't work - so probably best to just add an extra field for the time being.

app/Console/Commands/StoreExpireTournamentBanners.php
```diff
@@ -0,0 +1,44 @@
...
+
+class StoreExpireTournamentBanners extends Command
+{
+    protected $signature = 'store:expire-tournament-banners';
+
+    protected $description = 'Disables banner products of tournaments that have ended';
+
+    public function handle()
+    {
+        $tournamentIds = Tournament::from((new Tournament)->tableName(true))->bannerSalesEnded()->select('tournament_id');
```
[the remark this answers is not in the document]
notbakaneko: it's used as a subquery within a query on a different database so the explicit qualification is needed

app/Models/Store/Product.php
```diff
@@ -100,6 +103,12 @@ public function getDescriptionAttribute($value)
         }
     }
 
+    public function isAvailable() : bool
+    {
+        return $this->enabled
+            && $this->available_until === null ? true : $this->available_until->isFuture();
```
nanaya: should be put inside (brackets) otherwise explosions when not enabled and null `available_until`. Or `optional(x)->a() ?? default`.

app/Console/Commands/StoreExpireProducts.php
```diff
@@ -0,0 +1,41 @@
...
+class StoreExpireProducts extends Command
+{
+    protected $signature = 'store:expire-products';
+
+    protected $description = 'Disables products that should no longer be available.';
+
+    public function handle()
+    {
+        $count = Product
+            ::where('enabled', true)
```
nanaya: if wonder if `->where(enabled)->where(function($q){$q->whereIsNull(available)->orWhere(is available)})` works for filtering available items so this isn't needed anymore.

PR #4085, opened 2019-01-10
Return null instead of object with null fields if country is null.

### LINKED ISSUE #3875: User API response returns country with all null fields
See https://github.com/ppy/osu/issues/3701#issuecomment-437594797

notbakaneko:
closes #3875

would be better if `transform` itself could return null instead of array

---

### DIFF
--- a/app/Transformers/UserTransformer.php
+++ b/app/Transformers/UserTransformer.php
@@ -53,14 +53,18 @@ class UserTransformer extends Fractal\TransformerAbstract
     {
         $profileCustomization = $user->profileCustomization();
 
+        $country = $user->country_acronym === null
+            ? null
+            : [
+                'code' => $user->country_acronym,
+                'name' => $user->countryName(),
+            ];
+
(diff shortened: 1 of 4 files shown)


app/Transformers/UserTransformer.php
```diff
@@ -53,14 +53,18 @@ public function transform(User $user)
     {
         $profileCustomization = $user->profileCustomization();
 
+        $country = $user->country_acronym === null
```
nekodex: might want to use `presence()` here because `country_acronym` is not nullable (so table has empty strings)

(also might want to do that for the other null checks in this PR too)

nanaya: There's getter in the model for this attrirbute converting empty strings into null.

nekodex: Oh hrm, I think I might have been on the wrong/stale branch when checking this -- disregard!

PR #4087, opened 2019-01-11
New News Page

nanaya:
Missing:
- [x] query check
- [x] news listing hover effect
- [x] show more link theme (currently blends with background)
- [x] admin buttons


nekodex:
`@2k` :eyes:

PR #4117, opened 2019-01-20
Rename beatmap categories filter to sections

Joehuu:
This was the ~correct~ terminology.

---


peppy:
Not sure "sections" sounds much better to be honest..

Joehuu:
Renaming also in ppy/osu#4101.

After looking through some other places, both are used. Awaiting opinions.

Joehuu:
I'll leave it as no one is complaining atm.

PR #4124, opened 2019-01-24
Update licence header (and remove year)

peppy:


nanaya:
Conflicts fixed and updated using some terrible git surgeries 🔪

PR #4128, opened 2019-01-25
Fix username lookups with ints instead of strings

peppy:
In cases a user profile is hit using `user_id`, there is a potential that php/eloquent will be "smart" and not quote it as a string in the adjusted query, causing a complete index miss.

Fix has been applied server-side.

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -417,7 +417,7 @@ class User extends Model implements AuthenticatableContract
 
         switch ($lookup_type) {
             case 'string':
-                $user = self::where('username', $username_or_id)->orWhere('username_clean', '=', $username_or_id);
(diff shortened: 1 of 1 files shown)


app/Models/User.php
```diff
@@ -417,7 +417,7 @@ public static function lookup($username_or_id, $lookup_type = null, $find_all =
 
         switch ($lookup_type) {
             case 'string':
-                $user = self::where('username', $username_or_id)->orWhere('username_clean', '=', $username_or_id);
+                $user = self::where('username', strval($username_or_id))->orWhere('username_clean', '=', strval($username_or_id));
```
notbakaneko: should use `(string) $x` instead

PR #4138, opened 2019-01-30
Clean out routes

nekodex:
Pre-emptively fixes some issues that were discovered while creating automated browser/integration tests.

- Removes some orphaned routes
- Fixes some unnamed routes
- Unifies parameter naming (makes things consistent and will also be used by aforementioned tests to determine the type of each param during automated testing)

---


routes/web.php
```diff
@@ -245,7 +242,7 @@
 
 // FIXME: someone split this crap up into proper controllers
 Route::group(['as' => 'store.', 'prefix' => 'store'], function () {
-    Route::get('/', 'StoreController@getIndex');
+    Route::get('/', 'StoreController@getIndex')->name('index');
```
notbakaneko: this can probably be replaced by a redirect to `listing`

PR #4145, opened 2019-02-04
Don't show news posts from the future

cl8n:
osu! wiki wants to be able to merge news posts to the repo before they go live. is this all that is needed?


app/Models/NewsPost.php
```diff
@@ -65,7 +65,7 @@ public static function lookupAndSync($slug)
 
         $post->sync();
 
-        if ($post->page !== null) {
+        if ($post->page !== null && $post->published_at !== null && $post->published_at <= Carbon::now()) {
```
nanaya: use `->isPast()`

app/Models/NewsPost.php
```diff
@@ -77,7 +77,7 @@ public static function pageVersion()
 
     public static function search($params)
     {
-        $query = static::whereNotNull('published_at');
+        $query = static::whereNotNull('published_at')->where('published_at', '<=', Carbon::now());
```
nanaya: Can use a scope for this and `scopeDefault`.

cl8n: something like "scopePublished" and then scopedefault is that + orderby?

nanaya: yeah

PR #4152, opened 2019-02-04
Leave navigation tab bar on profile page visible on mobile view

### LINKED ISSUE #2033: Navigation of profile buttons disappear with narrow browser size
Not fun for mobile users that don't want to scroll through the whole bee movie on [my userpage](https://osu.ppy.sh/users/4236057)
[mobile](https://user-images.githubusercontent.com/30382015/34152110-6dd21df0-e4e8-11e7-8b94-a3386146b38f.PNG)
[desktop](https://user-images.githubusercontent.com/30382015/34151878-aed52f96-e4e7-11e7-830f-680fc6a5bc94.png)

notbakaneko:
closes #2033

Seems to...work

---


nekodex:
I'm not sure if we should be doing this, the design breaks when the line is forced to wrap and the header+bar ends up taking up a quarter of the screen on smaller devices...

e.g.:
simulator screen shot - iphone se - 2019-02-14 at 17 31 28

@arflyte suggests changing it to a swipeable tab bar instead, with the currently selected section being centred

PR #4154, opened 2019-02-04
Fix score badge in recent activity list

### LINKED ISSUE #3956: Grade icons in Recent section are barely discernible
wat

You can guess the grade by colour but S vs S+ is impossible to see. Even in "Recent plays" or "Ranks" with bigger icons S vs S+ is hard to see.
Some colours like green or yellow make it a bit hard to see the white text on it.

nanaya:
Resolves #3956.


nekodex:
The text in the SS+ button looks incorrectly aligned/cropped due to the lack of padding on the left:


Furthermore, the text in the icons look blurry because the characters have both incorrect (vertical) sub-pixel alignment and the images are being scaled down to in that list.

@arflyte said he'd take another look at it.

nanaya:
Updated to look like this instead

nekodex:
That SS+ still looks like it was accidentally cropped 😒

nekodex:
After some internal discussion, we've come up with an idea on how to better display grades/ranks across the game. @arflyte will work on exploring this and in the meantime this PR will be closed (as it's no longer relevant).

PR #4162, opened 2019-02-05
Highlight linked post

### LINKED ISSUE #2369: It can be unclear which post is being specifically linked to
For forum posts.
i.e.:
https://osu.ppy.sh/community/forums/posts/6472195
it's difficult to tell which post this link is directly taking to you. It should be highlighted in some fashion..

notbakaneko:
Add a border around the post that is linked or jumped to.

closes #2369

---

### DIFF
--- a/resources/assets/coffee/_classes/forum.coffee
+++ b/resources/assets/coffee/_classes/forum.coffee
@@ -189,10 +189,9 @@ class @Forum
     @highlightPost post
 
 
-  highlightPost: (post) =>
-    @highlightedPost?.classList.remove 'js-forum-post--highlighted'
-    @highlightedPost = post
-    @highlightedPost?.classList.add 'js-forum-post--highlighted'
+  highlightPost: (post) ->
+    $('.js-forum-post--highlighted').removeClass('js-forum-post--highlighted')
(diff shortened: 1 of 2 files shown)


resources/assets/less/bem/forum-post.less
```diff
@@ -42,6 +42,11 @@
     opacity: 0.5;
   }
 
+  .js-forum-post--highlighted & {
+    border: 3px solid @blue-darker;
+    margin: -3px -3px 2px -3px; // offset so content doesn't get squashed or shift around.
```
nanaya: do you mean `outline` (instead of border).

resources/assets/coffee/_classes/forum.coffee
```diff
@@ -185,6 +186,13 @@ class @Forum
 
     # using jquery smooth scrollTo will cause unwanted events to trigger on the way down.
     window.scrollTo window.pageXOffset, postTop
+    @highlightPost post
+
+
+  highlightPost: (post) =>
+    @highlightedPost?.classList.remove 'js-forum-post--highlighted'
```
nanaya: looks like it won't work quite right on navigation. Maybe just select all with that class instead?

PR #4172, opened 2019-02-08
Generate notification on some actions

nanaya:
Stuff event to its own table and broadcast it to redis instance at `REDIS_HOST_BROADCAST`.


nanaya:
This needs some cleaning and splitting.

PR #4173, opened 2019-02-08
Add custom renderer for markdown indexing

notbakaneko:
Generates the document used for indexing wiki markdown with a different processor. 
This allows greater control of the content used for indexing so we can do more customizations on the content for indexing, like selectively removing sections or increasing priority based on context, or collecting metadata and indexing it as a separate field.

The current output is not ideal (too many newlines for blocks), but can be tweaked later.

run `php artisan es:index-wiki` to re-index the wiki pages that are currently indexed.

---


app/Libraries/Markdown/ParsesHeader.php
```diff
@@ -0,0 +1,47 @@
...
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace App\Libraries\Markdown;
+
+use Symfony\Component\Yaml\Exception\ParseException as YamlParseException;
+use Symfony\Component\Yaml\Yaml;
+
+trait ParsesHeader
```
nanaya: not needed anymore?

app/Console/Commands/EsIndexWiki.php
```diff
@@ -0,0 +1,65 @@
...
+            $search = static::newBaseSearch()->searchAfter(array_values($cursor));
+            $response = $search->response();
+
+            foreach ($response as $hit) {
+                $page = new Page(null, null, $hit->source());
+                if ($page->page() === null) {
+                    continue;
+                }
+
+                (new EsIndexDocument($page))->handle();
```
nanaya: explodes on removed old page.

PR #4174, opened 2019-02-08
Add Browser Tests

nekodex:
This adds Laravel Dusk to perform browser testing for the site (running in parallel with the current phpunit tests on TravisCI to not increase test length). 

This goal of this sanity check is to hopefully eliminate the instances where broken JS makes it to production resulting in random parts of the site completely breaking (and/or not rendering at all...) without us realising for hours/days and/or until people complain to us about it.

Key Points:
- The Login and Logout flows are explicitly tested, in addition to all GET routes being programatically checked.
- The route definitions are used to make educated guesses as to what parameters should be used when testing the various routes.
- Pages are checked to ensure they don't 40x, 50x or throw any JavaScript errors. 
- The tests also ensures that admin paths are correctly restricted for non-admin users.
- Pages that should trigger verification are checked to ensure they correctly trigger verification (and those that shouldn't... don't). The verification flow itself is tested as part of this.

Possible improvements (for another day):
- Move route listing into a dataProvider to break up tests into smaller chunks
- Be smarter about scaffolding, ideally do so dynamically instead of having to manually define (investigate if possible using reflection?)
- Might be worth moving some functions from SanityTest into the base test case to use as helpers

I'm creating this PR a bit early so people can start reviewing it, there's still a few things left to fix before this can be merged:
- [x] fix broken `/status` page (will be in a separate PR)
- [x] check if `beatmapsets.discussion` should be failing or not
- [x] add factories for topic polls so relevant endpoints stop failing

---

This adds Laravel Dusk to perform automated browser testing for the site.


tests/Browser/SanityTest.php
```diff
@@ -0,0 +1,438 @@
...
+            $this->output("\n  /{$route->uri} (".(presence($route->getName()) ?? '???').')');
+
+            if (!present($route->getName())) {
+                $this->output(" \e[30;1m[SKIPPED]\e[0m");
+                $this->skipped++;
+                continue;
+            }
+
+            foreach ($bypass as $prefix) {
+                if (starts_with($route->uri, $prefix)) {
```
nanaya: `starts_with` accepts array so can just pass the array and no looping needed.

PR #4180, opened 2019-02-14
Fix group factory

nekodex:
Prevents the case where duplicate random numbers will result in primary key conflicts

---

### DIFF
--- a/database/factories/GroupFactory.php
+++ b/database/factories/GroupFactory.php
@@ -1,11 +1,11 @@
 <?php
 
+use App\Models\Group;
+
 $factory->define(App\Models\Group::class, function (Faker\Generator $faker) {
     return [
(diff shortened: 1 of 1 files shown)


database/factories/GroupFactory.php
```diff
@@ -1,11 +1,11 @@
...
+use App\Models\Group;
+
 $factory->define(App\Models\Group::class, function (Faker\Generator $faker) {
     return [
-        'group_id' => function () use ($faker) {
-            // try to avoid accidentally granting permissions based on hardcoded group ids
-            return $faker->numberBetween(40, 100);
-        },
+        // try to avoid accidentally granting permissions based on hardcoded group ids
+        'group_id' => max(Group::max('group_id'), 40) + 1,
```
nanaya: maybe in a function?

PR #4183, opened 2019-02-15
Add Shopify integration

notbakaneko:
Adds shopify integration

Requires the Shopify ProductVariant ids to be set on the products that should be linked to Shopify.
The javascript involved is loaded separately as the SDK included turns out to be fairly fat. The changes involved mean that some of what were previously `a` tags are now `button`s

Carts that have Shopify and non-Shopify items cannot be checked out.
Orders that are fulfilled through Shopify will redirect to Shopify for payment and invoices.

TODO:
- needs better handling when things go wrong.
- update with more of the callbacks from Shopify.

---


webpack.mix.js
```diff
@@ -22,8 +22,24 @@ const path = require('path');
 const webpack = require('webpack');
 const SentryPlugin = require('webpack-sentry-plugin');
 const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin');
+
+// might have already been called when mix starts? Won't override anything.
```
nanaya: maybe just

```
dotenv.config({path: `${__dirname}/.env`});
```

notbakaneko: yarn or webpack seems smart enough to run with the correct cwd, and doesn't look like we need to run this anymore since it's called by mix right when it starts, so it doesn't actually do anything anymore 🥉

resources/assets/lib/shopify-gid.ts
```diff
@@ -0,0 +1,33 @@
...
+export function encodeShopifyId(str: string) {
+  return btoa(str);
+}
+
+export function toShopifyProductId(id: string) {
+  return encodeShopifyId(`gid://shopify/Product/${id}`);
+}
+
+export function toShopifyVariantId(id: string) {
+  return encodeShopifyId(`gid://shopify/ProductVariant/${id}`);
```
nanaya:
```
number id => id (id of the thing)
base64 gid => gid (thing passed to api)
gid => NULL (graphql thing)
```

app/Libraries/OrderCheckout.php
```diff
@@ -38,10 +39,18 @@ class OrderCheckout
...
+    private $shopifyId;
+
+    public function __construct(Order $order, ?string $provider = null, ?string $shopifyId = null)
     {
+        if ($provider === Order::PROVIDER_SHOPIFY && $shopifyId === null) {
+            throw new InvariantException('shopify provider requires a checkout id.');
+        }
+
         $this->order = $order;
         $this->provider = $provider;
```
nanaya: should check it's valid provider listed in `allowedCheckoutTypes` (er, `allowedCheckoutProviders`?).

resources/views/store/orders/index.blade.php
```diff
@@ -59,8 +59,21 @@
                                 <span class="store-order__item-quantity">x{{ $item->quantity }}</span>
                         @endforeach
                     </ul>
-                    @if ($order->hasInvoice())
-                        <a class="store-order__link" href="{{ route('store.invoice.show', $order) }}">{{ trans('store.order.invoice') }}</a>
+                    @if ($order->isShopify())
+                        <button
+                            class="js-store-resume-checkout btn-osu-big"
+                            data-checkout-id="{{ $order->getShopifyCheckoutId() }}"
```
nanaya: I think it's fine calling `getTransactionId` directly.

resources/assets/lib/store.ts
```diff
@@ -0,0 +1,167 @@
...
+    let isValid = true;
+
+    const lineItems: LineItem[] = [];
+    $('.js-store-order-item').each((_, element) => {
+      const id = osu.presence(element.dataset.shopifyId);
+      if (id == null) {
+        isValid = false;
+      }
+
+      if (id != null) {
```
nanaya: `else` of above...?

resources/assets/lib/store.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  async beginCheckout(event: Event) {
+    if (event.target == null) { return; }
+
+    const orderId = osu.presence((event.target as HTMLElement).dataset.orderId);
+    if (orderId == null) {
+      throw new Error('orderId is missing');
+    }
+
+    const { isValid, lineItems } = this.collectShopifyItems();
```
nanaya: looks like belongs to `beginShopifyCheckout` (up to the `beginShopifyCheckout`, with confusing `lineItems` name which are actually Shopify line items) and add something like `data-provider` to the button.

Or could even its own class.

resources/assets/lib/store.ts
```diff
@@ -0,0 +1,167 @@
...
+      LoadingOverlay.hide();
+    }
+  }
+
+  async resumeCheckout(event: Event) {
+    if (event.target == null) { return; }
+
+    const target = event.target as HTMLElement;
+    const checkoutId = osu.presence(target.dataset.checkoutId);
+    if (checkoutId == null) {
```
nanaya: same as above, check provider name directly or in this case maybe call it `shopifyCheckoutId`?

...or transaction id.

Not sure if it makes sense having existence of checkout id means it's shopify checkout.

app/Models/Store/Product.php
```diff
@@ -201,6 +202,21 @@ public function scopeHasShipping($query)
...
+     * Returns the Shopify product variant GraphQL gid for this Product, null if it is not a Shopify item.
+     *
+     * @return string|null
+     */
+    public function getShopifyVariantGid() : ?string
+    {
+        return $this->isShopify() ? "gid://shopify/ProductVariant/{$this->shopify_id}" : null;
+    }
+
+    public function isShopify() : bool
```
nanaya: faking stock attribute value if shopify item might work...? :D (re: having wrong stock number)

resources/assets/lib/store.ts
```diff
@@ -0,0 +1,167 @@
...
+        provider: 'shopify',
+        shopifyId: checkout.id,
+      };
+
+      await osu.promisify($.post(laroute.route('store.checkout.store'), params));
+
+      window.location = checkout.webUrl;
+    } catch (error) {
+      // either error from Shopify or updating the order state failed.
+      // TODO: separate the handling of errors.
```
nanaya: more like needs fixing to show verification popup when user isn't yet 💃

app/Libraries/OrderCheckout.php
```diff
@@ -38,10 +39,18 @@ class OrderCheckout
...
+        $this->shopifyId = $shopifyId;
```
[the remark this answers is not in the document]
notbakaneko: it's not relevant to anything but shopify, non-shopify checkouts shouldn't be setting anything other than their provider name for reference in `transaction_id` even if something else is provided.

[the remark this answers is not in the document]
notbakaneko: Meanwhile; the shopify part is not relevant to non-shopify checkouts, so maybe a different class? ┐(￣ー￣)┌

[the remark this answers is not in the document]
nanaya: >non-shopify checkouts shouldn't be setting anything other than their provider name 

this? Maybe even put the check inside.

```
if provider in [ones requiring transactionid]
  explode if missing transactionid
  transactionid += ...
```

app/Http/Controllers/Payments/ShopifyController.php
```diff
@@ -0,0 +1,121 @@
...
+
+        // X-Shopify-Hmac-Sha256
+        // X-Shopify-Order-Id
+        // X-Shopify-Shop-Domain
+        // X-Shopify-Test
+        // X-Shopify-Topic
+
+        $orderId = $this->getOrderId();
+        if ($orderId === null) {
+            throw new Exception('missing orderId');
```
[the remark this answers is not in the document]
notbakaneko: This would normally be a result of an error on our end not setting it, so it should probably be a 500

resources/assets/lib/shopify-buy.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module 'shopify-buy' {
```
nanaya: apparently this is a thing https://yarnpkg.com/en/package/@types/shopify-buy

app/Libraries/OrderCheckout.php
```diff
@@ -63,18 +72,26 @@ public function getProvider()
...
+
+            if ($this->allowXsollaPayment()) {
+                $allowed[] = Order::PROVIDER_XSOLLA;
+            }
+
+            return $allowed;
         }
 
-        return $allowed;
+        return ['free'];
```
nanaya: another one using still using string

PR #4219, opened 2019-02-28
Add highlighting to wiki search results

### LINKED ISSUE #1655: Highlight relevant match in search results
It's a bit hard to make out why a search result appears when the match is inside a giant block of text

notbakaneko:
closes #1655


app/Models/Wiki/PageSearchResult.php
```diff
@@ -0,0 +1,63 @@
...
+namespace App\Models\Wiki;
+
+use App\Libraries\Elasticsearch\Hit;
+
+class PageSearchResult extends Page
+{
+    /** @var Hit */
+    private $hit;
+
+    public function __construct($path, $locale, Hit $hit)
```
nanaya: the first two parameter aren't needed.

PR #4231, opened 2019-03-04
Improved accept language detection

### LINKED ISSUE #4194: Brazil default language is wrong
The default language for brazil is set as "Português", thats the portugal version.

The correct one is "Português (Brasil)".

We have the same problem in the client too (wrong language as default)

image

notbakaneko:
closes #4194

seems to work; not sure the `locale` cookie's lifetime should be so long, though...

### DIFF
--- a/app/Http/Controllers/HomeController.php
+++ b/app/Http/Controllers/HomeController.php
@@ -139,7 +139,7 @@ class HomeController extends Controller
 
     public function setLocale()
     {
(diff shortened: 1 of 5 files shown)


Raymonf:
Wouldn't something like this be better than importing an entire package just for its language negotiation stuff?

http://www.thefutureoftheweb.com/blog/use-accept-language-header

notbakaneko:
replaced by #4292

PR #4237, opened 2019-03-05
Wiki main page

### LINKED ISSUE #2364: Implement Main Page for osu!wiki
there was this design somewhere but no idea if it's relevant anymore

LiquidPL:
Closes #2364.

---
Adds a customizable frontpage with a search box to the wiki.


resources/assets/coffee/react/wiki-search/suggestion.coffee
```diff
@@ -0,0 +1,69 @@
...
+
+    pathCleaned = @cleanPath @props.suggestion.path, @props.suggestion.title
+
+    div
+      className: osu.classWithModifiers 'wiki-search-suggestions__suggestion', ['highlighted' if @props.highlighted]
+      position: @props.position
+      onMouseEnter: @onMouseEnter
+      onMouseLeave: @onMouseLeave
+      onClick: @onClick
+      @formatTitle titleSplit, queryString
```
notbakaneko: text after the matching keyword gets truncated; e.g. `Some Tournament 2018` becomes `Some Tournament` when searching for `Tournament`, `Taiko Wars` truncated to `Taiko`, etc.
Use Elasticsearch to get the highlighting

app/Models/Wiki/MainPage.php
```diff
@@ -0,0 +1,132 @@
...
+    public function getContent(bool $force = false)
+    {
+        try {
+            $sections = OsuWiki::fetchContent('wiki/'.$this->pagePath());
+        } catch (GitHubNotFoundException $e) {
+            $sections = null;
+        }
+
+        if (present($sections)) {
+            $sections = $this->parseArray(Yaml::parse($sections));
```
notbakaneko: This can just go into the `try` block

app/Libraries/Search/WikiSearch.php
```diff
@@ -98,10 +102,19 @@ public function getQuery()
...
                 ]]);
+            }
         }
 
         return (new BoolQuery)
             ->must($langQuery)
-            ->must($matchQuery);
+            ->must($matchQuery)
+            ->mustNot(['match' => [
+                'queryable' => false,
```
notbakaneko: probably better to base on document type instead

LiquidPL:
It's probably better to start with a fresh PR in this case.

LiquidPL:
...or not, I need to think this through. I'll keep this close until I figure this out.

Issue #4232, opened 2019-03-04
Friends List Page


arflyte:
- [x] Convert user card to React
- [x] New user card
- [x] New listing format
- [x] New page format
- [ ] Extra card formats?

related issues
- #2463
- #1851

---

Here's the new friends page; comes with 4 different views.

The top bar groups the active status. Clicking Prev 50 or Next 50, slides the entire list instead of refreshing it in place.

The list tile shows the player's avatar, country flag, usertype colour (non-supporter blue, supporter pink, dev pink and so on), online status and online status colour.

Everything still requires polish, but the basic layout are there. User card designs are not final.

https://www.figma.com/file/l8dHiOeoeGKwaoJYc6h0I7/Web-Friends?node-id=0%3A1

![friend listing list](https://user-images.githubusercontent.com/3632750/53724334-ff4c5380-3eac-11e9-81ed-790da1397cc5.jpg)

Aergwyn:
Are topics like #3774 in the scope of this new design?

notbakaneko:
is that a new pager style? ಠ_ಠ

nekodex:
Are the user cover images being used here different aspect-ratio to the ones we already have generated? (i.e. Do we need to batch process a new size of user covers?)

arflyte:
The max width is 250px for 1x.

nekodex:
Note that we don't generate alternate user cover versions yet, only for beatmap covers. So this will need to be considered.

Ztumpie905509:
I think there could also be a list of followers. (A button to toggle from Mutuals and Followers ?)

It is quite hard to find who actually just followed me without any notifications.

cl8n:
I think the list view could include a column for notes, that would solve #4398 for groups and #3774 for friends list

cl8n:
the avatar-only view seems sort of useless since avatars alone aren't identifying (people change them all the time and half are just random anime girl of the month)

I think some kind of super small card would be better, like 
![53724334-ff4c5380-3eac-11e9-81ed-790da1397cc5](https://user-images.githubusercontent.com/9954349/57172389-066fdf00-6dd4-11e9-9ea5-c15bb03f8c5f.png) (spacing/sizes are all wrong but the idea is there)

peppy:
Agree that avatar-only is a bit pointless (unless it's a zero effort implementation).

PR #4249, opened 2019-03-11
Convert user card to React

notbakaneko:
Migrate user card implementation to React.
Also removes showing online/offline bar on the card when it's loading.

part 1 of #4232


resources/assets/lib/user-card-tooltip.tsx
```diff
@@ -0,0 +1,65 @@
...
+ *
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import * as React from 'react';
+import { UserCard } from 'user-card';
+
+interface PropsInterface {
+  userId: number | string;
```
nanaya: what's ` | string` doing in there.

resources/assets/lib/user-card-store.tsx
```diff
@@ -0,0 +1,58 @@
...
+  constructor(props: PropsInterface) {
+    super(props);
+
+    this.state = {
+      user: this.props.user,
+    };
+  }
+
+  componentDidMount() {
+    $.subscribe('store-supporter-tag:update-user.user-card-store', this.setUser.bind(this));
```
nanaya: Should use fat arrow like a good chunk of everything else.

resources/views/layout/_global_variables.blade.php
```diff
@@ -26,5 +26,20 @@
 @include ('layout._current_user')
 
 <div id="js-usercard__loading-template" class="hidden">
-    @include('objects._usercard', ['loading' => true])
+    {{-- This content is a placeholder so that qtip has something to fade in while the react component mounts --}}
```
nanaya: Can just be rendered at page load though will slow down things even more.

resources/assets/lib/user-card-tooltip.tsx
```diff
@@ -0,0 +1,65 @@
...
+      }));
+    } catch (error) {
+      return null;
+    }
+  }
+
+  render(): React.ReactNode {
+    if (this.state.user) {
+      return <UserCard user={this.state.user} />;
+    } else {
```
nanaya: Nothing is shown when the card is being loaded in this case.

resources/assets/lib/user-card-tooltip.tsx
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * This component's job is to get the data and bootstrap the actual UserCard component for tooltips.
+ */
+export class UserCardTooltip extends React.PureComponent<PropsInterface, StateInterface> {
+  constructor(props: PropsInterface) {
+    super(props);
+
+    this.state = {};
+
+    this.getUser().then((user) => {
```
naoey: react discourages side effects and `setState` in constructor. using `componentDidMount` is advised for initiating network calls.

also TS probably supports class properties for state initialisation without constructor 🤔

resources/assets/coffee/_classes/store-supporter-tag.coffee
```diff
@@ -106,42 +105,44 @@ class @StoreSupporterTag
...
 
+
   sliderValue: (price) ->
     price * @RESOLUTION
 
+
   updateSearchResult: =>
     if @searching
-      @usercard.innerHTML = $('#js-usercard__loading-template').html()
+      $.publish 'store-supporter-tag:update-user', null
```
nanaya: Can't `@user` be set to `null` when setting `@searching` to `false`? Then the path would be the same between searching and not. Also the `setUserInteraction` won't need any parameter as it can just refer to `@user?.id?` directly.

Or maybe even

```coffee
setUser: (user) =>
  @user = user
  $.publish ...
  ....value = @user?.id
  setUserInteraction()
```

resources/assets/coffee/_classes/user-card.coffee
```diff
@@ -62,37 +62,23 @@ class @UserCard
...
 
+    # react should override the existing content after mounting
+    card = $('#js-usercard__loading-template').children().clone()[0]
+    card.classList.replace 'js-react--user-card', 'js-react--user-card-tooltip'
+    delete card.dataset.reactTurbolinksLoaded
+    card.dataset.userId = userId
+
     options =
+      events:
+        render: ->
```
nanaya: kind of suspect `render: reactTurbolinks.boot` would work.

resources/assets/lib/user-card.tsx
```diff
@@ -0,0 +1,207 @@
...
+    let backgroundLink: React.ReactNode;
+
+    if (this.user.cover && this.user.cover.url) {
+      let backgroundCssClass = 'usercard__background';
+      if (!this.state.backgroundLoaded) {
+        backgroundCssClass += ' usercard__background--loading';
+      }
+
+      background =
+        <React.Fragment>
```
nanaya: I actually forgot that `<>` resolves to `<React.Fragment>` but it's a thing I've been using (the short syntax).

PR #4250, opened 2019-03-12
Add dummy env values for Shopify

notbakaneko:


webpack.mix.js
```diff
@@ -23,6 +23,7 @@ const webpack = require('webpack');
 const SentryPlugin = require('webpack-sentry-plugin');
 const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin');
 
+// pre-condition for DefinePlugin
```
nanaya: Is it because assigning `undefined` doesn't quite work?

Additionally `PAYMENT_SANDBOX` actually has default set somewhere below so it's not actually "required"...?

notbakaneko:
#4255 instead

PR #4253, opened 2019-03-12
Add endpoint to destroy sessions of specified user

nanaya:
- Method: `DELETE` (alternatively `_method: DELETE` form with plain `POST`)
- Path: `/_lio/user-sessions/{user}`
  - `user`: User ID
- Interop key header is needed (as usual)


nanaya:
Currently only doing the bare minimum for this to work. Actual cleanup of this thing shall be done later™.

app/Libraries/Session/Store.php
```diff
@@ -30,38 +30,77 @@ class Store extends \Illuminate\Session\Store
...
     /**
      * Get the redis key prefix for the given user (excluding cache prefix).
      *
      * @return string
      */
-    protected static function keyPrefix(int $userId = null)
+    public static function keyPrefix($userId)
     {
-        $userId = $userId ?? 'guest';
+        return 'sessions:'.$userId ?? 'guest';
```
notbakaneko: need brackets around `$userId ?? 'guest'` or it'll never coalesce

PR #4256, opened 2019-03-12
Rebalance func getUsernameAvailableAt

InvisibleSymbol:
The calculations for the time needed to unlock a username were flawed with the use of `$playCount * 0.75`. This resulted in users with over 50k playcount being literally unable to be unlocked in this century, taking over 102 years.

This pull request strives to help with this by switching to a function that quickly rises to a limit, this being `1095*(1-pow(M_E,-$play_count/11000))`, which tops at 3 years. I also changed the value being added to restricted users to reflect the other changes.

Here you can see a comparison of the old method (red line) and mine (green line), with x being the playcount and y being the days needed.

---

Make time needed to unlock a username more reasonable.


app/Models/User.php
```diff
@@ -395,7 +395,7 @@ public function getUsernameAvailableAt() : Carbon
 
         return $this->user_lastvisit
             ->addDays(static::INACTIVE_DAYS) //base inactivity period for all accounts
-            ->addDays($playCount * 0.75);    //bonus based on playcount
+            ->addDays(intval(1095*(1-pow(M_E,-$play_count/11000))));    //bonus based on playcount
```
samuelhklumpers:
```suggestion
            ->addDays(intval(1095*(1-pow(M_E,-$playCount/11000))));    //bonus based on playcount
```

app/Models/User.php
```diff
@@ -395,7 +395,7 @@ public function getUsernameAvailableAt() : Carbon
 
         return $this->user_lastvisit
             ->addDays(static::INACTIVE_DAYS) //base inactivity period for all accounts
-            ->addDays($playCount * 0.75);    //bonus based on playcount
+            ->addDays(intval(1095*(1-pow(M_E,-$playCount/11000))));    //bonus based on playcount
```
samuelhklumpers:
```suggestion
            ->addDays(intval(1095 * (1 - pow(M_E, -$playCount / 11000))));    //bonus based on playcount
```
Fix style

cl8n:
I think the original is fine though, it's strict for a reason

in your example, 50k playcount usually takes even the most active of players multiple years to collect, which I think should be grounds enough to reserve their own username for good

tybug:
50k playcount is an extreme example - it only takes `4866` playcount to reserve your name for 10 years. That seems a little too trivial to achieve for such a long reservation time.

InvisibleSymbol:
With the current system, my name is already reserved for 18 years, and I'm only a 6 digit... I can't agree this is a reasonable approach.

InvisibleSymbol:
Also, I don't respect the idea of locking usernames for good. Yeah, we all want Nathan on osu to be for ever on osu!, But do you seriously think that's going to be true in 50 years? I think after 3 years, it's reasonable that somebody else gets to use the username

FlyingStefan:
On the other hand it seems just reasonable you cannot take over someone's username that easily. In fact it's not obvious to have something like taking over the username of an inactive user. In the times of pre 2012 you were locked with one single name change, for the reason to change your terrible first username to something better. And the ideas was good. The purpose of this option is NOT to get these names from bigger players but mainly to sneak it from users with very very few playcounts and near no activity, when they played osu! for maybe one hour in total. 

idk to be honest, I think more elements should be considered into consideration when an username can be taken away. For example forum badges, post count and ranked beatmaps are a few things that are essential to consider.

InvisibleSymbol:
With these changes, I don't mean to enable people to change their name to one of a (ex-)famous person, I plan to solve the problem of users taking up names years after they quit, limiting the name choices to new people joining. The perfect algorithm would look online and see if the person had an impact/ is still being remembered. You can have fun writing something like that. 

I agree that it would be nice to look at forum badges/post count and ranked maps, but these things don't magically disappear if you overtake the name (if it's even possible to take the name of someone with a ranked map). We could implement checks for all these values, but how strong are they going to affect the duration of the lock? Even if we find a sweet spot, we would have to update the values every now and then to keep them reasonable.

I am suggesting a quick fix for a not-so-big problem with this pr.

cl8n:
usernames are already unavailable if the user has any badges, has any ranked maps, or is above a certain rank, fwiw

InvisibleSymbol:
well then, that solves one part of the concerns. Thanks for the information

Magnus-Cosmos:
I think adding and if statement or a condition based on playcount for years added for restricted players is also a good idea. Since restricted accounts with <1k playcount are usually multiaccounts of some type, thus there's no reason why their names should be locked for such a long time.

InvisibleSymbol:
The comment from @Magnus-Cosmos is something I would like to consider in this pr. 

One of the things I just noticed is that at `line 389 in User.php`, it sets the `AvailableAt` date to always be x years _from the current date_, basically meaning usernames of restricted user are locked for ever. Though I don't know if this is intentional, I'm properly going to change it so it's 3 years from `user_lastvisit`.

Poyo-SSB:
It is indeed intended that restricted usernames are unavailable indefinitely (to "prevent abuse"). I disagree with this, though—I'd like to see discussion regarding this behavior.

cl8n:
it's intended that they are locked forever. it also applies to users with non-default usergroups

InvisibleSymbol:
With my latest commit (https://github.com/ppy/osu-web/pull/4256/commits/c6a2458732d08159587e05b68a52518c53cb2dc2), I try to implement @Magnus-Cosmos idea. 

- Reserved users are locked forever
- Restricted users are locked for 3 years or the minimum of 180 days if their play count is under 1k.

Poyo-SSB:
Here's my proposal.

normal
restricted

In addition to making the curve shallower for restricted users, a linear bonus is also added for high play counts for long-term casual players.

[Graph](https://www.desmos.com/calculator/lrx21lotj8) (black = normal, purple = restricted): 

graph

A few benchmarks:

- 200 plays (beginner): ~8m (18d restricted)
- 2,000 plays (casual player): ~2y, 6m (~7m restricted)
- 5,000 plays (habitual player): ~4y, 8m (~1y, 2m restricted)
- 15,000 plays (median 5-digit player): ~8y, 1m (~3y, 3m restricted)
- 50,000 plays (median 4-digit player): ~9y, 11m (~7y, 2m restricted)
- 90,000 plays (median 3-digit player): ~11y, 2m (~9y, 5m restricted)
- 160,000 plays (median top-100 player): ~13y, 3m (~11y, 6m restricted)
- 250,000 plays (high-end playcount): ~16y (~13y, 7m restricted)

Note that usernames with a non-default usergroup or a high rank are indefinitely unavailable, so the only circumstance in which >75,000 plays would really be relevant is in the rare case of dedicated long-term casual players.

Again, here's the [graph](https://www.desmos.com/calculator/lrx21lotj8) for experimentation.

InvisibleSymbol:
Okay first off, I find your assumption of play count & user rank relationship to be quiet inaccurate.
In this plot (which is possible thanks to @tybug's Database) we can see the relation of rank (x) and play count (y). (Dataset of around 16k users at <500k rank)



Things we can observe here is:
- Basically nobody has a play count over 150k, even 100k is pretty rare
- 50k is more common than said, going down to even low 5 digits- high 6 digits
- The average seems to be at around 7k play count (With your approach this would be 6 years or 2 years with my approach)

Giving a user with 2k plays 2 years and 6 months seems way to high in my opinion, since this is easily achievable with little experience in the game. We're talking about a time range big enough for thousands of user to join. Again, this Pull Requests approach would be at around 1 year.

Also, in my next Plot, we can look at the inactive users (by total pp to play count)



Every point on the left side is an inactive user: An SQLite search showed that **there are about 8000 inactive users, which is about 50% of the Database.** 
Looking at only the inactive users, we see that the cut off seems to be at around 40k, with a lot of users quitting after playing less than 20k plays.



Your method would make it so many of these usernames take around 8 years to unlock, which is already getting in the unreasonable territory. Around 3 years seems good enough in my opinion, enough time to log on once to reset the counter if you really want to save your username.

One thing I like about your proposal is the usage of a different function for restricted users. I might implement this in this Pull Request. 

As I really like to use real examples, let's take my profile. I'm a 6 digit, around 1.5k pp and almost 10k play count. I have been playing this game for about 100 days so nothing really special.

- Old method : around 21 years already reserved.
- My method : around 2 years already reserved.
- Your method : around 7 years already reserved.

Quick note: Your profile never disappears (well if you cheat then sure, but normally not). If somebody takes your username, your profile still exists with username_old. 
(Example Profile with name being taken: https://osu.ppy.sh/users/2857358)

If you want to have a look at @tybug's Database, [here](https://kayo.moe/i1s2MXo5.db) is a link

tybug:
Just want to clarify that I created this dataset by picking random digits between 0 and the latest id, retrieving their user statistics through the api, and storing them if their rank is better than 500k (an arbitrary cutoff, as I believed I could extrapolate for lower ranked users quite well for what I wanted). Which means there shouldn't be any bias in the sampling, since users were chosen completely randomly. This dataset wasn't created for this explicit purpose, but I think it serves well.

Poyo-SSB:
My numbers were picked from a cursory glance at (osu!standard) performance rankings and random sampling of users. Having solid statistics is a lot more useful.

[I've adjusted the constants](https://www.desmos.com/calculator/silpb7o4xw) to more closely match your initial implementation and reflect the actual distributions.

tybug:
These graphs seem rather reasonable to me 👍

InvisibleSymbol:
Yep, the adjusted constant look perfect. Gonna commit them to this pr after I translate them into php.

app/Models/User.php
```diff
@@ -384,18 +384,29 @@ public static function checkWhenUsernameAvailable($username) : Carbon
...
         }, 0);
 
+        if ($this->group_id !== 2) {
+            //reserved usernames
+            return Carbon::now()->addYears(10);  //This will always be in the future, which is wanted
+        }
+
+        if ($this->user_type === 1) {
+            //restricted user
+            if ($playCount > 1000) {
```
Poyo-SSB: I think this was a typo, but this conditional probably isn't necessary at all because of the changes made.

InvisibleSymbol: With this I presume you mean we can use one calculation for both normal and restricted user?

InvisibleSymbol: Oh nvm, I got it. I will remove the play count > 1000 statement

InvisibleSymbol:
Due to a copy and paste error, the restricted specific curve was overwritten with the default one, which was fixed in the latest commit 8271950.

Poyo-SSB:
Is there any opposition to this change? I imagine that the previous indefinite limit was not arbitrary.

peppy:
The original was not "flawed", it was intentionally harsh.

Please add comments to the code explaining the ranges and bonuses (a comment saying "this is a bonus" is absolutely useless).

Also this should be adjusted so a playcount of 50000 has 5 years. The rest can probably scale with that.

Poyo-SSB:
[Here](https://www.desmos.com/calculator/zsqbuymfut) are further-adjusted constants to place 50k plays at 5 years for unrestricted players.

InvisibleSymbol:
>Please add comments to the code explaining the ranges and bonuses (a comment saying "this is a bonus" is absolutely useless).

@peppy Can you explain what you mean with "ranges and bonuses"? Should I explain what the formula does? If that's the case @Poyo-SSB would need to help me, since I don't understand it myself...

Poyo-SSB:
So both of the formulae are exponential decay functions with the identity 1-e^{-x}. They have a constant multiplier of 1580, which causes the formulae to flatten out at around 1580 days (~4.3 years). x is then divided by the constant value 5900 which controls how quickly the formulae flatten out (about 40,000 plays). The formula for a restricted user furthermore has a 0.35 multiplier on x, causing it to flatten out more slowly (closer to 80,000 plays). A linear bonus of 8x/5900 is added to the normal formula while the restricted user's formula has a bonus of 6x/5900. This is to reward long-term play. Finally, the normal formula has a constant inactivity period of 180 days added.

InvisibleSymbol:
If the comments are okay, I believe this is ready to merge (I have a more in-depth version [here](https://gist.github.com/InvisibleSymbol/deca44757757b47deff0d5c9aa37b063), but I found it to be to long).

Poyo-SSB:
Looks good to go. 👍

nanaya:
where did the numbers ("~4.3 years", "flattening out" point and its difference between restricted and not, and the linear "bonus") come from?

Poyo-SSB:
4.3 years just happens to be the way that my formula works out where an unrestricted user gets five years for 50k plays as suggested by peppy. It is, indeed, completely arbitrary. An interactive graph of my formulae can be found [here](https://www.desmos.com/calculator/zsqbuymfut).

app/Models/User.php
```diff
@@ -388,14 +388,37 @@ public static function checkWhenUsernameAvailable($username) : Carbon
 
     public function getUsernameAvailableAt() : Carbon
     {
-        if ($this->group_id !== 2 || $this->user_type === 1) {
+        $playCount = array_reduce(array_keys(Beatmap::MODES), function ($result, $mode) {
```
nanaya: Looks like I forgot to mention that this should use `$this->playCount()` instead.

app/Models/User.php
```diff
@@ -388,14 +388,37 @@ public static function checkWhenUsernameAvailable($username) : Carbon
...
+            $minDays = static::INACTIVE_DAYS;
+            $expMod = 1;
+            $linMod = 1;
+        }
+
+        // This is a exponential decay function with the identity 1-e^{-$playCount}.
+        // The constant multiplier of 1580 causes the formula to flatten out at around 1580 days (~4.3 years).
+        // $playCount is then divided by the constant value 5900 causing it to flatten out at about 40,000 plays.
+        // Furthermore, when the user is restricted, the formula has a 0.35 multiplier on $playCount,
+        // causing it to flatten out more slowly and a linear bonus of 0.75 to reward long-term players.
```
nanaya: The linear bonus should be mentioned in previous part instead and only mention the bonus is lower for restricted user here.

PR #4259, opened 2019-03-14
Split web client localization files

### LINKED ISSUE #4251: Split client-side translations into per-locale bundles
The client should only need to download and load the translations for the current and fallback locales instead of everything. 
Even better would be the ability to only load the translations needed.

notbakaneko:
Localization files are now split per-language instead of being all dumped into the same file;
Only the current and fallback locales are loaded now.

closes #4251


extract-locales.js
```diff
@@ -0,0 +1,79 @@
...
+const path = require('path');
+const mkdirp = require('mkdirp');
+
+const buildPath = path.resolve(__dirname, 'resources/assets/build');
+const localesPath = path.resolve(buildPath, 'locales');
+const messagesPath = path.resolve(buildPath, 'messages.json');
+
+function extractLanguages() {
+  console.log('Extracting localizations...')
+  mkdirp.sync(localesPath);
```
nanaya: this should be in `writeTranslations` instead. And maybe followed by removing existing `.js` files as it might be filled with unrelated stuff.

extract-locales.js
```diff
@@ -0,0 +1,79 @@
...
+  return languages;
+}
+
+function getAllMesssages() {
+  const content = fs.readFileSync(messagesPath);
+
+  return JSON.parse(content);
+}
+
+function runLangJs()
```
nanaya: I think something like `generateTranslationJson` is more descriptive.

extract-locales.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+}
+
+runLangJs();
+writeTranslations(extractLanguages());
+
+// copy lang.js
+fs.copyFileSync(
+  path.resolve(__dirname, 'vendor/mariuzzo/laravel-js-localization/lib/lang.min.js'),
+  path.resolve(__dirname, 'resources/assets/js/lang.js')
```
nanaya: in `buildPath` instead?

.gitignore
```diff
@@ -82,8 +81,9 @@ sftp-settings.json
 /bootstrap/cache/
 
 /resources/assets/js/messages.js
```
nanaya: not needed anymore

extract-locales.js
```diff
@@ -0,0 +1,79 @@
...
+function getAllMesssages() {
+  const content = fs.readFileSync(messagesPath);
+
+  return JSON.parse(content);
+}
+
+function runLangJs()
+{
+  const { spawnSync } = require('child_process');
+  spawnSync('php', ['artisan', 'lang:js', '--json', messagesPath], { stdio: 'inherit' });
```
nanaya: looks like docker needs updating to handle this as currently `yarn` container doesn't have `php` and vice versa. It's already not quite working smoothly as one needs to be re-run (`yarn` container, to build correct js) after the other (`php` container, to generate the `messages.js`) but at least it works in the end.

I think the separation should be `php-fpm` and `assets-builder` instead of `php` and `yarn`. Then maybe even make language (and route) js generation as part of webpack.

extract-locales.js
```diff
@@ -0,0 +1,79 @@
...
+
+function getAllMesssages() {
+  const content = fs.readFileSync(messagesPath);
+
+  return JSON.parse(content);
+}
+
+function runLangJs()
+{
+  const { spawnSync } = require('child_process');
```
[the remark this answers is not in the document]
notbakaneko: scoped to function ┐(ツ)┌

extract-locales.js
notbakaneko: Another problem to be solved another time ™️

PR #4279, opened 2019-03-24
Move up wiki GitHub button on mobile

### LINKED ISSUE #4215: GitHub button on the osu! wiki blocks title on mobile
screenshot_20190227-101046

gary-kim:
Closes #4215

### DIFF
--- a/resources/assets/less/bem/osu-page-header.less
+++ b/resources/assets/less/bem/osu-page-header.less
@@ -142,7 +142,7 @@
     align-items: center;
 
     right: 10px;
-    bottom: 10px;
+    bottom: 50px;
 
     @media @desktop {
       right: 0;
--- a/resources/assets/less/bem/osu-page-header.less
+++ b/resources/assets/less/bem/osu-page-header.less
@@ -142,7 +142,7 @@
     align-items: center;
 
     right: 10px;
-    bottom: 50px;
+    bottom: 10px;
 
     @media @desktop {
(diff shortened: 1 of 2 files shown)


gary-kim:
I had not thought of that. Sounds like a good idea.

notbakaneko:
the forum uses the same class for each post as well, so the element style should probably have a wiki-specific `--wiki` modifier as we currently do not want the buttons on the forum post to be vertical

gary-kim:
Sorry, I didn't see that change. This should now modify it in the wiki only.
Let me know if you'd like me to squash the commits.

resources/assets/less/bem/osu-page-header.less
```diff
@@ -301,6 +311,10 @@
     position: absolute;
     bottom: 10px;
 
+    @media @mobile {
```
notbakaneko: This should be scoped only to the wiki's version of the header `.@{top}--wiki`

resources/assets/less/bem/osu-page-header.less
```diff
@@ -156,6 +156,16 @@
     .@{top}:hover & {
       opacity: 1;
     }
+
+    & .forum-post-actions {
```
notbakaneko: this should be defined in `forum-post-actions.less` instead and scoped to just the wiki

gary-kim:
The last commit has the requested changes.

PR #4284, opened 2019-03-26
Change username validation short-circuiting order

notbakaneko:
Does the reserved username validation before the time availability checks.

Fixes the misleading "username will be available in x days" when they're actually not.

Also removes the flaky rank check on inactive users since ranks stop getting updated when they're inactive


peppy:
Is it intentional that this is WIP?

PR #4287, opened 2019-03-29
Add notification system

### LINKED ISSUE #1168: Notifications System

nanaya:
Take 2. Resolves #1168.

Todo:
- [x] ~~Figure out if there's less redundant way to generate the notifications~~*
- [x] Maybe merge `MarkNotificationsRead` jobs
- [x] Looks like the notifications server will throw error and explode if disconnection and ping happen at the same time
- [x] ~~Figure out better way to handle the hardcoded link to PM (old site)~~*
- [x] Better notification server logging
- [x] ~~Docker?~~*
- [x] Disconnect websocket on logout
- [x] ~~Empty~~ Hide notifications on logout
- [x] Ignore event from self

\* maybe later


resources/assets/lib/globals.d.ts
```diff
@@ -55,10 +55,13 @@ interface OsuCommon {
...
   presence: (str?: string | null) => string | null;
   promisify: (xhr: JQueryXHR) => Promise<any>;
-  timeago: (time: string) => string;
+  timeago: (time?: string) => string;
   trans: (...args: any[]) => string;
-  urlPresence: (url: string) => string;
+  transChoice: (key: string, count: number, replacements?: any, locale?: string) => string;
+  urlPresence: (url?: string) => string;
   uuid: () => string;
+  formatNumber: (num: number, precision?: number, options?: Intl.NumberFormatOptions, locale?: string) => string;
```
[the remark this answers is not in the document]
nanaya: one is less null than the other.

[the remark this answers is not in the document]
notbakaneko: overloads should be declared as functions

app/Http/Controllers/NotificationsController.php
```diff
@@ -0,0 +1,88 @@
...
+    public function __construct()
+    {
+        parent::__construct();
+
+        $this->middleware('auth');
+    }
+
+    public function index()
+    {
+        $limit = 51;
```
nekodex: constant/config?

nekodex: constant/config?

resources/assets/lib/notification-widget/item.tsx
```diff
@@ -0,0 +1,252 @@
...
+      return '/forum/ucp.php?i=pm&folder=inbox';
+    }
+
+    let route: string = '';
+    let params: any;
+
+    if (this.props.items.length === 1) {
+      switch (item.name) {
+        case 'beatmapset_discussion_post_new':
+          route = 'beatmap-discussions.show';
```
nekodex: I wonder if these URLs should be generated server-side to prevent the game-client having to generate all these too?... Having said that, how will the game client handle clicking on these? I imagine for beatmapset/etc it should probably display in-game and not open the website?

PR #4294, opened 2019-04-01
Add support for searching on Beatmapset title prefix

notbakaneko:
Support prefix search of terms from 3 to 8 characters; if it's not found with 8, it's probably not going to found with more.

Also limits the partial matching of multiple terms to the more relevant fields.

### DIFF
--- a/app/Libraries/Search/BeatmapsetSearch.php
+++ b/app/Libraries/Search/BeatmapsetSearch.php
@@ -54,9 +54,15 @@ class BeatmapsetSearch extends RecordSearch
 
         if (present($this->params->queryString)) {
             $terms = explode(' ', $this->params->queryString);
-            // results must contain at least one of the terms and boosted by containing all of them.
-            $query->must(QueryHelper::queryString($this->params->queryString, $partialMatchFields, 'or', 1 / count($terms)));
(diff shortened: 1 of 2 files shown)


app/Libraries/Search/BeatmapsetSearch.php
```diff
@@ -48,12 +48,14 @@ public function __construct(?BeatmapsetSearchParams $params = null)
      */
     public function getQuery()
     {
+        static $partialMatchFields = ['artist', 'artist.*', 'artist_unicode', 'creator', 'description^0.5', 'title', 'title.raw', 'title.*', 'title_unicode', 'tags^0.5'];
```
nanaya: missing at least difficulty names (and maybe other fields?).

peppy: @notbakaneko we gonna add difficulty here? probably a good idea (`version`)

notbakaneko: it's already handled by the other part of the query_string matcher; `$partialMatchFields` limits the fields combined in elasticsearch's query_string cross_field when doing partial terms matching

PR #4296, opened 2019-04-01
Add option to keep forum poll results hidden until the poll ends

cl8n:
suggested in #osu-loved to help prevent bandwagoning or fraud votes

probably a good idea for any poll where the outcome is important

### DIFF
--- a/app/Http/Controllers/Forum/TopicsController.php
+++ b/app/Http/Controllers/Forum/TopicsController.php
@@ -445,12 +445,12 @@ class TopicsController extends Controller
     private function getPollParams()
     {
         return get_params(request(), 'forum_topic_poll', [
+            'hide_results:bool',
             'length_days:int',
             'max_options:int',
             'options:string_split',
             'title',
             'vote_change:bool',
-            'hide_results:bool',
         ]);
     }
 
--- a/app/Http/Controllers/Forum/TopicsController.php
+++ b/app/Http/Controllers/Forum/TopicsController.php
@@ -450,6 +450,7 @@ class TopicsController extends Controller
             'options:string_split',
             'title',
             'vote_change:bool',
+            'hide_results:bool',
         ]);
     }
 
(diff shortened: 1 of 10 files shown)


database/migrations/2019_04_01_202237_add_poll_hide_results_to_topics.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('phpbb_topics', function (Blueprint $table) {
+            $table->boolean('poll_hide_results')->default(0);
```
nanaya: Better placed together with other poll related columns.

app/Libraries/OsuAuthorize.php
```diff
@@ -954,6 +954,25 @@ public function checkForumTopicPollEdit($user, $topic)
...
+
+        if ($user !== null && $topic->posts()->withTrashed()->first()->poster_id === $user->user_id) {
+            return 'ok';
+        }
+
+        if ($topic->pollEnd() === null || $topic->pollEnd()->isPast()) {
+            return 'ok';
+        }
+
+        if (!$topic->poll_hide_results) {
```
nanaya: Do this earlier as it requires least amount of work.

app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -450,6 +450,7 @@ private function getPollParams()
             'options:string_split',
             'title',
             'vote_change:bool',
+            'hide_results:bool',
```
nanaya: can use a bit of sorting

app/Models/Forum/TopicPoll.php
```diff
@@ -107,6 +111,7 @@ public function save()
                 'poll_length' => ($this->params['length_days'] ?? 0) * 3600 * 24,
                 'poll_max_options' => $this->params['max_options'],
                 'poll_vote_change' => $this->params['vote_change'] ?? false,
+                'poll_hide_results' => $this->params['hide_results'] ?? false,
```
nanaya: Defaults can be set in `fill` method.

<sub>And then no one remembered about that until now hence other entries also setting the defaults here.</sub>

PR #4301, opened 2019-04-02
Add storyboard icon to beatmap card

### LINKED ISSUE #2457: No storyboard icon on beatmap card
It seems that the video icon goes with storyboard.

Joehuu:
Resolves #2457. Reworded issue title as it was the same as PR title. Also not sure why that issue had a design tag.

Not tested, but it should work.  Old site uses `far` (regular) icon, but putting `fas` (solid) for now.


resources/assets/coffee/react/_components/beatmapset-panel.coffee
```diff
@@ -102,9 +102,12 @@ class @BeatmapsetPanel extends React.PureComponent
             src: beatmapset.covers.card
           div className: 'beatmapset-panel__image-overlay'
           div className: 'beatmapset-panel__status-container',
-            if beatmapset.video or beatmapset.storyboard
+            if beatmapset.video
               div className: 'beatmapset-panel__video-icon',
                 i className: 'fas fa-film fa-fw'
+            if beatmapset.storyboard
+              div className: 'beatmapset-panel__storyboard-icon',
```
nanaya: This class doesn't exist and thus not properly styled. If anything this should use same class as video (renamed, perhaps).

Joehuu: `__extra-icon`?

nanaya: ...yes

PR #4318, opened 2019-04-09
Let topic covers be disabled per forum

### LINKED ISSUE #4289: Thread covers and their misusage
Right, so I want to share my thoughts regarding thread covers. The feature has been around for a while now and from what I've seen, if anyone ever uploaded one, its random, has nothing to do with the thread's topic and ends up getting removed. While this is wasting server space (we're talking about images 2000px large, MBs in size), we're compressing avatars to JPEGs to save KBs. Something is amiss here. Could potentially limit this feature to supporters or the team only, since I see this viable for events/official stuff, not someone making a thread in the Help subforum, uploading an anime wallpaper as a cover just because they can.

cl8n:
closes #4289, they can still be viewed but only set by mods


app/Libraries/OsuAuthorize.php
```diff
@@ -916,28 +918,54 @@ public function checkForumTopicWatch($user, $topic)
         return 'ok';
     }
 
-    public function checkForumTopicCoverEdit($user, $cover)
+    /**
+     * @param  User $user
+     * @param  Topic|TopicCover $object
+     * @return string
+     */
+    public function checkForumTopicCoverEdit($user, $object)
```
nanaya: authentication cache key needs to be updated to handle multiple types of object.

PR #4319, opened 2019-04-09
Delete scores and ratings on disqualify

cl8n:
already happens on qualify, but since it doesn't go both ways, some people end up with weird things like permanent #1s due to mappers leaving their maps in pending/grave after they've already been qualified once


app/Http/Controllers/BeatmapDiscussionPostsController.php
```diff
@@ -128,8 +128,7 @@ public function store()
             if ($resetNominations) {
                 $events[] = BeatmapsetEvent::NOMINATION_RESET;
             } else {
-                $disqualify = $discussion->beatmapset->isQualified() &&
-                    priv_check('BeatmapsetDisqualify', $discussion->beatmapset)->can();
+                $disqualify = priv_check('BeatmapsetDisqualify', $discussion->beatmapset)->can();
```
nanaya: although you added check in `disqualify` method, new disqualify event will always be generated because there's no check whether or not it's actually disqualified.

app/Models/Beatmapset.php
```diff
@@ -562,6 +562,22 @@ public function setApproved($state, $user)
...
 
+    public function disqualify($user)
+    {
+        if (!$this->isQualified()) {
+            return false;
+        }
+
+        DB::transaction(function () use ($user) {
+            $this->setApproved('pending', $user);
+            $this->userRatings()->delete();
```
nanaya: Better in `setApproved` method itself so it can be used on both disqualify and qualify.

nanaya:
The branch needs updating.

cl8n:
done

app/Models/Beatmapset.php
```diff
@@ -560,6 +561,28 @@ public function setApproved($state, $user)
...
+    }
+
+    public function disqualify($user, $post)
+    {
+        if (!$this->isQualified()) {
+            return false;
+        }
+
+        DB::transaction(function () use ($user) {
+            BeatmapsetEvent::log(BeatmapsetEvent::DISQUALIFY, $user, $post)->saveOrExplode();
```
nanaya: `$post` isn't defined here.

PR #4323, opened 2019-04-10
Implement the QAT restructure

cl8n:
- new usergroup for full (not probationary) BNs (they should be added to both group 28 and 32)
- full BNs can disqualify
- GMT can disqualify
- probationary BNs cannot nominate sets with more than one game mode
- probationary BNs cannot qualify sets that have no nominations from full BNs

---


Ephemeralis:
if this does go in as-is, please signal in ops to me/mao/mo so we can time the deploy with a newspost explaining the changes

app/Libraries/OsuAuthorize.php
```diff
@@ -374,6 +374,16 @@ public function checkBeatmapsetNominate($user, $beatmapset)
             return $prefix.'owner';
         }
 
+        if ($user->isProbationaryBN()) {
```
Xinnoh: This should be checking if a user has either the qat or full bn usergroup like line 437.
A full standard bn that is in probation for mania should be allowed to nominate a standard/taiko hybrid.

cl8n: there aren't mode-specific usergroups for BN/probation so for the purpose of the website your example user should just be a full BN

otherwise this works as intended

cl8n:
yeah this should release at the same time as https://github.com/ppy/osu-wiki/pull/2160 and https://github.com/ppy/osu-wiki/pull/2162. also someone has to make the new usergroup & let NAT add people to it (if that stays )

nanaya:
Feels weird having a subtractive permission.

cl8n:
hmm... yeah it does, I wanted to keep all of the BNs in one group together for the site listing, so this is how it turned out. I'll try flipping the probationary group into "full" BNs instead

WalterToro:
With the QAT moving into the NAT, we discussed that the GMT will be in charge of handing inappropriate content on beatmaps, so to make our jobs easier and hot depending of a NAT member, we decided that Global Moderators will have a DQ button that will be used when something inappropriate (song content, background or any other asset) is observed, so that should be added to this PR.

osuMao:
> With the QAT moving into the NAT, we discussed that the GMT will be in charge of handing inappropriate content on beatmaps, so to make our jobs easier and not depending of a NAT member, we decided that Global Moderators will have a DQ button that will be used when something inappropriate (song content, background or any other asset) is observed, so that should be added to this PR.

While it's true that the GMT will handle inappropriate content on beatmaps, it has not been decided that the GMT will receive a disqualification button yet. This has not been part of the proposal this implementation is based on (and possibly needs discussion with the community as the rest of the proposal did).
Also if we were to give a disqualify button to the GMT, I feel like we could just fuse both groups as permissions woud essentially be identical (i.e. disband group 7, move all NATs to the GMT and give them the NAT title so that people could distinguish them)

cl8n:
I'm still lost as to why the NAT needs to be able to moderate site-wide... their description implies that they're just BNs who manage other BNs (and all of that management is being done on pishi's site, not this one). doesn't make any sense for these people to be, for example, moderating unrelated forums or handing out silences. if they want to moderate the site, they should follow the process to be part of the GMT like any other moderator has

also yeah Walter I think disqualifying would be necessary to moderate beatmaps' content

currently this PR is just making the changes described in the proposal document, but if it were up to me
- NAT doesn't need any permissions (just add them to BN+fullBN usergroup), but their group is still useful for the site listing
- GMT should be able to disqualify mapsets

cl8n:
mao said gmt disqualifying is ok, this should be ready for review now

abraker95:
> I'm still lost as to why the NAT needs to be able to moderate site-wide... their description implies that they're just BNs who manage other BNs (and all of that management is being done on pishi's site, not this one). doesn't make any sense for these people to be, for example, moderating unrelated forums or handing out silences. if they want to moderate the site, they should follow the process to be part of the GMT like any other moderator has

I don't why NAT needs side-wide moderation permissions either. Can this be addressed?

TheMoMan:
NAT will be required to moderate how BNs behave in terms of behaviour and that they are following the rules. They will oversee modding discussions so the ability to clean those up will be needed, and silence accordingly for things like kudosu abuse. They will also require the perms to manage the beatmap management forum. Finally, although focus of this is shifting to GMT, I don't see why the NAT cannot also handle cases of inappropriate beatmap submissions.

cl8n:
all of the things you mention are fine, but I'm confused as to why NAT is supposed to have moderation permissions that, as far as I can tell, they don't need, like for unrelated forums, comments sections, in-game chat, possibly more that I'm forgetting. and on the other hand I don't really see why (full) BNs shouldn't be able to do most of the things you list either. and at that point is there any difference between full BNs and NAT, as far as osu-web is concerned?

anyway for this PR I just want to apply the proposal but maybe someone can open a new issue if they think permissions could be done better

notbakaneko:
It's not finalized yet

Joehuu:
Will changing translation strings break crowdin?

notbakaneko:
the changes will be ignored and overriden from crowdin

cl8n:
should be fine if the translations can be synced at the same time as merging this right? I think peppy did that once before

if it's easier to make all of the changes on crowdin I can revert that

notbakaneko:
The ones where the key has change should probably be removed.
For the ones where the key hasn't changed, the text should be updated on crowdin; otherwise they'll be reverted if crowdin hasn't been updated; or just remove them and leave them untranslated.

PR #4336, opened 2019-04-15
Ignore env files

nanaya:
Not sure if should just ignore entire `.env.*` instead :thinking:


peppy:
Or just `.env*` maybe?

PR #4339, opened 2019-04-15
Style notification popup scrollbar

nanaya:


notbakaneko:
Incidentally, dark mode means the scrollbar colours don't match

PR #4359, opened 2019-04-19
Add sentry for frontend

nanaya:


nanaya:
Closed because old API key.

PR #4372, opened 2019-04-23
Update UserCard classList using more compatible functions

notbakaneko:
To reduce error noise on Sentry


notbakaneko:
rolled up into #4373

Issue #4379, opened 2019-04-25
set bookmark with mail in help forum automatically


Lyawi:
some people ask for help, we are trying our best and they disappear forever, probably not used to forums and never finding their thread again. if we comment they could get poked with an email + notif in forum at least. someone new who is not familiar with forums probably wont set the subscription. 

besides this, even if you just commented a thread to help someone it would be nice if its automatically bookmarked too. atm we are used to check the threads all the time if they replied to us, but it could be simpler in the future (getting a notif?)..

E223:
I think your proposal in #4332 would be a much better way of achieving what you want with this proposal. With that, one could "poke" the person that posted the topic, which would in turn notify them. 

I disagree with automatically bookmarking threads in Help. While it might be useful for people that post and then forget about their post, people who comment in these threads _generally_ tend to be slightly more knowledgeable about the site's functionalities, and can opt-in to receiving notifications for replies if they choose to. If anything, this should be a toggleable setting in my opinion.

Lyawi:
hmm, toggleable sounds ok to me, i wanna follow threads i replied to, without setting that bookmark all the time. an email all the time is kinda annoying, but i can live with it for now. if its toggleable, maybe it should affect the whole forum instead of help only then, tbh

peppy:
I think this is a pretty good idea as a forced default for help threads.

Flanster:
#4332 is already plenty as mentioned above, they can get a notification from **quoting their reply** and/or **mentioning their name**. Autobookmarking everything sounds bothersome.

cl8n:
I would say the opposite Flanster, because why would you post a thread if you didn't want to read the replies to it (on any forum, not just Help)? I don't think I've ever posted a thread without bookmarking it immediately after, can you explain why that would be a bothersome feature

Lyawi:
in some cases you dont even quote if its only one question, and especially if youre replying to this thread the first time. lets take https://osu.ppy.sh/community/forums/topics/899476?start=7057436 as a example. i can understand the notifs could bother people that use the forums, but i do too and i would find it helpful. if its toggleable it should be a thing? and only then. just my opinion

Flanster:
> I would say the opposite Flanster, because why would you post a thread if you didn't want to read the replies to it (on any forum, not just Help)? I don't think I've ever posted a thread without bookmarking it immediately after, can you explain why that would be a bothersome feature

Because I do the opposite, I've barely needed to bookmark anything as I can keep track of it.

[...]

PR #4393, opened 2019-04-30
Add option to automatically subscribe to topics when posting

cl8n:
possible solution for #4379

---


peppy:
I still think that this should be forced for the help forum, on top of potentially having this option as well.

Also this thing used to exist with phpbb, and still exists in the database (so you don't need to add a new column) - check `user_notify` I believe.

cl8n:
since the problem seems to be that newer users won't follow up on their threads, I think this option could be set true for everyone at first? or is that too forceful idk

will add the help forum thing if not that instead^

peppy:
The help forum one should bypass the setting 100%. The actual user setting was switched to "off" by default because of email spam, but once we have the ability to discern emails from notificaitons we'll probably set it to "on" for notifications "off" for email by default.

cl8n:
I can't reproduce the error here D:

PR #4395, opened 2019-05-01
Disallow emails differing only by case

### LINKED ISSUE #4394: Changing email from account settings is case sensitive.
The website lets you change your email address to another one that is already in use if you use some capital letters.

For example, if abcde@gmail.com is already in use by an account, it's possible to enter Abcde@gmail.com, so 2 accounts can use the same email.

notbakaneko:
fixes #4394

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -1691,7 +1691,8 @@ class User extends Model implements AuthenticatableContract
                 $this->validationErrors()->add('user_email', '.invalid_email');
             }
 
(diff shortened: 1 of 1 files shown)


nanaya:
isn't the column set to case insensitive collation already...?

peppy:
It's utf8_bin currently. I'm performing a migration to alter it.

Not sure if this change will be required afterwards (nor if it would work due to the tolower only being on one half of the comparison).

notbakaneko:
looking at it again, it's not going to work against existing nonlowercased addresses without the collation change, but changing the collation also means this shouldn't be needed anyway due to the unique constraint

PR #4412, opened 2019-05-07
New friends page design

notbakaneko:
- uses new friends page
- adds a sorter (currently only alphabetical by username and recent online status)
- adds online status filter (other states to be added when we add support for them)
- update group user listing to same design
- changing filtering or sort should update the url

see #4232

Sorting and filtering are client-side.

Known issues:
- group listing doesn't show the description of the group yet
- currently uses css class names from the changelog page for the filter; renaming them and moving them around appears to be a bit messy, so I'm leaving it as a separate task. 
- incremental display of users coming soon™️


resources/assets/lib/sort.tsx
```diff
@@ -0,0 +1,57 @@
...
+      }
+
+      return (
+        <button
+          className={cssClasses}
+          data-value={value}
+          key={value}
+          onClick={this.props.onSortSelected}
+        >
+          {osu.trans(`users.sort.${value}`)}
```
nanaya: suddenly `users` key in a seemingly generic `Sort` component.

resources/assets/lib/user-list.tsx
```diff
@@ -0,0 +1,170 @@
...
+        <p className='update-streams-v2__row update-streams-v2__row--version'>{text}</p>
+      </a>
+    );
+  }
+
+  private get sortedUsers() {
+    const users = this.filteredUsers.slice();
+
+    switch (this.state.sortMode) {
+      case 'username':
```
nanaya: More like `case 'last_visit'`.

resources/assets/lib/user-list.tsx
```diff
@@ -0,0 +1,170 @@
...
+  readonly state: State = {
+    filter: this.filterFromUrl,
+    sortMode: this.sortFromUrl,
+  };
+
+  onSortSelected = (event: React.SyntheticEvent) => {
+    const value = (event.currentTarget as HTMLElement).dataset.value;
+    const url = osu.updateQueryString(null, { user_sort: value });
+
+    Turbolinks.controller.pushHistoryWithLocationAndRestorationIdentifier(url, Turbolinks.uuid());
```
nanaya: Should be `advanceHistory` so it doesn't reload on navigation.

resources/assets/lib/user-list.tsx
```diff
@@ -0,0 +1,170 @@
...
+      </a>
+    );
+  }
+
+  private get sortedUsers() {
+    const users = this.filteredUsers.slice();
+
+    switch (this.state.sortMode) {
+      case 'username':
+        return users.sort((x, y) => moment(y.last_visit || 0).unix() - moment(x.last_visit || 0).unix());
```
nanaya: Or using `Moment#diff`.

resources/assets/lib/user-list.tsx
cl8n: probably switch the two return statements instead, does the same thing but would make clear which one is the default

nekodex:
Pretty sure this section's background is the wrong colour:


If you don't want to change the changelog/`.upsteam-streams` stuff yet, I'd at least look at adding on an additional class/override to set it to the right colour temporarily?

PR #4417, opened 2019-05-08
Additional beatmapset nomination tests

notbakaneko:
additional tests for #4416;

checks limited BNs can't make the qualifying nomination if someone from NAT or BNG has not nominated.

### DIFF
--- a/tests/Models/BeatmapsetTest.php
+++ b/tests/Models/BeatmapsetTest.php
@@ -82,10 +82,7 @@ class BeatmapsetTest extends TestCase
         $this->assertTrue($beatmapset->fresh()->isQualified());
     }
 
-    /**
-     * @doesNotPerformAssertions
-     */
-    public function testLimitedBNGNominatingBNGNominated()
+    public function testLimitedBNGQualifyingNominationBNGNominated()
     {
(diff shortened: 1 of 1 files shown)


tests/Models/BeatmapsetTest.php
```diff
@@ -86,6 +88,61 @@ public function testQualify()
...
+        $user->userGroups()->create(['group_id' => UserGroup::GROUPS['nat']]);
+        $beatmapset->nominate($user);
+
+        $nominator = factory(User::class)->create();
+        $nominator->userGroups()->create(['group_id' => UserGroup::GROUPS['bng_limited']]);
+
+        priv_check_user($nominator, 'BeatmapsetNominate', $beatmapset)->ensureCan();
+    }
+
+    public function testLimitedBNGNominatingLimitedBNGNominated()
```
cl8n: this test would fail if `Beatmapset`'s `requiredNominationCount()` were greater than 2, since the check for fullBN/NAT nominations is only relevant if the next nomination is the final one. likewise the other tests would pass but not with a useful result

not sure if this is actually an issue but the solution would be to generate `$beatmapset->requiredNominationCount() - 1` nominations first before doing that final priv_check

PR #4438, opened 2019-05-10
Remove GMT and NAT ability to move beatmapsets to Loved

cl8n:
not really sure why this existed in the first place. was only used two times, once by Noffy (QAT) because I asked her to when I was afk, and once by Pawsu (GMT) as a joke or something that was reverted

---

### DIFF
--- a/app/Libraries/OsuAuthorize.php
+++ b/app/Libraries/OsuAuthorize.php
@@ -353,7 +353,7 @@ class OsuAuthorize
     {
         $this->ensureLoggedIn($user);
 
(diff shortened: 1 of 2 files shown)


Ephemeralis:
it would be great in the future if you actually told someone that you were removing these permissions before going ahead with it

cue today where a gmt was forwarded a map to unlove and was confronted with a distinct lack of a permission they previously had which abruptly vanished into the ether

cl8n:
nobody could unlove previously, nobody can unlove now

PR #4444, opened 2019-05-13
Add block user and report user options to user card

notbakaneko:
`z-index: unset !important` to force the auto-incrementing `z-index` from qtip to be ignored


resources/assets/lib/user-card.tsx
```diff
@@ -157,17 +159,25 @@ export class UserCard extends React.PureComponent<PropsInterface, StateInterface
...
       <>
-        <a
-          className='simple-menu__item js-login-required--click'
-          href={laroute.route('messages.users.show', { user: this.user.id })}
-          onClick={dismiss}
-        >
-          {osu.trans('users.card.send_message')}
-        </a>
+        {
+          this.canMessage ? (
```
nanaya: just noticed this checks for user blocks but no popup is shown for blocked users anyway

resources/assets/lib/user-card.tsx
```diff
@@ -157,17 +159,25 @@ export class UserCard extends React.PureComponent<PropsInterface, StateInterface
...
+              href={laroute.route('messages.users.show', { user: this.user.id })}
+              onClick={dismiss}
+            >
+              <span className='fas fa-envelope' />
+              {` ${osu.trans('users.card.send_message')}`}
+            </a>
+          ) : null
+        }
+
+        <BlockButton onClick={dismiss} modifiers={['inline']} userId={this.user.id} wrapperClass='simple-menu__item' />
```
nanaya: this and report user below aren't shown for guest, resulting in empty menu.

resources/assets/lib/user-card.tsx
notbakaneko: cards already created for a user still show even after they're blocked

PR #4461, opened 2019-05-17
Add notifications for chat

nekodex:


app/Jobs/BroadcastNotification.php
```diff
@@ -186,6 +187,29 @@ private function onBeatmapsetResetNominations()
...
+    {
+        $channel = Channel::findOrFail($this->object->channel_id);
+        $users = $channel->users()->pluck('user_id');
+
+        $this->receiverIds =
+            array_values(
+                array_filter(
+                    $users->toArray(),
+                    function ($k) {
+                        return $k !== $this->source->user_id;
```
nanaya: This is already done somewhere else (when creating per-user notification and before sending it to client).

app/Models/Chat/Channel.php
```diff
@@ -166,14 +172,24 @@ public function addUser(User $user)
...
     {
-        UserChannel::where([
+        $userChannel = UserChannel::where([
             'channel_id' => $this->channel_id,
             'user_id' => $user->user_id,
-        ])->delete();
+        ]);
+
+        if ($this->isPM()) {
+            event(new UserSubscriptionChangeEvent('remove', $user, $userChannel->channel));
```
nanaya: `$userChannel` is still a builder. Call `first()` first?

resources/lang/en/notifications.php
```diff
@@ -51,6 +51,17 @@
...
         ],
 
+        'channel' => [
+            '_' => 'Chat',
+
+            'channel' => [
+                '_' => 'New message',
+                'channel_message' => ':username says ":title"',
+                'channel_message_compact' => ':title',
+                'channel_message_group' => 'from :username',
```
nanaya: this looks like specific for pm instead of channel message? (also in some other places like url generator)

PR #4462, opened 2019-05-17
Update score badges to latest design

### LINKED ISSUE #3956: Grade icons in Recent section are barely discernible

### LINKED ISSUE #4420: F as score-rank icon if you have rank D ingame
I recently noticed that the score-rank for rank D maps doesn't match with the rank that's displayed on the website. I don't know if this is intentional or not and I also couldn't find any information about it.

wrong-score-rank-icon

I tracked down the potential issue and found this to probably be the culprit of the mismatched score icon:

https://github.com/ppy/osu-web/blob/master/resources/assets/less/bem/score-rank-v2.less#L44

If this is indeed a bug I'd like to submit a pull request in which I change the score to display the D now and include the F icon for the css `class  score-rank-v2--F` like the others above.

nanaya:
Remove old badges and adjust things. Including return of the D (fixes #4420). Also probably fixes #3956.


Aergwyn:
Does also fix #3956?

nanaya:
Yeah. Hopefully.

Issue #435, opened 2016-03-20
Join date on forum posts is aligned wrong


peppy:
Shouldn't be bottom anchored.

![](https://puu.sh/nMWVx/3029801280.png)

nanaya:
this should go to @arflyte since that's how he designed it ¯\\\_(ツ)_/¯

suzumotoHikari:
Note: You can ignore this reply if this is against on the flyte's design.
Also, i am newbie on this kind of development especially coding less. :)

On the file `forum-post.less` on the line 284, on `&__info-panel-extra`, i just changed the value of `justify-content` to `initial`. Rather, removing the `justify-content` might work as well.

Also, you can remove it. As a probe, i had an image below showing the output.
Before i changed the `justify-content` value to `initial`: http://puu.sh/pEowt/3e41469534.PNG
I changed the `justify-content` value to `initial`: http://puu.sh/pEoyv/961b58c0c9.PNG

PR #4470, opened 2019-05-18
Align join date to top on forum posts

Carlos-T:
Fixes [#435](https://github.com/ppy/osu-web/issues/435)
So no space between flag and join date.
image


cl8n:
pretty sure it's intended to be bottom-anchored

Carlos-T:
What's the issue point then? a max-height? :) a common behaviour should be defined, no matter how long is the post.

peppy:
@arflyte

arflyte:
Updating the poster detail area design

PR #4481, opened 2019-05-21
Add method to rank beatmapset

nanaya:
I figured this should come first before migrating the actual job here.


app/Models/Beatmapset.php
```diff
@@ -700,6 +702,30 @@ public function love(User $user)
...
+    public function rank()
+    {
+        if (!$this->isQualified()) {
+            return false;
+        }
+
+        DB::transaction(function () {
+            $this->events()->create(['type' => BeatmapsetEvent::RANK]);
+
+            $this->setApproved('ranked', null);
```
cl8n: should also delete scores (maybe user ratings too? not sure what currently happens to those on rank)

nanaya: already handled in setApproved method.

PR #4494, opened 2019-05-24
Add ability to close PMs

nekodex:
Leaving/closing a PM channel will hide it until a new message arrives in the channel.

Messaging the same user again will show the past chat history (if present and not culled), although right now the experience of re-opening a chat is a bit awkward and requires sending a message before the history is reloaded/shown (will fix later™).

### DIFF
--- a/resources/assets/lib/chat/chat-api.ts
+++ b/resources/assets/lib/chat/chat-api.ts
@@ -68,6 +68,21 @@ export default class ChatAPI {
     });
   }
 
+  partChannel(channelId: number, userId: number) {
+    return new Promise((resolve, reject) => {
+      $.ajax(laroute.route('chat.channels.part', {
+        channel: channelId,
+        user: userId,
+      }), {
+        method: 'DELETE',
+      }).done((response) => {
+        resolve(response);
+      }).fail((error) => {
+        reject(error);
+      });
+    });
+  }
+
   sendMessage(channelId: number, message: string): Promise<ApiResponses.SendMessageJSON> {
     return new Promise((resolve, reject) => {
       $.post(laroute.route('chat.channels.messages.store', {channel: channelId}), {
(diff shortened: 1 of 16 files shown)


resources/assets/lib/stores/channel-store.ts
```diff
@@ -107,6 +114,20 @@ export default class ChannelStore implements DispatchListener {
     });
   }
 
+  @computed
+  get channelList(): Channel[] {
+    let channels: Channel[] = [];
+    channels = channels.concat(this.nonPmChannels);
+    channels = channels.concat(this.pmChannels);
+
+    return channels;
```
nanaya: Just one line

    return [...this.a, ...this.b]

?

resources/assets/lib/chat/conversation-list-item.tsx
```diff
@@ -33,33 +33,45 @@ export default class ConversationListItem extends React.Component<any, {}> {
...
-      <a href='#' className={className} onClick={this.switch}>
-        <div className='chat-conversation-list-item__unread-indicator' />
-        <img className='chat-conversation-list-item__avatar' src={conversation.icon} />
-        <div className='chat-conversation-list-item__name'>{conversation.name}</div>
-        <div className='chat-conversation-list-item__chevron'>
-          <i className='fas fa-chevron-right' />
-        </div>
-      </a>
+      <div className={className}>
+        <a href='#' className={`${baseClassName}__close-button`} onClick={this.part}>
```
nanaya: is there reason this isn't button? (so `preventDefault` isn't needed)

resources/assets/lib/chat/chat-api.ts
```diff
@@ -68,6 +68,21 @@ export default class ChatAPI {
     });
   }
 
+  partChannel(channelId: number, userId: number) {
+    return new Promise((resolve, reject) => {
```
nanaya: `$.ajax` has `.then` and `.catch` since 3.0 so the wrapping isn't needed anymore.

nekodex: noted

PR #4495, opened 2019-05-24
Ignore not found errors from audio player

### LINKED ISSUE #4493: Properly handle missing/invalid audio files
Should ignore such errors and not stuck in play state.

Example: https://osu.ppy.sh/beatmapsets/42861

```
NotSupportedError: Failed to load because no supported source was found.
```

notbakaneko:
closes #4493


nanaya:
Can use better handling of the error instead of stuck at play state.

PR #4505, opened 2019-05-27
Use same translation and order for home navigation

### LINKED ISSUE #4484: Capitalisation, ordering and styles of tabs on dashboard changes
image

image

nanaya:
Capitalisation style uses old version for now. Fixes #4484.

### DIFF
--- a/resources/views/home/_user_header_nav.blade.php
+++ b/resources/views/home/_user_header_nav.blade.php
@@ -46,20 +46,6 @@
 
     <li class="page-mode__item">
         <a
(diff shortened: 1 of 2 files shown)


resources/views/home/_user_header_nav.blade.php
```diff
@@ -46,41 +46,41 @@ class="
 
     <li class="page-mode__item">
         <a
-            href="{{ route('account.edit') }}"
```
nanaya: nice diff.

PR #4512, opened 2019-05-27
Less gradients score badge, and return of silver

nanaya:


peppy:
Update the ?reqhash please

Issue #4516, opened 2019-05-28
Tournament banners blurry


peppy:
normal:
![image](https://user-images.githubusercontent.com/191335/58446867-5e80d380-813d-11e9-83ec-1d6b080273e1.png)
hovered:
![image](https://user-images.githubusercontent.com/191335/58446871-60e32d80-813d-11e9-97c3-ef5b7ed25d31.png)

both are shocking.

[example user](https://osu.ppy.sh/users/9314998)

nekodex:
Partially fixed by #4564 (for other people's profiles). A proper fix is forthcoming (that involves moving the spinner elsewhere)

peppy:
Was the proper fix implemented already?

notbakaneko:
It's blurry if you use safari and view it on your own profile; the spinner needs to be moved out of the header background to anywhere else that won't overlap with the banner. An option that was discussed was to move the spinner to the image selection itself where the checkmarks currently are

peppy:
I guess that's a no, then. Should we leave this in the June milestone? Is it more effort than it's worth to fix?

peppy:
Moving out of milestone as it is now limited in scope and only affects safari.

nanaya:
Can use an update on this :thinking: it probably has been fixed thanks to the (not-so) recent header change.

peppy:
Tested on retina and non, looks fixed 👍

PR #4520, opened 2019-05-28
Fix tournament banner scaling

nanaya:
GPU accelerating it doesn't quite work. May fix (doesn't for safari) #4516 though need to be checked on another system/browser.


Joehuu:
Isn’t fix reference supposed to be #4516?

nanaya:
Indeed

PR #4522, opened 2019-05-28
Dark comments page

nanaya:
I figured this should be dealt with first. Haven't added any snippet for permalinks hence still moderator-only.


Flanster:
Meanwhile forums are with a dark background and white post body shining so bright you need sunglasses. Good priorities right there.

peppy:
@Flanster updating forums is quite a large task, and usage is relatively low. We are prioritising this page because it needs some changes to allow users to receive notifications for replies to comments, which is quite a high priority IMO.

PR #4523, opened 2019-05-29
Refactor beatmap listing page

### LINKED ISSUE #4314: Default beatmap search should probably include loved category
Feels a bit weird that it doesn't, as it doesn't really mention anywhere that it's being filtered down and looks like it's just failing to match.

image

image

notbakaneko:
- fix input box, filter options and search results going out of sync
- fix back/forward navigation showing wrong results
- fix url changing on next page when navigating away
- url history now restores properly
- search state/results moved to mobx and handled independently of UI components, so things like quick search can reuse any existing results
- category query parameter switched to `string` instead of `number`
- default search includes maps in `loved` state; closes #4314


resources/assets/lib/beatmap-search-filters.ts
```diff
@@ -0,0 +1,134 @@
...
+    return this.selectedValue('sort');
+  }
+
+  @computed
+  get query() {
+    return this.sanitizedQuery;
+  }
+
+  set query(value: string | null) {
+    if (osu.presence(value) === osu.presence(this.sanitizedQuery)) {
```
nanaya: `presence` call for right side is not needed. `value` can probably be processed before doing the comparison. And confirm whether it should be compared `trim`-ed or not (which then assigned to the `trim`-ed value anyway?)

resources/assets/coffee/react/beatmaps/search-panel.coffee
```diff
@@ -235,5 +237,14 @@ export class SearchPanel extends React.PureComponent
       @inputRef.current.focus()
 
 
-  submit: (query) ->
-    $(document).trigger 'beatmap:search:filtered', query: query.trim()
+  mountPortal: (portal, root) ->
+    # clean up any existing element when navigating backwards.
+    existingElement = document.getElementById(portal.id)
+    existingElement?.parentNode?.removeChild existingElement
```
nanaya: Maybe just use `remove()` directly.

resources/assets/lib/beatmaps/beatmap-search.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  private append(key: string, data: SearchResponse) {
+    this.appendBeatmapsets(key, data.beatmapsets);
+    this.cursors.set(key, data.cursor);
+    this.totals.set(key, data.total);
+  }
+
+  private appendBeatmapsets(key: string, data: BeatmapsetJSON[]) {
+    const beatmapsets = this.getOrCreate(key);
+    for (const beatmapset of data) {
```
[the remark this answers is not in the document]
notbakaneko: We're dropping support for Edge <14 since barely any users use it, so we don't need to use le in `for-of` anymore

PR #4525, opened 2019-05-30
Linkify urls in chat

### LINKED ISSUE #3881: Links aren't clickable in Chat

nekodex:
The displayed URLs will have http/https stripped (for display only).

To remain consistent with this, `discussionLinkify` (used by the beatmap discussions) was also updated.

fixes #3881


resources/assets/coffee/osu_common.coffee
```diff
@@ -170,8 +170,8 @@
     el.outerHTML
 
 
-  linkify: (text) ->
-    return text.replace(osu.urlRegex, '<a href="$1" rel="nofollow">$1</a>')
+  linkify: (text, newWindow = false) ->
+    return text.replace(osu.urlRegex, "<a href=\"$1\" rel=\"nofollow\"#{if newWindow then ' target=\"_blank\"' else ''}>$2</a>")
```
nanaya: Should add a `noopener` rel. Also maybe replace the double quotes inside with single quotes (and the escape inside target=_blank isn't needed).

`return` keyword optional.

resources/assets/coffee/_classes/beatmap-discussion-helper.coffee
```diff
@@ -54,6 +54,8 @@ class @BeatmapDiscussionHelper
             # different beatmapset, format: 1234#567
             linkText = "#{targetBeatmapsetDiscussions.beatmapsetId}##{targetBeatmapsetDiscussions.discussionId}"
 
+      linkText = linkText ? displayUrl
```
nanaya: Or just `linkText ?= displayUrl`

resources/assets/lib/models/chat/message.ts
```diff
@@ -38,6 +38,11 @@ export default class Message {
     this.sender = new User(-1); // placeholder user
   }
 
+  @computed
+  get parsedContent(): string {
+    return osu.linkify(this.content, true);
```
nanaya: should escape the content before passing to linkify (see comment on `discussionLinkify` function).

PR #4529, opened 2019-05-31
Add feed for news

nanaya:
In atom format :atom: 

Can use a bit fancier author name.

~~Still needs fixing so it actually work with discord thingy~~. Apparently it's problem on their end.


resources/views/news/index-atom.blade.php
```diff
@@ -0,0 +1,47 @@
...
+    <link rel="self" type="application/atom+xml" href="{{ request()->fullUrl() }}" />
+
+    <title>osu!news</title>
+    <icon>{{ config('osu.static') }}/apple-touch-icon.png</icon>
+
+    <updated>{{ json_time(optional($posts->last())->published_at ?? now()) }}</updated>
+
+    @foreach ($posts as $post)
+        <entry>
+            <id>tag:osu,2019:news/{{ $post->getKey() }}</id>
```
nekodex: shouldn't the tags (here and above) be `tag:osu.ppy.sh` or something more specific?

nekodex:
So uh, I just noticed that `metadata.blade.php` [already has an include for rss](https://github.com/ppy/osu-web/pull/4529/files#diff-ffcfad5e911ddf04bc44b7aa11aeb07aR110)?

PR #4535, opened 2019-06-02
Change the capitalization of permalinks in comments

TPGPL:
Changed the capitalization of permalinks in comments as it isn't fully correct in every language to capitalize the first letter of *every* word.

---

### DIFF
--- a/resources/assets/less/bem/comment.less
+++ b/resources/assets/less/bem/comment.less
@@ -67,7 +67,11 @@
     &--active {
       color: @green;
     }
-
+    
+    #--permalink {
+      text-transform: lowercase;
+    }
+      
     .@{_top}--changelog & {
       .link-blue-light();
     }
--- a/resources/assets/less/bem/comment.less
+++ b/resources/assets/less/bem/comment.less
@@ -68,11 +68,6 @@
       color: @green;
     }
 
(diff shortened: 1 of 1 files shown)


notbakaneko:
Can you provide examples of this?

TPGPL:
@notbakaneko currently permalink is Polish is translated as "odnośnik bezpośredni" and with capitalize it becomes "Odnośnik Bezpośredni" which isnt fully correct because capitalizing first letter of every word doesnt make sense in this language

nekodex:
I'm thinking that all the comment actions should probably just be lower-case instead. Not sure why they were title-cased at all in the first place?


(see comment 1 vs comment 2)

TPGPL:
should be ready now

nekodex:
The other items (reply, edit, etc) are currently capitalised in the translation files, so they'll need to be transformed to lowercase, see:

qcl12345:
I think this also ended up affecting the reply username as well when it probably shouldn't have?

e.g. https://osu.ppy.sh/comments/868955 and https://osu.ppy.sh/comments/868968 
The usernames that these people are replying to both have capital letters in them so they probably shouldn't be changed.

peppy:
Reported as #4565.

PR #4538, opened 2019-06-03
Typescript code quality updates

notbakaneko:
set ~`member-ordering` and~ `member-access` to be the same as osu-notification-server

`member-ordering` now uses an explicit order; getters and setters should follow immediately after fields, but there isn't a tslint rule for those.

rename `PropsInterface` and `StateInterface` to just `Props` and `State`


nekodex:
Might want to double check what is being done here... for example, look at `channel.ts`, all the `@observable` properties being pushed down into the middle of the file... ?

PR #4539, opened 2019-06-03
Allow searching by user and beatmapset id

### LINKED ISSUE #4517: Can't search for users by ID
Works for beatmaps / beatmapsets but not users:

image

notbakaneko:
Scores exact matching id as first result.
Also removes the search length restriction except on forum and wiki text content, as we don't really need that restriction.

closes #4517


resources/views/home/search.blade.php
```diff
@@ -68,7 +68,7 @@ class="search-header__input js-search--input"
                 @else
                     <div class="search-result">
                         <div class="search-result__row search-result__row--notice">
-                            {{ trans('home.search.missing_query', ['n' => config('osu.search.minimum_length')]) }}
+                            {{ trans('home.search.missing_query', ['n' => 1]) }}
```
notbakaneko: just so existing translations don't show `minimum :n characters`

nekodex: Probably worth just renaming the key to force a new translation?

PR #4541, opened 2019-06-04
Change the chat font to a legible one

### LINKED ISSUE #4540: Chat content font is exo (it probably shouldn't be)
All other content fonts are using more legible fonts. Chat feels out of place. @arflyte please confirm intent.

TPGPL:
Closes #4540

### DIFF
--- a/resources/assets/less/bem/chat-message-group.less
+++ b/resources/assets/less/bem/chat-message-group.less
@@ -78,7 +78,7 @@
     display: flex;
     word-break: break-word;
     justify-content: space-between;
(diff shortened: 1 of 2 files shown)


resources/assets/less/bem/chat-message-group.less
```diff
@@ -78,6 +78,7 @@
     display: flex;
     word-break: break-word;
     justify-content: space-between;
+    font-family: 'Open Sans', sans-serif;
```
notbakaneko:
```suggestion
    font-family: @font-content;
```

PR #4566, opened 2019-06-11
Generate event text for Bancho

### LINKED ISSUE #4562: Generated events don't populate text_clean field
This has not been an issue until recently, but now that the ranking process is handled by this system, it is causing in-game chat `#announce` lines to not appear.

Current:
- `text_clean` is not populated by `Event::generate`

Expected:
- `text_clean` should contain a (roughly) markdown version of `text`.
- Bancho reads from this field and nulls it once it has transferred it to in-game chat.

notbakaneko:
closes #4562

generates text in the format:
`[http://osuweb.test/beatmapsets/402995 Taneda Risa - Wareta Ringo (TV edit)] by [http://osuweb.test/users/475002 iamapen] has just been ranked!`

### DIFF
--- a/app/Models/Event.php
+++ b/app/Models/Event.php
@@ -95,8 +95,13 @@ class Event extends Model
                 $userUrl = e(route('users.show', $beatmapset->user, false));
                 $approval = e($beatmapset->status());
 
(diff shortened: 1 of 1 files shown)


peppy:
Confirming this is working as expected 👍

PR #4570, opened 2019-06-13
Restrict the viewing of history for private matches

### LINKED ISSUE #4491: Multiplayer match history links are brute-forceable
### Issue
Because multiplayer match history links are chronologically ordered, a dedicated user could easily abuse the osu! api to search for particular lobbies that have match data which are beneficial to them, particularly involving tournament play. I have been made aware that someone actually wrote a script in node/js to automate this.

### How this works from my understanding
Assuming that I am a player in a tournament, I can take beatmap IDs [from the mappool] or user IDs [of an opposing team member] and input it to the script. The script, when run, then finds me multiplayer matches which the map is played / the player is participating In, which gives me data to utilise (which maps are the opponents bad at? Pick that map!), resulting in me getting a slight advantage in my match against the opponents.)

### Impact
This has been an issue for a long time, however it has been recently popularised to some extent, people using it in team-based community tournaments. As players practice together before matches, a MP link is generated for their practice lobby. Those MP links are being easily found by opponents, using it to strategise their matches and revolving picks around maps that the opponents are bad at.

This *could* also happen during official tournaments too, as teams also practice on Bancho.

### Possible solutions
1. **Add a limit to how fast you can query for multiplayer lobbies using the API**
One issue with this is that if you know the start time of the lobby, you can just keep reducing the x in multiplayer/matches/xxxxxxx until you find what you are looking for.

2. **Private match history links**
For example, you could private multiplayer matches that you created using the "!mp make" command, to prevent abuse. Not sure how tedious that is from a development standpoint.

Not sure how much of this is actually an issue, but I am curious to hear the osu! team's thoughts on this. Please do not hesitate to ask if you have any questions, I am sure that there is people who are willing to chime in about this issue.

nekodex:
A private match that doesn't have 'public match history' enabled requires a user to have participated in the match to view the history.

NOTE: a bancho/client update and manual sql migration is required before this goes live.

fixes #4491

### DIFF
--- a/app/Libraries/OsuAuthorize.php
+++ b/app/Libraries/OsuAuthorize.php
@@ -28,7 +28,6 @@ use App\Models\Forum\Authorize as ForumAuthorize;
 use App\Models\Forum\Topic;
 use App\Models\Forum\TopicCover;
 use App\Models\Multiplayer\Match as MultiplayerMatch;
-use App\Models\Multiplayer\Match;
 use App\Models\User;
 use App\Models\UserContestEntry;
 use Carbon\Carbon;
--- a/app/Libraries/OsuAuthorize.php
+++ b/app/Libraries/OsuAuthorize.php
@@ -1143,11 +1143,11 @@ class OsuAuthorize
 
         $this->ensureLoggedIn($user);
 
-        if (!$match->hadPlayer($user)) {
-            return 'unauthorized';
+        if ($user->canModerate() || $match->hadPlayer($user)) {
+            return 'ok';
         }
 
-        return 'ok';
+        return 'unauthorized';
     }
 
     public function checkUserSilenceShowExtendedInfo($user)
--- a/app/Libraries/OsuAuthorize.php
+++ b/app/Libraries/OsuAuthorize.php
@@ -28,6 +28,7 @@ use App\Models\Forum\Authorize as ForumAuthorize;
 use App\Models\Forum\Topic;
 use App\Models\Forum\TopicCover;
 use App\Models\Multiplayer\Match as MultiplayerMatch;
+use App\Models\Multiplayer\Match;
 use App\Models\User;
 use App\Models\UserContestEntry;
 use Carbon\Carbon;
@@ -1133,6 +1134,21 @@ class OsuAuthorize
         } else {
             return $prefix.'no_access';
         }
+    }
+
+    public function checkMatchView($user, $match)
+    {
+        if (!$match->private) {
+            return 'ok';
+        }
+
+        $this->ensureLoggedIn($user);
+
+        if (!$match->hadPlayer($user)) {
+            return 'unauthorized';
+        }
+
+        return 'ok';
     }
 
     public function checkUserSilenceShowExtendedInfo($user)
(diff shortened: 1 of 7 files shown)


app/Libraries/OsuAuthorize.php
```diff
@@ -1135,6 +1135,21 @@ public function checkUserShow($user, $owner)
...
 
+    public function checkMatchView($user, $match)
+    {
+        if (!$match->private) {
+            return 'ok';
+        }
+
+        $this->ensureLoggedIn($user);
+
+        if (!$match->hadPlayer($user)) {
```
WalterToro: Probably a comment, I would let GMT/NATs and other elevated users to check these private matches for moderation issues (checking cheaters in multiplayer rooms, etc.)

peppy: admin already overrides everything.

tests/Controllers/MatchesControllerTest.php
```diff
@@ -0,0 +1,97 @@
...
+        factory(Event::class)->states('create')->create([
+            'match_id' => $this->privateMatch->match_id,
+        ]);
+        $this->privateMatchRoute = route('matches.show', $this->privateMatch->match_id);
+    }
+
+    public function testPublicMatchLoggedOut() // OK
+    {
+        $this
+            ->actingAs($this->user)
```
notbakaneko: logged out actingAs user ?_?

tests/Controllers/MatchesControllerTest.php
```diff
@@ -0,0 +1,97 @@
...
+
+    public function testPublicMatchLoggedInNotParticipated() // OK
+    {
+        $this
+            ->actingAs($this->user)
+            ->get($this->publicMatchRoute)
+            ->assertStatus(200);
+    }
+
+    public function testPublicMatchLoggedInParticipated() // OK
```
notbakaneko: is this test missing the event?

Negaosu:
Are matches public by default or private by default?

peppy:
Public by default. You will see an option in the osu! client starting next week.

Negaosu:
not sure if the change is done or not yet but I found a bug. I created a lobby and played some maps (https://osu.ppy.sh/community/matches/52736343) and it won't let me see the match history even though I'm logged in - I get this https://i.imgur.com/Wx3IwVC.png

This has happened to a few others I know as well, not sure what could have caused that.

peppy:
Thanks. We've identified the issue.

peppy:
@Negaosu please confirm it's working correctly now

Negaosu:
Yep it works correctly. Thanks!

Negaosu:
> @Negaosu please confirm it's working correctly now

Yep it works correctly now, thanks :D

(also a quick suggestion is maybe that any "refs" of the match can also be allowed to view the lobby page? so for example if one of the team members can't make the practice lobby, or if someone is hosting a lobby that they only want specific people to be able to see, they could do !mp addref [name] and they'd be able to see the page? just food for thought)

Negaosu:
@peppy @nekodex 

sorry for tag, just a quick (arguably needed xd) suggestion

sometimes not the whole team is able to make it to a practice lobby, but it will still be helpful for them to see the link. could you add a command like !mp access [username] or something along those lines which would allow the specified user to see the mp link (even if they are not currently online in game)?

thanks a lot!

peppy:
this is out of scope for now. please use file > save page in your browser or take a screenshot to share with people not in the room.

PR #4577, opened 2019-06-17
Disable beatmap download in listing when not available

### LINKED ISSUE #4560: Search results don't correctly handle disabled download beatmaps
image

image

This came up on discussion of implementing this in lazer, and realising the required attributes are missing from the API> For beatmap panels, we should only require the `download_disabled` attribute, to disable the button. The rest of the information can remain on the beatmap info page.

notbakaneko:
`availability` is now always included in the response

closes #4560


resources/assets/coffee/react/beatmapset-page/header.coffee
```diff
@@ -142,15 +142,15 @@ export class Header extends React.Component
 
           el BeatmapsetMapping, beatmapset: @props.beatmapset
 
-          if currentUser.id? && @props.beatmapset.availability
+          if currentUser.id?
```
nanaya: `parts-removed` is incorrectly visible on normal maps.


disabled | url | description
---|---|---
false | empty | normal (no availability info box)
false | present | parts-removed
true | empty | disabled, no info link
true | present | disabled, with info link

resources/assets/coffee/react/beatmapset-page/header.coffee
```diff
@@ -142,17 +147,19 @@ export class Header extends React.Component
...
               className: 'beatmapset-header__availability-info',
               if @props.beatmapset.availability.download_disabled
                 osu.trans 'beatmapsets.availability.disabled'
-              else
-                osu.trans 'beatmapsets.availability.parts-removed'
-
-              if @props.beatmapset.availability.more_information
-                div className: 'beatmapset-header__availability-link',
-                  a href: @props.beatmapset.availability.more_information, target: '_blank', osu.trans 'beatmapsets.availability.more-info'
+              else if @props.beatmapset.availability.more_information?
```
nekodex: This be backwards - should be checking if `.more_information` exists and falling back to the default message only if it's missing.

PR #4582, opened 2019-06-18
Add wikilink to the article in the restriction banner

### LINKED ISSUE #4581: Make the "Account Restrictions" wiki page more visible to restricted users
It would be nice to add https://osu.ppy.sh/help/wiki/Help_Centre/Account_Restrictions to the banner that appears on the website when users are restricted, as it could prevent confusion for users who don't know what a restriction is or why their account is not in good standing.

TPGPL:
Intends to resolve #4581.


peppy:
Thanks for the contribution but it is obvious there is zero testing or effort put into this change. Please test before making a pull request.

notbakaneko:
this can be reopened when it gets updated

PR #4608, opened 2019-07-02
Update PHP dependencies

nanaya:
~~Part 2 would be upgrading commonmark~~. Haven't even tested this one apart of tinker starts fine and some pages still load.

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -151,7 +151,7 @@ class Handler extends ExceptionHandler
 
         app('sentry')->configureScope(function ($scope) use ($e, $userContext) {
             $scope->setUser($userContext);
-            $scope->setTag('http_code', $this->statusCode($e));
+            $scope->setTag('http_code', (string) $this->statusCode($e));
         });
 
         $ref = app('sentry')->captureException($e);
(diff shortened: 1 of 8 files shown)


nanaya:
commonmark update isn't going to happen yet as the table extension still hasn't been updated.

app/Jobs/RegenerateBeatmapsetCover.php
```diff
@@ -71,11 +72,9 @@ public function handle()
             Datadog::increment(['thumbdonger.processed', 'thumbdonger.error']);
             Log::warning("[beatmapset_id: {$this->beatmapset->beatmapset_id}] Cover regeneration FAILED.");
             if (config('osu.beatmap_processor.sentry')) {
-                $tags = [
-                    'beatmapset_id' => $this->beatmapset->beatmapset_id,
-                ];
-                $client = new Raven_Client(config('osu.beatmap_processor.sentry'), ['tags' => $tags]);
-                $client->captureException($e);
+                $client = ClientBuilder::create(['dsn' => config('osu.beatmap_processor.sentry')])->getClient();
+                $scope = (new Scope)->setTag('beatmapset_id', $this->beatmapset->beatmapset_id);
```
notbakaneko: should have an explicit cast to `string`

app/Exceptions/Handler.php
```diff
@@ -154,9 +149,12 @@ private function reportWithSentry($e)
             ];
         }
 
-        Sentry::user_context($userContext);
+        app('sentry')->configureScope(function ($scope) use ($e, $userContext) {
+            $scope->setUser($userContext);
+            $scope->setTag('http_code', $this->statusCode($e));
```
notbakaneko: should have an explicit cast to `string` since `statusCode` should return `int` only

PR #4612, opened 2019-07-03
Add authorized oauth clients management

notbakaneko:
Adds a list of oauth clients a user has granted permissions to, and enables them to revoke the permissions granted. Revoking access to a client will revoke ALL tokens granted by that user to the client.

Also removes a bunch of oauth routes we're not supporting at this time.


resources/assets/lib/oauth/authorized-client.tsx
```diff
@@ -0,0 +1,98 @@
...
+      </div>
+    );
+  }
+
+  renderPermissions() {
+    const scopes = Array.from(this.props.client.scopes).sort();
+    return (
+      <>
+        <div>{osu.trans('oauth.authorized-clients.scopes_title')}</div>
+        <ul className='oauth-scopes'>
```
nanaya: looks a bit weird when there's no permission but not sure if worth fixing.

resources/assets/less/bem/authorized-client.less
```diff
@@ -0,0 +1,68 @@
...
+
+  margin-bottom: 10px;
+  padding-bottom: 10px;
+  border-bottom: 1px solid #444;
+
+  &__actions {
+    align-items: flex-start;
+  }
+
+  &__button {
```
nanaya: button vs div default text alignment

resources/assets/lib/oauth/client-json.ts
```diff
@@ -0,0 +1,27 @@
...
+ *    osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+ *    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *    See the GNU Affero General Public License for more details.
+ *
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export interface ClientJSON {
+  created_at: string;
```
nanaya: Not in transformer? Doesn't seem to be used either.

resources/assets/lib/oauth/client-json.ts
```diff
@@ -0,0 +1,27 @@
...
+  id: number;
+  name: string | null;
+  revoked: boolean;
```
nanaya: Same as `created_at`.

PR #4620, opened 2019-07-04
Limit kudosu per user per beatmapset

nanaya:


dorothy3242:
Unless I'm reading this wrong, 10 per set seems to be extremely low, as even a shorter set can get ~15 points by a modder if checked for all issues.

50 could be a better limit as it's high enough to still reward super in depth mods on maps that need them, but low enough to prevent people from spamming 200 points on one mapset to farm kds.

Setting the limit too low can result in users cutting mods short with the limit in mind.

Icekalt:
Heavily agree to noffy since putting max kds that low will affect modder with modding less in general. The benefits from that rule would way too low since kds abuse isnt a common thing (since its pretty useless right now). If you gonna implement that rule you really should aim for sth like 35-50 kds.

peppy:
The value there is a variable. It is a variable. Variables change.

nanaya:
Need to remember setting correct limit before deploying @peppy 

```bash
BEATMAPSET_DISCUSSION_KUDOSU_PER_USER=10
```

peppy:
Yep thanks, already have.

PR #4636, opened 2019-07-10
Site visual revamp (aka night mode-ish)

nekodex:
This implements the new colour system that @arflyte has devised, wherein all colours are derived from a single hue, with only the saturation+luminance varying:

As a result, pages will now use 'relative' colours for their designs - meaning that if the base hue is changed (e.g. page is moved to another section, etc), the design will 'just work' with no additional changes:

rainbows-optimized

It's all dynamically calculated in css from the base hue that's defined as a css var.

This PR also goes through and unifies all the sections of the site to use this new colour system. No more random pages on the site using their own colours/shades of gray/etc.

There's a few sections that were intentionally left out in an effort to make this PR smaller:
* admin section
* beatmap cards
* /matches/ page
* store
* forums (there's a redesign waiting to be implemented)
* support osu! page (there's a redesign waiting to be implemented)
* anchor tag styling unification - right now their styling is still all over the place...

Due to the size of this site/change, I'm sure I probably missed a section or a popup or something - so please click around and let me know if I missed anything.

I'll probably do another pass in a separate PR to do cleanup of old variables/colours/etc.

As this was a rather necessarily large PR, it might help to look at each commit instead of the overview of files changed...


resources/views/users/verify.blade.php
```diff
@@ -16,8 +16,10 @@
     along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
 --}}
 @extends('master', [
+    // Verification doesn't inherit from App\Controller, thus these variables aren't set. Thus we set them here:
     'currentSection' => 'error',
     'currentAction' => '401',
+    'currentSectionHue' => section_to_hue_map('error'),
```
nanaya: wouldn't it be easier to just call `section_to_hue_map` inside the layout itself :question:

PR #4678, opened 2019-07-17
Prevent null plural translation

nanaya:
I guess this is better than having null when the translation is incomplete.

<sub>It will also allow us to be lazy not checking plural translations.</sub>


resources/assets/coffee/osu_common.coffee
```diff
@@ -327,16 +327,32 @@
 
 
   transChoice: (key, count, replacements = {}, locale) ->
+    locale ?= currentLocale
+
     if !osu.transExists(key, locale)
-      locale = fallbackLocale
-      initialLocale = Lang.getLocale()
+      return osu.transChoice(key, count, replacements, fallbackLocale)
```
notbakaneko: infinite loop when key doesn't exist at all.

PR #4680, opened 2019-07-17
New theme fixes

### LINKED ISSUE #4653: New color scheme in site causes very hard to read texts
Referring to the recent change in color scheme, there are a lot of parts that has very hard to read texts due to extremely similar text colors. This could impact the readability of certain texts, especially to those who have visual impairments, or those devices that has problems with color display.

Examples include:
Screenshot_2019-07-12-18-07-19-548_org mozilla fennec_aurora
1) the username color for beatmap nominators
2) Post dates

image
Lots of people are complaining about this too and this is on the main page, the ranked dates.

Also another thing to mention is the background color and the text color being similar (like the main page with blue bg color and blue text and modding page with purple text and purple by color) would severely decrease the readability of the texts for those who have visual impairment on particular colors.

nekodex:
First pass of fixes for the new theme, mainly focusing on addressing the low-contrast text issue.

- A new F1 colour variant is now being used in place of B1 for text, which should hopefully fix the issues people have had with the more subdued text.
- Also changes the graphs on user profiles back to yellow to return some colour
- Various other colour tweaks/fixes for stuff that I stumbled across

resolves #4653

More to come...

### DIFF
--- a/resources/assets/less/bem/simple-menu.less
+++ b/resources/assets/less/bem/simple-menu.less
@@ -27,7 +27,7 @@
   .own-layer();
   .default-box-shadow();
   border-radius: @_border-radius;
-  background-color: @osu-colour-b5;
+  background-color: @osu-colour-b6;
   padding: @_padding-vertical @_gutter;
   display: flex;
   flex-direction: column;
@@ -70,7 +70,6 @@
     .own-layer();
     will-change: transform, opacity;
     transition: none;
-    background-color: @osu-colour-b6;
     border-radius: @border-radius-large;
   }
 
@@ -178,7 +177,7 @@
     flex: none;
 
     &:hover {
-      background-color: #111;
+      background-color: @osu-colour-b5;
       .highlight-stripe();
     }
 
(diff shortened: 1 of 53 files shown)


resources/views/beatmapset_watches/index.blade.php
```diff
@@ -109,7 +109,7 @@ class="beatmapset-watches__cover"
                 @endif
             </table>
 
-            @include('objects._pagination_v0', ['object' => $watches])
+            @include('objects._pagination_v0', ['object' => $watches, 'modifiers' => ['light-bg']])
```
nanaya: the page seems to be dark?

resources/assets/less/bem/simple-menu.less
```diff
@@ -178,8 +177,12 @@
     flex: none;
 
     &:hover {
-      background-color: #111;
+      background-color: @osu-colour-b5;
       .highlight-stripe();
+
+      .@{_top}--nav2 & {
+        background-color: transparent;
```
nanaya: b6 for the new main menu, b5 for the rest

PR #4683, opened 2019-07-18
Return UserTransformer counts as primitives

### LINKED ISSUE #4640: User model unnecessarily exposes counts as arrays
```
favourite_beatmapset_count
graveyard_beatmapset_count
loved_beatmapset_count
ranked_and_approved_beatmapset_count
unranked_beatmapset_count
```

notbakaneko:
Return just the value instead of wrapping the value in an array.
Client will need to be updated if it relied it on being wrapped @peppy 

closes #4640


nekodex:
btw talk to @nanaya about adding a changelog to the apidocs listing breaking changes

PR #4684, opened 2019-07-18
Dark design pass on admin section

nekodex:
Order listing page left bright as it's not really used ATM.

Also adds an 'osu-table' class that will be used to standardise simple tables (in the style of the beatmap leaderboards).

### DIFF
--- a/resources/assets/less/bem/admin-contest-entry.less
+++ b/resources/assets/less/bem/admin-contest-entry.less
@@ -56,7 +56,6 @@
 
     &--button {
       text-align: center;
-      background-color: @osu-colour-b5 !important;
     }
   }
 }
--- a/resources/assets/less/bem/admin-contest-entry.less
+++ b/resources/assets/less/bem/admin-contest-entry.less
@@ -27,12 +27,17 @@
 
   &__user-column {
     width: 250px;
+    padding-left: 10px;
   }
 
   &__avatar {
     display: inline-block;
     width: 32px;
     margin-right: 5px;
+  }
+
+  &__preview {
+    padding: 20px;
   }
 
   &__user-link {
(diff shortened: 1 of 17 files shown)


resources/assets/less/bem/admin-contest-entry.less
```diff
@@ -51,6 +56,7 @@
 
     &--button {
       text-align: center;
+      background-color: @osu-colour-b5 !important;
```
nanaya: or just remove `osu-table__cell` element for this column?

resources/views/admin/contests/show.blade.php
```diff
@@ -72,11 +72,13 @@
             <dd class="contest">
                 <div class="contest__description">{!! markdown($contest->description_enter) !!}</div>
             </dd>
+            <br />
             <dt class="admin-contest__meta-row">Voting Description</dt>
             <dd class="contest">
                 <div class="contest__description">{!! markdown($contest->description_voting) !!}</div>
             </dd>
             @if ($contest->extra_options !== null)
+                <br />
```
nanaya: `<dl>` may not contain `<br />`...

[Reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl)

resources/assets/less/bem/osu-layout.less
```diff
@@ -166,9 +165,75 @@
...
+        color: white;
+      }
+
+      h1 {
+        font-size: 28px;
+      }
+
+      h2 {
+        font-size: 16px;
+        display: inline-block;
```
nanaya: this probably should be `width: max-content` instead so it'll still occupy the entire row instead of becoming inlined element.

nanaya: and a `max-width: 100%` :dancer:

PR #4698, opened 2019-07-21
Fix multiplayer match history page not updating automatically

### LINKED ISSUE #4176: Multiplayer match history not auto update match result
It only update on minor event (player join or left) but not update the match result. The new match result only show when user manually refresh the page. I'm using Edge 42.17134.1.0 and Chrome 72.0.3626.96.

Also it can noticeable on "long" time gap between events where the result should be there 

when it's not updated
after refresh (the correct one)

LiquidPL:
Fixes #4176.

Also includes a set of artisan commands handy for debugging said mp history pages.


app/Transformers/Multiplayer/GameTransformer.php
```diff
@@ -35,6 +35,7 @@ class GameTransformer extends Fractal\TransformerAbstract
     public function transform(Game $game)
     {
         return [
+            'game_id' => $game->game_id,
```
nanaya: Just `id`.

app/Console/Commands/MultiplayerEventCreate.php
```diff
@@ -0,0 +1,80 @@
+<?php
```
nanaya: All "testing" commands should be a separate pull request. And maybe as factory, not commands.

PR #4699, opened 2019-07-21
Add translations for attribute strings in model validation

### LINKED ISSUE #4575: Post exceeded maximum length message
I managed to hit the character limit with the completed skins compendium. I assume I'm probably the first one to do so, because this just seems like some debug message and not something an user should have seen. I doubt anyone else will see it, but it would still be nice to have a proper text here.

https://puu.sh/DGoyj/dd79ec7468.png

LiquidPL:
Fixes #4575.


LiquidPL:
Did that, and also went around the code and added translations for attributes for stuff that used it.

app/Libraries/ValidationErrors.php
```diff
@@ -41,7 +41,7 @@ public function add($column, $rawMessage, $params = null) : self
         }
         $rawMessage = $this->keyBase.$rawMessage;
 
-        $params['attribute'] = $column;
+        $params['attribute'] = trans($this->keyBase.$this->prefix.'.'.$column);
```
nanaya: - should check if translation exists
- group attribute/column translations so they're not mixed with actual error messages (inside `...{$this->prefix}.attributes...` or something).

app/Libraries/ValidationErrors.php
```diff
@@ -43,6 +45,12 @@ public function add($column, $rawMessage, $params = null) : self
 
         $params['attribute'] = $column;
 
+        $attributeKey = $this->keyBase.$this->prefix.'.attributes.'.$column;
+
+        if (Lang::has($attributeKey)) {
```
nanaya: Use `trans_exists` as Lang::has doesn't handle crowdin's empty string translation.

(and maybe optionally make it so `$params['attribute']` only set once)

nanaya: Or maybe it's fine.

PR #4712, opened 2019-07-24
Add verification link

nanaya:
Click the link and the open verification box will be magically verified. Or something.

### DIFF
--- a/app/Events/UserSessionEvent.php
+++ b/app/Events/UserSessionEvent.php
@@ -33,10 +33,9 @@ class UserSessionEvent extends NotificationEventBase
         return new static('logout', $userId, compact('keys'));
     }
 
-    public static function newVerificationRequirementChange($userId, $key, $isRequired)
+    public static function newVerificationRequirementChange($userId, $isRequired)
     {
         return new static('require_verification', $userId, [
-            'key' => $key,
             'requires_verification' => $isRequired,
         ]);
     }
--- a/app/Events/UserSessionEvent.php
+++ b/app/Events/UserSessionEvent.php
@@ -35,7 +35,7 @@ class UserSessionEvent extends NotificationEventBase
 
     public static function newVerificationRequirementChange($userId, $isRequired)
     {
-        return new static('require_verification', $userId, [
+        return new static('verification_requirement_change', $userId, [
             'requires_verification' => $isRequired,
         ]);
     }
--- a/app/Events/UserSessionEvent.php
+++ b/app/Events/UserSessionEvent.php
@@ -28,12 +28,12 @@ class UserSessionEvent extends NotificationEventBase
     public $data;
     public $userId;
 
-    public static function onLogout($userId, $keys)
+    public static function newLogout($userId, $keys)
     {
         return new static('logout', $userId, compact('keys'));
     }
 
-    public static function onVerified($userId, $key)
+    public static function newVerified($userId, $key)
     {
         return new static('verified', $userId, compact('key'));
     }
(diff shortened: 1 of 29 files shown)


app/Http/Middleware/VerifyPrivilegedUser.php
```diff
@@ -24,6 +24,6 @@ class VerifyPrivilegedUser extends VerifyUser
 {
     public function requiresVerification($request)
     {
-        return $this->auth->user() === null || $this->auth->user()->isPrivileged();
+        return session()->get('requires_verification') ?? optional(auth()->user())->isPrivileged();
```
notbakaneko: need some way of invalidating `requires_verification` on privilege change

app/Events/UserSessionEvent.php
```diff
@@ -22,28 +22,39 @@
...
 
-class UserLogoutEvent extends NotificationEventBase
+class UserSessionEvent extends NotificationEventBase
 {
     public $action;
+    public $data;
     public $userId;
-    public $channelName;
+
+    public static function onLogout($userId, $keys)
```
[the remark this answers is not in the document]
nanaya: Can use some suggestion :dancer:

app/Libraries/Session/Store.php
```diff
@@ -194,7 +193,7 @@ public function currentUserSessions()
             $sessionMeta[$id]['device'] = $agent->device();
             $sessionMeta[$id]['platform'] = $agent->platform();
             $sessionMeta[$id]['browser'] = $agent->browser();
-            $sessionMeta[$id]['verified'] = isset($session['verified']) && $session['verified'] === UserVerification::VERIFIED;
+            $sessionMeta[$id]['verified'] = (bool) ($session['verified'] ?? false);
```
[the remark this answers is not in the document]
nanaya: Preventing failure when the key doesn't exist

app/Libraries/UserVerificationState.php
```diff
@@ -0,0 +1,213 @@
...
+        $session->start();
+
+        return new static([
+            'user' => User::find($params['userId']),
+            'session' => $session,
+            'legacySessionQueryWhere' => $params['legacySessionQueryWhere'],
+        ]);
+    }
+
+    public function __construct(array $params)
```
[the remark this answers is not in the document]
nanaya: One of the variable name is long.

app/Http/Middleware/VerifyPrivilegedUser.php
nanaya: Not sure how to cleanly handle that on notification server side.

Probably easier to just kick out the user's session.

app/Events/UserSessionEvent.php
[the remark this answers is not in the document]
notbakaneko: `newLogout`?

app/Http/Middleware/VerifyPrivilegedUser.php
notbakaneko: if it's for the notification server, the web side shouldn't need to rely on the session value?

nanaya: Both should rely on same thing...?

notbakaneko: one checks per-request and the other is a persistent connection, I don't think it should be considered the same?

nanaya: More like per session...

notbakaneko: moving a logged in user to a privileged group currently triggers a verification on the next request, but this change makes it no longer trigger the verification

peppy: Not sure this is a real issue, to be honest.

notbakaneko: Either:
- Privilege changes should force logout a user,
- the session value should be invalidated,
- or the session value should not be used

PR #4719, opened 2019-07-25
Add url support to profile badges

### LINKED ISSUE #4211: Add url support for profile badges
Would be nice to be able to get to the origin of the bad (ie. a wiki page for the tournament or a news page for the award).

Requires column to be added to the database.

LiquidPL:
Fixes #4211.

### DIFF
--- a/resources/assets/coffee/react/profile-page/badges.coffee
+++ b/resources/assets/coffee/react/profile-page/badges.coffee
@@ -24,16 +24,11 @@ export class Badges extends React.PureComponent
   render: =>
     div className: 'profile-badges',
       for badge in @props.badges
-        badgeEl = div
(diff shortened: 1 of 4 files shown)


resources/assets/coffee/react/profile-page/badges.coffee
```diff
@@ -17,15 +17,23 @@
...
-import { div, img } from 'react-dom-factories'
+import { div, img, a } from 'react-dom-factories'
 el = React.createElement
 
 export class Badges extends React.PureComponent
   render: =>
     div className: 'profile-badges',
       for badge in @props.badges
-        div
+        badgeEl = div
```
nanaya: Set element type based on whether it should be a link or not.

PR #4722, opened 2019-07-26
Add proper meta description for contests

TPGPL:
image

changed the meta description for contests as i don't think the current one is user-friendly

also part of #220


nanaya:
The new description is redundant with page title. Not sure if useful.

nanaya:
Better than just duplicating the title.

PR #4723, opened 2019-07-26
Display creator's current username at beatmap panel

### LINKED ISSUE #1975: Mapper namechanges show on information page but not beatmap search
example 1:
https://osu.ppy.sh/beatmapsets/468938#osu/1002859
beatmap search will display their old name
image
beatmap information page will display their current name
image

versus ingame:
image

It should still be displayed with the name they had when the mapset was ranked, so that it is consistent with what is seen in-game.

Example 2:
https://osu.ppy.sh/beatmapsets/375877#osu/1460396 the mapper changed his name to MariahCarey and back specifically so this beatmap could be ranked under the name MariahCarey to match the artist field, but this is not displayed on the new web's information page.. so .. uhh.. rip.

TPGPL:
intends to resolve #1975


TPGPL:
closing until i figure out the solution

PR #4724, opened 2019-07-26
Add direct link for beatmap packs

### LINKED ISSUE #4246: Keep direct links for Beatmap Packs on new website
At the moment you can post a link directly from a beatmap pack from the old website design, the format is 
`https://osu.ppy.sh/p/packlist/?n=<letter><number>#<letter><number>` and is a pretty handy feature, it's not that much used, yet I work a lot with that.

This however is not possible for the new forum design, and I would love if this could be kept. The current URL on the new website is `https://osu.ppy.sh/beatmaps/packs<letter><number>` which is pretty easy to use it but doesn't seem to work.

cl8n:
fixes #4246 , external links should use /beatmaps/packs/{id} or /beatmaps/packs/{tag}


app/Http/Controllers/BeatmapPacksController.php
```diff
@@ -55,4 +67,14 @@ public function show($id)
 
         return view('packs.show', compact('pack', 'sets', 'mode'));
     }
+
+    private function indexLink(BeatmapPack $pack) : string
+    {
+        $type = $pack->type();
+        $indexInPagination = BeatmapPack::getPacks($type)->get()->search($pack);
```
nanaya: Not sure if want to do this. It doesn't expand either. Might as well just display expanded single item.

app/Http/Controllers/BeatmapPacksController.php
```diff
@@ -38,11 +39,22 @@ public function index()
...
+        if (is_numeric($idOrTag)) {
+            $pack = BeatmapPack::findOrFail($idOrTag);
+        } else {
+            $pack = BeatmapPack::where('tag', $idOrTag)->firstOrFail();
+        }
+
+        return ujs_redirect($this->indexLink($pack));
+    }
+
+    public function raw($id)
```
nanaya: the view is still called `show`.

resources/views/packs/index.blade.php
```diff
@@ -52,6 +52,7 @@
             <div class="beatmap-packs__list js-accordion">
                 @foreach ($packs as $pack)
                     <div class="beatmap-pack js-beatmap-pack js-accordion__item" data-pack-id="{{ $pack['pack_id'] }}">
+                        <a id="pack-{{ $pack['pack_id'] }}"></a>
```
nanaya: This doesn't work and blocks things above it.

nanaya:
Also the list items should be changed to link instead of plain div.

app/Http/Controllers/BeatmapPacksController.php
```diff
@@ -38,11 +39,22 @@ public function index()
...
         return view('packs.index')
-            ->with('packs', $packs->paginate(20)->appends(['type' => $type]))
+            ->with('packs', $packs->paginate(static::PER_PAGE)->appends(['type' => $type]))
             ->with('type', $type);
     }
 
-    public function show($id)
+    public function show($idOrTag)
+    {
+        if (is_numeric($idOrTag)) {
```
notbakaneko:
```php
>>> is_numeric('1e2')
=> true
>>> is_numeric('1.2')
=> true
>>> is_numeric('1')
=> true
>>> ctype_digit('1.2')
=> false
>>> ctype_digit('1e2')
=> false
>>> ctype_digit('1')
=> true
```

notbakaneko:
Closing due to lack of updates

peppy:
Reopening this one, should be easy to complete. @cl8n can you let us know if you plan on fixing this up? If not one of us will get it updated and merged.

cl8n:
not in the near future

iirc I didn't continue with the PR mostly because this seemed like a dumb workaround, but I couldn't figure out a better solution (nanaya commented there too)
https://github.com/ppy/osu-web/blob/218f097379fe01cd31d2b0079d2c6d8f5e815397/app/Http/Controllers/BeatmapPacksController.php#L71-L79

peppy:
Can someone look into this? It looks to be a prerequisite of a [news post](https://github.com/ppy/osu-wiki/pull/2667)

notbakaneko:
I think we should just go with adding a page that can show a single pack with a link back to the index.

peppy:
Sounds fine if that's the easiest way to resolve this.

nanaya:
Updated to
- show single item when accessed directly
- add link to them
- reworked css classes

app/Models/BeatmapPack.php
```diff
@@ -66,6 +66,11 @@ public function downloadUrl()
         return $this->downloadUrls()[0];
     }
 
+    public function type() : string
+    {
+        return array_search($this->tag[0], static::$tagMappings, true);
+    }
+
```
notbakaneko: Don't need this, I think?

PR #4729, opened 2019-07-27
Add meta description for tournaments

TPGPL:
part of #220


nanaya:
Same as other PR, it's redundant with page title.

TPGPL:
beatmaps use basically the same page description as this PR and tournaments dont have their own page description so i think it is fine to have it

nanaya:
The one in beatmapset is equally useless. There's no need to copy that.

PR #3560, opened 2018-08-04
Fix font issue for zh and zh-tw

kj415j45:

PR #4736, opened 2019-07-29
Add fonts specific to zh-tw

Raymonf:
This commit makes the website use Microsoft JhengHei instead of Microsoft YaHei for Traditional Chinese (`zh-tw`).

I'm not sure what `Hiragino Sans GB` is, but the equivalent for Traditional Chinese is `Hiragino Sans TC`, apparently.

The glyphs for JhengHei adhere to the standards that Taiwan has put out. In general, the font is lighter and increases readability (in my opinion) for Traditional Chinese.

Note for historical purposes: #3560 is the commit that added the `zh` fonts for both Simplified and Traditional Chinese.


resources/assets/less/variables.less
```diff
@@ -22,6 +22,7 @@
 @font-default: 'Exo 2', 'Helvetica Neue', Tahoma, Arial, 'Hiragino Kaku Gothic ProN', Meiryo, 'Microsoft YaHei', 'Apple SD Gothic Neo', sans-serif;
 @font-default-vi: Exo, 'Exo 2', 'Helvetica Neue', Tahoma, Arial, 'Hiragino Kaku Gothic ProN', Meiryo, 'Microsoft YaHei', 'Apple SD Gothic Neo', sans-serif;
 @font-default-zh: 'Exo 2', 'Helvetica Neue', Tahoma, Arial, 'Hiragino Sans GB', 'Microsoft YaHei', 'Apple SD Gothic Neo', system-ui, sans-serif;
+@font-default-zh-tw: 'Exo 2', 'Helvetica Neue', Tahoma, Arial, 'Hiragino Sans TC', 'Microsoft JhengHei', 'Apple SD Gothic Neo', system-ui, sans-serif;
```
notbakaneko: `Hiragino Sans GB` is a MacOS font, `Hiragino Sans TC` isn't

Raymonf: Ah! I didn't realize it was bundled in macOS. Then, I'll change it to PingFang TC.

PR #4752, opened 2019-08-03
Remove spinner from non-voted entries on contest page

### LINKED ISSUE #4734: Contest entry loading icon shows on every entry when voting
chrome_2019-07-28_17-52-42
this spinning loading icon shows on every entry except the one being voted for. the star turns a lighter color after the vote request finishes, so I feel like this loading icon doesn't need to exist at all

picture is taken right after clicking on the vote button

LiquidPL:
Fixes #4734.

### DIFF
--- a/resources/assets/coffee/react/contest-voting/art-entry-list.coffee
+++ b/resources/assets/coffee/react/contest-voting/art-entry-list.coffee
@@ -35,6 +35,7 @@ export class ArtEntryList extends BaseEntryList
         key: index,
         displayIndex: index,
         entry: entry,
+        clickedEntry: @state.clickedEntry
         waitingForResponse: @state.waitingForResponse,
         options: @state.options,
         contest: @state.contest,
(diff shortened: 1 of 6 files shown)


nekodex:
See my comment on the parent issue - the spinner shouldn't be removed, just the behaviour corrected.

nekodex:
I'm going to close this PR because the proposed change results in a worse user experience than the current implementation.

Some background as to why:
When the page was first written, it was done in such a way that only one voting action could be performed at once (i.e. no async voting). That's why the spinner was made visible on all the entries at the same time - to give clear feedback to the user that you can't currently make another vote.

This PR removes that feedback without removing the simultaneous voting limitation, which makes things more confusing for the user. For example, take the case where a user with high latency internet places a vote and then tries to interact with other voters before the first vote has gone through... they will be left wondering why nothing is happening when they click/tap.

A proper fix for this would be to update the voting system to be able to handle asynchronous voting (remembering to keep track of outgoing votes vs max number of votes for the current contest and rolling back votes when errors occur, etc), but that's more complicated which is why it wasn't done initially.

PR #4753, opened 2019-08-03
Improve suffixed numbers formatter

TPGPL:
Adds an ability to use shortened suffixes for numbers

---


nanaya:
There's no use for this at the moment.

PR #4756, opened 2019-08-04
Create a global user badge component with title

### LINKED ISSUE #4690: Add tooltips with full group name where abbreviations are used for display
It doesn't seem to be taking too much place while it used to be that way and it would be more clear I think.

TPGPL:
Resolves #4690

### DIFF
--- a/resources/assets/coffee/react/_components/user-badge.coffee
+++ b/resources/assets/coffee/react/_components/user-badge.coffee
@@ -1,0 +1,26 @@
+###
+#    Copyright (c) ppy Pty Ltd <contact@ppy.sh>.
+#
+#    This file is part of osu!web. osu!web is distributed with the hope of
+#    attracting more community contributions to the core ecosystem of osu!.
+#
+#    osu!web is free software: you can redistribute it and/or modify
+#    it under the terms of the Affero GNU General Public License version 3
+#    as published by the Free Software Foundation.
+#
+#    osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+#    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#    See the GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+###
+
+import * as React from 'react'
+import { div } from 'react-dom-factories'
+bn = 'user-group-badge'
+
+export UserBadge = ({type}) ->
+  div
+    className: osu.classWithModifiers(bn, type)
+    title: osu.trans "common.group_name.#{type}"
--- a/resources/assets/coffee/react/_components/user-badge.coffee
+++ b/resources/assets/coffee/react/_components/user-badge.coffee
@@ -19,8 +19,19 @@
 import * as React from 'react'
 import { div } from 'react-dom-factories'
 bn = 'user-group-badge'
+group_names =
+  bot: 'Automated Message'
+  bng: 'Beatmap Nominator'
+  bng_limited: 'Probationary Beatmap Nominator'
+  dev: 'Game Developer'
+  gmt: 'Global Moderation Team'
+  mapper: 'Beatmap Creator'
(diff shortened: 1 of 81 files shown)


resources/lang/en/beatmap_discussions.php
```diff
@@ -78,4 +78,11 @@
         'everyone' => 'Everyone',
         'label' => 'Filter by user',
     ],
+
+    'user_title' => [
+        'mapper' => 'Beatmap Creator',
+        'bng' => 'Beatmap Nominator',
+        'bng_limited' => 'Probationary Beatmap Nominator',
+        'nat' => 'Nomination Assessment Team Member',
```
VenixOSU: not sure if "member" is needed

resources/assets/coffee/react/beatmap-discussions/post.coffee
```diff
@@ -113,7 +113,9 @@ export class Post extends React.PureComponent
             div
               className: "#{bn}__user-badge"
               if userBadge?
-                div className: "user-group-badge user-group-badge--#{userBadge}"
+                div
+                  className: "user-group-badge user-group-badge--#{userBadge}"
+                  title: osu.trans "beatmap_discussions.user_title.#{userBadge}" if !@props.user.is_bot
```
notbakaneko: why no title for bot?

resources/lang/en/beatmap_discussions.php
```diff
@@ -78,4 +78,11 @@
         'everyone' => 'Everyone',
         'label' => 'Filter by user',
     ],
+
+    'user_title' => [
```
notbakaneko: missing all the other groups
https://github.com/ppy/osu-web/blob/8e2da62880bda8a163265b410666bf06a11da0d8/resources/assets/coffee/osu_common.coffee#L257

resources/assets/coffee/react/beatmap-discussions/post.coffee
TPGPL: i thought its self explanatory already

notbakaneko: If there's going to be a tooltip there, it should be for all of them

VenixOSU: to be honest, i think not having a tooltip for bot is reasonable, it's a special case and also tooltip would need to say same as badge what's quite dumb

TPGPL:
added a special bot tooltip which can give more context about it i think

nekodex:
BTW these tooltips shouldn't be specific to beatmap modding - the badges are going to be used globally so the tooltips need to also be global.

Also the group names should NOT be translated, they should be using whatever names they've been given in the database.

nekodex:
You're still using translations for the group name...

resources/lang/en/common.php
```diff
@@ -108,18 +108,6 @@
         'target' => 'drop here to upload',
     ],
 
-    'group_name' => [
-        'bot' => 'Automated Message',
```
VenixOSU: I'm pretty sure that "automated message" won't really work globally, would be better to replace it with something more universal so it doesn't fit on modding discussions (and possibly some other message stuff) only.

resources/assets/coffee/react/_components/user-badge.coffee
```diff
@@ -19,8 +19,19 @@
 import * as React from 'react'
 import { div } from 'react-dom-factories'
 bn = 'user-group-badge'
+group_names =
+  bot: 'Automated Message'
```
VenixOSU: I'm pretty sure that "automated message" won't really work globally, would be better to replace it with something more universal so it doesn't fit on modding discussions (and possibly some other message stuff) only.

peppy: also pretty sure this one will want to be localisable

TPGPL: so finally should those strings be localisable or not?

TPGPL:
i don't think i will be able to do that as i don't know typescript 🤷‍♂

notbakaneko:
This needs to be typescript

PR #4765, opened 2019-08-07
Add new column for beatmap discussion sorting order

nekodex:
Should stop editing/deleting/restoring posts from bumping discussions in 'last updated' view.

(also lays the groundwork for some future optimisations)


database/migrations/2019_08_07_070216_add_last_post_at_to_beatmap_discussions.php
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('beatmap_discussions', function (Blueprint $table) {
+            $table->timestamp('last_post_at')->nullable();
+            $table->index(['user_id', 'updated_at']);
```
nanaya: Upon closer inspection this probably doesn't belong here (wrong migration name etc). Also missing drop index.

PR #4769, opened 2019-08-08
Profile Integration of Modding Tab

nekodex:
Initial restyle/rewrite of the modding 'tab' on the user profile to actually look like part of the site instead of a placeholder page.

Known Issues:
- The 'see more' links currently just go to search pages instead of paginating in-page (Not sure if we actually want to paginate in-page though...?)
- ~~Changes from editing/deleting a post/discussion won't be reflected until the page is refreshed (fix coming)~~
- User Beatmaps are still shown on the main profile page instead of the modding for now, (probably makes sense to move them?)
- Some of the page structure is duplicated from the main user profile with slight modifications - depending on how the final modding profile ends up, possibly will be completely different (or can be combined/subclassed).


resources/assets/coffee/react/modding-profile/votes.coffee
```diff
@@ -0,0 +1,89 @@
...
+            className: "#{bn}__user-link"
+            href: laroute.route('users.modding.index', user: user.id) + '#votes'
+            span
+              className: "#{bn}__user-text u-ellipsis-overflow"
+              user.username
+
+        div
+          className: "#{bn}__user-badge"
+          if userBadge?
+            div className: "user-group-badge user-group-badge--#{userBadge}"
```
TPGPL: could reuse the element from #4756

nekodex: Sure, if it gets merged before this PR.

resources/assets/coffee/react/modding-profile/events.coffee
```diff
@@ -0,0 +1,75 @@
...
+                    div
+                      className: "beatmapset-event__content"
+                      dangerouslySetInnerHTML:
+                        __html: osu.trans "beatmapset_events.event.#{@typeForTranslation(event)}",
+                          'user': @props.users[event.user_id].username
+                          'discussion': (if discussionId then "<a href='#{discussionLink}'>##{discussionId}</a>" else '')
+                          'text': (if event.discussion then _.truncate(event.discussion.startingPost.message, {length: 100}) else '[no preview]')
+
+                    div
+                      className: 'beatmap-discussion-post__info'
```
nanaya: element without block

resources/assets/less/bem/modding-profile-vote-card.less
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  &__user {
+    flex: 1;
+    display: flex;
+    flex-direction: column;
+    align-items: flex-start;
+    justify-content: center;
+    margin: 0 5px;
+    width: 150px;
```
nanaya: this probably should be set in vote column instead? (275-150-avatar-stripe)px

PR #4783, opened 2019-08-11
Add beatmap information button to discussions

### LINKED ISSUE #4063: Add button to redirect to beatmap page on discussion page
example

Something like that, maybe? Cause placing it on title is hard to notice + no feedback to user if we hover it over the title that indicate it redirect us to beatmap page if we click on it

TPGPL:
Resolves #4063


TPGPL:
isn't it vertically centered by default though from the __icon class?

notbakaneko:
it should line up horizontally with the icon on the other button

PR #4785, opened 2019-08-13
Add various factories written for testing purposes

LiquidPL:
Adds a few model factories I've written for testing stuff.


database/factories/Multiplayer/EventFactory.php
```diff
@@ -37,3 +37,9 @@
         'text' => 'PART',
     ];
 });
+
+$factory->state(App\Models\Multiplayer\Event::class, 'game', function (Faker\Generator $faker) {
+    return [
+        'text' => 'test game',
```
nekodex: Should probably also create/associate a game with this event for it to be valid. Also `user_id` should be `null` for this event type.

database/factories/Multiplayer/GameFactory.php
```diff
@@ -0,0 +1,27 @@
...
+
+use Carbon\Carbon;
+
+$factory->define(App\Models\Multiplayer\Game::class, function (Faker\Generator $faker) {
+    $beatmap = App\Models\Beatmap::inRandomOrder()->first();
+
+    return [
+        'beatmap_id' => $beatmap->beatmap_id,
+        'start_time' => Carbon::now()->subSeconds($beatmap->total_length),
+        'play_mode' => $faker->numberBetween(0, 3),
```
nekodex: I'd just use the beatmap's playmode to ensure we're using a valid playmode for the beatmap

PR #4786, opened 2019-08-14
New forum listing style

nanaya:
Topic page itself is still bright light :sun_with_face:


resources/views/forum/forums/index.blade.php
```diff
@@ -16,53 +16,57 @@
...
 
-                <h1 class="osu-page-header__title">
-                    {{ trans("forum.title") }}
-                </h1>
-            </div>
-        </div>
-    </div>
+    <div class="osu-page osu-page--forum">
+        @foreach($forums as $category)
+            <div id="forum-{{ $category->forum_id }}" class="forum-list">
```
notbakaneko: needs offset?

PR #4790, opened 2019-08-15
Add sockets extensions to php docker image

LiquidPL:
Sending chat messages didn't work without it.


nanaya:
What's the error you're getting?

LiquidPL:
Well it does work, as in the message is created, but the post message endpoint returns a 500 with some datadog error. 

https://pastebin.com/qs2n0stY

PR #4807, opened 2019-08-19
LESS Styles code cleanup

jhdcruz:
### Changelogs
 - Fixed mismatch property-name.
 - Removed duplicate properties.
 - Shorthanded some duplicate/multiple properties.
 - Removed unit-identifier `px` to properties with `0` due to styling redundancy.

### DIFF
--- a/docs/css/style.css
+++ b/docs/css/style.css
@@ -722,7 +722,7 @@ body {
   transition-duration: 230ms;
 }
 .tocify-wrapper .tocify-focus {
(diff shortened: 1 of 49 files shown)


peppy:
Is there a reason this PR was deleted?

jhdcruz:
> Is there a reason this PR was deleted?

Outdated & haven't tested. Planning to make a new one, which is updated and tested, if I have more free time and finished the setup for this.

peppy:
ok, thanks for the clarification!

PR #4812, opened 2019-08-21
Remove country filtering for score rankings

notbakaneko:
country filtering is supposed to be only for performance ranking, apparently?

### DIFF
--- a/app/Http/Controllers/RankingController.php
+++ b/app/Http/Controllers/RankingController.php
@@ -70,7 +70,7 @@ class RankingController extends Controller
                 abort(404);
             }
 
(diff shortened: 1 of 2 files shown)


Natsukiii11:
i know this is probably way too late but may i ask why you removed this? i dont see any point in removing this, it's a feature that i personally used daily and im sure im not the only one. i dont see any positives in removing this feature so, id just like to know, thanks

peppy:
It was bringing our databases to a halt as it wasn't actually intended to be made available. If you enjoyed this feature, please open a feature request stating the use case you had for it and we will consider adding proper support.

PR #4815, opened 2019-08-21
Add users.read API scope

### LINKED ISSUE #4798: Add scope to users endpoints
Please add a scope to the `api/v2/users/{user}/` endpoints to make them accessible.

cl8n:
assuming these routes are ready to be made public

closes #4798

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -51,6 +51,13 @@ class UsersController extends Controller
 
         if (is_api_request()) {
             $this->middleware('require-scopes:identify', ['only' => ['me']]);
+            $this->middleware('require-scopes:users.read', ['only' => [
+                'beatmapsets',
+                'kudosu',
+                'recentActivity',
+                'scores',
+                'show',
+            ]]);
         }
 
         $this->middleware(function ($request, $next) {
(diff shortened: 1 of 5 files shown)


notbakaneko:
If it's for public-only data, I'm not sure it should be going through the third-party authorization grant flow? (and the scope would probably be called `public` or something)

Aergwyn:
If I had a bot that would use public endpoints would that just send client id and secret in the headers to associate the request? Completely skipping the authorization or how did you imagine it?

notbakaneko:
Client ID and secret are used to get the oauth token; how that token is obtained and the permissions granted depend on whether it's a third-party client acting on behalf of a user or not.

notbakaneko:
Closing this as this is not how we want the scopes and authorization to work; non-privileged public data should be using a public grant that doesn't allow the token to act as the use that authorized it.

nekodex:
To elaborate on what notbakaneko said:

Currently there's privileged information being returned by the API (e.g. account moderation info/metadata inside `/users/xxx` responses for users within the moderation group/etc) that we don't want third-party clients to get by default when granting basic "user read" permissions.

So we're gonna move out all the privileged info into separate permission group(s) first before we open up endpoints that could expose that data (so that for example, moderators don't accidentally leak privileged info to third-parties without realising).

PR #4820, opened 2019-08-24
Add overflowing property for basic beatmap stats

### LINKED ISSUE #1879: Avoid beatmap data overflowing off the page
https://osu.ppy.sh/beatmapsets/701416#mania/1484517
screen shot 2017-12-03 at 5 09 14 pm

TPGPL:
Resolves #1879 (temporary solution)


notbakaneko:
We decided that silly values shouldn't be shown instead #4819.

Also this just end up cutting of the display

PR #4821, opened 2019-08-24
Fix ranking stats getting cut on mobile views

### LINKED ISSUE #4795: "Global Ranking" and "Country Ranking" gets cut awkwardly in the new Modding panel when viewed on mobile devices
As suggested by the title; when I tried to open the new modding panel for my profile from a mobile device, I noticed that the "Global Ranking" and "Country Ranking" display was getting cut awkwardly in the bottom by the medals tab as shown in my screenshot(s) below.

I've tried to replicate it using various devices and browsers and the problem still persists, by the way :

a) Using Firefox Browser, on a smartphone

image

b) Using Firefox Browser, on a tablet

image

c) Using Puffin Browser, on a tablet

image

TPGPL:
Resolves #4795 

Fixes 2 problems:
- ranking stats touching border of the box

image

- level badge overflowing on smaller screens

image

---

### DIFF
--- a/resources/assets/coffee/react/modding-profile/detail-bar.coffee
+++ b/resources/assets/coffee/react/modding-profile/detail-bar.coffee
@@ -66,7 +66,7 @@ export class DetailBar extends React.PureComponent
 
         @renderExtraMenu()
 
-      div className: "#{bn}__column #{bn}__column--right",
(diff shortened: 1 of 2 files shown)


nekodex:
This is not the correct way to go about fixing the underlying problem, this just covers it up. See #4895.

PR #4836, opened 2019-08-27
Disallow deleting posts after a discussion is resolved

### LINKED ISSUE #4557: Remove ability to delete comments after resolution
how it will look like:
x9coswp
its confusing

notbakaneko:
closes #4557


resources/assets/coffee/react/beatmap-discussions/discussion.coffee
```diff
@@ -203,12 +204,20 @@ export class Discussion extends React.PureComponent
...
       lastEditor: @props.users[post.last_editor_id]
-      canBeEdited: @props.currentUser.is_admin || @isOwner(post)
+      canBeEdited: @props.currentUser.is_admin || canBeEdited
       canBeDeleted: canBeDeleted
       canBeRestored: canModeratePosts
       currentUser: @props.currentUser
 
 
+  resolvedSystemPostId: =>
+    if !@_resolvedSystemPostId?
```
nanaya: should be reset on update to posts array

app/Models/BeatmapDiscussionPost.php
```diff
@@ -163,6 +163,31 @@ public function user()
...
+            return false;
+        }
+
+        // The only system post type currently implemented is 'resolved', so we're making the assumption
+        // the next system post is always going to be either a resolve or unresolve.
+        // This will have to be changed if more types are added.
+        $systemPost = static::where('system', true)
+            ->where('id', '>', $this->id)
+            ->where('beatmap_discussion_id', $this->beatmap_discussion_id)
+            ->first();
```
nanaya: Shouldn't it find the last resolving post instead of earliest state change post?

1
2
3-resolve
4
5-unresolve
6
7-resolve
8
9-unresolve
10
11

In this case only 8+ should be editable, shouldn't it? With current check 4 is editable.

PR #4859, opened 2019-08-28
Add new forum index cover

nanaya:


notbakaneko:
The colour shift when the image isn't large enough is very jarring:
image

Issue #4759, opened 2019-08-07
Ability to sort friends based on rank


AqilCont:
I would like another option on the Friends page where you can also sort friends by Rank, as on stable and Lazer.
![image](https://user-images.githubusercontent.com/32044067/62588377-899d4700-b893-11e9-8fd5-3a533db1bbda.png)

peppy:
Sounds like an easy addition.

PR #4867, opened 2019-08-29
Add ability to sort friends by rank

venix12:
Adds an option to sort friends by default gamemode rank 
TODO: make eager-load work properly

fixes https://github.com/ppy/osu-web/issues/4759

### DIFF
--- a/app/Transformers/UserCompactTransformer.php
+++ b/app/Transformers/UserCompactTransformer.php
@@ -48,7 +48,7 @@ class UserCompactTransformer extends Fractal\TransformerAbstract
             'is_supporter' => $user->isSupporter(),
             'last_visit' => json_time($user->displayed_last_visit),
             'pm_friends_only' => $user->pm_friends_only,
-            'default_gamemode_rank' => $user->statistics($user->getPlaymodeAttribute($user->playmode), true)->rank ?? -1,
+            'default_gamemode_rank' => $user->getModeRank($user->getPlaymodeAttribute($user->playmode)) ?? -1,
         ];
     }
 
--- a/app/Transformers/UserCompactTransformer.php
+++ b/app/Transformers/UserCompactTransformer.php
@@ -48,7 +48,7 @@ class UserCompactTransformer extends Fractal\TransformerAbstract
             'is_supporter' => $user->isSupporter(),
             'last_visit' => json_time($user->displayed_last_visit),
             'pm_friends_only' => $user->pm_friends_only,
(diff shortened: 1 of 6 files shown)


app/Transformers/UserCompactTransformer.php
```diff
@@ -48,6 +48,7 @@ public function transform(User $user)
             'is_supporter' => $user->isSupporter(),
             'last_visit' => json_time($user->displayed_last_visit),
             'pm_friends_only' => $user->pm_friends_only,
+            'user_gamemode_rank' => $user->statistics($user->getPlaymodeAttribute($user->osu_playmode), 'true')->rank,
```
notbakaneko: `statistics` can return `null`

notbakaneko: `$user->playmode`

notbakaneko: `'true'` is a string not boolean; also this ends up being `false` in the statistics conditional

notbakaneko: this ends up being the rank for the user's selected default game mode, so the property name sounds misleading.

resources/assets/lib/user-list.tsx
```diff
@@ -44,6 +44,10 @@ function usernameSortAscending(x: User, y: User) {
   return x.username.localeCompare(y.username);
 }
 
+function rankSortDescending(x: User, y: User) {
+ return x.user_gamemode_rank > y.user_gamemode_rank;
```
notbakaneko: properties added to the type should be added to `globals.d.ts`

notbakaneko: `compareFunction` for `Array.sort` should return a `number`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

resources/lang/en/sort.php
```diff
@@ -26,6 +26,7 @@
     'old' => 'Old',
     'top' => 'Top',
     'username' => 'Username',
+    'rank' => 'Rank',
```
notbakaneko: keys should be alphabetical

resources/assets/lib/user-list.tsx
```diff
@@ -63,6 +67,8 @@ export class UserList extends React.PureComponent<Props> {
     switch (this.state.sortMode) {
       case 'username':
         return users.sort(usernameSortAscending);
+      case 'rank':
```
notbakaneko: case rank before username

VenixOSU:
Will work on eager-load soon

venix12:
would be better if i do this in a new pr i believe

PR #4885, opened 2019-09-03
New style for checkbox/radio button

nanaya:


resources/assets/less/bem/osu-switch-v2.less
```diff
@@ -16,53 +16,64 @@
...
+      border-radius: 0.5em;
+
+      &::after {
+        .fas();
+        content: '\f00c';
+      }
+    }
+
+    .@{_top}__input[type=radio] + & {
+      border-radius: 10000px;
```
nekodex: why not just 50%?

resources/assets/less/bem/osu-switch-v2.less
```diff
@@ -16,53 +16,64 @@
...
 
-    .@{_top}__input:checked ~ & {
-      visibility: visible;
+      &::after {
+        content: '';
+        background-color: currentColor;
+        border-radius: 10000px;
```
nekodex: as above

resources/views/accounts/_edit_notifications.blade.php
```diff
@@ -26,20 +26,15 @@
...
-                        >
-                        <span class="osu-checkbox__box"></span>
-                        <span class="osu-checkbox__tick">
-                            <i class="fas fa-check"></i>
-                        </span>
-                    </div>
+                    @include('objects._switch', [
+                        'attributes' => [
+                            'name' => 'user[user_notify]',
+                            'type' => 'checkbox',
```
nekodex: what do you think about just defaulting to `checkbox` with an option to turn it into radio? saves having to pass through `'type' = 'checkbox'` in all of the places

PR #4886, opened 2019-09-03
Use inter when using torus

nanaya:
Nothing is actually using it at the moment, so ┐(°～° )┌


resources/assets/less/functions.less
```diff
@@ -331,11 +331,16 @@
...
 
 .content-font() {
-  font-family: @font-content;
+  .content-font-base();
   line-height: 1.35;
   color: #444;
 }
 
+.content-font-base() {
+  font-family: @font-content;
```
nekodex: probably not necessary with the line below?

resources/views/forum/forums/index.blade.php
```diff
@@ -19,7 +19,7 @@
     'legacyNav' => false,
     'pageDescription' => trans('forum.title'),
     'searchParams' => ['mode' => 'forum_post'],
-    'useTorusFont' => true,
+    'use2019Font' => true,
```
nekodex: probably worth inverting this and having it be `useLegacyFont` or something instead?

resources/views/forum/forums/show.blade.php
```diff
@@ -23,7 +23,7 @@
         'mode' => 'forum_post',
     ],
     'titlePrepend' => $forum->forum_name,
-    'useTorusFont' => true,
+    'use2019Font' => true,
```
nekodex: as above

resources/views/master.blade.php
```diff
@@ -52,8 +52,9 @@ class="
     >
         <style>
             :root {
-                @if ($useTorusFont ?? false)
+                @if ($use2019Font ?? false)
```
nekodex: etc

PR #4927, opened 2019-09-11
Fix missing null check on modding profile page

### LINKED ISSUE #4925: Regular users unable to access one profile's modding tab
https://osu.ppy.sh/users/5099768/modding
Multiple regular users (5+) have reported not being able to see Net0's modding tab and are getting this instead ->
image

They are experiencing this problem on Net0's profile specifically. The theory is that something possibly went wrong with visibility permissions when he was unrestricted, as GMT and NAT are able to view the page, but regular users can not. We have already tried disabling extensions etc, but this made no difference. The users are able to view modding sections on other people's profiles just fine, but not Net0's.

nekodex:
Not actually directly related to permissions - GMTs/etc can see more event types which pushed the offending event(s) off the page.

fixes #4925


cl8n:
the permission checks in `ModdingHistoryController` aren't consistent across the pages either. maybe for another PR but that was something I discovered when looking into this broken kd event causing the problem in #4925 . the discussion is hidden from me on /modding, but I can see it just fine on /modding/votes-received

PR #4934, opened 2019-09-11
Reporting improvements

### LINKED ISSUE #4768: Add report button to modding discussions posts and replies
Would be good if simply formatted as a link like on comment sections.

like this
image
and put it here:
image

Currently lacking the options makes it basically required to manually report issues spotted to mods, which can be discouraging to reporting anything at all.

LiquidPL:
Fixes #4768.
Adds reporting of beatmap discussion posts. Also adds the reason dropdown to comment and score reporting, with reasons that make sense for each included.

---
* Added a report option to the beatmap discussion posts.
* Added the reason dropdown to comment and score reporting.


notbakaneko:
We're making changes to how reporting is handled, so I just made the changes directly to this branch

PR #4957, opened 2019-09-15
Fix chat switching to first channel on new message when new DM is opened

### LINKED ISSUE #4905: Chat will automatically redirect to #osu when open a new chat
When the new chat to a specified player is opened, when someone sent a new message in #osu, it will automatically jump to #osu. But this seems to be happened to new chat, existed chat doesn't affected.
 Here is a clip showing the issue: https://streamable.com/0rfmu

LiquidPL:
Fixes #4905.


peppy:
Please give a reason when closing PRs (for clarity).

LiquidPL:
notbakaneko opened #5034, so I assumed that this is the PR that's gonna go forward.

PR #4967, opened 2019-09-17
Fix social media icons alignment on landing page

notbakaneko:
~Not sure if should change the sizes to match the ones in the navigation header which have better horizontal alignment because of the sizes~
Now with  matching sizes


notbakaneko:
image
ಠ_ಠ

PR #4973, opened 2019-09-18
Update polymorphic name data

nanaya:
Should be done after ~~#4971~~ #5026 is deployed.


notbakaneko:
is this still bocked?

PR #4976, opened 2019-09-18
Fix bbcode newlines

nanaya:
Only remove up to one newline after the closing tag. This is already done for `[notice]` box but nothing else.


notbakaneko:
There should probably be additional test cases added to the parsing tests for this?

PR #4978, opened 2019-09-18
Hide deleted comments on the comments listing page

### LINKED ISSUE #4695: Hide deleted comments on the comments listing page
otherwise it will look like that 
image

LiquidPL:
Fixes #4695.

Currently without a toggle, but it should be fairly easy to add once it's needed.

---


app/Http/Controllers/CommentsController.php
```diff
@@ -74,7 +74,10 @@ public function index()
 
         $commentBundle = new CommentBundle(
             $commentable ?? null,
-            ['params' => request()->all()]
+            [
+                'params' => request()->all(),
+                'includeDeleted' => isset($commentable),
```
nanaya: this should just be set for non-json request instead of here.

app/Libraries/CommentBundle.php
```diff
@@ -120,10 +121,16 @@ public function commentIncludes()
...
     {
         if (isset($this->commentable)) {
-            return $this->commentable->comments();
+            $query = $this->commentable->comments();
         } else {
-            return Comment::select();
+            $query = Comment::select();
         }
+
+        if (!$this->includeDeleted) {
```
nanaya: this looks like should be in `getComments`.

nanaya:
Why is there `includeParent` stuff.

LiquidPL:
I messed up a merge.

PR #4983, opened 2019-09-19
Update bbcode style

nanaya:
- [x] Requires #4976 (though not quite directly related)


Magnus-Cosmos:
Is there a reason why SPOILER is added before box, I always assumed boxes didn't necessarily mean they were spoilers (since I've used them as drop downs to organize content, and I'm sure others have as well)

PR #5017, opened 2019-09-25
"Automatic" width for upload image button

### LINKED ISSUE #5016: "upload image" button icon is cut off in some languages
The "upload image" button can have the icon cut off because its text is too long in some languages.

In English (the icon looks ok):
English

In Italian (the icon looks cut off):
Italian

Raymonf:
This makes the upload image button grow to as large as it needs, but keeping a minimum size of `@avatar-size--base`.

It fixes #5016

This might cause issues with design consistency, though.

### DIFF
--- a/resources/assets/less/bem/btn-osu-big.less
+++ b/resources/assets/less/bem/btn-osu-big.less
@@ -51,8 +51,7 @@
   }
 
   &--account-edit {
-    display: inline-block;
-    max-width: @avatar-size--base;
+    min-width: @avatar-size--base;
     padding: 7px 10px;
     text-align: left;
   }
--- a/resources/assets/less/bem/btn-osu-big.less
+++ b/resources/assets/less/bem/btn-osu-big.less
@@ -51,8 +51,8 @@
   }
 
   &--account-edit {
-    display: block;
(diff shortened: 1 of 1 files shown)


resources/assets/less/bem/btn-osu-big.less
```diff
@@ -51,8 +51,8 @@
   }
 
   &--account-edit {
-    display: block;
-    width: @avatar-size--base;
+    display: inline-block;
+    max-width: @avatar-size--base;
```
notbakaneko: did you mean `min-width`

resources/assets/less/bem/btn-osu-big.less
```diff
@@ -51,8 +51,8 @@
   }
 
   &--account-edit {
-    display: block;
-    width: @avatar-size--base;
+    display: inline-block;
```
notbakaneko: the base block is already `display: inline-block`, so the entire line can be removed.

resources/assets/less/bem/btn-osu-big.less
Raymonf: I did... my nonexistent dyslexia strikes again

PR #5023, opened 2019-09-26
Add a toggle for hiding deleted forum posts

### LINKED ISSUE #4808: Add a toggle to show/hide deleted posts (Moderator-only)
Was about to do a post for https://github.com/ppy/osu-web/issues/3607 but as it seems to be solved already I've done a new issue.

Deleted posts are visible to moderators in regular forum threads, and there are plenty of extreme cases where people either quote posts (which isn't prefectly prevented yet) or just don't contribute anyhow to the thread, and it would be good if deleted posts can be toggled on and off for moderators too. The feature works for the comment section in news posts and also works for the modding section so it should potentially be possible to apply that for forum threads too.

As reference, cases like these aren't too uncommon
unknown (2)
unknown (1)
unknown

LiquidPL:
Fixes #4808.


app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -221,6 +221,7 @@ public function show($id)
         $postEndId = get_int(Request::input('end'));
         $nthPost = get_int(Request::input('n'));
         $skipLayout = Request::input('skip_layout') === '1';
+        $showDeleted = Request::input('show_deleted', 'true') === 'true';
```
nanaya: parameter name should be `with_deleted` to match the others.

resources/views/forum/topics/delete.blade.php
```diff
@@ -18,5 +18,11 @@
 @extends('forum.topics.replace_delete_button', ['countDifference' => -1])
 
 @section('moderatorAction')
-    $el.addClass("js-forum-post--hidden");
+    $deletedToggle = document.querySelector(".js-forum-topic-moderate--toggle-deleted");
+
+    if ($deletedToggle.dataset.showDeleted === "1") {
+        $el.addClass("js-forum-post--hidden");
+    } else {
+        $el.remove();
```
nanaya: this should use the same logic as the one in `replace_delete_button`.

resources/views/forum/topics/delete.blade.php
```diff
@@ -18,5 +18,11 @@
 @extends('forum.topics.replace_delete_button', ['countDifference' => -1])
 
 @section('moderatorAction')
-    $el.addClass("js-forum-post--hidden");
+    $deletedToggle = document.querySelector(".js-forum-topic-moderate--toggle-deleted");
```
nanaya: `$` prefix is for jquery collection

resources/views/forum/topics/delete.blade.php
```diff
@@ -18,5 +18,11 @@
...
+
+    if ($deletedToggle.dataset.showDeleted === "1") {
```
nanaya: can probably just use `forum.showDeleted()`

resources/assets/coffee/_classes/forum.coffee
```diff
@@ -197,6 +203,12 @@ class @Forum
     $(post).addClass('js-forum-post--highlighted')
 
 
+  toggleDeleted: =>
+    Turbolinks.visit osu.updateQueryString null,
+      show_deleted: if @showDeleted() then 'false' else 'true'
```
nanaya: this should use 0/1 like the others

LiquidPL:
Also I realized that the load more links don't update after fetching new posts, they still point to the posts that were initially loaded. Gonna make an issue for that.

resources/assets/coffee/_classes/forum.coffee
```diff
@@ -214,7 +224,12 @@ class @Forum
 
 
   postUrlN: (postN) ->
-    "#{document.location.pathname}?n=#{postN}"
+    url = "#{document.location.pathname}?n=#{postN}"
+
+    if @showDeleted()
```
nanaya: default is true already - with_default=0 is gone with this.

PR #5031, opened 2019-09-27
Add API endpoint to favourite/unfavourite a beatmapset

### LINKED ISSUE #4760: Add API endpoint to set user favourite status of a beatmap
We need this to update favourite value of beatmapset from the client. (Using default request requires web authentication)

notbakaneko:
closes #4760

was going to update and add the api doc for the list of favourites but realized there's more work to be done with where they're supposed to go and what's supposed to be returned, so that will be done separately.


nanaya:
Can use conflict resolve.

PR #5040, opened 2019-10-01
Fix chat not scrolling to the last message on load

### LINKED ISSUE #4897: Chat doesn't scroll to last unread (or bottom) on load
On load or switching chat channel, the scroll position is always the top. Should be bottom (or last unread).

LiquidPL:
Fixes #4897.

Still not sure if that's how this should be done, also we probably want to eventually track scroll position for each channel separately.


nekodex:
Yeah, so, I suggest you test this more.

This change makes _every_ incoming message forcibly trigger a scroll to the end of chat, making it impossible to scroll up and read the backlog.

notbakaneko:
superseded by #5404

PR #5048, opened 2019-10-02
Fix featured artist tracks sort order

nanaya:
MySQL doesn't support repositioning null values and laravel has been silently converting it to desc (and they changed it in 5.8 to throw error instead).


notbakaneko:
Could order by `display_order is null asc, display_order asc` to keep nulls at the end; depends how much we want nulls to be last  ┐(ツ)┌

nanaya:
Not sure if want to sort by boolean (which is helpfully converted into int by mysql).

PR #5058, opened 2019-10-03
Add a webhook to automate wiki updates

### LINKED ISSUE #2763: Decrease the wiki cache update time
Currently it takes too long to update the wiki. Should probably be changed into instant updating?

### LINKED ISSUE #3647: Add a new Wiki Admins group with an ability to force wiki updates
should be a temporary solution to #2763?

LiquidPL:
Fixes #2763. Closes #3647.

~~Adds a github webhook at `/wiki-update-webhook`, and a `WIKI_WEBHOOK_SECRET` env variable storing the webhook secret.~~ using the currently existing changelog webhook instead


app/Http/Controllers/WikiController.php
```diff
@@ -65,14 +65,25 @@ public function update($path)
     {
         priv_check('WikiPageRefresh')->ensureCan();
 
-        (new Wiki\Page($path, $this->locale()))->refresh();
+        (new Wiki\Page($path, $this->locale()))->forget();
 
         return ujs_redirect(Request::getUri());
     }
 
+    public function updateWebhook()
```
nanaya: Update the one in GithubImporter to handle this instead. And make sure of correct repository name, branch, event type, etc.

app/Libraries/OsuWiki.php
```diff
@@ -63,6 +65,24 @@ public static function fetchContent($path)
...
     }
 
+    public static function getUpdatedFiles($old, $new)
+    {
+        $diff = GitHub::repo()
+            ->commits()
+            ->compare(static::USER, static::REPOSITORY, $old, $new);
+
+        return array_filter($diff['files'], function ($file) {
+            return substr($file['filename'], 0, 4) === 'wiki';
```
nanaya: Use `starts_with`. Also add trailing slash (`wiki/`).

app/Jobs/UpdateWiki.php
```diff
@@ -0,0 +1,92 @@
...
+                $this->getObject($file['previous_filename'])->forget();
+            }
+
+            if ($status === 'removed' || $status === 'modified') {
+                $object->forget();
+            }
+
+            if ($status === 'modified') {
+                $object->get();
+            }
```
nanaya: this is a very long way to do

1. forget
2. forget previous_filename if renamed
3. get if not removed

app/Http/Controllers/WikiController.php
LiquidPL: Should this also use the currently existing webhook (ChangelogController@github) rather than this new one?

nanaya: yes

app/Jobs/UpdateWiki.php
```diff
@@ -0,0 +1,81 @@
...
+        // splits github path into wiki path, filename (locale in case of pages), file extension
+        preg_match('/^(?:wiki\/)(.*)\/(.*)\.(.{2,})$/', $path, $matches);
+
+        if (OsuWiki::isImage($path)) {
+            $path = $matches[1].'/'.$matches[2].'.'.$matches[3];
+
+            return new Image($path);
+        } else {
+            return new Page($matches[1], $matches[2]);
+        }
```
nanaya: This errors on `wiki/redirect.yaml` (or any file that's directly inside `wiki/`).

Probably should add handler for that as well.

And maybe something for news while at it since it's part of same repository and push event.

app/Jobs/UpdateWiki.php
```diff
@@ -0,0 +1,87 @@
...
+
+        foreach ($files as $file) {
+            $status = $file['status'];
+
+            $object = $this->getObject($file['filename']);
+
+            if ($object) {
+                // because otherwise we'd have newsposts being pointlessly removed and readded into the database
+                if (get_class($object) !== 'NewsPost' || (get_class($object) === 'NewsPost' && $status !== 'removed')) {
+                    $object->forget();
```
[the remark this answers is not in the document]
LiquidPL: Is this issue holding the PR back? Because I've looked into this, and:

* for wiki pages, both `forget()` and `get()` dispatch the respective jobs on the queue, and as far as I know they should be executed in the order they were dispatched,
* for images, the functions are completely synchronous, so it shouldn't be an issue at all.

Or do you mean that I should change the wiki page jobs to be handled synchronously in here?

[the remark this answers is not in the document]
nanaya: looks like what I observed was something slightly different. If the page previously indexed, the index will be deleted by `forget` but as it's done later in queue, when `get` is called, the page is still in index and thus isn't updated.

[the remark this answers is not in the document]
nekodex: BTW, I wouldn't assume that there will only ever be a single job worker running (and by extension that things will run in order).

If we ever need to scale up the number of workers in the future, jobs will be running concurrently and not necessarily in the order they were created.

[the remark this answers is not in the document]
LiquidPL: yeah, I see now, it was because `Page::get()` won't dispatch an index if a page is already in index

app/Libraries/WikiRedirect.php
```diff
@@ -35,19 +38,29 @@ public function normalizePath($path)
...
 
         return $this->cache['redirect'][$this->normalizePath($path)] ?? null;
     }
+
+    public function forget()
+    {
+        Cache::forget(static::CACHE_KEY);
+    }
+
+    public function get()
```
nanaya: the signature doesn't match WikiObject.

PR #5064, opened 2019-10-06
Update wiki.php

pimpG:
small translation

### DIFF
--- a/resources/lang/pt-br/wiki.php
+++ b/resources/lang/pt-br/wiki.php
@@ -41,4 +41,11 @@ return [
             'default' => 'Versão em inglês',
         ],
     ],
+    'main' => [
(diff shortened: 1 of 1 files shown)


TPGPL:
Thanks for your contribution, although all web translation changes are made through [Crowdin](https://crowdin.com/project/osu-web). so you should make changes there instead

pimpG:
this is not available to translate there it seems?
it shows pt-br translation is 100%

TPGPL:
it's because Crowdin hasn't been updated with the newest page changes (e.g. wiki main page) so there are no strings for that yet; you should wait till it gets updated and then suggest translations

peppy:
crowdin translations are updated on a weekly basis. if it's not available please wait until the next update.

PR #5070, opened 2019-10-07
Handle star difficulty per-mode when searching

### LINKED ISSUE #4545: Suggested difficulty doesn't work for hybrids
When searching by suggested difficulty for taiko/mania/catch (I can't test osu since i'm unranked lol), it will include hybrid maps that also meet the star rating requirement.

eg: My suggested difficulty is 6.89* in ctb (way too high but that's another issue), these maps will show up in my results
https://osu.ppy.sh/beatmapsets/819149
Standard mode: 6.74*, Catch: 5.43*
https://osu.ppy.sh/beatmapsets/703957
Standard mode: 7.05*, Catch: 4.16*

edit: When the standard diffs are converted to ctb, they are still only around 5* which shouldn't show up

notbakaneko:
- [x] Blocked until index updated with #5069

Updates the search queries to handle difficulties per-mode. The current behaviour just matches any play mode and sorts with the min/max of all the modes; this changes it to only use the selected mode and sort only based on that mode, using converted difficulties if included.

Closes #4545

Known issues: the beatmapset card still shows the difficulty of the original mode instead of the converted difficulty


nanaya:
conflicts :boom:

PR #5072, opened 2019-10-07
Give topics the option to allow double posting

### LINKED ISSUE #4537: Allow double posting in the "Report a Qualified beatmap here!" thread
When filing beatmap reports, it is more organised to make one post per beatmapset. This makes it clearer when there are multiple different maps being reported, instead of potentially mistaking two reports of two separate maps for one long report.

Issue is, you can't double-post in this thread on the new website easily. You can on the old forum. This is discouraging to users checking a larger amount of qualified maps and reporting multiple. While disallowing double-posting  makes sense for most of the forums, it does not in this use case.

**It should be allowed to double-post on the new website in this thread for any user.**

cl8n:
closes #4537

not sure if something temporary like this is wanted at all, figured I'd make a PR anyway since it only takes a minute. the idea is you'd set [this thread](https://osu.ppy.sh/community/forums/topics/447428)'s `topic_bumped` to 2 and then people can double post

### DIFF
--- a/app/Libraries/OsuAuthorize.php
+++ b/app/Libraries/OsuAuthorize.php
@@ -1236,7 +1236,7 @@ class OsuAuthorize
             return $prefix.'locked';
         }
 
-        if ($topic->isDoublePostBy($user)) {
+        if (!$topic->allowsDoublePosting() && $topic->isDoublePostBy($user)) {
             return $prefix.'double_post';
         }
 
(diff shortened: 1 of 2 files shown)


cl8n:
abraker says this is wanted on a per-forum level in addition to the qualified thread so I'll redo this with that in mind.

PR #5073, opened 2019-10-07
Centralize the wiki main page styling

LiquidPL:
As requested by wiki people when we were initially merging the main page, this makes it more sensible to add customized styling to elements outside of the main panels with links.


notbakaneko:
I don't see anything in ppy/osu-wiki#2656 that this requires?

LiquidPL:
Actually I think that's right, I just overthinked it. The wiki-side PR still requires this one, so I guess I just blocked them both for good measure. Should be good to go.

resources/assets/less/bem/wiki-main-page.less
```diff
@@ -17,14 +17,10 @@
...
 .wiki-main-page {
+  @_top: wiki-main-page;
   background: @osu-colour-b5;
 
   &__blurb {
-    font-size: 12px;
-    text-align: center;
-    font-weight: 300;
-    color: @osu-colour-c1;
-
```
notbakaneko: I don't think this needs to be moved

resources/assets/less/bem/wiki-main-page.less
```diff
@@ -17,14 +17,14 @@
  */
 
 .wiki-main-page {
+  @_top: wiki-main-page;
```
notbakaneko:
```suggestion
```

resources/assets/less/bem/wiki-main-page.less
```diff
@@ -38,4 +38,30 @@
...
+    color: @osu-colour-c1;
+    font-size: 20px;
+    font-weight: 300;
+    font-style: normal;
+
+    .wiki-main-page-panel & {
+      margin-bottom: 40px;
+    }
+
+    .wiki-main-page-panel:first-child & {
```
notbakaneko: should be a modifier that categorizes why it's `font-size` should be different instead of `:first-child`

PR #5100, opened 2019-10-11
Fix color of beatmap search text

### LINKED ISSUE #5099: beatmap searching input text color unreadable
My issue is that the text color for when you are searching beatmaps is very hard to read. [See this picture](https://i.imgur.com/Kbcz6oU.png) Its very hard to distinguish the difference between the background color and the actual text color. while the placeholder text is [perfectly 100% readable.](https://i.imgur.com/p1yDNUS.png) I'd like to propose that the actual search text be the same color as the placeholder text for readability. I believe the color of the placeholder text is #757575

I tried to search through the osu-web code to find out where it inherits the color to propose a pull request myself... but I can't seem to find it.

nanaya:
Resolves #5099.

### DIFF
--- a/resources/assets/less/bem/beatmapsets-search.less
+++ b/resources/assets/less/bem/beatmapsets-search.less
@@ -97,6 +97,8 @@
     font-style: italic;
     font-weight: lighter;
 
+    color: @osu-colour-b3;
+
     &:focus {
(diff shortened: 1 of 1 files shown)


nl-tatatat:
Awesome job nanaya!

Issue #4784, opened 2019-08-11
Send notifications to all QAT when an issue is reported on a qualified beatmap


peppy:
Right now a forum thread is being used, which is all sorts of backwards.

This should address the issues raised in #4536.

ascendancelol:
Can this be sorted by mode? As in, could I only receive notifications for reports of my mode (catch) and not all modes? Seems much better than the report thread either way if it goes right to my notifs.

dorothy3242:
Reposting thoughts from the other issue plus some other stuff

there needs to be something that
- makes reported maps visible in one place
- ^ ideally alongside a summary of the problem that the reporter would submit
- keeps it intuitive to report maps for issues (easy to know the option exists)
- Should be able to alert **BNs and NATs alike** (QAT doesn't exist)
- Would automatically clear the reports for disqualified maps.

There should also be some way for BN/NAT to clearly mark if they feel a report is invalid. It doesn't need to remove the report, but it would make it clear a report has already been looked at so others can focus on other unchecked issues.

I think something like github reviews could work quite well. You "start a report", post multiple issues as part of it, and it shows up as one thing with a summary for others to easily look at.

TheMoMan:
Some functional requirements
- Notifications must be sent when a new issue has been reported.
- Notifications should be seperated by gamemode, so that relevant members can react to beatmaps of the appropriate mode.
- NAT must receive notifications from at least one gamemode.
- BN must have the ability to opt into receiving notifications for any gamemode.
- There should be an API for updates to this system (let discord bots repost updates etc)

Naxesss:
If reports are only made when people post a problem stamp, https://github.com/ppy/osu-web/issues/4405 is a prerequisite so that nominators can report beatmaps like anyone else when they're unsure of something and not have it be disqualified.

For the last point Mo mentioned, see https://github.com/ppy/osu-web/issues/4042.

Context:
We currently use a webhook to highlight any NAT or BN that opts-in when a beatmapset of their mode is posted in the report thread. When a beatmapset is disqualified or ranked, we remove the entry so that anyone looking at the channel can see what's relevant. Allowing this to track problems (and possibly suggestions) means we could adapt this accordingly. This information (who did the report, what map it was on, etc) is also added to a database and is something we use to judge Beatmap Nominator activity.

![](https://i.imgur.com/qllpBrH.png)

notbakaneko:
[...]

PR #5101, opened 2019-10-11
Send notification to NATs and BNs when a problem is posted on a qualified beatmap

notbakaneko:
requires
- [x] ppy/osu-notification-server#24

First step for #4784.

Just sends a notification to BNs and NATs when a problem is posted on a qualified beatmapset. There are no options for it yet.

### DIFF
--- a/app/Events/BeatmapsetDiscussionQualifiedProblemNotificationEvent.php
+++ b/app/Events/BeatmapsetDiscussionQualifiedProblemNotificationEvent.php
@@ -24,22 +24,25 @@ use App\Models\Notification;
 use Illuminate\Broadcasting\Channel;
 use Illuminate\Queue\SerializesModels;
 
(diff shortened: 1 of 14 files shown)


app/Events/BeatmapsetDiscussionQualifiedProblemNotificationEvent.php
```diff
@@ -0,0 +1,69 @@
...
+ *    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace App\Events;
+
+use App\Models\Notification;
+use Illuminate\Broadcasting\Channel;
+use Illuminate\Queue\SerializesModels;
+
+class BeatmapsetDiscussionQualifiedProblemNotificationEvent extends NotificationEventBase
```
nanaya: nothing in this seem beatmap discussion specific?

PR #5103, opened 2019-10-11
Add basic opengraph support

TPGPL:
Adds Opengraph usage support. Requirement for future opengraph PRs, so posting this now.

---


resources/views/layout/metadata.blade.php
```diff
@@ -28,6 +28,20 @@
...
+    <meta property="og:url" content="{{ Request::url() }}">
+    <meta property="og:title" content="{{ $opghTitle }}">
+    <meta property="og:description" content="{{ $opghDescription }}">
+
+    @if (isset($opghImage))
+        <meta property="og:image" content="{{ $opghImage }}">
+        <meta property="og:image:width" content="128">
+        <meta property="og:image:height" content="128">
+    @endif
+@endif
```
nanaya: Instead of using `$hasOpengraph` etc just make it

```
$opengraph = [
  'title' => ...,
  'image' => ...,
];
```

resources/views/layout/metadata.blade.php
```diff
@@ -28,6 +28,20 @@
 <meta name="keywords" content="osu, peppy, ouendan, elite, beat, agents, ds, windows, game, taiko, tatsujin, simulator, sim, xna, ddr, beatmania, osu!, osume">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
 
+@if ($hasOpengraph)
+    <meta property="og:site_name" content="{{ 'osu! » '.$opghSitename ?? 'osu!' }}">
```
nanaya: I think it should always be 'osu!'.

resources/views/layout/metadata.blade.php
```diff
@@ -28,6 +28,20 @@
...
+    <meta property="og:type" content="website">
+    <meta property="og:url" content="{{ Request::url() }}">
```
nanaya: this should be canonical url (e.g. `https://osu.ppy.sh/users/1` even when the user accesses `https://osu.ppy.sh/users/1?random=parameter`).

resources/views/layout/metadata.blade.php
```diff
@@ -28,6 +28,20 @@
...
+    <meta property="og:title" content="{{ $opghTitle }}">
+    <meta property="og:description" content="{{ $opghDescription }}">
+
+    @if (isset($opghImage))
+        <meta property="og:image" content="{{ $opghImage }}">
+        <meta property="og:image:width" content="128">
+        <meta property="og:image:height" content="128">
```
nanaya: first these attributes are optional and second nothing on current system has such image size

resources/views/layout/metadata.blade.php
TPGPL: it should use `issset($opengraph)` instead of `$hasOpengraph` then, right?

resources/views/layout/metadata.blade.php
TPGPL: wouldnt it be less misleading if for title "Ben Briggs", it would show `osu! » featured artists`?

resources/views/layout/metadata.blade.php
```diff
@@ -28,6 +28,18 @@
 <meta name="keywords" content="osu, peppy, ouendan, elite, beat, agents, ds, windows, game, taiko, tatsujin, simulator, sim, xna, ddr, beatmania, osu!, osume">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
 
+@if (isset($opengraph))
+    <meta property="og:site_name" content="{{ 'osu! » '.$opengraph['site_name'] ?? 'osu!' }}">
+    <meta property="og:type" content="website">
+    <meta property="og:url" content="{{ strtok(Request::url(), '?') }}">
```
nanaya: This should use `$canonicalUrl` (also used elsewhere) and check in the `isset` above.

resources/views/layout/metadata.blade.php
```diff
@@ -28,6 +28,18 @@
...
+    <meta property="og:title" content="{{ $opengraph['title'] }}">
+    <meta property="og:description" content="{{ $opengraph['description'] }}">
```
nanaya: This is optional. And should use `$pageDescription` (also used elsewhere).

resources/views/layout/metadata.blade.php
```diff
@@ -28,6 +28,18 @@
...
+
+    @if (isset($opengraph['image']))
```
nanaya: shouldn't have this check as `og:image` is required

resources/views/layout/metadata.blade.php
```diff
@@ -28,6 +28,18 @@
 <meta name="keywords" content="osu, peppy, ouendan, elite, beat, agents, ds, windows, game, taiko, tatsujin, simulator, sim, xna, ddr, beatmania, osu!, osume">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
 
+@if (isset($opengraph))
+    <meta property="og:site_name" content="{{ 'osu! » '.$opengraph['site_name'] ?? 'osu!' }}">
```
nanaya: this doesn't work (`'x'.$a ?? 'y'`, the `??` is for the entire `'x'.$a`, not just `$a`). Use a separate php block instead. And `site_name` for the key sounds a bit weird (and not quite exactly used as is either). `section` is probably fine.

resources/views/layout/metadata.blade.php
```diff
@@ -28,6 +28,24 @@
...
+    @endif
+
+    <meta property="og:type" content="website">
+
+    @if (isset($canonicalUrl))
+        <meta property="og:url" content="{{ $canonicalUrl }}">
+    @endif
+
+    <meta property="og:title" content="{{ $opengraph['title'] }}">
+    <meta property="og:description" content="{{ $pageDescription }}">
```
nanaya: this is optional and check if `$pageDescription` exists instead

resources/views/layout/metadata.blade.php
```diff
@@ -28,6 +28,24 @@
...
+@if (isset($opengraph))
+    @if (isset($opengraph['section']))
+        <meta property="og:site_name" content="{{ 'osu! » '.$opengraph['section'] }}">
+    @else
+        <meta property="og:site_name" content="{{ 'osu!` }}">
+    @endif
+
+    <meta property="og:type" content="website">
+
+    @if (isset($canonicalUrl))
```
nanaya: this is not optional so it should be part of the check above (or no check at all so whoever adding opengraph to pages will notice this)

resources/views/layout/metadata.blade.php
```diff
@@ -28,6 +28,24 @@
 <meta name="keywords" content="osu, peppy, ouendan, elite, beat, agents, ds, windows, game, taiko, tatsujin, simulator, sim, xna, ddr, beatmania, osu!, osume">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
 
+@if (isset($opengraph))
+    @if (isset($opengraph['section']))
+        <meta property="og:site_name" content="{{ 'osu! » '.$opengraph['section'] }}">
+    @else
+        <meta property="og:site_name" content="{{ 'osu!` }}">
```
nanaya: the content doesn't need to be inside `{{ }}`. And it's not even syntactically correct. Or make a `@php` block to set the correct `$siteName` and reduce duplicating the same property.

nanaya:
Continued at #5245.

PR #5111, opened 2019-10-15
Include comment message markdown in response

notbakaneko:
Missing from #4872; otherwise there's no markdown to parse


nanaya:
Probably should look into way of not sending both html and markdown for everything.

PR #5123, opened 2019-10-16
Replace register link with download

nanaya:


fmang:
@nanaya Are there alternatives to in-game registrations?

Some people can’t run osu! because of poor hardware, but can run alternative osu! implementations. Still, they need an osu! account to download beatmaps.

peppy:
If you can't run osu! then you don't need beatmaps.

PR #5128, opened 2019-10-16
Allow moderation of profile page

### LINKED ISSUE #4980: Add GMT/NAT members the ability to edit/lock people's userpages
Currently GMT/NAT members are unable to edit anyone's userpage besides their own, which is something very useful to partially delete content. We can't do it through old forum since it returns a non-existent file.

nanaya:
Resolves #4980.

- [x] Log to banlist thingy

### DIFF
--- a/app/Http/Controllers/AccountController.php
+++ b/app/Http/Controllers/AccountController.php
@@ -193,21 +193,6 @@ class AccountController extends Controller
         return $user->defaultJson();
     }
 
-    public function updatePage()
-    {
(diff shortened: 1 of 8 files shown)


peppy:
If this is to be merged, please make it admin-only. We can't have any old moderator able to edit another user's content.

peppy:
Never mind, let's keep things in-line with forum editability and change them both at once if we deem that more correct.

nanaya:
Oh well ┐(ツ)┌

PR #5131, opened 2019-10-16
Don't return full html for missing wiki images

nanaya:


notbakaneko:
should set Content-Type to `image/jpeg` and return blank imo

nanaya:
I'm just copying generic nginx behavior. Also pretty sure blank body is an invalid jpeg.

PR #5135, opened 2019-10-17
Fix the views when trying to PM yourself

notbakaneko:
You already can't create a new conversation with yourself anyway.

also fixes the test failure in #5071 when checking the routes

### DIFF
--- a/resources/assets/lib/chat.ts
+++ b/resources/assets/lib/chat.ts
@@ -19,7 +19,6 @@ import { PresenceJSON, SendToJSON } from 'chat/chat-api-responses';
 import MainView from 'chat/main-view';
 import * as _ from 'lodash';
 import Channel from 'models/chat/channel';
-import User from 'models/user';
 import core from 'osu-core-singleton';
 
 const dataStore = core.dataStore;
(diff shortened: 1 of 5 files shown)


resources/assets/lib/chat.ts
```diff
@@ -35,28 +34,26 @@ reactTurbolinks.register('chat', MainView, () => {
...
   if (!_.isEmpty(sendTo)) {
-    const target: User = dataStore.userStore.getOrCreate(sendTo.target.id, sendTo.target); // pre-populate userStore with target
-    let channel: Channel | null = dataStore.channelStore.findPM(target.id);
+    const target = dataStore.userStore.getOrCreate(sendTo.target.id, sendTo.target); // pre-populate userStore with target
+    let channel = dataStore.channelStore.findPM(target.id);
 
     if (channel) {
       initialChannel = channel.channelId;
-    } else {
+    } else if (!target.is(sendTo.target)) {
```
nanaya: Not sure in what case `target` - which is from `getOrCreate(...id, sendTo.target)` - is not `sendTo.target` ?_?

PR #5137, opened 2019-10-17
Limit notification of new problem on qualified maps

nanaya:
Only notify new problem if there's no currently open problems.

The whole thing should probably be moved out of controller.

### DIFF
--- a/app/Http/Controllers/BeatmapDiscussionPostsController.php
+++ b/app/Http/Controllers/BeatmapDiscussionPostsController.php
@@ -106,7 +106,9 @@ class BeatmapDiscussionPostsController extends Controller
     {
         $discussion = $this->prepareDiscussion(request());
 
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/BeatmapDiscussionPostsController.php
```diff
@@ -135,18 +137,34 @@ public function store()
...
 
+        $notifyQualifiedProblem = false;
+
+        if (!$disqualify && $discussion->beatmapset->isQualified() && $discussion->message_type === 'problem') {
+            $openProblems = $discussion
+                ->beatmapset
+                ->beatmapDiscussions()
+                ->ofType('problem')
+                ->where(['resolved' => false])
+                ->count();
```
notbakaneko: probably shouldn't count deleted discussions

PR #5143, opened 2019-10-18
Include preview in modding discussion notifications

### LINKED ISSUE #5139: More descriptive modding discussion notifications
Related to https://github.com/ppy/osu-web/issues/1925

image
saying "there's a problem" or "username posted on map" is practically no info at all. Watched beatmap comments actually preview the message, why not for modding discussion notifications? it would make it much easier to tell what the different notifications for one map is as well

without a preview, showing individual notifications for one map is redundant and pointless because it just says the same thing again and again.

Also, responses to problem posts on qualified maps are showing as "problem posted on qualified map" when that's totally inaccurate, a response to a problem is not the same as someone posting a problem. 
Like if person A posts a problem, then the mapper responds with a good reason why they won't change, that shows as two "problem posted on qualified map" notifications which is wrong and misleading!

notbakaneko:
closes #5139

also makes problems their own item instead of lumping them together with discussion post notifications


app/Jobs/BroadcastNotification.php
```diff
@@ -139,6 +139,7 @@ private function onBeatmapsetDiscussionPostNew()
         $this->receiverIds = static::beatmapsetReceiverIds($this->notifiable);
 
         $this->params['details'] = [
+            'content' => $this->object->message,
```
nanaya: this probably should be truncated (also the other one)

resources/lang/en/notifications.php
```diff
@@ -30,14 +30,18 @@
...
-                'beatmapset_discussion_qualified_problem_compact' => 'Problem on qualified beatmap',
+                'beatmapset_discussion_post_new' => 'New post on ":title" by :username: ":content"',
+                'beatmapset_discussion_post_new_compact' => 'New post by :username: ":content"',
                 'beatmapset_discussion_unlock' => 'Discussion on ":title" has been unlocked',
                 'beatmapset_discussion_unlock_compact' => 'Discussion was unlocked',
             ],
 
+            'beatmapset_problem' => [
+                '_' => 'Beatmap problem',
+                'beatmapset_discussion_qualified_problem' => ':username reported a problem ":title": ":content"',
```
nanaya: "on qualified beatmap"?

resources/assets/lib/notification-maps/category.ts
```diff
@@ -37,7 +37,7 @@ export function categoryGroupKey(item: Notification) {
 export const nameToCategory: CategoryMap = {
   beatmapset_discussion_lock: 'beatmapset_discussion',
   beatmapset_discussion_post_new: 'beatmapset_discussion',
-  beatmapset_discussion_qualified_problem: 'beatmapset_discussion',
+  beatmapset_discussion_qualified_problem: 'beatmapset_problem',
```
nanaya: need new category icon

resources/assets/lib/notification-maps/message.ts
```diff
@@ -23,7 +23,7 @@ export function messageCompact(item: Notification) {
   let message: string;
 
   const replacements = {
-    content: item.details.content,
+    content: item.details.content || '',
```
nanaya: I wonder if it's better saying no preview or something for those missing data. (also the other one)

notbakaneko: I'm not sure it'd make sense as the text would show up in between quotes, and I don't think it's worth adding special handling since it'll only show for notifications before the change that haven't been read.

We should also probably look at making things like content, image and url always be required when creating notifications?

PR #5154, opened 2019-10-21
Add a file to gitignore

LiquidPL:
It's been appearing in my repo for quite some time now.


nanaya:
Update the one in that directory instead

nanaya:
Not related to this pull request but there has been some talk on changing the default clockwork logging to sqlite (and you probably want to do that too so you won't have a directory potentially filled with thousands of json files).

PR #5157, opened 2019-10-22
Add an ability to pin comments

### LINKED ISSUE #5138: Add the option to mods to pin/highlight comments
Pin/highlight comments (whatever you wanna call it) was on disqus and helped mods to give a short note, redirect people to other places as needed, etc. So it can be helpful for mods to give a notice about whats happening in a comment section for everyone to look at.

LiquidPL:
Resolves #5138.


app/Transformers/CommentTransformer.php
```diff
@@ -53,6 +53,8 @@ public function transform(Comment $comment)
 
             'edited_at' => json_time($comment->edited_at),
             'edited_by_id' => $comment->edited_by_id,
+
```
notbakaneko:
```suggestion
```

routes/web.php
```diff
@@ -98,6 +98,9 @@
 Route::post('comments/{comment}/vote', 'CommentsController@voteStore')->name('comments.vote');
 Route::delete('comments/{comment}/vote', 'CommentsController@voteDestroy');
 
+Route::post('comments/{comment}/setPinned', 'CommentsController@setPinned')->name('comments.pin');
```
notbakaneko:
```suggestion
Route::post('comments/{comment}/pin', 'CommentsController@pin')->name('comments.pin');
```

routes/web.php
```diff
@@ -98,6 +98,9 @@
...
+Route::delete('comments/{comment}/setPinned', 'CommentsController@setPinned');
```
notbakaneko:
```suggestion
Route::delete('comments/{comment}/pin', 'CommentsController@pin');
```

app/Http/Controllers/CommentsController.php
```diff
@@ -250,6 +251,27 @@ public function update($id)
         return CommentBundle::forComment($comment)->toArray();
     }
 
+    public function setPinned($id)
```
notbakaneko:
```suggestion
    public function pin($id)
```

app/Libraries/OsuAuthorize.php
```diff
@@ -972,6 +972,17 @@ public function checkCommentVote(?User $user, Comment $comment) : string
         return 'ok';
     }
 
+    public function checkCommentPin(?User $user, Comment $comment) : string
+    {
+        $this->ensureLoggedIn($user);
+
+        if ($user->isAdmin()) {
+            return 'ok';
+        }
```
venix12: seems like this isn't needed as ``doCheckUser()`` always returns ``ok`` if user is admin

resources/assets/lib/models/comment.ts
```diff
@@ -76,6 +78,11 @@ export class Comment {
     return currentUser.is_admin || currentUser.can_moderate;
   }
 
+  @computed
```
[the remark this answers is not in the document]
notbakaneko: it sure does; also seems pointless

PR #5171, opened 2019-10-24
Refactor account playstyles setting

nanaya:
Update AccountEdit to also handle the task of AccountEditPlaystyle.

PR #5172, opened 2019-10-24
Notification option for new problem on qualified beatmapset

nanaya:
Depends on:
- [x] #5171

Some problems:
- [x] the model and table don't seem right but I'm not sure of any better way
  - I don't quite like the one row per option per user but I can't make one row per user with either using type text for column or one big json blob
  - the handling of details seems like will get pretty ugly pretty fast
  - will need some kind of cleanup for rows with empty details
    - no index so it will be fun (json can't be indexed)
- [x] pretty sure the wording for setting label need improvement


app/Http/Controllers/AccountController.php
```diff
@@ -174,6 +185,28 @@ public function updateEmail()
...
+    public function updateNotificationOptions()
+    {
+        $request = request();
+
+        $name = $request['name'] ?? null;
+        $params = get_params($request, 'user_notification_option', ['details:any']);
+
+        $option = auth()->user()->notificationOptions()->firstOrCreate(['name' => $name]);
+
+        if ($option === null) {
```
notbakaneko: `!$option->exists`? `firstOrCreate()` isn't going to return `null`

resources/views/accounts/_edit_notifications.blade.php
```diff
@@ -42,5 +42,42 @@
...
+                    class="account-edit-entry__checkboxes js-account-edit"
+                    data-account-edit-auto-submit="1"
+                    data-account-edit-type="array"
+                    data-url="{{ route('account.notification-options', [
+                        'name' => 'new_problem_on_qualified_beatmapset',
+                    ]) }}"
+                    data-field="user_notification_option[details]"
+                >
+                    @php
+                        $modes = $notificationOptions['new_problem_on_qualified_beatmapset']['modes'] ?? [];
```
notbakaneko: missing `['details']`?

resources/lang/en/accounts.php
```diff
@@ -67,6 +67,7 @@
     'notifications' => [
         'title' => 'Notifications',
         'topic_auto_subscribe' => 'automatically enable notifications on new forum topics that you create',
+        'beatmapset_discussion_qualified_problem' => 'notification for new problem on qualified beatmaps',
```
notbakaneko: `Receive notifications for new problems on qualified beatmaps` or something

nanaya: I think what the options are for need explaining as well :pensive:

PR #5180, opened 2019-10-25
Allow guest access for beatmap discussion listing pages

### LINKED ISSUE #5159: Discussion and reply routes are not yet public
https://osu.ppy.sh/beatmapsets/beatmap-discussions
https://osu.ppy.sh/beatmapsets/beatmap-discussion-posts

For guests, similar pages such as https://osu.ppy.sh/beatmapsets/events and https://osu.ppy.sh/comments are visible, but not the above. Am assuming it's because they are wip. 

Making these public would be useful for hooking up beatmap reports to discord so bns can be notified beyond the site itself (we had this before the thread got locked). This is important to minimize the risk of reports being missed and beatmaps needing unranks. BNs dealing with reports aren't on the site 24/7, but they are on discord, so it's the ideal notification platform.

As it looks currently I'll probably need to use sessions, but with them expiring it's not a reliable solution. Some api endpoint for events in general would be great (an issue for this already exists), but I'm guessing it would take longer to implement than just making these pages public, and nat+bns would like this back up asap.

nanaya:
Resolves #5159 if it's going to be a thing.

### DIFF
--- a/app/Http/Controllers/BeatmapDiscussionPostsController.php
+++ b/app/Http/Controllers/BeatmapDiscussionPostsController.php
@@ -39,7 +39,7 @@ class BeatmapDiscussionPostsController extends Controller
 
     public function __construct()
     {
-        $this->middleware('auth');
+        $this->middleware('auth', ['except' => 'index']);
 
         return parent::__construct();
     }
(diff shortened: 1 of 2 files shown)


Ephemeralis:
Any word on when this can go in? It's a roadblock for relay notification via discord bot at the moment, which has been a long requested feature from the BN team.

peppy:
Keep in mind this is my an API endpoint and not intended for not usage.

PR #5183, opened 2019-10-25
Add configuration for client-bundled beatmapsets

### LINKED ISSUE #5170: Move bundled beatmap IDs to env
As we will be updating them on a regular basis, it would be nice to have these in the .env file.

https://github.com/ppy/osu-web/blob/master/app/Models/Beatmapset.php#L155

nanaya:
Resolves #5170.

env entry is `BEATMAPSET_CLIENT_BUNDLE` with default copied from current hardcoded list.


peppy:
production env updated

PR #5189, opened 2019-10-27
Change Subtitle on osu!wiki Main Page

Circleen:
Peppy hated the old one, so I tried to replace it.

### DIFF
--- a/resources/lang/en/wiki.php
+++ b/resources/lang/en/wiki.php
@@ -42,6 +42,6 @@ return [
     ],
     'main' => [
         'title' => 'Knowledge Base',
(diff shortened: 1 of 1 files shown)


bdach:
I don't mean to be disparaging, but the suggested subtitle feels... redundant? As in it's restating the *Knowledge Base* title itself. I'm probably the furthest thing you could get from a copy-writer, so I don't feel great suggesting something alternative, but some of the other discord suggestions in `#osu-web` sounded ok.

peppy:
I opened an issue to discuss this. Please stop making PR.s

PR #5195, opened 2019-10-28
Add filter options for beatmap discussions listing

notbakaneko:
Adds 
- the ability to filter beatmap modding discussions based on whether the beatmap has been qualified or disqualified.
- an option to show only unresolved discussions


resources/lang/en/beatmap_discussions.php
```diff
@@ -38,9 +38,19 @@
         'form' => [
             '_' => 'Search',
             'deleted' => 'Include deleted discussions',
+            'only_unresolved' => 'Show only unresolved discussions',
             'types' => 'Message types',
             'username' => 'Username',
 
+            'status' => [
+                '_' => 'Status',
```
nanaya: "beatmap status"?

resources/views/beatmap_discussions/index.blade.php
```diff
@@ -63,6 +83,16 @@ class="simple-form__input"
                     </div>
                 </div>
 
+                <div class="simple-form__row simple-form__row--no-label">
```
nanaya: the padding seems a bit off

app/Models/BeatmapDiscussion.php
```diff
@@ -139,6 +155,41 @@ public static function search($rawParams = [])
         return ['query' => $query, 'params' => $params];
     }
 
+    public static function statusQuery($query, $status)
```
nanaya: This is a weird way to do scope.

app/Models/BeatmapDiscussion.php
```diff
@@ -124,6 +125,20 @@ public static function search($rawParams = [])
             $params['message_types'] = array_keys(static::MESSAGE_TYPES);
         }
 
+        $params['status'] = static::getValidStatus($rawParams['status'] ?? null);
```
nanaya: looks like this and some other "status"es should be called "beatmapset status"? Otherwise it looks like it's about status of the discussion...

PR #5198, opened 2019-10-29
Move medal icons to assets

### LINKED ISSUE #5194: Dynamic resources should probably be stored outside of git
Medals being the main one here, should be moved to the assets store rather than needing to be committed each time.

nanaya:
Resolves #5194.

- [x] set `USER_ACHIEVEMENT_ICON_PREFIX` (or maybe have a better default)
- [x] copy assets to new location
- [x] nginx rewrite for the moved files (for notification)
    ```
    location ^~ /images/badges/user-achievements/ {
      rewrite ^.*/([^/]+)$ https://...$1;
    }
    ```

After deployment in tinker (takes a bit long):
```php
App\Models\Notification::where(['name'=>'user_achievement_unlock'])->chunkById(1000, function($notifications) {
    foreach ($notifications as $notification) {
        $achievement = App\Models\Achievement::find($notification->details['achievement_id'] ?? null);
        if ($achievement !== null) {
            $notification->update(['details' => array_merge($notification->details, ['cover_url' => $achievement->iconUrl()])]);
        }
    }
});
```
(or make a tinker command?)


peppy:
Probably not worth having a command, will only ever be used once.

Ephemeralis:
```https://assets.ppy.sh/medals/web/``` now points to the exact same files present in ```/public/images/badges/user-achievements/``` on osu-web.

```https://assets.ppy.sh/medals/client/``` now points to the medals-client directory on osu-web-10 which contains enlarged exports of said medals for use explicitly in the client.

PR #5209, opened 2019-10-29
Add support for running browser tests in docker

LiquidPL:
Adds a selenium container to the docker setup, and an example dusk config that should work with it.

Also adds `docker-compose.override.yml` to gitignore since I happen to be using it for something.

Also there's an issue with this setup, mainly that you need to regenerate js routes before and after tests since the app url changes, and I'm not exactly sure where should I put that.


notbakaneko:
This isn't a direction we're looking at pursuing at this time unless the issues mentioned can be addressed

PR #5222, opened 2019-11-01
Truncate large numbers in beatmap stats

### LINKED ISSUE #1879: Avoid beatmap data overflowing off the page

TPGPL:
Resolves #1879

---


resources/assets/coffee/osu_common.coffee
```diff
@@ -214,7 +214,7 @@
 
 
   formatNumberSuffixed: (number, precision, options = {}) ->
-    suffixes = ['', 'k', 'm', 'b', 't']
+    suffixes = ['', 'k', 'm', 'b', 't', 'qa', 'qi', 'sx', 'sp', 'oc', 'no']
```
peppy: what the fuck

bdach: I think these new ones are supposed to stand for `quadrillion`, `quintillion`, `sextillion`, ..., `nonillion`, but yeah, these don't seem to be standardised in any way, so probably shouldn't be used. (Also are numbers that large even showing up anywhere?)

TPGPL: the issue requires at least quintillion in the suffixer, added more as some beatmaps might have even bigger numbers, can delete some if thats a problem

cl8n: maybe add an option to default to infinity symbol if it's beyond trillions

peppy: I like the infinity idea, if that's feasible (would need to hide the number itself too).

TPGPL: i could go with the infinity char if its not considered as a workaround, the original BPM will still be shown in a tooltip

PR #5234, opened 2019-11-04
Add wiki notice for articles in need of cleanup or rewrite

cl8n:
osu version of https://en.wikipedia.org/wiki/Template:Cleanup_rewrite. `outdated` will override this since it'd be a little redundant

I feel like these messages should be managed on osu-wiki but that's a discussion for some other time :running_man:


LiquidPL:
To be honest I feel like we should implement at least a rudimentary template system for it.

app/Models/Wiki/Page.php
```diff
@@ -227,6 +227,11 @@ public function isOutdated() : bool
         return $this->page()['header']['outdated'] ?? false;
     }
 
+    public function needsCleanup() : bool
+    {
+        return $this->page()['header']['needs_cleanup'] ?? false;
```
notbakaneko:
```suggestion
        return $this->page['header']['needs_cleanup'] ?? false;
```

cl8n: `page` is only a function here isn't it

notbakaneko: `page()` has been removed, it's `page` now

PR #5236, opened 2019-11-05
Add client verification support

### LINKED ISSUE #2862: Add osu! client verification support
(Sorry if this is known/a duplicate - I tried search)

When logging in (first time) with the old osu! client we need to authorize our accounts, that means we are redirected to the old website which will be getting removed soon. I noticed osu!lazer doesn't require authorization however, so this may not be an issue (unless that feature is not yet implemented there).

LiquidPL:
Resolves #2862.

The confirmation dialog could probably use some design work though.
image

The verification route sits at `/home/account/verify-client?client_hash=<hash goes here>`.


nanaya:
Closing this as implementation for new web will be way simpler: just a form with button - with the involved session must be verified (using normal verification flow).

PR #5242, opened 2019-11-05
Add RSS for news

nanaya:
Opengraph related stuff will be added later.


resources/views/news/index-rss.blade.php
```diff
@@ -0,0 +1,44 @@
...
+                <title>{{ $post->title() }}</title>
+                <link>{{ route('news.show', $post->slug) }}</link>
+                <guid isPermaLink="false">{{ atom_id('news', $post->getKey()) }}</guid>
+                <pubDate>{{ $post->published_at->toRfc7231String() }}</pubDate>
+                <description>{!! $post->previewText() !!}</description>
+                @if ($post->firstImage() !== null)
+                    <enclosure
+                        url="{{ $post->firstImage(true) }}"
+                        length="300"
+                        type="image/jpeg"
```
cl8n: should try to infer the type, these aren't guaranteed to be jpeg (looks like already a few are png and one is gif)

nanaya: As it turned out it's not even useful.

app/Models/NewsPost.php
```diff
@@ -204,9 +204,23 @@ public function editUrl()
         return 'https://github.com/'.OsuWiki::USER.'/'.OsuWiki::REPOSITORY.'/tree/master/news/'.$this->filename();
     }
 
-    public function firstImage()
+    public function firstImage($absolute = false)
```
nanaya: as turned out this isn't needed for this PR but it'll be needed for opengraph stuff, so... ┐(°～° )┌

camellirite:
this is neat but is there a link to it that im missing? i want to have a neat little newsfeed on my desktop using rainmeter

LiquidPL:
https://osu.ppy.sh/home/news?format=rss

PR #5247, opened 2019-11-06
Fix country ranking API

### LINKED ISSUE #4794: Unable to get country rankings using via API
I'm trying to implement api for rankings in osu!, but here's a problem: it works fine for scores and performance, but country returns `null`. And error was Internal server error.
Request I was using: `https://osu.ppy.sh/api/v2/rankings/osu/country`

notbakaneko:
fixes #4794


peppy:
Is it possible to enforce `break` in `switch` via CI? Seems like it'd be worthwhile.

nanaya:
Is that an interesting way of asking to create test for this...?

bdach:
Tests are nice, but not allowing case fall-through would eliminate this entire class of bugs entirely, and it seems like something static analysis could do? (I might be wrong, never touched php in my life)

nanaya:
It's a valid php construct. There just isn't test for this endpoint (for some combination of parameters and data).

peppy:
I understand it's valid. Was a proposal to disallow usage of it as it is quite prone to human error (in c# they outright disallow it for this reason). If there aren't many usages it might be better for code quality to also disallow in osu-web.

nanaya:
There is no such thing it seems.

PR #5248, opened 2019-11-06
Create helper to make using group colours less tedious.

nekodex:
Figured it was worthwhile after seeing how many places we've duplicated this logic in.

Not sure about whether the group names should live in `user-group-badge.less` or not though...(should eventually stop being defined in CSS anyway... right?)

### DIFF
--- a/resources/assets/less/bem/user-group-badge.less
+++ b/resources/assets/less/bem/user-group-badge.less
@@ -27,87 +27,67 @@
   justify-content: center;
   align-items: center;
 
-  &--alumni {
-    color: @osu-colour-group-alumni;
-    &::before {
-      content: 'OLD';
+  // mapping group to display-name
+  .osu-group-names() {
+    @alumni: {
+      full: 'ALUMNI';
+      short: 'OLD';
+    }
+    @bng: {
+      full: 'BN';
+      short: 'BN';
+    }
+    @bng_limited: {
+      full: 'BN';
+      short: 'BN';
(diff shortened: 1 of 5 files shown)


resources/assets/less/bem/user-group-badge.less
```diff
@@ -27,87 +27,67 @@
...
-
-  &--mapper {
-    color: @osu-colour-blue-1;
-    &::before {
-      content: 'MAP';
-
-      @media @desktop {
-        content: 'MAPPER';
-      }
+    @mapper: {
```
nanaya: this isn't in the group list so it's not getting any styling...

resources/assets/less/bem/user-group-badge.less
```diff
@@ -27,87 +27,67 @@
...
+      full: 'MAP';
+      short: 'MAPPER';
```
peppy: full/short intended to be this way around?

resources/assets/less/bem/user-group-badge.less
```diff
@@ -27,87 +27,67 @@
...
-    color: @osu-colour-group-ppy;
-    &::before {
-      content: 'PPY';
+    @ppy: {
+      full: 'PPY';
+      short: 'PPY';
+    }
+    @support: {
+      full: 'SPT';
+      short: 'SUPPORT';
```
peppy: also here

PR #5254, opened 2019-11-07
Add a bundled filter to beatmapset listing

### LINKED ISSUE #5169: Website listing of beatmaps that can be bundled with the client
As a search feature/filter
After seeing [this comment thread](https://osu.ppy.sh/comments/956151)  , I think it would also be helpful to have a permanent list of maps that can be bundled with the client easily viewable on the website.

This way if people have different bundled maps and want to easily find the ones missing, or if someone reinstalls the game on a new pc and wants to refind their nostalgic bundled maps again, the website would help them do so.

image

could work as an "extra" filter option.

LiquidPL:
Fixes #5169.

### DIFF
--- a/app/Libraries/Search/BeatmapsetSearch.php
+++ b/app/Libraries/Search/BeatmapsetSearch.php
@@ -72,6 +72,7 @@ class BeatmapsetSearch extends RecordSearch
         $this->addGenreFilter($query);
         $this->addLanguageFilter($query);
         $this->addExtraFilter($query);
+        $this->addBundledFilter($query);
         $this->addStatusFilter($query);
 
         $nested = new BoolQuery;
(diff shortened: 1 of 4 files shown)


app/Libraries/Search/BeatmapsetSearch.php
```diff
@@ -129,6 +130,15 @@ private function addExtraFilter($query)
         }
     }
 
+    private function addBundledFilter($query)
+    {
+        if ($this->params->showBundled) {
+            $query->filter(['ids' => [
+                'values' => config('osu.beatmapset.client_bundle'),
```
notbakaneko: this setting is the ids of the current client bundle; the proposal was for all bundled maps, i.e. including maps previously bundled but no longer bundled - this should be stored in the database.

This setting is also no longer used nor updated and will be removed

notbakaneko:
as mentioned here https://github.com/ppy/osu-web/issues/5169#issuecomment-561737008, the list should be coming from the DB, but that list doesn't exist yet

PR #5255, opened 2019-11-07
Groundwork for beatmap discussion 'reviews'

nekodex:
This starts laying down some of the groundwork required for beatmap discussion 'reviews':
 - Adds a Markdown renderer (limited to 'review' discussion posts and only bold/italic for now)
 - Extends Markdown parsing/rendering to support 'embedded' discussions within reviews
 - Restructures how discussions are rendered so we can use a different layout for reviews and normal discussion posts
 - (as a bonus, this also updates the modding history results to look more like the actual discussion posts)

Known issues that will be fixed in later PRs:
 - Modding profile pages may not have the required discussions loaded to show the embeds properly (will just show a generic 'discussion missing' message for now).
 - Modding history results will show posts as unparsed/plaintext.
 - Reviews come up under `General (All difficulties)` currently instead of in their own tab
 - Issues/embeds aren't linked with their review (and vice-versa) yet

Review posts can't be created and aren't visible unless the `BEATMAPSET_DISCUSSION_REVIEW_ENABLED` environment variable is set. (don't do this in production yet)

For those playing at home, here's a little preview of how things are shaping up:


resources/assets/lib/beatmap-icon.tsx
```diff
@@ -0,0 +1,51 @@
...
+interface Props {
+  beatmap: Beatmap;
+  modifier?: string;
+  overrideVersion?: boolean;
+  showTitle?: boolean;
+}
+
+export const BeatmapIcon: FunctionComponent<Props> = ({beatmap, overrideVersion, showTitle, modifier}) => {
+  const difficultyRating = overrideVersion || BeatmapHelper.getDiffRating(beatmap.difficulty_rating);
+  const showTooltip = showTitle || !overrideVersion;
```
nanaya: this seem different to the original

PR #5256, opened 2019-11-07
Wiki Redirect library cleanup

nanaya:
- redirect is cached forever (with its own `cached_at` attribute)
- added `needsSync` and `sync` just like `NewsPost`
- fixed exception catcher class name
- conflict party :tada:  (with #5253)


peppy:
needs minor conflict resolution

PR #5268, opened 2019-11-09
Shorten licence header messages

### LINKED ISSUE #5267: Reduce size of licence header
While AGPLv3 recommends adding the full text as we are doing to every file it is overly cumbersome. This is luckily only a recommendation.

image

A minimal licence header - in line with our licence headers in other projects - is enough to satisfy their minimum criteria.

Proposal:

```
```

TPGPL:
Resolves #5267 

---


peppy:
Thanks, but if the proposal is to be agreed, I will handle this myself (along with providing a script to perform the update for people with in-flight PRs). As it is a proposal, a PR is not required for now.

PR #5270, opened 2019-11-09
Update medal achievement event to use transformer for detail

### LINKED ISSUE #5269: Missing Achivement icon on the Recent section
This is probably a regression of https://github.com/ppy/osu-web/commit/32c8c3193cad0385790a343e58e7218a421c8d3c
Example of Quick Draw achievement of recent section on my profile (https://osu.ppy.sh/users/6326135)
image
`srcset` attribute of the img tag is set as `srcset="undefined 1x, undefined 2x"`

nanaya:
Fixes #5269.

### DIFF
--- a/app/Models/Event.php
+++ b/app/Models/Event.php
@@ -275,10 +275,7 @@ class Event extends Model
         }
 
         return [
-            'achievement' => [
-                'slug' => $achievement->slug,
-                'name' => $achievement->name,
-            ],
+            'achievement' => json_item($achievement, 'Achievement'),
             'user' => $this->arrayUser($matches),
         ];
     }


app/Models/Event.php
```diff
@@ -275,10 +275,7 @@ public function parseMatchesAchievement($matches)
         }
 
         return [
-            'achievement' => [
-                'slug' => $achievement->slug,
-                'name' => $achievement->name,
-            ],
+            'achievement' => json_item($achievement, 'Achievement'),
```
peppy: is the second field just for display purposes? if so, medal is probably better

```suggestion
            'achievement' => json_item($achievement, 'Medal'),
```

PR #5275, opened 2019-11-11
Require verification to post or edit modding discussion post

notbakaneko:
~it _will_ cause the page to refresh and clear anything that's been entered into the text field after verification~

- [x] #5278


notbakaneko:
superseded by #5282

PR #5282, opened 2019-11-12
Require verification for all write actions

nanaya:
- [x] depends on #5279
- [x] fix tests
  - [x] add `asVerifiedUser` helper
  - [x] requires #5285
- [x] fix oauth authorization flow


peppy:
Is it possible that this be an environment variable toggle?

notbakaneko:
set `USER_POST_ACTION_VERIFICATION` in env

PR #5294, opened 2019-11-15
Add relationship between reviews and their child issues

nekodex:
Discussion UI is also updated to accommodate, with links going back and forth.

- [x] depends on #5255, merge and update this PR/branch first

PR #5297, opened 2019-11-15
Add tab for beatmap discussion reviews

nekodex:
Also hides the tab if reviews are not enabled (via the `.env` var)

- [x] depends on #5294, merge and then update this first


nanaya:
Just noticed the generated url for review page shouldn't have beatmap id in it.

PR #5312, opened 2019-11-21
Faster removal of existing scores when beatmapset gets ranked

notbakaneko:


peppy:
Considering that even when using the right index, when i ran manually it took around 80 minutes to run (250k scores), this alone likely isn't going to be enough.

Should we be increasing the job timeout?

notbakaneko:
The previous `order by score_id` would have made it a lot slower than it should have been since the index wouldn't have been in the same order.

peppy:
The above time was after changing the query for efficiency. It was still pretty slow just due to the sheer number of scores. We'll try 1 hour and see if it happens again.

PR #5329, opened 2019-11-23
Add rank sorting to the user listing

### LINKED ISSUE #4759: Ability to sort friends based on rank

venix12:
Fixes #4759

Adds the current user's mode rank filter to the user listing.


app/Http/Controllers/FriendsController.php
```diff
@@ -51,13 +51,17 @@ public function __construct()
 
     public function index()
     {
-        $friends = Auth::user()
+        $currentUser = Auth::user();
+        $currentMode = ucfirst($currentUser->playmode);
```
nanaya: should use `studly_case`

app/Transformers/UserCompactTransformer.php
```diff
@@ -70,6 +72,11 @@ public function includeCover(User $user)
         });
     }
 
+    public function includeCurrentModeRank(User $user)
+    {
+        return $this->primitive($user->statistics(Auth::user()->playmode)->rank ?? false);
```
nanaya: - check `Auth::user()` isn't null. Also maybe use `auth()` instead of `Auth` to skip an import
- `null` instead of `false`
- `rank` is an old unused column; use `globalRank()` instead

resources/assets/lib/sort.tsx
```diff
@@ -40,7 +40,12 @@ export class Sort extends React.PureComponent<Props> {
           key={value}
           onClick={this.props.onSortSelected}
         >
-          {osu.trans(`sort.${value}`)}
+          {value === 'rank'
+            ? <span>
+                <i className={`fas fa-extra-${currentUser.playmode}`} /> {osu.trans('sort.rank')}
```
nanaya: should be `fa-extra-mode-` (`fa-extra-` is the old icon)

resources/assets/lib/user-list.tsx
```diff
@@ -40,6 +40,14 @@ interface State {
   viewMode: ViewMode;
 }
 
+function rankSortDescending(x: User, y: User) {
+  if (x.current_mode_rank && y.current_mode_rank) {
```
nanaya: Should not really affect anything but `!= null` check avoids javascript casting fun.

app/Transformers/UserCompactTransformer.php
venix12: hmm other transformers are using an import but i can use ``auth()`` if it's really a deal ¯\\\_(ツ)_/¯

nanaya: newer codes generally use `auth()`, I think.

app/Transformers/UserCompactTransformer.php
```diff
@@ -70,6 +72,11 @@ public function includeCover(User $user)
         });
     }
 
+    public function includeCurrentModeRank(User $user)
+    {
+        return $this->primitive($user->statistics(Auth::user()->playmode ?? 'osu')->globalRank() ?? null);
```
nanaya: `null->x() ?? null` will still throw error.

resources/assets/lib/sort.tsx
```diff
@@ -40,7 +40,12 @@ export class Sort extends React.PureComponent<Props> {
           key={value}
           onClick={this.props.onSortSelected}
         >
-          {osu.trans(`sort.${value}`)}
+          {value === 'rank'
+            ? <span>
+                <i className={`fas fa-extra-mode-${currentUser.playmode || 'osu'}`} /> {osu.trans('sort.rank')}
```
nanaya: use `??` instead of `||`.

venix12: ``??`` won't work in JS, ``||`` is a proper way to do this i believe

app/Transformers/UserCompactTransformer.php
venix12: i tested and i believe it won't

nanaya: have you tested with someone without stats?

venix12: yeah

nanaya: that's weird, this is what I'm seeing

nanaya: have you checked this again? I'm still seeing error (see above).

venix12: been quite busy, can do today most likely

venix12: oh yeah, you are right, i accidentally didn't change ``rank`` to ``globalRank()`` in my testing environment

PR #5332, opened 2019-11-24
Uncap pp precision

pavlukivan:
I made this PR because of [this post](https://www.reddit.com/r/osugame/comments/e0xoel/mangomizer_sets_a_4994998_pp_score_but_it_says/)

SR is uncapped in API responses, so I think pp shouldn't be capped either


mangomizer:
Actually, wasn't SR capped recently as well? Like I remember a time when we could get SR to 12 decimal places, now it's capped to 6 s.f.

peppy:
4 decimal places is already more than enough. Strongly against this as being a step backwards.

cl8n:
not sure if you meant to "fix" the thing in the reddit post, but this doesn't do it. 500pp is just 499.5 rounded up and that's not a truncated value
https://github.com/ppy/osu-web/blob/61ae2ce5670b203cad287755d3a776f74df16cbd/app/Transformers/ScoreTransformer.php#L57

notbakaneko:
These aren't even used, it won't do anything

PR #5334, opened 2019-11-24
Update country names

cl8n:
Musty suggested "North Macedonia" in #osu-web a few days ago. the [Prespa agreement](https://en.wikipedia.org/wiki/Prespa_agreement) changed the country's name in February

changed some others along the way

- [x] Update `osu_countries`
- [x] Check lazer / stable hardcodings


bdach:
Maybe if we're doing these sorts of changes, might be worth looking at more recent changes that took place? Going off [wikipedia](https://en.wikipedia.org/wiki/Timeline_of_historical_geopolitical_changes#2010s):

* Swaziland is now named Eswatini
* The Czech Republic has proclaimed that the preferred English form is now "Czechia"
* Cape Verde wants to be called Cabo Verde even in English (we already have Côte d'Ivoire and Timor-Leste instead of the English-sounding variants)
* South Sudan separated from Sudan
* The Netherlands Antilles no longer exist, now split into Curaçao and Sint Maarten (that one's going to be a special headache if there are any users from there...)

cl8n:
I'll do the simple renames here. should be another issue for splitting/making countries

fwiw, South Sudan and Kosovo(another new country you didn't mention) already have flag images, SS.png and XK.png, but neither are in the database or used for registrations

peppy:
Just to be sure, this is only affecting the seeder right? So it will not be visible anywhere and requires database (and lazer codebase) changes?

notbakaneko:
This only changes the seeder, so I left it open for reference

peppy:
Has been updated everywhere.

PR #5336, opened 2019-11-25
Disable login for accounts with empty email addresses

notbakaneko:
Also moved the `logout` function to a helper


app/Http/Controllers/SessionsController.php
```diff
@@ -67,7 +67,7 @@ public function store()
     public function destroy()
     {
         if (Auth::check()) {
-            $this->logout();
```
nanaya: there's another call in `Account\SessionsController`

PR #5340, opened 2019-11-25
Reset inactive user when logged in from different country

nanaya:
- [x] Depends on #5343.


tests/Controllers/SessionsControllerTest.php
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+namespace Tests\Controllers;
+
+use App\Models\User;
+use Tests\TestCase;
+
+class SessionsControllerTest extends TestCase
+{
+    public function testCreate()
```
notbakaneko: I don't think that's what's being tested...

PR #5344, opened 2019-11-26
Record more things on various login attempts

nanaya:
That was weird. Slightly lenient password match check.

- [x] check this actually works
- [x] write tests (probably needs #5343)
- [x] depends on #5340


app/Models/LoginAttempt.php
```diff
@@ -35,27 +36,136 @@ class LoginAttempt extends Model
...
+    {
+        return optional($user)->getKey() ?? 0;
+    }
+
+    public static function hashInvalidPassword($password)
+    {
+        // The goal is just to allow vaguely matching password - for
+        // example when trying same password - to be excluded from being
+        // counted as additional attempt.
+        return substr(sha1('osu_unique_'.md5($password)), 8, 12);
```
notbakaneko: But vaguely matching passwords would result in completely different hashes..?

nanaya: as in there might be another password resulting same "hash"

notbakaneko: Only same password and collisions would result in the same result, and the collisions won't even be from similar values since the password is being hashed? The chance of a vaguely matching password returning the same result might as well be non-existent

nanaya: it's not "vaguely matching password" but "vaguely matching" password (as in just because it matches doesn't mean it's the same) :dancer:

notbakaneko: it's not allowing anything then ಠ_ಠ

PR #5354, opened 2019-11-29
Make user disabled page crowdin-translatable

nanaya:
And add enchant for the page while at it? Though the wording will be weird if it's disabled.

- [x] Updated to depend on #5361


resources/views/users/disabled.blade.php
```diff
@@ -19,6 +19,40 @@
...
+        </p>
+
+        <ul>
+            <li>
+                {!! trans('users.disabled.reasons.tos._', [
+                    'community_rules' => tag('a', [
+                        'href' => osu_url('user.rules'),
+                    ], trans('users.disabled.reasons.tos.community_rules')),
+                    'tos' => tag('a', [
+                        'href' => osu_url('legal.terms'),
```
notbakaneko: I think it's `route('legal', 'terms')` now

Issue #5358, opened 2019-12-01
[Question] Why doesn't "generate-localizations.js" use fs to make the directory?


GingkathFox:
Line: https://github.com/ppy/osu-web/blob/5b413ac15f9383c442409c502fbb2d111ae171e4/generate-localizations.js#L73

`fs` has `fs.mkdirSync`, why not use that instead of requiring a second module?

cl8n:
mkdirp also makes any parent directories that are missing

GingkathFox:
```
fs.mkdir('/tmp/a/apple', { recursive: true }, (err) => {
  if (err) throw err;
});
```
> The optional options argument can be an integer specifying mode (permission and sticky bits), or an object with a mode property and a recursive property indicating whether parent folders should be created. Calling fs.mkdir() when path is a directory that exists results in an error only when recursive is false.

notbakaneko:
because `recursive: true` is a `node.js 10.12+` feature and we weren't running 10.12+ at the time.

GingkathFox:
oh, ok. Opening a pull request then.

PR #5359, opened 2019-12-01
Updated to use `fs.mkdir` module

GingkathFox:
Updated the localization file to use the fs mkdir module. Also, maybe instead of a `for` loop you could use a `forEach`? I don't know much about maps, if yall are busy i'll figure it out.

Closes https://github.com/ppy/osu-web/issues/5358

- [x] update servers' nodejs to version 12


GingkathFox:
Fixed. Thanks for reminding me :D

PR #5362, opened 2019-12-02
Crowdin-compatible mails

nanaya:
- [x] depends on #5361
- [x] check if anything breaks
- [x] move mail-related translations to its own file

Now I think about it, we probably should centralize all the email translations so we can exclude it when exporting to javascript :thinking:


resources/lang/en/beatmapset_watches.php
```diff
@@ -32,12 +32,18 @@
...
+
+    'update_notice' => [
+        'email' => [
+            'content' => [
+                'opening' => 'Just letting you know that there has been a new update in beatmap ":title" since your last visit.',
+                'unwatch' => 'If you no longer wish to watch this beatmap, you can either click the "Unwatch" link found in the page above, or from the modding watchlist page:',
+                'visit' => 'Visit the discussion page here:',
+            ],
+        ],
+    ],
```
notbakaneko: delete?

resources/lang/en/forum.php
```diff
@@ -91,6 +87,16 @@
...
+    'reply' => [
+        'email' => [
+            'content' => [
+                'link' => 'Jump straight to the latest reply using the following link:',
+                'new_reply' => 'Just letting you know that there has been a new reply in ":title" since your last visit.',
+                'unsubscribe' => 'If you no longer wish to watch this topic, you can either click the "Unsubscribe topic" link found at the bottom of the topic above, or from topic subscriptions management page:',
+            ],
+        ],
+    ],
+
```
notbakaneko: delete?

PR #5368, opened 2019-12-04
Move group identifier to database

nanaya:
This moves group identifier to id (?) mapping from being hardcoded in php to database. Also pass more group detail to transformer. Preload group details on request to avoid extra queries.

No client side update yet.

- [x] merge #5369 first and fill in the new columns
- [x] figure out why dusk is failing :pensive: 

Table: `phpbb_groups`

- `identifier`: the key that corresponds to `group_id` in [UserGroup](https://github.com/ppy/osu-web/blob/ff8618daece03a1099346512196606993b4a0144/app/Models/UserGroup.php#L39)
- `short_name`: "short" values in [user-group-badge](https://github.com/ppy/osu-web/blob/ff8618daece03a1099346512196606993b4a0144/resources/assets/less/bem/user-group-badge.less) with matching identifier (`OLD` should probably be `ALM` or something)
- `display_order`: integers which value the order in [UserGroup](https://github.com/ppy/osu-web/blob/ff8618daece03a1099346512196606993b4a0144/app/Models/UserGroup.php#L39)
- `colour`: values in [here](https://github.com/ppy/osu-web/blob/ff8618daece03a1099346512196606993b4a0144/resources/assets/less/colors.less#L152), variables should be resolved to the actual values (except `var(--blah)`s)

The ones without matching values can be left blank.


notbakaneko:
Is this still waiting for the columns to be filled?

nanaya:
Added some hints on how to update the table :thinking:

...and fixed conflicts :frowning:

resources/assets/lib/quick-search/user.tsx
```diff
@@ -49,7 +49,7 @@ export default function User({ user }: { user: User }) {
 
         {user.group_badge == null ? null : (
           <div className='user-search-card__col user-search-card__col--icon'>
-            <div className={`user-group-badge user-group-badge--${user.group_badge}`} />
+            <div className={`user-group-badge user-group-badge--${user.group_badge.identifier}`} />
```
notbakaneko: renders empty badge if in default group or any other group without a badge

nanaya: not if group_type is set correctly in database :thinking:

PR #5369, opened 2019-12-04
Add columns needed for database-sourced group id

nanaya:
See #5368.


peppy:
Should be fine to run this one live whenever. Small table.

Issue #4896, opened 2019-09-05
Opening a chat notification should dismiss all related notifications


smoogipoo:
A notification pops up for each message a user sends you. If the notification for one message is clicked, the rest remain active, which feels really bad.

peppy:
This is the same deal for most notification types. Was supposed to be revisited but has not been yet.

PR #5384, opened 2019-12-06
Add support for marking whole notification stacks as read

notbakaneko:
Allows entire notification stacks to be marked as read in order to support the fix for #4896
The format for this (endpoint is still the same) is 
```ts
{
  object_type: string;
  object_id?: number;
  category?: string;
}
```

~~- [ ] database index needs to be added on `is_read`~~ didn't seem to have any effect on the query ?_?
- [x] unread count is wrong when an entire stack is marked as read


app/Http/Controllers/NotificationsController.php
```diff
@@ -173,4 +183,48 @@ private function endpointUrl()
 
         return $url;
     }
+
+    private function markReadByIds($ids)
+    {
+        $user = auth()->user();
+        if ($user->userNotifications()->whereIn('notification_id', $ids)->update(['is_read' => true])) {
+            event(new NotificationReadEvent($user->getKey(), $ids));
```
nanaya: might as well pass `['ids'=>$ids]` for `$params` and remove the second argument?

app/Http/Controllers/NotificationsController.php
```diff
@@ -173,4 +183,48 @@ private function endpointUrl()
...
+
+            return response(null, 204);
+        }
+
+        return response(null, 422);
+    }
+
+    private function markReadByNotificationIdentifier($params)
```
nanaya: maybe should look into updating `Jobs\MarkNotificationsRead`...?

resources/assets/lib/notifications/notification-identity.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export interface NotificationIdentityJson {
+  category?: string;
+  id?: number;
+  object_id?: number;
+  object_type: string;
+}
+
+export function resolveStackId(identity: NotificationIdentity) {
```
nanaya: not used?

resources/assets/lib/notifications/notification-identity.ts
```diff
@@ -0,0 +1,68 @@
...
+  return `${identity.objectType}-${identity.objectId}-${identity.category}`;
+}
+
+export function resolveIdentityType(identity: NotificationIdentity): NotificationIdentityType | undefined {
```
nanaya: might as well make a class?

resources/assets/lib/notifications/notification-identity.ts
notbakaneko: is used for notification history page

resources/assets/lib/notifications/notification-identity.ts
notbakaneko: needs to work on interfaces

PR #5394, opened 2019-12-11
Add migration for show_in_client on user_contest_entries

nekodex:
backported from production


peppy:
Added migration to table.

PR #5396, opened 2019-12-11
Update more pages to use new header

nanaya:
Some special cases:
- profile page:
  - a bit taller (+30px) to compensate the loss from title area
    - cover size updated to 2400x640
  - full opacity
  - old gutter size
  - loading spinner visibility not animated so it can be removed when not used
  - mode switcher pretty much hack job
- chat:
  - different gutter size to match channel icons position
  - height of nav+70px
- supporter:
  - height of 450px (matches old one)


resources/assets/lib/header-v4.tsx
```diff
@@ -53,20 +54,22 @@ export default class HeaderV4 extends React.Component<Props> {
...
-                  {this.props.subSection}
-                </span>
-              </div>
+            <div className='header-v4__title'>
+              <span className='header-v4__title-section'>
+                {this.props.section}
+              </span>
+              <span className='header-v4__title-action'>
+                {' '}
+                {this.props.subSection}
```
nekodex: can use interpolation:
```
` ${this.props.subSection}`
```
(although not sure if better or worse)

nanaya: It'll be gone anyway :dancer:

PR #5399, opened 2019-12-11
Mark as read fixes for the web chat client

### LINKED ISSUE #4896: Opening a chat notification should dismiss all related notifications

nekodex:
Fixes:
 - when initially loading the chat page, a mark-as-read for the starting channel is not always sent
 - while actively looking at a channel, mark-as-read isn't always sent reliably

resolves #4896


resources/assets/lib/chat/chat-orchestrator.ts
```diff
@@ -147,15 +152,23 @@ export default class ChatOrchestrator implements DispatchListener {
...
-    const lastRead = channel.lastMessageId;
+    const lastReadId = channel.lastMessageId;
 
     if (!channel.isUnread) {
       return;
     }
 
-    this.api.markAsRead(channel.channelId, lastRead)
+    // We don't need to send mark-as-read for our own messages, as the cursor is automatically bumped forward server-side when sending messages.
+    const lastSentMessage = channel.messages.find((message) => message.messageId === lastReadId);
```
nanaya: this probably should be from last index otherwise the loop will get longer over time?

resources/assets/lib/chat/chat-orchestrator.ts
```diff
@@ -18,7 +18,7 @@
 
 import {
   ChatChannelPartAction,
-  ChatChannelSwitchAction,
+  ChatChannelSwitchAction, ChatMessageAddAction,
```
nanaya: the others are one line per thing

PR #5401, opened 2019-12-11
Update header for home section and some other cleanups

nanaya:
- [x] continues from #5396


nekodex:
Capitalisation is sneaking back in again... we want it all lowercase, e.g.:

nanaya:
@arflyte said the new headers should use normal capitalisation though :thinking:

nekodex:
@peppy overruled and said to use lowercase (fwiw, we've had this discussion like 2-3 times already for the previous iterations of the header design)

PR #5404, opened 2019-12-12
Chat UI fixes (including autoscroll)

### LINKED ISSUE #4897: Chat doesn't scroll to last unread (or bottom) on load

nekodex:
Fixes:
 - Spelling error in class name `chat-conversation-list-seperator`
 - Some colours (certain links were still blue, 'limitations' text using wrong colour)
 - Autoscrolling when switching channels (either to the unread marker or the bottom of the conversation, depending)

fixes #4897, supersedes #5040


resources/assets/lib/chat/conversation-view.tsx
```diff
@@ -26,13 +26,27 @@ import { Spinner } from 'spinner';
...
+import DispatcherAction from '../actions/dispatcher-action';
+import DispatchListener from '../dispatch-listener';
 import MessageDivider from './message-divider';
 import MessageGroup from './message-group';
 
 @inject('dataStore')
+@inject('dispatcher')
 @observer
-export default class ConversationView extends React.Component<any, any> {
+export default class ConversationView extends React.Component<any, any> implements DispatchListener {
```
nanaya: just noticed the props and states are not typed at all.

resources/assets/lib/chat/conversation-view.tsx
```diff
@@ -45,8 +59,32 @@ export default class ConversationView extends React.Component<any, any> {
...
+
+    const unreadMarker = $(chatView).find(`.${this.baseCssClass}__${this.unreadMarkerClass}`);
+    this.unreadMarker = unreadMarker.length > 0 ? unreadMarker : undefined;
+
+    if (this.didSwitchChannel) {
+      if (this.unreadMarker) {
+        this.scrollToUnread();
+      } else {
+        this.scrollToBottom();
+      }
```
nanaya: I wonder if this check is better done in `scrollToUnread`.

resources/assets/lib/chat/conversation-view.tsx
```diff
@@ -45,8 +59,32 @@ export default class ConversationView extends React.Component<any, any> {
...
 
-    if (this.props.dataStore.uiState.chat.autoScroll) {
-      $(chatView).scrollTop(chatView.scrollHeight);
+    const dataStore = this.props.dataStore;
+    const channel = dataStore.channelStore.channels.get(dataStore.uiState.chat.selected);
+    if (!channel.loaded) {
+      return;
+    }
+
+    const unreadMarker = $(chatView).find(`.${this.baseCssClass}__${this.unreadMarkerClass}`);
```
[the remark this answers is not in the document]
nekodex: this... is using the ref? it's just jquery-ifying it here?

resources/assets/lib/chat/conversation-view.tsx
```diff
@@ -178,4 +216,26 @@ export default class ConversationView extends React.Component<any, any> {
       </div>
     );
   }
+
+  scrollToBottom = (): void => {
+    const chatView = this.chatViewRef.current;
+    if (chatView) {
+      $(chatView).scrollTop(chatView.scrollHeight);
```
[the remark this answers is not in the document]
nekodex: Apparently iOS Safari still doesn't like Element.scrollTo 🤷‍♂ 

(according to: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo)

resources/assets/lib/chat/conversation-view.tsx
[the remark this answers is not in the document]
nanaya: you mean jquerying the parent and then finding the thing. Instead of attaching ref to the thing itself (and maybe then jquerying it)

resources/assets/lib/chat/conversation-view.tsx
nekodex: Not sure, I figured it'd be easier to follow with this logic at a higher level? (in `componentDidUpdate`)

resources/assets/lib/chat/conversation-view.tsx
```diff
@@ -103,18 +145,29 @@ export default class ConversationView extends React.Component<any, any> {
...
+        unreadMarkerShown = true;
+
+        // If the unread marker is the first element in this conversation, it most likely means that the unread cursor
+        // is even further in the past, making the displayed marker somewhat useless (until we can back-load those
+        // past messages in)... thus we ignore it when auto-scrolling and just go to the bottom instead.
+        //
+        // TODO: Actually in hindsight, there's another scenario where the first element in the conversation is an
+        // unread marker - when you receive new PMs and have yet to read any. Will look to handle this case later...
+        if (!_.isEmpty(conversationStack)) {
+          this.assumeHasBacklog = true;
```
nanaya: shouldn't this be set to `false` before the loop?

resources/assets/lib/chat/conversation-view.tsx
```diff
@@ -103,18 +146,29 @@ export default class ConversationView extends React.Component<any, any> {
...
+      if (!unreadMarkerShown && message.messageId > dataStore.uiState.chat.lastReadId && message.sender.id !== currentUser.id) {
+        unreadMarkerShown = true;
+
+        // If the unread marker is the first element in this conversation, it most likely means that the unread cursor
+        // is even further in the past, making the displayed marker somewhat useless (until we can back-load those
+        // past messages in)... thus we ignore it when auto-scrolling and just go to the bottom instead.
+        //
+        // TODO: Actually in hindsight, there's another scenario where the first element in the conversation is an
+        // unread marker - when you receive new PMs and have yet to read any. Will look to handle this case later...
+        if (!_.isEmpty(conversationStack)) {
```
nanaya: looking closer, this check is wrong? It's when the stack is empty then there might be backlog?

```suggestion
        if (_.isEmpty(conversationStack)) {
```

PR #5405, opened 2019-12-12
Update default button colour

nanaya:
Current one is a bit bright. It doesn't even use "primary button color" in the color file.

### DIFF
--- a/resources/assets/less/bem/btn-osu-big.less
+++ b/resources/assets/less/bem/btn-osu-big.less
@@ -26,7 +26,7 @@
 
   display: inline-block;
 
-  background-color: @osu-colour-h2;
+  background-color: @osu-colour-l4;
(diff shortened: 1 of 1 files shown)


peppy:
Is this fixing a specific page/issue? Screenshots would be useful.

nekodex:
The new on-hover colour doesn't exactly look great (looks murky/washed out), may want to double-check with @arflyte before doing this

notbakaneko:
lower button in each is on hover

before:
image


after:
image

nanaya:
Buttons on forum bottom floating bar are already using this colour :thinking: 
Though the non-section-hue ones use different combination.

notbakaneko:
everyone else agreed the old colours look better and @arflyte said to keep the old colours

PR #5409, opened 2019-12-13
Fix pluralisation rule not applied correctly for certain language

nanaya:
Monkeypatch fun :monkey:


nanaya:
Only other language with two components are chinese and those don't have variations for plural so it should be fine :dancer:

PR #5410, opened 2019-12-13
Add semi-transparent placeholder default colour

nanaya:


peppy:
Does this fix a specific issue?

PR #5418, opened 2019-12-16
Add tag for grouping to datadog request_time metric

### LINKED ISSUE #5391: Datadog key request_time has no prefix and too many tags
Needs grouping at very least, as discussed. Also it's very hard to find because it has no `osu.web.` prefix.

notbakaneko:
- add prefix to metrics
- add section name to tag for grouping

closes #5391


nanaya:
Probably should fix the action tag while at it (removing controller namespace prefix, [converting to snake_case_controller:action_name because tag limitation](https://docs.datadoghq.com/tagging/)).

PR #5450, opened 2019-12-23
Stop reporting OperationTimeoutException to Sentry

notbakaneko:
The change to use `cache_remember_mutexed` caused the previous exception silencing behaviour to regress since `cache_remember_mutexed` catches and reports all exceptions. The side-effect ended up being the code it's calling can't handle the exception anymore


app/helpers.php
```diff
@@ -95,8 +95,15 @@ function cache_remember_mutexed(string $key, $seconds, $default, callable $callb
 
                 cache()->put($fullKey, $data, max($oneMonthInSeconds, $seconds * 10));
             } catch (Exception $e) {
-                // Log and continue with data from the first ::get.
-                log_error($e);
+                $handled = false;
```
nanaya: $handled = $exceptionHandler !== null && $exceptionHandler($e);

PR #5471, opened 2019-12-26
Proxy BBCode [audio]

### LINKED ISSUE #5470: BBCode [audio] sources aren't proxied
are they supposed to be? seems weird that img does this but audio doesn't

notbakaneko:
assumes a version of camo that can supports proxying audio like go-camo

- [x] make sure `--allow-content-audio` is enabled
- [x] also requires go-camo >= 2.1.0

closes #5470


peppy:
✅  go-camo is now 2.1.4

PR #5472, opened 2019-12-27
Encode args for redirecting in RedirectController

notbakaneko:
~~return 404 instead of dying if there aren't enough parameters found to generate a route to redirect to.~~
Encode args instead


cl8n:
this controller might not be necessary anymore since Route::redirect is a thing

nekodex:
Unless something's changed with Laravel, the whole reason for that controller existing was due to closures (iirc, used to re-map params for `Route::redirect`) in the route file being incompatible with Laravel's route caching (heck, weren't you the person that made that change? 👀)

notbakaneko:
we're still using `ujs_redirect` and still can't `route:cache` closures

cl8n:
the version of laravel being used at the time didn't have Route::redirect, seeing this PR just reminded me about it

nvm then just wasn't sure if it was still needed

nanaya:
This is laravel/framework#26715 problem. The current fix seems dangerous in case we forget to sync the redirect route parameter when the target is updated and have it silently breaks.

Normal parameter missing should already be handled correctly (as in no route match).

PR #5480, opened 2020-01-05
Delete NP.png

febrimuhnas:
Fix wrong width of Nepal flag


peppy:
It seems like you may have meant to make a different change than you did. If you're not confident in using git, please consider opening an issue with the updated resource and someone will tend to it for you.

febrimuhnas:
wait, i want to change the flag but i don't permitted to upload a new file. how i can change it?

peppy:
You would need to probably use a git client. Not the web interface.

PR #5486, opened 2020-01-08
Show contest header in contest post content

nanaya:
- header image inside the page as the header background is barely visible
- use new font


peppy:
Is this still a required chance? Reason it's stuck in review / conflict?

PR #5494, opened 2020-01-09
Fix missing hit_accuracy in spotlight ranking API

### LINKED ISSUE #5252: Rankings chart endpoint is missing hit_accuracy
Screenshot_3

notbakaneko:
closes #5252

Adds new attribute that uses `accuracy` if `accuracy_new` doesn't exist.
Not sure if should just change all uses of `accuracy_new` to `hit_accuracy`, though


peppy:
`accuracy` is incorrect in the case `accuracy_new` is present, so the way you have it looks fine. it uses an older calculation method in places like `user_stats`.

PR #5524, opened 2020-01-20
Make direct download button use beatmap id instead of beatmapset id

### LINKED ISSUE #5523: osu!direct download button uses set instead of beatmap
Makes "go to map" button in the game a pain to use

stanriders:
Fixes #5523 

I'm not familiar with coffeescript so I couldn't make sure if `@props.currentBeatmap.id` is a thing, hope it is


peppy:
The PR description makes it sound like you didn't test this. Could you explain why you were unable to test if so?

stanriders:
I don't have anything set up for it nor any knowledge of how I'm supposed to test things in osu-web (or php in general), so I've decided to just wing it since it's a really tiny change.

peppy:
If contributing again, please look into setting up a proper environment. It can take as little as 5 minutes and is a basic expectation for contributors.

stanriders:
Sure

PR #5535, opened 2020-01-23
Add autocomplete to wiki main page search bar

notbakaneko:
- [x] requires #5528

Adds basic autocomplete support when typing into the search input
image


resources/assets/lib/wiki-search-controller.ts
```diff
@@ -0,0 +1,107 @@
...
+  shiftSelectedIndex(direction: number) {
+    this.selectIndex(this.selectedIndex + direction, direction);
+  }
+
+  @action
+  updateQuery(query: string) {
+    this.saved = this.query = query;
+
+    if (this.query.trim().length === 0) {
+      this.suggestions = [];
```
nanaya: also need abort running request if any

resources/assets/lib/wiki-search-controller.ts
```diff
@@ -0,0 +1,125 @@
...
+interface SuggestionJSON {
+  highlight: string;
+  path: string;
+  title: string;
+}
+
+export class WikiSearchController {
+  @observable selectedIndex = -1;
+  @observable shouldShowSuggestions = false;
+  @observable suggestions = observable.array<SuggestionJSON>([]);
```
nanaya: I just noticed, the `@observable` decorator isn't needed here, is it?

notbakaneko: Technically, `@observable` is implicitly wrapping the array and `observable.array` is just there to infer the correct type so we can reassign the variable later without typescript errors

PR #5539, opened 2020-01-24
Add support for nested click menu

nanaya:
- convert Fade to typescript
- convert ClickMenu to typescript
  - change signature of extra parameter for showing menu event from `target` to `{target}`
- add support for nested click menu


resources/assets/lib/click-menu.ts
```diff
@@ -0,0 +1,135 @@
...
+    } else {
+      document.body.dataset.clickMenuCurrent = this.current;
+    }
+  }
+
+  show = (target?: string | null | undefined) => {
+    this.current = target;
+
+    const tree = this.tree();
+    const menus = Array.from(document.querySelectorAll('.js-click-menu[data-click-menu-id]'));
```
notbakaneko: apparently adding `dom.iterable` to `lib` in `tsconfig.json` removes the need to use `Array.from`

PR #5547, opened 2020-01-27
Untranslate supporter donation email

### LINKED ISSUE #5504: Supporter emails should not be directly translated
The support emails being sent out under my name should not be directly translated into the user's end language. I don't mind the translation being included, but it should follow the original english text for clarity, and mention that it is a user-provided translation.

notbakaneko:
Translation keys are set to hidden in crowdin; the PR removes the translations and unshares the the opening and closing line of the mail.

closes #5504

- [ ] use only one of #5547 or #5548


peppy:
Is it possible to include the translation below the non-translated content as mentioned in the issue?

nanaya:
#5548

PR #5549, opened 2020-01-27
Fix duration not being translated in supporter email

notbakaneko:
Mailer's locale hasn't been set when the object is constructed


notbakaneko:
already fixed in #5548

PR #5553, opened 2020-01-28
Add pluralization for number of times a post was edited

notbakaneko:
Reported in https://crowdin.com/translate/osu-web/27/en-it#633


resources/lang/en/forum.php
```diff
@@ -60,7 +60,7 @@
     'post' => [
         'confirm_destroy' => 'Really delete post?',
         'confirm_restore' => 'Really restore post?',
-        'edited' => 'Last edited by :user :when, edited :count times in total.',
+        'edited' => 'Last edited by :user :when, edited :count time in total.|Last edited by :user :when, edited :count times in total.',
```
nanaya: Although almost never needed, this should use `:count_delimited`

PR #5562, opened 2020-01-30
Endpoint for creating beatmap discussion 'reviews'

nekodex:
Part of the server-side requirements for the upcoming beatmap discussion editor. The expected document structure for a review is:
```ts
export interface DocumentBlock {
  text: string;
  type: 'paragraph' | 'embed';
}

export interface DocumentParagraph extends DocumentBlock {
  type: 'paragraph';
}

export interface DocumentIssueEmbed extends DocumentBlock {
  beatmap_id: number | null;
  discussion_type: 'praise' | 'problem' | 'suggestion';
  timestamp: string;
  type: 'embed';
}

export type BeatmapReviewBlock =
	DocumentIssueEmbed | DocumentParagraph;

export type BeatmapDiscussionReview = BeatmapReviewBlock[];
```
Basically just an array of 'blocks', with each block being of type 'paragraph' or 'embed', with relevant attributes.

A minimum of _x_ `DocumentIssueEmbed` blocks will be required to consider the document a valid review. (where _x_ is configurable in `.env`, but just defaults to 1 for now)

This is just the creation endpoint and doesn't handle editing yet. I'm also open to suggestions as to if there's a better path for these endpoints.


app/Http/Controllers/BeatmapDiscussionsController.php
```diff
@@ -120,6 +123,129 @@ public function restore($id)
...
+            // generate the post body now that the issues have been created
+            foreach ($document as $block) {
+                switch ($block['type']) {
+                    case 'paragraph':
+                        if (!$block['text']) {
+                            throw new \Exception(trans('beatmap_discussions.review.validation.paragraph_missing_text'));
+                        }
+                        // escape embed injection attempts
+                        $text = preg_replace('/%\[\]\(#(\d+)\)/', '%\[\]\(#$1\)', $block['text']);
+                        $output[] = "{$text}\n";
```
nanaya: I wonder if storing the document as is (with `embed` changed into `embed_link` or something) would be easier.

app/Libraries/BeatmapDiscussionReview.php
```diff
@@ -0,0 +1,145 @@
...
+namespace App\Libraries;
+
+use App\Exceptions\InvariantException;
+use App\Models\BeatmapDiscussion;
+use App\Models\BeatmapDiscussionPost;
+use App\Models\Beatmapset;
+use App\Models\User;
+use DB;
+
+class BeatmapDiscussionReview
```
nanaya: The model name really should've been BeatmapsetDiscussion(Post), so at least this should use the correct name :pensive:

PR #5591, opened 2020-02-05
Merge errors from discussion to post if discussion validation fails

nekodex:
Fixes validation errors getting lost when calling `saveOrExplode` on a new `BeatmapDiscussion`.


nekodex:
Maybe? But I'm not sure how useful that'd be in this case, seeing as that validation is tied to both the `BeatmapDiscussion` and the `BeatmapDiscussionPost`? 🤷‍♂

PR #5599, opened 2020-02-07
Add Ukraine localisation

peppy:
81% complete on [crowdin](https://crowdin.com/project/osu-web/uk#), which is more than zero.


resources/lang/uk/model_validation/fulfillments.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+return [
+    'username_change' => [
+        'only_one' => '',
+        'insufficient_paid' => '',
+        'reverting_username_mismatch' => '',
+    ],
+    'supporter_tag' => [
+        'insufficient_paid' => 'Сума платежу недостатня для придбання osu!supporter (: actual > :expected)',
```
notbakaneko: There shouldn't be a space in `:actual` but I also think this line isn't supposed to be translated.
I've already fixed it on crowdin.

peppy: Have pulled down updates 👍

PR #5627, opened 2020-02-13
Add user title on forum post info

### LINKED ISSUE #5612: Add user title on forum
Seems like it would work nicely, especially for people with some kind of custom title, would mix great with usergroup badges informing about usergroup so it's not confusing, some examples of how i see it in mind, would be probably pretty much straightforward to implement if deemed fitting:

 

It's probably also to note it would have a good informing role for support team

venix12:
Fixes #5612


resources/views/forum/topics/_post_info.blade.php
```diff
@@ -40,6 +40,12 @@ class="forum-post-info__row forum-post-info__row--username js-usercard"
             data-user-id="{{$user->user_id}}"
             href="{{ route("users.show", $user) }}"
         >{{ $user->username }}</a>
+
+        @if ($user->title())
```
nanaya: Although usually fine, conditional test using plain string in php should be avoided. Use `isset` check.

venix12: seems like ``isset`` won't work in this context, will go for a regular check instead lol

nanaya:
You need to either allow edits from maintainers or update your branch with master.

venix12:
uh yes, seems like for some reason i didn't do it 🤔 both done anyways

nanaya:
Except it's already outdated :thinking:

PR #5606, opened 2020-02-07
New json format for beatmap discussion reviews

nekodex:
Dependant on:
- [x] #5562
- [x] #5604

PR #5631, opened 2020-02-14
Add timestamp parsing/linkifying in review markdown blocks

nekodex:
depends on
- [x] #5606


resources/assets/lib/beatmap-discussions/timestamp-plugin.ts
```diff
@@ -0,0 +1,64 @@
...
+    const match = value.substr(fromIndex).search(/[0-9]{2}/);
+
+    return match < 0 ? match : match + fromIndex;
+  }
+
+  function inlineTokenizer(eat: Eat, value: string, silent?: true): Node | boolean | void {
+    const regex = new RegExp(/((\d{2,}:[0-5]\d[:.]\d{3})( \((?:\d[,|])*\d\))?)/);
+    const result = regex.exec(value);
+
+    if (!result || result.index !== 0) {
```
nanaya: the `.index` check probably isn't needed if the regex starts with `^`

resources/assets/lib/beatmap-discussions/review-post.tsx
```diff
@@ -53,6 +57,9 @@ export class ReviewPost extends React.Component<Props> {
                 <div className='beatmapset-discussion-message' {...props}/>
               </div>;
             },
+            timestamp: (props) => {
+               return <a className='beatmapset-discussion-message__timestamp' {...props}/>;
```
nanaya: maybe use the simpler syntax for this (`(x) => <a ... />`)

PR #5639, opened 2020-02-17
Use count query for first place scores

notbakaneko:
Use count query for first place scores instead of fetching all the rows and creating models and counting them.


peppy:
Was closing this intentional? Explanation would be useful.

nanaya:
Deleted wrong branch :small_airplane:

notbakaneko:
Moving other WIP fixups to another branch since it makes more sense there.
Hopefully I won't accidentally delete that branch too ᕕ( ᐛ )ᕗ

PR #5642, opened 2020-02-18
Add breaking change note for beatmap and build stream json response update

nanaya:
Forgot about this yesterday when updating the json responses.


peppy:
I think there's probably a few others we missed.

nanaya:
Yeah I need to remember about this :pensive:

PR #5645, opened 2020-02-18
Allow restricted users to see their own scores

### LINKED ISSUE #3694: Some areas of restricted users' profiles display incorrectly
Likely due to shared filtering logic, restricted users are unable to see certain portions of their profile as expected.

notbakaneko:
closes #3694


app/Http/Controllers/BeatmapsController.php
```diff
@@ -69,30 +69,28 @@ public function scores($id)
...
-
-            if ($user !== null) {
-                $score = (clone $query)->where('user_id', $user->user_id)->first();
 
-                if ($score !== null) {
+                if (isset($score)) {
                     $results['userScore'] = [
                         'position' => $score->userRank(compact('type', 'mods')),
                         'score' => json_item($score, 'Score', ['user', 'user.country']),
                     ];
```
notbakaneko: @peppy @smoogipoo need some confirmation if lazer can handle `null` position and if there's any difference if `userScore` is not set versus `userScore: null` for a beatmap's scores

peppy: not being present will result in a position of 0. null will require a change at our end to support it.

notbakaneko: we're going to go with not setting `position` if it's not available for now

PR #5650, opened 2020-02-20
Fix user best performance returning empty

notbakaneko:
broken by #5541

Fixes best performance showing empty when cache gets refreshed after being viewed by a guest .
Also moving the overrides to the params classes for requests since non-request params should be useable without any authentication (e.g. in when console)

defaulting `isLoginRequired` to `true` was not the best idea (・_・ヾ

### DIFF
--- a/app/Libraries/Search/BeatmapsetSearchRequestParams.php
+++ b/app/Libraries/Search/BeatmapsetSearchRequestParams.php
@@ -47,6 +47,8 @@ class BeatmapsetSearchRequestParams extends BeatmapsetSearchParams
         '8' => 'loved',
     ];
 
+    private $requestQuery;
+
     public function __construct(array $request, ?User $user = null)
     {
         parent::__construct();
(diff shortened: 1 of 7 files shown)


app/Libraries/Search/PostSearchRequestParams.php
```diff
@@ -34,4 +34,9 @@ public function __construct(array $request, User $user)
         $this->forumId = get_int($request['forum_id'] ?? null);
         $this->includeSubforums = get_bool($request['forum_children'] ?? null);
     }
+
+    public function isLoginRequired(): bool
+    {
+        return true;
+    }
```
nanaya: this is probably a bit confusing when a guest accesses the relevant page.

app/Libraries/Search/BeatmapsetSearchRequestParams.php
```diff
@@ -134,6 +136,11 @@ public static function getAvailableFilters()
         return compact('extras', 'general', 'genres', 'languages', 'modes', 'played', 'ranks', 'statuses');
     }
 
+    public function isLoginRequired(): bool
+    {
+        return present($this->requestQuery);
+    }
```
nanaya: Interestingly, this doesn't actually affect `shouldReturnEmptyResponse` for this class :upside_down_face:

PR #5652, opened 2020-02-20
Auto-generated wiki sitemap

notbakaneko:
Auto generated wiki sitemap; not sure if it's worth keeping the old cached sitemap while a new one is being generated

path order is a bit busted at the moment because it's a case-sensitive sort


nanaya:
conflict...

resources/views/wiki/sitemap.blade.php
```diff
@@ -0,0 +1,69 @@
...
+    as published by the Free Software Foundation.
+
+    osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+    See the GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+--}}
+@extends('master')
```
nanaya: can use a prepend title "sitemap" (also maybe inside the page?)

.travis.yml
```diff
@@ -56,6 +56,7 @@ install:
   - wget -q --waitretry=10 --retry-connrefused -T 60 -O - http://127.0.0.1:9200
   - php artisan es:create-search-blacklist
   - php artisan es:index-documents --yes
+  - php artisan es:index-wiki --create-only --yes
```
nanaya: add to `migrate:fresh` command as well?

app/Libraries/Wiki/WikiSitemap.php
```diff
@@ -0,0 +1,113 @@
...
+            'titles' => $this->titles,
+        ];
+    }
+
+    private function parse(Hit $hit)
+    {
+        $page = Page::fromEs($hit);
+        $key = $hit->source('locale').'/'.$hit->source('path');
+
+        $this->titles[$key] = $page->title();
```
nanaya: At least one page is missing parent, resulting in error due to nonexistent page title.

This page: https://osu.ppy.sh/help/wiki/Tournaments/SOLMT/2019

Has no: https://osu.ppy.sh/help/wiki/Tournaments/SOLMT

(this one should probably be fixed in `osu-wiki`)

app/Libraries/Wiki/WikiSitemap.php
```diff
@@ -0,0 +1,113 @@
...
+class WikiSitemap
+{
+    public $titles = []; // stores the titles
+    public $sitemap = []; // stores the tree structure
+
+    public static function allPagesSearch()
+    {
+        return (new BasicSearch(Page::esIndexName(), 'wiki_sitemap'))
+            ->query(['exists' => ['field' => 'page']])
+            ->sort(new Sort('_id', 'asc'));
```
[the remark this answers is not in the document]
nanaya: as it turned out, it's because the full `_id` has `/$locale.md` suffix.

resources/views/wiki/_sitemap_section.blade.php
```diff
@@ -0,0 +1,38 @@
...
+    osu!web is distributed WITHOUT ANY WARRANTY; without even the implied
+    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+    See the GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with osu!web.  If not, see <http://www.gnu.org/licenses/>.
+--}}
+
+@php
+    $page = is_array($sitemap) ? $parent : $sitemap;
```
nanaya: Only use `$parent` as the value will always be the same?

```suggestion
    $page = $parent;
```

`$page` can probably even be removed. Or rename `$parent` to `$page` as it took me a while wondering why link for `$parent` is generated here instead of the caller (parent).

(and [as I said before](https://github.com/ppy/osu-web/pull/5652#discussion_r390757197), I don't see point filling [the value with `$page->path` instead of just null](https://github.com/ppy/osu-web/pull/5652/files#diff-897db912bb97ca161df702078958e918R111) as they may be overwritten by array anyway)

PR #5663, opened 2020-02-21
Show time ago for statuses in beatmap page 

notbakaneko:
to match the game client


nanaya:
@peppy will need to update crowdin once this is merged

huupoke12:
Is the missing bold in the time intentional?
image

nl-tatatat:
make it bold

MBomb-osu:
Could there be a little more clarification on what "to match game client" actually means? I'm assuming it's something with lazer, however I feel it'd actually be better to do it the other way around, and change the game client to say exact dates instead. Having the information of date ranked right in front of you I'd say is much more helpful, and precise, than saying "x months ago", and it looks a lot cleaner on the site too. I'd also say working out a time period from a date is a lot easier than working out a date from a time period (e.g. If someone asked me what 9 months ago was, it'd take me a few seconds to work it out, however if someone asked me when May was, I would easily be able to say 9 months ago within a second or two), so this change feels quite odd to me.

ascendancelol:
Dunno what the purpose of this was, the exact date of things was far more convenient to look at at a glance. if this was made to be consistent with something in-game, I'd suggest making it work the other way to where the exact date is shown instead of "time since". 

If I'm looking through my maps quickly, I'd much rather see that my map was ranked on February 1, 2020 instead of 26 days ago. This is especially the case for maps farther back when time is no longer measured in days but months or years. Even if you get said info by mousing over, there should be less searching for simple information please!

JBHyperion:
@MBomb-osu I think "match game client" refers to the "dates ago" you see on profile history, e.g. receiving kudosu, submitted beatmaps, scores (also applies to the beatmap listing top scores), as well as comments and discussions sections, etc. which all tend to "months ago" when > 28 days.

nl-tatatat:
Yeah JBH, but for those metrics you actually want to see how many days/months ago it was. For map submissions, you want to see if your map is reaching the graveyard once it reaches ~28 days ago, for kudosu you want to see how recently the last mod was, and for scores you want to see about how long ago the score was submitted. You don't need/want an exact date for those things. I think you DO want exact dates while browsing beatmaps though.

JBHyperion:
@nl-tatatat I didn't say it was a good idea, I just theorised why it was done this way (format consistency)

Agree it would be better to have exact dates since record-keeping purposes are something the "collective" care more about for beatmaps than they do for the other things I listed (which are more of interest to the individual), but maybe I'm just used to it this way.

peppy:
Please open an issue stating all the places on the site you’d want to see absolute dates. Discussion shouldn’t happen in merged PRs

PR #5665, opened 2020-02-21
Add support for declaring permissions on transformers

notbakaneko:
Allows permissions to be declared at the transformer level.
Using a transformer without the required permission will return empty; properties that require a permission that current user doesn't have for will not be rendered.

- [x] includes/requires #5603

- now also fixes some cases where a `null` is passed to an nested include


tests/Transformers/BeatmapDiscussionTransformerTest.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        parent::setUp();
+
+        $mapper = factory(User::class)->create();
+        $beatmapset = factory(Beatmapset::class)->states('with_discussion')->create([
+            'user_id' => $mapper->getKey(),
+        ]);
+
+        $this->beatmapDiscussion = $beatmapset->beatmapDiscussions()->first();
+        $this->beatmapDiscussion->update(['deleted_at' => now()]);
```
nanaya: missing variable declaration, also maybe call it `deletedBeatmapDiscussion`

tests/Transformers/BeatmapDiscussionPostTransformerTest.php
```diff
@@ -0,0 +1,91 @@
...
+use App\Models\User;
+use Tests\TestCase;
+
+class BeatmapDiscussionPostTransformerTest extends TestCase
+{
+    /** @var BeatmapDiscussion */
+    protected $beatmapDiscussion;
+
+    /** @var BeatmapDiscussionPost */
+    protected $post;
```
nanaya: `deletedPost`?

PR #5673, opened 2020-02-23
Remove Ubuntu quick setup script

### LINKED ISSUE #5672: Quick Ubuntu development setup script is outdated
I just planned to start contributing and set up fresh Ubuntu VM. After using priovided bootstrap.sh it seems that php is outdated there. I will try to increment it, retry and provide adequate PR for that.

image

Tetriando:
Since the Ubuntu quick development setup script is outdated and probably unreliable to keep eye on since there are other ways like docker, the proposition is to get rid of it completely and let developers setup and adjust their environment manually if needed.

I also adjusted some parts in README.md and SETUP.md besides strictly the Ubuntu setup. I changed the parts that I'm aware of so far. If there is more please let me know. If those changes should be separate from this PR, also let me know for the future.

Fixes #5672


Tetriando:
For example I was thinking if Node dependency mentioned in docs shouldn't be at least in version 10 LTS since 8 LTS is already deprecated.

PR #5691, opened 2020-02-28
Font and style updates

nanaya:
- RIP exo (and open sans)
- dark beatmapset card
- torus bold is now actually semibold (800 for actual bold, 900 for heavy)
- still need to remove a bunch of `.link-default()`s (later)
- etc


yansaan:
Why not change download beatmaps icon, is not cool!

Philosoph228:
wtf this font is painful. Return exo2 back!!!

PR #5694, opened 2020-02-28
Fixed the issue with time of loved not showing in the Germanic translation

marcostudios:
As said on https://osu.ppy.sh/community/forums/topics/1028649 the time when the map was loved isn't showing due to the missing ':timeago', any German speaking user could check if the line makes sense? Thank you :)


bdach:
I'm pretty sure it is not proper to apply this directly to source, as translations are sourced from [crowdin](https://crowdin.com/project/osu-web). This change, if correct, should be applied there to avoid this change being overwritten by the next translation update (check git history of modified file).

peppy:
Yep, this needs to be done via crowdin.

PR #5700, opened 2020-03-02
Add users.read API scope

notbakaneko:
this is pretty much #4815 rebased onto the new transformer permissions stuff + additional tests

- [x] requires #5699


resources/lang/en/api.php
```diff
@@ -33,5 +33,9 @@
         'friends' => [
             'read' => 'See who you are following.',
         ],
+
+        'users' => [
+            'read' => 'Read the public profile of any user.',
```
nanaya: ...on your behalf.

Can also use some rearrangements as it may sound a bit weird.

PR #5722, opened 2020-03-06
Adjust expert difficulty colour to be closer to black

peppy:


celerih:
Screenshot_20200306-073415~2

Issue is now when you hover over an Expert+ diff you don't get any text (since it's coloured like the icon amd u get black over black) so it doesn't show star rating

peppy:
yep

PR #5735, opened 2020-03-09
Name all API routes

notbakaneko:
Provide all API routes with a name so we can use `route()` to create the url during tests.


notbakaneko:
Worked out a way to setup route tests without the name, so this isn't needed anymore.

PR #5746, opened 2020-03-10
Select closest available channel after parting a channel

### LINKED ISSUE #5624: Chat should choose next conversation, not previous, when closing
image

notbakaneko:
Selects the next available channel when leaving a channel, instead of always defaulting to the first one.
This should also handle when leaving channels via other tabs/windows/clients.

closes #5624


resources/assets/lib/chat/chat-orchestrator.ts
```diff
@@ -74,33 +74,29 @@ export default class ChatOrchestrator implements DispatchListener {
...
     }
 
-    const channelList = channelStore.channelList;
-    if (channelList.length > 0) {
-      // TODO: switch to next 'closest' conversation instead of first in list
-      dispatch(new ChatChannelSwitchAction(channelList[0].channelId));
-    } else {
-      channelStore.loaded  = false;
-    }
+    const nextIndex = Math.min(index, channelList.length - 1);
```
nanaya: maybe add check `>= 0` as well (or use lodash' clamp)

resources/assets/lib/chat/chat-orchestrator.ts
```diff
@@ -74,33 +74,29 @@ export default class ChatOrchestrator implements DispatchListener {
...
 
-  focusNextChannel() {
-    const channelStore = this.rootDataStore.channelStore;
-    const channel = channelStore.get(this.rootDataStore.uiState.chat.selected);
-    if (channel && (channel.exists || channel.newChannel)) {
-      return;
+  focusChannelAtIndex(index: number) {
+    const channelList = this.rootDataStore.channelStore.channelList;
+    if (channelList.length === 0) {
+      return this.rootDataStore.channelStore.loaded = false;
```
nanaya: is it actually intended to return `false` or just for line saving :thinking:

notbakaneko: line saving 😔

PR #5757, opened 2020-03-12
Remove excessive zalgo text input

### LINKED ISSUE #5398: Beatmap comments cannot handle Zalgo text overflow
Source map in question: https://osu.ppy.sh/beatmapsets/1077085#osu/2253921

notbakaneko:
Experimental removal of excessive combination characters from input while preserving valid combinations.

closes #5398


app/Models/Comment.php
```diff
@@ -93,6 +93,11 @@ public function replies()
         return $this->hasMany(static::class, 'parent_id');
     }
 
+    public function setMessageAttribute($value)
+    {
+        return $this->attributes['message'] = unzalgo($value, 2);
```
nanaya: default level instead of explicit?

app/Models/User.php
```diff
@@ -1849,6 +1861,13 @@ public function isValid()
             }
         }
 
+        if ($this->isDirty('user_twitter') && present($this->user_twitter)) {
+            // https://help.twitter.com/en/managing-your-account/twitter-username-rules
+            if (!preg_match('/^[a-zA-Z0-9_]{1,15}$/i', $this->user_twitter)) {
```
nanaya: `/i` and `a-zA-Z` seems a bit redundant

PR #5769, opened 2020-03-13
Convert ShowMoreLink component to Typescript

nanaya:
One component at a time. I have no idea how to cleanly handle `RefObject` without using `any` thanks to the returned element being different depending on the parameter. I prefer export default for things like this but the existing structure doesn't use it ┐(･_･)┌

- [x] depends on #5768


resources/assets/lib/show-more-link.tsx
```diff
@@ -0,0 +1,81 @@
...
+      </span>
+    </>
+  );
+
+  let onClick = props.callback;
+  let url = props.url;
+
+  if (props.loading) {
+    onClick = undefined;
+    url = undefined;
```
notbakaneko: might as well return the `span` here; setting to `undefined` in one part of the `if` and checking the same variables `== null` in a nested part of the same conditional looks misleading

PR #5790, opened 2020-03-17
Update documentation for getting OAuth tokens

notbakaneko:


robinknaapen:
It's maybe nice to mention `state` for security purposes. https://auth0.com/docs/protocols/oauth2/oauth-state

PR #5791, opened 2020-03-17
Add API scope for reading beatmaps

notbakaneko:
- adds `beatmaps.read` scope
- helper to export and import list of declared middleware on api routes
- more specific scope tester


notbakaneko:
we're going to move to using a more generic public use scope instead

PR #5806, opened 2020-03-21
remove outdated info about kudosu stars

pavlukivan:
they were replaced with hypes, right? In that case available kudosu info is irrelevant nowadays.


peppy:
If this is to be removed, the underlying translation keys should be to. Also please give a visual context of which part you're removing (screenshot or something similar).

pavlukivan:
Before | after
image

pavlukivan:
Alternatively, it can be replaced with something like "kudosu earned last month" which would be somewhat useful

venix12:
well, wouldn't kudosu earned still make some sense? guess the intended direction is to ever make it useable?

pavlukivan:
I meant "replace kudosu available with kudosu earned last month", not "replace kudosu earned with kudosu earned last month"

venix12:
ah yes, kudosu available i mean, sorry

pavlukivan:
well, that's not up to me to decide

notbakaneko:
@peppy are we going to go ahead with removing this?

peppy:
Hmm... I guess there's no harm in hiding it for the time being.

Issue #5781, opened 2020-03-15
Modding "Kiai" Word Issue


bowlingfan:
Problem details: While I was modding this map: https://osu.ppy.sh/beatmapsets/1079916#osu/2348754, I suddenly see that whenever I post "(time) Kiai's end is not snapped by (number)ms" It says that it failed to save the post. I was so confused. I even showed that any other word except kiai was working well.

Video or screenshot showing the problem:
https://streamable.com/fp3hc

osu! version: Stable 20200304.1 (latest)

peppy:
Can you check your console to see what error you're getting?

bowlingfan:
I have no idea where to find that.

peppy:
In chrome: Right click and choose "inspect", then go to the network tab and reproduce your issue. You should see an error request appear, where you can get further details (screenshots of request/response would be a good start)

bowlingfan:
Ok, what I see is

beatmap-discussion | Stat: 422 | Type: xhr | Size: 548 B | Time: 117 ms

What do I look for? I'm not sure on where for the Network tab in the inspect element.

peppy:
Can you provide a screenshot showing the request/response portion?

bowlingfan:
https://imgur.com/rNDVXri

peppy:
@nanaya i remember you looking into this one possibly? something regarding beatmap length

nanaya:
It should show more useful error message now (or soon-ish). See if it helps.

peppy:
Closing for now. @bowlingfan please comment if you are still having issues here.

PR #5812, opened 2020-03-23
Improve message of beatmap discussion posting errors

nanaya:
The exception is already conveniently handled somewhere else.


peppy:
Related to #5781?

PR #5814, opened 2020-03-23
Fix script used by docker

### LINKED ISSUE #5809: Docker setup is not working
Steps:
1. `$ export UID`
2. `$ docker-compose up`

Error:
image

Sometimes there is a different error but I can't reproduce this reliably:
image

There is also this line getting spammed at the end of terminal, is this in any way related?
```
notification-server_1  | OAuth public key hasn't been created. Sleeping for 5 seconds before starting again...
```

Trying to troubleshoot:
- Accessing localhost:8080 redirects to localhost:8080/home so php is working.
- I logged in to mysql container and `mysql.user` table doesn't have `osuweb` user
- In mysql container db_setup.sh and db_user.sql exists inside /docker-entrypoint-initdb.d
- Looks like db_user.sql is not running for some reason. (Changing the `database` volume name also doesn't work, but I'm not sure if it was supposed to work in the first place.)

nanaya:
The script is run in weird way (presumably to skip exporting variables?) and result in the main shell having nounset and errexit set, breaking the init script.

The script is simple enough already so the options aren't really needed.

Should resolve #5809.

### DIFF
--- a/bin/db_setup.sh
+++ b/bin/db_setup.sh
@@ -1,11 +1,5 @@
 #!/bin/sh
 
-set -e
-set -u
-
-# goes to top directory.
-cd "$(dirname "${0}")/.."
-
 for db in osu osu_store osu_mp osu_chat osu_charts osu_updates; do
   echo "CREATE DATABASE ${db} DEFAULT CHARSET utf8mb4" | mysql -u root
 done


taylankasap:
I can confirm this fixes my issue.

PR #5815, opened 2020-03-23
Fix full height text editor in forum overlapping with header

### LINKED ISSUE #5803: mobile: forum post editor gets pushed under the header
Experiencing this on the Google Chrome mobile app. The post editor can get the top pushed under the header, and there is no way to move it back down for visibility. This thankfully doesn't hamper usability, but is very awkward for user experience. 

Screenshot_20200320-103634
Screenshot_20200320-103644

taylankasap:
Fixes #5803


peppy:
CI fail looks like a false positive?

PR #5824, opened 2020-03-26
Simplify chat sort messages function

nanaya:
Noticed this when poking around chat related classes.


taylankasap:
I don't know the number of items here so this might be over-optimization but shouldn't you call uniqBy first so sortBy is called on fewer items?

PR #5826, opened 2020-03-26
Remove web-only properties from User API response

### LINKED ISSUE #5644: rankHistory in user json should be part of transformer
[Problematic line](https://github.com/ppy/osu-web/blob/e2040db1d7debf97b165366c909744535410cf14/app/Http/Controllers/UsersController.php#L297).

For starter it's not even `snake_case`.

It should be moved inside transformer so it's centralised and easier to document.

And It'll be a breaking change because lazer is already using it :tada: 

... :slightly_frowning_face:

notbakaneko:
- [x] osu!lazer client should be updated to not parse the properties removed from API response
- `UserTransformer` now a superset of `UserCompactTransformer`
- typescript side now has `UserJSONExtended` and `UserJSON` respectively
- some permissions-related properties used explicitly for web removed from OAuth responses: `friends`, `is_admin`, `is_bng`, `is_full_bn`, `is_gmt`, `is_limited_bn`, `is_moderator`, `is_nat`, `is_restricted`
- `User` and `GroupJSON` moved to modules
- `rankHistory` moved into transformer (closes #5644)
- `UserCompactTransformer` now returns a default cover instead of `null`s if included

changes extracted from another branch that adds general public scope to API, so come commits may be loopy =)


app/Transformers/UserTransformer.php
```diff
@@ -5,96 +5,104 @@
...
+        'is_gmt' => 'IsNotOAuth',
+        'is_limited_bn' => 'IsNotOAuth',
+        'is_moderator' => 'IsNotOAuth',
+        'is_nat' => 'IsNotOAuth',
+        'is_restricted' => 'IsNotOAuth',
+    ];
+
+    public function __construct()
+    {
+        static $includes;
```
nanaya: looking again, I wonder if it's better to define all the includes in `UserCompact` instead.

notbakaneko: that does sound more useful 🤔

nanaya:
what if all the additional attributes are includes and `UserTransformer` literally only contains `$defaultIncludes`... 🤔 

(probably need some kind of private `$profile` variable and function so it's not generated repeatedly)

peppy:
I don't think lazer will need any updates for this (it parses only when data is present). May be best to go ahead with merge and release and we can test for any regressions lazer side due to (potentially?) missing data.

notbakaneko:
if there's no problems with parsing if the property doesn't exist then it should be fine 🤔

peppy:
oops, i thought this was already merged..

PR #5831, opened 2020-03-28
Fix incorrectly named supporter environment variable

peppy:


.env.example
```diff
@@ -239,4 +239,4 @@ SITE_SWITCHER_JS_HASH=
 #BEATMAPSET_UPLOAD_BONUS_PER_RANKED_MAX=2
 #BEATMAPSET_UPLOAD_ALLOWED_SUPPORTER=8
 #BEATMAPSET_UPLOAD_BONUS_PER_RANKED_SUPPORTER=1
-#BEATMAPSET_UPLOAD_BONUS_PER_RANKED_LIMIT_MAX=2
+#BEATMAPSET_UPLOAD_BONUS_PER_RANKED_SUPPORTER_MAX=2
```
nanaya: Use `_SUPPORTER` suffix just like other variables? Also matches the actual config name.

And match sample value with the default.

```suggestion
#BEATMAPSET_UPLOAD_BONUS_PER_RANKED_MAX_SUPPORTER=12
```

config/osu.php
```diff
@@ -51,7 +51,7 @@
         'upload_bonus_per_ranked_max' => get_int(env('BEATMAPSET_UPLOAD_BONUS_PER_RANKED_MAX')) ?? 2,
         'upload_allowed_supporter' => get_int(env('BEATMAPSET_UPLOAD_ALLOWED_SUPPORTER')) ?? 8,
         'upload_bonus_per_ranked_supporter' => get_int(env('BEATMAPSET_UPLOAD_BONUS_PER_RANKED_SUPPORTER')) ?? 1,
-        'upload_bonus_per_ranked_max_supporter' => get_int(env('BEATMAPSET_UPLOAD_BONUS_PER_RANKED_LIMIT_MAX')) ?? 12,
+        'upload_bonus_per_ranked_max_supporter' => get_int(env('BEATMAPSET_UPLOAD_BONUS_PER_RANKED_SUPPORTER_MAX')) ?? 12,
```
nanaya: Same as above

```suggestion
        'upload_bonus_per_ranked_max_supporter' => get_int(env('BEATMAPSET_UPLOAD_BONUS_PER_RANKED_MAX_SUPPORTER')) ?? 12,
```

PR #5843, opened 2020-03-31
Add beatmap default download option

### LINKED ISSUE #2079: Beatmap search listing missing ability to download a set without video
Curent website a small "no vid" link below the download button when viewing maps from the beatmap listing
This is totally missing on the new web, so if someone doesn't want the video they can't download that osz straight from the listing, which makes the listing less convenient for those that rarely download with video.

### LINKED ISSUE #5458: Beatmap panels should use osu!direct links if user is supporter
Doesn't make sense to download the map as an `.osz` if I have direct.

<sup><sub>Also prevents downloading the map for the fourth time</sub></sup>

nanaya:
No the most intuitive way as no one will discover it until they check settings page.

Resolves #5458, and kind of resolves #2079.


app/Transformers/UserTransformer.php
```diff
@@ -274,6 +274,7 @@ public function includeUserPreferences(User $user)
             $customization = $user->profileCustomization();
 
             return [
```
notbakaneko, line 276: I think this can be replaced with `return $customization->only(...)`?

PR #5845, opened 2020-03-31
Fix country mismatch in inactive user test

taylankasap:
In `SessionsControllerTest::testLoginInactiveUser` the created user has random country but ForceReactivation class detects the request country as `null`. Null obviously doesn't match with the user's country and it redirects to password reset page so the test fails.

If you need to test if the created user has the same country as the request, there should be a new testcase for that.

How on earth does this test pass on your setups?

_I run tests from inside of the container with `bin/phpunit -d memory_limit=-1`_

### DIFF
--- a/tests/Controllers/SessionsControllerTest.php
+++ b/tests/Controllers/SessionsControllerTest.php
@@ -5,6 +5,7 @@
 
 namespace Tests\Controllers;
 
+use App\Models\Country;
 use App\Models\LoginAttempt;
 use App\Models\User;
 use Tests\TestCase;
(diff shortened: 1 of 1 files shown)


taylankasap:
(I tinkered too much with this PR now it looks way too bloated. Let's ignore that.)

Is this what you had in mind regarding the issue?

tests/Controllers/SessionsControllerTest.php
```diff
@@ -27,12 +27,15 @@ public function testLogin()
     public function testLoginInactiveUser()
     {
         $password = 'password1';
-        $user = factory(User::class)->create(compact('password'));
+        $country_acronym = factory(\App\Models\Country::class)->create(['acronym' => '__'])->acronym;
```
nanaya: - no need to specify the acronym
- variable names should be camelCased
- import the class instead of specifying full name

taylankasap: > no need to specify the acronym

If I don't specify the acronym it gives duplicate error for me because countries exist in my case, that's the whole point of this PR.

> variable names should be camelCased

Compact won't work with camelCase, do you want me to ditch the compact here?

nanaya: > If I don't specify the acronym it gives duplicate error for me because countries exist in my case, that's the whole point of this PR.

```php
(Country::first() ?? factory(Country::class)->create())->getKey()
```

> Compact won't work with camelCase, do you want me to ditch the compact here?

yes

PR #5854, opened 2020-04-01
Prevent language/genre Any from being set

### LINKED ISSUE #5850: Setting map genre and language to Any should not be possible
Currently you can add the genre and language to a mapset if you are the owner of the set or have permissions.
Although there is currently no check if you selected the Any option in the dropdown, which results you being able to set the following:

nanaya:
Resolves #5850.

I was thinking adding column like `settable` to indicate whether it can be used for beatmapset but I couldn't find a good name so it's id-based for now. Id 0 is kind of special already so it should be fine...? 🤷‍♂ 

Also fixed validation message.

### DIFF
--- a/app/Models/Beatmapset.php
+++ b/app/Models/Beatmapset.php
@@ -1172,12 +1172,12 @@ class Beatmapset extends Model implements AfterCommit, Commentable
     {
         $this->validationErrors()->reset();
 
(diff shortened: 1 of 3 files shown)


notbakaneko:
I just noticed it should probably be possible to be able to set only one of the options if necessary and leaving the other as unspecified?

PR #5870, opened 2020-04-04
updated the store page to reflect changes

nl-tatatat:
No longer say that friends ranking is unavailable (because its functional),
and no longer say that the max upload limit with supporter is 12, because its actually 20.

I hope I did this right, haha.


nl-tatatat:
Will be changed from
https://i.imgur.com/9OvVk1o.png

To

https://i.imgur.com/kOv4nNO.png

nl-tatatat:
woops... I made a mistake.

PR #5873, opened 2020-04-05
Clarify docs for oauth/token parameters

### LINKED ISSUE #5872: Getting 400 on oauth/token
Hello, after I've obtained code from redirect url params during authorization, I'm trying to exchange code to refresh_token based on docs: https://osu.ppy.sh/docs/index.html#authorize-users-for-your-application:

```python
def do_auth(code):
    params = {
        'grant_type': 'authorization_code',
        'code': code,
        'client_id': CLIENT_ID,
        'client_secret': CLIENT_SECRET,
        'redirect_uri': THIS_HOST + '/redirect_osu'
    }
    url = 'http://osu.ppy.sh/oauth/token'
    response = requests.post(url, params=params)
```

but I got this:
```
layout.errors.400.error

The authorization grant type is not supported by the authorization server.

layout.errors.400.description
```

Is there anything I'am doing wrong?

UncleGoogle:
Clarify that required parameters should be passed in request body not as params in url
Resolves #5872

### DIFF
--- a/resources/views/vendor/apidoc/partials/info.blade.php
+++ b/resources/views/vendor/apidoc/partials/info.blade.php
@@ -124,7 +124,7 @@ fetch("{!! route('oauth.passport.token') !!}", {
 
 `POST {!! route('oauth.passport.token') !!}`
 
-Parameters
+Form-data parameters passed in the request body:
 
 Name          | Type   | Description
 --------------|--------|-------------------------------


notbakaneko:
This is unnecessary; `Parameters` is a standard way of referring to values to be sent in a `POST` request. Additionally, for `POST` requests, parameters are __expected__ to be sent in the body, not as part of the _query string_.
Furthermore, it's not limited to just `form-data`.

UncleGoogle:
I see, thanks for clarifying a few things.

I was just confused with the docs as few lines above there were exactly the same word `Parameters` for the query string (grant type code doc). But right - GET typically should not contain body, so `Parameters` resolves to "query string" by careful reader.

Additionally, I was pretty sure that putting those in headers is also valid way o doing POST for oauth. And I've found https://www.oauth.com/oauth2-servers/creating-documentation/ :
>When client authentication is required in a request, such as in the Authorization Code grant, there are two ways your service can accept the client ID and secret in the request. Your service can accept the authentication in an HTTP Basic Auth header using the client ID as the username and secret as the password, or by accepting the strings in the post body as client_id and client_secret. **It is up to your service whether you want to accept either or both of these methods, so you need to tell your developers how you expect them to include this authentication in requests.**

Enitoni:
I spent around 2 days ripping my hair out over this. I really think there should be an explicit clarification that in this particular example you need to use the request body. It is not obvious to a beginner or even an experienced person that you have to use one or the other depending on the request method.

peppy:
I'd suggest you do some initial reading on http protocol. POST requests are implicitly in the aforementioned format. Adding documentation of this in API docs is unnecessary (beyond the specification that the request type is a POST).

Enitoni:
If multiple people are complaining about this, there is clearly an issue. I do not understand what is so hard about just adding one single line that tells people to put the data in the request body. It seems like such an easy fix and there would not be any downsides to it. And then you won't have to worry about it ever again.

Docs should be explicit and easy to understand. Do not just assume that people know every single detail.

peppy:
On checking some various other API docs (github, digitalocean, cloudflare, twitter), there is no case where this is mentioned.


I think a more correct solution here would be to add the example field, missing for the aforementioned request. If you read down further on the page, examples show the parameters being POST form data, not url query string.

PR #5875, opened 2020-04-06
Update time display

### LINKED ISSUE #3762: Years should not round up at 0.5

### LINKED ISSUE #5859: Better or no rounding for silence timer on bullet list
1ez
One place it says the silence is only one hour, and one hour is past. Silence is still active because APPARENTLY it was originally 90 minute silence, but the site rounds it as one hour, which is rather confusing,

nanaya:
- use momentjs for timeago display - as it turned out timeago formatter can be easily replaced
- update momentjs time rounding ([defaults for reference](https://momentjs.com/docs/#/customization/relative-time-threshold/))
  - 119 minutes &rarr; 119 minutes
  - 120 minutes &rarr; 2 hours
  - 47 hours &rarr; 47 hours
  - 48 hours &rarr; 2 days
  - 61 days &rarr; 61 days
  - 62 days &rarr; 2 months
  - 23 months &rarr; 23 months
  - 24 months &rarr; 2 years

The rounding update should probably be updated with something better.

I also learned moment can display fraction like "2.5 years" instead of whole number although "7.2 hours" probably isn't useful (or is it?); unfortunately there's no way to tell the function what unit it's rounding. Different rounding can also be used.

Although probably look into migrating momentjs to date-fns [for better formatting](https://date-fns.org/v2.11.1/docs/formatDistance)?

Maybe resolves #3762? And resolves #5859?


peppy:
Those rounding cutoffs seem good for now. Not sure if we want fractions.

notbakaneko:
I looked at migrating to `date-fns` previously; the main issue at the time was the lack of localizations

nanaya:
It now has localization. There doesn't seem to be easy way to load the translation files separately though.

notbakaneko:
they only had a few of the languages we supported before 😞, that might have changed.

PR #5880, opened 2020-04-07
Generate https urls in api authentication docs

notbakaneko:
People just copy these urls so they should probably be https.
I wonder if it's better to have an option to just change the default route generation to https, instead of this roundabout way of getting https from `route`


notbakaneko:
`route` doesn't generate https or relative url by default when run on command line

PR #5885, opened 2020-04-08
Add per-route throttling

notbakaneko:
Just simple per-path for now, actual ~~per-group~~ and multi-level decay can come later since that involves overriding more things.

The decay time can't be changed until the previous timer expires.

per-group configuration throttling doesn't actually seem possible in a rational manner with the way middleware and routing works in laravel...


nanaya:
probably should also include http method as well for the key? Otherwise it's more like throttle-path.

notbakaneko:
closed in favour of #5886

PR #5886, opened 2020-04-08
Throttle password reset requests

notbakaneko:
prefixing is apparently a new thing in laravel 6


peppy:
What's the rate defined here? 20 per 1440 minutes?

nanaya:
20 per day, yes.

PR #5900, opened 2020-04-12
Log beatmap genre and language changes

cl8n:
- [x] requires #5963

NAT wants BN to be able to change these soon, and need some way to keep track of changes if anything goes wrong

I'm not sure what to do about /beatmapsets/events or /users/modding bc they rely on a blade template that's missing a lot of the functionality compared to the react version. should i duplicate the logic for now or move to react ?


notbakaneko:
> I'm not sure what to do about /beatmapsets/events or /users/modding bc they rely on a blade template that's missing a lot of the functionality compared to the react version. should i duplicate the logic for now or move to react ?
These should be moved to react; there's a react version of of modding profile events for the user modding profile page, which is like a budget version of the blade, which itself is basically a budget version of of the actual `beatmap-discussions/event` component, so these should actually be reconciled together sooner rather than later...

cl8n:
cool, thnx

one thing I wanted to change there was allowing event message templates to have event-specific properties defined on the `comment`, like in this PR I have `:old` and `:new` in the messages but it needed a specific check for these events.

PR #5907, opened 2020-04-13
Add options for notification delivery

notbakaneko:
This seems more complicated than it should be 🤔 

Allow push notification options to be set on the existing mail notification options.
Includes options for receiving beatmapset disqualify notifications (separate from problems reported)

Also fixes updating notification setting not showing any error on failure.

Since it uses #5923 to send per-user notifications, it doesn't rely on ppy/osu-notification-server#25 anymore >_>;;
- [x] instead, now depends on ppy/osu-notification-server#27


app/Jobs/BroadcastNotification.php
```diff
@@ -164,17 +186,35 @@ private function onBeatmapsetDiscussionQualifiedProblem()
...
+
+        $notificationOptions = UserNotificationOption
+            ::where(['name' => Notification::BEATMAPSET_DISQUALIFY])
+            ->whereNotNull('details')
+            ->get();
+
+        $this->receiverIds = static::beatmapsetWatcherUserIds($this->object);
+
+        foreach ($notificationOptions as $notificationOption) {
+            if (count(array_intersect($notificationOption->details['modes'] ?? [], $modes)) > 0) {
```
nanaya, line 214: should this check `push` option as well? Considering it's part of `supportsNotifications` check.

Although there isn't any way to set it from frontend. Or maybe should check the one in `BEATMAPSET_MODDING`...? :thinking:

app/Jobs/BroadcastNotification.php
```diff
@@ -34,12 +34,31 @@ class BroadcastNotification implements ShouldQueue
...
     private $source;
 
-    private static function beatmapsetReceiverIds($beatmapset)
+    private static function beatmapsetWatcherUserIds($beatmapset)
     {
-        return $beatmapset
-            ->watches()
-            ->pluck('user_id')
-            ->all();
+        return static::filterBeatmapsetModdingUserIds(
```
nanaya, line 39: I wonder if it's better to have a method in notification object itself (or its own class) to get all the watching users, both global and per-item

resources/assets/less/bem/osu-switch-v2.less
```diff
@@ -12,6 +12,10 @@
   text-transform: none;
   margin: 0 5px 0 0;
 
+  .account-edit-entry__grid & {
+    margin: 0;
+  }
```
nanaya, line 17: maybe a modifier instead?

app/Http/Controllers/AccountController.php
```diff
@@ -169,20 +171,32 @@ public function updateEmail()
...
 
-        if ($option->update($params)) {
-            return response(null, 204);
-        } else {
-            return response(['form_error' => [
-                'user_notification_option' => $option->validationErrors()->all(),
-            ]]);
-        }
+                $option = auth()->user()->notificationOptions()->firstOrCreate(['name' => $key]);
+                if (!$option->update($params)) {
```
nanaya, line 189: looking again, I probably should've just written this but while there's no `updateOrExplode`, there's `->fill($params)->saveOrExplode()`.

app/Http/Controllers/AccountController.php
```diff
@@ -169,20 +171,32 @@ public function updateEmail()
...
+                $params = get_params($value, null, ['details:any']);
 
-        if ($option->update($params)) {
-            return response(null, 204);
-        } else {
-            return response(['form_error' => [
-                'user_notification_option' => $option->validationErrors()->all(),
-            ]]);
-        }
+                $option = auth()->user()->notificationOptions()->firstOrCreate(['name' => $key]);
```
nanaya, line 188: this probably should use `firstOrNew` for slightly fewer queries?

app/Jobs/BroadcastNotification.php
notbakaneko: probably move this in a separate PR

tests/BeatmapsetDisqualifyTest.php
```diff
@@ -0,0 +1,178 @@
...
+        Event::assertDispatched(NewPrivateNotificationEvent::class, function (NewPrivateNotificationEvent $event) use (&$dispatchedCount) {
+            if ($event->notification->name === Notification::BEATMAPSET_DISQUALIFY) {
+                $this->assertSame(array_unique($event->getReceiverIds()), $event->getReceiverIds());
+                $dispatchedCount++;
+            }
+
+            return true;
+        });
+
+        $this->assertSame(1, $dispatchedCount);
```
nanaya, line 50:
```
$events = Event::dispatched(...::class, function (...) {
  if (...) {
    $this->assertSame(...);

    return true;
  }

  return false;
});

$this->assertSame(1, $events->count());
```

`assertDispatched` is literally above just `> 0` instead of `=== 1`.

app/Jobs/BroadcastNotification.php
notbakaneko: No, this option doesn't have anything to do with delivery mode

PR #5910, opened 2020-04-14
Add cutoff for showing relative time in beatmap info

### LINKED ISSUE #5682: Beatmap pages should show accurate dates for their data
Dropping in from replies on [this tweet](https://twitter.com/Typobad/status/1232787599238062083), I personally think the change in the beatmap description is better to be changed back to its old functionality.

While on user profiles it's alright to have less accurate time indicators shown for account creation and last login, for beatmaps these dates have a lot more meaning to them. The date of a beatmap, when combined with the beatmap creator's name, can sometimes quickly give a player information about the map they are playing. Within a year's time a mapper's style or skill in mapping can change a lot, when you see the difference between perhaps a early [2010 Happy30 map](https://osu.ppy.sh/beatmapsets/13223#osu/53554) and [one from just a few months later](https://osu.ppy.sh/beatmapsets/17670#osu/62963) a player might recognize them as probably not being similar due to the experience a mapper can gain between this time.

To tl;dr: the dates between profiles and beatmap descriptions serve different purposes which I think are important enough to keep in mind, this being inconsistent between pages isn't bad (as long as stuff on the same page is the same) if it means giving more important information to the user from the start (as opposed to having to hover over the "x time ago" to see the accurate date.)

notbakaneko:
- cuts off at 4 weeks
- add `timeago` support to `TimeWithTooltip`
- switch to react component for rendering

closes #5682

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -756,7 +756,7 @@ function timeago($date)
     $display_date = i18n_time($date);
     $attribute_date = json_time($date);
 
-    return "<time class='js-timeago timeago' datetime='{$attribute_date}'>{$display_date}</time>";
+    return "<time class='js-timeago' datetime='{$attribute_date}'>{$display_date}</time>";
 }
 
 function link_to_user($id, $username = null, $color = null, $classNames = null)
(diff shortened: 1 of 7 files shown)


nanaya:
hi @arflyte

PR #5911, opened 2020-04-14
Update audio player

### LINKED ISSUE #4096: Audio previews in news article don't pause when starting a new one
Example: 
https://osu.ppy.sh/home/news/2019-01-14-new-featured-artist-onumi

May want to look at hooking up our custom player (which does support this).

### LINKED ISSUE #4739: Add a volume control button on the Featured Artist page
It would be nice to have a button to control the volume on the Featured Artist webpage, on the bottom of the page, kinda like Soundcloud does on their website, since when you preview a song, the sound is pretty high and there is no way to manage the volume, unless using the Windows volume panel, which is quite inconvenient. 

Thanks!

nanaya:
Resolves #4739 and resolves #4096.

The audio player sure has grown quite big 🤔 

Using css vars for template instead of react because I'm not sure if re-rendering the player(s) every 1/60s through react is good idea 🕺  (and I still need to sync state with another component anyway)

The design on forum page can use some fixing though. Hi @arflyte


arflyte:
Hmm, can we make the player appear above the navigation bar for now? I'll think of a better way to fix this.

nanaya:
> Is this supposed to autoplay the next track...on any page?

yes

> I can't help but think the player should tell what's playing when it auto plays the next track or jump back to it

maybe later for beatmapset and featured artists page. That'll require a bit more changes but shouldn't be too difficult.

Though it won't be possible in some cases where there's just no metadata for them like news posts and forum. Unless we're just showing the file url.

PR #5916, opened 2020-04-15
Don't send global notification for new problem on qualified map if it also disqualifies

### LINKED ISSUE #5184: Nominators shouldn't receive notifications on disqualified beatmaps
Currently, disqualifying a map notifs the entire BNG, as shown in this image below.
IMG_20191026_053231
[Discussion linked](https://osu.ppy.sh/beatmapsets/1002819/discussion/-/generalAll#/1232171)

This should not ping entire BNG because it is no longer qualified and its pretty useless to notif as its most likely not going to have a huge discussion, just the mapper needs to fix something.

This also against the idea of "reporting qualified beatmap" as it is no longer qualified, and its no longer a report because its already disqualified. Disqualification happens mostly due to issues that already present that needs to be fixed before continuing to be in qualified/ranked, so there is no need to ping the BNG again once its disqualified.

nanaya:
Resolves #5184.

As mentioned in the issue, there's nothing to check for BNGs if the post also disqualifies the map.

### DIFF
--- a/app/Http/Controllers/BeatmapDiscussionPostsController.php
+++ b/app/Http/Controllers/BeatmapDiscussionPostsController.php
@@ -137,7 +137,7 @@ class BeatmapDiscussionPostsController extends Controller
 
         $notifyQualifiedProblem = false;
 
-        if ($discussion->beatmapset->isQualified() && $discussion->message_type === 'problem') {
+        if (!$disqualify && $discussion->beatmapset->isQualified() && $discussion->message_type === 'problem') {
             $openProblems = $discussion
                 ->beatmapset
                 ->beatmapDiscussions()
--- a/tests/Controllers/BeatmapDiscussionPostsControllerTest.php
+++ b/tests/Controllers/BeatmapDiscussionPostsControllerTest.php
@@ -780,8 +780,8 @@ class BeatmapDiscussionPostsControllerTest extends TestCase
         return [
             [
                 'bng',
-                [Notification::BEATMAPSET_DISQUALIFY, Notification::BEATMAPSET_DISCUSSION_POST_NEW, Notification::BEATMAPSET_DISCUSSION_QUALIFIED_PROBLEM],
-                [],
+                [Notification::BEATMAPSET_DISQUALIFY, Notification::BEATMAPSET_DISCUSSION_POST_NEW],
+                [Notification::BEATMAPSET_DISCUSSION_QUALIFIED_PROBLEM],
             ],
             [
                 null,


peppy:
I checked through the history, but it seems like this new change is actually putting things in the correct place.

Should notify:

- If a qualified beatmap has a problem reported, that should fire a notification
- If a qualified beatmap is disqualified, that should fire a notification

Shouldn't notify:

- If a disqualified beatmap has a problem reported, that should **not** fire a notification.
- If a qualified beatmap has a problem reported (and that reported problem triggers a disqualification due to the user being in a special user group) only one disqualification notification should fire, not both the "problem" and "disqualification".

nanaya:
Apparently I forgot to mention this is about global notification, the one that triggers everyone who enabled "receive notifications for new problem on qualified beatmaps".

And this disables global notification for "If a qualified beatmap is disqualified".

If someone watches the beatmapset they'll be notified when:

- someone posts something
- beatmapset state is changed (in addition to new post above)

If they doesn't watch the beatmapset but enabled global notification:

- someone posts new problem on qualified map but doesn't disqualify (because they don't have the permission to)
- someone posts new problem on qualified map which also disqualifies it (because they have the permission to) &larr; this is the one being removed in this PR

peppy:
Aha, I see. That makes sense when re-reading the issue this is closing. Let me check in with modding community to make sure this is what we want.

peppy:
image

Is it much extra effort to add this as a second option?

nanaya:
That's already in the correct context, right? As in for global notification, not when someone watches a beatmapset. There's no global notification for nomination resets at the moment.

peppy:
Yeah, I guess it's another way of phrasing the one that you are looking to remove in this PR.

nanaya:
ಠ_ಠ

I'm asking because the comment says:

> On-dq notifs personally help me keep track of the validity of BNs **nomination resets**, ...

There's no global notification for nomination resets at the moment, those are only for people watching the beatmapset.

peppy:
The one you're disabling is global though, right? That's what they've been using until now, so the proposal is making this a new toggle rather than outright disabling (am I missing something logically?)

notbakaneko:
I think, essentially, we need to have more options for global (implicit, non-watching) notifications?

Currently, beatmap modding-related notifications require the user to have watching enabled on the beatmap, _except_ for new problems reported on qualified beatmaps, which is sent to anyone with the option enabled.

There isn't a disqualification notification sent if you aren't watching the beatmap - it's currently lumped in together with the new problem on qualified map notification, which this PR will disable, after which you'll only get the disqualification notification if you are watching the map, but, it sounds like you should be able to get the notification for the disqualification even if you're not explicitly watching the map (but not receive 2 if you are).

peppy:
yeah, i'd see there being two checkboxes in settings: the existing one (which only fires on non-disuqalifying problems) and a new one (which fires on disqualifications)

nanaya:
Closing this and back to #5184.

PR #5919, opened 2020-04-16
Beatmap page style adjustments

### LINKED ISSUE #5166: "Hype" and "Report" buttons on beatmap information page are not the same size
image
Hype beatmap is just slightly thinner than Report beatmap, the result looks unpolished.

nanaya:
- remove titles which are never shown and seems redundant
- adjust margin between items
- full width button because it almost take half of the box width and there are variations in width between locales (up to almost full width)

Resolves #5166.

### DIFF
--- a/resources/assets/coffee/react/beatmapset-page/hype.coffee
+++ b/resources/assets/coffee/react/beatmapset-page/hype.coffee
@@ -82,19 +82,16 @@ export class Hype extends React.PureComponent
     return unless @props.beatmapset.status == 'qualified'
 
     if @userCanDisqualify()
-      title = osu.trans('beatmapsets.show.hype.disqualify.button_title')
       buttonParams =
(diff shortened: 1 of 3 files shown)


resources/assets/coffee/react/beatmapset-page/hype.coffee
```diff
@@ -75,28 +75,34 @@ export class Hype extends React.PureComponent
...
+          className: "#{bn}__button"
+          @renderReportButton()
+
+
+  renderReportButton: =>
+    return unless @props.beatmapset.status == 'qualified'
+
+    if @userCanDisqualify()
+      buttonParams =
+        text: osu.trans 'beatmaps.nominations.disqualify'
```
venix12, line 88:
```suggestion
        text: osu.trans('beatmaps.nominations.disqualify')
```

PR #5925, opened 2020-04-16
Adjust height of profile page detail bar

### LINKED ISSUE #5713: "Friends" buttons are not the same height
For some reason they are not equal to each other. So when you switch between your page and someone else's the content below this buttons is "jumping".

_Your profile_:
image

_Friend's profile_:
image

_Someone else's profile_:
image

### LINKED ISSUE #5775: Profile detail bar elements are slightly off when collapsing / expanding
The whole element becomes 1px shorter when expanded, causing slight misalignment of the friends button and the level badge:

nanaya:
Resolves #5713.

...except not really. The original reported bug has been (accidentally) fixed in #5653.

This PR further fixes the height change that occurs when toggling rank detail view due to one of the bar contents being slightly taller than expected. That pushes the entire container height when it's rendered, resulting in the buttons being moved around.

and resolves #5775

### DIFF
--- a/resources/assets/less/bem/profile-detail-bar.less
+++ b/resources/assets/less/bem/profile-detail-bar.less
@@ -12,6 +12,7 @@
 
   @media @desktop {
     padding: 10px @gutter-profile-page-desktop;
+    height: 60px;
   }
 
   &__column {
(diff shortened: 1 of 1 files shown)


TheWildTree:
Also fixes #5775.

PR #5936, opened 2020-04-22
Switch to multiple user channel notification

### LINKED ISSUE #5923: Switch to multiple user channel notification
(Title is tentative)

Currently, notifications are sent to a specific channel where you have to explicitly opt-in to 'watching' the object you want to get notifications for, or a 'global' notification where you receive notifications of a single type from any object (ironically, sent via a per-user channel). The issue is it's not possible to do both for a notification type, and also not have it send duplicate notifications if you have the watch and global one enabled at the same time, which was an issue when doing #5174. 

This has come up again lately due to discussions around #5916, #5184, and the need to implement more settings for notification delivery options #5907 and more notifications in the future.

notbakaneko:
closes #5923

backports the changes to `RedisBroadcaster` from Laravel 6.12+ which uses Lua to send to multiple pubsub channels at once. This effectively makes all current notifications per-user with the fanning out handled by Redis instead of the job worker; i.e. a single message is sent to Redis.

A side-effect of this is all notification broadcasts are effectively per-user and only get 'sent' if it's enabled.


app/Models/Chat/Channel.php
```diff
@@ -200,7 +198,6 @@ public function receiveMessage(User $sender, string $content, bool $isAction = f
 
         if ($this->isPM()) {
             $this->unhide();
-            broadcast_notification(Notification::CHANNEL_MESSAGE, $message, $sender);
```
nanaya, line 203: this probably shouldn't be removed?

PR #5945, opened 2020-04-24
Move time formatting to javascript

### LINKED ISSUE #5928: Forum poll end time is in UTC
`Polling will end at April 17, 2020 at 4:03:43 PM UTC`
should probably use the same time formatting as everything else.

nanaya:
Resolves #5928.


resources/assets/lib/localtime.ts
```diff
@@ -0,0 +1,62 @@
...
+    if (!(parent instanceof HTMLElement)) {
+      return [];
+    }
+
+    if (parent instanceof HTMLTimeElement && parent.classList.contains('js-localtime')) {
+      return [parent];
+    } else {
+      // Casting is needed because the compiler doesn't detect result of
+      // 'time.class-name' query as array of time elements.
+      return [...parent.querySelectorAll('time.js-localtime')] as HTMLTimeElement[];
```
notbakaneko, line 47:
```suggestion
      return [...parent.querySelectorAll<HTMLTimeElement>('time.js-localtime')];
```

PR #5947, opened 2020-04-24
Review editor for beatmap modding discussions

nekodex:
Okay, this is the big one - sorry in advance for how fat the PR ended up being. 😅

What's a review? Well, it's our take on the concept of a 'code review', but for Beatmaps. It allows a modder to group together discussions/issues with additional text for context/comment/etc, without undermining the ability for a mapper to see and tackle each issue individually if they prefer.

Hopefully this will both streamline the modding process and add back a personal touch.

Screen Shot 2020-04-24 at 8 23 49 pm

There's a few little issues remaining here and there (that I'll fix in seperate pull requests as to not bloat this any further), but this should hopefully be in a good enough state to start reviewing...

Interface/tab for reviews won't be visible unless `.env` has `BEATMAPSET_DISCUSSION_REVIEW_ENABLED=true`

Also I could do with some feedback about what size reviews we want to allow:
```
BEATMAPSET_DISCUSSION_REVIEW_MAXIMUM_BLOCKS=10
BEATMAPSET_DISCUSSION_REVIEW_MINIMUM_ISSUES=1
```
_(a block is each paragraph of text or embedded issue)_

**Things to note:**
 - Formatting is only bold/italic for now, but we can add more later easily enough
 - Formatting can only be applied _outside_ of embedded discussions (intentional)
 - Nearby issues detection doesn't exist yet
 - No feedback for character limits or block requirements until after posting
 - You can't delete discussions while editing a review, only 'unlink' them.
 - You can't 'resolve' a review, only the discussions/issues within it (might be worth showing the review as 'resolved' if all the stuff inside is?)

**Known issues (will fixes these in follow-up PRs):**
 - Formatting toolbar shouldn't appear inside embedded discussions
 - Hitbox on hover-menu is a bit wonky and needs some tweaking
 - ~~If a timestamp exists, it will always be a timestamped discussion (should be changed to happen only if the line starts with the timestamp)~~ (addressed in this PR)
 - ~~Doesn't respect discussion locking yet~~ (addressed in this PR)

...and I've probably missed some other stuff 🤷‍♂️


resources/assets/lib/beatmap-discussions/editor-beatmap-selector.tsx
```diff
@@ -0,0 +1,63 @@
...
+      id: 'all',
+      label: osu.trans('beatmaps.discussions.mode.scopes.generalAll'),
+    });
+
+    this.props.beatmaps.forEach((beatmap: Beatmap) => {
+      if (beatmap.deleted_at) {
+        return;
+      }
+
+      this.menuOptions.push({
```
nanaya, line 36: the props might never be changed but it sure is kind of weird seeing this being permanently set based on initial props

iptq:
looks clean!

resources/assets/lib/beatmap-discussions/new-review.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+declare global {
+  interface Window {
+    stickyHeader: StickyHeader;
+  }
+}
+
+export default class NewReview extends React.Component<Props, State> {
+  initialValue: string;
+  placeholder: string = '[{"children": [{"text": "placeholder"}], "type": "paragraph"}]';
```
nanaya, line 37: neither seem to be used. It's passed to `Editor` which then ignore it.

resources/assets/lib/beatmap-discussions/editor-discussion-component.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+interface Props extends RenderElementProps {
+  // attributes taken from RenderElementProps, but extended with contentEditable
+  attributes: {
+    contentEditable?: boolean;
+    'data-slate-inline'?: true;
+    'data-slate-node': 'element';
+    'data-slate-void'?: true;
+    dir?: 'rtl';
+    ref: any;
```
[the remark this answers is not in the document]
nekodex: `// attributes taken from RenderElementProps, but extended with contentEditable`

resources/assets/lib/beatmap-discussions/editor-beatmap-selector.tsx
nekodex: Yeah not sure, the idea was to save regenerating the menu on every render... probably fine just moving it to render? ...but then again the same time the page is already pretty heavy, so 🤷‍♂️

resources/assets/lib/beatmap-discussions/editor-issue-type-selector.tsx
```diff
@@ -0,0 +1,58 @@
...
+export default class EditorIssueTypeSelector extends React.Component<Props> {
+  static contextType = SlateContext;
+  menuOptions: MenuItem[];
+
+  constructor(props: Props) {
+    super(props);
+
+    this.menuOptions = [];
+    selectableTypes.forEach((type) => {
+      this.menuOptions.push({
```
nanaya, line 35: or `this.menuOptions = selectableTypes.map(...)`

resources/assets/less/bem/beatmap-discussion-editor.less
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  &__input-area {
+    .content-font();
+
+    max-height: 70vh;
+    overflow-y: auto;
+
+    padding: 15px;
+    padding-bottom: 50px;
```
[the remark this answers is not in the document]
nekodex: I feel like we encountered this same issue/"intended-behaviour" with the chat UI?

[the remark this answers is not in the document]
nanaya: yeah, it was fixed by adding the padding to the last element instead of the container

resources/assets/lib/beatmap-discussions/editor-toolbar.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { createPortal } from 'react-dom';
+import { Editor, Range, Text, Transforms } from 'slate';
+import { ReactEditor, useSlate } from 'slate-react';
+
+export const EditorToolbar = () => {
+  const bn = 'beatmap-discussion-editor-toolbar';
+  const ref = useRef({} as HTMLDivElement);
+  const editor = useSlate();
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
nekodex: Yeah, it's based on [this example](https://github.com/ianstormtaylor/slate/blob/master/site/examples/hovering-toolbar.js)...

I guess the difference is I've since extracted it out of the parent (Editor) because it was getting too big... but they're kinda coupled in their behaviour?

resources/assets/lib/beatmap-discussions/editor.tsx
```diff
@@ -0,0 +1,467 @@
...
+    }
+  }
+
+  componentWillUpdate(): void {
+    this.cache = {};
+  }
+
+  decorate = (entry: NodeEntry) => {
+    const [node, path] = entry;
+    const ranges: TimestampRange[] = [];
```
nanaya, line 100: `decorate` sounds like a generic name but the returned ranges are strictly timestamp ranges?

resources/assets/lib/beatmap-discussions/editor.tsx
```diff
@@ -0,0 +1,467 @@
...
+        value: parseFromMarkdown(this.props.document, this.props.discussions),
+      });
+    }
+  }
+
+  toggleItalic = () => {
+    this.toggleMark('italic');
+  }
+
+  toggleMark = (format: 'bold' | 'italic') => {
```
nanaya, line 407: this and `toggleFormat` (why different name?) in `ToolbarButton` behaves slightly differently:

1. create a text with bolded word in the middle (`something **like ** this`)
2. highlight the whole text
3. toggle bold

result:
- bolded text if toggled using mod-b
- unbolded text if toggled using toolbar

resources/assets/lib/beatmap-discussions/editor.tsx
```diff
@@ -0,0 +1,467 @@
...
+  toggleBold = () => {
+    this.toggleMark('bold');
+  }
+
+  toggleEditMode = (enabled: boolean) => {
+    if (!this.props.document || !this.props.discussions || _.isEmpty(this.props.discussions)) {
+      return;
+    }
+
+    if (enabled) {
```
nanaya, line 396: looks like this can simply check `this.props.editing`.

resources/assets/lib/beatmap-discussions/editor.tsx
nanaya: actually, why are they two different functions?

resources/assets/lib/beatmap-discussions/editor.tsx
```diff
@@ -0,0 +1,467 @@
...
+            beatmaps={this.sortedBeatmaps()}
+            readOnly={this.state.posting}
+            {...props}
+          />
+        );
+        break;
+
+      case 'link':
+        el =  <a href={props.element.url} rel='nofollow'>{props.children}</a>;
+        break;
```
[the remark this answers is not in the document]
nekodex: `autolink-plugin.ts`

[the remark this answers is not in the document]
nanaya: but that's for remark and editor doesn't seem to use it? ?_? What should I type/do to trigger this?

resources/assets/lib/beatmap-discussions/new-review.tsx
```diff
@@ -0,0 +1,136 @@
...
+          </div>
+        </div>
+      </div>
+    );
+  }
+
+  setSticky = (sticky = true) => {
+    this.setState({
+      cssTop: this.cssTop(sticky),
+    });
```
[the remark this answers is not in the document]
nekodex: I just implemented it the same way as it's done in `new-discussion.coffee` for consistency, kinda feel like refactoring how it behaves is out of scope for this PR...

resources/assets/coffee/_classes/beatmap-helper.coffee
```diff
@@ -51,6 +51,6 @@ class @BeatmapHelper
 
   @sort: (beatmaps) ->
     if beatmaps[0].mode == 'mania'
-      _.orderBy beatmaps, ['convert', 'cs', 'difficulty_rating'], ['desc', 'asc', 'asc']
+      _.orderBy beatmaps, ['mode_int', 'convert', 'cs', 'difficulty_rating'], ['asc', 'desc', 'asc', 'asc']
```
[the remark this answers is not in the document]
nekodex: ugh, so... group by mode, sort each mode individually, then recombine?

resources/assets/lib/beatmap-discussions/new-review.tsx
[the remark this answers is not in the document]
nanaya: probably should at least add comment on both sides so they're updated in sync

nekodex:
@nanaya ready for another review pass, issues should be all fixed 🤞

resources/assets/lib/beatmap-discussions/editor-discussion-component.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  componentDidMount = () => {
+    // reset timestamp to null on clone
+    Transforms.setNodes(this.context, {timestamp: null}, {at: this.path()});
+  }
+
+  componentDidUpdate = () => {
+    const path = this.path();
+
+    if (this.props.element.beatmapId !== 'all') {
```
nanaya, line 42: is this the part that's supposed to handle this? Except it doesn't work and gives out "timestamp is specified" error when posted.

(`beatmapId` is undefined for general all)

resources/assets/lib/beatmap-discussions/editor-discussion-component.tsx
```diff
@@ -0,0 +1,139 @@
...
+import { ReactEditor } from 'slate-react';
+import { BeatmapsetJson } from '../beatmapsets/beatmapset-json';
+import BeatmapJsonExtended from '../interfaces/beatmap-json-extended';
+import EditorBeatmapSelector from './editor-beatmap-selector';
+import EditorIssueTypeSelector from './editor-issue-type-selector';
+import { SlateContext } from './slate-context';
+
+interface Props extends RenderElementProps {
+  // attributes taken from RenderElementProps, but extended with contentEditable
+  attributes: {
```
nanaya, line 16: the comment is now rather confusing because it's not extending anything anymore...

Remove?

resources/assets/lib/beatmap-discussions/editor-helpers.ts
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        review.push(doc);
+        break;
+    }
+  });
+
+  // strip last block if it's empty (i.e. the placeholder that allows easier insertion at the end of a document)
+  const lastBlock = review[review.length - 1];
+  if (lastBlock.type === 'paragraph' && !osu.present(lastBlock.text)) {
```
nanaya, line 89: is this a non-`stringify` version of [this check](https://github.com/ppy/osu-web/blob/62b5fb082a4ba9a68aba941076aaf7cac29ce5ce/resources/assets/lib/beatmap-discussions/editor.tsx#L225)? Perhaps decide on one and move it to its own helper function.

resources/assets/lib/beatmap-discussions/editor-toolbar.tsx
```diff
@@ -0,0 +1,139 @@
...
+import { SlateContext } from './slate-context';
+
+const bn = 'beatmap-discussion-editor-toolbar';
+
+export class EditorToolbar extends React.Component {
+  static contextType = SlateContext;
+  ref: React.RefObject<HTMLDivElement>;
+  scrollContainer: HTMLElement | undefined;
+  private scrollTimer: number | undefined;
+  private readonly throttledUpdate: (() => null) & _.Cancelable; // TODO: is there a better way of typing this?
```
nanaya, line 19: Just assign the value here (`private readonly throttledUpdate = _.throttle(...)`)

Can also be done for the `ref` above I believe.

resources/assets/lib/beatmapsets/beatmapset-json.ts
```diff
@@ -3,6 +3,7 @@
 
 import GenreJson from 'interfaces/genre-json';
 import LanguageJson from 'interfaces/language-json';
+import BeatmapJson from '../interfaces/beatmap-json';
```
nanaya, line 6: this must've been automated because it's weirdly different compared to other imports right above it 🙃

resources/assets/lib/beatmap-discussions/editor-helpers.ts
nekodex: no?

nanaya: upon closer inspection, the other one checks entire document. Although it can still probably be done by doing something like `doc.every(isEmptyBlock)`

Shmiklak:
Italic formatting doesn't work properly, example: https://osu.ppy.sh/beatmapsets/1085819/discussion/-/reviews#/1623552

PR #5958, opened 2020-04-27
Record volume settings

### LINKED ISSUE #5949: Audio player: remember last used volume setting.
On the audio player, if I change the volume to be maximum instead of half, this setting is only remembered so long as the osu website is open in a particular visit. Closing the tab and revisiting osu later resets the volume to half. 

This can be very inconvenient and kind of defeats the point of volume control, as people could be having issues with the map previews blasting their speakers again or being too quiet despite setting the volume on a previous visit.

nanaya:
Resolves #5949.

I'm not sure if the detail is good enough:
- there are two settings: per-browser and per-user; both are set when changing volume
- per-browser settings have higher priority over per-user
- the settings are only applied on the very first load
- as we're not clearing local storage on login, it may override per-user settings (although it doesn't save it for the user)


resources/assets/lib/osu-audio/settings.ts
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import { route } from 'laroute';
+import Main from './main';
+
+export default class Settings {
```
notbakaneko, line 7: nothing in this class seems to require being declared as an arrow function on a class field, regular function prototypes should be preferred; simple `get`/`set` can just use getters and setters

PR #5961, opened 2020-04-28
Only autoplay next track within same group

### LINKED ISSUE #5940: Stop autoplaying the next song on literally everything
I wanted to see what a map was on the BSS and a completely unrelated song that I had no interest in was shoved into my ears once it finished.
It is good for newsposts, but please don't use it on the beatmap listing.

nanaya:
Doesn't resolve　#5940.


nanaya:
There are people on both sides, so ┐(°～° )┌

PR #5963, opened 2020-04-28
Refactor beatmap modding event component

notbakaneko:
Moves the different react/non-react variations into one component.

Some of the other stuff in `ModdingHistoryController` can probably make use of the new `ModdingHistoryEventsBundle` but I haven't gotten around to moving them yet.


resources/assets/lib/interfaces/beatmapset-event-json.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import { BeatmapsetJson } from 'beatmapsets/beatmapset-json';
+
+export default interface BeatmapsetEventJson {
+  beatmapset?: BeatmapsetJson;
+  comment: any;
```
nanaya, line 8: add TODO about fixing the model so there's no plain string anymore

app/Libraries/ModdingHistoryEventsBundle.php
```diff
@@ -0,0 +1,274 @@
...
+        }
+
+        $votesGiven = ($this->votes['given'] ?? collect())->pluck('user_id')->toArray();
+        $votesReceived = ($this->votes['received'] ?? collect())->pluck('user_id')->toArray();
+
+        $userIds = array_merge(
+            $userIds,
+            $this->posts->pluck('user_id')->toArray(),
+            $this->posts->pluck('last_editor_id')->toArray(),
+            $this->events->pluck('user_id')->toArray(),
```
nanaya, line 252: looks like bug waiting to happen when someone moves `getUsers` call to before those are assigned (also `$this->votes` and `$this->discussions` above).

resources/assets/lib/beatmap-discussions/event.tsx
```diff
@@ -0,0 +1,144 @@
...
+    const cover = this.props.event.beatmapset?.covers?.list;
+    let discussionLink = route('beatmapsets.discussion', { beatmapset: this.beatmapsetId });
+    if (this.discussionId != null) {
+      discussionLink = `${discussionLink}#/${this.discussionId}`;
+    }
+
+    return (
+      <div className='beatmapset-event'>
+        <a href={discussionLink}>
+          <img className='beatmapset-activities__beatmapset-cover' src={cover} />
```
nanaya, line 75: I think this is time to do something with this random element without its block.

notbakaneko:
I can't seem to trigger any errors with my data set, do you have the error?

PR #5973, opened 2020-04-30
Add different translation for ended contests without end date

### LINKED ISSUE #5965: Contest listing localization problems with "ended"
In English it sounds fine to write _ended + date_, but in other languages (e.g. German or French) you need an auxiliarly word like _endete + am + datum_ in German or _terminé + le + date_ in French.

Problems occur when in the original version it just says _ended_, which should be translated as simply _endete/beendet_ or _terminé_, respectively. However, it's translated with the auxiliary word, which is not needed.

English:
grafik
German:
grafik

Maybe a second "ended" on Crowdin would be helpful.

nanaya:
Resolves #5965.

### DIFF
--- a/app/Models/Contest.php
+++ b/app/Models/Contest.php
@@ -163,7 +163,11 @@ class Contest extends Model
             case 'voting':
                 return i18n_date($this->voting_starts_at).' - '.i18n_date($this->voting_ends_at);
             default:
-                return trans('contest.dates.ended', ['date' => i18n_date($this->voting_ends_at)]);
+                if ($this->voting_ends_at === null) {
+                    return trans('contest.dates.ended_no_date');
+                } else {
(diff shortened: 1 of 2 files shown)


app/Models/Contest.php
```diff
@@ -163,7 +163,11 @@ public function currentPhaseDateRange()
             case 'voting':
                 return i18n_date($this->voting_starts_at).' - '.i18n_date($this->voting_ends_at);
             default:
-                return trans('contest.dates.ended', ['date' => i18n_date($this->voting_ends_at)]);
+                if ($this->voting_ends_at === null) {
+                    return trans('contest.dates.ended_no_date');
+                } else {
+                    return trans('contest.dates.ended', ['date' => i18n_date($this->voting_ends_at)]);
```
notbakaneko, line 169: this should be `js_localtime` now?

PR #5981, opened 2020-05-01
Add more endpoints to public api

notbakaneko:
- `public` scope added
- `users.read` scope removed, anything that required this scope now uses `public` instead.
- tests now use a route+scope file
- snake case `createdAt` in `EventTransformer`
- added a helper to read/dump routes and their declared scopes
`php artisan route:convert --from-{csv|json}=source --to-{csv|json}=destination`
omitting the destination dumps json to the console.
omitting the source reads the routes and scopes from code.
`GET` and `HEAD` are currently duplicated but they're generated from the same route, so probably will look at addressing that in the future...

current api v2 routes:

route | method | scope required
-- | -- | --
api/v2/comments | GET |  
api/v2/comments/{comment} | GET |  
api/v2/changelog/{stream}/{build} | GET |  
api/v2/changelog | GET |  
api/v2/changelog/{changelog} | GET |  
api/v2/seasonal-backgrounds | GET |  
api/v2/beatmaps/{id}/scores | GET | public
api/v2/beatmaps/lookup | GET | public
api/v2/beatmaps/{beatmap} | GET | public
api/v2/beatmapsets/search/{filters?} | GET | public
api/v2/beatmapsets/lookup | GET | public
api/v2/beatmapsets/{beatmapset} | GET | public
api/v2/friends | GET | friends.read
api/v2/me/{mode?} | GET | identify
api/v2/news | GET | public
api/v2/news/{news} | GET | public
api/v2/rankings/{mode}/{type} | GET | public
api/v2/spotlights | GET | public
api/v2/users/{user}/kudosu | GET | public
api/v2/users/{user}/scores/{type} | GET | public
api/v2/users/{user}/beatmapsets/{type} | GET | public
api/v2/users/{user}/recent_activity | GET | public
api/v2/users/{user}/{mode?} | GET | public
api/v2/wiki/{page?} | GET |


app/Libraries/RouteScopesHelper.php
```diff
@@ -0,0 +1,142 @@
...
+                array_splice($middlewares, $index ? $index + 1 : count($middlewares), 0, $newString);
+            }
+
+            $route['middlewares'] = $middlewares;
+        }
+    }
+
+    public function fromCsv(string $filename)
+    {
+        $csv = array_map('str_getcsv', file($filename));
```
nanaya, line 54: does `fgetcsv` not work?

app/Providers/AuthServiceProvider.php
```diff
@@ -54,7 +54,7 @@ public function boot()
         Passport::tokensCan([
             'friends.read' => trans('api.scopes.friends.read'),
             'identify' => trans('api.scopes.identify'),
-            'users.read' => trans('api.scopes.users.read'),
```
nanaya, line 57: there are quite a few active `users.read` token at the moment. Are we going to force them to reauthorize?

Although I also noticed the scope check is currently an `all` (as in requires all the specified scopes) instead of `any` (requires any of the specified scopes)

tests/RouteScopesTest.php
```diff
@@ -5,53 +5,128 @@
...
         foreach (Route::getRoutes() as $route) {
+            /** @var \Illuminate\Routing\Route $route */
             if (!starts_with($route->uri, 'api/')) {
                 continue;
             }
 
-            // uri will still have the placeholders and wrong verbs, but we don't need them.
-            $request = Request::create($route->uri, 'GET');
-            $this->app->instance('request', $request); // set current request so is_api_request can work.
+            foreach ($route->methods() as $method) {
```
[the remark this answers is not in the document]
notbakaneko: but it can already use it at this point of the test...?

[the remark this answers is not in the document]
nanaya: Calling it in data provider means the test for each routes can be generated and run individually without manually collecting test exceptions

app/Providers/AuthServiceProvider.php
notbakaneko: yes, the grant is different

app/Libraries/RouteScopesHelper.php
notbakaneko: this doesn't require dealing with the file handle

peppy:
For existing grants what will they get after this is merged? Errors?

nanaya:
Error of missing required scope.

peppy:
And this will apply to basically 100% of current api consumers, correct?

nanaya:
Only for endpoints requiring `users.read` scope which are stuff under `/users/:id`.

peppy:
aha, so not `/me` then? i think that's fine.

Issue #3913, opened 2018-11-22
Disable textbox instead of not sending messages while muted


TPGPL:
![image](https://user-images.githubusercontent.com/19843905/48916642-5f316800-ee83-11e8-9887-3258cf49a53e.png)
![image](https://user-images.githubusercontent.com/19843905/48916652-65bfdf80-ee83-11e8-847a-1936c454ceb0.png)

nanaya:
Pretty sure this has been implemented a while back.

PR #5982, opened 2020-05-01
Show chat muted notice on existing channels

### LINKED ISSUE #5954: No note when replying to user blocking messages from non-friends
If a user is blocking messages from people not on their friends list (or has you blocked?) and messages you first, the note that normally appears under their name explaining why you can't message them does not appear. So when you go to reply, all of your messages fail/403 with no explanation and you are left wondering why that is the case. For example:
  

Typically you would see something like this:

nanaya:
Resolves #5954.

I sure hope the `moderated` column in Channel is properly set.


nekodex:
Uh, unless I'm mistaken... pretty sure nothing actually sets that moderated flag yet 😒

(IIRC it was only half-implemented as an emergency switch or something)

nanaya:
Well it doesn't matter if it's all false as the correct value for pm (which is the relevant part to this pr) is filled in differently 💯

app/Models/Chat/UserChannel.php
```diff
@@ -147,6 +152,7 @@ function ($userChannel) use ($byChannelId, $byUserId, $channelMessageIds, $userI
                     $userActual = $targetUser->userScoped;
                     $presence['icon'] = $userActual->user_avatar;
                     $presence['name'] = $userActual->username;
+                    $presence['moderated'] = !priv_check_user($user, 'ChatStart', $userActual)->can();
```
notbakaneko, line 155: should probably match this with sending messages which uses `ChatChannelSend`

nanaya: I think last I checked it is way more query heavy compared to this one

notbakaneko: it should probably include the `moderated` state still since that's part of the check when sending even if it's a pm channel

TPGPL:
Resolves #3913 as well?

PR #6008, opened 2020-05-14
Collapse long comments

### LINKED ISSUE #5948: Arrow on top right of comments should also collapse top level comment
Or you have long comments that can't be hidden. I'm proposing reddit behavior.

JNAPTBfeN1

hbowpdG4Ic

nanaya:
Resolves #5948?

There's no such thing as lines in html though, so...

The counting is done by wrapping it inside a container which width is the maximum allowed site-wide based on `@container-sm`.

Line height attribute is based on the line height of the first element. Images are hidden as we're not going to wait until they finish loading and they may also be shorter on narrower size due to keeping aspect ratio. Extra padding and margins, mixed line height, and various other factors will contribute in causing incorrect result.

Also note that this forces repaint so don't call this during initial react render or otherwise scroll position will be reset to 0. Mainly noticeable on browser history navigation.

And probably will slow things down (a bit?) thanks to the line counting and comments being rendered twice on initial render :snail:


resources/assets/coffee/react/_components/comment.coffee
```diff
@@ -55,32 +57,48 @@ export class Comment extends React.PureComponent
...
+      @setState lines: countMinLines(@props.comment.messageHtml)
+
+
   render: =>
     el Observer, null, () =>
       @children = uiState.getOrderedCommentsByParentId(@props.comment.id) ? []
       parent = store.comments.get(@props.comment.parentId)
       user = @userFor(@props.comment)
       meta = commentableMetaStore.get(@props.comment.commentableType, @props.comment.commentableId)
+      longContent = @state.lines? && @state.lines.count >= CLIP_LINES + 2
```
notbakaneko, line 87: could use a note on the `+ 2`

nanaya:
the difference is tiny for 7 lines and I don't want to go down the rabbit hole on figuring out which line-height to be used.

Also that part is already mentioned in the usage comment:
```
 * Extra padding and margins, mixed line height, and various other factors
 * will contribute in causing incorrect result.
```

notbakaneko:
so it should be `estimateNumLines` since it's not counting any minimum number of lines then?

PR #6010, opened 2020-05-14
Fix slow beatmap events listing

notbakaneko:
caused by #5963; the query builder causing the issue didn't change but ended up used in a condition that doesn't work very well.

- `BeatmapDiscussionPost::scopeVisible()` should not be called without filtering by `user_id`, and especially not when the condition has `whereHas('user' ...->default())`. The discussion post listing didn't use `visible()` to filter so it didn't have the same issue.
- posts not actually needed on listing page, so no longer fetched

Turns out the implicit `deleted_at` and `active` global scopes on `Beatmapset` has no index covering them making `whereHas('beatmapset')` not so great to use; if only one condition is included, it's fine, but having both is bad. 
Then there's inconsistency where `BeatmapDiscussion::beatmapset()` includes `withTrashed()` but `BeatmapsetEvent::beatmapset()` does not which is why some other places weren't affected...

### DIFF
--- a/app/Libraries/ModdingHistoryEventsBundle.php
+++ b/app/Libraries/ModdingHistoryEventsBundle.php
@@ -225,6 +225,10 @@ class ModdingHistoryEventsBundle
     private function getPosts()
     {
         return $this->memoize(__FUNCTION__, function () {
+            if (!$this->withExtras) {
+                return collect();
+            }
+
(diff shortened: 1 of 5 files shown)


notbakaneko:
Implicit global query scope turns out to be a trap waiting to explode in your face ಠ_ಠ

PR #6013, opened 2020-05-15
Show all user badges in profile page

### LINKED ISSUE #5739: Display multiple usergroup-badges in profile for users in multiple usergroups
uwu

people who belong to multiple usergroups should have their badges in profile.

since you can basically have a single user title, and some titles/groups have priority over others, having multiple usergroup badges would help identifying the user's roles.

nanaya:
Resolves #5739.

I'm not sure if all the badges also should be shown for modding discussion and forum post. Currently they still show only the first badge.


app/Models/User.php
```diff
@@ -1304,19 +1304,30 @@ public function blockedUserIds()
...
+            foreach (array_unique($ids) as $id) {
+                $group = app('groups')->byId($id);
+
+                if (optional($group)->display_order !== null) {
+                    $groups[] = $group;
+                }
+            }
+
+            usort($groups, function ($a, $b) {
+                return $a->display_order < $b->display_order ? -1 : 1;
```
notbakaneko, line 1327: can just be `$a->display_order - $b->display_order`?

PR #6020, opened 2020-05-17
Update permissions for setting genre and language

cl8n:
> [1:00 PM] yaspo: nat/gmt always: yes
> bn: wip/pending yes, qualified also yes
> mapper on anything that is their own and hasn't been nominated, correct
> also yeah it should be a lock

"lock" meaning you can't nominate without setting these first

to this I added project loved for loved, ~~and a BN who is also the mapset owner is treated as just the owner.~~ (yaspo said let BN do their thing even on their own maps)


cl8n:
I can't reproduce test errors here https://travis-ci.com/github/ppy/osu-web/jobs/335883681

notbakaneko:
Test database starts empty so it has no genres or languages by default

app/Libraries/OsuAuthorize.php
```diff
@@ -712,7 +724,18 @@ public function checkBeatmapsetMetadataEdit(?User $user, Beatmapset $beatmapset)
...
+        }
+
+        if ($user->isBNG() && in_array($beatmapset->approved, $bnEditable, true)) {
+            return 'ok';
+        }
+
+        if ($user->getKey() === $beatmapset->user_id &&
+            in_array($beatmapset->approved, $ownerEditable, true) &&
+            !$beatmapset->hasNominations()
+        ) {
```
notbakaneko, line 738: This check should be changed so the user gets a useful error message when the beatmapset has nominations instead of the generic unauthorized error.

tests/Models/BeatmapsetTest.php
```diff
@@ -214,4 +232,22 @@ private function fillNominationsExceptLast(Beatmapset $beatmapset, string $group
...
+    }
+
+    protected function tearDown(): void
+    {
+        Genre::find(Genre::UNSPECIFIED)->delete();
+        $this->fakeGenre->delete();
+        $this->fakeLanguage->delete();
+
+        parent::tearDown();
+    }
```
notbakaneko, line 252: also, this isn't needed

PR #6022, opened 2020-05-18
Limit fetching friends/foes data

nanaya:
Only fetch the ones to be used by ChatStart's hasBlocked/hasFriended.

### DIFF
--- a/app/Models/Chat/UserChannel.php
+++ b/app/Models/Chat/UserChannel.php
@@ -97,8 +97,12 @@ class UserChannel extends Model
             ->where('channels.type', '=', 'PM')
             ->with([
                 // only fetch data related to $user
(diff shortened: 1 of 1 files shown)


app/Models/Chat/UserChannel.php
```diff
@@ -95,7 +95,15 @@ public static function presenceForUser(User $user)
             })
             ->join('channels', 'channels.channel_id', '=', 'user_channels.channel_id')
             ->where('channels.type', '=', 'PM')
-            ->with(['userScoped.friends', 'userScoped.blocks'])
+            ->with([
+                // only fetch data related to $user, to be used by ChatStart privilege check
```
notbakaneko, line 99: I think it should mention it's where `$user` is the target of the friend/block since it's not getting any of the users `$user` has friended or blocked?

PR #6026, opened 2020-05-18
Fix username change test for user ID 1

cl8n:
always fails for me because I have a user with ID 1

should be safe to pull the existing user because this test is only checking that user_id is *one of* the validation errors

### DIFF
--- a/tests/Libraries/ChangeUsernameTest.php
+++ b/tests/Libraries/ChangeUsernameTest.php
@@ -16,7 +16,7 @@ class ChangeUsernameTest extends TestCase
 {
     public function testUserCannotBeRenamed()
     {
-        $user = $this->createUser(['user_id' => 1]);
+        $user = User::find(1) ?? $this->createUser(['user_id' => 1]);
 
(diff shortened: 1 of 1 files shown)


notbakaneko:
Tests are meant to be run with an empty database.

peppy:
If that's the case, can we throw an early warning when test are run while the database isn't empty?

As an aside, how do you guys run this when developing? Would be useful to document how it's intended to run tests during development process, since it sounds like it won't just work running it while you're working?

cl8n:
I thought it'd be okay to have this find-or-create logic because that's about what the Groups library is doing during tests, and it seems like a small convenience with no drawback

I did this in #6020 for the same reason, should I change it to a usual factory()->create ?

https://github.com/ppy/osu-web/blob/05dd422d5695d81fc4f2e2314e6759d38cae452d/tests/Models/BeatmapsetTest.php#L240

notbakaneko:
I'd rather we address this with properly documenting how tests should be run #6095 instead of adding additional workarounds in tests that we've been working on removing.

PR #6027, opened 2020-05-19
Fix livestream page

### LINKED ISSUE #6014: Live stream listing doesn't list Live streams
The [live stream listing](https://osu.ppy.sh/community/livestreams) doesn't list anything even tho multiple players are streaming for multiple hours already

nanaya:
Twitch updated their API and now requires proper OAuth tokens.

Fixes #6014.

Not the smartest way as the token is re-fetched when the cache expires (ignoring `expires` value of the token itself), but shouldn't be too bad 🤷 

Additional environment variable is now needed (@peppy):

```
TWITCH_CLIENT_SECRET=
```

To obtain (if missing):
- go to dev &rarr; console &rarr; applications (https://dev.twitch.tv/console/apps)
- manage the app
- click "new secret" (as there doesn't seem to be any way to view existing secret?)


peppy:
Added to .env for next deploy.

PR #6039, opened 2020-05-22
Trigger disqualification/reset of nominations when a review contains problem(s)

### LINKED ISSUE #6087: Adding discussions by editing reviews should create notifications
when adding a new suggestion/problem/praise by editing a review, a new notification should be created. the user can't know if a new discussion was created otherwise, and original notification doesn't update with # of issues

nekodex:
Also triggers notifications for aforementioned events.

- [x] depends on #5947

fixes #6087


nanaya:
conflicts :shrug:

resources/assets/lib/beatmap-discussions/editor-helpers.ts
```diff
@@ -31,6 +31,17 @@ export const toggleFormat = (editor: ReactEditor, format: string) => {
   );
 };
 
+export const slateDocumentContainsProblem = (input: SlateNode[]) => {
+  let containsProblem = false;
+  input.forEach((node) => {
```
nanaya, line 36: A bit simpler with [`Array#some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some).

app/Libraries/BeatmapsetDiscussionReview.php
```diff
@@ -97,6 +101,22 @@ public static function create(Beatmapset $beatmapset, array $document, User $use
...
                 ->update(['parent_id' => $review->getKey()]);
 
+            // handle disqualifications and the resetting of nominations
+            if ($problemPost) {
+                $resetNominations = $beatmapset->isPending() &&
+                    $beatmapset->hasNominations() &&
+                    priv_check_user($user, 'BeatmapsetResetNominations', $beatmapset)->can();
+
+                if ($resetNominations) {
+                    BeatmapsetEvent::log(BeatmapsetEvent::NOMINATION_RESET, $user, $problemPost)->saveOrExplode();
```
nanaya, line 111: looks like this should [broadcast reset nomination event](https://github.com/ppy/osu-web/blob/04df7c7a15f96357f3631a0157655a7661a54f70/app/Http/Controllers/BeatmapDiscussionPostsController.php#L170) as well?

There's also missing broadcast of [global notification on first new problem on qualified beatmapset](https://github.com/ppy/osu-web/blob/04df7c7a15f96357f3631a0157655a7661a54f70/app/Http/Controllers/BeatmapDiscussionPostsController.php#L185).

resources/assets/lib/beatmap-discussions/editor.tsx
```diff
@@ -378,6 +385,28 @@ export default class Editor extends React.Component<Props, State> {
...
+    const canReset = currentUser.is_admin || currentUser.is_nat || currentUser.is_bng;
+    const willReset =
+      this.props.beatmapset.status === 'pending' &&
+      this.props.beatmapset.nominations && this.props.beatmapset.nominations.current > 0 &&
+      docContainsProblem;
+
+    let confirmed = true;
+
+    if (canDisqualify && willDisqualify) {
+      confirmed = confirm(osu.trans('beatmaps.nominations.reset_confirm.disqualify'));
```
nanaya, line 401: skip `confirmed` variable and return immediately? Also in a quick glance it looks like there might be up to two confirmations (except it's not actually possible).

resources/assets/lib/beatmap-discussions/editor.tsx
```diff
@@ -247,21 +252,23 @@ export default class Editor extends React.Component<Props, State> {
...
-        data: { document: this.serialize() },
-        method: 'POST',
+    if (this.showConfirmationIfRequired()) {
+      this.setState({posting: true}, () => {
+        this.xhr = $.ajax(route('beatmapsets.discussion.review', {beatmapset: this.props.beatmapset.id}), {
+          data: {document: this.serialize()},
+          method: 'POST',
+        });
+
+        this.xhr.then((data) => {
```
nanaya, line 262: The indentation seems a bit weird on `.always` and `.catch`.

Additionally, `.done` instead of `.then` and `.fail` instead of `.catch` should return `JQueryXHR` so everything can be chained during initial assignment `this.xhr = `.

app/Libraries/BeatmapsetDiscussionReview.php
```diff
@@ -97,6 +101,22 @@ public static function create(Beatmapset $beatmapset, array $document, User $use
             BeatmapDiscussion::whereIn('id', $childIds)
                 ->update(['parent_id' => $review->getKey()]);
 
+            // handle disqualifications and the resetting of nominations
+            if ($problemPost) {
```
nanaya, line 105: This is for posting new review, but what about when updating an existing one creating new problem embed?

nanaya: Related issue: #6087.

app/Libraries/BeatmapsetDiscussionReview.php
```diff
@@ -246,6 +267,44 @@ private static function createPost($beatmapsetId, $discussionType, $message, $us
...
+        if ($resetNominations) {
+            BeatmapsetEvent::log(BeatmapsetEvent::NOMINATION_RESET, $user, $problemDiscussion)->saveOrExplode();
+            broadcast_notification(Notification::BEATMAPSET_RESET_NOMINATIONS, $beatmapset, $user);
+            $beatmapset->refreshCache();
+        }
+
+        if ($disqualify) {
+            $beatmapset->disqualify($user, $problemDiscussion);
+        } else {
+            if ($priorOpenProblemCount === 0) {
```
nanaya, line 298: not quite obvious but this is missing check `$beatmapset->isQualified()`

PR #6044, opened 2020-05-23
Change a scope from apiv2 reference 

ilsubyeega:
Since [this pull request](https://github.com/ppy/osu-web/pull/5981) made change some of the scope.


resources/views/vendor/apidoc/partials/info.blade.php
```diff
@@ -192,7 +192,7 @@
 --------------|-------------------------------
 identify      | Allows reading of the public profile of the authorizing user (`/me`).
 friends.read  | Allows reading of the authorizing user's friend list.
-users.read    | Allows reading of public user profiles on behalf of the authorizing user.
+public        | Allows reading of public user profiles on behalf of the authorizing user.
```
aspriddell, line 195:
```suggestion
public        | Allows reading of publicly available data on behalf of the authorizing user.
```
`public` scope can access things that aren't necessarily tied to a user, like spotlights or beatmaps

ilsubyeega: Thank you
committed

PR #6057, opened 2020-05-26
Fix unable to send message to closed PMs

### LINKED ISSUE #6043: Calling api chat/new will throw a 403 error if a channel already exists for a PM conversation
Currently a 403 is returned the second time this is called for the same user. This should gracefully succeed for private messages. The lazer client is unaware of whether there's an old PM channel lurking in the database (that has since been "left" by the user).

The code can return the existing channel and allow this operation to continue.

notbakaneko:
- rejoin PM channel if can `ChatStart`

closes #6043

`OsuAuthorize` cache should probably be reset before each request instead

### DIFF
--- a/app/Libraries/Chat.php
+++ b/app/Libraries/Chat.php
@@ -56,6 +56,8 @@ class Chat
             });
 
             Datadog::increment('chat.channel.create', 1, ['type' => $channel->type]);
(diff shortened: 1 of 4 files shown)


notbakaneko:
#6056 instead

Issue #6001, opened 2020-05-11
user_id is still mediumint in migrations


smoogipoo:
Should be `int`? I don't know what live is currently.

nanaya:
yes, the migration is rather outdated at the moment.

PR #6062, opened 2020-05-27
Synchronize database schema with production

nanaya:
# BACKUP YOUR DATABASE BEFORE RUNNING THE MIGRATION

This isn't full sync (as in it won't generate 0 diff between dev and prod dump) but should be pretty close.

Could be cleaned up further but not sure if worth it 🤷 
I kind of suspect this won't work on people's existing dev db. Sure could use some test on that.

Differences:
- character set and collation dump cleanup (see below)
- most of the partitions aren't copied over (some are rolloing so it's pointless trying to copy that)
- some unused tables (for osu-web) aren't copied over
- additional manual changes are needed on prod (see below)
- `osu_mirrors` has interesting default value on prod

Dump cleanup: (vim command)
```
:%s/ CHARACTER SET utf8mb4 / /g
:%s/ CHARACTER SET utf8mb4,/,/g
:%s/ CHARACTER SET utf8 / /g
:%s/ CHARACTER SET utf8,/,/g
:%s/ COLLATE utf8mb4_0900_ai_ci / /g
:%s/ COLLATE utf8mb4_0900_ai_ci,/,/g
:%s/ COLLATE utf8mb4_general_ci / /g
:%s/ COLLATE utf8mb4_general_ci,/,/g
:%s/ COLLATE utf8_general_ci / /g
:%s/ COLLATE utf8_general_ci,/,/g
:%s/ CHARSET=utf8mb4 / /g
:%s/ CHARSET=utf8mb4;/;/g
:%s/ CHARSET=utf8 / /g
:%s/ CHARSET=utf8;/;/g
:%s/ COLLATE=utf8mb4_0900_ai_ci / /g
:%s/ COLLATE=utf8mb4_0900_ai_ci;/;/g
:%s/ COLLATE=utf8mb4_general_ci / /g
:%s/ COLLATE=utf8mb4_general_ci;/;/g
:%s/ COLLATE=utf8_general_ci / /g
:%s/ COLLATE=utf8_general_ci;/;/g
:%s/ tinyint(\d*) / tinyint /g
:%s/ smallint(\d*) / smallint /g
:%s/ mediumint(\d*) / mediumint /g
:%s/ int(\d*) / int /g
:%s/ bigint(\d*) / bigint /g
:%s/ AUTO_INCREMENT=\d* / /g
```

(note that the resulting dump won't be valid. An additional `ENGINE=.*` removal might be needed)

Additional change on production (at least one is pretty big, I think):
```sql
-- should not be needed anymore as there's already similar index for (`user_id`,`updated_at`)
ALTER TABLE osu.beatmap_discussions DROP INDEX beatmap_discussions_user_id_foreign;
-- beatmap_id somehow has AUTO_INCREMENT attribute in production. Probably large table?
ALTER TABLE osu.osu_leaders_taiko CHANGE beatmap_id beatmap_id mediumint unsigned NOT NULL;
-- fixing mysql doing weird thing
ALTER TABLE osu_store.notification_requests CHANGE created_at created_at timestamp NULL DEFAULT NULL;
-- also fixing mysql doing weird thing
ALTER TABLE osu_store.notification_requests CHANGE updated_at updated_at timestamp NULL DEFAULT NULL;
```

Depends on #6052.
Closes https://github.com/ppy/osu-web/issues/6001.

- [x] add `2020_05_15_083037_sync_structure` to production `migrations` table

### DIFF
--- a/database/migrations/2020_05_15_083037_sync_structure.php
+++ b/database/migrations/2020_05_15_083037_sync_structure.php
@@ -1,0 +1,452 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class SyncStructure extends Migration
+{
+    const MODE_SUFFIXES = ['', '_fruits', '_mania', '_taiko'];
+
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('artist_albums', function (Blueprint $table) {
+            $table->dropForeign('artist_albums_artist_id_foreign');
+            $table->foreign('artist_id')
+                ->references('id')
+                ->on('artists')
+                ->onDelete('restrict')
+                ->onUpdate('restrict');
+        });
+        Schema::table('artist_tracks', function (Blueprint $table) {
+            $table->dropForeign('artist_tracks_artist_id_foreign');
+            $table->foreign('artist_id')
+                ->references('id')
+                ->on('artists')
+                ->onDelete('restrict')
+                ->onUpdate('restrict');
+            $table->dropForeign('artist_tracks_album_id_foreign');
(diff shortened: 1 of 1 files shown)


peppy:
Production changes have been applied.

peppy:
Does this require special handling from my end? Just usual migration row insertion?

notbakaneko:
just the migration row

PR #6064, opened 2020-05-28
Add endpoint returning playlist item's scores

notbakaneko:
Scores are returned in `total_score` descending order, with submission order tiebreaker.

No pagination yet.

- <del>add migration `2020_05_28_072624_add_multiplayer_scores_high_top_scores_index` </del> small enough should be fine being done during deploy (-nanaya)
  ```
  alter table `multiplayer_scores_high` add index `top_scores`(`playlist_item_id`, total_score DESC, `score_id`);
  ```


nanaya:
The table seems small enough it can probably be run during deploy.

PR #6076, opened 2020-06-01
Centre add button on beatmap reviews for now

peppy:
Just to make it feasible to target the button until we come up with something better.


peppy:
As an alternative, just have it never hide (and maybe style it to fit better, using a blue shade rather than white).

nanaya:
This doesn't actually center it for taller blocks (like embeds) :upside_down_face:

peppy:
Yeah, was aware of that. Just making it semi-usable with minimal effort.

Let's go with your fix.

PR #6091, opened 2020-06-02
Workaround cleaning up user score in elasticsearch

nanaya:
At least it didn't choke when I tried this on someone with ~1000 scores so it's probably safe for now?

Also this isn't the correct fix.

A full fix should do query without any hidden filter or aggregates on both es and db side and actually delete everything outdated in the es instead of first top _n_.


peppy:
What is this fixing again?

nanaya:
the score cleanup endpoint not completely cleaning up scores unless it's hit multiple times.

peppy:
Right, thanks.

PR #6092, opened 2020-06-02
Improve visibility of spaces in username in some places

### LINKED ISSUE #5942: usernames with multiple consecutive spaces get compressed
some examples: 
- https://osu.ppy.sh/users/723190 `-     -` shows up as `- -`
- https://osu.ppy.sh/users/1537821 `Bad   Apple` shows up as `Bad Apple`
- https://osu.ppy.sh/users/325462 `Bad  apple` shows up as `Bad apple`

Now these names can't be created anymore, but since they still exist I think they should be fixed.
Especially confusing when searching for "Bad Apple"

nanaya:
This is very not thorough.

Also I know I haven't updated page title for beatmap discussion page even though I updated the one inside the page itself.

Maybe enough to resolve #5942?

### DIFF
--- a/resources/assets/coffee/react/profile-page/header-info.coffee
+++ b/resources/assets/coffee/react/profile-page/header-info.coffee
@@ -36,7 +36,7 @@ export class HeaderInfo extends React.PureComponent
       div className: "#{bn}__details",
         h1
           className: "#{bn}__name"
-          span className: 'u-ellipsis-overflow', @props.user.username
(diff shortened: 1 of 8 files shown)


lmnyx:
Shouldn't this be applied to all possible user inputs or atleast replacing a lot of spaces to one in, i.e. forum post, profile bio, modding, map description, etc.?

peppy:
It's being applied to the places that matter the most. The users affected by this are legacy users – it's not supported in current username rules, so it's a lesser issue.

PR #6104, opened 2020-06-03
Trim topic title

### LINKED ISSUE #6100: Disable blank titles on forum posts
Currently it is possible to only use a space for the title of your post, making the line for the title disappear, messing up the layout. The time gets shifted upwards and when you edit the title the input field is misaligned

nanaya:
Resolves #6100.

### DIFF
--- a/app/Models/Forum/Topic.php
+++ b/app/Models/Forum/Topic.php
@@ -421,7 +421,7 @@ class Topic extends Model implements AfterCommit
 
     public function setTopicTitleAttribute($value)
     {
-        $this->attributes['topic_title'] = preg_replace('/(^\pZ+|\pZ+$)/u', '', trim($value));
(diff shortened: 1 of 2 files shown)


app/Models/Forum/Topic.php
```diff
@@ -419,6 +419,11 @@ public function setTopicLastPosterColourAttribute($value)
         $this->attributes['topic_last_poster_colour'] = ltrim($value, '#');
     }
 
+    public function setTopicTitleAttribute($value)
+    {
+        $this->attributes['topic_title'] = preg_replace('/(^\pZ+|\pZ+$)/u', '', trim($value));
```
notbakaneko, line 424: When `/u` is enabled `\s` becomes `\t\n\f\r\p{Z}` then `trim` shouldn't be needed anymore

PR #6115, opened 2020-06-04
Force line-wrapping for text within review embeds that are too long

### LINKED ISSUE #6099: In modding reviews, long texts, e.g. links, inside discussion does not wrap properly.
In modding reviews, long texts that cannot be broken down with space or hyphens, e.g. links, inside discussion does not wrap properly.

Landscape:
image

Portrait:
image

nekodex:
fixes #6099

### DIFF
--- a/resources/assets/less/bem/beatmap-discussion-review-post-embed-preview.less
+++ b/resources/assets/less/bem/beatmap-discussion-review-post-embed-preview.less
@@ -32,6 +32,7 @@
 
   &__message-container {
     flex: 1;
+    word-break: break-word;
 
     @media @desktop {
       align-self: center;


nanaya:
So I looked into the `word-break: break-word` because I vaguely remember it's deprecated. It is deprecated but the replacement ([`overflow-wrap: anywhere`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap)) isn't supported by Safari yet :pensive:

nekodex:
We're already using `word-break: break-word` in a bunch of places, so probably fine to just update them all at once when browser support is there?

notbakaneko:
the main reason we started moving away from `word-break: break-word` to a more convoluted wrapping model was because firefox didn't support `break-word`, but now it does ಠ_ಠ

nanaya:
Probably should update all the `word-wrap: break-word` to `word-break: break-word` (and remove `min-width`?)

PR #6116, opened 2020-06-04
Refactor BroadcastNotification job

notbakaneko:
Split notification job into separate classes where the class only needs to contain things relevant to that notification, instead of a bunch of unrelated code in the same class.

- [x] ppy/osu-notification-server#28


app/Jobs/Notifications/BroadcastNotificationBase.php
```diff
@@ -0,0 +1,137 @@
...
+     * In most cases this is a deduplicated list that excludes the user id that
+     * triggered the notifications. This should be overriden in cases where the source user id shouldn't be removed.
+     * e.g. UserAchievementUnlock.
+     */
+    public function getReceiverIds(): array
+    {
+        return array_values(array_unique(array_diff($this->getListeningUserIds(), [optional($this->source)->getKey()])));
+    }
+
+    public function getSource(): ?User
```
nanaya, line 91: never used?

app/Jobs/Notifications/CommentNew.php
```diff
@@ -0,0 +1,45 @@
...
+
+namespace App\Jobs\Notifications;
+
+use App\Exceptions\InvalidNotificationException;
+use App\Models\Follow;
+use App\Models\User;
+
+class CommentNew extends BroadcastNotificationBase
+{
+    public function __construct($object, User $source)
```
nanaya, line 14: isn't `$object` always `Comment`

app/Jobs/Notifications/ForumTopicReply.php
```diff
@@ -0,0 +1,53 @@
...
+
+use App\Models\Forum\Post;
+use App\Models\User;
+use App\Models\UserNotificationOption;
+
+class ForumTopicReply extends BroadcastNotificationBase
+{
+    public function __construct(Post $object, User $source)
+    {
+        parent::__construct($object, $source);
```
nanaya, line 16: I wonder if having `$notifiable` and `$post` (or `$comment` etc) instead of `getNotifiable` and `$object` would be easier to read.

Something like
```php
$this->post = $post;

parent::__construct($post->topic, $source);
```

app/Jobs/Notifications/ForumTopicReply.php
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public function getDetails(): array
+    {
+        return [
+            'title' => $this->notifiable->topic_title,
```
nanaya, line 22: continuing from above, there's no `$this->notifiable` here

app/Jobs/Notifications/UserAchievementUnlock.php
```diff
@@ -0,0 +1,44 @@
...
+namespace App\Jobs\Notifications;
+
+use App\Models\Achievement;
+use App\Models\User;
+
+class UserAchievementUnlock extends BroadcastNotificationBase
+{
+    public function __construct(Achievement $object, User $source)
+    {
+        parent::__construct($object, $source);
```
nanaya, line 15: the old one has source reset to new user so notification server will still forward it (there's user id check in `user-connection.ts`). Though considering notifications are now sent to per-user channel maybe the check is not needed anymore in the notification server side and this can stay the same user?

PR #6118, opened 2020-06-04
Add Sentry alert when user registered from unknown country

nanaya:
We're assuming every users have associated with a country in various places.


peppy:
Not sure sentry is the correct place for this, but let's see how it goes.

Issue #4050, opened 2018-12-31
Change the behavior of [profile] bbcode tags


cl8n:
Even though these tags are supposed to be a convenient way to link users, they're usually not preferred because the link will break if the user changes their username

I think it would make more sense to convert these at post time to ordinary links (that go to a profile by user ID), and then just have the user card show on hover for any /u/ or /users/ link

peppy:
Should definitely be converted at time of post.

PR #6137, opened 2020-06-08
Convert [profile] bbcode to [url] at post time

cl8n:
backend part of #4050 . now on the forum they end up missing the bold and user card since they're plain urls, but that should be the behavior of any url with a user link

### DIFF
--- a/app/Libraries/BBCodeForDB.php
+++ b/app/Libraries/BBCodeForDB.php
@@ -230,11 +230,13 @@ class BBCodeForDB
 
                 if ($user === null) {
                     $name = $this->extraEscapes($m[1]);
+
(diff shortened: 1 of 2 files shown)


nanaya:
Before you go on creating a url parser for that user card, maybe try extending the syntax to allow passing uid `[profile=uid]uname[/profile]` and automatically add it on store if missing.

cl8n:
that'd work too, but I was thinking the user card would be wanted even if you wrote in a user [url]. if that's not the case I'll change

nanaya:
I have a feeling adding support for user card on any (forum post?) links will be rather complex.

nanaya:
Apparently I forgot to mention that I need to see how the front-end change look like first before continuing this.

PR #6138, opened 2020-06-08
Add option to allow guest search

### LINKED ISSUE #6121: Config option to enable anonymous beatmap listing searches
Originally mentioned in https://github.com/ppy/osu-web/issues/5924

I'd really like a way to enable anonymous beatmap listing searches for my deploys. This used to be possible once-upon-a-time via changing `app/Libraries/Search/BeatmapsetSearchRequestParams.php`, but that technique doesn't seem to work anymore :(

nanaya:
I probably broke a few things in the process. Resolves #6121.

Had to call it "advanced search" because pagination without filter/keyword search for guest is also "search".


smoogipoo:
Looks to be working fine.

PR #6170, opened 2020-06-12
Perform portal cleanup when navigating

### LINKED ISSUE #6162: Editor toolbar in beatmap discussion review stuck on navigation
One can collect toolbars with a few link navigation & back from discussion page (it seems somehow fine using browser's forward after first nav & back).

Probably because Portal component never cleans up on unmount.

nekodex:
fixes #6162


resources/assets/lib/portal.tsx
```diff
@@ -1,7 +1,37 @@
...
+  }
+
+  addPortal = () => document.body.appendChild(this.container);
+
+  componentDidMount() {
+    this.addPortal();
+
+    $(document).on(`turbolinks:before-cache.${this.uuid}`, () => {
+      this.removePortal();
+    });
```
nanaya, line 29:
```suggestion
    $(document).on(`turbolinks:before-cache.${this.uuid}`, this.removePortal);
```

resources/assets/lib/portal.tsx
```diff
@@ -1,7 +1,37 @@
...
+  }
+
+  componentWillUnmount = () => $(document).off(`turbolinks:before-cache.${this.uuid}`);
```
nanaya, line 32: this should call `removePortal` as well for when it's unmounted in other cases like unselecting the text.

resources/assets/lib/portal.tsx
```diff
@@ -1,7 +1,39 @@
...
+    $(document).on(`turbolinks:before-cache.${this.uuid}`, this.removePortal);
+  }
+
+  componentWillUnmount = () => {
+    this.removePortal();
+
+    $(document).off(`turbolinks:before-cache.${this.uuid}`);
+  }
+
+  removePortal = () => document.body.removeChild(this.container);
```
nanaya, line 36: this causes error on navigation (presumably as it tries to run this on unmount):

```
NotFoundError: Node.removeChild: The node to be removed is not a child of this node
```

Issue #5852, opened 2020-03-31
Increase the variety of languages and genres available


Raidencio:
in light of the recent update that allows mappers to set their own genres and language, it is widely known that the number of both of those are fairly limited, and some songs have to be classified into "other" or "novelty", while there could be much more accurate genres to describe them

Some of the new genres could be:
- Vocaloid
- Metal
- Classical/Orchestral
- Dubstep 
- Psytrance
- Drum and bass
- Breakcore
...

(will expand this list with community input)

Some of the new languages could be the ones available in the page (https://i.imgur.com/oyJGQpn.png)

notRealazy:
adding more **broad** genres would be nice, but i don't think we should try to delve into more precise ones, mostly looking at electronic genres. if we add dnb, dubstep, breakcore, people might wonder why we don't add drumstep, riddim, mashcore, or other genres like techno/house/etc. 

more genres = good, too many precise genres = bad

Raidencio:
makes sense, what broad genre should we add that encompasses multiple electronic subgenres?

currently "Electronic" is a little ?? because you can pretty much stuff anything in there that is not recorded in a studio

frukoyurdakul:
At its current state I think the suggestions only contain "Vocaloid, Metal, Classical / Orchestral, Breakcore" as the rest can fall into more general categories.

dorothy3242:
I think
Metal, Classical, Folk
would be good additions for musical genres
a lot of folk and classical stuff currently get categorized weirdly due to lacking categories, and metal gets thrown into rock for lack of better tag. Some of electronic's weirdness is a result of lacking tags for piano pieces as well. 

I agree it shouldn't be too specific unless the genre system was revamped entirely - allowing the user base to add the genre tags, easily accessible genre trees, allowing more than 1 genre to be assigned to a map, etc. Not really sure how beneficial that would actually be.

Vocaloid would be kind of weird since it encompasses all genres, and Vocaloid itself is presently dying down in favor of other singing synthesizer programs (CeVIO, piapro, synthesizer V, NEUTRINO, etc). We'd be better off recategorizing existing Vocaloid tracks to other genres appropriately.

frukoyurdakul:
[...]

Issue #6019, opened 2020-05-17
Add "Unspecified" language for beatmaps, and make it the default setting


cl8n:
default genre and language for maps is Unspecified and Other respectively, but the default to Other isn't very helpful because that actually is a valid language setting on a lot of maps. searching for language Other is useless right now

I think it would be best to

- add "Unspecified" language
- make that the default for submitted maps
- set all maps with genre+language Unspecified+Other to Unspecified+Unspecified

cl8n:
resolved

PR #6176, opened 2020-06-13
Add new beatmap genres and languages

cl8n:
part of #5852

note changes to language display order, I kept instrumental and other at the end

and ofc needs live db edits to change anything


cl8n:
gonna have to add similar things for #6019 so I'll do it in this PR too

app/Models/Language.php
```diff
@@ -18,6 +18,8 @@ class Language extends Model
     protected $primaryKey = 'language_id';
     public $timestamps = false;
 
+    const UNSPECIFIED = 14;
```
peppy, line 21: should we make this 0 instead?

PR #6205, opened 2020-06-18
Combine database-based cursor handling to a helper

nanaya:
This will help adding the cursor pagination for multiplayer scores.

I was hoping it'd net negative line changes but it's not ( ﾟ ヮﾟ)


app/Models/Model.php
```diff
@@ -77,6 +77,17 @@ public function refresh()
...
+    public function scopeCursorSort($query, array $sort, ?array $cursor)
+    {
+        if (empty($cursor)) {
+            foreach ($sort as $sortItem) {
+                $query->orderBy($sortItem['column'], $sortItem['order']);
+            }
+        } else {
+            $query->cursorWhere($cursor);
+        }
+    }
```
notbakaneko, line 89: this part seems weird since `$sort` gets ignored if `$cursor` is set but `$cursor` always seems to be `$sort` + extra `value` key when not `null`, making it pretty much `scopeCursorWhere` where the `where` part is skipped if there's no value? 🤔 ❓

PR #6210, opened 2020-06-19
Add interop endpoint for user creation

nanaya:
- url is `_lio/users`, method is `POST`
- copy some attributes (ip address, country, password, email) from another user with `source_user_id` parameter
- setting additional attributes:
  - user group (`user[group]` parameter, by group `identifier`)
  - country (`user[country_acronym]` parameter)
- ip address set to `127.0.0.1`

Removed `Legacy` part in the new namespace because it (and some others) isn't actually legacy :shrug:


app/Http/Controllers/InterOp/UsersController.php
```diff
@@ -0,0 +1,50 @@
...
+            $sourceUser = User::find($request['source_user_id']);
+
+            if ($sourceUser === null) {
+                abort(404, "Can't find User specified in source_user_id");
+            }
+
+            if (!isset($params['user_email'])) {
+                $sourceEmailParts = explode('@', $sourceUser->user_email);
+                $params['user_email'] = "{$sourceEmailParts[0]}+{$params['username']}@{$sourceEmailParts[1]}";
+            }
```
notbakaneko, line 32: will explode if $params['username'] wasn't set

PR #6212, opened 2020-06-19
Add scores#show endpoint and include surrounding scores and position where relevant

### LINKED ISSUE #6102: Timeshift scores indexing returns too many scores
Currently seems to just return all scores in the one response, which is going to be a lot of data when spotlights get released. We'll probably want a similar cursor to other requests, however:

1. In the case that the user completes a beatmap, we probably want to index N scores around the user's score. We'd need to have the ability to re-index in both directions, depending on the direction in which the user scrolls the display (so we could be loading scores > user score or scores < user score).
2. In the case that the user opens up the results for a beatmap, we probably want to start at the top score?

I think both of these could be done by sending a nullable `score_id` to the server in the initial request.

nanaya:
- [x] depends on #6209

The surrounding scores and position which should resolve #6102.


nanaya:
I can't convince mysql to use the new index and the filesort is like for 11 rows anyway and `filesort` apparently also means in-memory sorting.

notbakaneko:
It's still using the existing index to get a limited set of rows first, so I guess it should be fine 🤔

PR #6223, opened 2020-06-23
Add multiplayer rooms and scores to public scope

### LINKED ISSUE #6055: Make timeshift lobby scores available via public scope API
In lights of the upcoming changes to the Beatmap Spotlights system, we need a feasible way to extract the individual scores on beatmaps of timeshift lobbies in osu!lazer. For that, we would need to be able to target our API call to a specific timeshift lobby and pull the information we need. 

There are currently viable API v2 endpoints such as `/rooms/owned` and `/rooms/{room_id}/playlist/{playlist_id}/scores`, however executing the GET request with `public` scope authentication returns a generic "Invalid scope(s) provided." error message. I assume the information is not accessible via any scope right now.

notbakaneko:
I thought I did this already but apparently not 🤔 
There should be anything particularly special in the response.

closes #6055

### DIFF
--- a/app/Http/Controllers/Multiplayer/RoomsController.php
+++ b/app/Http/Controllers/Multiplayer/RoomsController.php
@@ -14,7 +14,7 @@ class RoomsController extends BaseController
     public function __construct()
     {
         $this->middleware('auth');
-        $this->middleware('require-scopes:public', ['only' => ['index', 'leaderboard', 'public']]);
+        $this->middleware('require-scopes:public', ['only' => ['index', 'leaderboard', 'show']]);
     }
 
     public function index()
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Multiplayer/RoomsController.php
```diff
@@ -14,6 +14,7 @@ class RoomsController extends BaseController
     public function __construct()
     {
         $this->middleware('auth');
+        $this->middleware('require-scopes:public', ['only' => ['index', 'leaderboard', 'public']]);
```
nanaya, line 17: action `#public`? ?_?

Issue #4317, opened 2019-04-09
Simpler forum authorization


cl8n:
I wanted to look into replacing phpbb forum auth soon (as mentioned by lots of comments saying "this is temporary" etc), but I'm not sure which features are needed

a comment says that user-specific ACLs are no longer used but I don't have info beyond that. I think all i'd need is a list of the permission types still in use and whether or not "roles" are necessary anymore

peppy:
One prerequisite for this is to turn off the old forums, i believe. And I think we are actually ready to do this.

As for roles, we still do have forum-group specific roles (ie support team with help forum access) so we'd likely need to consider that. A table with `group_id`, `forum_id` and then a number of `bool` columns for specific permissions may be enough

peppy:
cc/ @nanaya in case he can think of any edge cases or special requirements i may have missed.

cl8n:
by roles I meant AuthRoles which seem to give global permissions for usergroups. to my knowledge all of this is already taken care of by OsuAuthorize so I think we can get rid of it?

nanaya:
Some checks in OsuAuthorize still use old forum acl.

peppy:
This can be reconsidered now, if there is interest. Removing ACL completely would be a huge step forward.

PR #6232, opened 2020-06-25
Add badge for moderators of forums

cl8n:
you can't tell who moderators of forums are unless they're in one of the global mod groups, so this adds a fake-group badge to show that. works and looks similar to "Mapper" for beatmap discussions

### DIFF
--- a/resources/views/forum/topics/_post.blade.php
+++ b/resources/views/forum/topics/_post.blade.php
@@ -19,6 +19,13 @@
     }
 
     $user = $post->userNormalized();
+    $group = isset($topic) && $user->isModeratorForForum($topic->forum)
+        ? (object)[
+            'colour' => 'hsl(200, 60%, 50%)',
+            'group_name' => "{$topic->forum->forum_name} Forum Moderator",
+            'short_name' => 'MOD',
+        ]
+        : ($user->visibleGroups()[0] ?? null);
 ?>
 <div
     class="js-forum-post {{ $post->trashed() ? 'js-forum-post--hidden' : '' }} forum-post"
--- a/resources/views/forum/topics/_post.blade.php
+++ b/resources/views/forum/topics/_post.blade.php
@@ -32,7 +32,7 @@
     data-post-id="{{ $post->getKey() }}"
     data-post-position="{{ $options["postPosition"] }}"
 >
-    @include('forum.topics._post_info', compact('user'))
+    @include('forum.topics._post_info', compact('user', 'group'))
 
     <div class="forum-post__body js-forum-post-edit--container">
         <div class="forum-post__content forum-post__content--header">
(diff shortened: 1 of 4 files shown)


cl8n:
I've gotten the (very) occasional PMs where someone is confused about not being able to make new threads on the Loved forum, it seems like the confusion comes from me looking like a normal user and I make all the threads there. that was the immediate reason I just made this change, but in general it seems like forum moderators are pretty identifiable--- they've all got badges of some kind and you can look at their group listings

I don't have a very strong opinion about this but I thought it'd be making things more consistent.

peppy:
Lets go with this 👍 .

peppy:
On further internal discussion, it's looking like all forum moderators are already in a group which is exposed, apart from the "loved" forum group. Is there another case you are aware of @cl8n?

resources/views/forum/topics/_post.blade.php
```diff
@@ -19,6 +19,13 @@
     }
 
     $user = $post->userNormalized();
+    $group = isset($topic) && $user->isModeratorForForum($topic->forum)
+        ? (object)[
+            'colour' => 'hsl(200, 60%, 50%)',
```
notbakaneko, line 24: Where's this colour coming from?

resources/views/forum/topics/_post.blade.php
```diff
@@ -19,6 +19,13 @@
     }
 
     $user = $post->userNormalized();
+    $group = isset($topic) && $user->isModeratorForForum($topic->forum)
```
notbakaneko, line 22: should also explicitly pass `group` into the `@include('forum.topics._post_info', ...` part

resources/views/forum/topics/_post.blade.php
cl8n: https://github.com/ppy/osu-web/blob/b460e4853e0689420b7888a33d106b8ee33c1eff/resources/assets/lib/beatmap-discussions/mapper-group.ts#L7
was just to get it working, idk if that makes sense for design

cl8n:
if there haven't been any changes since `moderator_groups` was added to forums, then yeah loved is the only one. maybe this display would make more sense after https://github.com/ppy/osu-web/issues/4317#issuecomment-481101956, where you might have certain users moderating forums without belonging to a group

notbakaneko:
this was waiting for some feedback from @peppy, I think?

cl8n:
@peppy 🏓

peppy:
I honestly don't know if this is necessary or just added complexity. Tempted to lean in the direction of #2 based on how few people this affects.

peppy:
I think the consensus here is that this is *probably* not required, so going to close for the time being. We can re-visit if this applies to more user groups in the future.

PR #6237, opened 2020-06-26
Add OAuth client credentials grant support

### LINKED ISSUE #5798: Allow client grant API access
Going through client auth process isn't convenient for serverside services, but currently the only API client that is allowed to use password grant type is lazer and iirc using lazer id/secret isn't allowed.

notbakaneko:
Adds support for OAuth `client_credential` tokens. 
Tokens using `client_credential`s do not have a `User` assigned and are treated as guest users.

Originally, this was going to extend the `CheckCredentials` middleware, but I also wanted the OAuth token being applied for authentication and the route token scope checks to be separate so that it was easier to isolate issues.

Using the default `SessionGuard` for API requests produces the desired result, but it still doesn't behave the way it should internally - it works because the previous middleware uses a null session driver and removes the cookies.
`api` guard is now useless.

closes #5798

- [x] requires #6233


app/Http/Middleware/AuthApi.php
```diff
@@ -6,35 +6,74 @@
...
+    public function __construct(ResourceServer $server, TokenRepository $repository)
+    {
+        $this->server = $server;
+        $this->repository = $repository;
     }
 
-    public function handle(Request $request, Closure $next)
+    public function handle($request, Closure $next)
     {
         auth()->shouldUse('api');
```
nanaya, line 34: I haven't checked deeper but I suspect you can remove this and set user manually when validating token instead. Something like this: https://0paste.com/72919-e7aaf5a7.txt

Although even then this still does query same token twice similar to https://github.com/laravel/passport/issues/382 but in this class' `validTokenFromRequest` instead (in addition to oauth library's).

nanaya: at over 600 queries for beatmapsets search endpoint using client cred, I think this needs to be fixed

PR #6239, opened 2020-06-26
Temporary workaround for querying beatmapset search result from different connection

nanaya:
Quality code™.


peppy:
I'm going to leave this open and deploy direct to server for now 👍

peppy:
We'll probably use the new db setup to handle this instead (using proxysql + query rules).

peppy:
Can resolve this externally now.

PR #6244, opened 2020-06-29
Fix logout incorrectly called on guards without the method

notbakaneko:
`RequestGuard` used by API doesn't have a `logout()` method. This wasn't an issue previously because the `CheckUserBanStatus` middleware didn't apply properly in API requests due to the user not being resolved early enough.


nanaya:
Probably should fix the response for api as well

PR #6259, opened 2020-07-03
Add block count display to the review editor

nekodex:
Also disables posting if block limit is exceeded (for new reviews, not when editing yet... will do that separately)

Jul-03-2020 19-17-51


peppy:
I'd hope we have the count set high enough this never becomes an issue. Not sure how useful having it displayed is.

nanaya:
I wonder if we should just exclude paragraphs from being counted toward number of blocks.

nekodex:
Can you elaborate on why you think this is unnecessary?

I'd argue that giving an indication to the user that they're about to hit a limit is infinitely better than throwing an ugly error popup up after the fact...

Is the issue that it's always visible? If so I can hide it until the user starts approaching the limit or something?

peppy:
As above, I think the limit should be so high the user should not care about hitting it

nekodex:
It's already high enough that most users won't hit it, but that doesn't make it impossible to hit.

This gives better feedback so they can address it _before_ attempting to post and getting slapped in the face.

peppy:
I guess only showing when getting really close (last 10) or something?

nanaya:
conflicts :pensive:

resources/assets/less/bem/beatmap-discussion-editor.less
```diff
@@ -50,6 +74,13 @@
     }
   }
 
+  &__inner-block-count {
```
nanaya, line 77: collides with the new insertion bar and a bit too close to the unlink button

PR #6271, opened 2020-07-06
Add option to display beatmap in original language

### LINKED ISSUE #3217: Add option to display beatmap metadata in original language
Currently we're not displaying beatmap titles/etc in their original language. An account-wide setting should probably be added to allow users to toggle whichever they'd prefer.

notbakaneko:
Maybe a more easily accessible toggle instead of stowed away in the settings page would be better? 🤔 
(and maybe better/more generic option name)

closes #3217


resources/assets/lib/utils/beatmap-helper.ts
```diff
@@ -69,6 +70,15 @@ export function getDiffRating(rating: number) {
   return 'expert-plus';
 }
 
+// TODO: should make a Beatmapset proxy object or something
+export function getTitle(beatmapset: BeatmapsetJson) {
```
nanaya, line 74: php side needs this as well for (at least) `beatmapsets.show` and `scores.show` (html) page title.

app/Models/Beatmapset.php
```diff
@@ -1088,6 +1088,16 @@ private function getBBCode()
         return new BBCodeFromDB($description, $post->bbcode_uid, $options);
     }
 
+    public function getDisplayTitle()
+    {
+        $showOriginal = auth()->user()->userProfileCustomization->options['beatmapset_title_show_original'] ?? false;
```
nanaya, line 1093: I just remembered about title in notification. I think that should follow this setting as well?

(And thus this function probably should accept argument `$user` and use it instead. Accessing session user within model seems bad)

Also the attribute can be accessed by `...Customization->beatmapset_title_show_original`.

Going a bit further, the `userProfileCustomization` can use `$user->profileCustomization()` instead to be guaranteed to be not null. And get the default directly from the relevant model.

notbakaneko: `profileCustomization`'s conditional insert sounds like a terrible idea with notifications for other users

nanaya: Apparently it's already being called when loading the page itself ( ﾟ ヮﾟ)

notbakaneko: which has nothing to do with the user receiving the notification where the setting is going to be on?

nanaya: hmm maybe just update it to stop creating a new entry. Still weird defining default for same thing in two different places.

notbakaneko: I wonder if it's better to include `title_unicode` in the notification instead, otherwise, the previous notifications would be displaying based on whatever the setting was when it was sent instead of what the current setting is (and being completely ignored for the ones already without the extra data)

nanaya: sure, if you can figure out how to do it with mail notification :upside_down_face: 

(although that sounds a bit like waste of space)

PR #6278, opened 2020-07-08
Add category and pagination to multiplayer rooms

### LINKED ISSUE #6226: Pinnable timeshift rooms
With the upcoming spotlights, it would be nice to have the ability to pin timeshift rooms to the top of the listing. This should probably just be a database flag that's returned via the `/rooms` endpoint.

nanaya:
- nowhere to set it apart of manually editing database
- use parameter `category` in rooms listing endpoint to filter by category (normal or spotlight)
- add support for cursor based pagination. Sort option is implied by existing parameter `type`
- cursor for more results will need to be built manually
  - for type `ended`, following are required:
    - `cursor[ends_at]`
    - `cursor[id]`
  - for other types:
    - `cursor[id]`
- I can't add those cursor hint in the response without breaking current api ( ﾟ ヮﾟ)

Resolves #6226.

### DIFF
--- a/app/Models/Multiplayer/Room.php
+++ b/app/Models/Multiplayer/Room.php
@@ -60,7 +60,7 @@ class Room extends Model
 
         switch ($mode) {
             case 'ended':
-                $query->ended()->orderBy('ends_at', 'desc');
+                $query->ended();
                 $sort = 'ended';
                 break;
             case 'participated':
--- a/app/Models/Multiplayer/Room.php
+++ b/app/Models/Multiplayer/Room.php
@@ -6,6 +6,7 @@
 namespace App\Models\Multiplayer;
 
 use App\Exceptions\InvariantException;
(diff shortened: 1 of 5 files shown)


app/Models/Multiplayer/Room.php
```diff
@@ -33,12 +34,66 @@ class Room extends Model
...
+                $query->hasParticipated($user);
+                break;
+            case 'owned':
+                $query->startedBy($user);
+                break;
+            default:
+                $query->active();
+        }
+
+        $category = presence(get_string($params['category'] ?? null));
```
notbakaneko, line 76: Apparently the plan is not to have multiple categories in the same list, so this should default to `normal` or something (also solves issue of index not being used when ordering by `ended_at`)

smoogipoo:
Tested the categories client-side and it's going to work :+1:

PR #6279, opened 2020-07-08
Timeshift rankings

nanaya:
- [x] Depends on #6278
- [x] Decide what the tab should be called (currently "multiplayer")


peppy:
"multiplayer" is fine for now

PR #6283, opened 2020-07-09
Fix navigation after blocking a user

### LINKED ISSUE #6280: Viewing a blocked profile and then going onto a different page causes the website to not load
Once you block someone and leave the page, it causes the web page you're going to to not load entirely, forcing you to refresh. This bug has been around for a while and I thought someone had reported it but I couldn't find it anywhere and it's incredibly easy to reproduce.

**How to reproduce:**

Step 1: [Block someone.](https://i.imgur.com/vKUvfJ2.png)

Step 2: Go to a different page. Any page on the website.

Step 3: Congratulations. [Bug reproduced.](https://i.imgur.com/ZryKzt1.png)

Hitting F5, as mentioned, makes the page load. I've managed to replicate this both on Mozilla Firefox and on Google Chrome, as well as mobile and different computers. It doesn't matter if you're running scripts like osuplus or any Greasemonkey scripts either, this bug persists in all use cases.

notbakaneko:
fixes #6280

### DIFF
--- a/resources/assets/coffee/react/_components/notification-banner.coffee
+++ b/resources/assets/coffee/react/_components/notification-banner.coffee
@@ -38,7 +38,8 @@ export class NotificationBanner extends React.PureComponent
 
 
   removePortalContainer: =>
-    notificationBanners[0].removeChild @portalContainer
+    if @portalContainer.offsetParent?
+      notificationBanners[0].removeChild @portalContainer
 
 
   createPortalContainer: =>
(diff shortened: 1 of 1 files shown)


resources/assets/coffee/react/_components/notification-banner.coffee
```diff
@@ -38,7 +38,8 @@ export class NotificationBanner extends React.PureComponent
 
 
   removePortalContainer: =>
-    notificationBanners[0].removeChild @portalContainer
+    if @portalContainer.offsetParent?
+      notificationBanners[0].removeChild @portalContainer
```
nanaya, line 42: `.offsetParent` also sometimes returning null even if parent still exists aside...

```suggestion
    @portalContainer.remove()
```

PR #6287, opened 2020-07-10
Fix timestamps of embedded issues not being displayed when editing a review

nekodex:


nanaya:
I can't reproduce the problem :question:

nekodex:
Jul-22-2020 14-57-59

resources/assets/lib/beatmap-discussions/review-document.ts
```diff
@@ -73,7 +73,7 @@ export function parseFromJson(json: string, discussions: Record<number, BeatmapD
           }],
           discussionId: discussion.id,
           discussionType: discussion.message_type,
-          timestamp: discussion.timestamp,
+          timestamp: discussion.timestamp ? BeatmapDiscussionHelper.formatTimestamp(discussion.timestamp) : null,
```
nanaya, line 76: This doesn't work when timestamp is `00:00.000` (because 0 is falsy).

Also the helper returns `undefined`  for null-ish timestamp so the check here can be skipped altogether. The function signature will need an update though. And maybe update the helper to return `null` instead.

resources/assets/lib/beatmap-discussions/editor-discussion-component.tsx
```diff
@@ -26,7 +26,9 @@ export default class EditorDiscussionComponent extends React.Component<Props> {
 
   componentDidMount = () => {
     // reset timestamp to null on clone
-    Transforms.setNodes(this.context, {timestamp: null}, {at: this.path()});
+    if (this.editable()) {
```
nanaya, line 29: looks like it is reset again when navigated using keyboard (by pressing left arrow key several times)

PR #6298, opened 2020-07-13
Additionally sort by score id for multiplayer room rankings

### LINKED ISSUE #6293: New spotlight leader board (seemingly) randomly changes leader board position of users with identical score
When two players have identical scores, they change leader board place seemingly randomly. Not sure why this happens.

[Here](https://imgur.com/gallery/vjSwylP) is an imgur album showing me and idealism changing leaderboard position, despite neither of us setting a score.

Afaik there is no lazer version/log for this.

nanaya:
Updating the table may take a while :thinking: 

`updated_at` can't be used because any attempts will update the row.

Resolves #6293.


nanaya:
Seems fine to be run directly as it's not too big.

PR #6302, opened 2020-07-14
Fix overflow in footer

### LINKED ISSUE #5819: Incorrect site width on full hd+ screens (mobile)
On my Honor 8x smartphone, the screen resolution is 2340x1080, which is why the site is displayed wider than the actual screen resolution
Screenshot_20200325_145123_com android chrome
Screenshot_20200325_145119_com android chrome

gagahpangeran:
Fix #5819 

**Before**

image

**After**

image

### DIFF
--- a/resources/assets/less/bem/footer.less
+++ b/resources/assets/less/bem/footer.less
@@ -24,13 +24,14 @@
     margin-bottom: 5px;
     display: flex;
     justify-content: center;
+    flex-wrap: wrap;
   }
 
   &__link {
     .link-plain();
     font-weight: bold;
     color: white;
-    margin: 0 10px;
+    margin: 5px 10px;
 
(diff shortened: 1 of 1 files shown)


resources/assets/less/bem/footer.less
```diff
@@ -24,13 +24,14 @@
...
     justify-content: center;
+    flex-wrap: wrap;
   }
 
   &__link {
     .link-plain();
     font-weight: bold;
     color: white;
-    margin: 0 10px;
+    margin: 5px 10px;
```
nanaya, line 34: this adds extra margin on normal pages and especially messes up footer on chat page

gagahpangeran: I'll change it to mobile only

PR #6305, opened 2020-07-14
Throttle sending chat read marker

nanaya:
steps:
- open busy chatroom
- receive multiple messages in single update

result:
- multiple mark as read requests are sent (one for each messages)

This PR adds a second of waiting before finally sending the mark read request.


resources/assets/lib/chat/chat-orchestrator.ts
```diff
@@ -157,28 +158,38 @@ export default class ChatOrchestrator implements DispatchListener {
...
+
+      const lastSentMessage = channel.messages[channel.messages.length - 1];
+      if (lastSentMessage && lastSentMessage.sender.id === window.currentUser.id) {
         channel.lastReadId = lastReadId;
-      })
-      .catch((err) => {
-        console.debug('markAsRead error', err);
-      });
+
+        return;
```
notbakaneko, line 179: need to unset the channel from `markingAsRead` here as well or it'll get stuck when the latest message is yours when it runs.

PR #6306, opened 2020-07-14
Change view ranking filter header

### LINKED ISSUE #6201: Add a clear distinction for filter headers
image

Currently the filter headers `Variant` and `Show` blend into other filter options and it's not obvious which one is not the option.

gagahpangeran:
Closed #6201 

I only change based on the design and don't modify `Sort` component to prevent changes in other page.

**Desktop**

image

**Mobile**

image

The mobile view is cover up half of the screen. Should I make it horizontal?

### DIFF
--- a/resources/assets/less/bem/ranking-filter.less
+++ b/resources/assets/less/bem/ranking-filter.less
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 .ranking-filter {
+  @top: ranking-filter;
   --gutter: 10px;
   display: flex;
   flex-direction: column;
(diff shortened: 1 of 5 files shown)


notbakaneko:
I think the vertical padding can be reduced all around and the title text should be a bit smaller.
Title when in mobile view should probably be inline

gagahpangeran:
image

image

resources/assets/less/bem/ranking-filter.less
```diff
@@ -13,17 +14,38 @@
...
+      display: block;
+
       @media @desktop {
         flex: 1;
       }
+
+      .@{top}__item--title {
+        padding: 0;
+        margin-bottom: 10px;
+      }
```
notbakaneko, line 48: This should be inside `&--title` as `.@{top}__item--full &` instead.

gagahpangeran:
Feels empty

image

I can't check spotlight page yet because the spotlight seeder is error.

gagahpangeran:
All done! Please check again.

PR #6309, opened 2020-07-15
Fix quick search bug when using space keyword only

### LINKED ISSUE #6308: Quick search bug when click space button
- Click quick search from any page.
- Click `space` button.
- The quick search popup suddenly close and search icon is gone.
- After that quick search popup doesn't appear when you click at any page and redirect to search page.
- If you open beatmap listing page after this happen and then open it again, the page becomes blank.
- Refresh the web page can solve this issue.

**Video** : https://streamable.com/u2rhc0

I can reproduce it using firefox dev 79.0b6, chrome 83.0.4103.116, and firefox android 68.10.1.

gagahpangeran:
fix #6308 

Should I change this

https://github.com/ppy/osu-web/blob/eeaecc44c06a4a657765a5eb4a0180a74a93e2ce/resources/assets/lib/quick-search/worker.ts#L25-L30

to this

```ts
interface SearchResult {
  beatmapset?: SearchResultBeatmapset;
  forum_post?: SearchResultSummary;
  user?: SearchResultUser;
  wiki_page?: SearchResultSummary;
}
```

to check another potential undefined?

### DIFF
--- a/resources/assets/lib/quick-search/main.tsx
+++ b/resources/assets/lib/quick-search/main.tsx
@@ -77,7 +77,7 @@ const otherModes: ResultMode[] = ['forum_post', 'wiki_page'];
   }
 
   private count(mode: ResultMode) {
-    if (this.props.worker.searchResult === null) {
+    if (this.props.worker.searchResult === null || this.props.worker.searchResult[mode] === undefined) {
       return 0;
     }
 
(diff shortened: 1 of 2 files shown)


gagahpangeran:
Ah I see..
I check in `beatmapset search filters` and there's this line

https://github.com/ppy/osu-web/blob/3613a3c9ea9129b211398d0e3694076e9371996a/resources/assets/lib/beatmapset-search-filters.ts#L55

Is it okay to sanitize the query like that?

nanaya:
Not much difference

PR #6322, opened 2020-07-16
Don't generate notifications for bots

### LINKED ISSUE #6320: Notifications shouldn't be sent to banchobot
BanchoBot (uid3) has 159,752 unread notifications. There should be a rule against sending notifications to bots in general.

nanaya:
Resolves #6320.


nanaya:
wait, that's not the correct latest commit :thinking: I updated that to return only bot and then diffing locally through php

PR #6324, opened 2020-07-17
Add support for play mode variant to tournament

nanaya:
Should be pretty straightforward: set `play_mode_variant` column (`4k` or `7k` for mania) as needed. Empty string isn't a valid variant. Null (value, not "null" string) is.


peppy:
Need this in ASAP for upcoming tournament use @notbakaneko

PR #6341, opened 2020-07-20
Add user cover to beatmaps score response

### LINKED ISSUE #6338: User cover missing in /beatmaps/{id}/scores response
See https://github.com/ppy/osu/issues/9618

gagahpangeran:
Fix #6338


app/Http/Controllers/BeatmapsController.php
```diff
@@ -76,7 +76,7 @@ public function scores($id)
                 }
 
                 $results = [
-                    'scores' => json_collection($query->visibleUsers()->forListing(), 'Score', ['beatmap', 'user', 'user.country']),
+                    'scores' => json_collection($query->visibleUsers()->forListing(), 'Score', ['beatmap', 'user', 'user.country', 'user.cover']),
```
nanaya, line 79: userProfileCustomization needs to be preloaded otherwise this will cause 49 additional queries.

nanaya:
Also userScore still need that additional transformer include otherwise user own score won't have cover metadata.

PR #6344, opened 2020-07-20
Add Client Credentials Grant to API documentation

### LINKED ISSUE #6313: API documentation doesn't mention new client grant support
Also the "public use" in this sentence should probably just be dropped:

image

notbakaneko:
also adds a note about resetting the client secret revoking issued tokens.

- [x] also going to require #6352 for documentation on callbacks to make sense

closes #6313


resources/views/vendor/apidoc/partials/info.blade.php
```diff
@@ -154,6 +158,60 @@
...
+});
+```
+
+`POST {!! route('oauth.passport.token') !!}`
+
+Parameters
+
+Name          | Type   | Description
+--------------|--------|-------------------------------
+client_id     | number |  The Client ID you received when you [registered]({{ route('account.edit').'#new-oauth-application' }})
```
nanaya, line 192: random extra space

resources/views/vendor/apidoc/partials/info.blade.php
```diff
@@ -194,7 +252,17 @@
 friends.read  | Allows reading of the authorizing user's friend list.
 public        | Allows reading of publicly available data on behalf of the authorizing user.
 
-`identify` is the default scope and always implicitly provided.
+`identify` is the default scope for the [Authorization Code Grant](#authorization-code-grant) and always implicitly provided. The [Client Credentials Grant](#client-credentials-grant) does not currently have any default scopes.
+
+
+## Managing an OAuth application
```
nanaya, line 258: "an" not needed? :thinking:

resources/views/vendor/apidoc/partials/info.blade.php
```diff
@@ -154,6 +158,60 @@
 refresh_token | string | The refresh token.
 
 
+## Client Credentials Grant
+
+The client credential flow provide a way for developers to get access tokens without user permissions; as such, these tokens are considered as guest users.
```
nanaya, line 163: "getting access tokens without user permissions" sounds a bit dangerous :upside_down_face:  "with no" instead of "without" perhaps? Or some other better wordings.

notbakaneko: ...those both mean the same thing 🤔

PR #6348, opened 2020-07-21
Include artist when changing display preferences for Beatmap

### LINKED ISSUE #3217: Add option to display beatmap metadata in original language

notbakaneko:
missing from #6271
closes #3217

Not sure if I should also rename the option 🤔

### DIFF
--- a/resources/assets/coffee/react/profile-page/beatmap-playcount.coffee
+++ b/resources/assets/coffee/react/profile-page/beatmap-playcount.coffee
@@ -4,6 +4,7 @@
 import * as React from 'react'
 import { a, div, h2, h3, img, p, small, span, strong } from 'react-dom-factories'
 import { StringWithComponent } from 'string-with-component'
+import { UserLink } from 'user-link'
 import { getArtist, getTitle } from 'utils/beatmap-helper'
 el = React.createElement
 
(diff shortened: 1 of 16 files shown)


gagahpangeran:
> Not sure if I should also rename the option thinking

Maybe rename it to 'show beatmap metadata in original language' so it's more clear.

resources/assets/coffee/react/profile-page/beatmap-playcount.coffee
```diff
@@ -31,26 +34,36 @@ export class BeatmapPlaycount extends React.PureComponent
...
-                    class: "#{bn}__mapper-link js-usercard"
-                    'data-user-id': beatmapset.user_id
+              el StringWithComponent,
+                pattern: osu.trans 'beatmapsets.show.details.mapped_by'
+                mappings:
+                  ':mapper':
+                    span
+                      className: "#{bn}__mapper-link"
+                      key: 'mapper'
+                      el UserLink,
```
nanaya, line 62: the color is wrong because direct styling (on `a`) has higher precedence than inherited (from `__mapper-link`)

PR #6350, opened 2020-07-21
Fix chat input and no conversations page width mobile

gagahpangeran:
**Before**

image

**After**

image


gagahpangeran:
In addition when I debug this, I also found this issue.

**Before**

image

**After**

image

PR #6351, opened 2020-07-22
Add read index to user notifications

nanaya:
Might be useful for marking chat pm notifications read.

- [x] add `2020_07_22_054643_add_user_read_index_to_user_notifications` to migrations in production table as the index is already there


peppy:
Good to go at your discretion.

PR #6352, opened 2020-07-22
Allow OAuth applications with empty callbacks

### LINKED ISSUE #6345: Allow empty Application Callback URL
The `client_credentials` flow doesn't require the use of `redirect_uri`, for these cases, having an empty Application Callback URL should be allowed. `authorization_code` flows should still require a `redirect_uri`.

notbakaneko:
OAuth applications with no application callback urls (`redirect_uri`) should be allowed since the `client_credentials` grant doesn't make use of them. `redirect_uri` is still required when making requests for `authorization_code` grants; previously, not providing one in the request would error because it wouldn't match the required `redirect_uri`.

closes #6345


app/Http/Controllers/Passport/ApproveAuthorizationController.php
```diff
@@ -0,0 +1,22 @@
...
+use Laravel\Passport\Http\Controllers\ApproveAuthorizationController as PassportApproveAuthorizationController;
+
+class ApproveAuthorizationController extends PassportApproveAuthorizationController
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function approve(Request $request)
+    {
+        abort_if(!present(trim($request['redirect_uri'])), 400, trans('model_validation.required', ['attribute' => 'redirect_uri']));
```
nanaya, line 18: the authorization form only has client id and state so this always fail

PR #6353, opened 2020-07-22
Remove messages of silenced users

### LINKED ISSUE #6133: Chat doesn't hide messages of silenced users
image

image

User has been muted, but his messages weren't removed from the chat

nanaya:
At few millions of account history entries, adding the index may take a while :thinking: 

This doesn't actually delete anything from database so the messages will return on reload. The process is supposedly handled differently on the silence action itself.

Resolves #6133.

- [x] add migration to production `2020_07_22_045416_add_timestamp_index_to_user_account_history`

```sql
CREATE INDEX osu_user_banhistory_timestamp_index ON osu_user_banhistory (timestamp);
```


app/Http/Controllers/Chat/ChatController.php
```diff
@@ -143,7 +144,24 @@ public function updates()
 
         $messages = $messages->get()->reverse();
 
-        if ($messages->isEmpty()) {
+        $silenceQuery = UserAccountHistory::bans()->limit(100);
+        $lastHistoryId = get_int($params['since_history'] ?? null);
+
+        if ($lastHistoryId === null) {
+            $previousMessage = Message::find($since);
```
notbakaneko, line 151: what's the reason for not using `$messages->last()` (first?)

app/Http/Controllers/Chat/ChatController.php
```diff
@@ -143,7 +144,24 @@ public function updates()
 
         $messages = $messages->get()->reverse();
 
-        if ($messages->isEmpty()) {
+        $silenceQuery = UserAccountHistory::bans()->limit(100);
+        $lastHistoryId = get_int($params['since_history'] ?? null);
```
notbakaneko, line 148: and this param should probably be `history_since` or `silences_since`

app/Http/Controllers/Chat/ChatController.php
nanaya: because there might be bans before the earliest current latest message?

PR #6374, opened 2020-07-28
Add notifications for replies to comments

### LINKED ISSUE #4355: Notification for comment replies
1.) replies to your own comments should send you notifications by default. (maybe be disable-able). helps to continue a conversation.

2.) option to watch the comment sections for specific beatmaps and news articles. maybe could be done using a bookmark feature like forum threads, or a watch feature like beatmap discussions. then any comments posted on the watched beatmap/news article will send you notifications.

reasoning for b: as a mapper i want to know when people comment on stuff without having to manually check every map i have on a regular basis. please make my dreams come true. let me watch the comments for beatmaps. thank you have a nice day

notbakaneko:
Adds the option to receive notifications when someone replies to your comment (defaults to false). This option does not depend on the comment watches, meaning if you're not watching the comments section and have this option enabled, you will still receive the notification. Delivery mode is handled by `comment_new`'s settings, since it uses the same notification with custom text client-side.

This doesn't use the originally planned addition of a separate delivery option for replies as implementing that resulted in some confusing non-sensical combinations.

closes #4355


app/Models/UserNotificationOption.php
```diff
@@ -27,6 +27,7 @@ class UserNotificationOption extends Model
     const BEATMAPSET_MODDING = 'beatmapset:modding'; // matches Follow notifiable_type:subtype
     const DELIVERY_MODES = ['mail', 'push'];
     const FORUM_TOPIC_REPLY = Notification::FORUM_TOPIC_REPLY;
+    const COMMENT_REPLY = Notification::COMMENT_REPLY; // FIXME: temporary until mail notification changes get merged in
```
nanaya, line 30: and then it's time to fix this

app/Jobs/Notifications/CommentNew.php
```diff
@@ -5,46 +5,21 @@
...
-
-        $this->comment = $comment;
+        $userIds = Follow::whereNotifiable($this->comment->commentable)
+            ->where(['subtype' => 'comment']);
 
-        if ($this->comment->commentable === null) {
-            throw new InvalidNotificationException("comment_new: comment #{$this->comment->getKey()} missing commentable");
+        if ($this->comment->parent_id !== null) {
+            $userIds->where('user_id', '<>', $this->comment->parent->user_id);
+            broadcast_notification(Notification::COMMENT_REPLY, $this->comment, $this->source);
```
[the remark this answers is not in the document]
notbakaneko: I actually skipped fixing this at the time, because, if you were watching the comment thread but chose to receive only mail notifications for replies, what should show up in your push notifications?

resources/views/accounts/_edit_notifications.blade.php
```diff
@@ -28,6 +28,33 @@
                     </div>
                 </label>
             </div>
+
+            <div
+                class="account-edit-entry account-edit-entry--no-label js-account-edit js-user-preferences-update"
```
nanaya, line 33: `js-user-preferences-update` causes error because the endpoint doesn't return anything.

nanaya (approved):

PR #6380, opened 2020-07-29
Check client version for multiplayer

nanaya:
Adds simple check.


app/Http/Controllers/Multiplayer/Rooms/Playlist/ScoresController.php
```diff
@@ -163,6 +164,12 @@ public function store($roomId, $playlistId)
     {
         $room = Room::findOrFail($roomId);
         $playlistItem = $room->playlist()->where('id', $playlistId)->firstOrFail();
+
+        $clientHash = presence(request('version_hash'));
+        abort_if($clientHash === null, 422, 'missing client version');
+        $client = Build::where('hash', hex2bin($clientHash))->firstOrFail();
+        abort_if(!$client->allow_ranking, 403, 'invalid client version');
```
notbakaneko, line 171: I would think non-existent and non-rankable clients should be returning the same error?

PR #6383, opened 2020-07-29
Fix wrong css url encoding

nanaya:
Plain `encodeURI` doesn't work because it may cause double encoding when passed percent-encoded url and it doesn't even escape single quotes.


notbakaneko:
Shouldn't the urls being used here be not pre-encoded in the first place?

nanaya:
Not directly related but generating non-percent-encoded url is actually impossible with `route` helpers :thinking: The same if you try copying from browsers link or url bar.

If anything, we should probably assume the input is percent-encoded and just wrap it with double quotes.

PR #6385, opened 2020-07-29
Remove global relative positioning

nanaya:
This seems to net some performance improvement in some places (large discussion page for example).

`u-relative` utility has been added for places where it's the only styling needed.

- [x] depends on #6471 as I'm not fixing styling for old buttons
- [x] depends on #6477 for similar reason as above


notbakaneko:
For reference, with this, I can have hundreds of chat channels open and the hover effect no longer lags behind the cursor.

PR #6387, opened 2020-07-30
Allow using separate redis for cache

nanaya:
Environment variable `CACHE_REDIS_HOST` (and `_PORT`). It uses `REDIS_HOST` if empty (follows previous behavior).


peppy:
✅  updated env

PR #6389, opened 2020-07-30
Change text in chatbox only persist per channel

### LINKED ISSUE #6372: Text in chatbox should not be shared across every channel
## Current behavior
Screen-Recording-2020-07-28-11-43-17

This happens to both channels and PMs.

## Expected behavior
Text in chatbox should be confined to a single chat that it was typed in. The reasoning is that, for instance, if you are chatting in a PM with someone and accidentally presses send while switching chat tabs, private information may inadvertently be made public.

gagahpangeran:
Closed #6372 

Only persist in chat page only using local state. The text will be cleared when refresh or change page.

If the behavior wants to be the same as the forum reply (using local storage), I'll change it.


gagahpangeran:
> The current input text for the channel can be stored in the `Channel` object instead?

Ah yes, that will do.

PR #6392, opened 2020-07-30
Persistent "Hide Deleted Posts" setting

### LINKED ISSUE #6218: Make "Hide Deleted Posts" persistent
Hide Deleted Posts definitely should be persistent across pages.
It is important for moderators that actually want to browse the forums while being e.g. in a screenshare with other non-moderators or perhaps even on a livestream. You can currently work around this by adding custom css, however that is not really a permanent solution.

notbakaneko:
`with_deleted` query string will still override the saved setting, but won't change it, unless the option is toggled.

I was looking at streamlining some of the `with_deleted` handling but it turned out to be quite messy keeping the url param, blade and js state in sync, so I ended up skipping it.

closes #6218

### DIFF
--- a/resources/assets/coffee/_classes/forum.coffee
+++ b/resources/assets/coffee/_classes/forum.coffee
@@ -202,7 +202,7 @@ class @Forum
           forum_posts_show_deleted: +!@showDeleted()
     .done (user) =>
       $.publish 'user:update', user
-      Turbolinks.visit osu.updateQueryString @postUrlN(@currentPostPosition)
+      Turbolinks.visit @postUrlN(@currentPostPosition)
 
 
   initialScrollTo: =>
--- a/resources/assets/coffee/_classes/forum.coffee
+++ b/resources/assets/coffee/_classes/forum.coffee
@@ -199,7 +199,7 @@ class @Forum
       method: 'PUT'
       data:
         user_profile_customization:
-          forum_posts_show_deleted: +!@showDeleted()
+          forum_posts_show_deleted: !@showDeleted()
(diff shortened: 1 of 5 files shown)


resources/assets/coffee/_classes/forum.coffee
```diff
@@ -193,8 +193,17 @@ class @Forum
...
+    return if !@showDeleted()? # you don't see this option unless you're a moderator, anyway
+
+    $.ajax laroute.route('account.options'),
+      method: 'PUT'
+      data:
+        user_profile_customization:
+          forum_posts_show_deleted: +!@showDeleted()
+    .done (user) =>
+      $.publish 'user:update', user
+      Turbolinks.visit osu.updateQueryString @postUrlN(@currentPostPosition)
```
nanaya, line 205: `updateQueryString` not doing anything.

resources/assets/coffee/_classes/forum.coffee
```diff
@@ -193,8 +193,17 @@ class @Forum
...
   toggleDeleted: =>
-    Turbolinks.visit osu.updateQueryString @postUrlN(@currentPostPosition),
-      with_deleted: +!@showDeleted()
+    return if !@showDeleted()? # you don't see this option unless you're a moderator, anyway
+
+    $.ajax laroute.route('account.options'),
+      method: 'PUT'
+      data:
+        user_profile_customization:
+          forum_posts_show_deleted: +!@showDeleted()
```
nanaya, line 202: cast to int not needed

PR #6398, opened 2020-07-31
Add play count for unranked beatmapset card

### LINKED ISSUE #6381: Add play count for unranked beatmapset card
I suggest to add the play count at top right beatmapset card like ranked beatmapset. Currently play count is only displayed in beatmapset page but not displayed in the card.

image

Also related with #6375 so user can see the play count when sorting.

gagahpangeran:
Closed #6381


gagahpangeran:
Swap the column maybe if two column?

**Desktop**

Screenshot from 2020-08-06 13-27-09

**Mobile**

Screenshot from 2020-08-06 13-27-54

nanaya:
That will be a bit confusing when there are both ranked and unranked maps on the list as the numbers will move around.

gagahpangeran:
Ah, I get it.

PR #6406, opened 2020-08-02
Fix tooltip not hiding on mobile

### LINKED ISSUE #5532: Some info popups not hiding on unhover
Example, showing three beatmap name popups at the same time:

To reproduce, just open the popups without doing anything else (they can only be opened on beatmap titles that don't fit the page width). They can be hidden by hovering over them again, but that's counter intuitive and annoying.

### LINKED ISSUE #5533: Hover-sensitive elements not automatically losing focus on mobile.
For example, if you press a mod selection button on beatmap scores page, it shows as hovered even when you release it. 
To make it realize it's released, you need to tap a different position on the screen. It does make a little sense with popups (Like the one that shows how many hours you played when you hover on playtime) because it's a bit harder to trigger them on mobile without releasing the screen (it tries to copy if you move your fingers a bit), however even in that case I think it's better to hide them because they can be unintuitive to hide.

gagahpangeran:
Fix #5532

~~Probably also fix #5533 (let me know if this is not the solution).~~


gagahpangeran:
Will investigate #5533 later

PR #6420, opened 2020-08-03
Change forum toolbar to scrollable in mobile

### LINKED ISSUE #4949: Forum footer bar overflow on mobile
image

Samsung Galaxy S5 as set device, but clips on most devices

### LINKED ISSUE #5373: Mobile forum moderation buttons push things off-screen
might be fixed by #4273 but not sure

would be nice if one button in the bottom-left corner could expand into a menu, like the mobile nav menu

Screenshot_20191204-192244

gagahpangeran:
Fix #4949 and fix #5373 as temporary solution until there's new design (?) in the future.

Because that issue also bother me in mobile.


resources/assets/less/bem/forum-topic-nav.less
```diff
@@ -25,6 +25,7 @@
     position: absolute;
     bottom: 0;
     left: 0;
+    overflow-x: auto;
```
nanaya, line 28: `forum-topic-nav` block can use `u-fancy-scrollbar` class so the scrollbar isn't oversized on desktop browsers.

gagahpangeran: Probably the scrollbar will not show in desktop because most likely it will never overflow, but I'll add it.

PR #6435, opened 2020-08-05
Fix song title overflow in beatmap page mobile

gagahpangeran:
**Before**

image

**After**

image


resources/assets/less/bem/beatmapset-header.less
```diff
@@ -85,6 +85,7 @@
       line-height: 1;
       margin-bottom: 5px;
       margin-top: 10px;
+      word-wrap: break-word;
```
nanaya, line 88: `word-break: break-word` is slightly more reliable (mainly it doesn't depend on the element to have explicit max width)

PR #6437, opened 2020-08-05
Add documentation for user API endpoint

nanaya:
No object structure documentation yet because it's very long ( ﾟ ヮﾟ)


app/Http/Controllers/UsersController.php
```diff
@@ -249,11 +358,67 @@ public function scores($_userId, $type)
...
+     * page                                 | |
+     * previous_usernames                   | |
+     * rankHistory                          | For specified mode.
+     * ranked_and_approved_beatmapset_count | |
+     * replays_watched_counts               | |
+     * scores_first_count                   | For specified mode.
+     * statistics                           | For specified mode. Inluces `rank` and `variants` attributes.
+     * support_level                        | |
+     * unranked_beatmapset_count            | |
+     * user_achievements                    | |
```
notbakaneko, line 415: I think we need a better way of documenting includes rather than just chucking them a ll in here, though

PR #6438, opened 2020-08-05
Fix add new embed show discussion deleted in beatmap discussion history page

### LINKED ISSUE #6140: Adding new embed post in discussion listing shows it as deleted
0. add new review (skip if there's one recently)
1. go to `/beatmapsets/beatmap-discussions`
2. edit the review, add new embed
3. save

Result is the new embed shown as "discussion deleted" instead of the actual content.

gagahpangeran:
fix #6140 

So this happened because the value of `DiscussionContext` is not updated after add new embed, so when `ReviewPostEmbed` consume the new value, the object of new embed is undefined.

\<rant\>
Somehow my solution for this bug works. Debugging in this component is very hard because almost all of the array and object operation are mutable and very unsafe. I think it should be refactored so the operation is immutable.
Also when I stop the elastic search service, I always got weird 422 error when add or delete the review. I didn't get 422 error when the elastic search service is up. But after few hours of debugging my machine started lagging because the RAM was full and even my swap was used.
Screenshot from 2020-08-05 16-38-46
\<\/rant\>


resources/assets/coffee/react/beatmap-discussions-history/main.coffee
```diff
@@ -42,6 +42,7 @@ export class Main extends React.PureComponent
 
     discussions = @state.discussions
     users = @state.users
+    relatedDiscussions = @state.relatedDiscussions
```
nanaya, line 45: you can reduce state modification by cloning them, I guess (with `[..]`).

gagahpangeran: Should I change for the the other state too?

nanaya: yes, might as well

PR #6417, opened 2020-08-03
Fix mail notification digest links

notbakaneko:
Each collection is supposed to be able to contain multiple links, e.g.

```
The discussion on "some beatmapset" has new updates:
/beatmapsets/{id}/discussion#/563
/beatmapsets/{id}/discussion#/5643
/beatmapsets/{id}/discussion#/123124
```

PR #6439, opened 2020-08-05
Omit some notifications from notification mail digest

notbakaneko:
Exclude the more generic `xyz has updates` type notifications from the mail digest if they've already been sent and the user hasn't performed any action that marks it as 'read'. This mostly applies to forum replies.

Notifications for items that are more specific or can be part of another item (like comments that can be anywhere) are still sent, but these generally have more specific links to the items themselves and shouldn't have been re-sent anyway.

includes #6417


app/Jobs/Notifications/BeatmapsetDiscussionPostNotification.php
```diff
@@ -16,6 +16,15 @@ abstract class BeatmapsetDiscussionPostNotification extends BroadcastNotificatio
 
     protected $beatmapsetDiscussionPost;
 
+    public static function getBaseKey(Notification $notification): string
+    {
+        $category = Notification::nameToCategory($notification->name);
+
+        return $category === 'beatmapset_discussion'
+            ? "{$notification->notifiable_type}.{$category}.beatmapset_discussion_post_new"
```
nanaya, line 24: it says "post_new" but it also includes discussion lock/unlock :thinking: 

just generic `_` maybe?

app/Jobs/Notifications/BeatmapsetDiscussionPostNotification.php
```diff
@@ -16,6 +16,15 @@ abstract class BeatmapsetDiscussionPostNotification extends BroadcastNotificatio
 
     protected $beatmapsetDiscussionPost;
 
+    public static function getBaseKey(Notification $notification): string
+    {
+        $category = Notification::nameToCategory($notification->name);
```
nanaya, line 21: maybe should add `category` function on notification.

nanaya (approved):

PR #6444, opened 2020-08-06
Fix review block counter in some pages

nanaya:
Context everywhere ( ﾟ ヮﾟ)


resources/assets/coffee/react/beatmap-discussions-history/main.coffee
```diff
@@ -90,34 +91,35 @@ export class Main extends React.PureComponent
...
-                el Discussion,
-                  discussion: discussion
-                  users: @users()
-                  currentUser: currentUser
-                  beatmapset: discussion.beatmapset
-                  isTimelineVisible: false
-                  visible: false
-                  showDeleted: true
-                  preview: true
+    el ReviewEditorConfigContext.Provider, value: @props.reviewsConfig,
```
nekodex, line 94: Should probably at least be consistent with the other contexts below (i.e. `value: ...` on the following line, or changing them to match)

resources/assets/coffee/react/modding-profile/main.coffee
```diff
@@ -148,65 +149,66 @@ export class Main extends React.PureComponent
...
-
-          div className: "osu-layout osu-layout--full#{if isBlocked && !@state.forceShow then ' osu-layout--masked' else ''}",
-            el Header,
-              user: @state.user
-              stats: @state.user.statistics
-              userAchievements: @props.userAchievements
-
-            div
-              className: 'hidden-xs page-extra-tabs page-extra-tabs--profile-page js-switchable-mode-page--scrollspy-offset'
+    el ReviewEditorConfigContext.Provider, value: @props.reviewsConfig,
```
nekodex, line 152: Should probably at least be consistent with the other contexts below (i.e. value: ... on the following line, or changing them to match)

PR #6456, opened 2020-08-08
Fix recent favourites users not updated after favouriting beatmapset

### LINKED ISSUE #6250: After favouriting beatmapset with 0 favourites, favourites starts showing nothing
a
Should not make it hover-able or update the favorites after favouriting.

gagahpangeran:
fix #6250


nanaya:
handle current user separately by creating an updated list of favourited users:
- if user has favourited, make sure they're in the list
- remove them from the list otherwise

make sure to show remaining number correctly as well; from what I gather `@favouritesToShow` is rather useless unless you want to trim out the list after making updated list (while keeping current user in it).

notbakaneko:
#6720

PR #6471, opened 2020-08-12
Remove old buttons and clean up fileupload

nanaya:
- remove `btn-osu` and `btn-osu-lite`
- simplify `fileupload` class and move stuff around


resources/assets/coffee/react/contest-entry/user-entry.coffee
```diff
@@ -23,7 +23,12 @@ export class UserEntry extends React.Component
   render: ->
     div className: 'contest-userentry contest-userentry--ok',
       if !@props.locked
-        a className: 'btn-osu btn-osu--textlike btn-osu--stick-right', href: '#', 'data-confirm': osu.trans('common.confirmation'), title: osu.trans('common.buttons.delete'), onClick: @delete,
+        button
+          className: 'contest-userentry__delete'
+          type: 'button'
+          'data-confirm': osu.trans('common.confirmation')
```
notbakaneko, line 29: I think this actually causes the handler for `button` to not trigger?

PR #6473, opened 2020-08-12
Make chat channel with moderator visible even if blocked

### LINKED ISSUE #6462: Channel not visible when receiving message from blocked moderator
This is caused by an oversight on my part in #6424. Moderators are able to send messages to users that are blocking them which causes a notification to appear, but the channel is still being hidden by a check in [UserChannel.php](https://github.com/ppy/osu-web/blob/1b5640d2ba3ac8330f2013d08d7b4f3358d4b2c2/app/Models/Chat/UserChannel.php#L160) that I missed.

I'd be happy to fix this, but I'm not sure which behavior is wanted. From what I can tell there are two options, either:
1. Don't allow moderators to message users blocking them at all (not super ideal for both GMT and NAT, but also likely a rare thing to encounter)
2. Even if the user is blocking the moderator, don't hide the channel from them if they receive a message

notbakaneko:
Makes chats from moderators visible even if user tries to block them.
Still can't reply unless you unblock them >_>.

This also involved way too many changes _just_ to preload the necessary user groups for the extra `isModerator()` check.
Also removed all the squeezing properties of from other tables onto different models.

closes #6462


app/Models/Chat/UserChannel.php
```diff
@@ -89,91 +84,105 @@ public static function presenceForUser(User $user)
...
-        );
+
+            return $presence;
+        });
 
         // strip out the empty [] elements (from restricted/blocked users)
         return array_values(array_filter($collection));
     }
+
+    private static function userChannelsForUser(User $user)
```
nanaya, line 184: `forUser` is enough, isn't it? UserChannel is already its class name after all. And maybe as scope?

app/Models/Chat/UserChannel.php
```diff
@@ -89,91 +84,105 @@ public static function presenceForUser(User $user)
...
 
-                    // override channel icon and display name in PMs to always show the other party
-                    $userActual = $targetUser->userScoped;
-                    $presence['icon'] = $userActual->user_avatar;
-                    $presence['name'] = $userActual->username;
-                    // ideally this should be ChatChannelSend but it involves too many queries
-                    $presence['moderated'] = $presence['moderated'] || !priv_check_user($user, 'ChatStart', $userActual)->can();
+                // hide if target is restricted or blocked unless blocked user is a moderator.
+                if (!$targetUser
+                    || $user->hasBlocked($targetUser) && !$targetUser->isModerator()) {
```
nanaya, line 166: Should this check for isAdmin as well? Currently admins can send message to everyone.

nanaya (changes requested):

nanaya (approved):

PR #6478, opened 2020-08-13
Fix forum post reply read notification not being sent properly

notbakaneko:
Payload was being sent in the no longer supported format and thus being ignored by the notification widget.


app/Jobs/MarkNotificationsRead.php
```diff
@@ -31,56 +28,44 @@ public function __construct($object, $user)
...
+            'category' => Notification::nameToCategory(Notification::FORUM_TOPIC_REPLY),
+            'object_id' => $notifiableId,
+            'object_type' => $notifiableType,
+        ];
+
+        if ($count > 0) {
+            event(new NotificationReadEvent($this->user->getKey(), [
+                'notifications' => [$notificationIdentity],
+                'read_count' => $count,
+                'timestamp' => now(),
```
nanaya, line 67: it looks like this will mark everything client-side as read even though it's only up to the post's `post_time` server-side? (which may be quite a bit before `now()` and the topic may have further unread posts)

app/Jobs/MarkNotificationsRead.php
```diff
@@ -31,56 +28,44 @@ public function __construct($object, $user)
...
-
-        $this->notificationTime = $this->object->post_time;
-    }
-
     public function handle()
     {
-        try {
-            if ($this->object instanceof Beatmapset) {
-                // do nothing
-            } elseif ($this->object instanceof ForumPost) {
```
nanaya, line 50: rename class to `MarkForumTopicReplyNotificationsRead`?

app/Models/Notification.php
```diff
@@ -93,6 +93,16 @@ public function source()
         return $this->belongsTo(User::class);
     }
 
+    public function toIdentityJson()
+    {
+        return [
+            'category' => static::nameToCategory($this->name), // TODO: change to category after https://github.com/ppy/osu-web/pull/6439
```
nanaya, line 99: and then it's time to fix this

nanaya (changes requested):

nanaya (approved):

PR #6479, opened 2020-08-13
Truncate forum post username

### LINKED ISSUE #6470: Truncate forum post username
e

notsoli:
Uses u-ellipsis-overflow to truncate forum post usernames if they overflow out of their container. In order to make u-ellipsis-overflow work, a width of 100% is set, similarly to some of the other elements in the post info (avatar, flag).

Before:
chrome_2020-08-13_15-18-02

After:
chrome_2020-08-13_15-18-55

Resolves #6470


resources/assets/less/bem/forum-post-info.less
```diff
@@ -75,6 +75,9 @@
     }
 
     &--username {
+      width: 100%;
+      display: block;
```
notbakaneko, line 79: this is not supposed to show up on mobile view

solitairecreations: oops, you're right
i think removing the display: block solves this issue
```suggestion
```

notbakaneko (approved):

PR #6481, opened 2020-08-14
Add warning indicator to review issues that are near existing posts

nekodex:
Screen Shot 2020-08-14 at 8 22 41 pm


resources/assets/lib/beatmap-discussions/editor-discussion-component.tsx
```diff
@@ -48,36 +64,190 @@ export default class EditorDiscussionComponent extends React.Component<Props> {
...
+      this.destroyTooltip();
     }
   }
 
+  componentWillUnmount() {
+    this.destroyTooltip();
+  }
+
+  createTooltip = (event: (React.MouseEvent | React.TouchEvent)) => {
+    const target = event.currentTarget as HTMLElement;
```
nanaya, line 88: The cast probably isn't needed if the signature has `<HTMLElement>`.

nanaya (changes requested):

resources/assets/lib/beatmap-discussions/editor-discussion-component.tsx
```diff
@@ -48,36 +64,190 @@ export default class EditorDiscussionComponent extends React.Component<Props> {
...
+        timestamp,
+      };
+    }
+
+    return this.cache.nearbyDiscussions?.discussions;
+  }
+
+  nearbyDraftEmbeds = (drafts: SlateNode[]) => {
+    const timestamp = this.timestamp();
+    if (timestamp == null || !drafts || drafts.length === 0) {
```
nanaya, line 167: just noticed `!drafts` (should) always return false here.

resources/assets/lib/beatmap-discussions/editor-discussion-component.tsx
```diff
@@ -107,36 +277,42 @@ export default class EditorDiscussionComponent extends React.Component<Props> {
...
-                className={`${bn}__timestamp`}
+                className={`${this.bn}__timestamp`}
                 contentEditable={false} // workaround for slatejs 'Cannot resolve a Slate point from DOM point' nonsense
               >
                 <span title={canEdit ? timestampTooltip : ''}>
                   {this.props.element.timestamp || osu.trans('beatmap_discussions.timestamp_display.general')}
                 </span>
               </div>
               {unsavedIndicator}
+              <DraftsContext.Consumer>{this.nearbyIndicator}</DraftsContext.Consumer>
```
nanaya, line 297: I've mentioned this somewhere else but it looks like storing this to a variable and using it here and in the other place just like `unsavedIndicator` may work and save rendering same thing twice.

nanaya (changes requested):

nanaya (changes requested):
Clickable timestamp is probably not the best idea when editing existing post. It's fine(ish) when creating a new one as it's saved to local storage but not quite when editing existing post.

nekodex:
Hrm, I suppose those timestamps should really be a (nested?) popup/preview or something to save having to navigate there and back... guess I'll just make them not-links in edit mode for now and figure out a better way later.

nanaya (approved):

PR #6482, opened 2020-08-14
Add missing preload for forum topic page

nanaya:
Found when checking other stuff.

There's still at least one more but it's rare case (one extra query for displaying polls) and I'm not yet sure how to properly cache it.


nanaya:
probably just need to cast it to null if zero :thinking:

PR #6514, opened 2020-08-19
Add recaptcha

nekodex:
An all-or-nothing implementation to begin with.

Toggle via `.env` - if these two variables are set then it's turned on, otherwise it's turned off:
```
RECAPTCHA_SECRET=
RECAPTCHA_SITEKEY=
```
[get keys here](http://www.google.com/recaptcha/admin) (reCAPTCHA v2, "I'm not a robot" Checkbox)


resources/views/layout/metadata.blade.php
```diff
@@ -107,6 +107,18 @@
     <script src="/vendor/js/moment-locales/{{ $momentLocale }}.js" data-turbolinks-track="reload"></script>
 @endif
 
+@if (config('captcha.sitekey') !== '' && config('captcha.secret') !== '')
+    <script src="https://www.recaptcha.net/recaptcha/api.js?render=explicit&onload=renderCaptcha" async defer></script>
```
nanaya, line 111: should set the locale as well. And maybe use NoCaptcha's `renderJs`? Also where did you get the url from? The documentation says [`google.com`](https://developers.google.com/recaptcha/docs/display)

resources/lang/en/users.php
```diff
@@ -84,6 +84,7 @@
         'failed' => 'Incorrect sign in',
         'forgot' => 'Forgotten your password?',
         'info' => 'Please sign in to continue',
+        'invalid-captcha' => 'Captcha invalid, refresh the page and try again.',
```
nanaya, line 87: translation keys are `snake_case`d

resources/assets/less/bem/login-box.less
```diff
@@ -47,6 +47,12 @@
       width: 300px;
       margin-top: 5px; // sync with Nav2#centerPopup
     }
+
+    &--captcha {
+      @media @desktop {
+        width: 340px;
```
nanaya, line 53: this is 4 pixels off (the width is 304 but 340px only provides 300px wide content)

app/Http/Controllers/SessionsController.php
```diff
@@ -49,6 +50,21 @@ public function store()
...
+        if (config('captcha.sitekey') !== '' && config('captcha.secret') !== '') {
+            $token = presence($params['g-recaptcha-response'] ?? null);
+            $validCaptcha = false;
+
+            if ($token !== null) {
+                $validCaptcha = NoCaptcha::verifyResponse($token);
+            }
+
+            if (!$validCaptcha) {
+                DatadogLoginAttempt::log('invalid_captcha');
```
nanaya, line 62: should it log missing captcha separately from invalid ones?

nanaya (changes requested):

resources/views/layout/metadata.blade.php
nekodex: This issue suggests that google.com is blocked in China: https://github.com/anhskohbo/no-captcha/pull/144

Also, https://developers.google.com/recaptcha/docs/faq
```
Can I use reCAPTCHA globally?

Yes, please use "www.recaptcha.net" in your code in circumstances when "www.google.com" is not accessible.

First, replace <script src="https://www.google.com/recaptcha/api.js" async defer></script> with <script src="https://www.recaptcha.net/recaptcha/api.js" async defer></script>
After that, apply the same to everywhere else that uses "www.google.com/recaptcha/" on your site.
```

I explicitly didn't use renderJs, because it uses google.com.

nanaya: maybe document that somewhere

resources/assets/less/bem/login-box.less
nekodex: Looks like 304px is to account for the drop shadow on their light-mode theme, setting the box to 344px makes the input boxes above the recaptcha box too wide... so I guess 342px to match the visual widths?

nanaya (approved):

PR #6519, opened 2020-08-19
Wrap long username instead of truncating

### LINKED ISSUE #6516: User card doesn't appear after username is hovered once in forum
I was browsing in forum and found this issue.

- Open a thread in forum.
- Hover username, then user card will appear.
- Hover again the same username, user card doesn't appear anymore.

[video](https://streamable.com/zrqety)

notbakaneko:
Having multiple qtips on the same element where one of them involves disabling all the tooltip interaction on the element turns out to be not a very good idea. 🤔 

Also fixes the usercard position since 100% width ends up always pushing it too far to the side.

`word-break: break-word` actually overrides `overflow-wrap`, but Safari
doesn't support `overflow-wrap: anywhere` yet.

closes #6516

### DIFF
--- a/resources/assets/less/bem/forum-post-info.less
+++ b/resources/assets/less/bem/forum-post-info.less
@@ -75,7 +75,6 @@
     }
 
     &--username {
-      overflow-wrap: anywhere;
       word-break: break-word;
 
       font-size: @font-size--title-small-3;
--- a/resources/assets/less/bem/forum-post-info.less
+++ b/resources/assets/less/bem/forum-post-info.less
@@ -75,8 +75,9 @@
     }
 
     &--username {
-      width: 100%;
-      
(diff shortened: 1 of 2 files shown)


resources/assets/less/bem/forum-post-info.less
```diff
@@ -75,8 +75,9 @@
     }
 
     &--username {
-      width: 100%;
-      
+      overflow-wrap: anywhere;
```
nanaya, line 78: I think the plan is to only use `word-break: break-word` for now and then mass replace with this once Safari catches up. It's currently redundant as the `word-break` below does the same thing (and being web, it will probably stay for like another 100 years).

nanaya (changes requested):

nanaya (approved):

PR #6537, opened 2020-08-24
Always show beatmap play success chart

### LINKED ISSUE #6432: Display pass/fail chart for pending+graveyarded beatmaps
Currently on pending and graveyarded beatmaps, the section will just say "unranked beatmap"
image

This information should be displayed now as plays are tracked for pending and unranked maps. 

It's also important to make this section functional, as presently per-difficulty playcount is not accessible on the webpage without this panel.
image

Having per-difficulty playcount visible would be beneficial for programs like loved to see which difficulties in a set are actually being played.

nanaya:
Resolves #6432.


notbakaneko (approved):

notbakaneko (changes requested):
suddenly conflicts 🤔

notbakaneko (approved):

PR #5322, opened 2019-11-22
React-ify beatmap discussion history page

nekodex:
Uses similar logic to the modding profile page to structure discussion data in such a way that they can be rendered by the beatmap-discussion react components - this way we will no longer have two distinct code paths to maintain for rendering beatmap discussions (in javascript and php...).

This also combines the users-specific discussion history page (previously at `/users/<id>/modding/discussions`) with the main index because they were functionally identical anyway.

Not sure if it's worth adding a redirect for it?

- [x] depends on #5307

PR #6538, opened 2020-08-24
Fix user parameter not set properly on beatmap events listing page pagination links

### LINKED ISSUE #6409: alternating page results in beatmap events removes user filter when it's used
as title says
https://osu.ppy.sh/beatmapsets/events?user=pimp&min_date=&max_date=

if i try to click on page 2 it will redirect me to this link https://osu.ppy.sh/beatmapsets/events?limit=20&page=2#events which is basically page 2 without user filter. it was supposed to show me this result https://osu.ppy.sh/beatmapsets/events?user=pimp&page=2&min_date=&max_date=

nanaya:
Resolves #6409.

Although I have a feeling we should just remove the page (and some other user-specific listing pages) entirely and direct the link from modding profile to normal events listing page instead. Not sure why it's not done yet. Should I? :thinking: 

Now I think about it there's blocking problem for user forum post listing which requires keyword for normal search but it's not the case for beatmap modding related stuff.

- [x] decide this or the other one

### DIFF
--- a/app/Http/Controllers/Users/ModdingHistoryController.php
+++ b/app/Http/Controllers/Users/ModdingHistoryController.php
@@ -70,6 +70,7 @@ class ModdingHistoryController extends Controller
         $user = $this->user;
 
         $bundle = ModdingHistoryEventsBundle::forListing($user, $this->searchParams);
+        $bundle->paginationIncludeUserParam = false;
         $jsonChunks = $bundle->toArray();
         $paginator = $bundle->getPaginator();
         $params = $bundle->getParams();
--- a/app/Libraries/ModdingHistoryEventsBundle.php
+++ b/app/Libraries/ModdingHistoryEventsBundle.php
@@ -18,6 +18,8 @@ class ModdingHistoryEventsBundle
 {
     const KUDOSU_PER_PAGE = 5;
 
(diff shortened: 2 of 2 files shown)


nekodex:
Yeah, probably should remove the remaining user-specific listing pages... only reason they're not gone yet is because we haven't got around to it yet.

Might be worth just merging this in the meantime unless you think removing those pages won't take too long?

nanaya:
Shouldn't be too difficult... I think

notbakaneko:
I think we also wanted the user's modding profile stuff to always be accessible via `users/{user_id}/modding/thing` and not `thing?user={user_id}`

nanaya:
So it turned out the posts listing page doesn't have search form yet (which is probably why it's split in the first place).

As for whether or not it should be separate, I don't think there's specific discussion for it apart of being mentioned a bit in #5322 which also updates the link for one specific part to use general listing.

I don't think there's point having them separate though. It just complicates the already almost nonexistent design.

nekodex:
Not sure if it really matters whether it's mounted at the user-specific path or not? I guess we can add a redirect from there if required?

notbakaneko:
closed in favour of #6558

PR #6542, opened 2020-08-24
Remove manual event throttling on scroll and resize

nanaya:
Check on Firefox and Chrome indicated those aren't needed anymore.
One fewer thing to worry about.

Also fixes PopupMenu having wrong event name.

Tested by replacing `return if running` with

```
if running
  console.log 'skipping run', eventName
  return
```


notbakaneko (approved):
Seems like they all fire at the same rate as part of the spec (and has been for years) 🤔 
https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-processing-model

PR #6544, opened 2020-08-25
Event cleanup in Popup Menu component

nanaya:
- wrong namespace because single quote
- nothing removes the listener on `$(window)`


notbakaneko (approved):

notbakaneko (changes requested):
now this one has conflicts, what 🤔

PR #6549, opened 2020-08-25
Add troubleshooting help on download page

nanaya:
Not sure if should link to forum or some wiki page :thinking: 

Also adds enchant button.

Reworked the page a bit as well while at it.


notbakaneko (commented):
but the forum link doesn't show either if enchant id isn't set 🤔

nanaya:
I'm not conditionaling and adding more translation just for that :neutral_face:

notbakaneko (approved):

notbakaneko (approved):

PR #6553, opened 2020-08-26
Group beatmap discussion moderation buttons separately and add colour

### LINKED ISSUE #6550: Delete and Nominate Buttons shouldn't be right next to one another
If you have the BN and GMT/NAT Usergroup, nominating a map will look like this:
example

both the nominate button and delete button will ask the user whether or not they're sure of what they're doing. This is all fine, but the messages and buttons are so similar and close together that it's basically an accident just waiting to happen (wanted to nominate a map? Whoops, it's gone)

Can you swap this button's position with the "Lock discussion" button which is far less destructive if misfired?

nekodex:
Screen Shot 2020-08-26 at 1 34 47 pm

Hopefully this should make the buttons easier to identify and prevent people from accidentally clicking the wrong ones...

fixes #6550

### DIFF
--- a/resources/assets/coffee/react/beatmap-discussions/nominations.coffee
+++ b/resources/assets/coffee/react/beatmap-discussions/nominations.coffee
@@ -393,7 +393,7 @@ export class Nominations extends React.PureComponent
     el BigButton,
       text: osu.trans 'beatmaps.nominations.disqualify'
       icon: 'fas fa-thumbs-down'
-      modifiers: ['warn']
+      modifiers: ['warning']
       props:
         onClick: @focusNewDiscussionWithModeSwitch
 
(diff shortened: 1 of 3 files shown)


resources/assets/coffee/react/beatmap-discussions/nominations.coffee
```diff
@@ -40,13 +40,15 @@ export class Nominations extends React.PureComponent
...
         div className: "#{bn}__item #{bn}__item--nominators", @nominatorsList()
       div className: "#{bn}__items #{bn}__items--buttons",
-        div className: "#{bn}__item", @feedbackButton()
-        div className: "#{bn}__item", @hypeButton()
-        div className: "#{bn}__item", @disqualifyButton()
-        div className: "#{bn}__item", @nominationButton()
-        div className: "#{bn}__item", @loveButton()
-        div className: "#{bn}__item", @deleteButton()
-        div className: "#{bn}__item", @discussionLockButton()
+        div className: "#{bn}__items--grouping",
```
nanaya, line 43: what is this modifier without element

resources/assets/less/bem/beatmap-discussion-nomination.less
```diff
@@ -45,6 +45,11 @@
       --item-margin: 2px;
       flex-wrap: wrap;
```
nanaya, line 46: may want to move this to the grouping <del>modifier</del> element (for mobile)

nanaya (changes requested):

nanaya (approved):

PR #6554, opened 2020-08-26
Locally cache groups data

nanaya:
File cache because apc is a pecl extension. Usual redis for versioning mechanism.

Initial cache may generate a few duplicates depending on the timing but hopefully it doesn't generate too many cache files. Especially tests (hence array driver).


notbakaneko (changes requested):
- How does the cache get invalidated?
- How do previously cached entries expire/get removed?
- Should have note on why contents are stored locally since it's still hitting redis every time to get the version.

peppy:
Not sure why it needs a version at redis in the first place? The only time this needs to be invalidated is when a new group is added right? I can just run artisan `cache:clear` or something (or just have it update on redeploy only).

notbakaneko:
`cache:clear` deletes everything in the default cache and only that one; if it's redis, it flushes everything in the selected db.

peppy:
yeah, i know i shouldn't run that command, but potentially another one that doesn't.

nanaya:
It'll need to be run on all servers. And that may get more annoying when using docker or just when having more servers.

This way resetting the cache can be done from any one instance and can even be added to web/admin interface.

peppy:
I guess, but also adding one RTT that we're trying to get rid of.

nanaya:
might as well move the whole table to env file?

peppy:
i think we should just go with this for now.

notbakaneko (approved):

PR #6555, opened 2020-08-26
Replace paragraph about no items with counter

### LINKED ISSUE #2681: Hide map sections if a user has 0
On the old website, the drop down catergories for favorite/loved/ranked/pending/graveyarded beatmaps would not display if a user had 0 maps in that catergory.
image
but on the new website it does.. this
image
Which is wasting space for telling you there's literally no information here.
If a user has 0 of every type like the example above, the beatmap section should be hidden in general, just like how the "Account Standing" section is not displayed if a user has no recent infringements.

nanaya:
Resolves #2681.

Not sure if too many because I also apply it for ranks, historical, and modding profile's votes ( ﾟ ヮﾟ)


nekodex:
What about these sections? (modding profile)

Screen Shot 2020-08-26 at 7 03 33 pm

Is the collapsing of stuff only intended for subsections?

nanaya:
Yeah I didn't change that because it's not in subsection. And it's a bit beyond what the original issue is asking for.

nekodex:
Well, I'm clarifying because the linked issue says:
`If a user has 0 of every type like the example above, the beatmap section should be hidden in general, just like how the "Account Standing" section is not displayed if a user has no recent infringements.`

...but @arflyte's mockup didn't seem to take that into account? only the case where a subsection is empty?

arflyte:
As for sections, if it's empty, then maybe it's best just completely hide it (especially discussion and post). We can just leave the ones that we think people would want to see even if it's empty. I'd personally think it's best to hide all section with no items.

nanaya:
So the modding profile looks like this when everything is empty? :thinking: Should I hide the beatmaps count on the right as well?

(unrelated but what's ranking stats doing there)

 @arflyte

arflyte:
I haven't really looked into the modder area yet. Keep it hidden, for now, I guess. I'll look into the modder section soon.

nekodex:
Uhhh, that looks horrible and seems like a bad idea from a usability point of view? If I didn't know any better, I'd have assumed that the page had broke and failed to render...

I feel like we should collapse down to the header at most (possibly with a zero count like the subsections?), but not have them completely go away?

nanaya:
I'd say it's better showing them as is for now?

nekodex (approved):

PR #6558, opened 2020-08-27
Link user modding events to general listing page and fix pagination links

### LINKED ISSUE #6409: alternating page results in beatmap events removes user filter when it's used

nanaya:
Resolves #6409 by removing the user-specific page entirely. Alternative to #6538.

The main problem I can see is the form says "username" but it's using user id as it's more reliable (number-only username shouldn't have been allowed in the first place).

- [x] decide this or the other one


notbakaneko (approved):
Let's go with this one since it involves making less special cases and exceptions 🤔

PR #6561, opened 2020-08-27
More thorough user best score check and cleanup

notbakaneko:
Checks all the user's scores now instead of just the displayed aggregates. Should be computationally cheaper on the elasticsearch side since performing the aggregation requires loading the segment with the scores anyway.

Also changed `Search::searchAfter` to clear the response of the current search so that the existing search object can be reused at it was intended for cursoring instead of creating a new one.


notbakaneko:
There's another option where we can just queue all the score ids into the score indexer's queue and it'll do the right thing.

app/Libraries/UserBestScoresCheck.php
```diff
@@ -38,12 +49,32 @@ public function __construct(User $user)
...
     public function check(string $mode)
     {
+        $this->dbIdsFound = 0;
+        $this->esIdsFound = 0;
+
         $clazz = Best\Model::getClassByString($mode);
 
-        $esIds = $this->user->beatmapBestScoreIds($mode, 2000);
-        $dbIds = $clazz::default()->whereIn('score_id', $esIds)->pluck('score_id')->all();
+        $search = $this->newSearch('osu');
```
nanaya, line 57: nice mode

nanaya (changes requested):

nanaya (approved):

PR #6563, opened 2020-08-27
Allow restricted users to update some preferences

### LINKED ISSUE #6562: Adjusting/Muting the Volume and play next track automatically gives access denied when restricted
Trying to change the volume and clicking the "play next track automatically" button on song previews gives you the "access denied" warning. pretty sure it shouldnt? https://streamable.com/o6krvy

notbakaneko:
Allows personal preferences like sort orders, notification options, audio player settings, etc to be updated.

closes #6562


nekodex (changes requested):
There's still some options that 403 that probably shouldn't:
- Notifications -> `automatically enable notifications on new forum topics that you create`
- Stuff under Privacy and Playstyles (they also 403, but don't show an error popup)

nekodex (approved):

PR #6453, opened 2020-08-07
Fix beatmap discussion height

gagahpangeran:
Fix #6447

PR #6565, opened 2020-08-28
Cap height of message-area to prevent textarea overflow

### LINKED ISSUE #6447: Writing Box in Beatmap Discussion bugs on larger posts.
I was writing some mod post just now which got a bit longer, so i used "General (This difficulty)" which has no character limit. But at some point i noticed that the box does not get larger anymore and the writing goes "beyond" it. 

To see here: 
https://i.imgur.com/pyqv8nx.png

I closed my browser and tried if this happens again and it does.
So i tried to reproduce it on another beatmap, this time one of mine. And it happened again:
https://i.imgur.com/aSTADHd.png

I am using the latest version of Chrome, if this information helps and only have one extension on it. 

Furthermore i tried to reproduce it with another browser (Used Microsoft Edge because i have no other oops)
And the problem happened again.

nekodex:
This is a temporary/band-aid fix to stop the textarea overflowing (on desktop at least)... the mobile layout of the page, though, is hella broken and needs reworking.

supersedes #6453 and fixes #6447


resources/assets/less/bem/beatmap-discussion-new.less
```diff
@@ -72,7 +72,7 @@
     .reset-input();
     resize: vertical;
     width: 100%;
-    max-height: 70vh;
+    max-height: calc(60 * var(--vh, 1vh));
```
nanaya, line 75: This is still not quite correct when unpinned:



bonus (on discussion review composer, unpinned):



It seems like the actual problem is there are multiple `max-height` defined which don't quite add up properly. The one in `beatmap-discussion-new` should be removed?

nanaya (changes requested):

nanaya (approved):

PR #6566, opened 2020-08-28
Fix verification dialog error message position

notbakaneko:
Should display under the input box instead of all the way to the side.

Although, it seems weird that this text has `position: absolute ` at all 🤔 
It'd only benefit multi-line text not changing the layout, but then, if there's too many lines, it'd just start overlapping the other text? `min-height` could be used to maintain a minimum space for 2 lines but still allow more lines to expand the space 🤷

### DIFF
--- a/resources/assets/less/bem/user-verification.less
+++ b/resources/assets/less/bem/user-verification.less
@@ -56,6 +56,7 @@
     margin: 10px 0;
 
     &--key {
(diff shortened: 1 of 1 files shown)


peppy:
Merging this as a high priority fix. Please apply any review as a follow-up PR fix.

PR #6569, opened 2020-08-28
Fix position of close button in modding reply box

### LINKED ISSUE #6568: Close button in modding reply goes to post header instead of reply
title, this also overlaps with the collapse button.

notbakaneko:
closes #6568


peppy:
Merging this as a high priority fix. Please apply any review as a follow-up PR fix.

PR #6575, opened 2020-08-30
Add marker and warning for beatmaps with NSFW content

### LINKED ISSUE #5902: Label 18+ content with a “explicit” mark
Add the ability to identify "18+" content (more likely "13+" since we don't allow 18+ on osu! in the first place)

cl8n:
closes #5902

adds:

- ability to mark maps as NSFW with metadata editor
- a warning screen before viewing a NSFW map
- exclusion of NSFW maps from search
- an account setting to turn off the warning/search stuff

design and language are just my guesses what I thought looked okay, the warning is very similar styling to the report form

- [x] deployed #7014
- [x] run `php artisan es:index-documents --types=beatmapsets --yes`


peppy:
Thanks for the contribution!

One think I'll mention initially is that we should probably find a better term than NSFW, unless that's commonly used on other sites aimed at a younger audience?

"Potentially offensive content" or something?

peppy:
I can't think of a better acronym/term to use internally though, so I guess nsfw works fine from that perspective. Open to suggestion, though.

ribbanya:
"[Explicit (content)](https://en.wikipedia.org/wiki/Parental_Advisory)" is used on songs and other media when concerning a younger demographic, at least in the US.

nanaya:
is it only for older stuff or we're still allowing new nsfw stuff?

peppy:
There are still new edge cases which this would apply to.

cl8n:
should only mods be allowed to change the map nsfw setting? right now it's the same as genre/language permission, so the mapper and BNs can change it too before it's ranked

Ephemeralis:
NSFW is the 2020 gold standard given its prevalence on platforms like Reddit. Probably no need to change it.

Other alternatives could be 18+, Mature (M), Explicit. Localizing this could be a good way to target specific regulation terms that are prominent in a given language (or culture associated with that language) to allow us to make this as flexible as it needs to be.

nanaya (dismissed):
- audio preview should be disabled for nsfw maps' card
- add explicit badge to the beatmap card. Color `orange-2`, font size 10px, padding `0.2em 1em`, background `b5`, border radius 10k px, margin left 10px, align center with title with +2px from top adjustment. Title itself should have ellipsis on overflow but the badge should always be shown
  
- also for beatmap page itself

peppy:
@cl8n i think this is pending on applying a review from your end!

Ephemeralis:
@cl8n @peppy @nanaya @notbakaneko can one or more of you take this over to get something out ASAP in order to stop things from boiling over?

notbakaneko (changes requested):
Wonder if the index and schema update should be split out so they can be run first? Otherwise, users will get a page of nothing on the listing until the index is updated.

notbakaneko (commented):
I'm not sure how useful hiding the result from the beatmap listing is when there's no option to change the setting there (and the cover is masked/blocked already).

There's also the issue of notifications since those just use the url embedded into the notification itself.

nanaya:
The notification part is probably a lost cause ( ﾟ ヮﾟ)

The closest thing I can think of is removing the cover altogether based on its nsfw state at the time of notification regardless of the user's nsfw settings.

peppy:
Let's just ignore notifications until someone complains about it.

nanaya:
Added option for the search filter. Currently boolean (`false` to exclude, `true` to include, `null` to follow user settings). Pondered about adding it as part of Extra or General section but they don't support a third state, only `true` and `default`.

notbakaneko (approved):

peppy:
Indexing has been run now, merge at your discretion.

PR #6579, opened 2020-08-31
Fix error reporting failure when connection to Paypal fails

notbakaneko:
~~Apparently `getData()` will return null in many cases which the Sentry logger does not like.~~
~~`getMessage()` should at least be an empty string in the worst case.~~

Looks like updated `sentry-laravel` package uses `?string` instead of `string` 🤔

### DIFF
--- a/app/Libraries/Payments/PaypalExecutePayment.php
+++ b/app/Libraries/Payments/PaypalExecutePayment.php
@@ -63,7 +63,7 @@ class PaypalExecutePayment
 
                 return $payment->execute($this->execution, $context);
             } catch (PayPalConnectionException $e) {
-                \Log::error($e->getData());
+                \Log::error($e->getMessage());
                 // TODO: get more context data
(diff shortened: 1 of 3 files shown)


app/Libraries/Payments/PaypalExecutePayment.php
```diff
@@ -63,7 +63,7 @@ public function run()
 
                 return $payment->execute($this->execution, $context);
             } catch (PayPalConnectionException $e) {
-                \Log::error($e->getData());
+                \Log::error($e->getMessage());
```
nanaya, line 66: I don't think `Log::error` sends sentry log...? :thinking:

notbakaneko: sentry-laravel add handlers to logging events and adds breadcrumbs when a message is logged

nanaya (approved):

PR #6580, opened 2020-08-31
Add correctly-named rank history data in user json

nanaya:
I'm not going to document `rankHistory` key =)

Duplicated until other api users update the thing.


notbakaneko (commented):
are you going to update the apidoc block 🤔

notbakaneko (approved):

PR #6585, opened 2020-09-01
Split off object structure documentation files

nanaya:
Instead of one big file. List of files to be included thankfully can be autogenerated.


peppy (approved):
looks logical

PR #6594, opened 2020-09-03
Simplify dealing with twitch stream user data

nanaya:
`user_name` from streams listing is usable for both display name and generating url.


peppy (approved):
looks sane

notbakaneko (approved):

PR #6595, opened 2020-09-03
Fix missing beatmap fail rate charts when switching difficulties

### LINKED ISSUE #6570: Beatmap page data display issues (for lack of a better title)
GIF that shows bug in action:
YHlqgGOK6P

I used [my own beatmap](https://osu.ppy.sh/beatmapsets/1221830#osu/2541504) but this should be reproducible on other sets as well.

Steps to reproduce:

1. Click on another difficulty in the set that has no data available (as of this issue's filing it would be the Normal difficulty on my set)
2. Go to a difficulty that *does* have data available (Extra diff on my set has some)
3. Notice failure points don't show at all (if you refresh the page on the Extra diff, it should show again)

I'm on Microsoft Edge, version 84.0.522.63.

Please let me know if more information is needed, thanks!

nanaya:
- ~~fix broken chart when the data is all zero~~
- fix missing chart when toggling between beatmap with and without plays (resolves #6570)
- ~~hide data for converted maps (I think those pass/play counts in osu_beatmaps don't include converts? @peppy )~~

Split off the fixes.

### DIFF
--- a/resources/assets/coffee/_classes/stacked-bar-chart.coffee
+++ b/resources/assets/coffee/_classes/stacked-bar-chart.coffee
@@ -2,7 +2,7 @@
 # See the LICENCE file in the repository root for full licence text.
 
 class @StackedBarChart
-  constructor: (area, @options = {}) ->
+  constructor: (@area, @options = {}) ->
     @margins =
       top: 0
       right: 0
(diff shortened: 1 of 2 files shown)


peppy:
The counts include all plays at the moment, regardless of the target game mode (so they do include converts).

notbakaneko (approved):

PR #6601, opened 2020-09-04
Remove redundant link styling

nanaya:
Also removes light pagination in the process.

That was a lot of styles. My general guide was if it's not accompanied by `.reset-input` then it probably can be removed. Hopefully I didn't miss any weird cases ( ﾟ ヮﾟ)


notbakaneko (approved):
`link-default` is being applied globally on `a` anyway, I think there'd be other problems if applying `reset-input` and `link-default` again 🤔

PR #6607, opened 2020-09-07
Fix failtime chart when the data is all zero

nanaya:
Split from #6595.


notbakaneko (changes requested):
not that it makes a difference as long as all the values are integers >= 0, but wouldn't it make more sense to cap the maximum of the graph to `max(1, @max)`?

notbakaneko (approved):

PR #6609, opened 2020-09-07
Clean up getting params from request

notbakaneko:
every call to `offsetGet` on `Request` ends up calling `all()`, repeatedly creating new value arrays.


nanaya (approved):
There seems to be a bit more places can be fixed but this can go in first.

PR #6611, opened 2020-09-08
Move some manual file copy to webpack

nanaya:
File loader is a thing nowadays and work with plain css etc. And plain css loading also works without having to specify inline.


resources/assets/less/app.less
```diff
@@ -39,7 +40,6 @@
 @import "colors";
 @import "variables";
 
-@fa-font-path: "/vendor/fonts/font-awesome";
-@extra-font-path: "/fonts/extra";
-@torus-font-path: "/fonts/torus";
-@venera-font-path: "/fonts/venera";
+@extra-font-path: "../fonts/extra";
+@torus-font-path: "../fonts/torus";
+@venera-font-path: "../fonts/venera";
```
notbakaneko, line 45: should declare have some `@font` alias for the path so we don't have to do this `../` thing 🤔

notbakaneko (changes requested):

resources/assets/less/app.less
nanaya: Eh it's only three for now :shrug:

notbakaneko: I mean you can add `alias: { @font: .... }` to webpack and it will resolve `url(~@font/torus/etc)`

notbakaneko (approved):

PR #6619, opened 2020-09-09
Beatmap discussions mobile overflow/usability fixes

nekodex:
Fixes a bunch of overflows and layout breakages when using a mobile device (esp. smaller ones) with the beatmap discussions system.

Reduces padding/spacing in some places and adds a bunch of horizontal-scroll in others to handle overflow (which is a technique we're using elsewhere on the site, but not sure how I feel about doing this everywhere without any obvious indicator that it's a thing... maybe it's a common enough pattern that people know to swipe sideways when there's overflow?)

Things that I didn't get around to finishing that should probably be addressed later:
- "pinning" the new discussion/review pane feels useless on mobile - it should probably just fill the screen and instead be a floating show/hide toggle or something
- related to the previous point, being able to scroll the page underneath while the pane is "pinned" feels pointless on mobile (not enough screen real-estate to actually reference discussions), so we probably should just disable scrolling of the page while the pane is visible to avoid a double-scrolling situation that appears to be doing nothing


notbakaneko (approved):

resources/assets/less/bem/page-mode.less
```diff
@@ -76,6 +83,12 @@
       margin-top: 5px;
       vertical-align: middle;
     }
+
+    @media @mobile {
+      &:last-child {
+        padding-right:10px;
```
notbakaneko, line 89: whoops, missing a space
```suggestion
        padding-right: 10px;
```

notbakaneko (changes requested):

notbakaneko (approved):
if only `scrollIntoViewIfNeeded` was supported by Firefox :thinking:

notbakaneko (approved):

PR #6620, opened 2020-09-10
Change default for comment reply notification option

### LINKED ISSUE #6618: Notifications for replies to comments defaults to off
image

This should not be the case.

nanaya:
Temporary fix until someone figures out better place to set the default.

Resolves #6618.


notbakaneko (changes requested):
`testCommentReplyNotification` also needs to be updated

notbakaneko (approved):

PR #6624, opened 2020-09-10
Don't render beatmap discussion starting post twice

nanaya:
Speed up rendering a bit by achieving current design through grid and not rendering the message and accompanying components twice.

- [x] depends on #6622


notbakaneko (approved):
solved by ~~flexbox~~ grid

PR #6626, opened 2020-09-10
Skip rendering hidden beatmap discussions

nanaya:
The side effect is per-item state (mainly expand/collapse) is gone when switching filters.
I doubt anyone care enough about it (and the state is already gone by switching another things) so it should be fine.


notbakaneko:
If only there was some way to store state with the discussion object 🤔

nanaya:
The collapse state is put back in #6640.

notbakaneko (approved):

PR #6630, opened 2020-09-11
Add FastRandom PRNG

cl8n:
will be required for logic of selecting bundled maps, and maybe more stuff in the future since this is used for a few different things in the osu client. there's a lot of weird bit masking going on here because PHP doesn't have the data types for fastrandom...

### DIFF
--- a/app/Libraries/FastRandom.php
+++ b/app/Libraries/FastRandom.php
@@ -1,0 +1,82 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace App\Libraries;
+
+use InvalidArgumentException;
+
+/**
+ * A fast PRNG used in the osu! client
+ *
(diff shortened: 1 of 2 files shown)


peppy (approved):
Implementation seems correct as far as I can tell.

notbakaneko:
Algorithm needs to be documented somewhere else; link is no longer live.

peppy:
Just remove the link and rely on the lazer one, it's been proven accurate.

peppy (approved):

peppy:
Honestly we could go without this implementation and rely on PHP's instead, if that's a decided direction. The client will eventually be updated to be server-dependent to get this bundled listing, so this is temporary at best.

nekodex:
I'm confused as to why we need to port over the exact PRNG algorithm from lazer? Do the results need to be exactly identical to lazer for some reason? We already have a seeded shuffle (see `seeded_shuffle` in `helpers.php`) that could be used if a deterministic randomisation is required?

peppy:
It's to match what stable is using to decide on bundled maps. This allows server and client to match without making changes to stable.

cl8n:
well that's a very unhelpful diff since I moved stuff around. I renamed `nextInt`=>`nextInt32` and `nextNonnegativeInt`=>`nextInt31`, because "nonnegative" makes it seem like `nextInt` produced negative values, but that only happens on 32bit php which I assume nobody is using

PR #6641, opened 2020-09-14
Remove unused variable

nanaya:
Not sure what was in it before :thinking:


notbakaneko (approved):
looks like user groups for badges 🤔

PR #6644, opened 2020-09-14
Beatmapset download popup

nanaya:
So this is a thing. There are a few problems with the box:

- paragraph spacing is messed up for languages with longer lines than English
- pippi background is white
- I'm not sure what that line above the action button is for
- the action button isn't wrapped so entire window is rather broken on smaller window
  - probably fine if nobody downloads map through phone? <sub>(cue people with only mobile internet and can't tether so downloading maps from phone)</sub>

### DIFF
--- a/resources/assets/coffee/_classes/beatmapset-download-observer.coffee
+++ b/resources/assets/coffee/_classes/beatmapset-download-observer.coffee
@@ -2,12 +2,14 @@
 # See the LICENCE file in the repository root for full licence text.
 
 class @BeatmapsetDownloadObserver
-  targetSelector: '.support-osu-popup'
+  targetSelector: '.js-support-osu-popup'
   container: '#popup-container'
-  wrapperClass: 'empty-popup'
 
   constructor: ->
     $(document).on 'click mouseup', '.js-beatmapset-download-link', @quotaCheck
+
+
+  $currentPopup: => $(@targetSelector)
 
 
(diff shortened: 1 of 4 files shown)


peppy:
That dialog really needs a redesign. Looks almost like scammy advertising as it is...

peppy:
@arflyte you said you were going to look at this, did you make any progress? we discussed having it display somewhere inline on the page, like the changelog.

arflyte:
Figma_200928-005514

Thinking of having a pop from the bottom GDPR style. Check the link below for samples.

https://www.figma.com/file/vicEqgDFJYngk9PTAxVMng/Beatmap-Set-Popup?node-id=0%3A1

peppy:
Looks better, but I'd rather it's not something people have to actively close. Just something that exists inline in the page and you can ignore if you don't care.

peppy:
@arflyte do you see replies if i don't highlight you ever time? i guess not?

arflyte:
GAH. I missed this. So, since the notification is the same size as the normal page, it can be placed at the top too like other inline notifications. If you think the pippi makes the notification too big, she can be resized (or removed).

peppy:
In further discussion, since this isn't being displayed and likely will never be, it's probably best to destroy the code for now.

I think if we do add something like this back, it will be non-popup, similar to what's done on the changelog pages. But I'm not sure that will happen either.

PR #6650, opened 2020-09-15
Reduce initial work on beatmap discussion post render

nanaya:
Delay rendering/calculating various things. Barely measurably useful :upside_down_face:


notbakaneko (approved):
Apparently using `renderToStaticMarkup` browser-side is ok now? 🤷

PR #6653, opened 2020-09-15
Fix error when viewing discussions through user modding profile

cl8n:
page broke in #6627 because `readPostIds` can be undefined, and previously `_.includes` was silently handling it


cl8n:
rip travis 😔

PR #6655, opened 2020-09-15
Remove user brick background

### LINKED ISSUE #6647: backgrounds of brick user cards are distracting
probably more of a personal nitpick if anything but posting here to see if anyone else feels the same way. the brick user cards show user banners in the background, but they're so small that I can barely tell what any of them are, and it ends up looking more like random colors that make the text stand out less for no reason

[with background (current)](https://user-images.githubusercontent.com/9954349/93160957-58ae5900-f6c6-11ea-90c2-a750ec25ccb1.png) vs [without background](https://user-images.githubusercontent.com/9954349/93160966-5f3cd080-f6c6-11ea-8244-3d7133339c0c.png)

I find that without the backgrounds it's a bit easier to visually parse a large list. it would also make the page load faster without all of the images

cl8n:
closes #6647 , if that's the best way to handle it


resources/assets/less/bem/user-card-brick.less
```diff
@@ -4,18 +4,17 @@
...
-  margin: var(--item-margin);
-  flex: none;
+  .link-white();
+  .link-plain();
 
+  margin: var(--item-margin);
   width: min-content;
   max-width: 100%;
-  background-color: @_background-color;
+  background-color: osu-hsla(@osu-colour-b4, 0.8);
```
nanaya, line 17: Don't use `osu-hsla` anymore.

```suggestion
  background-color: hsla(var(--hsl-b4), 0.8);
```

resources/assets/less/bem/user-card-brick.less
```diff
@@ -4,18 +4,17 @@
 .user-card-brick {
   @_top: user-card-brick;
   @_fade-duration: 220ms;
-
-  @_background-color: @osu-colour-b6;
   @_border-radius: @border-radius-large;
 
   --border: transparent;
```
nanaya, line 9: not exactly this but `::after` should be combined with main block as it was only needed because of the background layers.

Additionally, it can be switched to use plain border and adjust the padding.

nanaya (changes requested):

nanaya (approved):

PR #6658, opened 2020-09-16
Faster discussion timeline chart

nanaya:
Simple enough without needing for d3.


peppy:
do we use d3 elsewhere?

nanaya:
yeah, charts in profile page, online users at dashboard/landing page, changelog usage, etc.

resources/assets/lib/beatmap-discussions/chart.tsx
```diff
@@ -0,0 +1,56 @@
...
+  praise: 'praise',
+  problem: 'problem',
+  suggestion: 'suggestion',
+};
+
+export default function Chart(props: Props) {
+  const items: React.ReactNode[] = [];
+
+  if (props.duration !== 0) {
+    forOwn(props.discussions, (discussion) => {
```
notbakaneko, line 24: I think this can be `for...of Object.values(props.discussions)` instead?

notbakaneko (changes requested):

notbakaneko (approved):

PR #6660, opened 2020-09-16
Move back to upstream Slate

nekodex:
The fixes we forked to address have been since been merged in, so we can stop maintaining our seperate fork.

Some changes were made to the typing of custom attributes on `Element` (from any -> unknown), so they need to be manually typecast now when used (at least until the discussions about Slate implementing declaration merging and/or generics to handle custom attributes better result in... an outcome).


nanaya:
Needs merge master and additional type fixes.

nanaya (approved):

PR #6665, opened 2020-09-17
Better duration format when the file hasn't been loaded

### LINKED ISSUE #6659: Audio Player shows zero as total length
Audio players seem to show `0:00` as their total length:

Zero length total
---
And the correct total value appears when you start the track:

Updated length total

nanaya:
Or if the duration is unknown.

Resolves #6659.

### DIFF
--- a/resources/assets/less/bem/audio-player.less
+++ b/resources/assets/less/bem/audio-player.less
@@ -8,7 +8,7 @@
   --play-button: @fa-var-play;
 
   --current-time: '0:00';
-  --duration: '-:--';
+  --duration: '--:--';
   --progress: 0;
 
   height: @_height;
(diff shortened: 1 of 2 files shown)


nekodex (changes requested):
I think `--:--` is probably a better default? Pretty sure it's more commonly used as a placeholder for time? `-:--` looks weird to me

Not sure about other media players, but VOX uses `--:--` at least:
Screen Shot 2020-09-18 at 6 20 36 pm

Also seems odd to show `0:00 / --:--`
Screen Shot 2020-09-18 at 6 31 21 pm
Maybe hide the current time altogether and just show `--:--` until it's loaded?

nekodex (approved):

PR #6667, opened 2020-09-17
Make backspace and delete behave the same inside of empty nodes (i.e. delete the node)

### LINKED ISSUE #6460: Deleting first paragraph in mod review is difficult
if a paragraph is the first part of a mod review, the only way to delete it is by pressing backspace at the beginning of the next issue (shown [here](https://www.dropbox.com/s/djhsasbpldzxofx/cmJiVEofAx.mp4?dl=0)). there should be a better way to do this?

### LINKED ISSUE #6520: More consistent backspace/delete behavior on discussion review editor
Currently, both delete and backspace behave differently depending on the current and adjacent blocks

### backspace
previous | current | result
---|---|---
empty | empty | deletes current block
empty | filled | deletes previous block &larr; :question: 
filled | empty | deletes current block
filled | filled | deletes current block, moves current text to previous &larr;  :question:  
none (current is first) | empty | deletes current block

### delete
current | next | result
---|---|---
empty | empty | deletes current block
empty | filled | deletes current block
filled | empty | deletes next block &larr;  :question: 
filled | filled | deletes next block, moves next text to current &larr;  :question: 
empty | none (current is last) | deletes current block

I think at least when current is empty, both backspace and delete should delete current block. I'm not sure about filled/filled, backspace's empty/filled, and delete's filled/empty situation though.

nekodex:
After Playing around with the scenarios in #6520, I'm thinking it feels best if we only make a change to the delete/backspace behaviour within empty nodes (to remove the current node), but otherwise just leave the other cases as is?

They kinda make sense with how I'd expect delete and backspace to work w.r.t. text before/after the cursor? I'm open to suggestions if anyone has a good argument for if/how they should behave differently.

fixes #6460, fixes (partially) #6520


notbakaneko (changes requested):
Right now, if the current block is not empty, it collapses the next block into it - but if it's empty it does backwards instead which seems confusing, even more so if the blocks are different types.
out

I think how deleting should behave needs more discussion in general:
https://github.com/ppy/osu-web/issues/6520#issuecomment-698198316

nanaya (changes requested):
Things explode if you press delete/backspace on empty review. Press clear, select editor, press delete/backspace on that empty post. Console will log bunch of errors and nothing can be typed in.

nanaya:
Handling pressing delete/backspace on non-empty block can be done in separate PR (once we figure out what to do in those cases).

resources/assets/lib/beatmap-discussions/editor.tsx
```diff
@@ -179,6 +180,11 @@ export default class Editor extends React.Component<Props, State> {
     if (!this.props.editMode) {
       const content = JSON.stringify(value);
 
+      // prevent document from becoming empty (and invalid) - ideally this would be handled in `withNormalization`, but that isn't run on every change
+      if (value.length === 0) {
+        value = this.emptyDocTemplate;
+      }
```
nanaya, line 186: this should be done regardless of `editMode` prop as same problem happens when editing reviews

nanaya (changes requested):

nanaya (approved):

PR #6671, opened 2020-09-17
Replace hardcoded values in ranking logic

cl8n:


app/Console/Commands/ModdingRankCommand.php
```diff
@@ -69,7 +69,7 @@ private function rankAll($modeInt)
         $toBeRankedQuery = Beatmapset::qualified()
             ->withoutTrashed()
             ->withModesForRanking($modeInt)
-            ->where('queued_at', '<', now()->subDays(7));
+            ->where('queued_at', '<', now()->subDays(Beatmapset::MINIMUM_DAYS_FOR_RANKING));
```
nekodex, line 72: May as well also move `Beatmapset::MINIMUM_DAYS_FOR_RANKING` to config while we're at it?

nekodex (changes requested):

nekodex (approved):

PR #6675, opened 2020-09-18
Remove slack notification for forum updates

nanaya:
Leaving slack endpoint config because it's also used for store notification slack... which also isn't used? ?_?


peppy:
I think we can remove the store one too..

notbakaneko (approved):

PR #6682, opened 2020-09-23
Add username prefix search

### LINKED ISSUE #6652: Allow longer user prefix searches
Searching for long usernames doesn't show the users until typing the full name. This makes search rather useless.

For whatever reason this was fixed only for beatmaps (https://github.com/ppy/osu-web/issues/6284). It should be applied to *all* search types.

notbakaneko:
closes #6652

- add prefix searching
- removes ngram filter (current settings deprecated in future versions of elasticsearch and had dubious usefulness anyway)

Index will need to be updated:
`php artisan es:index-documents --types=users --cleanup --yes`


nanaya:
hmm, this change means xxultralaser_1986xx can't be searched with "ultralas" keyword anymore :thinking:

peppy:
not sure how common such searches are. is that already the case for beatmap searching?

nanaya:
it is already the case for beatmaps. Although for username I think it's more common to not have space between words?

peppy:
How about underscores? Is that still considered a breaking character?

notbakaneko:
It doesn't split on underscore, that requires a custom tokenizer plugin or newer Elasticsearch (6.8+) for native support

nanaya (approved):

PR #6683, opened 2020-09-24
Replace StyleCI

nekodex:
This PR replaces StyleCI with PHP_CodeSniffer (running via GitHub Actions)

As part of this, the PR also fixes the codebase to make it compliant with PSR-2 (and most of the way to PSR-12, except when in conflict with Laravel's code standard) as to not immediately fail after the switchover.

A few benefits of this switch:
- More accurate linting (StyleCI's checks weren't very strict and regularly let non-compliant code through)
- Code quality checks that pick up issues that aren't strictly code-standard related (unused vars/use statements, etc)
- GitHub annotations will now show the issues on PRs themselves without having to go off-site to view diffs
- No longer reliant on checks that can only run server-side, now the checks can be run locally (`$ vendor/bin/phpcs`, with a fixer: `$ vendor/bin/phpcbf`)

Main downside is that it takes longer to run (~1.5-2min)

Note that StyleCI will fail on this PR because the last commit removes `.styleci.yml`, for reference the only failure is:
Screen Shot 2020-09-24 at 10 43 24 pm
...which contradicts PSR-12, thus ignorable.

- [ ] disable StyleCI


notbakaneko (commented):
Tentatively approving, also seems a lot more usable than StyleCI.

peppy:
Any rule changes can be discussed post merge, since they all look quite sane.. Let's get this in and see how it behaves (hard to judge prior to merge).

PR #6684, opened 2020-09-24
Add option to disable minimum play count on verified sessions for certain actions

notbakaneko:
defaults to true (allows verified sessions to skip play count requirement)
add `USER_MIN_PLAYS_ALLOW_VERIFIED_BYPASS=false` to .env to disable the play count bypass.

Effects
- chat
- posting to beatmap discussions
- posting to forum

known issues:
~~Failing the check when not verified will automatically send a verification email out - the issue is, on chat, the verification dialog doesn't show when the message fails to send and doesn't show what the error is.~~ It should only attempt verification if the option is enabled, now. Chat still has the issue where if doesn't show the reason for error.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -208,6 +208,7 @@ SITE_SWITCHER_JS_HASH=
 # USER_POST_ACTION_VERIFICATION=true
 # USER_MAX_LOGIN_ATTEMPTS=10
 # USER_MIN_PLAYS_FOR_POSTING=10
+# USER_MIN_PLAYS_ALLOW_VERIFIED_BYPASS=true
 # space delimited, list of groups (the identifier) which user can be renamed when inactive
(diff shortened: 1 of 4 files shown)


nekodex:
Not sure if we ever really want 'verified' to be allowed to bypass the minimum play count?

Digging through the history on `OsuAuthorize`, it _seems_ like the verification bypass was put there to allow chat from the API(lazer) and InterOp(bancho) endpoints to work (not sure if it was intentional to bypass the min playcount though?) (ref: #5352, maybe @nanaya can confirm?)

Which kinda brings up the question as to what the actual behaviour should be for new users attempting to chat in-game, or new users who exclusively play on lazer (which afaik doesn't update playcount)?

nekodex (approved):
Gonna just merge this as it's required to block spamming - will extract discussion about behaviour out into a seperate issue.

PR #6686, opened 2020-09-25
Add support for user registration email domain block list

notbakaneko:
Migrates a missing feature from old web.
Check is similar to the one for usernames and `phpbb_disallow`.
The format is `*@domain` or even `*@domain*` where `*` becomes the regex equivalent of `.*?`
`ban_start` is not checked.

`isValidEmail()` is a bit janky as it'll just append additional errors on multiple calls.

Not sure if the migration should just go into the base tables, but then we added this `2020_05_15_083037_sync_structure` 🤔 

- [ ] add `2020_09_25_092826_create_phpbb_banlist` to migrations

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -1848,14 +1848,7 @@ class User extends Model implements AuthenticatableContract, HasLocalePreference
         }
 
         if ($this->isDirty('user_email') && present($this->user_email)) {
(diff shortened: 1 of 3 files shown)


nekodex (approved):
Seems to work fine, can clean up later as necessary

PR #6698, opened 2020-09-29
Update documentation for rankings API endpoint

### LINKED ISSUE #6692: Api documentation should mentation how you can filter leaderboards by country
https://osu.ppy.sh/docs/index.html#get-ranking doesn't mention anywhere that you can use the `country` parameter to filter by Country.

nanaya:
- remove `authenticated` badge
- use proper format for `urlParam`
- add missing `queryParam`
  - I think those are all...?

Resolves #6692.


nekodex (changes requested):
I think you missed `page`... although, maybe we should have a generic section explaining pagination (the types employed and how to use cursors, etc)? Then we can just link to it with "has pagination" or something to save having to repeatedly explain?

Also maybe the default should be listed for optional params that have one (like `filter` for example)

nekodex (approved):

PR #6699, opened 2020-09-30
Fix the 'user card' padding/alignment of modding posts

### LINKED ISSUE #6694: Replies to modding posts don't seem to be aligned/spaced correctly
Example:
image

If you look, my reply isn't exactly aligned with the parent comment.

My browser version is Edge 85.0.564.63.

Thanks.

nekodex:
fixes #6694, a regression from #6619


nanaya (commented):
I don't remember if the alignment originally looks like this with replies aligned a bit more to the right following indentation :thinking:

nekodex:
I just put back the `min-width: 150px;` that I took out, but only for desktop (see: https://github.com/ppy/osu-web/pull/6619/files#diff-2fdb1a47e9622771aaeed5d75331eaa6L40)

nanaya (approved):

PR #6705, opened 2020-09-30
Add correct momentjs locale mapping for Norwegian

nanaya:
Er ( ﾟ ヮﾟ)

(using `nb` / Bokmål as it seems to be the most common and has some words in it in current general `no` translations)


notbakaneko (approved):
the way they've named the locales seems like all kinds of wtf 😒

PR #6706, opened 2020-09-30
Use the same user lookup logic for user cards as user profiles

### LINKED ISSUE #6637: User card can't find users who renamed
https://osu.ppy.sh/users/Byakugan249 this link will show

but loading it will go to the correct user: https://osu.ppy.sh/users/660484

nekodex:
I'm not sure how ideal this is, seeing as there's a greater than zero chance that hovering a username will show a user card for a (seemingly different) user(name)... but at least it will be consistent and show the same user as the profile you'd get from clicking on the link? 

(...I guess one option would be to add the previously-known-as thingy to the user card?)

fixes #6637


notbakaneko:
I think `lookupWithHistory` should be updated to skip the additional history lookup if doing an id-only lookup

app/Models/User.php
```diff
@@ -472,6 +472,11 @@ public static function lookupWithHistory($usernameOrId, $type = null, $findAll =
             return $user;
         }
 
+        // don't perform username change history lookup if we're searching by ID
+        if (ctype_digit((string) $usernameOrId)) {
+            return null;
+        }
```
notbakaneko, line 478: Usernames and name changes can be all digits; when it's an id-only lookup then then there's no point in doing the username history lookup since it's not looking for a name.

notbakaneko (changes requested):

notbakaneko (approved):

app/Models/User.php
```diff
@@ -478,6 +478,11 @@ public static function lookupWithHistory($usernameOrId, $type = null, $findAll =
             return $user;
         }
 
+        // don't perform username change history lookup if we're searching by ID
+        if ($type === 'id') {
+            return null;
+        }
```
nanaya, line 484: I don't see how this will ever be true considering `$type` is passed as null everywhere...? :thinking:

notbakaneko: Problem for the future 🤔

PR #6707, opened 2020-09-30
Add user id to profile bbcode tag

### LINKED ISSUE #4050: Change the behavior of [profile] bbcode tags

nanaya:
Continues #6137, resolves #4050. Doesn't change the link output format so user card is still shown as expected.


notbakaneko (commented):
it occurs to me that something like `[profile=idOfUserA]nameOfUserB[/profile]` could be misleading 🤔

notbakaneko (approved):

Issue #6442, opened 2020-08-06
beatmap listing sorted by playcount isn't always in order


cl8n:
it's most apparent in pending, where just a few pages down the maps don't seem to be in order of playcount at all. maybe related to how often the es index gets updated

peppy:
I don't believe it is ever getting updated by playcount (there's no update schedule for it any more).

nanaya:
The endpoint has been added. All that's left is for the score processor or something related to hit it every now and then with list of beatmap ids. @peppy

peppy:
I've added reindexing support via the existing submission code in a bit of a temporary way, but should be enough to keep these relative up-to-date 👍 .

PR #6717, opened 2020-10-01
Add support of bulk indexing of documents through LIO

notbakaneko:
Add support for bulk indexing via `_lio` in order to update the beatmapset playcount (#6442) stored in elasticsearch.

Currently limited to `Beatmapset` and `User`; it _does not_ update anything else other than the model's corresponding entry in elasticsearch.

Payload format is json:
`POST /_lio/indexing/bulk`
```jsonc
{
  "beatmapset": [1,2,3], // list of ids
  "user": [1,2,3],
}
```
or equivalent `beatmapset[]=1&beatmapset[]=2&beatmapset[]=3&...etc`


app/Jobs/EsIndexDocumentBulk.php
```diff
@@ -0,0 +1,49 @@
...
+
+        $this->className = $className;
+        $this->ids = $ids;
+    }
+
+    public function handle()
+    {
+        $dummy = new $this->className();
+
+        $models = $this->className::esIndexingQuery()->whereIn($dummy->getKeyName(), $this->ids)->get();
```
nanaya, line 37: chunk?

app/Http/Controllers/LegacyInterOpController.php
```diff
@@ -70,6 +72,27 @@ public function indexBeatmapset($id)
         return response(null, 204);
     }
 
+    public function indexDocuments()
```
nanaya, line 75: move to its own controller in InterOp namespace

(should also look into splitting the rest as well one day™)

app/Jobs/EsIndexDocumentBulk.php
```diff
@@ -0,0 +1,49 @@
...
+class EsIndexDocumentBulk implements ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, Queueable;
+
+    protected $className;
+    protected $ids;
+
+    public function __construct(string $className, array $ids)
+    {
+        if (!new $className() instanceof Indexable) {
```
nanaya, line 25: `is_a` is a thing (and can accept class name string)

nanaya (changes requested):

nanaya (approved):

Issue #5477, opened 2020-01-03
include multiple download mirrors for packs


FlyingStefan:
It has been removed with #2235 but it was also requested by the people at the same time to have an alternative to Mediafire and having now MEGA as platform to download packs would require this support back. Additionally now that I have noticed some packs from Mediafire got recently unavailable and had to be reuploaded.

I understand it's a bit inconvenient to come up with that but I didn't know this has been intentionally removed with the new website.

peppy:
I would actually prefer we just move everything to mega, or host them ourselves. Mediafire is a pretty shoddy service.

FlyingStefan:
If you can provide around 200gb of room for all packs (and upcoming packs in the next months) I'd be more than happy to move them.

peppy:
The idea behind hosting them elsewhere was that there was added redundancy was osu! to ever explode into flames. I think there may still be value in having that level of redundancy.

Would mega work as a solution? I'm happy to pay for an account there to provide ample space for hosting. At least it is ad-free.

FlyingStefan:
Mega is a great idea! Thanks to the generosity of the osu!dan project a plan of 4TB could be paid out for the Mega account so for now it's fine.

And I'll gladly accept your offer once the current plan runs out.

peppy:
Let me know if anything needs to be done from our end.

nanaya:
So what's the plan with this? Is it still relevant or single place as of current is fine?

FlyingStefan:
this is actually good timing, because the MEGA subscription is about to run out very soon and I'd require some assistance for that. 

Also, regarding alternative download mirror: I have set up a mediafire mirror with all beatmap packs till october 2020 and this should be added into the description of the beatmap pack page, in case people prefer this platform for whatever reason or have problems to use MEGA: https://app.mediafire.com/t01qat9p8m4wc

peppy:
@FlyingStefan is it fine to close this issue based on switching to mega for hosting here?

Joehuu:
Packs are now hosted directly, closing.

PR #6729, opened 2020-10-05
Simplify beatmap pack url render

nanaya:
Transition to single url is done. Kind of, depending #5477.


notbakaneko (approved):
Assuming there's only a single url.

PR #6730, opened 2020-10-05
Include more user profile related counts

### LINKED ISSUE #6633: Recent scores counter disappears when theres atleast 1 score
e
e

nanaya:
Resolves #6633. Some of the counts may not reflect what can actually be fetched thanks to pagination hard limit.

The count seems to be fast enough (including on account with 60k+ plays) so it should be fine...? :shrug:


notbakaneko (changes requested):
conflicts

notbakaneko (approved):

PR #6731, opened 2020-10-05
Refactor web chat client

notbakaneko:
A precursor to overall cleaning up the Chat API before adding chat scopes. Eventually, we want to stop needing the client to constantly poll for updates (partially because getting the updates is kind of inefficient, even if there were no updates) and handle updates on a per-channel basis and/or have them pushed out from the server.

A lot of this PR is moving things and removing parts no longer necessary.
In general:
- chat worker no longer sends chat message; now handled by channel store.
- chat orchestrator removed, state modification moved to the stores that 'own' them instead of anyone getting to set them;
- `Promise` wrappers around `jqXHR` no longer needed.
- `Message.update()` removed because it was just assigning to itself in places it was used.
- minimizing the number of references to `RootDataStore`; Store dependencies are now as explicit as possible - trying to avoid cases where a class goes up it's dependency tree just to do back down into another tree, or ferrying variables across multiple levels. In these cases it should probably be using an application-level context to access them or dispatch a message.
- When something dispatches a `DispatcherAction`, it's task related to it is effectively 'done' (its unit-of-work is compete) - it shouldn't make the assumption that the event runs; e.g. if:
  ```
  *typing*
  <SelectChannelEvent>
  <SendMessage>
  ```
  there should be no assumption that the channel was changed before sending the message.
- more use of `async`/`await`
- generally trying to limit the use of `DispatcherAction`s unless it's for decoupling like breaking dependency chains, or the result of an external message (e.g. received from websocket or another tab); if a component can observe state from a store it already has a direct dependency on, that is preferable to handling `DispactherAction`


notbakaneko:
this has been split up and more changes added into other PRs

PR #6733, opened 2020-10-06
Add links to user titles

### LINKED ISSUE #3706: Add links to user titles
for example, Beatmap Nominator could link to the group or wiki page, and Featured Artist could link to their artist listing

cl8n:
resolves #3706, downside is if you want to re-use the same title with multiple URLs (like "Featured Artist" from the issue) you'd have to make more ranks.

### DIFF
--- a/resources/assets/less/bem/profile-info.less
+++ b/resources/assets/less/bem/profile-info.less
@@ -170,6 +170,7 @@
   }
 
   &__title {
+    .link-white();
     font-size: @font-size--title-small-4;
     padding: 0 0 5px;
     line-height: normal;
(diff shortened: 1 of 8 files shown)


resources/assets/less/bem/profile-info.less
```diff
@@ -170,6 +170,7 @@
   }
 
   &__title {
+    .link-white();
```
nanaya, line 173: the title should probably not be full width

nanaya (changes requested):

nanaya:
Conflicts.

nanaya (approved):

PR #6734, opened 2020-10-06
Add total and unread counts to modding watchlist

### LINKED ISSUE #4148: add info to the modding watchlist page header
image
currently, the modding watchlist header has a no-height box on the right side
so my proposal is:
add information to it, just like 
image
the forum subscriptions header states the number of subscribed threads, and the number of ones with new replies.
on the modding discussions page, it could be better if the box was used to stay the number of subscribed beatmaps and the number of ones with new replies!

cl8n:
resolves #4148


app/Models/BeatmapsetWatch.php
```diff
@@ -65,6 +65,11 @@ public function scopeRead($query)
         });
     }
 
+    public function scopeUnread($query)
+    {
+        $query->whereColumn('last_read', '<', 'last_notified');
```
nanaya, line 70: Nothing currently updates `last_notified` so the result is wrong.

nanaya (changes requested):

app/Models/BeatmapsetWatch.php
cl8n: huh... I just flipped scopeRead/isRead expecting it to work, how does `isRead()` have the right result on live site?

nanaya: It's not currently used anywhere.

nanaya: er, sorry, it is used somewhere. But it's currently wrong as the original thing that updates the column has been replaced with something else.

nanaya (approved):

Issue #6689, opened 2020-09-25
Multi-mode beatmapsets only require 2 nominations


Ephemeralis:
Currently, a set that has difficulties from all 4 game modes only requires the standard two nominations to proceed into qualified. Logically speaking, shouldn't a hybrid set require 2 nominations per game mode present? If 4 game modes are present, 8 nominations should be required, with enforcement from NAT end to ensure that appropriately qualified BNs are awarding those nominations.

I've asked around as to why this isn't already the case and have been met with a mixture of "the current system doesn't support it" and "it's tradition that it's always been this way". We should address this. The fact that giant hybrid sets are entering Ranked with presumably cover to have less than normal oversight is a huge problem.

cl8n:
"the current system doesn't support it" for the same reason as #5484

cl8n:
should nominators of multiple modes be allowed to choose which modes they're nominating a map for? or just one nominate button to do all of the modes they're allowed to?

Ephemeralis:
> should nominators of multiple modes be allowed to choose which modes they're nominating a map for? or just one nominate button to do all of the modes they're allowed to?

The former. Some hybrid BN may opt to only pass judgment on a particular mode they're qualified in, nothing is requiring them to check every mode if they don't want to.

Greenshellus:
I'm generally open to this, one issue I'm seeing with this however is that people will feel less open towards creating hybrid sets in the first place, as one of the perks actually was to only require 1 BN per mode to confirm a set's readiness. If we make it so it always requires 2 BNs for _each_ mode, it's gonna be much more of a hassle for mappers to manage.

Then again, hybrids should get the same treatment in terms of checking each mode carefully as regular single-mode sets do. As it is now, ranking a big 20 GDs Standard set with a side of 3 half-assed Taiko diffs for example is somewhat easy for the amount of difficulties there are, as only one BN has to look at all the Standard diffs, while the other BN only skims through those three diffs in less than an hour and can call it a day. That's not proper checking of a set, that's dumb.


If we implement this though, I'd like to make a proposal that goes hand-in-hand with it.


Allow Probationary BNs to nominate hybrid sets as well.

The big issue with Probationees and hybrid sets always was that they have to take responsibility for the respective mode they're nominating on their own, while ranking any set requires _at least_ one Full BN to check a set. Which, obviously, works for single-mode sets as two BNs cover the same thing, but not so much for hybrid sets, as you'll only have one BN per mode.

[...]

PR #6747, opened 2020-10-08
Add ability to associate user group members with playmode(s)

### LINKED ISSUE #6681: Add 'new sort' by in BN pages by game mode they're nominated
I referred to this page : https://osu.ppy.sh/groups/28

It would be helpful in my opinion especially with new mappers when they searched bn for nominated.

nekodex:
aka: The ability to mark NAT/BN members as responsible for specific game modes

This is a prerequisite for the nomination changes, as discussed in #6689 and sets up the usergroup/playmode mapping.

Enabling playmode support is done at a per-group level, while the playmodes themselves are defined on the user's membership relation (`phpbb_user_group`) - this means that we can assign playmodes at a per-usergroup-per-user level. e.g. a user can be a NAT for osu! mode but also a BN for osu!+osu!taiko (not sure if this will ever happen, but it's supported)

This PR also has some visible changes:
- playmode(s) are now shown in the user group badge (if enabled for the group and the member has playmodes defined):
Screen Shot 2020-10-08 at 3 43 13 pm
- user badges have also been added to user cards (to make group membership more obvious):
Screen Shot 2020-10-08 at 3 46 02 pm
- ability added to filter the groups page by playmode has been added (if `has_playmodes ` is set for the group):
Screen Shot 2020-10-08 at 3 57 34 pm

resolves #6681

---

Things to be aware of when deploying:
- [x] check if migrations need to be done beforehand manually (mainly concerned about `phpbb_user_group`)
- [x] group cache needs to be cleared after setting `has_playmodes` in DB: run `app('groups')->resetCache()` via tinker, (this will need to be done again whenever `has_playmodes` is toggled on a group)
- [x] will need to construct/run some DB queries to set the relevant playmodes for current NAT/BN/etc members (update: constructed, just needs to be run after deploy)


nekodex:
marking as blocked because I noticed some edge case issues

nanaya (changes requested):

resources/assets/coffee/osu_common.coffee
```diff
@@ -262,6 +262,18 @@
       window.scrollTo position[0], position[1]
 
 
+  playmodeToInt: (playmode) ->
```
nanaya, line 265: not used?

resources/assets/lib/globals.d.ts
```diff
@@ -62,12 +62,14 @@ interface OsuCommon {
   classWithModifiers: (baseName: string, modifiers?: string[]) => string;
   diffColour: (difficultyRating?: string | null) => React.CSSProperties;
   groupColour: (group?: GroupJSON) => React.CSSProperties;
+  intToPlaymode: (int: number) => string;
```
nanaya, line 65: doesn't exist?

app/Models/User.php
```diff
@@ -1373,19 +1373,34 @@ public function visibleGroups()
...
-                $group = app('groups')->byId($id);
+            foreach ($this->userGroups as $userGroup) {
+                $cachedGroup = app('groups')->byId($userGroup->group_id);
+                if (!$cachedGroup) {
+                    continue;
+                }
+
+                $group = clone $cachedGroup;
+
+                if ($group && $group->has_playmodes) {
```
nanaya, line 1385: the first [truthy](https://www.php.net/manual/en/language.types.boolean.php#language.types.boolean.casting) check is rather useless. Similarly the `optional` wrapper below. Additionally there's no need to `clone` if `display_order` is `null`.

nanaya (changes requested):

nanaya (changes requested):

resources/docs/source/includes/_structures/group.md
```diff
@@ -1,10 +1,18 @@
...
+short_name      | string | Short name of the group for display.
+description     | string | |
+colour          | string | |
+
+
+If [Group](#group) is returned as part of a [User](#user) object, it may also contain:
+
+Field           | Type     | Description
+----------------|----------|------------------------------------------------------------
+playmodes       | string[] | [GameModes](#gamemode) which the member is responsible for, e.g. in the case of BN/NAT
```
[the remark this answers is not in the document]
nekodex: attribute doesn't exist at all unless it's inside user, which is why I listed it separately... but yeah, also nullable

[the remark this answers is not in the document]
nanaya: the transformer doesn't imply that though? :thinking:

nanaya (approved):

nanaya:
`phpbb_user_group` seems to be only about 500MB so it should be fine even run from migration.

peppy:
Remaining two steps will happen after deploy.

PR #6749, opened 2020-10-08
Mark notifications (the model) as read direct from its store

nanaya:
Noticed this weird dependency on WidgetNotificationStackStore for handling notifications as read (especially when it's by notification id).

Move responsibility of marking Notification objects from WidgetNotificationStackStore (and partially from NotificationStackStore) to NotificationStore. 

This slows things down a bit for type/stack read as the NotificationStore doesn't have the grouping so it has to loop the entire array.

As a bonus, this fixes bug where "clear :type" <del>for non-"all" type</del> from notification popup widget sometimes doesn't mark notifications in history page read.

### DIFF
--- a/resources/assets/lib/models/notification-stack.ts
+++ b/resources/assets/lib/models/notification-stack.ts
@@ -53,10 +53,6 @@ export default class NotificationStack implements NotificationReadable {
     return `${this.objectType}-${this.objectId}-${this.category}`;
   }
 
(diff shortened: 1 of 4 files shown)


notbakaneko (approved):
The grouping was supposed to be handled by the stack store to avoid the filtering but it looks like filtering this way seems to have little impact on overall performance compared to the render itself.

PR #6751, opened 2020-10-09
Save Paypal API transaction id when processing orders

notbakaneko:
Save more metadata when processing Paypal payments. This is mostly useful for looking up failed orders, etc.

The main changes are the transaction's `PAYID` number (which is what their API uses) is stored when starting the transaction, and the IPN transaction number is written even if the order fails processing.

Moving forwards, the payment provider and the transaction reference id from the provider will be recorded in separate columns. In the case of Paypal, the PAYID is used for the reference since this is the one that's available at the start of the transaction and supports API lookups.
This PR doesn't make any changes to how they are currently read and used, but moving forwards we'll want to read from `provider` and `reference` instead of `transaction_id` on `Order`.

### DIFF
--- a/app/Http/Controllers/Payments/PaypalController.php
+++ b/app/Http/Controllers/Payments/PaypalController.php
@@ -66,6 +66,8 @@ class PaypalController extends Controller
         $order = Order::where('user_id', Auth::user()->user_id)->processing()->findOrFail($orderId);
         $command = new PaypalCreatePayment($order);
         $link = $command->getApprovalLink();
+        // getId() is only available after the payment is created which getApprovalLink() calls.
+        $order->update(['tracking_code' => $command->getPayment()->getId()]);
 
         return $link;
     }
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Payments/PaypalController.php
```diff
@@ -66,6 +66,8 @@ public function create()
         $order = Order::where('user_id', Auth::user()->user_id)->processing()->findOrFail($orderId);
         $command = new PaypalCreatePayment($order);
         $link = $command->getApprovalLink();
+        // getId() is only available after the payment is created which getApprovalLink() calls.
+        $order->update(['tracking_code' => $command->getPayment()->getId()]);
```
nanaya, line 70: wouldn't this be shown to the user if the order status changes to shipped/delivered?

notbakaneko: I put it there because I didn't want to have to add another column to store this but looks like it needs another column 😕

nanaya (approved):

PR #6758, opened 2020-10-13
Simplify notifications history header links

nanaya:
Seems weird using `observe` instead of simply `computed` for the links :thinking:


notbakaneko (approved):
The difference is a `@computed` getter still runs every time it's called even if it hasn't changed while the `observe` only runs when `currentFilter` has changed. (edit: apparently not when it's declared on an `@observer`? 🤔 )

It's ok here since it's only returning unobserved values and the calculation is relatively small.

PR #6763, opened 2020-10-14
Move chat channel and message handling to store

notbakaneko:
more pieces of #6731

- `ChatWorker` only polls for updates now.
- channel and message handling all moved to channel store


resources/assets/lib/stores/channel-store.ts
```diff
@@ -188,4 +192,62 @@ export default class ChannelStore {
 
     this.loaded = true;
   }
+
+  @action
+  private handleChatChannelNewMessages(channelId: number, json: MessageJSON[]) {
+    const messages = json.map((messageJson) => {
+      if (messageJson.sender != null) this.userStore.getOrCreate(messageJson.sender_id, messageJson.sender);
```
nanaya, line 199: (probably not this line but just so it can be marked resolved)

User info is missing on initial load

resources/assets/lib/models/chat/channel.ts
```diff
@@ -133,16 +138,13 @@ export default class Channel {
...
-  updateMessage(message: Message) {
-    const messageObject = _.find(this.messages, {uuid: message.uuid});
-    if (messageObject) {
-      messageObject.update(message);
-      if (messageObject.errored) {
-        messageObject.messageId = messageObject.uuid; // prevent from being culled by uniq sort thing
-      } else {
-        messageObject.persist();
-      }
+  updateMessage(message: Message, json: MessageJSON | null) {
```
nanaya, line 141: ((again) probably not this line but just so it can be marked resolved)

Message order seems a bit weird. First message is sent and then it moves up (last message should be "what's with the order")

nanaya (changes requested):

resources/assets/lib/stores/channel-store.ts
notbakaneko: this change might have been dependent on changes to channel loading 🤔

resources/assets/lib/models/chat/channel.ts
nanaya: probably something to do with server and client time slightly off and the message not updated with actual timestamp anymore

nanaya (approved):

PR #6768, opened 2020-10-16
Remove display of user lastfm

nanaya:
The integration has been broken for a while (and never got added to new web) so remove it.

Keeping backend removal separate as lazer seems to be still using it.


notbakaneko (approved):

Poliwrath:
removed from lazer in https://github.com/ppy/osu/pull/10518/

peppy:
Also lazer can handle missing elements, so won't be a problem with live releases. Should be fine to remove.

PR #6775, opened 2020-10-17
Match docker's elasticsearch image version with composer.json

Techno-coder:
Current elasticsearch PHP version is 6.7.2 updated in d21fb10c6130c3cad4596acb1ea1b29fa15e6167 but docker-compose.yml was not similarly updated, breaking search and seed functionality under the Docker image. This commit rectifies this missing change.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -85,7 +85,8 @@ services:
     image: redis:latest
 
   elasticsearch:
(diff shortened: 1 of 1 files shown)


cl8n:
can go up to 6.8.12 without breaking changes, that's what travis is using now

PR #6783, opened 2020-10-20
Clean up some react and mobx warnings

nanaya:
Noticed them when running tests.

### DIFF
--- a/resources/assets/lib/chat/conversation-view.tsx
+++ b/resources/assets/lib/chat/conversation-view.tsx
@@ -80,7 +80,10 @@ export default class ConversationView extends React.Component<Props> implements 
       this.didSwitchChannel = false;
     } else {
       snapshot = snapshot ?? blankSnapshot();
-      const prepending = this.firstMessage !== this.currentChannel?.messages[0];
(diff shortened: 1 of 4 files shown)


resources/assets/lib/chat/conversation-view.tsx
```diff
@@ -80,7 +80,10 @@ export default class ConversationView extends React.Component<Props> implements
       this.didSwitchChannel = false;
     } else {
       snapshot = snapshot ?? blankSnapshot();
-      const prepending = this.firstMessage !== this.currentChannel?.messages[0];
+      const currentChannelFirstMessage = (this.currentChannel != null && this.currentChannel.messages.length > 0)
+        ? this.currentChannel?.messages[0]
+        : null;
+      const prepending = this.firstMessage !== currentChannelFirstMessage;
```
notbakaneko, line 86: should probably add a `firstMessage` getter to Channel that does the check instead.

notbakaneko (changes requested):

notbakaneko (approved):

PR #6800, opened 2020-10-22
Add language menu for wiki page

### LINKED ISSUE #6493: Display a list of existing languages in osu!wiki articles
Example:
image

osu!wiki will only display the current web language. Some people have more than two languages, showing all available languages suitable for people to choose their suitable third language, and also helping translators to compare.

nanaya:
There's (existing) problem #6817 where clicking any of the wiki links inside the page will revert back to user default language.

Resolves #6493.

- [x] depends on #6799


notbakaneko (approved):

app/Models/Wiki/Page.php
```diff
@@ -159,6 +160,21 @@ public function __construct($path, $locale, $requestedLocale = null)
         $this->defaultSubtitle = array_pop($defaultTitles);
     }
 
+    public function availableLocales()
```
notbakaneko, line 163: should exclude empty pages

notbakaneko (changes requested):

notbakaneko (approved):

PR #6803, opened 2020-10-22
Show only chat notifications when clicking chat icon

### LINKED ISSUE #5773: Show chat notifications when clicking chat button
having a button for chat but showing chat notifications in a different place is really weird, especially when the chat button shows the number of chat messages you have unread.

Clicking on the chat notifications under notifications also redirect you to the relevant chat, but clicking the chat button doesn't.

so basically, proposed solution:

show _chat_ notifications when clicking the chat button, not the notifications button.
OR...
don't show number of chat messages on the chat button. Keep that for notifications only.

notbakaneko:
Originally, this was going to use just the single notification and default to the chat notifications when clicking on the chat icon, switching to all or previously selected filter when clicking on the notifications icon. However, getting the React components and `click-menu` to sync their visibility correctly turned into a complete disaster, especially when having to handle the differences between the desktop and mobile versions.

~~So instead, this just duplicates the notification widget and adds support for showing only one type of notifications in it. The obvious issues here are:~~
~~- duplicate rendering of the widgets~~
~~- notification widget still includes chat notifications~~

Chat notifications are excluded from the widget.

Known issues:
`Clear all` on the notification widget current doesn't animate due to it's state not being the one being updated at the moment.

closes #5773


nekodex (changes requested):
If we're splitting out chat notifications, we may as well go all the way and filter them out of the main notification widget too? It feels pretty silly going half-way and having them displayed in two places at once?

nekodex (approved):

PR #6809, opened 2020-10-26
Add OAuth scope for sending chat PMs

### LINKED ISSUE #6359: Use cases for creating/sending chat PMs through API
forum pms are being used to send automatically generated info-heavy messages (BN evaluations/reports and individual feedback for monthly beatmapping contests), which can total around 100 pms a month. project loved also sends similar template forum pms but i'm not sure if they'd use this feature

manually sending these with new chat involves sending a few messages in a row because of length limits or sending a link to the full content of the message

https://osu.ppy.sh/docs/index.html?javascript#create-new-pm this seems like a doable way to handle these situations

notbakaneko:
closes #6359

~~This is for the very specific case in the issue of sending PMs; more general use chat scopes will come later.~~

A specific scope for chat bots is added, `bot`; only chat bots are allowed to use it.
The PR also adds support for `client_credentials` having an OAuth resource owner (basically, the client), but it's currently only applied for the `bot` scope as it's a breaking change if applied to all `client_credentials` (guest becoming non-guest).

I was looking at adding more generic tests that covered all routes but it turned out not to be so practical on routes that have additional permissions like multiplayer requiring the user to be in the room, etc. This wouldn't be an issue if we could run just the middleware part of the route (and manually constructing the middleware for the request would defeat the purpose of testing it).


nekodex (changes requested):
After discussing with @peppy, we don't really want to open `chat.write` permissions up to third parties, esp. not the ability for them to send messages on behalf of users.

As per issue #6359, we only want bot users to have access to send messages - and only as themselves.

So I imagine the flow should probably be something like:
- Bot owner logs into to the site as the bot and creates an oauth app (so it's owned by the bot account
- Use the client credentials grant method for the above (bot-owned) app to get a token
- Send message via the API using the bot's token

peppy:
Not sure about the "only as themselves" part – @pishifat mentioned he may want to send messages on behalf of another authorised user (which is fine to support – as long as the overall permissions are limited to bot oauth sources).

@pishifat if you can confirm this is the intention that would be helpful too.

pishifat:
sending as an authorized user would be ideal since there's situations where replies are needed. for NAT uses, these replies are most relevant to the user sending the message through this system, so sending as a user would be best even if the bot could receive replies

nekodex (changes requested):
I'm thinking that renaming the scope to `bot` is probably a bad idea, especially when the description of the scope is opaque and doesn't explain to the authorising user exactly what permissions they're granting the app.

Screen Shot 2020-11-20 at 4 12 30 pm
Seeing the above doesn't exactly inform you that "act as a chat bot" actually means "send chat messages on your behalf" in this context...

Probably better to just keep the scope as `chat.write` with the appropriate description, but with the `isBot()` restriction on requesting that scope?

app/Models/OAuth/Token.php
```diff
@@ -43,6 +60,38 @@ public function revoke()
...
+            throw new MissingScopeException([], 'Tokens without scopes are not valid.');
+        }
+
+        if ($this->isClientCredentials() && in_array('*', $this->scopes, true)) {
+            throw new MissingScopeException(['*'], '* is not allowed with Client Credentials');
+        }
+
+        if (in_array('bot', $this->scopes, true)) {
+            if (count($this->scopes) !== 1) {
+                throw new MissingScopeException(['bot'], 'bot scope is not allowed with other scopes.');
```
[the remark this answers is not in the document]
notbakaneko: `bot` allows grants using `client_credentials` to have a user assigned to the `auth()` guard (otherwise, none of the things relying on `auth()->user()` would work) which is incompatible with how `client_credentials` grants currently work, where everything it uses an anonymous user.

`bot` is a special scope that
- is only available to OAuth clients owned by users in the chat bot group.
- is only allowed with `client_credentials`.
- lets the request have an auth'd user.

the alternative is some combination scope like `chat.write identify`, but it's still only going to be available to bots only.

There's also going to be the issue when using `client_credentials`, the `client_secret` is going to have to be deployed to whatever environment is going to be requesting the token; that passport's scopes aren't restricted by client and any arbitrary scope can be requested is an issue - this is one of the reasons `client_credentials` grants are currently forced to be guest only.

app/Models/OAuth/Token.php
```diff
@@ -43,6 +70,58 @@ public function revoke()
...
+        }
+
+        if ($this->canDeletage()) {
+            if (!$this->client->user->isBot()) {
+                throw new InvalidScopeException('This scope is only available for chat bots.');
+            }
+
+            if (!$this->isClientCredentials()) {
+                throw new InvalidScopeException('bot scope is only valid for client_credentials tokens.');
+            }
```
nekodex, line 111: I wonder if it's worth rewriting these conditionals to group the `isClientCredentials()` stuff together - might make it more obvious which restrictions apply to which auth type?

nekodex (changes requested):
seems fine, typos notwithstanding

app/Models/OAuth/Token.php
notbakaneko: These are grouped by scope/permission requested, grouping by `isClientCredentials()` just means more scope and user bot checks need to be added or duplicated to the conditional since the token type is the exception not the rule ("you can do `a` unless in case of `x`", vs. "in case of `x` you can't do `a`...unless `y` but not if `c`")

resources/views/vendor/apidoc/partials/info.blade.php
```diff
@@ -247,22 +251,34 @@
 </aside>
 
 
+## Resource Owner
+
+The `Resource Owner` is the user that a token acts on behalf of.
+
+For [Authorization Code Grant](#authorization-code-grant) tokens, the Resource Owner is the user authorizing the token.
+
+[Client Credentials Grant](#client-credentials-grant) tokens do not have a Resource Owner (i.e. is a guest user), unless they have been granted the {{ ApidocRouteHelper::scopeBadge('bot') }} scope. The Resource Owner of tokens with the {{ ApidocRouteHelper::scopeBadge('bot') }} scope is the owner of the OAuth Application that was granted the token. Currently, only [Chat Bot]({$wikiUrl})s are allowed to request the {{ ApidocRouteHelper::scopeBadge('bot') }} scope.
```
nekodex, line 260: that wiki url is broken

nekodex (approved):

PR #6812, opened 2020-10-27
Support composer 2

notbakaneko:
This should be the minimum required to support composer 2 🤔 

- `sentry-laravel` needs to be updated because its dependencies need to be updated to support composer 2.
- `platform` set to php 7.3 so `ocramius/package-versions` doesn't get installed (requires 7.4); `composer/package-versions-deprecated` is used instead; `ocramius/package-versions` is required by `sentry/sentry`.
- `laravel/framework` updated to ~~latest~~ 6.18.26 to support updated `composer.lock` format without the cookie change. Notification server will need to be updated to handle the new cookie ppy/osu-notification-server#32


notbakaneko:
Looks like the format of the cookie value has changed which causes problems for the notification server, in addition to invalidating all existing cookies

nekodex (approved):
Everything seems to be in order when testing locally, may have to double check if we need to update the sentry DSN to the new-format one after deploying though

PR #6814, opened 2020-10-27
Return json-formatted 404 for json requests on invalid routes

nanaya:
Seems more logical returning json if the request explicitly requested json format even on invalid routes...? :thinking: 

Now I looked at the response not sure if empty array (`[]`) makes sense though as opposed to maybe empty object (`{}`).

### DIFF
--- a/app/Http/Controllers/FallbackController.php
+++ b/app/Http/Controllers/FallbackController.php
@@ -22,7 +22,7 @@ class FallbackController extends Controller
 
     public function index()
     {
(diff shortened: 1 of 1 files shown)


notbakaneko:
`[]` is just the easiest way to make sure the `Content-Type` of the response is `application/json`, it should probably be something like `return response()->json(null, 404)` instead :thinking:

notbakaneko (approved):

PR #6815, opened 2020-10-27
OAuth Token Revocation

### LINKED ISSUE #6784: Enable oauth revocation endpoint
Apparently the builtin endpoint is currently not enabled.

The endpoint seems to not revoke refresh tokens though? And the later version added different endpoint which actually revoke the refresh token as well :question:

### LINKED ISSUE #6785: Broadcast oauth revocation event to websocket
Revoking oauth tokens should be broadcasted and handled similarly to logout event so websocket connections can disconnect properly for oauth authenticated users.

nanaya:
Revoking by specifying access/refresh token will be slightly more involved.

Added here:
- endpoint to revoke current access token (and its refresh token)
- publish revocation event to notification (prefixed with `oauth:` which is already handled by notification server)
- also revoke tokens when removing user sessions through interop endpoint

Resolves #6785, resolves #6784.


notbakaneko (changes requested):
delete token endpoint won't work for tokens with regular (non-`*`) scopes?

Probably should have tests for
- asserting token requested to be revoked can no longer be used (checking if making request using token fails or some other combination of tests)
- logout event/payload is broadcasted since it's used by a separate service

notbakaneko (approved):

PR #6825, opened 2020-10-29
Ensure getting enough beatmap scores for listing

### LINKED ISSUE #6764: Beatmap leaderboard is cut off well before 50 scores
On this beatmap: https://osu.ppy.sh/beatmapsets/639281#osu/1417193
The leaderboard stops at 13 scores, when there are supposed to be way more.
adrenaline leaderboard bug
Checking in-game, there are around 3k scores.
screenshot3543
This issue also happens on the country and friend leaderboards.

nanaya:
Resolves #6764.


app/Models/Score/Best/Model.php
```diff
@@ -68,25 +68,34 @@ public function macroForListing()
...
+
+                foreach ($baseResult as $entry) {
+                    if (isset($result[$entry->user_id])) {
+                        continue;
+                    }
+
+                    $result[$entry->user_id] = $entry;
+
+                    if (count($result) >= $limit) {
+                        break 2;
```
peppy, line 93: this is... not syntax i can get behind (do we use this elsewhere?). maybe worth splitting out into its own function and using `return` instead?

peppy (changes requested):

notbakaneko (approved):

PR #6826, opened 2020-10-29
Add API rate limiting and scores endpoint to API

### LINKED ISSUE #6781: Make api/v2/scores/ available via public scope API
I'm looking to get, for a given score id, both the score metadata as well as the replay data via the api. I can do the latter over the old api, but the former will likely [not be implemented there](https://github.com/ppy/osu-api/issues/304). 
The api endpoint [`api/v2/scores/{mode}/{score}/download`](https://osu.ppy.sh/docs/index.html#apiv2scoresmodescoredownload) looks very promising, but using it with `public` scope results in the error `Invalid scope(s) provided.` for me, similar to #6055, so I assume it is currently not accessible with any scope.

notbakaneko:
- rate limit applied per OAuth token, which means getting a new token obviously resets the limit, so there should probably be a limit on that too :thinking:
- configurable by changing `API_THROTTLE_GLOBAL` env.
- defaults to 1200 requests/min; it is not a rolling window (unlike the rate limit for chat).
- replay downloads default to 10/min `API_THROTTLE_SCORES_DOWNLOAD`

and since there are no skippable `before_action` filters and skipping middleware can't be applied at the controller level, I opted not to exclude the 404 fallbacks from throttling (needs the controller to be excluded specified directly in the middleware)

Rate bursts can be supported by stacking multiple middlewares with different settings, but laravel will only apply the response header of the first middleware run; this issue can be seen when hitting the download rate limit - it'll return a `429` code but still show the limit of the first middleware in the headers.

This doesn't seem like a good way to define multiple throttle rules, honestly 🤔 

closes #6781


.env.example
```diff
@@ -1,3 +1,5 @@
+# API_THROTTLE=60,1,api
```
peppy, line 1: this can be defined by any number of route prefixes?

notbakaneko: what do you mean?

peppy: just interested in how this spec works. might be worth linking the laravel doc for it?

ie. can we specify multiple throttling rules in a single line>

notbakaneko: No, the arguments are `rate limit`, `reset interval`, `key prefix`. The prefix is there because if there is an existing matching timer key in redis, it will use the expiry on that one. Multiple rules on the same line will need a different syntax.

peppy: The use case here is specifically for the replay retrieval endpoint. Is support for that planned to be added at a later date? Might be worth considering it before merging this.

app/Http/Controllers/Controller.php
```diff
@@ -23,6 +23,9 @@ abstract class Controller extends BaseController
 
     public function __construct()
     {
+        if (is_api_request()) {
+            $this->middleware('throttle:'.config('osu.api.throttle.global').':');
```
nanaya, line 27: is there something with setting this here instead of in the route file itself?

nanaya (changes requested):

nanaya (approved):

PR #6827, opened 2020-10-29
Add scores endpoint to API

### LINKED ISSUE #6781: Make api/v2/scores/ available via public scope API

notbakaneko:
closes #6781
- [ ] requires #6826 (for rate limiting)


notbakaneko:
combined with #6826

PR #6838, opened 2020-10-31
Add ability to see user's comments

### LINKED ISSUE #4112: Ability to see a list of comments made by user on maps
With Disqus you were able to see the list of comments you have made on beatmaps, but that is now gone. So I suggest adding it as a link in the user's profile page in a similar way you can see the list of forum posts a user has made. Something like "Contributed ### forum posts and ### map comments":

venix12:
Resolves #4112


resources/assets/coffee/react/profile-page/links.coffee
```diff
@@ -50,6 +50,13 @@ export class Links extends React.PureComponent
 
 
   textMapping =
+    comments_count: (val, user) ->
+      count = osu.transChoice 'users.show.comments_count.count', val
+      url = "#{laroute.route('comments.index')}?author_id=#{user.id}"
```
notbakaneko, line 55: Use `route`'s second parameter to handle the query string

notbakaneko (changes requested):

notbakaneko (changes requested):
I forgot to mention it before, it should show the username somewhere as a title (and also in the breadcrumbs?) when viewing a single user's comments.

notbakaneko:
Waiting for feedback from @arflyte since there's already existing pages to show a user's forum posts and modding posts with form inputs

venix12:
For what it's worth, I based the design on the current single comment one

app/Transformers/UserTransformer.php
```diff
@@ -30,6 +30,7 @@ public function transform(User $user)
         $profileCustomization = $this->userProfileCustomization($user);
 
         return array_merge($result, [
+            'comments_count' => $user->comments->count(),
```
notbakaneko, line 33: this should be `$user->comments()->count()` otherwise it's just fetching all the comments

notbakaneko (changes requested):
Looking at this again, I think maybe it's better if the breadcrumbs was `listing > username > comments` and username was a link back to the user's profile page, instead of the title link? There are separators between top level comments in the listing but there isn't one for the title; on the pages with a title, they're all single comments and their following replies.

venix12:
you just mean to remove the title and change breadcrumbs to ``listing > username > comments``, right?

notbakaneko:
yes

peppy:
I'd say go ahead with the current design since it looks okay and we don't have anything better yet. Agree with the proposed breadcrumb changes.

app/Libraries/CommentBundle.php
```diff
@@ -140,6 +144,11 @@ public function commentsQuery()
     public function countForPaginator()
     {
         $query = $this->commentsQuery();
+
+        if ($this->params->userId !== null) {
+            $query->where('user_id', $this->params->userId);
+        }
```
notbakaneko, line 150: this should just go in `commentsQuery`

app/Libraries/CommentBundle.php
```diff
@@ -167,6 +176,10 @@ private function getComments($query, $isChildren = true, $pinnedOnly = false)
 
         $query->with('commentable')->cursorSort($sort, $cursor ?? null);
 
+        if ($this->params->userId !== null) {
+            $query->where('user_id', $this->params->userId);
+        }
```
notbakaneko, line 181: same as above

database/migrations/2020_11_04_144136_add_index_to_user_id_on_comments.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('comments', function (Blueprint $table) {
+            $table->index('user_id');
```
notbakaneko, line 20: This index should be `user_id, created_at desc, id desc` because of the listing order (see `2020_09_09_042555_update_comments_index`).

notbakaneko (changes requested):

database/migrations/2020_11_04_144136_add_index_to_user_id_on_comments.php
```diff
@@ -17,7 +17,7 @@ class AddIndexToUserIdOnComments extends Migration
     public function up()
     {
         Schema::table('comments', function (Blueprint $table) {
-            $table->index('user_id');
+            $table->index(['user_id', DB::raw('created_at DESC'), DB::raw('id DESC')], 'created_at_desc');
```
notbakaneko, line 20: There's already an existing index with this name

notbakaneko (changes requested):

database/migrations/2020_11_04_144136_add_index_to_user_id_on_comments.php
notbakaneko: name it `user_comments` or something

database/migrations/2020_11_04_144136_add_index_to_user_id_on_comments.php
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('comments', function (Blueprint $table) {
+            $table->index([DB::raw('created_at DESC'), DB::raw('id DESC')], 'created_at_desc');
```
notbakaneko, line 32: also this should be `$table->dropIndex(name of index);`

database/migrations/2020_11_04_144136_add_index_to_user_id_on_comments.php
venix12: oh right, I got you wrong

notbakaneko (approved):

PR #6846, opened 2020-11-04
Increase opaqueness of user-group badges to aid legibility

nekodex:
Decreases the transparency of the group badges in general, more so on probationary ones as they can get near impossible to read in some cases, e.g:


peppy:
Did you mean "Increases"?

nekodex:
Uh yeah, increases. (I think my brain wanted to type 'decreases transparency')

Screen Shot 2020-11-04 at 2 53 03 pm

Almost wondering if we should bump the opacity of the user-card overlays?

eenpersoon64:
think its worth mentioning here that the alumni tag for people that are alumni + probation probably shouldnt be transparent on their userpage


this seems to be the case in the search box

nanaya:
That one is a different bug

resources/assets/less/bem/user-group-badge.less
```diff
@@ -5,7 +5,7 @@
   @_top: user-group-badge;
   border-radius: 20px;
   font-size: 10px;
-  background-color: osu-hsla(@osu-colour-b6, 0.75);
+  background-color: osu-hsla(@osu-colour-b6, 0.9);
```
notbakaneko, line 8: I don't think the opacity of the background needs to be adjusted if the `probationary` modifier is adjusting the whole opacity?

notbakaneko (approved):

PR #6853, opened 2020-11-05
Correctly handle user group's pending state

### LINKED ISSUE #6852: Newly registered users are being added to REGISTERED group with pending status
`user_pending` should be set to 0 when a user is added to a group. Doesn't really affect osu-web systems, but can cause weirdness like [seen here](https://github.com/ppy/osu-web/issues/6851).

I've applied the fix retroactively and will do so once more once this has been fixed in the registration process.

nanaya:
Unifies user group handling and set pending to false by default. Additionally fix group adder/remover previously checking user token.

Resolves #6852.


peppy (dismissed):

app/Models/User.php
```diff
@@ -909,14 +924,29 @@ public function defaultGroup()
...
+                $ret[$key][] = $userGroup->group_id;
+            }
+
+            return $ret;
         });
     }
 
     // check if a user is in a specific group, by ID
-    public function isGroup($group)
+    public function isGroup($group, $checkToken = true)
```
notbakaneko, line 942: I'd name the param to something else that more accurately reflects whether the function is being used as a permission check (in no groups if using oauth), or as an existence check. 
`isGroupPending` probably doesn't need the same options since it wouldn't make sense to do a permission check on a pending group? :thinking:

notbakaneko (changes requested):

app/Models/User.php
nanaya: can use some function name suggestion :thinking:

peppy:
Reminder to self that I need to re-run a batch query again after this is merged.

app/Models/User.php
notbakaneko: `$forPermission` or `checkPermissions` or something? Or just have a separate `hasGroupPermission` function instead 🤔

notbakaneko (approved):

PR #6854, opened 2020-11-05
Add API endpoint to fetch users by id

### LINKED ISSUE #6842: Add ability to query details for multiple users in a single API request
The osu! client needs to be able to quickly retrieve details for a subset of users that are to be displayed on screen (via a list of `user_id`s). This is relatively high priority as we need to use it for the spectator UI.

nanaya:
Just a simple `/users` index requiring `ids` parameter for it to return anything.

Due to how array parameter is passed, the query string may get a bit long...

Resolves #6842.


peppy:
For sanity I'd probably set an upper limit of 32 or 50 or something around that region.

nekodex (approved):
Even with 50 lots of `&ids[]=xxxxxx`, it's unlikely that we'll hit the 2000 char upper bound recommendation for query strings anyway

PR #6860, opened 2020-11-09
Rollback update of laravel and other packages

nanaya:
Something with it made chat system slower.

Also downgraded explicit jwt requirement for minimal change to `composer.lock` file (compared to before the update).


peppy (approved):

notbakaneko (approved):
Travis, stop timing out pls.

PR #6861, opened 2020-11-09
Speed up marking notifications read

nanaya:
The test I did previously was either incomplete or I read something wrong... or maybe because it's actually fixed by mysql 8.0.22 and that's what's I'm using in my dev.


nanaya:
That was from a wrong repo :frowning:

PR #6866, opened 2020-11-10
Add job to delete user notifications

nanaya:
The question is if

- it's worth handling things gracefully (this PR, the easy way)
  - cleaning up 200M+ rows will take a while
- or do a simple rows delete and figure out how to signal the websocket client to fetch new state and reconcile the data
  - alternatively, dispose current state and start fresh
  - this will involve either:
    - signaling everyone currently connected
    - or fetch `user_id` involved in deletion beforehand

Or both, simple rows delete for now (and break clients) and do things gracefully afterwards? If we set the threshold for initial delete large enough it should only affect minimal number of people (if at all).

This still doesn't do anything to the larger `notifications` table. Although once the `user_notifications` table gets slimmer cleaning up the larger one will be just simple row delete.


peppy:
Already mentioned this multiple times, but the burden of doing large deletes like this on database (and replication) load is quite high, which is why we don't do it anywhere these days. What's wrong with the plan to add partitioning to the tables? Is there a reason not to do it this way? Last we discussed I thought we agreed on that direction.

Trying to avoid needless effort and future issues if/when load increases. Partitions are tried and proven.

notbakaneko:
What's the partitioning scheme going to be?

peppy:
time-based, using `created_at` (see `update_partitions` in `osu-web-10` or `describe table` for any table which already uses this, such as `osu_scores`, `osu_games`, `messages` etc.)

nanaya:
The deletes are done in small batches based on primary key so it shouldn't take up much resources to do.

Doing partition drop will require front end to either figure out the difference before and after the deletes or reset the state. Considering how much people will be actually affected by this I'm not sure it's worth the effort adding specific support for this - the main breakages from what I can tell is just incorrect unread count (and potentially showing deleted notifications but those shouldn't be a problem?).

Actually if we don't mind showing incorrect unread count for some people (who weirdly enough keep notifications unread for long time) we can just do the partition way.

nanaya:
maybe just try running this and see how it goes until the table is partitioned? I've also added option to disable it by setting the max delete per run to 0. @peppy

notbakaneko (approved):
it seems to work 🤔

notbakaneko (approved):

PR #6880, opened 2020-11-17
Include locale in wiki path

### LINKED ISSUE #6817: Wiki articles links should use the locale of the current article
It seems weird that following clicking through a link of an article in one language ends up reverting the language choice.

nanaya:
Resolves #6817.

Changes:
- `/help` part is removed because it's a bit too long otherwise
- images now go under its own path (`/wiki/images/`)
- locale is part of the path, after `/wiki`, before page name (`/wiki/en/Welcome`)
- when showing fallback translation, locale in url and selector button shows the requested locale
  - link to english is shown in the menu
- all links to another wiki page inside the page are relative path so going to another page will keep trying requested locale
  - for example if `/wiki/id/Beatmap` doesn't exist, it'll show the page in English but url stays `/wiki/id/Beatmap` and all wiki links inside still point to `/wiki/id/Page_Name`
  - locale menu will show `id` as current locale, and `en` is listed in the menu
- all pages under `Legal/` are always shown under `/legal/<locale>/Page_Name`
  - accessing under `/wiki` will redirect
  - generating such link using `wiki_url` will still generate url under `/legal/`
  - `index` link in header is removed, `Legal` ("listing") page is the top link instead
- sitemap is now under locale path as well `/wiki/<locale>/Sitemap`, complete with locale menu


nanaya:
Whoever reviewing this should tell me if they want this large PR to be split or not :eyes:

notbakaneko (changes requested):
Going to something like `/api/v2/wiki/en/images/shared/news/banners/monthly-beatmapping-contest.png`
returns a redirect to... `/wiki/images/images/shared/news/banners/monthly-beatmapping-contest.png`

nanaya:
I'm not sure if there's any point having images under api endpoint.

app/Http/Controllers/WikiController.php
```diff
@@ -27,26 +28,56 @@ class WikiController extends Controller
...
-            return ujs_redirect(wiki_url());
+        // redirect to default page if missing all parameters
+        if ($locale === null) {
+            return ujs_redirect(wiki_url(null, $this->locale()));
         }
 
+        // if images slip through the markdown processing, redirect them to the correct place
         if (OsuWiki::isImage($path)) {
-            return $this->showImage($path);
+            return ujs_redirect(route('wiki.image', concat_path([$locale === 'images' ? null : $locale, $path])));
```
notbakaneko, line 44: Shouldn't this also skip adding the locale to the path for images if it's a valid locale?

notbakaneko (approved):

PR #6885, opened 2020-11-19
Add button for reporting forum post

### LINKED ISSUE #4597: Add report button to forum posts
Would be useful to have a report button option for forum posts to either report really old bumps or just a keyboard spam message or something offensive.
Example:
report button 2_000000
report extra spicy_000001

nanaya:
Resolves #4597.


peppy:
Would be great if we could get this one in as a priority (allows updating recently published reporting instructions which are... quite cumbersome).

notbakaneko (approved):
seems working

PR #6892, opened 2020-11-20
Add link for 'report abuse' to help menu

### LINKED ISSUE #6890: Reporting Abuse article should be displayed under 'help' header
[A highly upvoted comment](https://osu.ppy.sh/comments/1320799) from the recent abuse-related newspost made by peppy makes a good point - we should probably include a link to the abuse reporting article under the help header for extra visibility.

```https://osu.ppy.sh/help/wiki/Reporting_Bad_Behaviour/Abuse```

### PR DESCRIPTION
**Pull request by nekodex**
resolves #6890

### DIFF
--- a/resources/lang/en/layout.php
+++ b/resources/lang/en/layout.php
@@ -77,9 +77,9 @@ return [
         ],
         'help' => [
             '_' => 'help',
+            'getAbuse' => 'report abuse',
             'getFaq' => 'faq',
             'getRules' => 'rules',
-            'getAbuse' => 'report abuse',
             'getSupport' => 'no, really, i need help!',
             'getWiki' => 'wiki',
         ],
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by peppy — APPROVED** [review]

File: resources/lang/en/layout.php [hunk 1]
```diff
@@ -79,6 +79,7 @@
             '_' => 'help',
             'getFaq' => 'faq',
             'getRules' => 'rules',
+            'getAbuse' => 'report abuse',
```
nanaya, line 82: Translation keys are sorted. Also not sure about calling it `getAbuse` :thinking: (and I'm not sure why this entire section is still using `get` prefix either)

nanaya (changes requested):

peppy (approved):

PR #6903, opened 2020-11-24
Fix not being able to delete grouped notifications

### LINKED ISSUE #6900: Wrong parameter passed for batch notification destroy
Recent update broke deleting notifications by identity as the frontend for deleting is still sending identities in older format.

Sentry Issue: [OSU-WEB-7SX](https://sentry.io/organizations/ppy/issues/2043495116/?referrer=github_integration)

notbakaneko:
closes #6900

### DIFF
--- a/app/Libraries/Notification/BatchIdentities.php
+++ b/app/Libraries/Notification/BatchIdentities.php
@@ -44,13 +44,20 @@ class BatchIdentities
         return $obj;
     }
 
-    public static function scrubIdentities(array $params)
+    public static function scrubIdentities($params)
     {
-        $identities = array_map(function ($param) {
-            return is_array($param) ? static::scrubIdentity($param) : null;
-        }, $params);
+        if (!is_array($params)) {
+            return [];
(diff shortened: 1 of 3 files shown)


nanaya:
May want to fix the backend error as well when invalid parameter is passed.

app/Libraries/Notification/BatchIdentities.php
```diff
@@ -46,6 +44,15 @@ public static function fromParams(array $params)
         return $obj;
     }
 
+    public static function scrubIdentities(array $params)
+    {
+        $identities = array_map(function ($param) {
+            return is_array($param) ? static::scrubIdentity($param) : null;
+        }, $params);
+
+        return array_values(array_filter($identities));
```
nanaya, line 53: maybe foreach and only add to array if not null instead of this map/filter/values chain?

nanaya (changes requested):

nanaya (approved):

PR #6909, opened 2020-11-27
Add per-user new beatmap notification

### LINKED ISSUE #5635: Subscribing to mapper
There are many mappers that make great maps. Wouldn't it be great if you could be notified when they submit a new map?

I imagine there would be a page where you can manage your subscription for each mapper. You would be able to filter subscriptions by gamemode, star rating, CS (or keymode), OD, AR, and map length so you could receive notification for the types of maps you want. I don't know if filters should be global or per subscription.

nanaya:
No listing page yet ( ﾟ ヮﾟ)

Actual event broadcast must be done by whatever handling beatmapset upload.

Also it currently comes up in notification popup under "profile" due to being grouped by user.

Resolves #5635. No filtering though.

Endpoint for broadcasting new map:

```
POST /_lio/beatmapsets/:id/broadcast-new
```


resources/assets/lib/follow-user-modding-button.tsx
```diff
@@ -0,0 +1,118 @@
...
+  follow: boolean;
+  loading: boolean;
+}
+
+const bn = 'user-action-button';
+
+export default class FollowUserModdingButton extends React.Component<Props, State> {
+  private buttonRef = React.createRef<HTMLButtonElement>();
+  private eventId = `follow-user-modding-button-${osu.uuid()}`;
+  private xhr: JQueryXHR | null = null;
```
notbakaneko, line 25: just `private xhr?: JQueryXHR`? it's never unset

resources/assets/lib/interfaces/current-user.ts
```diff
@@ -8,6 +8,7 @@ import GameMode from './game-mode';
 
 export default interface CurrentUser extends UserJsonExtended {
   blocks: UserRelationJson[];
+  follow_user_modding: number[];
```
notbakaneko, line 11: I wonder if nested object would be more useful for this, but then again we don't know if more `follow` ids would go in there 🤔

notbakaneko (changes requested):

notbakaneko (approved):

PR #6917, opened 2020-12-01
fix grammar in testflight page

tybug:


Joehuu (approved):

peppy (approved):

peppy:
Should really be localisable but this will do for now.

PR #6925, opened 2020-12-02
Fix front-end part of full BN nomination check

### LINKED ISSUE #6922: Probationary BN cannot nominate a map nominated by a full BN
Jerry (probationary taiko bn) apparently can't nominate https://osu.ppy.sh/beatmapsets/1121548/discussion, which is currently nominated by frukoyurdakul (full taiko bn). Probably related to the new nomination restrictions from https://github.com/ppy/osu-web/pull/6832.

nanaya:
Resolves #6922.

### DIFF
--- a/resources/assets/lib/beatmapsets/beatmapset-json.ts
+++ b/resources/assets/lib/beatmapsets/beatmapset-json.ts
@@ -14,7 +14,7 @@ interface BeatmapsetCovers {
 }
 
 export function isBeatmapsetNominationEvent(x: BeatmapsetEvent): x is BeatmapsetNominationEvent {
-  return x.type === 'nominate' && x.comment != null && Array.isArray(x.comment.modes);
+  return x.type === 'nominate' && Array.isArray(x.comment?.modes);
 }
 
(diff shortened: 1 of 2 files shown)


resources/assets/lib/beatmapsets/beatmapset-json.ts
```diff
@@ -13,6 +13,17 @@ interface BeatmapsetCovers {
   slimcover: string;
 }
 
+export function isBeatmapsetNominationEvent(x: BeatmapsetEvent): x is BeatmapsetNominationEvent {
+  return x.type === 'nominate' && x.comment != null && Array.isArray(x.comment.modes);
```
nekodex, line 17:
```suggestion
  return x.type === 'nominate' && Array.isArray(x.comment?.modes);
```

nanaya: But that will call `isArray` on `undefined` :thinking:

nekodex: which returns false anyway?

nekodex (approved):

PR #6931, opened 2020-12-03
New multiplayer room type

nanaya:
Add parameter `category` of `realtime` (also for listing). `ends_at` parameter can be skipped.

And then I noticed there's no way to indicate number of people inside the room :thinking: 

Todo:
- [x] test

### DIFF
--- a/app/Http/Controllers/Multiplayer/Rooms/Playlist/ScoresController.php
+++ b/app/Http/Controllers/Multiplayer/Rooms/Playlist/ScoresController.php
@@ -190,7 +190,6 @@ class ScoresController extends BaseController
     public function update($roomId, $playlistId, $scoreId)
     {
         $room = Room::findOrFail($roomId);
-        // todo: check against room's end time, check within window of start_time + beatmap_length + x
 
         $playlistItem = $room->playlist()
             ->where('id', $playlistId)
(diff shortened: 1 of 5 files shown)


smoogipoo:
I think realtime rooms should not be returned if `category` is not present. Since these are technically different systems (at least client side), "any" only applies to playlists:

```
Multiplayer categories:
- realtime

Playlists categories:
- any
- normal
- spotlight
```

peppy:
we should probably just *require* a type to be specified, or make it part of the route?

smoogipoo:
Seems like omitting the `category` parameter doesn't return any rooms. It should return rooms with `category = NULL | spotlight`. The client does not send `category` to retrieve normal rooms, but it can do so (becomes `NULL | normal | spotlight`).

nanaya:
Do you have any active rooms...?

smoogipoo:
Yeah I do:

image

After forcing 'normal' category in the DB:

image

nanaya:
oh huh I messed up the enum defaults

peppy:
@notbakaneko can you do a check over this? would like to get it live for further testing of the full (backend) deployment.

notbakaneko (changes requested):

notbakaneko (approved):

peppy:
Migrations are fine to run via deploy.

PR #6932, opened 2020-12-03
Separate web socket connection handling from notifications

notbakaneko:
Decouples the web socket handling is decoupled from notifications handling.
The goal is to make the web socket worker more of a generic standalone interface for realtime messages that delegates message processing to other components, so that we don't have to cram everything into the websocket handler. 

`SocketWorker` manages the web socket connection and forwards any relevant events, while the notification worker parses and dispatches only notification-related events. The notification worker could be simpler if it didn't need to handle the whole `firstLoadedAt` thing which is there to handle events for notifications that haven't been loaded yet.

Other changes are:
- notification icons always show even if not connected to the web socket yet
- different message in widget if notifications haven't loaded yet, instead of "No notifications"
(it still does the `...` flashing from the component being remounted on navigation)


resources/assets/lib/chat-icon.tsx
```diff
@@ -49,7 +43,7 @@ export default class ChatIcon extends React.Component<Props> {
   }
 
   private unreadCountDisplay() {
-    if (this.props.worker.hasData) {
+    if (core.socketWorker.hasConnectedOnce) {
```
nanaya, line 46: this results in `...` turning into `0` before finally showing the correct number. Maybe it should still consult notification worker after all?

resources/assets/lib/notification-widget/main.tsx
```diff
@@ -187,7 +186,15 @@ export default class Main extends React.Component<Props, State> {
     });
 
     if (nodes.length === 0) {
-      const transKey = this.controller.currentFilter == null ? 'notifications.all_read' : 'notifications.none';
+      let transKey = 'notifications.loading';
+      if (!core.socketWorker.hasConnectedOnce) {
+        transKey = 'notifications.loading';
+      } else if (this.controller.currentFilter == null) {
+        transKey = 'notifications.all_read';
```
nanaya, line 193: similar to the count display, this will be shown when websocket is connected but the initial data is still being downloaded

nanaya (changes requested):

resources/assets/lib/notification-widget/main.tsx
nanaya: this one is still not fixed :sleeping:

nanaya (approved):

Issue #1566, opened 2017-09-23
osu!wiki - add markup for id attributes in headers


MegaApplePi:
Is it possible to add markup for `id` attributes to the headers? This will allow us to consistently section link to any locale of an article.
The syntax looks like this: `## Easy {#ez}`

Other attributes like classes would be ignored.

ref: <https://michelf.ca/projects/php-markdown/extra/#spe-attr>

PR #6934, opened 2020-12-04
Support beatmapset rule violation message

notbakaneko:
Renders a different message and link to the wiki if `download_disabled_url` is set to `rule_violation`

### DIFF
--- a/resources/assets/coffee/react/beatmapset-page/header.coffee
+++ b/resources/assets/coffee/react/beatmapset-page/header.coffee
@@ -190,27 +190,28 @@ export class Header extends React.Component
 
 
   renderAvailabilityInfo: =>
-    if currentUser.id? && @hasAvailabilityInfo()
-      href = if @props.beatmapset.availability.more_information == 'rule_violation'
-               "#{route('wiki.show', locale: currentLocale, path: 'Rules')}#beatmap-submission-rules"
-             else
-               @props.beatmapset.availability.more_information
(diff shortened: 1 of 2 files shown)


resources/assets/coffee/react/beatmapset-page/header.coffee
```diff
@@ -201,6 +189,30 @@ export class Header extends React.Component
             timeElapsed: @props.timeElapsed
 
 
+  renderAvailabilityInfo: =>
+    if currentUser.id? && @hasAvailabilityInfo()
```
nanaya, line 193: inverse condition and return instead for less indented stuff

resources/assets/coffee/react/beatmapset-page/header.coffee
```diff
@@ -201,6 +189,30 @@ export class Header extends React.Component
...
+      href = if @props.beatmapset.availability.more_information == 'rule_violation'
+               "#{route('wiki.show', locale: currentLocale, path: 'Rules')}#beatmap-submission-rules"
```
nanaya, line 195: not quite fixable but the hash only works for english locale until #1566 is fixed :pensive:

nanaya (changes requested):

resources/assets/coffee/react/beatmapset-page/header.coffee
notbakaneko: commonmark still doesn't support named anchors :|

nanaya (approved):

PR #6950, opened 2020-12-07
Fix country name formatting for the State of Palestine

### LINKED ISSUE #6742: Country naming for State of Palestine should be fixed
According to the [International Organisation for Standardization](https://www.iso.org/obp/ui/#iso:code:3166:PS) country name is defined as `Palestine, State of` while the osu! website states `Palestinian Territory Occupied` which goes against the ISO 3166 formatting. This needs to be fixed accordingly.

zeusminus:
closes #6742

Fixes ISO-3166 naming for the State of Palestine. This is also my first PR in this repository, so let me know if I edited the wrong file or if there are other files that need to be addressed regarding this fix.


peppy:
This seems fine, but will also need to be updated in the database to actually change the display on the site itself. I'll attend to that part.

peppy (approved):

peppy:
We inserted credits into travis but travis no provide.

PR #6953, opened 2020-12-08
Combine watchlist pages

nanaya:
"subtype" sure is looking more like "type" now. And my previous `modding` "subtype" for following mapper doesn't make sense anymore unless there's better name for beatmapset discussion/nomination watch ( ﾟ ヮﾟ)a

This will be the base for adding comment and mapper watchlists.


resources/lang/en/follows.php
```diff
@@ -4,6 +4,18 @@
 // See the LICENCE file in the repository root for full licence text.
 
 return [
+    'forum_topic' => [
+        'title' => 'forum topic',
```
notbakaneko, line 8: I think naming these after what their sections are named instead of the code makes more sense.

resources/assets/lib/home-links.ts
```diff
@@ -0,0 +1,29 @@
...
+    },
+    {
+      active: active === 'friends@index',
+      title: osu.trans('friends.title_compact'),
+      url: route('friends.index'),
+    },
+    {
+      active: active === 'follows@index',
+      title: osu.trans('follows.index.title_compact'),
+      url: route('follows.index', { subtype: 'comment' }),
```
notbakaneko, line 21: but there's no comments here yet?_?

resources/assets/lib/home-links.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import { route } from 'laroute';
+
+export default function homeLinks(active: string) {
+  return [
+    {
+      active: active === 'home@index',
```
notbakaneko, line 9: can use route name instead?

notbakaneko (changes requested):

resources/lang/en/follows.php
nanaya: what do you mean?

notbakaneko: topics/beatmap discussions

nanaya: but modding has been called such :thinking: 
also "forum watchlist" sounds more like watchlist for forum instead of topic

notbakaneko (approved):

PR #6955, opened 2020-12-08
Display mapping followers count

### LINKED ISSUE #6942: Display of how many "beatmap subscriptions" your profile has.
With the recent implementation of the mapper subscription feature, the natural thought of where we could see how many people have subscribed to us arose. This could be implemented in the same way the friends button is:
https://i.imgur.com/G0T8YuB.png
However, with the bell icon instead of the person. With the number of subscriptions a person has. However, on your own profile, it would obviously not allow you to subscribe to yourself.

venix12:
Resolves #6942

based on friends follower count workflow
~~not really sure if "modding followers" is the best name :thinking:~~

- [x] change modding -> mapping after #6960 is merged


peppy:
rather than "modding followers", it should probably just be "x subscribed"

venix12:
guess i can just go for "subscribers" instead since the count is already displayed on the button itself

nanaya:
I'm thinking of renaming "modding" to "mapping" because it's not really about following the user's modding activity.

peppy:
Yeah, seems weird that it was named Modder in the first place.

venix12:
Yes, it would make much more sense, guess it's a thing for another pull request though since it would require changing naming all around the code, including changes from #6909 and #6953 (when merged) which is likely out of scope for this one likely.

venix12:
Should be ready for review now, I believe. :eyes:

nanaya (changes requested):

resources/lang/en/follows.php
```diff
@@ -13,6 +13,7 @@
     ],
 
     'mapping' => [
+        'disabled' => 'subscribers',
```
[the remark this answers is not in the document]
venix12: https://github.com/ppy/osu-web/blob/866c4b1c45826ecbb100bf516536765910b5b449/resources/lang/en/friends.php#L10-L14

was basing on this, would likely make sense to change though, yes :thinking:

venix12:
changed ``subscribers`` to ``mapping subscribers`` as well since it would be less confusing i believe

resources/assets/lib/follow-user-mapping-button.tsx
```diff
@@ -43,29 +53,40 @@ export default class FollowUserMappingButton extends React.Component<Props, Stat
   }
 
   render() {
-    if (currentUser.id == null || currentUser.id === this.props.userId) {
+    if (currentUser.id === this.props.userId && !this.props.alwaysVisible) {
```
nanaya, line 56: this incorrectly shows the button when `alwaysVisible` is false and user is guest

resources/assets/lib/follow-user-mapping-button.tsx
```diff
@@ -43,29 +53,40 @@ export default class FollowUserMappingButton extends React.Component<Props, Stat
...
       return null;
     }
 
-    const title = osu.trans(`follows.mapping.${this.state.follow ? 'to_0' : 'to_1'}`);
+    const currentUserProfile = currentUser.id === this.props.userId;
+    const notLoggedIn = currentUser.id == null;
```
nanaya, line 61: maybe just combine an `||` of these as `canToggle` and use it where needed (always visible check above, title and disabled state below)

resources/assets/lib/follow-user-mapping-button.tsx
```diff
@@ -43,29 +53,40 @@ export default class FollowUserMappingButton extends React.Component<Props, Stat
...
+
+    let title = osu.trans(`follows.mapping.${this.state.follow ? 'to_0' : 'to_1'}`);
+    if (currentUserProfile || notLoggedIn) title = osu.trans(`follows.mapping.followers`);
```
nanaya, line 64: can be done at once with ternary operator `const title = someCheck ? osu.trans(...) : ....`

nanaya (changes requested):

nanaya (approved):

PR #6958, opened 2020-12-09
Show spinner when clear all notifications

### LINKED ISSUE #6901: Spinner isn't shown when marking all notifications read
Reproduction steps:

0. have some unread notifications
1. open notification widget
2. clear all

The check mark should turn into spinner but it currently doesn't.

gagahpangeran:
Resolved #6901

spin

Method [markTypeAsRead](https://github.com/ppy/osu-web/blob/0bf85478f31fd4b881344cda6b264dd608f88df8/resources/assets/lib/notifications/notification-controller.ts#L65) was never called because the `name` of `NotificationType` is `null`, so `isMarkingAsRead` was never set to `true`.

### DIFF
--- a/resources/assets/lib/notifications/notification-controller.ts
+++ b/resources/assets/lib/notifications/notification-controller.ts
@@ -63,12 +63,11 @@ export default class NotificationController {
 
   @action
   markCurrentTypeAsRead() {
+    this.type.markTypeAsRead();
     if (this.type.name == null) {
       for (const name of this.typeNamesWithoutNull) {
         this.store.getOrCreateType({ objectType: name }).markTypeAsRead();
       }
-    } else {
-      this.type.markTypeAsRead();
     }
   }
 


notbakaneko (changes requested):
This marks chat read as well, which it should not; only the notification types listed in the widget itself should be marked as read. It also shouldn't be queuing the same type to be marked as read twice.

peppy:
Was this somehow forgotten about? It looks like the changes since last review have been applied.

notbakaneko (changes requested):
- This doesn't address the issue for `notifications-index`
- this type of logic should be handled by `NotificationController` which should be updated accordingly
- CI results should be checked

peppy:
@gagahpangeran is it safe to say you're dropping this issue?

gagahpangeran:
Yes for now. I just clean up my branch.

I'll revisit it later or someone can pick it up.

PR #6959, opened 2020-12-09
Migrate to Github Actions from TravisCI

### LINKED ISSUE #6928: Switch CI from travis to github actions
Travis is now charging out the ass for their service, even for open source usage. We should look to switch to github actions within the next month or two to avoid unnecessarily paying them.

notbakaneko:
Moves the rest of our CI away from Travis.

Github Actions doesn't support yaml anchors or shared steps, making it inconvenient to split the tests into separate jobs to allow them to run in parallel. The slowest part of the build is running the migrations which also happens to be the part where there isnt' currently a good way to cache with the current setup.

closes #6928


nanaya:
conflict :eyes:

.github/workflows/tests.yml
```diff
@@ -0,0 +1,120 @@
...
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Configure sysctl limits
+        run: |
+          sudo swapoff -a
+          sudo sysctl -w vm.swappiness=1
+          sudo sysctl -w fs.file-max=262144
+          sudo sysctl -w vm.max_map_count=262144
```
nanaya, line 92: can use comment on these settings (is it for elasticsearch?)

bin/db_setup.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
 for db in osu osu_store osu_mp osu_chat osu_charts osu_updates; do
-  echo "CREATE DATABASE IF NOT EXISTS ${db} DEFAULT CHARSET utf8mb4" | mysql -u root
-  echo "CREATE DATABASE IF NOT EXISTS ${db}_test DEFAULT CHARSET utf8mb4" | mysql -u root
+  echo "CREATE DATABASE IF NOT EXISTS ${db} DEFAULT CHARSET utf8mb4" | mysql -u root $@
+  echo "CREATE DATABASE IF NOT EXISTS ${db}_test DEFAULT CHARSET utf8mb4" | mysql -u root $@
```
nanaya, line 5: doesn't really affect how it's currently used but it should be quoted for correct usage
```suggestion
  echo "CREATE DATABASE IF NOT EXISTS ${db} DEFAULT CHARSET utf8mb4" | mysql -u root "$@"
  echo "CREATE DATABASE IF NOT EXISTS ${db}_test DEFAULT CHARSET utf8mb4" | mysql -u root "$@"
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,64 @@
...
+        uses: actions/cache@v2
+        id: yarn-cache
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Install Composer dependencies
+        run: composer config -g github-oauth.github.com 98cbc568911ef1e060a3a31623f2c80c1786d5ff && composer install --dev --prefer-dist --no-progress --no-suggest
```
nanaya, line 53: (probably should remove the deprecated composer options as well)

```suggestion
        run: >-
          composer config -g github-oauth.github.com 98cbc568911ef1e060a3a31623f2c80c1786d5ff
          && composer install --dev --prefer-dist --no-progress --no-suggest
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      - name: Install js dependencies
+        run: yarn --frozen-lockfile
+
+      - run: yarn lint
```
nanaya, line 58: maybe use specific github action as well if easy enough :thinking:

nanaya (changes requested):

nanaya (approved):

PR #6964, opened 2020-12-10
Allow starting chat if verification allows minimum plays to be skipped

### LINKED ISSUE #6884: Visiting chat doesn't trigger verification and instead shows a misleading message
Visiting chat with an unverified session doesn't trigger verification and prevents you from being able to send messages...

You get a disabled input box with the generic 'access denied' message (which doesn't mention verification at all)
Screen Shot 2020-11-18 at 9 44 47 pm

However, if you trigger verification manually - i.e. going to account settings, verifying, then returning to chat (and refreshing), the error message disappears and you are able to send messages again.

gagahpangeran:
Edit: Upon investigation, the issue of triggering verification appears to have already been solved by previous changes. The remaining issue is when the minimum plays requirement is allowed to be skipped when verified if the option is enabled. When allowed to be skipped, it should allow input and trigger verification when trying to send the message; when the option to skip is disabled, it should still prevent input.

Closed #6884


peppy:
That test failure is going to require some work (the browser test does not look to be aware of the verification flow).

notbakaneko (changes requested):
It's not explained clearly in the issue but verification for chat is conditional based on `OsuAuthorize::checkChatChannelSend`; verification is triggered when sending the message if required; the issue is, in some scenarios you can't attempt to send anything because the input is disabled - it should prompt you to verify in these cases.

gagahpangeran:
So from my understanding, the scenario is like this.
- User login with unverified session.
- User chat to someone and the input is always enabled except for these reason.
Screen Shot 2020-11-18 at 9 44 47 pm
- User send the chat and then verification popup appear.

Is that right?

So the scenario is like chat in public channel (?)

chat

peppy:
If it simplifies things at all, we could potentially just move verification to the point of entering the chat display? It would mean that private messages are better protected by account verification too (which I believe is not a thing right now?)

gagahpangeran:
The weird thing is I can only reproduce the issue in local. I cant reproduce the issue in website when login in incognito tab, the input is never disabled and the message is not appeared.

Try to chat to someone in unverified session:
- Website
Screenshot from 2020-12-10 13-35-44

- Local
Screenshot from 2020-12-10 13-35-48

Will investigate more.

notbakaneko:
@gagahpangeran yes

notbakaneko:
If verification is going to be required for viewing chat, then it needs to be required for notifications and possibly the websocket as well.

peppy:
@notbakaneko Let's try and keep it to on send for now. That sounds like a lot of pain we should probably consider at a later point in time, if ever.

gagahpangeran:
So after some investigation, the issue can be appeared when user is in unverified session **and the playcount is bellow minimum**. The issue is not appeared when user is in unverified session and has playcount above minimum. That's why I can't reproduce the issue in production website.

In `OsuAuthorize::checkChatStart`, there check for `ensureHasPlayed` and can be bypass if `USER_MIN_PLAYS_ALLOW_VERIFIED_BYPASS` set to `true` in config.

https://github.com/ppy/osu-web/blob/baf4356bf167c411ea8a4a676521bcccf46670e7/app/Libraries/OsuAuthorize.php#L781-L787

When user is in unverified session and never played (the playcount is bellow minimum), `ensureHasPlayed` throws error and set the `AuthorizationResult::can` to false.

https://github.com/ppy/osu-web/blob/baf4356bf167c411ea8a4a676521bcccf46670e7/app/Libraries/OsuAuthorize.php#L1758-L1764

https://github.com/ppy/osu-web/blob/baf4356bf167c411ea8a4a676521bcccf46670e7/app/Libraries/OsuAuthorize.php#L66-L76

Because of that, `$json['can_message']` in `ChatController` is set to false and it is consumed by react component as `channel.moderated` and then the issue appeared.

https://github.com/ppy/osu-web/blob/baf4356bf167c411ea8a4a676521bcccf46670e7/app/Http/Controllers/ChatController.php#L30

https://github.com/ppy/osu-web/blob/baf4356bf167c411ea8a4a676521bcccf46670e7/resources/assets/lib/chat.ts#L30

https://github.com/ppy/osu-web/blob/baf4356bf167c411ea8a4a676521bcccf46670e7/resources/assets/lib/chat/conversation-view.tsx#L243-L245

Can you confirm this? @notbakaneko 

I have two proposed solution:
- Always trigger verification popup when enter chat when user is not in verified session (current PR). I'll change the test if this is approved.
- The chat will never be disabled and the verification popup only appear when user try to send message. I'll change this PR and removed `ensureHasPlayed` in `checkChatStart`.

notbakaneko (dismissed):
It should still show the message and prevent input if the min plays condition is not met. Verification allowing the in plays requirement to be bypassed is a toggle and not always enabled.

nanaya (approved):

PR #6966, opened 2020-12-10
Add watchlist listing for comment and mapper

nanaya:
depends on:
- [x] #6953
- [x] #6960

Will probably need a bit more cleanup once those are merged.


notbakaneko (changes requested):
comment and mapper are missing the watchlist submenu highlight.

(query analyzer output for recent ids also looks fun)

notbakaneko (approved):

PR #6968, opened 2020-12-10
Include required OAuth scopes in apidoc

notbakaneko:
Lists the required scopes for the route based on the middleware declared so we don't have to keep updating it manually.
Also changes `Requires authentication` to mean requiring an OAuth token and also auto-tags it; `@authenticated` no longer has to be added to the apidoc comments, which may not have matched in some cases, anyway.

The new label colours are just arbitrary complimentary colours of the existing `#3a87ad`

- [x] needs some better marking/explanation of routes only available to osu!lazer client


app/Libraries/ApidocRouteHelper.php
```diff
@@ -0,0 +1,62 @@
...
+            $this->routeScopes[$key] = $route;
+        }
+    }
+
+    public function getScopes(string $uri, string $method)
+    {
+        return $this->routeScopes[$this->keyFor($uri, $method)]['scopes'];
+    }
+
+    public function hasScopes(string $uri, string $method)
```
nanaya, line 44: not used?

resources/views/vendor/apidoc/partials/info.blade.php
```diff
@@ -43,6 +43,8 @@
 
 # Authentication
 
+Routes marked with this label <span class="authenticated">Requires authentication</span> require authentication.
```
nanaya, line 46: There's still confusion whether "authentication" here means actual user (authorization grant etc) or just client grant is enough.

app/Libraries/ApidocRouteHelper.php
```diff
@@ -0,0 +1,62 @@
...
+        foreach ($routeScopesHelper->toArray() as $route) {
+            $key = $this->keyFor($route['uri'], $route['method']);
+            $route['scopes'] = array_filter($route['scopes'], function ($scope) {
+                return $scope !== 'any';
+            });
+            $this->routeScopes[$key] = $route;
+        }
+    }
+
+    public function getScopes(string $uri, string $method)
```
nanaya, line 39: maybe add "lazer only" label when empty? Will need to differentiate with `any` though.

nanaya (changes requested):

resources/views/vendor/apidoc/partials/route.blade.php
```diff
@@ -45,16 +50,21 @@
...
-<small class="authenticated">Requires authentication</small>
-@endif
-
 {!! $topDescription !!}
 
 ### HTTP Request
 @foreach($route['methods'] as $method)
 `{{$method}} {{ $displayUri }}`
 
+<span class="scope-list">
```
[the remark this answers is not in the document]
notbakaneko: routes can have different scopes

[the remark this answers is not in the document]
nanaya: do you mean different method for same route can have different scopes...?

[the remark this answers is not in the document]
nanaya: "different" as in "GET"/"HEAD" pair and "PATCH"/"PUT"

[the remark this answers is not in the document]
notbakaneko: why is that even a thing 🗑️

app/Libraries/ApidocRouteHelper.php
```diff
@@ -0,0 +1,78 @@
...
+    public function requiresAuthentication(array $route)
+    {
+        if (
+            in_array('GET', $route['methods'], true)
+            && starts_with("{$route['uri']}/", RequireScopes::NO_TOKEN_REQUIRED)
+        ) {
+            return false;
+        }
+
+        return in_array('require-scopes', $this->routeScopes[static::keyFor($route)]['middlewares'], true);
```
nanaya, line 76: It looks like `api` routes are all wrapped inside `require-scopes` so this will always return true?

also doesn't `$route` already contain `$route['middlewares']`? :thinking:  This function could even be a static function.

app/Libraries/ApidocRouteHelper.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public function getScopeTags(array $route)
+    {
+        if (!$this->requiresAuthentication($route)) {
+            return [];
+        }
+
+        return $this->routeScopes[static::keyFor($route)]['scopes'];
+    }
```
nanaya, line 65: can't this be done as part of the loop in constructor?
```
if ($this->requiresAuthentication($route)) {
  // filter scopes and add oauth etc
} else {
  $route['scopes'] = [];
}
```

nanaya (changes requested):

nanaya (approved):

PR #6969, opened 2020-12-10
Show 'no conversations yet' when close all chat channel

gagahpangeran:
Found this issue when working on another PR

Before:

emptybefore

After:

emptyafter


resources/assets/lib/stores/channel-store.ts
```diff
@@ -230,6 +230,10 @@ export default class ChannelStore {
     }
 
     this.channels.delete(channelId);
+
+    if (this.channels.size <= 0) {
+      this.loaded = false;
+    }
```
nanaya, line 236: The fix should be done by fixing `MainView` to check channel size instead.

nanaya (changes requested):
Should

nanaya:
Additionally, the initial presence loader in `lib/chat.ts` should check if it's an array instead of empty because otherwise the load state would be wrong (channel store with empty presence/channel should still be marked as `loaded`... I think? @notbakaneko ).

https://github.com/ppy/osu-web/blob/ac2a44f6f13415a9a8d5db92c50297085734ede0/resources/assets/lib/chat.ts#L13

resources/assets/lib/stores/channel-store.ts
gagahpangeran: So this should check the channel size instead of loaded?

https://github.com/ppy/osu-web/blob/ac2a44f6f13415a9a8d5db92c50297085734ede0/resources/assets/lib/chat/main-view.tsx#L37

nanaya: yes

nanaya (approved):

PR #6977, opened 2020-12-14
Always use current time for room starts_at

nanaya:
Never sent by the client. Maybe.


peppy (approved):

notbakaneko:
Doesn't this make it not possible to prepare rooms for future plays?

peppy:
Yep, that's out of scope for now. We can add it back if/when there's a requirement for that (the client doesn't currently allow choosing the start time, nor does it send it, so this is just a consistency check).

PR #6982, opened 2020-12-14
Persistent notification widgets and icons

nanaya:
No more `...` count on simple navigation. Also reworked react component load state management.


notbakaneko (approved):
no more flashy `...`

PR #6990, opened 2020-12-16
Cleanup beatmapset discussion related routes

notbakaneko:
Moving the modding discussions urls to be under `beatmapset/discussions` to be inline with our naming scheme.
`/beatmapsets/beatmap-discussions/...` -> `/beatmapsets/discussions/...`
`/beatmapsets/beatmap-discussion-posts/...` -> `/beatmapsets/discussions/posts/...`

`votes` index was already using the correct url which actually made it not obvious the url already exists.
Requests to old urls will be redirected to the new one; support for `307` redirects have been added which should tell the browser to preserve the HTTP method when redirecting. I'm not sure how search engines treat `307`s, otherwise, it'd be easier to default to `307` 🤔 .

Upcoming API routes for discussions will be added according to new names under `/api/v2/`.

(I didn't rename the controllers yet because that's another pile of fun)


nanaya (approved):
Seems fine :eyes: (famous last word)

PR #6993, opened 2020-12-16
Reduce data queried for mobile navbar usercard

notbakaneko:
User card doesn't need all the extra data from `User::defaultJson()`.
I thought it was weird the data for `currentUser` kept getting queried twice 🤔

### DIFF
--- a/resources/views/layout/header_mobile/user.blade.php
+++ b/resources/views/layout/header_mobile/user.blade.php
@@ -6,7 +6,7 @@
     @if (Auth::check())
         <div
             class="navbar-mobile-item__main js-react--user-card"
-            data-user="{{ json_encode(json_item(auth()->user(), 'UserCompact', ['cover', 'country', 'groups'])) }}"
+            data-is-current-user="1"
         ></div>
 
         <a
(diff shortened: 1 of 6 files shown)


resources/views/layout/header_mobile/user.blade.php
```diff
@@ -6,7 +6,7 @@
     @if (Auth::check())
         <div
             class="navbar-mobile-item__main js-react--user-card"
-            data-user="{{ json_encode(Auth::user()->defaultJson()) }}"
+            data-user="{{ json_encode(json_item(auth()->user(), 'UserCompact', ['cover', 'country', 'groups'])) }}"
```
nanaya, line 9: how about add mount option like `data-user-current="1"` and then check that in props function in `register-components` :thinking: 

Also I noticed the one for `user-card-store` doesn't even need to pass the user at all (also it's missing group lol).

<sub>and then there's unused `objects._usercard` view...</sub>

nanaya (changes requested):

nanaya (approved):

PR #7001, opened 2020-12-17
Add better error message when change profile for silenced user

### LINKED ISSUE #6944: Trying to change profile when silenced is really bad UX

gagahpangeran:
related: #6944


nanaya (approved):
The better message helps but doesn't fix the main problem of bad ui in general when silenced.

PR #7003, opened 2020-12-17
Add alert and disable forms in account edit when user is silenced

### LINKED ISSUE #6944: Trying to change profile when silenced is really bad UX

venix12:
resolves #6944

the alert message could possibly be more expanded


nanaya (changes requested):

resources/views/forum/_post_toolbar.blade.php
```diff
@@ -2,7 +2,12 @@
     Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
     See the LICENCE file in the repository root for full licence text.
 --}}
-<div class="post-box-toolbar">
+<div
+    class="post-box-toolbar"
+    @if (isset($disabled) && $disabled === true)
+        disabled
```
nanaya, line 8: `div` doesn't have `disabled` attribute. This should go either as class name (as modifier perhaps) or data attribute.

nanaya (changes requested):

nanaya (approved):

PR #7009, opened 2020-12-19
Show correct error message when changing username to different casing

### LINKED ISSUE #7005: Wrong displayed message in the rename page when...
When I type my name correctly in this page: https://osu.ppy.sh/store/products/32 , where we have to type a name, I tried to put my own name (Corne2Plum3):
bug2
So there is a special message who say this is already my username.

But if I change some uppercase and lowercase, the message "This name is already in use" is displayed instead of the message above.
bug1
I'm sure there is no other user called like this because when I searched my own username in the player search, I have only 1 result.

_Don't ask me WHY I did this._

_Also, I don't want to change my username, I did this just for fun._

austin-leung:
Fixes #7005

Displays same username error message rather than username is in use. We had trouble trying this out in the local web environment, but the php unit tests reflect that it works.

Co-authored by @angelaz1

### DIFF
--- a/app/Libraries/ChangeUsername.php
+++ b/app/Libraries/ChangeUsername.php
@@ -53,7 +53,7 @@ class ChangeUsername
             return $this->validationErrors()->addTranslated('username', static::requireSupportedMessage());
         }
 
-        if ($this->username === $this->user->username) {
+        if (User::cleanUsername($this->username) === $this->user->username_clean) {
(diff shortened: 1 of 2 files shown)


PercyDan54:
A suggestion: don't create PRs from `master` branch, create a new branch for this

notbakaneko (approved):

PR #7013, opened 2020-12-21
Enforce ends_at for some multiplayer rooms

nanaya:
To prevent orphan room when the user is disconnected after creating the room but before connecting to the multiplayer websocket server.


notbakaneko (commented):
these rooms will show up in the `ended` rooms listing, is that fine?

nanaya:
The websocket server has been modifying the column directly I think so there's no change in that regard.

notbakaneko (approved):

PR #7017, opened 2020-12-22
Select channel updates in chunks

notbakaneko:
It seems to take less memory but it's strange that it's hitting the memory limit before even selecting any large models?


notbakaneko:
probably should leave this as a last resort 🤔

Issue #6945, opened 2020-12-07
Artisan fails to create a lazer oauth client


smoogipoo:
Not sure what the proper name of this client is so please rename the issue as appropriate. It's the client that's not assigned a user id and used for lazer authentication. Fails with:

```
[smgi@smgi osu-web]$ docker-compose run --rm php artisan passport:client --password
Creating osu-web_php_run ... done

 What should we name the password grant client? [osu!web Password Grant Client]:
 >


   Symfony\Component\Debug\Exception\FatalThrowableError  : Call to a member function oauthClients() on null

  at /app/app/Models/OAuth/Client.php:59
    55|         $this->validationErrors()->reset();
    56|
    57|         if (!$this->exists) {
    58|             $max = config('osu.oauth.max_user_clients');
  > 59|             if ($this->user->oauthClients()->thirdParty()->where('revoked', false)->count() >= $max) {
    60|                 $this->validationErrors()->add('user.oauthClients.count', '.too_many');
    61|             }
    62|         }
    63|

  Exception trace:

  1   App\Models\OAuth\Client::isValid()
      /app/app/Models/OAuth/Client.php:124

  2   App\Models\OAuth\Client::save()
      /app/vendor/laravel/passport/src/ClientRepository.php:124

  Please use the argument -v to see more details.
```
Getting around this issue by disabling that code block works as expected, so perhaps it's enough to not do `$this->user->oauthClients()` when not specifying a user id?

PR #7023, opened 2020-12-23
Update SETUP documentation

ThePooN:
Added a few points that can potentially help future devs deploying environments like https://dev.ppy.sh

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -187,6 +187,11 @@ $ yarn run watch
 
 You can watch emails being sent from `osu-web` by watching the `storage/logs/laravel.log` file. Make sure the `MAIL_DRIVER` is set to `log`.
 
+## Use the API from osu!
+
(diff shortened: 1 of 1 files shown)


SETUP.md
```diff
@@ -183,6 +183,15 @@ To continuously generate assets as you make changes to files (less, coffeescript
 $ yarn run watch
 ```
 
+## Email
+
+You can watch emails being sent from `osu-web` by watching the `storage/logs/laravel.log` file. Make sure the `MAIL_DRIVER` is set to `log`.
+
+## Use the API from osu!
+
+To connect from osu!(lazer) via the API offered by osu-web, you need to create an OAuth client from your user settings. Then, open the `oauth_clients` table and set the `personal_access_client` and `password_client` columns to `1`.
```
venix12, line 192: pretty sure that entire thing could be done using artisan passport commands (would require #6945 to be fixed in the first, though)

nanaya (approved):

PR #7032, opened 2020-12-29
Fix placeholder image for test

nanaya:
Could use another placeholder service but it's simple enough to just host your own.


notbakaneko (approved):
maybe CI should run a local placeholder service 🧨

PR #7050, opened 2021-01-09
Clarify API key origin in setup procedure

bdach:
Possibly obvious, but still tripped me up when setting up. Wasn't expecting the key to be only available on the old site.

I had other issues - aside from me screwing up and assuming `docker-compose` will come from the same PPA as `docker`, resulting in me having a 2017 version of the former with the 2020 version of the latter and refusing to work with a relative path in `docker-compose.yml` (which I consider irrelevant to the setup document), `APP_KEY` was initially blank for some reason - but on consulting with another person that was setting up recently, that doesn't seem to be something universal. Might have been a weird linux thing? No idea.

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -173,7 +173,7 @@ Note that if you use the bundled docker-compose setup, yarn/webpack will be alre
 $ php artisan migrate:fresh --seed
 ```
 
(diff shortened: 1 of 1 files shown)


frenzibyte (approved):
tripped me up as well, looks clear now. 👍🏼

peppy (approved):

PR #7060, opened 2021-01-13
Correctly display deleted user

nanaya:
Seems a bit more verbose than it should be :thinking:


resources/assets/lib/user-card.tsx
```diff
@@ -176,7 +177,7 @@ export class UserCard extends React.PureComponent<Props, State> {
       background = <div className={overlayCssClass} />;
     }
 
-    if (this.isUserLoaded) {
+    if (this.isUserLoaded && !this.user.is_deleted) {
```
notbakaneko, line 180: this combination can be made `isUserVisible` property or something, or just have different render for deleted user.

notbakaneko (changes requested):
Does this need to handle the profile page, indexing, groups, etc?

nanaya:
it should only appear in high scores. Everything else is deleted.

notbakaneko (approved):

PR #7064, opened 2021-01-14
Fix various wiki redirect

### LINKED ISSUE #7035: Wiki abbreviation redirects are incorrectly parsed as locale codes
One case found by pure accident is that https://osu.ppy.sh/wiki/AR will always redirect you to the **ar**abic version of the article, https://osu.ppy.sh/wiki/ar/Beatmapping/Approach_rate.

nanaya:
- case insensitive locale check
- no default page if only locale passed
- check redirect of `/$locale/$path` if `/$path` redirect doesn't exist

Resolves #7035.


notbakaneko (changes requested):
`/wiki/AR` -> `/wiki/ar/Main_Page`
`/wiki/ar` -> `/wiki/en/Beatmapping/Approach_rate`

is it supposed to behave like that?

notbakaneko (approved):

PR #7066, opened 2021-01-14
Simpler forum search

### LINKED ISSUE #2786: Add thread title to results in user's post history
It's helpful to know which thread something was posted in without having to click on it.

### LINKED ISSUE #7021: Sorting forum search results by post time sorts by thread creation time instead of time of post containing the result
It make sense for "post time" to sort by post instead of thread. If this is an intended sorting option, there should be an additional sorting option named "thread creation".

Example of this behavior:

nanaya:
Seems working. Full forum reindex is required for title search to work again.

Resolves #7021. Resolves #2786.

And more. Or less? :thinking: 

- search returns posts instead of topics
- title data is stored in first post
- `users/{user}/posts` uses same search as general forum post


app/Models/Elasticsearch/TopicTrait.php
```diff
@@ -50,16 +50,6 @@ public function getEsId()
...
-            'topic_id' => $this->topic_id,
-            'poster_id' => $this->topic_poster,
-            'forum_id' => $this->forum_id,
-            'post_time' => json_time($this->topic_time),
-            'search_content' => $this->topic_title,
-            'type' => 'topics',
-        ];
-
-        return $values;
+        return ['type' => 'topics'];
```
notbakaneko, line 53: the entire thing doesn't seem to be needed anymore.

notbakaneko (changes requested):
`type` looks like it can be removed from the schema since the join isn't used anymore.
It doesn't look like `Topic` needs to be indexed at all?

Also, I'm not sure if allowing searching post text with single character query string is a good idea 🤔

nanaya:
type is needed so current index can still be used when the new index is being rebuilt.

> Also, I'm not sure if allowing searching post text with single character query string is a good idea :thinking: 

it's to allow searching user posts without entering query. I suppose I could separate it again but I don't think there's much harm in it?

notbakaneko:
I think indexing `Topic` and `filter(['term' => ['type' => 'posts']])` can be skipped and `type` not being added to the new schema; we can probably live with the topic title appearing as individual results while the index is rebuilt?

A matching `parent` document doesn't seem to need to exist and the extra field will just be ignored for the new index

nanaya:
A further cleanup is needed including some logic simplification for indexing and others and I'm thinking of doing them all in followup PR because otherwise this PR would get rather large.

notbakaneko (changes requested):
this will need to handle missing `topic_title` https://github.com/ppy/osu-web/blob/f69111fe4ab3ad4a839c71472ef89e1e2b78c715/app/Libraries/Elasticsearch/Hit.php#L54

rest seems fine 🤔

notbakaneko (approved):

Issue #6480, opened 2020-08-14
Multiple places where English terms are used instead of translated ones


Phippe:
- [x] Beatmap discussion history: Genre/language changes don't use their respective translated terms (e.g. Unspecified, English)<img src="https://user-images.githubusercontent.com/34287258/90236845-99932380-de23-11ea-9bc6-1980247a95da.png"  width="500"/>

- [x] Permissions when asking for authorisation <img src="https://user-images.githubusercontent.com/34287258/99915328-5807f380-2d03-11eb-8c41-d14eab693bb3.jpeg" width="400"/>

- [x] Recent events on profile page: Sentence announcing qualification or ranking of beatmap uses English terms (e.g. qualified)<img src="https://user-images.githubusercontent.com/34287258/90254033-66ac5800-de42-11ea-9eb1-a0a20e9b3649.png"  width="500"/>

- [x] Supporter email: Duration of osu!supporter is in English (e.g. months)<img src="https://user-images.githubusercontent.com/34287258/91457552-c19f6f80-e884-11ea-8d9a-579411cfcc3c.jpg"  width="400"/>

PR #7070, opened 2021-01-15
Show translated scope description

nanaya:
One of #6480.


notbakaneko (approved):
correct locale not being set when the service provider runs seems to make setting the description there pointless 🤔

PR #7085, opened 2021-01-18
Config cleanup

### LINKED ISSUE #7026: assets urls in config are unused and should proabably be removed
https://github.com/ppy/osu-web/blob/c0cf70631844c545ea8f939f38413d2dbfd45b73/config/osu.php#L24-L27

seem to no longer be used anywhere. might be a good time to go through and purge any others we don't use any more.

nanaya:
- [x] `MINI_ASSETS_URL` is now `S3_MINI_URL`

Resolves #7026.

Mainly only checked `osu.php`.


peppy:
I've updated the S3 URL and also gone through and removed any of the deprecated keys (only the `SITE_SWITCHER` one was in use).

peppy (approved):

PR #7093, opened 2021-01-20
Better default difficulty for beatmap page

### LINKED ISSUE #7059: Recommended difficulty doesn't apply on loading a beatmap from the listing
The selected beatmap should be the recommended one (via https://github.com/ppy/osu-web/blob/ce1db2df867d564e4ee082a3f2e01cf032987394/app/Models/User.php#L1690-L1698), on entering the beatmap info page from the listing

nanaya:
Resolves #7059.

Side effect is it defaults to 1.0 for guests and users with no stats for the specific mode (previously it always shows the last difficulty).

Another side effect is it also changes default for beatmap discussion. Which I just noticed should link to currently viewed difficulty? :thinking:


notbakaneko (commented):
If a set has multiple modes but not your default, which should it select? 🤔

nanaya:
> If a set has multiple modes but not your default, which should it select? :thinking: 

...the same as current behaviour but with matching difficulty?

notbakaneko (approved):

PR #7095, opened 2021-01-21
Include user statistics for all modes in index

### LINKED ISSUE #7038: Provide user global rank in each ruleset on the multiple users lookup request
osu!lazer regularly uses the *multiple users lookup* API request (`api/v2/users/?ids[]=x`) when retrieving a set of user data to display them on certain components in a condensed manner, one of which is the multiplayer room's participant panels list. 

The problem is, which leads to this report: Naturally, each participant should have his global rank displayed next to him, and the rank should update in-line with the room's selected beatmap ruleset, which this endpoint doesn't provide currently.

Therefore I propose that the *multiple users lookup* request shall pass in the global ranks for each ruleset per user, to allow for displaying them as such.

nanaya:
<del>Doing it this way makes it a bit more difficult to have custom includes in the future :thinking: </del> no more nesting

Also need to separate `rank` include values as one require additional query while the other doesn't.

Resolves #7038.


app/Transformers/UserStatisticsTransformer.php
```diff
@@ -28,6 +28,7 @@ public function transform(UserStatistics\Model $stats = null)
                 'progress' => $stats->currentLevelProgressPercent(),
             ],
 
+            'global_rank' => $stats->globalRank(),
```
nanaya, line 31: Apparently I used `global_rank` for variants (in this file) and `rank_global` for scores 🤨

app/Transformers/UserStatisticsTransformer.php
```diff
@@ -49,6 +50,7 @@ public function transform(UserStatistics\Model $stats = null)
         ];
     }
 
+    // TODO: split this into global_rank (already exists) and country_rank
```
notbakaneko, line 53: it seems weird to want to split this but then have stats from multiple modes in a `statistics_all` field.

notbakaneko (changes requested):
We should look at standardizing the response fields to either a subtree
```
type: {
  variant1: {},
  variant2: {},
  variant3: {}
}
```
or flat at the top level
```
type_variantX: {}, (or variantX_type?)
```

notbakaneko (approved):

PR #7102, opened 2021-01-22
Proxy markdown first image

### LINKED ISSUE #5243: News header images aren't proxied
except on the user dashboard, since they go through `background_image()`

nanaya:
Resolves #5243.


notbakaneko (commented):
should this have a null check for `proxy_media`?

nanaya:
hmm, `getUrl` only returns string it seems :thinking:

notbakaneko (approved):

PR #7106, opened 2021-01-22
Add watchlisted mappers beatmaps search filter

### LINKED ISSUE #7081: osu! direct category for subscribed mappers.
Fairly self explanatory. It would be really nice if subscribed mappers would get their own osu! direct listing. This would make subscribing to mappers a lot more practical and interactive, and also make it easier to download all maps from a mapper, and not just the most recent ones.

This could of course also be expanded to the website's beatmap listing.

nanaya:
Resolves #7081.

Can use some input on the wording for the filter name:


notbakaneko (approved):

notbakaneko:
maybe `Followed mappers`? Or maybe we need a better term for mappers a user is following 🤔

iptq:
I think "subscribed" mappers has a universally understood meaning since a really old bot that has the same functionality was called osu subscribe bot and osuplus also has subscribe to mapper buttons.

nanaya:
Should it be "By subscribed mappers" or just "Subscribed mappers"? (subscribed? followed?)

notbakaneko:
`Subscribed mappers`

notbakaneko (approved):

PR #7108, opened 2021-01-22
Update laravel

nanaya:
Seems working... so far.

### This will log out all users due to change in how laravel handles cookies.


notbakaneko (approved):

peppy:
Should be fine to merge this one. I will make a reminder to upgrade our redis backend when deploying the next release.

PR #7109, opened 2021-01-22
Add follows page to the language folders

zeusminus:
For some reason, `follows.php` wasn't added to the language folders besides English, which prevented the file to be localized on Crowdin. This PR aims to resolve the issue, *hopefully*. 

I'm guessing this change needs to be synchronized on Crowdin as well, so tagging @peppy here.

### DIFF
--- a/resources/lang/ar/follows.php
+++ b/resources/lang/ar/follows.php
@@ -1,0 +1,38 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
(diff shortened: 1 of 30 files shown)


peppy:
Should this not happen in the opposite direction? Crowdin creates these files, not vice-versa.

nanaya:
I think you need to add the new `follows.php` file in each languages directory when updating the translation :thinking:

peppy:
Let me run my usual sync script and see what happens.

peppy:
#7110 includes actual translated content, will use that instead.

PR #7112, opened 2021-01-25
Remove unindexed date order from user events query

notbakaneko:
Not sure why they were being ordered by `date`, there's no index on `date`; `event_id` should be in the correct order?


notbakaneko:
#7111

PR #7113, opened 2021-01-25
Send user notification digest by user_notification.id

nanaya:
There's currently no index on user_notification.notification_id.

The migration should probably work as is and not take too long.

### DIFF
--- a/app/Jobs/UserNotificationDigest.php
+++ b/app/Jobs/UserNotificationDigest.php
@@ -87,14 +87,18 @@ class UserNotificationDigest implements ShouldQueue
 
     private function getNotifications()
     {
-        return Notification
-            ::whereHas('userNotifications', function ($q) {
-                $q->where('user_id', $this->user->getKey())
-                    ->where('is_read', false)
-                    ->hasMailDelivery();
-            })
+        $notificationIds = $this
+            ->user
+            ->userNotifications()
+            ->where('is_read', false)
+            ->hasMailDelivery()
             ->where('id', '>', $this->fromId)
             ->where('id', '<=', $this->toId)
+            ->select('notification_id')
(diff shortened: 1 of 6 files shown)


database/migrations/2021_01_25_050834_add_count_last_mail_user_notification_id_sent.php
```diff
@@ -0,0 +1,64 @@
...
+     */
+    public function up()
+    {
+        $lastNotificationIdSent = Count::where('name', 'last_mail_notification_id_sent')->first();
+
+        if ($lastNotificationIdSent !== null) {
+            $lastNotificationId = $lastNotificationIdSent->count;
+
+            // prevent existing job from sending more digests
+            $lastNotificationIdSent->update(['count' => $lastNotificationId * 10000]);
```
notbakaneko, line 25: shouldn't this just be some fixed number close to max bigint?

app/Jobs/UserNotificationDigest.php
```diff
@@ -87,14 +87,18 @@ private function filterNotifications(Collection $notifications)
...
+            ->userNotifications()
+            ->where('is_read', false)
+            ->hasMailDelivery()
             ->where('id', '>', $this->fromId)
             ->where('id', '<=', $this->toId)
+            ->select('notification_id')
+            ->pluck('notification_id');
+
+        return Notification
+            ::whereIn('id', $notificationIds)
```
notbakaneko, line 101: `Notification::whereIn('id', $subquery)` instead of sending long list of ids? (besides `select` being redundant with `pluck`)

notbakaneko (changes requested):

notbakaneko (approved):

PR #7115, opened 2021-01-25
Add option to fetch user score rank from cache server

nanaya:
Relevant settings:
```
SCORES_RANK_CACHE_MIN_USERS=35000
SCORES_RANK_CACHE_SERVER_URL=http://127.0.0.1
```

Also relevant for dev:
```
# enable non-caching rank server
# don't forget to also set the SERVER_URL and MIN_USERS
SCORES_RANK_CACHE_LOCAL_SERVER=1
```


notbakaneko (approved):
Should probably look at making timeouts configurable; 1 seconds for connect seems kind of long for production.

peppy:
Hmm, actually now that you mention that, the (non-connect) timeout should be set to 10 here. Otherwise for a potential long lookup it will fail after 5 seconds and perform the same query a second time (via osu-web) even though the rank cache server may return within the following 5 seconds.

I'm fine with the connect timeout being lower - it should be pretty fast from our end, but yeah, the request timeout should either be 10 or configurable

PR #7118, opened 2021-01-27
Track user notifications cleanup

nanaya:
Add count to `osu.web.notifications_cleanup.user_notifications`.

### DIFF
--- a/app/Console/Commands/UserNotificationsCleanup.php
+++ b/app/Console/Commands/UserNotificationsCleanup.php
@@ -8,6 +8,7 @@ namespace App\Console\Commands;
 use App\Libraries\Notification\BatchIdentities;
 use App\Models\User;
 use App\Models\UserNotification;
+use Datadog;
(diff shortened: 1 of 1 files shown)


app/Console/Commands/UserNotificationsCleanup.php
```diff
@@ -59,6 +60,7 @@ public function handle()
                 );
                 $deleted = count($notificationIds);
                 $deletedTotal += $deleted;
+                Datadog::increment(config('datadog-helper.prefix_web').'.notifications_cleanup.user_notifications', $deleted);
```
notbakaneko, line 63: I think you mean
```suggestion
                Datadog::increment(config('datadog-helper.prefix_web').'.notifications_cleanup.user_notifications', 1, null, $deleted);
```
2nd param is sample rate

notbakaneko (changes requested):

notbakaneko (approved):

PR #7121, opened 2021-01-27
Fix cached rank check

### LINKED ISSUE #7120: Wrong global and local stats on a beatmap (Osu!)
As soon as I look at the detailed statistics of a map on my profile my global and local rank is wrong and always the same.
Let's say I got 530th global on a beatmap, but instead of 530th global it says 23050 global and the exact same number for local rank

I found out that this not only the case with my plays, but if I go the country ranking list and choose a random profile, I'll find wrong detailed statistics about a level relating global and local ranking as well...
Capture4
Capture5
Capture6

Capture2

nanaya:
The check was kinda very wrong :upside_down_face: 

Resolves #7120.

### DIFF
--- a/app/Models/Score/Best/Model.php
+++ b/app/Models/Score/Best/Model.php
@@ -144,11 +144,7 @@ abstract class Model extends BaseModel
     {
         $server = config('osu.scores.rank_cache.server_url');
 
-        if (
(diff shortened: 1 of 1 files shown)


peppy (approved):
oops.

PR #7127, opened 2021-01-28
Add job to delete notifications

nanaya:
Should be fine deleting up to notification id of earliest user notification (further subtracted by 1000) :shrug: 

Hopefully nothing breaks on that delete with limit and primary key constraint.


app/Console/Commands/NotificationsCleanup.php
```diff
@@ -0,0 +1,56 @@
...
+        $total = config('osu.notification.cleanup.max_delete_per_run');
+
+        if ($total === 0) {
+            return;
+        }
+
+        $perLoop = min($total, static::MAX_PER_LOOP);
+        $loops = $total / $perLoop;
+
+        $maxNotificationId = optional(UserNotification::first())->notification_id;
```
notbakaneko, line 33: `first()` by itself doesn't guarantee order

app/Console/Commands/NotificationsCleanup.php
```diff
@@ -0,0 +1,56 @@
...
+
+        $maxNotificationId -= static::NOTIFICATION_ID_BUFFER;
+
+        $this->line("Deleting notifications up to {$maxNotificationId}");
+
+        $deletedTotal = 0;
+
+        for ($i = 0; $i < $loops; $i++) {
+            $deleted = Notification::where('id', '<', $maxNotificationId)->limit($perLoop)->delete();
+            Datadog::increment(config('datadog-helper.prefix_web').'.notifications_cleanup.user_notifications', 1, null, $deleted);
```
notbakaneko, line 49: should be different from the user notification cleanup key?

notbakaneko (changes requested):

notbakaneko (approved):

PR #7132, opened 2021-01-29
Fix new PM channel disappearing from presence update

### LINKED ISSUE #7131: Starting a new PM may send message to wrong person
I've hit this multiple times now, and it's very dangerous.

Steps:

- Have some PM channels already open
- Send a new PM from someone's profile
- Start typing a message or two quickly
- Note that the channel changes while you are typing the message, potentially sending subsequent messages to a different person

I believe this may only work if you are sending the new message to someone you don't already share message history with.

https://user-images.githubusercontent.com/191335/106233082-05821080-6239-11eb-820f-51d11997e18f.mp4

notbakaneko:
closes #7131

This happens if a new PM is sent in between the request and response of the presence update (and only if there are new updates):
1. request presence update
2. send new pm
3. pm channel transferred
4. receive presence update response (does not contain new pm)
5. new pm channel disappears
6. new pm channel reappears after next presence update


resources/assets/lib/models/chat/channel.ts
```diff
@@ -23,6 +23,7 @@ export default class Channel {
   @observable moderated: boolean = false;
   @observable name: string = '';
   @observable newPmChannel = false;
+  @observable newPmChannelTransient = false;
```
nanaya, line 26: does it need to be `@observable`?

nanaya (changes requested):

nanaya (approved):

PR #7135, opened 2021-01-29
Persistent show deleted comments option

### LINKED ISSUE #7045: Make "Show Deleted" persistent
Similar story as with #6218, just with comments this time.
Main reason for this is that you can not see deleted comments on a permalink, since you can not able "Show Deleted" on permalinks. Alternative would be adding "show deleted" to comments you visit via permalink.
But in addition to that it probably should just be a persistent setting anyways.

nanaya:
Resolves #7045.


resources/assets/coffee/react/_components/comments-manager.coffee
```diff
@@ -81,8 +81,14 @@ export class CommentsManager extends React.PureComponent
...
   toggleShowDeleted: =>
+    nextState = !uiState.comments.isShowDeleted
+
     runInAction () ->
-      uiState.comments.isShowDeleted = !uiState.comments.isShowDeleted
+      uiState.comments.isShowDeleted = nextState
+
+    $.ajax laroute.route('account.options'),
+      method: 'PUT'
+      data: user_profile_customization: comments_show_deleted: nextState
```
notbakaneko, line 91: it's time to just move this to the store

notbakaneko (changes requested):

resources/assets/coffee/react/_components/comments-manager.coffee
nanaya: or maybe when splitting it off from `UIStateStore`. The entirety of ui store for comment is currently part of `UIStateStore`. Or more like everything is part of it.

notbakaneko (approved):

PR #7138, opened 2021-01-29
User notification loading optimizations

notbakaneko:
This tries to optimize user notification loading:
- the grouping fields from `notifications` are duplicated to `user_notifications` to make querying them simpler and avoid combining `group by` with a dependent query; 
- `user_notifications` uses `category` instead of `name` since that's what the notifications are grouped by;
- try to reduce the number of queries used to load notifications; currently, each notification stack uses multiple separate queries to get the various pieces of data needed; due to pagination and count displays, it's not practical to have an index for every query combination, not to mention `count` ends up requiring some sort of scan, anyway.
- if a notification stack is smaller than the page size, the order they're returned in doesn't matter since there's no pagination to deal with; these can be batched together in the same query.

There are some combination of indices that help with the `group by` queries, but the need to support ordering for  pagination and counts end up making them useless; instead, we just amortize the cost of getting the stack size together with it.

Some empirical differences when loading the notifications page:
Before:
image

After:
image

### DIFF
--- a/app/Jobs/Notifications/BroadcastNotificationBase.php
+++ b/app/Jobs/Notifications/BroadcastNotificationBase.php
@@ -175,9 +175,13 @@ abstract class BroadcastNotificationBase implements ShouldQueue
 
             foreach ($deliverySettings as $userId => $delivery) {
                 $userNotification = $notification->userNotifications()->create([
-                    'delivery' => $delivery,
-                    'user_id' => $userId,
+                    'category' => $notification->category,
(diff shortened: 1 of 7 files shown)


notbakaneko:
supposed to be draft because how to actually add the columns to production need to be worked out 👀

peppy:
Is this still just waiting on adding columns to a table? That can be organised quite easily, so go ahead with a review if required.

PR #7145, opened 2021-02-01
Beatmap discussion watchlist better sort and other updates

### LINKED ISSUE #2017: Sort Modding Watchlist Page by Most Recently updated beatmap
Currently the maps on the page are being sorted from oldest to newest initially uploaded.

I think having the most recently updated maps at the top of the list would make more sense, with less recently updated maps as you go down. This way maps more relevant to looking at a page dedicated to keeping you up-to-date with new map posts would be at the top.

Just like the bookmarks or subscription pages for forum threads, threads most recently posted in are at the top of the list.

nanaya:
Resolves #2017.

Also reduces number of queries and added a new column indicating last update time.

Other watchlist pages can also use better sorting which will be done separately. I thought there's open issue for it but I couldn't find it :question:


notbakaneko (approved):
no index on the sorting but
```
> nanaya
there's more things that'll be slow if the user has many watches
```

PR #7146, opened 2021-02-01
Disable expensive chat presence group by query

notbakaneko:
~Breaks the the unread markers on channels; we probably want to try link those with notifications instead or something~ 🤔 
`last_message_id` is now saved with channel, `first_message_id` may return later (there's an endpoint that includes it but web chat doesn't use it yet).

Also includes a fix for polling skipping messages in some cases.

`last_message_id` is only used for the unread markers (and the `since` in polling, it seems).
`first_message_id` is only used to handle the `show more` state - it seems better to just eat the extra request and return no messages instead.

- [x] add `2021_02_02_073616_add_last_message_id_to_channels` to migrations


nanaya:
There's no notification for public channels though. I wonder if it's better to cache the values in channel table instead.

peppy:
I did notice the inline comment suggesting the last_id should be done client side

https://github.com/ppy/osu-web/blob/dabf703b776ebf225588a55cc962bde2d268f5ef/app/Models/Chat/UserChannel.php#L82-L83

I guess this is referring to doing one initial lookup then tracking from that point onwards client side?

notbakaneko:
Well, once you start loading messages in the channel, the last id can be inferred, so it's a matter to providing a placeholder until then.

peppy:
Either way, tracking in the channels table sounds like a valid approach.

resources/assets/lib/models/chat/channel.ts
```diff
@@ -51,6 +52,17 @@ export default class Channel {
     return this.messages[this.messages.length - 1];
   }
 
+  @computed
+  get lastMessageId() {
```
nanaya, line 56: looks like `lastReadId` ended up a bit behind with this as it's only updated before the actual sent message id is set.

nanaya (changes requested):

nanaya:
The new column (and migration entry) should also be added ahead of time.

resources/assets/lib/chat/chat-api-responses.ts
```diff
@@ -25,11 +25,16 @@ export interface ChannelJson {
 
 // This is the version used by 'presence'.
 export interface ChannelJsonExtended extends ChannelJson {
-  first_message_id: number;
   last_message_id: number;
   last_read_id: number;
 }
 
+export interface ChatInitialJson {
+  last_message_id: number | null;
```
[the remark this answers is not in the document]
notbakaneko: turns out relying only on client side max id resulting in subtle polling bug 🤔

resources/assets/lib/chat/chat-worker.ts
```diff
@@ -39,9 +39,7 @@ export default class ChatWorker implements DispatchListener {
 
     this.updateXHR = true;
 
-    const maxMessageId = this.channelStore.maxMessageId;
```
nanaya, line 42: apparently there's nothing else using `maxMessageId` anymore so it can be removed?

nanaya (dismissed):
Can use one cleanup but otherwise fine if just want to deploy quickly

nanaya (approved):

peppy:
Column and migration have been added.

peppy (approved):

PR #7160, opened 2021-02-03
Add search API endpoint

### LINKED ISSUE #2162: API: Add ability to search for users
Like Beatmapsets searching in rankings (users) would be great.

This would include
a query (possibly looking into name and location?),
sorting (name aka alphabetical, rank and location),
and ordering (asc / desc).

nanaya:
Resolves #2162. Additional search options will be added later.


notbakaneko (changes requested):
this returns nothing for `user` when using client credentials token since `auth()->check()` doesn't include presence of oauth token

notbakaneko (approved):

PR #7166, opened 2021-02-04
Check user overall room attempts for room max attempts limit

nanaya:
This changes the room rule. Not sure if highest allowed max attempts of 32 is enough with this change.


notbakaneko:
can you have more playlist items than attempts ?_?

nanaya:
Maybe if the playlist isn't intended to be played in full.

peppy:
I think we can set a higher number than 32. Maybe 128 for the time being? Configurable maybe.

And yep, having the attempt count lower than items in room is a valid and interesting customisation (players would need to choose which items they play to maximise their ending score) so we do want to support that.

Issue #2455, opened 2018-02-16
Edit profile when hovering near the details on user page


Joehuu:
![](https://puu.sh/zpjOR/dcaaf20154.png)
![](https://puu.sh/zpkip/e6578c0af1.png)

Xinnoh:
You can access the page to edit this information only one web page away via settings, the old site requires two so less of an issue

Joehuu:
Should this be a thing? The feeling of pressing a button near it is better than navigating through settings.

peppy:
It might be handy to have a link for this kind of thing, where it's not completely obvious  how to edit them.

willowsnow:
could instead just make them editable textboxes based on whether or not the user is logged in ^^

arflyte:
![a8eb94cd46](https://user-images.githubusercontent.com/3632750/36579880-e469ea12-18a8-11e8-8129-c52560367502.png)
Editable textboxes wouldn't fit in tight space. Added direct link to profile edit page for now, before we implement the new profile page design.

PR #7167, opened 2021-02-04
Add an edit button to the profile links section

iptq:
Trying out something simple to learn the codebase, so I'd appreciate any kind of advice on this.

This patch adds an Edit button to the right side of the links section, as requested in https://github.com/ppy/osu-web/issues/2455. When clicked, it just navigates to the account edit page where the user can make the changes they wanted to make.

image

This is what it looks like when there's only one row:

image

Closes https://github.com/ppy/osu-web/issues/2455


resources/assets/less/bem/profile-links.less
```diff
@@ -36,4 +42,27 @@
     font-weight: 600;
     word-wrap: break-word;
   }
+
+  .edit-profile-links {
```
nanaya, line 46: this should conform with [bem](https://css-tricks.com/bem-101/). Either as a new block (in its own file) or as element of `profile-links` (`profile-links__actions`)

resources/assets/less/bem/profile-links.less
```diff
@@ -36,4 +42,27 @@
...
+    position: absolute;
+    top: 0;
+    right: 0;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: 70px;
```
nanaya, line 53: this should be `@gutter-profile-page` on default and `@gutter-profile-page-desktop` on desktop. Check `page-extra__actions`.

resources/assets/coffee/react/profile-page/links.coffee
```diff
@@ -103,6 +103,14 @@ export class Links extends React.PureComponent
       for row, j in rows when _.compact(row).length > 0
         div key: j, className: "#{bn}__row #{bn}__row--#{j}", row
 
+      div
+        className: 'edit-profile-links'
+        a
+          href: laroute.route('account.edit')
+          button
```
nanaya, line 110: `button` can't be inside `a`. I don't see the point anyway (apart of skipping some styling).

resources/assets/less/bem/profile-links.less
```diff
@@ -36,4 +42,27 @@
...
+      color: @osu-colour-c1;
+      button {
+        background-color: @osu-colour-b2;
+        outline: none;
+        border: none;
+        border-radius: 50%;
+        width: 40px;
+        height: 40px;
+      }
+    }
```
nanaya, line 66: just use `profile-page-toggle`. The styling will need a bit of update so it also works with `a` element.

nanaya (dismissed):

peppy:
@arflyte can you confirm you're okay with there being three permanently displaying edit buttons on the profile page with this addition?

iptq:
Oh shoot, I meant for the edit button to only display in case the user
is viewing their own userpage. I'll implement that tomorrow.

peppy:
I am presuming that is already the case (haven't checked your logic). Even in such a case, there are now three edit buttons in total (see two existing ones in screenshot):

20210208 190205 (Safari app)

arflyte:
Since it's just for the player's own profile, yeah. Although I definitely need to refine the edit button a lil bit more.

notbakaneko (approved):

PR #7168, opened 2021-02-04
Per-playlist item max attempts

### LINKED ISSUE #7150: A `maximum_attempts` field should be added per playlist item
Currently, there is a maximum attempt value for the room as a whole, but going forward we also want to allow customisations like this to each playlist item. It should work in the same way as the room level one (and both should be checked if they are both applied).

nanaya:
Resolves #7150.

- [x] continues from #7166


tests/Models/Multiplayer/RoomTest.php
```diff
@@ -70,17 +71,50 @@ public function testMaxAttemptsReached()
...
-        $room->startPlay($user, $playlistItem);
+        $room->startPlay($user, $playlistItem2);
         $this->assertTrue(true);
 
         $this->expectException(InvariantException::class);
-        $room->startPlay($user, $playlistItem);
+        $room->startPlay($user, $playlistItem1);
+    }
+
+    public function testMaxAttemptsForItemReached()
```
notbakaneko, line 91: This looks like it's just testing the same thing twice, the previous test checks that the room limit is hit, this one should probably be checking the playlist limit is hit but not the room limit, so maybe it should be testing is other plays can still be started?

notbakaneko (changes requested):

tests/Models/Multiplayer/RoomTest.php
nanaya: it's already testing whether other item can still be started, just no assertion (now added).

notbakaneko (approved):

PR #7169, opened 2021-02-05
Add current user score/stats to room json

### LINKED ISSUE #7148: Local user statistics should be returned with a playlist room request
A call to `/api/v2/rooms/<room_id>` should return some statistics about the local user for display purposes. Currently the one I'm interested in is attempt count as calculated below:

https://github.com/ppy/osu-web/blob/3169b33ccc4c540be5f20136393ad5f00d635ff9/app/Models/Multiplayer/Room.php#L342

More may come as we build on the system. We also need attempt count per playlist item in the near future, so adding that as well may be a good measure.

nanaya:
Not sure if the playlist item attempts attribute makes sense but there's no playlist item stats at the moment. There's high score but I don't know if it makes sense adding "attempts" attribute to that. Also probably will eventually need to cache that instead of manual count?

Resolves #7148.


app/Transformers/Multiplayer/RoomTransformer.php
```diff
@@ -35,6 +37,14 @@ public function transform(Room $room)
         ];
     }
 
+    public function includeCurrentUserScore(Room $room)
+    {
+        $user = auth()->user();
```
notbakaneko, line 42: needs to handle null user

notbakaneko (changes requested):

notbakaneko (approved):

PR #7170, opened 2021-02-05
Move notifications endpoint

nanaya:
Avoids old sessions from retrying with invalid sessions and just generally better without `/home` prefix.


notbakaneko:
should be web socket ones be updated separately?

notbakaneko (approved):

PR #7174, opened 2021-02-05
Adjust topic watches when the topic is moved

### LINKED ISSUE #4882: Forum notifications shown to users that don't have access to the topic/subforum
The user who took the picture below (https://osu.ppy.sh/users/2964029) originally bookmarked [this thread](https://osu.ppy.sh/community/forums/topics/381215) when they had access to it and are still receiving notifications even though they can no longer access it.

I'm not sure if topic watches are updated when a thread is moved (they probably should be if they aren't), but at the very least the notification shouldn't be shown if you can't see the topic the notification is for.

nanaya:
Resolves #4882.


app/Models/Forum/Topic.php
```diff
@@ -249,18 +250,22 @@ public function moveTo($destinationForum)
...
+            // this relies on dispatcher always reloading the model
+            ->select(['poster_id', 'post_id'])
+            ->each(function ($post) {
+                dispatch(new UpdateUserForumCache($post->poster_id));
+                dispatch(new EsIndexDocument($post));
+            });
+
+        dispatch(new UpdateUserForumTopicFollows($this));
+
+        return $ret;
```
notbakaneko, line 268: isn't this just going to be always `true` 🤔

notbakaneko (changes requested):

notbakaneko (approved):

PR #7181, opened 2021-02-08
Fix error on displaying score of deleted user

nanaya:
Use placeholder country for (hard-)deleted users.


app/Models/Score/Best/Model.php
```diff
@@ -293,7 +293,7 @@ public function scopeWithType($query, $type, $options)
     {
         switch ($type) {
             case 'country':
-                $countryAcronym = $options['countryAcronym'] ?? $options['user']->country_acronym;
+                $countryAcronym = $options['countryAcronym'] ?? $options['user']->country_acronym ?? 'XX';
```
notbakaneko, line 296: I would think `$query->none()` would make more sense for non-existent users but I guess this works fine as long as there aren't any scores matching `XX` 🤔

notbakaneko (approved):

PR #7182, opened 2021-02-08
Add job to fix display order of inactive channels

### LINKED ISSUE #7175: Archived web chats are being sorted in reverse chronological order
Already opened chats are displaying in reverse order, meaning all recent messages I've received are now at the very bottom of the chat list. This is especially troublesome as a BN, given I cannot find any of the recent messages due to the sheer amount of messages I have received and left open in new chat. 

Newly opened chats do move to the top as intended, but once they're "archived" again they're sent back to the very bottom.

I made a recording where I open a recent chat and then refresh the page, to demonstrate the behavior:

https://user-images.githubusercontent.com/12388394/107062518-2528b400-67b8-11eb-9b6a-18c8fd128a46.mp4

notbakaneko:
resolves #7175 by setting `last_message_id` on channels that haven't had it set yet.

`php artisan chat:channel-set-last-message-id`
takes `--chunk-size` (default 1000) `--delay` (default 1) between chunks as optional parameters; allows have no delay with `--delay=0`


app/Console/Commands/ChatChannelSetLastMessageId.php
```diff
@@ -0,0 +1,58 @@
...
+            return;
+        }
+
+        $this->line("Updating chat channels without last_message_id with {$delay}s delay between chunks of {$chunkSize}...");
+
+        $query = Channel::where('last_message_id', null);
+        $progress = $this->output->createProgressBar($query->count());
+        $query->chunkById($chunkSize, function ($chunk) use ($delay, $progress) {
+            foreach ($chunk as $channel) {
+                $lastMessageId = optional(Message::where('channel_id', $channel->getKey())->last())->getKey();
```
nanaya, line 35: `$channel->messages()->max('message_id')`?

app/Console/Commands/ChatChannelSetLastMessageId.php
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    private function getOptionValue(string $name, int $default)
+    {
+        $value = $this->option($name);
+        if (!ctype_digit($value)) {
+            $value = $default;
+        }
+
+        return get_int($value);
```
nanaya, line 56: not sure if it's that much more useful over just `return get_int($this->option($name)) ?? $default;`

app/Console/Commands/ChatChannelSetLastMessageId.php
```diff
@@ -0,0 +1,58 @@
...
+        $delay = $this->getOptionValue('delay', 1);
+        $chunkSize = $this->getOptionValue('chunk-size', 1000);
+        if ($chunkSize < 1) {
+            return;
+        }
+
+        $this->line("Updating chat channels without last_message_id with {$delay}s delay between chunks of {$chunkSize}...");
+
+        $query = Channel::where('last_message_id', null);
+        $progress = $this->output->createProgressBar($query->count());
```
nanaya, line 32: not sure about counting unindexed column. At least it currently only takes 800ms so maybe not too bad if only meant to be run once...

app/Console/Commands/ChatChannelSetLastMessageId.php
```diff
@@ -0,0 +1,58 @@
...
+namespace App\Console\Commands;
+
+use App\Models\Chat\Channel;
+use App\Models\Chat\Message;
+use Illuminate\Console\Command;
+
+class ChatChannelSetLastMessageId extends Command
+{
+    const MAX_PER_LOOP = 1000;
+    const NOTIFICATION_ID_BUFFER = 1000;
```
nanaya, line 15: nice consts

nanaya (changes requested):

app/Console/Commands/ChatChannelSetLastMessageId.php
```diff
@@ -0,0 +1,46 @@
...
+        $delay = $this->option('delay') ?? 1;
+        $chunkSize = $this->option('chunk-size') ?? 1000;
+        if ($chunkSize < 1) {
+            return;
+        }
+
+        $this->line("Updating chat channels without last_message_id with {$delay}s delay between chunks of {$chunkSize}...");
+
+        $progress = $this->output->createProgressBar();
+        Channel::where('last_message_id', null)->chunkById($chunkSize, function ($chunk) use ($delay, $progress) {
```
nanaya, line 28: may still want to not filter by `last_message_id` here as `chunkById` doesn't put any upper bound :thinking: 

Either that or the null check below is kinda redundant.

nanaya (changes requested):

app/Console/Commands/ChatChannelSetLastMessageId.php
notbakaneko: should be fine since it's going to scan all the rows anyway 🤔

nanaya (approved):

PR #7187, opened 2021-02-09
More frequent user reindex

### LINKED ISSUE #7185: Recent active user search doesn't quite work
(`recent_only=1` parameter for user search)

It searches based on `user_lastvisit` in the elasticsearch but the entry is only updated on renames and registrations.

It currently "works" on live site because we recently reindexed the whole thing but it'll get stale as time passes.

Maybe should just let user to be reindexed every time it's updated?

Edit: users hiding their presence shouldn't have the actual value indexed

nanaya:
Also hides last visit time accordingly.

Resolves #7185.


peppy:
How often will this be updated now? On a quick check it looks like we're updating the session every page visit (but not sure about database..? can't find this anywhere).

nanaya:
Every minute at most per user as the update is rounded to 0 second.

peppy:
I'm not sure if this is going to cause load issues on ES, but I do know that at bancho's end we batch and limit updates to once every n minutes per user because the database overhead is considerable.

Is osu-web ever updating the database `user_lastactive` currently?

nanaya:
That's the update frequency for user_lastvisit.

peppy:
Hmm, I see. Maybe worth documenting where that's happening in code because it's a bit hard to find.

Also worth noting that this won't reindex if a user is active in-game but not on the site. Probably something to consider in the future (or handle via one of lazer's osu-web API callbacks).

notbakaneko (changes requested):

notbakaneko (commented):
This probably isn't an issue as long as indexing is always done by `queue:work` or in a persistent shared process, but `$mappings = $this->esFilterFields();` should probably be removed since it needs to read the columns from database, and the assignments for the extra fields moved into the loop or switched to a lookup mapping.

notbakaneko (approved):

PR #7189, opened 2021-02-09
Add match API endpoints

### LINKED ISSUE #6906: API route for community/matches/:id scrapping
Hi,
Would be grateful if you would create route for scrapping results from that page:
https://osu.ppy.sh/community/matches/60883767
or extend the GET /rooms/:id route to get the ID for the community match page.

If it's hard to understand, please try to make  **GET /rooms/1** request and then, try to find it on website.
https://osu.ppy.sh/community/matches/1 - it does not work, even when I provide the "channel_id" value https://prnt.sc/vptqlq

It's important for script that is nearly done, I mentioned it in another issue, right here:
https://github.com/ppy/osu-web/issues/6882

thanks

nanaya:
Resolves #6906. Except no documentation :eyes: Are we still calling this `Matches`?

This uses cursor pagination so people can browse to their heart's content without slowing down things. I think.

Only non-private rooms included in listing for now as (it seems) there's no index to quickly list own rooms. Individual private rooms can still be viewed with correct permissions.


peppy:
Seems fine, but not sure about allowing the ASC sort direction? Wouldn't this just return the oldest matches in the database (and since we nuke old ones, it would be a very random and non-useful selection).

notbakaneko (approved):

PR #7200, opened 2021-02-12
Fix multiplayer allowed mods check failing

smoogipoo:
The user's mods list contains both the required mods and the allowed mods they selected.
```
allowed_mods = HR
required_mods = DT

mods = [DT, HR]
```
Please advise if there are tests to be updated here or if this is improper code style w.r.t. performance and all that :).


notbakaneko:
What you're saying is `required_mods` are implicitly allowed?

smoogipoo:
Yeah, `required_mods` always need to be there, `allowed_mods` is just whatever the user picks as extra on top of `required_mods`.

notbakaneko (approved):

PR #7201, opened 2021-02-12
Move per-mode statistics to its own transformer

### LINKED ISSUE #7197: Ruleset statistics returned by /v2/users/?ids[]=<id> are not easily deserializable lazer-side
Currently `statistics_{osu/taiko/...}` properties from `/v2/users/?ids[]=<id>` are added directly to the main user JSON object:
```json
{
    "users": [
        {
            "id": 14210502,
            "statistics_fruits": {
                 ...
            },
            "statistics_mania": {
                 ...
            },
            "statistics_osu": {
                 ...
            },
            "statistics_taiko": {
                 ...
            },
        ...
        }
    ]
}
```

Which makes deserializing this lazer-side complicated than usual, as it is not simply determinable what the property names are, without writing extra logic to get all properties that start with "statistics_" and parse them properly.

Therefore I request making these statistics be contained in a single property called `all_statistics` or otherwise, that makes deserializing this lazer-side pretty straightforward, given how the deserializer used there does its job:
```json
{
    "users": [
        {
            "id": 14210502,
            "all_statistics": {
                "fruits": {
                    ...
                },
                "mania": {
                    ...
                },
                "osu": {
                    ...
                },
                "taiko": {
                    ...
                },
            }
        ...
        }
    ]
}
```

nanaya:
Resolves #7197.

<del>`statistics_all` name is kind of a lie because it might not contain all modes :eyes: </del>
<del>Maybe should just default include all modes? Or rename to something else? (`statistics_modes`?)</del>
Now `statistics_rulesets`.

This relies on transformer renderer to trust `availableIncludes` list and call the intended method right away instead of doing `method_exists` check first. [The comment here says "check method name" but it's not actually done](https://github.com/thephpleague/fractal/blob/9c9c493edcf0f5438555e5b515a790749a26c807/src/TransformerAbstract.php#L187).


notbakaneko:
If anything, `statistics` only being a single mode that can be different is a lie and is the one that needs renaming 🤔

peppy:
We'd prefer it only includes statistics which exist for the user. Renaming to `statistics_rulesets` might be best?

notbakaneko (approved):

PR #7204, opened 2021-02-12
Add storyboard icon on beatmap page

### LINKED ISSUE #7184: "Has storyboard" indicator in beatmap info page
They shows only at beatmap listing page and not at beatmap info page. People who open a beatmap link directly without going from beatmap listing won't know if that beatmap has storyboard or not until they check it themselves in game.
image

nanaya:
Resolves #7184.


notbakaneko (approved):

Hiviexd:
imo you could've done the same for the video icon too, feels kinda weird seeing the storyboard icon in the beatmap page but not the video one

PR #7205, opened 2021-02-12
Exclude period from end of url when parsing

### LINKED ISSUE #7161: Link parsing in chat includes the period as part of the link
As demonstrated below:

notbakaneko:
plus some other common ones.

Was going to try excluding all punctuation `\p{P}` with some url exclusions, but javascript doesn't seem to support character class subtraction and I think the regex is already complicated enough without adding more workarounds for it.

Also Safari doesn't support negative/positive look behinds, etc.

closes #7161

As an aside, seeing how complicated a regex for accurately parsing urls in arbitrary strings is, I think it'd be more useful having a less accurate parser that's simpler with easier to customize rules for usability.


resources/assets/coffee/osu_common.coffee
```diff
@@ -3,7 +3,7 @@
 
 @osu =
   isIos: /iPad|iPhone|iPod/.test(navigator.platform)
-  urlRegex: /(https?:\/\/((?:(?:[a-z0-9]\.|[a-z0-9][a-z0-9-]*[a-z0-9]\.)*[a-z][a-z0-9-]*[a-z0-9](?::\d+)?)(?:(?:(?:\/+(?:[a-z0-9$_\.\+!\*',;:@&=-]|%[0-9a-f]{2})*)*(?:\?(?:[a-z0-9$_\.\+!\*',;:@&=-]|%[0-9a-f]{2})*)?)?(?:#(?:[a-z0-9$_\.\+!\*',;:@&=/?-]|%[0-9a-f]{2})*)?)?))/ig
+  urlRegex: /(https?:\/\/((?:(?:[a-z0-9]\.|[a-z0-9][a-z0-9-]*[a-z0-9]\.)*[a-z][a-z0-9-]*[a-z0-9](?::\d+)?)(?:(?:(?:\/+(?:[a-z0-9$_\.\+!\*',;:@&=-]|%[0-9a-f]{2})*)*(?:\?(?:[a-z0-9$_\.\+!\*',;:@&=-]|%[0-9a-f]{2})*)?)?(?:#(?:[a-z0-9$_\.\+!\*',;:@&=/?-]|%[0-9a-f]{2})*)?)?(?:[^\.,;:\s])))/ig
```
nanaya, line 6: the `;` exclusion causes incorrect result when it encounters trailing `&gt;` (or other html entities)

nanaya (changes requested):

nanaya (approved):

PR #7209, opened 2021-02-15
Support null output for empty transforms

notbakaneko:
Returns `null` instead of `[]` if the transformer and its includes are empty. This is an issue with the `UserStatisticsRulesetsTransformer` if there are no statistics since empty arrays get serialized as `[]`

Serializing empty `stdClass` results in proper empty object `{}`, however, the transformer library used can't handle anything but array which would require a lot more stuff to be reimplemented.

I don't think there's any case where the main transformer returns a non-associative array with values, so an option shouldn't be necessary.

### DIFF
--- a/app/Libraries/Transformers/Scope.php
+++ b/app/Libraries/Transformers/Scope.php
@@ -23,7 +23,13 @@ class Scope extends Fractal\Scope
             }
         }
 
-        return parent::fireTransformer($transformer, $data);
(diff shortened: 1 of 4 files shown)


nanaya:
would this return null for empty collections as well? :thinking:

notbakaneko:
empty collections should still return `[]`

nanaya (approved):

PR #7218, opened 2021-02-16
Allow deleting topic

### LINKED ISSUE #1584: Can't soft-delete forum topic/threads
Only way currently seems to be individually deleting all the posts... which is not great. Guessing this was an oversight.

nanaya:
As part of a different cleanup, found this kinda should be resolved as I don't want to deal with non-permanent first post. There may still be problem with post positioning but that can be fixed later™

Resolves #1584.

- [x] depends on #7213


notbakaneko (changes requested):
Is deleting the first post going to delete the topic now?

If you're a moderator and view a topic where the original first post was previously deleted, the cached first post id won't match the first post, causing the show more link to endlessly spin

notbakaneko (approved):

PR #7221, opened 2021-02-17
Fix discussion tabs label

nanaya:
Not sure how I missed this :upside_down_face:


notbakaneko (approved):
I thought I changed that and deleted the translation 🤔

PR #7223, opened 2021-02-17
Add `expired` property to multiplayer playlist items

smoogipoo:
I was debating whether to name this `played` or `expired`, but picked `expired` in the end as it could be used for more than multiplayer. For example if a host is ever to be able to "delete" playlist items then we could also set `expired = 1` in that case and preserve scores/etc. Whether such functionality should exist or not is a different question altogether, but I hope the example demonstrates why I didn't want to tie this property down to _just_ the playlist item having been "played" in multiplayer.

### DIFF
--- a/app/Http/Controllers/Multiplayer/Rooms/Playlist/ScoresController.php
+++ b/app/Http/Controllers/Multiplayer/Rooms/Playlist/ScoresController.php
@@ -193,6 +193,8 @@ class ScoresController extends BaseController
             ->where('id', $playlistId)
             ->firstOrFail();
 
+        abort_if($playlistItem->expired, 422, 'expired playlist item');
+
         $roomScore = $playlistItem->scores()
             ->where('user_id', auth()->user()->getKey())
             ->where('id', $scoreId)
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/Multiplayer/Rooms/Playlist/ScoresController.php
```diff
@@ -193,6 +193,8 @@ public function update($roomId, $playlistId, $scoreId)
             ->where('id', $playlistId)
             ->firstOrFail();
 
+        abort_if($playlistItem->expired, 422, 'expired playlist item');
```
nanaya, line 196: This probably should be done in room's (`app/Models/Multiplayer/Room.php`) `assertValidStartPlay`

nanaya (changes requested):

database/migrations/2021_02_17_073637_add-expired-column-to-playlist-items.php
```diff
@@ -0,0 +1,35 @@
...
+class AddExpiredColumnToPlaylistItems extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('multiplayer_playlist_items', function (Blueprint $table) {
```
peppy, line 19: should be fine to let this run synchronously

peppy (approved):

nanaya (approved):

PR #7225, opened 2021-02-17
Add cursor pagination for forum posts

### LINKED ISSUE #5078: Load more links in forum threads always point to the originally loaded posts
They should probably update to point to new posts with each fetched update.

### LINKED ISSUE #7247: Deleted Threads trying to load "show more" forever
If a thread is deleted it tries to load stuff via "show more" infinetly
85e2FukaD8

nanaya:
- support for `cursor` parameter
- old `start` and `end` (and `n`) stay because they serve different purpose
- preloading posts now doesn't involve calculating post position
- update load more links (resolves #5078)
- correctly handle case when there's no more posts to load (resolves #7247)

<del>Note that the final sorting can probably be smarter. Maybe I'll do that when adding the API thing.</del> Easy enough so added here instead.

- [x] depends on #7218


app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -272,63 +274,83 @@ public function show($id)
...
+            $jumpTo = $firstPost->getKey();
+
+            if ($cursorHelper->getSortName() === 'id_asc') {
+                if ($jumpTo !== $topic->topic_first_post_id) {
+                    $extraSort = 'id_desc';
+                }
+            } else {
+                $extraSort = 'id_asc';
+            }
+            if (isset($extraCursorHelper)) {
```
notbakaneko, line 322: always false?

app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -272,63 +274,83 @@ public function show($id)
...
+                if ($post !== null) {
+                    $params['cursor'] = ['id' => $post->getKey() - 1];
+                    $params['sort'] = 'id_asc';
+                }
+            } elseif ($params['start'] !== null) {
+                $params['cursor'] = ['id' => $params['start'] - 1];
+                $params['sort'] = 'id_asc';
+            } elseif ($params['end'] !== null) {
+                $params['cursor'] = ['id' => $params['end'] + 1];
+                $params['sort'] = 'id_desc';
```
notbakaneko, line 295: The 404 guards near the start of the function should probably be done even earlier and the `$params` building split out of this long function.

notbakaneko (changes requested):

notbakaneko (approved):
seems working 🤔

PR #7227, opened 2021-02-18
Different default posts sort based on existence of query string

### LINKED ISSUE #7226: User's post history not sorted by post time
The ``/users/{user}/posts`` page used to be sorted with the most recent posts first, which it probably should still be.

nanaya:
Resolves #7226.


notbakaneko (commented):
The user posts page seems to be missing the sort bar?

nanaya:
Because I may nuke that page soon.

notbakaneko (approved):

PR #7229, opened 2021-02-18

peppy:


resources/lang/ja/beatmapsets.php
```diff
@@ -12,7 +12,7 @@
     ],
 
     'download' => [
-        'limit_exceeded' => '',
+        'limit_exceeded' => 'スピードを落として、もっと遊ぼう。',
```
peppy, line 15: man... we need more japanese tanslators

PR #7230, opened 2021-02-18
Forum topic viewing API

### LINKED ISSUE #7235: Reduce work when loading more forum posts
`TopicsController#show` contains some things that are not needed when loading more posts like cover, polls, and featured votes. Those should be skipped accordingly.

nanaya:
Topic viewing only for now. Not sure if extra attributes are needed (cover, poll, featured votes).

And should accessing using API mark it as read?

- [x] depends on #7225
- [x] decide whether to use `search` or `params` the return data's sort info

Resolves #7235 because it's simple enough and this PR already moves things around.


app/Transformers/Forum/TopicTransformer.php
```diff
@@ -0,0 +1,33 @@
...
+namespace App\Transformers\Forum;
+
+use App\Models\Forum\Topic;
+use App\Transformers\TransformerAbstract;
+
+class TopicTransformer extends TransformerAbstract
+{
+    public function transform(Topic $topic)
+    {
+        return [
```
notbakaneko, line 15: probably should include post count as well?

notbakaneko (changes requested):
Could use some alphabetical key ordering.
Also seems strange this API returns 404 error when querying with a cursor past the end of the posts when others return empty array + other related data, especially considering the response includes `topic` and  this the the `topic` endpoint. 🤔

notbakaneko (commented):
I wonder if `search` or `query` works better given the context, `params` seems too generic

notbakaneko (approved):

PR #7238, opened 2021-02-19
Allow creating and editing forum post and topic through API

### LINKED ISSUE #6072: Add API endpoints for viewing and posting on the forum
I'm currently scraping the normal website to automate things for project loved https://osu.ppy.sh/community/forums/120 and I think abraker does the same to get statistics about Off-Topic (?)

would be nice if this could be done via API.

nanaya:
Resolves #6072. This adds a new oauth scope `forum.write`.

- [x] depends on #7230

(should be noted that no moderation permissions are granted)


app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -415,6 +432,32 @@ public function show($id)
...
+     * Create a new topic.
+     *
+     * ---
+     *
+     * ### Response Format
+     *
+     * Field  | Type                       | Includes
+     * ------ | -------------------------- | --------
+     * topic  | [ForumTopic](#forum-topic) | |
+     * post   | [ForumPost](#forum-post)   | body
```
notbakaneko, line 448: I missed it in the previous one, but these should match, either `[ForumTopic](#forumtopic)` or `[Forum Topic](#forum-topic)`.

notbakaneko (approved):

PR #7239, opened 2021-02-19
Hotfix for restricted/deleted user in discussion listing

notbakaneko:
Discussion listing didn't include hidden users when viewed by a moderator (but discussion on beatmap page still shows restricted users for everyone?).

Includes some hotfixes for handling user missing from the props - this is all in the process of being refactored away.


resources/assets/coffee/react/beatmap-discussions/discussion.coffee
```diff
@@ -167,7 +168,7 @@ export class Discussion extends React.PureComponent
 
     topClasses = "#{vbn} #{vbn}--#{type}"
     topClasses += " #{vbn}--inactive" if score != 0
-    user = @props.users[@props.discussion.user_id]
+    user = @props.users[@props.discussion.user_id] ? @props.users[null]
```
nanaya, line 171: at least around line 290 of this file also need this

https://github.com/ppy/osu-web/blob/f4dd0e1a74635ea61acb6e2263ecfa7264a6beae/resources/assets/coffee/react/beatmap-discussions/discussion.coffee#L290-L291

Then maybe here (and there's random `currentUser.id?` boolean comparison against number)

https://github.com/ppy/osu-web/blob/f4dd0e1a74635ea61acb6e2263ecfa7264a6beae/resources/assets/coffee/react/modding-profile/posts.coffee#L55-L56

nanaya (changes requested):

nanaya (approved):

PR #7245, opened 2021-02-22
Remove redundant rank data

### LINKED ISSUE #7240: Remove redundant user global rank attributes
- `UserStatistics#pp_rank`
- `UserStatistics#rank.global`
  - used in profile page (and maybe somewhere else as well?)
  - move country rank outside as well
- `UserCompact#current_mode_rank`
  - used in user listing (friends, groups)
  - include full statistics instead...?

nanaya:
Resolves #7240. `rank.country` will be removed later once everything is moved to use `country_rank`.

I'm not sure the current usage state in osu though :thinking: This assumes they moved to use `global_rank` already.

### DIFF
--- a/app/Http/Controllers/FriendsController.php
+++ b/app/Http/Controllers/FriendsController.php
@@ -8,6 +8,7 @@ namespace App\Http\Controllers;
 use App\Jobs\UpdateUserFollowerCountCache;
 use App\Models\User;
 use App\Models\UserRelation;
+use App\Transformers\UserCompactTransformer;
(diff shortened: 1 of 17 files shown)


frenzibyte:
Yes, osu!lazer has moved to always use `global_rank` instead of `ranks.global`.

notbakaneko (changes requested):
maybe mention the removal in `Breaking Changes` section 🤔

notbakaneko (approved):

PR #7253, opened 2021-02-24
Refactor database cursor helper

### LINKED ISSUE #7233: Easier Database Cursor Helper Usage
A few recent observations on what can be improved about that helper to allow simpler usage of it.

- migrate remaining usage of `cursorWhere` to `cursorSort` (or whatever the simplified function equivalent is)
- allow generation of the helper direct from model (with trait perhaps)
- allow skipping manual helper creation (by accepting both string and helper for the `cursorSort` scope)
- remove intermediate cursor format (the one generated by `DbCursorSort#prepare`) and only use it internally
  - default format should be just plain `array` (the one accepted by `#prepare` function itself)
  - that should also allow removal of `#prepareNext` function

nanaya:
Resolves #7233.

RIP `cursorWhere` <del>(and my forum update PRs)</del>.


notbakaneko (changes requested):

notbakaneko (commented):
`scopeCursorSort` accepting `string` or helper and `array` or `Model` seems weird 😔 .
Seems like it's expected behaviour really should be documented or params converted before passing in.

(would be better if we could return the helper from `scope` function, though 😔 )

notbakaneko (approved):

PR #7266, opened 2021-02-26
Add modding discussion listing endpoints to API

### LINKED ISSUE #4042: Expose all discussion, reply and vote events
The /discussions, /votes-received, /votes-given and /posts routes are only available through the /user route, whereas it would be beneficial seeing an overview of all activity through /beatmapsets for these in the same way the /events route works.

The following /user routes currently exist:
- https://osu.ppy.sh/users/{user}/modding/events
- https://osu.ppy.sh/users/{user}/modding/discussions
- https://osu.ppy.sh/users/{user}/modding/votes-received
- https://osu.ppy.sh/users/{user}/modding/votes-given
- https://osu.ppy.sh/users/{user}/modding/posts

However, of all the following /beatmapset routes, only the first exists:
- ~~https://osu.ppy.sh/beatmapsets/events~~
- ~~https://osu.ppy.sh/beatmapsets/discussions~~ (edit: now exists as /beatmap-discussions)
- ~~https://osu.ppy.sh/beatmapsets/votes-received~~
- ~~https://osu.ppy.sh/beatmapsets/votes-given~~ (edit: all votes are /beatmapsets/discussions/votes)
- ~~https://osu.ppy.sh/beatmapsets/posts~~ (edit: now exists as /beatmap-discussion-posts)

Adding the last four of the above list would not only allow the QAT to better be able to keep track of beatmapset activity for purposes such as moderation (it's difficult moderating the upvote system without a proper overview, for instance), but also the overall community to better involve themselves wherever things are happening (whether that be through actually going to the link or by being kept up to date of anything happening relevant to them through a discord bot or similar). Overall I could see this opening up a lot of possibilities for community members to contribute that would otherwise be difficult to work with normally.

2020 Edit: These should ideally be api endpoints for the purposes I mentioned above, see https://github.com/ppy/osu-web/issues/5842#issuecomment-606377139.

notbakaneko:
This was originally going to be part of a larger refactor moving all the listing pages to React and have consistent data structures, however, that's turning out to be quite a bit more involved than expected, so, just the API part is being PR'ed first with less complete data structures.

I'm not sure how useful including extra data is from a API-only consumer point of view, though, as you could want multiple levels of data, or none at all 🤔 

Also includes a change to the visibility of discussions by restricted users in the listings to be more inline with the behaviour on the beatmap page itself.

`sort` param changed to use `_` instead of `-` as the delimiter to be in line with other endpoints.

closes #4042


nanaya (changes requested):

app/Libraries/BeatmapsetDiscussionPostsBundle.php
```diff
@@ -0,0 +1,90 @@
...
+            $beatmapsetIds = $this->getPosts()->pluck('beatmapset.beatmapset_id')->unique()->values();
+
+            return Beatmapset::whereIn('beatmapset_id', $beatmapsetIds)->get();
+        });
+    }
+
+    private function getPosts()
+    {
+        return $this->memoize(__FUNCTION__, function () {
+            $this->search = BeatmapDiscussionPost::search($this->params);
```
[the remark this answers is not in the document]
notbakaneko: That'd still be weird because `getSearch` would be including a query builder that could be modified before `getPosts` actually calls it, instead of returning the just search that `getPosts()` used.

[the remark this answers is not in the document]
notbakaneko: There doesn't seem to be a need for query to be returned though 🤔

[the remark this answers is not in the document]
nanaya: looks like a better solution is to split off the search function to params cleaner and query builder

app/Libraries/BeatmapsetDiscussionPostsBundle.php
```diff
@@ -0,0 +1,80 @@
...
+            'cursor' => $this->getCursor(),
+            'posts' => json_collection($this->getPosts(), new BeatmapDiscussionPostTransformer()),
+            'users' => json_collection($this->getUsers(), new UserCompactTransformer()),
+        ];
+    }
+
+    private function getBeatmapsets()
+    {
+        return $this->memoize(__FUNCTION__, function () {
+            return $this->getPosts()->pluck('beatmapDiscussion.beatmapset')->uniqueStrict('id')->values();
```
nanaya, line 37: Preload for this seems missing. Or maybe it should be `->pluck('beatmapset')` instead?

Also there's nothing linking post and beatmapset with current json structure :upside_down_face:

resources/docs/source/includes/_structures/beatmapset_discussion_post.md
```diff
@@ -0,0 +1,16 @@
+## BeatmapsetDiscussionPost
+
+Represents a post in a [BeatmapsetDiscussion](#beatmapsetdiscussion).
+
+Field                 | Type                     | Description
+--------------------- | ------------------------ | -----------
+beatmap_discussion_id | number                   | |
```
nanaya, line 7: this should be renamed eventually...

nanaya (changes requested):

resources/docs/source/includes/_structures/beatmapset_discussion_post.md
notbakaneko: wonder if we should just duplicate a `beatmapset_discussion_id` field on everything

nanaya: :thinking: is it even currently used for anything

nanaya (approved):

PR #7272, opened 2021-03-02
Remove skype

### LINKED ISSUE #7259: Remove skype from user profiles
Not needed in this day and age.

tzx:
Resolves #7259

I removed some of the translation stuff, but I'm not sure if that's okay. I don't know anything about laravel.


peppy:
I believe you should only have to remove the localisation string for english; the rest will automatically be removed on next crowdin sync.

tzx:
> I believe you should only have to remove the localisation string for english; the rest will automatically be removed on next crowdin sync.

I see. I shouldn't need to add them back right?

peppy:
Probably fine.

notbakaneko (changes requested):
These should be removed as well:
https://github.com/ppy/osu-web/blob/b9bbb4b/app/Http/Controllers/AccountController.php#L128
https://github.com/ppy/osu-web/blob/b9bbb4b/app/Models/User.php#L217
https://github.com/ppy/osu-web/blame/b9bbb4b/resources/docs/source/includes/_structures/user.md#L45
https://github.com/ppy/osu-web/blame/b9bbb4b/resources/docs/source/includes/_structures/user.md#L183

tzx:
There are still instances of `user_msnm` in the `@property` of some Models. Are the models' data just pulled from a database, meaning I shouldn't remove these right?

There's also the `user_msnm` localization in other languages but I removed it for english, so I believe they should be removed for others automatically.

notbakaneko:
> There are still instances of user_msnm in the @property of some Models
Those are fine to leave in for now

notbakaneko (approved):

Issue #6805, opened 2020-10-23
Remove web chat update polling


notbakaneko:
Chat message and channel updates currently happen by polling `/chat/updates`; the overhead on this is fairly high even if there are no updates. These should be pushed over via the websocket server instead.

peppy:
This is mostly resolved now? @notbakaneko

notbakaneko:
Update polling has been removed #8121

Handling silences and restrictions over websockets should probably have its own issue.

PR #7281, opened 2021-03-04
Add show chat channel to API

notbakaneko:
This is going to be needed as we move towards removing chat presence polling (#6805).
Although, final response isn't finalized yet. 
Also, nothing in `presence` is going to be updated to use any of the new stuff.


resources/docs/source/includes/_structures/chat_channel_show.md
```diff
@@ -0,0 +1,53 @@
+## ChatChannelShow
```
nanaya, line 1: is it going to be used anywhere else? Seems rather weird separating it from the route itself

nanaya (changes requested):

nanaya (approved):

PR #7297, opened 2021-03-11
Move beatmap attributes to per-beatmap document

nanaya:
This will allow searching for beatmaps with specific combination of attributes.

For example, searching for beatmap with difficulty rating of 3 and drain rate of 8 using the old structure may return beatmapset which contain two different beatmaps, one with diff 3 (but maybe drain 7) and another one with drain 8 (but maybe diff 4).

This is required to match lazer behavior for advanced query search.

Removed the old attributes except playmodes because it's the field still in use by legacy web although I think it could be easily updated to use the nested one?

### DIFF
--- a/config/schemas/beatmaps.json
+++ b/config/schemas/beatmaps.json
@@ -25,39 +25,14 @@
           "dynamic": false,
           "type": "nested",
           "properties": {
-            "beatmap_id": {
-              "type": "long"
-            },
-            "convert": {
-              "type": "boolean"
-            },
-            "difficultyrating": {
-              "type": "double"
-            },
-            "playmode": {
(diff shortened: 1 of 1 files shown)


peppy:
I'm willing to update the legacy code to work with the new location. Or you can, I think you have access.

nanaya:
Apparently nested document can't be easily searched using query string syntax which legacy web uses.

peppy:
okay let's not.

notbakaneko (approved):

notbakaneko (commented):
reindex not necessary until #7300

notbakaneko (approved):

PR #7298, opened 2021-03-11
Include sort parameter in beatmapset search response

nanaya:
With advanced search, non-blank `query`/`q` parameter doesn't always default to `relevance` anymore. Duplicating the parsing logic to the frontend doesn't seem to make sense either so it's probably best to render current sort option based on server response.

Well, I tried updating the frontend part to do it but it's a bit more complicated. So here's just the response for starter.


peppy:
Seems like something we want to get merged now that advanced search is already live?

nanaya:
The frontend also need to be hooked up. And then there's semi-related bug recently found where sort option is reset when search query is updated.

app/Libraries/Search/BeatmapsetSearchRequestParams.php
```diff
@@ -274,21 +283,23 @@ private function parseQuery(): void
...
 
-        if (empty($field)) {
-            $this->sorts = $this->getDefaultSort($order);
-        } else {
-            $this->sorts = $this->normalizeSort(new Sort($field, $order));
+        if (!in_array($this->sortOrder, ['asc', 'desc'], true)) {
+            $this->sortOrder = 'desc';
         }
 
+        $this->sorts = $this->normalizeSort(new Sort($mappedField, $this->sortOrder));
```
notbakaneko, line 299: maybe make `remapSortField` non-static (and rename?) and assign `$this->sortField` there? The way it bounces between `$mappedField` and `$this->sortField` (and then `sortOrder` in the middle) reads a bit confusing.

notbakaneko (changes requested):

app/Libraries/Search/BeatmapsetSearchRequestParams.php
nanaya: doing that in the `normalizeSort` (now `setSorts`) seems like a better idea :man_dancing:

notbakaneko (approved):

PR #7306, opened 2021-03-13
More spacing between pp and time columns in beatmap leaderboards

### LINKED ISSUE #7250: More spacing between pp and time columns in beatmap leaderboards
When pp is in the 1000's the values in the two columns start to blend together. Look at 8th place

Link to map these scores are on: https://osu.ppy.sh/beatmapsets/800102#mania/1679790

reykim7124:
Minimize the column miss, and enlarge the pp column to space it between the time column

Before changes:
image

After changes:
image

Closes #7250


nanaya (approved):
This doesn't fix the underlying problem of the columns having no padding at all but I guess this fixes specifically the mentioned issue.

reykim7124:
True, maybe for a permanent fix so that this won't happen again, the table column could probably be made to autofit the content instead of a fixed width per column.

nanaya:
tables are already automatically sized regardless of `width`. It's just missing any kind of padding.

PR #7313, opened 2021-03-15
Mostly uniform page side padding

nanaya:
This is originally just for profile page so the beatmapset card can be slightly wider but then I figured may as well go all the way and remove the old gutter size. Not removing `-v2` labels yet because it's easier to check. Maybe for another PR.

Notable exceptions are

- beatmapset listing page
- forum topic (left side and reply box)
- floating header on forum topic and beatmapset search

Design adjustments:
- "+1" label for comment upvote button has been moved to above the button because otherwise it doesn't fit.
- title icon in changelog alignment [on mobile](https://s.myconan.net/2021-03/firefox_2021-03-15_19-59-07.png) is different compared to [desktop](https://s.myconan.net/2021-03/firefox_2021-03-15_19-59-43.png)
- smaller action buttons on profile page


resources/assets/less/variables.less
```diff
@@ -51,15 +51,12 @@
 @padding-big: (20px + @logo-size + 10px);
 
 @gutter-profile-page: @gutter-v2;
-@gutter-profile-page-desktop: @gutter-desktop;
+@gutter-profile-page-desktop: @gutter-v2-desktop;
```
notbakaneko, line 54: just use `@gutter-v2` and `@gutter-v2-desktop` directly? (maybe we can use css variables next time profile gutter changes again 🤔 )

notbakaneko (changes requested):

notbakaneko (approved):
eh, probably should do the changes separately

Issue #6974, opened 2020-12-11
Migrate from TSlint to ESlint


notbakaneko:
TSLint has been deprecated for a while and stopped accepting updates since Dec 1st.
Apparently, ESLint can do annotations on Github Actions via `setup-node` and since we've just moved our entire CI to Github actions #6959, now is a good time as any to look at switching

nanaya:
I'll start slowly fixing the errors one by one every now and then.

nanaya:
done in #8003

PR #7314, opened 2021-03-15
Pre-eslint migration type fixes

notbakaneko:
1 of ?

Part of migrating to use ESLint (#6974).
`typescript-eslint` also has more quite a fair bit more thorough and comprehensive checks than `tslint` does, and it turns out that quite a fair number of things need fixing up.

This fixes some of the more obvious and simpler to fix up typing issues.


nanaya (changes requested):

resources/assets/lib/utils/beatmapset-discussion-helper.ts
```diff
@@ -29,7 +29,7 @@ export function propsFromHref(href: string) {
 
   if (targetUrl.host === window.location.host) {
     const target = BeatmapDiscussionHelper.urlParse(targetUrl.href, null, { forceDiscussionId: true });
-    if (target?.discussionId != null) {
+    if (target?.discussionId != null && target?.beatmapsetId != null) {
```
nanaya, line 32: the typing for the function being wrong aside, the second `?` is not needed because it's already guaranteeed by the first one?

nanaya (changes requested):

nanaya (approved):

PR #7315, opened 2021-03-15
Cache some user related lookup operations

nanaya:
Found this when adding favourite button on card. Also updated other lookups while at it.

Bonus point if the servers are updated to have the native extension installed instead of using polyfill (`php7.x-ds`) (native is automatically prioritised over polyfill if installed).

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -22,6 +22,7 @@ use App\Traits\Validatable;
 use Cache;
 use Carbon\Carbon;
 use DB;
+use Ds\Set;
 use Egulias\EmailValidator\EmailValidator;
 use Egulias\EmailValidator\Validation\NoRFCWarningsValidation;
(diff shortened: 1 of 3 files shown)


app/Models/User.php
```diff
@@ -1523,17 +1524,29 @@ public function nominationModes()
 
     public function hasBlocked(self $user)
     {
-        return $this->blocks->where('user_id', $user->user_id)->count() > 0;
+        $blockIds = $this->memoize(__FUNCTION__, function () {
```
notbakaneko, line 1527: I feel like these should be memoized as `__FUNCTION__` 🤔

nanaya: still close enough and saves me thinking of name :shrug:

notbakaneko (approved):

PR #7322, opened 2021-03-16
Return beatmap checksum in API response by default

### LINKED ISSUE #6777: Including checksum of Beatmap object on api v2
[this code hides the 'checksum' attribute](https://github.com/ppy/osu-web/blob/2d5561aca34b3ad71f7524e3a4a5633b3b6863a3/app/Models/Beatmap.php#L62).
I think if this attribute is included in Beatmap object from api v2, services such as beatmap mirrors will be able to check for updates without using api v1.

LiquidPL:
Fixes #6777.


app/Transformers/BeatmapCompactTransformer.php
```diff
@@ -17,6 +17,10 @@ class BeatmapCompactTransformer extends TransformerAbstract
         'max_combo',
     ];
 
+    protected $defaultIncludes = [
```
LiquidPL, line 20: One thing I'm not sure about is whether this should be here or in `BeatmapTransformer`.

ilsubyeega: im not very sure but if this compacted beatmap should have checksum as default, i think it should be in transform function (and remove related code)

LiquidPL: Good point, I've left it here for now so it doesn't break any instances that depend on the `checksum` include, but if we actually decide on keeping it here it should be removed, yes.

notbakaneko (changes requested):
given the supposed use-case for this, it should be in `BeatmapTransformer` instead

omkelderman:
in the api documentation is the checksum field mentioned as optional attribute of `BeatmapCompact`: https://osu.ppy.sh/docs/index.html#beatmapcompact. I dont know exactly how the Transformer logic works but to me it looks like the most logical place would be inside the `BeatmapCompactTransformer` where it is right now?

LiquidPL:
It's optional in `BeatmapCompactTransfomer`, but now we'll make it default in `BeatmapTransformer`.

notbakaneko (approved):

PR #7331, opened 2021-03-18
Fix notification not sent in some cases

### LINKED ISSUE #7325: Problem / DQ notifications rarely, if ever work on the website
I confirmed this with a significant number of people. They may have notifications enabled for their mode, or for every mode. In each case, many of them have not received any notifications for DQs or problems on qualified maps via the osu website in months.  Some receive notifications only sporadically.

For instance, I have all modes enabled for reports, and I have osu! enabled for disqualification notifications

image

Here is my notifications for beatmaps up to a few days ago:
image

Noted in green are report notifications.

But it's missing a LOT of things
By comparing to a #mapfeed history hosted by bots in discord servers and double checking the map discussion histories to be sure, there have been **24** disqualification events for osu! maps in the past 5 days, of which **0** showed up in my notifications. 
Similar goes for the problem notifications, only a few of many many more actually notified me. This is a serious issue as it means my reliable reporting tools are only available off platform, and the on-platform ones are broken.

For example https://osu.ppy.sh/beatmapsets/1228487/discussion/2822288/general#/2218546  https://osu.ppy.sh/beatmapsets/1395277/discussion/2880013/general#/2218508 these never notified me.

**Conclusion:**

Something is broken with them, and they aren't working as intended.

nanaya:
Mainly for "global" type notifications. Avoid doing large `whereIn` query.

Resolves #7325.

### DIFF
--- a/app/Jobs/Notifications/BroadcastNotificationBase.php
+++ b/app/Jobs/Notifications/BroadcastNotificationBase.php
@@ -9,6 +9,7 @@ use App\Events\NewPrivateNotificationEvent;
 use App\Exceptions\InvalidNotificationException;
 use App\Models\Notification;
 use App\Models\User;
+use App\Models\UserGroup;
 use App\Models\UserNotification;
 use App\Models\UserNotificationOption;
 use App\Traits\NotificationQueue;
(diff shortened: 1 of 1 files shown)


app/Jobs/Notifications/BroadcastNotificationBase.php
```diff
@@ -80,39 +81,51 @@ private static function applyDeliverySettings(array $userIds)
...
+        $allBotUserIds = UserGroup
+            ::where('group_id', $botGroupId)
+            ->select('user_id');
+
+        // only consider users with bot as their primary group
         $botUserIds = User
-            ::whereIn('user_id', $userIds)
-            ->where('group_id', app('groups')->byIdentifier('bot')->getKey())
+            ::where('group_id', $botGroupId)
+            ->whereIn('user_id', $allBotUserIds)
```
notbakaneko, line 128: This part seems unnecessary - if going by the primary group, the `User::isBot` check is just `$this->group_id === app('groups')->byIdentifier('bot')->getKey();`
The old query was just limiting the result set to the list of users given instead of selecting all bot users.

I wonder if the list of bot users is something worth caching? 🤔

notbakaneko (changes requested):

app/Jobs/Notifications/BroadcastNotificationBase.php
nanaya: yeah there's no index on `group_id` on users table. The query should be simple enough (primary key lookup on two tables) and the amount of data is minimal so it should be fine doing this query (or at least not much worse compared to the current one) :shrug:

notbakaneko (approved):

PR #7334, opened 2021-03-19
Split out discord endpoints for user reports into two categories

peppy:
This was a request from the moderation team, who don't deal with cheating reports but have them clogging up the content moderation channel.

Requires `.env` update.


nanaya (approved):

TicClick:
@peppy as **no** reports reach the respective Discord channel now (doesn't matter if the reason is cheating or not), this seems to be deployed. could you please check if everything is okay and the destination for non-cheating reports is what it was before?

PR #7354, opened 2021-03-22
Remove old wiki index on database reset

nanaya:
I ended up with loads of old wiki index after a while. The other index already cleans up itself.

I wonder if it's because the new index is empty but anyone who cares about the content can reindex the thing afterwards or should just not run "fresh all" command.


peppy (approved):
seems logical

PR #7355, opened 2021-03-22
Improve matching for artist and mapper search

### LINKED ISSUE #7333: Artist and creator advanced search improvements
It currently works in a rather limited way. Here are list of things need to be added:

- prefix and keyword index like title for `artist`, `artist_unicode`, and `creator` fields (I noticed `title_unicode` is also missing those index)
- `artist_unicode` needs to be counted in for artist search (similar to general search)
- both artist and creator search should allow term/exact match

(and maybe more things I missed/forgot)

nanaya:
Resolves #7333.


app/Libraries/Search/BeatmapsetSearch.php
```diff
@@ -33,7 +33,19 @@ public function __construct(?BeatmapsetSearchParams $params = null)
...
     {
-        static $partialMatchFields = ['artist', 'artist.*', 'artist_unicode', 'creator', 'title', 'title.raw', 'title.*', 'title_unicode', 'tags^0.5'];
+        static $partialMatchFields = [
+            'artist',
+            'artist.*',
+            'artist_unicode',
+            'artist_unicode.*',
+            'creator',
+            'title',
+            'title.raw',
```
nanaya, line 43: not sure if useful considering there's also `title.*`

notbakaneko (changes requested):
Is the query `artist:"the best"` supposed to search for 
- `the best`
- `the` or `best` or `the best`

app/Libraries/Search/BeatmapsetSearch.php
notbakaneko: Should be fine to remove 🤔

peppy:
I would hope it searches for `the best` only.

notbakaneko:
I forgot to mention but should `artist:"the best"` match `the best one`?

peppy:
Yep

notbakaneko (approved):

PR #7360, opened 2021-03-23
More pre-eslint updates

notbakaneko:
part 2 of ???

Follow up to #7314

This is largely `eslint --fix` + adding equivalent eslint ignore rules.
`EditorToolbar.renderButton` refactored out to remove `tslint:disable-next-line:jsx-no-lambda` (because you can't have eslint and tslint disable next line together as they do exactly what they say)


nanaya:
conflicts :pensive:

resources/assets/lib/beatmap-discussions/editor-toolbar-button.tsx
```diff
@@ -0,0 +1,28 @@
...
+interface Props {
+  format: string;
+}
+
+export const EditorToolbarButton = (props: Props) => {
+  const context = React.useContext(SlateContext);
+  const handleClick = React.useCallback((event: React.SyntheticEvent<HTMLButtonElement>) => {
+    event.preventDefault();
+    toggleFormat(context, props.format);
+  }, []);
```
nanaya, line 17: shouldn't this depend on `props.format`? And maybe `context`? I wonder if it's better to just move it back and read from `dataset` instead.

resources/assets/lib/osu-audio/main.ts
```diff
@@ -85,7 +85,7 @@ export default class Main {
 
   audio = new Audio();
   private currentSlider?: Slider;
-  private durationFormatted: string = `'0:00'`;
+  private durationFormatted = '\'0:00\'';
```
nanaya, line 88: ಠ_ಠ

double quotes maybe?

nanaya (changes requested):

nanaya (approved):

PR #7367, opened 2021-03-24
Show pp even if 0

### LINKED ISSUE #7308: Certain scores incorrectly displaying pp values as "processing"
Currently, mania scores in some cases can be worth exactly 0pp. I believe this is causing them to indefinitely show their pp value as "still being calculated" in the recent plays section of user profiles, score pages, beatmap listings, etc.

A bunch of example scores affected by this can be found on [this map](https://osu.ppy.sh/beatmapsets/1071024#mania/2241878) with the "No Fail" mod selected (though it should be noted this can occur without No Fail as well).

notbakaneko:
closes #7308

..unless `0 pp` has some other special meaning as well


peppy:
Zero should not have any meaning apart from "not much pp" aka rounded down to zero.

nanaya (approved):

PR #7369, opened 2021-03-24
Add option to disable client check

nanaya:
Resolves #7349.

- [x] depends on #7368

PR #7371, opened 2021-03-24
Add basic score submission

nanaya:
Not much check is done. Similar to the one in multiplayer except `ruleset_id` parameter is also required for initial creation.

- [x] depends on #7369 (branched off wrong branch whoops)

Endpoints:
- `POST api/v2/beatmaps/{beatmap}/solo/scores`
- `PUT api/v2/beatmaps/{beatmap}/solo/scores/{score}`


app/Transformers/Solo/ScoreTransformer.php
```diff
@@ -0,0 +1,31 @@
...
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace App\Transformers\Solo;
+
+use App\Models\Solo\Score;
+use App\Transformers\TransformerAbstract;
+
+class ScoreTransformer extends TransformerAbstract
```
ilsubyeega, line 11: can you mind adding the `user` attribute? I think the client-side don't have user_id property atm
and possibly required to get scores by API v2 for future

+) suggestion: add `beatmap` attribute as availableIncludes (for getting scores list from single user etc)

peppy: Are you sure you're in the right place? The client already knows the user and beatmap because that's what it used to... submit the score, no?

ilsubyeega: I thought the Solo/ScoreTransformer to fetch additional score lists, But it could be fetched by Normal ScoreTransformer one (modifying some values). sorry

app/Libraries/ScoreCheck.php
```diff
@@ -0,0 +1,37 @@
...
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace App\Libraries;
+
+use App\Exceptions\InvariantException;
+
+class ScoreCheck
+{
+    public static function assertCompleted($score)
```
notbakaneko, line 12: `$score` could do with an interface

app/Http/Controllers/Solo/ScoresController.php
```diff
@@ -0,0 +1,74 @@
...
+        ]);
+        $params['ended_at'] = now();
+        $params['mods'] = Mod::parseInputArray($params['mods'] ?? [], $score->ruleset_id);
+
+        $score->complete($params);
+
+        return json_item($score, 'Solo\Score');
+    }
+
+    private function getBeatmapOrFail($beatmapId)
```
notbakaneko, line 70: `static`?

notbakaneko (changes requested):

app/Libraries/ScoreCheck.php
nanaya: how do one interface eloquent attributes :thinking:

notbakaneko: I was thinking of just an `interface` with common `@property` in the doc block for typings 🤔

notbakaneko (approved):

ilsubyeega:
investigating how to play this on osu!lazer, is this intended to get url like this?
`https://osu.ppy.sh/api/v2/beatmaps/1831220/solo/scores`

peppy:
No.

ilsubyeega:
> No.

https://github.com/ppy/osu-web/blob/master/tests/api_routes.json#L50 says submission URL is this line, plus I had some problems about investigating, I would like to create new issue

peppy:
Have your fun elsewhere, please.

Issue #7373, opened 2021-03-24
Undocumented fields returned by the api


tybug:
Several objects have fields returned by the api which aren't mentioned in the docs:

### BeatmapsetCompact

* nsfw

### CommentBundle

* cursor

### User

* title_url

peppy:
Would be great if anyone willing to help out with documentation could chip in. These should be very easy to add even for a non-laravel-developer.

PR #7376, opened 2021-03-25
Add unmentioned fields to docs

knightss27:
Fixes issue #7373.

Should these fields also be defined inside of the ts interface definitions in `/resources/lib`? I mention this as you can see these fields being constructed in the relevant php files, but the ts definitions for both CommentBundleJson and UserJson are missing the fields mentioned here, but I have no clue whether that is purposeful or not.


peppy (approved):

peppy (dismissed):

resources/docs/source/includes/_structures/comment_bundle.md
```diff
@@ -91,6 +91,7 @@ Field             | Type                                  | Description
 ----------------- | ------------------------------------- | --------------------------------------------------------------
 commentable_meta  | [CommentableMeta](#commentablemeta)[] | ID of the object the comment is attached to
 comments          | [Comment](#comment)[]                 | Array of comments ordered according to `sort`;
+cursor            | [Cursor](#cursor)[]                   |
```
notbakaneko, line 94:
```suggestion
cursor            | [Cursor](#cursor)                     |
```

cursor should just be a `Cursor`

notbakaneko (changes requested):

notbakaneko (approved):

Issue #7343, opened 2021-03-20
Reset button in me! should probably have a confirmation + be coloured red


Purplegaze:
I'm always afraid I'm going to click this instead of "Preview" whenever I'm trying to make big changes to my userpage. Considering it clears all your changes without giving you a chance to undo, I think it should visually look different and have some sort of confirmation once you press it (or be removed entirely, because just refreshing the page and opening up me! again has the same result)

![](https://cdn.discordapp.com/attachments/247078580381679627/822759042665283584/unknown.png)

peppy:
I think just removing it is fine. If a user wants to start over, they can hit the "cancel" button.

nanaya:
Removed.

PR #7377, opened 2021-03-25
Remove reset button from bbcode editor

knightss27:
Per #7343. I think this should work for simply removing the button from the UI. Unless I'm looking in the wrong place, I can't seem to find any code where the `@sendOnChange` with `type: 'reset'` is actually being handled as in `user-page-editor.coffee` the switch for onChange only handles cancel and save events. If that's not the case then there may be more reset code that needs removing.


nanaya:
I wonder if it's better to show confirmation dialog for reset button instead.

peppy:
Honestly I'd just let the user hit cancel then edit again. It's such an uncommon flow that I don't think the button is needed.

nanaya (approved):

PR #7378, opened 2021-03-26
Add callback support for Ajax errors

nanaya:
Still needs some more test but seems fine from quick check.

This will simplify handling verification errors and any jquery ajax errors in general.

Full replacement for existing `ajaxError` and `emitAjaxError` and some other weirdness will be done later™. Some are done here for example.


resources/assets/lib/utils/html.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+export function createClickCallback(target: unknown) {
```
nanaya, line 4: this function name and the filename itself is questionable :eyes:

notbakaneko (approved):

PR #7388, opened 2021-03-28
Fix null artist_unicode and title_unicode fields

### LINKED ISSUE #7380: artist_unicode and title_unicode are null on old beatmaps
I did `JSON.parse(document.getElementById('json-beatmapset').textContent)` in browser's developer console, and I just noticed that the artist_unicode and title_unicode are null, while other (newer) beatmaps has the value (usually same value as non-unicode property)
Beatmap URL: https://osu.ppy.sh/beatmapsets/34348#osu/111680
I think it's just because it's an old map (submitted in 2011), but I think it shouldn't be null and should show the same value as a non-Unicode one instead (just like any other maps).

For reference: https://osu.ppy.sh/beatmapsets/1348446#osu/2792231
This map has non-null artist_unicode and title_unicode.

notbakaneko:
closes #7380

Automatically fallback to `artist` and `title`.

Wonder if they should be named something else 🤔 
(yes the function order is weird)

### DIFF
--- a/app/Models/Beatmapset.php
+++ b/app/Models/Beatmapset.php
@@ -1289,7 +1289,7 @@ class Beatmapset extends Model implements AfterCommit, Commentable, Indexable
     {
         $profileCustomization = $user->userProfileCustomization ?? new UserProfileCustomization();
         if ($profileCustomization->beatmapset_title_show_original) {
-            return presence($this->title_unicode) ?? $this->title;
+            return $this->title_unicode;
         }
 
         return $this->title;
@@ -1304,6 +1304,11 @@ class Beatmapset extends Model implements AfterCommit, Commentable, Indexable
         }
 
         return Forum\Post::find($topic->topic_first_post_id);
+    }
+
+    public function getTitleUnicodeAttribute()
+    {
+        return presence($this->attributes['title_unicode']) ?? $this->title;
     }
 
     public function freshHype()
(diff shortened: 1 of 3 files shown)


app/Models/Beatmapset.php
```diff
@@ -1275,11 +1275,16 @@ private function getBBCode()
         return new BBCodeFromDB($description, $post->bbcode_uid, $options);
     }
 
+    public function getArtistUnicodeAttribute()
+    {
+        return $this->attributes['artist_unicode'] ?? $this->artist;
```
nanaya, line 1280: Also for the other one

```suggestion
    public function getArtistUnicodeAttribute($value)
    {
        return $value ?? $this->artist;
```

resources/assets/lib/beatmapsets/beatmapset-json.ts
```diff
@@ -93,6 +93,6 @@ export interface BeatmapsetJson {
   nsfw: boolean;
   status: BeatmapsetStatus;
   title: string;
-  title_unicode: string | null;
+  title_unicode: string;
```
nanaya, line 96: should update `artist_unicode` one as well?

nanaya (changes requested):

resources/assets/lib/beatmapsets/beatmapset-json.ts
notbakaneko: huh, I thought I updated it 🤔

nanaya (approved):

PR #7397, opened 2021-03-31
Fix display of discussion post editor name

### LINKED ISSUE #7345: Deleted modding posts say edited [deleted user] (Bug?)
It's really confusing and weird, seems like a bug

image
this post was by Jerry, deleted by Jerry, but says it was also edited by [deleted user] and as a moderator I have no idea what that is supposed to mean. It seems to be showing on any post deleted by the poster.

nanaya:
Deleted user object no longer has null id.

Resolves #7345.


notbakaneko (changes requested):
https://github.com/ppy/osu-web/blob/ea710d7b3beeee648c4a18f83d60f6e6f5b155e6/resources/assets/coffee/react/beatmap-discussions/discussion.coffee#L290
This looks like a mistake and should probably be
```coffee
lastEditor: @props.users[post.last_editor_id] ? @props.users[null] if post.last_editor_id?
```

notbakaneko (approved):

PR #7405, opened 2021-04-02
Cleanup observers when unmounting components

notbakaneko:
TIL `disposeOnUnmount` is a thing in mobx.
Only works on react components, though...

### DIFF
--- a/resources/assets/lib/beatmaps/main.tsx
+++ b/resources/assets/lib/beatmaps/main.tsx
@@ -6,7 +6,7 @@ import AvailableFilters from 'beatmaps/available-filters';
 import HeaderV4 from 'header-v4';
 import { isEqual } from 'lodash';
 import { IValueDidChange, Lambda, observe } from 'mobx';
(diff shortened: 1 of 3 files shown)


resources/assets/lib/beatmaps/main.tsx
```diff
@@ -27,11 +27,11 @@ export class Main extends React.Component<Props> {
   constructor(props: Props) {
     super(props);
 
-    this.observerDisposers.push(observe(controller, 'searchStatus', this.searchStatusErrorHandler));
+    disposeOnUnmount(this, observe(controller, 'searchStatus', this.searchStatusErrorHandler));
   }
 
   componentDidMount() {
-    this.observerDisposers.push(observe(controller, 'searchStatus', this.scrollPositionHandler));
```
nanaya, line 34: `observerDisposers` should be removed?

nanaya (changes requested):

nanaya (approved):

PR #7407, opened 2021-04-02
Fix usage of Record types

nanaya:
As I recently learned that `Record` with `string` or `number` as key results in wrong typing:

```typescript
const map: Record<string, string> = {};

// this is valid according to typescript (up to 4.2.3)
const entry: string = map.x;
```

In such cases the correct typing should be `Partial<Record<string, string>>`.

This PR also adds typing for `UserPreferences`.


resources/assets/lib/notifications/worker.ts
```diff
@@ -39,9 +39,9 @@ export default class Worker implements DispatchListener {
   @observable waitingVerification = false;
 
   @observable private firstLoadedAt?: Date;
-  private timeout: Record<string, number> = {};
-  private xhr: Record<string, JQueryXHR> = {};
-  private xhrLoadingState: Record<string, boolean> = {};
+  private timeout: Partial<Record<string, number>> = {};
+  private xhr: Partial<Record<string, JQueryXHR>> = {};
+  private xhrLoadingState: Partial<Record<string, boolean>> = {};
```
notbakaneko, line 44: These should should probably have explicit keys listed instead of being `Partial`. Or changed since the only thing in them seems to be `loadMore` 🤔

notbakaneko (changes requested):

resources/assets/lib/notifications/worker.ts
nanaya: Even with explicit keys those will still be partial (or explicit allow undefined/null) otherwise they must be initialized with dummy value (which seems worse).

notbakaneko (approved):
`notifications/worker` really shouldn't be using `Record` for state but I might be updating some of it as part of a chat update, so that can stay for now.

PR #7418, opened 2021-04-05
Eager load beatmap for beatmapset discussion listing

notbakaneko:
`BeatmapDiscussionShow` permission calls `BeatmapShow` which needs beatmap.

Model cache sure would be useful here since a superset of the beatmaps is going to be loaded elsewhere...


nanaya (changes requested):
There's another place missing eager load:

https://github.com/ppy/osu-web/blob/fde8d24b7060fd246756bce069fc6f3214346b57/app/Libraries/BeatmapsetDiscussionsBundle.php#L57

Should include `beatmapset` for when it's trying to authorize showing deleted beatmaps in transformer.

nanaya (approved):

PR #7420, opened 2021-04-05
Update Beatmapset Json interface

nanaya:
Only adding those used by the new beatmapset card :eyes:  (or probably more)

### DIFF
--- a/resources/assets/lib/interfaces/beatmapset-extended-json.ts
+++ b/resources/assets/lib/interfaces/beatmapset-extended-json.ts
@@ -13,7 +13,7 @@ interface NominationsSummaryInterface {
   required: number;
 }
 
-export interface BeatmapsetExtendedJson extends BeatmapsetJson {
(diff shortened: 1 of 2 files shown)


resources/assets/lib/interfaces/beatmapset-extended-json.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import { BeatmapsetJson } from 'beatmapsets/beatmapset-json';
+
+interface AvailabilityInterface {
+  download_disabled: boolean;
+  download_disabled_url: string;
```
notbakaneko, line 8: this is `more_information` in the transformer ?_?

notbakaneko (changes requested):

notbakaneko (approved):

PR #7422, opened 2021-04-05
New beatmapset card

nanaya:
Depends on:
- [x] new audio player attribute #7421
- [x] updated beatmapset json interface #7420 
- [x] css-controlled circular progress #7419
- [x] typescript `formatNumberSuffixed` #7417

[Preview](https://osunext.dev.myconan.net/beatmapsets/)


Phippe:
I think clicking _anywhere_ on the bottom bar to see the difficulties is more intuitive, at least on mobile:

image

notbakaneko:
conflicts 😔

resources/assets/lib/utils/beatmapset-helper.ts
```diff
@@ -15,3 +22,32 @@ export function nominationsCount(nominations: BeatmapsetNominationsInterface, ty
...
+  .fail((xhr: JQuery.jqXHR, status: string) => {
+    // undo faked change
+    beatmapset.has_favourited = !add;
+    beatmapset.favourite_count += add ? -1 : 1;
+
+    error(xhr, status, retryCallback);
+  })
+  .done((data: FavouriteResponse) => {
+    beatmapset.favourite_count = data.favourite_count;
+  });
```
notbakaneko, line 52: This indent seems fine but eslint is going to complain about it (and the rule itself has many bugs with constructs like this); the easiest way around it seems to be to put the call on the same line like `}).fail` instead.

resources/assets/lib/utils/beatmapset-helper.ts
```diff
@@ -15,3 +22,32 @@ export function nominationsCount(nominations: BeatmapsetNominationsInterface, ty
...
+
+  if (core.userLogin.showIfGuest(retryCallback)) return;
+
+  const add = !beatmapset.has_favourited;
+
+  // fake immediate change
+  beatmapset.has_favourited = add;
+  beatmapset.favourite_count += add ? 1 : -1;
+
+  $.ajax(route('beatmapsets.favourites.store', { beatmapset: beatmapset.id }), {
```
notbakaneko, line 37: should be prefixed with `void` to mark the promise as being unhandled apparently https://github.com/typescript-eslint/typescript-eslint/blob/v4.17.0/packages/eslint-plugin/docs/rules/no-floating-promises.md

notbakaneko (changes requested):
function assignments to class properties should have the semicolon at the end; unfortunately there's no option to turn that off https://github.com/typescript-eslint/typescript-eslint/issues/2633

notbakaneko (approved):

PR #7425, opened 2021-04-06
Add ESLint config

notbakaneko:
Add config for eslint so we can at least start using it without jumping about branches and not have to keep updating with new merges.

There's still a lot of problems to go through, so tslint is still the default for now.


nanaya (changes requested):
Mostly already mentioned somewhere else.

nanaya:
conflicts :pensive:

.eslintrc.js
```diff
@@ -0,0 +1,263 @@
...
+        },
+      ],
+      "@typescript-eslint/semi": [
+        "error",
+        "always",
+      ],
+      "@typescript-eslint/type-annotation-spacing": "error",
+      "@typescript-eslint/unbound-method": "warn", // TODO: some calls are intentionally unbounded...
+      "@typescript-eslint/unified-signatures": "error",
+      "no-fallthrough": "off",
```
nanaya, line 134: non-typescript ones should be in the global rules?

.eslintrc.js
```diff
@@ -0,0 +1,263 @@
...
+      "@typescript-eslint/no-parameter-properties": "off",
+      "@typescript-eslint/no-unsafe-assignment": "warn",
+      "@typescript-eslint/no-unsafe-call": "warn",
+      "@typescript-eslint/no-unsafe-member-access": "warn",
+      "@typescript-eslint/no-unsafe-return": "warn",
+      "@typescript-eslint/no-unused-expressions": "error",
+      "@typescript-eslint/no-use-before-define": "off",
+      "@typescript-eslint/prefer-for-of": "error",
+      "@typescript-eslint/prefer-function-type": "error",
+      "@typescript-eslint/quotes": [
```
nanaya, line 113: should be in global for non-ts as well (the non-ts version). And a few other rules as well. [This may be useful](https://github.com/standard/eslint-config-standard-with-typescript/blob/a74134d087afb8eb2707635b85558b31226d2dd3/src/index.ts#L4).

.eslintrc.js
```diff
@@ -0,0 +1,263 @@
...
+  overrides: [{
+    env: {
+      browser: true,
+      node: false,
+    },
+    extends: [
+      "plugin:@typescript-eslint/recommended",
+      "plugin:@typescript-eslint/recommended-requiring-type-checking",
+      "plugin:react/recommended",
+      "plugin:react-hooks/recommended",
```
nanaya, line 21: global?

package.json
```diff
@@ -3,6 +3,7 @@
   "scripts": {
     "dev": "yarn run development",
     "development": "cross-env NODE_ENV=development webpack --progress --hide-modules --config=webpack.config.js",
+    "eslint": "eslint 'resources/assets/lib/**/*.{js,jsx,ts,tsx}' *.js",
```
nanaya, line 6: To include `.eslintrc.js`

```suggestion
    "eslint": "eslint 'resources/assets/lib/**/*.{js,jsx,ts,tsx}' '*.js'",
```

.eslintrc.js
```diff
@@ -0,0 +1,263 @@
...
+      ],
+      "@typescript-eslint/semi": [
+        "error",
+        "always",
+      ],
+      "@typescript-eslint/type-annotation-spacing": "error",
+      "@typescript-eslint/unbound-method": "warn", // TODO: some calls are intentionally unbounded...
+      "@typescript-eslint/unified-signatures": "error",
+      "no-fallthrough": "off",
+      "no-invalid-this": "off",
```
nanaya, line 135: can use some comment on why rules are turned off (whether it's permanent or a todo)

nanaya (changes requested):

nanaya (approved):

PR #7430, opened 2021-04-07
Add user parameter type

### LINKED ISSUE #7428: Add equivalent of V1 `type` parameter when looking up users via API
from v1 docs:

> - `type` - specify if `u` is a user_id or a username. Use `string` for usernames or `id` for user_ids. Optional, 
default behaviour is automatic recognition (may be problematic for usernames made up of digits only).

`GET /users/{user}/{mode?}` behaves like v1's automatic recognition right now, so numeric usernames are an issue

nanaya:
Resolves #7428.


notbakaneko (changes requested):
`UsersController::checkUsernameExists` should be updated

notbakaneko (approved):

PR #7471, opened 2021-04-12
Limit showing 12 difficulties on beatmap card

### LINKED ISSUE #7454: Difficulties on new Beatmap Cards going outside boundaries of the Beatmap Cards
This is probably best described through an example:
image
There are so many difficulties on each of the 4 modes that it goes outside of the Beatmap Card

Beatmap in Example: https://osu.ppy.sh/beatmapsets/1194435

nanaya:
Resolves #7454.


peppy (approved):
Let's get this one merged and let the users test.

PR #7473, opened 2021-04-12
Fix beatmap card difficulties popup

### LINKED ISSUE #7450: New beatmap cards go off layout on user profile

nanaya:
Resolves #7450.

This renders the popup on document root so it can have its own independent z-index. A bunch of additional things are required:

- lazy rendering of the popup
- js-based transition handling (as you can't animate from nothingness) using react-transition-group
- split off styling and everything from the panel
- absolute-positioned popup and fake it hard enough so it looks like part of the panel
  - this may fail if the card is inside another scrolling container or fixed element
  - not that I remember  any of such cases
  - will be added as necessary (probably rather difficult)


notbakaneko:
conflicts 🙃

resources/assets/less/bem/beatmaps-popup.less
```diff
@@ -0,0 +1,49 @@
...
+    display: grid;
+    grid-gap: 10px;
+  }
+
+  &__group {
+    display: grid;
+    grid-template-columns: minmax(0, 1fr);
+    grid-gap: 2px;
+  }
+
```
notbakaneko, line 48:
```suggestion
```

notbakaneko (changes requested):
seems fine; there was some weirdness in safari but it vanished by itself and now I can't replicate it anymore 🤔

notbakaneko (approved):

PR #7478, opened 2021-04-13
Larger cover on beatmap card

nanaya:
:shrug:


peppy:
I guess we'll also deploy this and see how the reaction is? Personally I kind of prefer the smaller one, but this does give better context by showing the "full" background.

PR #7480, opened 2021-04-13
Add ability to report beatmapsets

### LINKED ISSUE #7327: Report beatmaps from beatmap page
covering reports for stolen or nsfw content. this can be handled using a similar format to user or score web reports

venix12:
resolves #7327

only unwanted content and beatmap stealing for now (can add more easily), also adds ``BeatmapsetMenu`` component that could be used for other things in the future as well :eyes:


peppy:
I don't know where the term "Beatmap Stealing" comes from, but that shouldn't be what this report form is used for. That should be handled via a DMCA request exclusively.

venix12:
Well, since a long time there are cases of people just taking one's beatmap and submitting it (as-is or slightly modified), it used to be handled by the moderation teams with a proper silence.

peppy:
Let's not do that any more. It's not the correct process.

resources/assets/lib/beatmapsets-show/beatmapset-menu.tsx
```diff
@@ -0,0 +1,34 @@
...
+                key='report'
+                reportableId={beatmapset.id.toString()}
+                reportableType='beatmapset'
+                user={beatmapset.user}
+            />
+        );
+
+        return (
+            <PopupMenuPersistent>
+                {children}
```
notbakaneko, line 30: Separate variable is kind of unnecessary with only a single item with no other significant logic.

resources/assets/coffee/react/beatmapset-page/header.coffee
```diff
@@ -184,6 +185,11 @@ export class Header extends React.Component
 
             @renderLoginButton()
 
+            if currentUser.id?
```
notbakaneko, line 188: Also shouldn't show the option to the beatmapset owner.

notbakaneko (changes requested):
This looks kind of odd, to be honest
image

The dots are also just floating there with no feedback at all

venix12:
what would you suggest instead of those 3 dots, perhaps a button with them inside?

peppy:
@arflyte can you check on https://github.com/ppy/osu-web/pull/7480#pullrequestreview-677168712 and propose a quick design for this?

arflyte:
On it. Gimme a moment.

arflyte:
I think we can use the same design as the fav button but with the exclamation triangle. Something similar can be easily added for the new beatmap info.

venix12:
well, according to https://github.com/ppy/osu-web/issues/7327#issuecomment-818384764 it should be hidden under a dropdown menu

notbakaneko:
I think the idea is that the button shouldn't be prominent and should be more discreet and subdued in appearance like on the use profile page? Thoughts, @arflyte?

peppy:
@arflyte still waiting for a response from you on this one. as per the previous comments, we don't want this button to be prominent because it is seldom used.

arflyte:
Missed this. Haven't give this much thought yet, as I was focused on the newer version of the page. We can use the style that we used on the profile page. Keep the ellipsis and put them in a transparent black circle so that it doesn't feel like it's floating. Use the same height as the buttons. Make the circle darker on hover.

resources/assets/lib/beatmapsets-show/beatmapset-menu.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { BeatmapsetJson } from 'beatmapsets/beatmapset-json';
+import { PopupMenuPersistent } from 'popup-menu-persistent';
+import * as React from 'react';
+import { ReportReportable } from 'report-reportable';
+
+interface Props {
+  beatmapset: BeatmapsetJson;
+}
+
+export const BeatmapsetMenu = (props: Props) => (
```
notbakaneko, line 13: `export default function`

notbakaneko (changes requested):

notbakaneko (approved):

PR #7482, opened 2021-04-13
Fix invalid timestamp breaking review editor on load

### LINKED ISSUE #7481: Modding reviews are unusably bugged in cases of using timestamps on general tab
Reported by numerous people. This is feature breaking and many people are quitting using reviews because they simply can't use reviews with this bug.

In some cases, trying to post a review will render an error, then the review tab will become inaccessible. The review tab will only be viewable for that user again if they log out, and their review will not post.

This is triggered by using timestamps on the general tab for problem or suggestion stamps. There are many reasons timestamps would be used on the general tab such as
- In error when drafting the review and forgetting to set a difficulty
- Pointing out issues that apply to all difficulties at that time
- Storyboard mods
- timing issues 
- etc

I've made a video recreating the issue which can be seen here:
https://www.youtube.com/watch?v=UW_VlWA3zBM

I used https://osu.ppy.sh/beatmapsets/1419424/discussion/-/reviews this map as my test subject (since it's mine)

notbakaneko:
closes #7481

~~Removes the timestamp resetting in #6287 which ends up breaking the node parsing~~
~~Incidentally, I can't reproduce the original issue in #6287?~~

using `undefined` seems fine, `null` is not

~~edit2: This means timeline validation on posting to all difficulties doesn't quite work since it's not explicitly `null` anymore (which needs different fix)~~
Makes no sense anyway, since you can post timestamps to general?


resources/assets/lib/beatmap-discussions/editor-beatmap-selector.tsx
```diff
@@ -50,11 +50,9 @@ export default class EditorBeatmapSelector extends React.Component<Props> {
   }
 
   select = (id: string) => {
-    const beatmapId = id !== 'all' ? parseInt(id, 10) : 'all';
+    const beatmapId = id !== 'all' ? parseInt(id, 10) : undefined;
 
-    if (beatmapId) {
```
notbakaneko, line 55: not sure what the original is checking 🤨

resources/assets/lib/beatmap-discussions/editor.tsx
```diff
@@ -464,7 +464,7 @@ export default class Editor extends React.Component<Props, State> {
           // clear invalid beatmapId references (for pasted embed content)
           const beatmap = typeof node.beatmapId === 'number' ? this.props.beatmaps[node.beatmapId] : undefined;
           if (beatmap == null || beatmap.deleted_at != null) {
-            Transforms.setNodes(editor, {beatmapId: null}, {at: path});
+            Transforms.setNodes(editor, {beatmapId: undefined}, {at: path});
```
nanaya, line 467: it looks like setting node prop to `null` will actually unset the prop (thus `undefined`). As there's no check here, the original `setNodes` will just keep looping as `node.beatmapId` will always be `undefined` and `undefined !== null`.

Not sure if it's slate thing or something here changed the behaviour.

nanaya (changes requested):

resources/assets/lib/beatmap-discussions/editor.tsx
nanaya: Probably this

https://github.com/ianstormtaylor/slate/blob/6a1376332bbd2567336c444c57c1e64fdf706feb/packages/slate/src/transforms/general.ts#L202

notbakaneko: https://github.com/ianstormtaylor/slate/blob/95f402c59414331b2eeca9a19bd2c73c0ab6cd6c/packages/slate/src/transforms/node.ts#L637
This would imply using `undefined` is better to avoid constantly triggering `set_node` since `node[k]` is going to be `undefined` instead of `null`.

There's also the cases where we have checks like `timestamp !== this.props.element.timestamp`

nanaya: yeah, although probably should just check if it needs updating at all in the first place (which I removed two weeks ago and caused this bug)

nanaya (changes requested):

resources/assets/lib/beatmap-discussions/editor-discussion-component.tsx
```diff
@@ -72,7 +72,6 @@ export default class EditorDiscussionComponent extends React.Component<Props> {
 
       Transforms.setNodes(this.context, {timestamp}, {at: path});
     } else {
-      Transforms.setNodes(this.context, {timestamp: null}, {at: path});
```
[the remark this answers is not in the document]
notbakaneko: This is the same as existing behaviour...although it makes no sense since timestamps don't error when directly posting to general 🤔

nanaya (approved):

PR #7484, opened 2021-04-14
Delete beatmapset forum topic upon beatmapset deletion

venix12:
part of #7465

PR #7485, opened 2021-04-14
Add delete beatmapset LIO endpoint

### LINKED ISSUE #7465: Add legacy IO endpoint for deleting a beatmap
Required for administrative tasks.

```
                LegacyHelper.RunLegacyIO($"beatmapsets/{BeatmapSetID}", "DELETE");
```

- Should run the standard deletion process with banchobot as the user if required (for logging purposes).
- Should return an error if the beatmap is in a state it can't be deleted.

venix12:
- [x] depends on #7484
resolves #7465


nanaya (changes requested):

routes/web.php
```diff
@@ -521,6 +521,7 @@
     Route::apiResource('users', 'InterOp\UsersController', ['only' => ['store']]);
 
     Route::group(['namespace' => 'InterOp'], function () {
+        Route::delete('beatmapsets/{beatmapset}', 'BeatmapsetsController@destroy')->name('beatmapsets.destroy');
```
nanaya, line 524: use `::resource`

tests/Controllers/InterOp/BeatmapsetsControllerTest.php
```diff
@@ -0,0 +1,55 @@
...
+            ->delete($url)
+            ->assertStatus(204);
+
+        $beatmapset->refresh();
+        $topic->refresh();
+
+        $this->assertTrue($beatmapset->trashed());
+        $this->assertTrue($topic->trashed());
+        $this->assertSame($eventBeforeCount, Event::count());
+        $this->assertSame($logBeforeCount + 1, Log::where('log_operation', 'LOG_BEATMAPSET_DELETE')->count());
```
nanaya, line 53: These are already tested elsewhere. Check the user in the generated log instead.

nanaya (changes requested):

nanaya (approved):

PR #7495, opened 2021-04-16
Prevent redirect loop on invalid topic

### LINKED ISSUE #7464: Incorrect status on updated beatmap on beatmap info page
I don't know if anyone else has had this issue but it occured on my own mapset:
https://osu.ppy.sh/beatmapsets/1368018

You can see that it's updated properly but the tag on the right shows "Graveyarded"
I have tried to reupdate several times a week ago and also now, but it still stay the same status everytime it was updated

For some reason this status is handled properly if I check it in my user page under pending beatmaps:
https://osu.ppy.sh/users/5075660

I hope this does not affect nomination process

nanaya:
Accessing topic with no posts currently causes a redirect loop.

I'm not sure what the correct response should be in such case...

Resolves #7464 (osu-web side, beatmapset getting in weird state is fixed separately)

### DIFF
--- a/app/Http/Controllers/Forum/TopicsController.php
+++ b/app/Http/Controllers/Forum/TopicsController.php
@@ -323,7 +323,12 @@ class TopicsController extends Controller
             if ($skipLayout) {
                 return response(null, 204);
             } else {
-                return ujs_redirect(route('forum.topics.show', $topic));
+                // make sure topic has posts at all otherwise this will be a redirect loop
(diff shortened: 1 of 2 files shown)


peppy:
404 seems fine. Probably safe to mark this as closing https://github.com/ppy/osu-web/issues/7464? I'll fix the beatmap's state itself at a database level.

notbakaneko (approved):

PR #7503, opened 2021-04-19
Typescript PlaymodeTabs component

nanaya:
Somehow ended up here because I wanted to reuse the ordered group by mode beatmaps function :thinking:


notbakaneko (changes requested):
Links are disabled, this doesn't seem right 🤔 
image

notbakaneko (approved):

PR #7504, opened 2021-04-19
Gracefully handle missing user in channels

notbakaneko:
Instead of everything just erroring.
Also, finally removes chat's custom user link rendering.
(and fixes usercard missing tooltip position)

### DIFF
--- a/resources/assets/lib/chat/message-group.tsx
+++ b/resources/assets/lib/chat/message-group.tsx
@@ -1,12 +1,11 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
 // See the LICENCE file in the repository root for full licence text.
 
-import { route } from 'laroute';
-import * as _ from 'lodash';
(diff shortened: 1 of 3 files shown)


resources/assets/lib/chat/message-group.tsx
```diff
@@ -30,9 +29,9 @@ export default class MessageGroup extends React.Component<Props, any> {
     return (
       <div className={className}>
         <div className='chat-message-group__sender'>
-          <a className='js-usercard' data-tooltip-position='top center' data-user-id={sender.id} href={route('users.show', { user: sender.id })}>
```
nanaya, line 33: apparently the `tooltip-position` is supposed to work except it's not anymore when it's converted to typescript. Should probably at least keep the parameter intact for when it's supported again.

also `top center` is not the default for user card. It's hardcoded to `right center` at the moment which is the original default.

nanaya (changes requested):

nanaya (approved):
Not sure about link without href but should do for now I guess :thinking:

PR #7506, opened 2021-04-19
Ordered beatmap grouping

nanaya:
The main advantage of using object is it's easier to transform to and from json but using map allows using grouping result directly without mapping from modes first.

- [x] depends on #7503


notbakaneko (changes requested):
seems fine, although not sure about setting the target to es2020 🤔 

also conflicts

notbakaneko (approved):

Issue #4728, opened 2019-07-26
Google Chrome internal website search does not fully work


NoffyNoffeh:
If you type in osu.ppy.sh in the google chrome url/search bar then hit space, it gives you the option to search the website using its own search quickly.

However, if you actually try to do this, you get taken to osu's search page without your search term actually there.
![2019-07-26_16-08-55](https://user-images.githubusercontent.com/33286824/61978685-f1b07b00-afbf-11e9-82f7-17717dca4120.gif)

Doing the google chrome quick search should already fill out your search term once you are taken to osu's search page.

notbakaneko:
Chrome adds this automatically based on a heuristic, there isn't anything in osu-web that provides omnibox support

peppy:
Is this safe to close as not-our-issue?

cl8n:
you can add support for it explicitly, I just told noffy she should try adding this herself because all that's needed is https://developer.mozilla.org/en-US/docs/Web/OpenSearch

muekoeff:
Sample OpenSearch-file which can be referenced by inserting `<link rel="search" type="application/opensearchdescription+xml" href="/path/to/opensearch.xml" title="osu! beatmap listing" />` into `<head>`:

```
<?xml version="1.0" encoding="UTF-8"?>
<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/" xmlns:moz="http://www.mozilla.org/2006/browser/search/">
	<ShortName>osu! beatmap listing</ShortName>
	<Description>osu! beatmap listing</Description>
	<Url method="get" template="https://osu.ppy.sh/beatmapsets?q={searchTerms}" type="text/html">
	</Url>
	<Image height="16" type="Image/x-icon" width="16">https://osu.ppy.sh/favicon.ico</Image>
	<moz:SearchForm>https://osu.ppy.sh/beatmapsets</moz:SearchForm>
</OpenSearchDescription>
```

LittleEndu:
Shouldn't something like ``https://osu.ppy.sh/home/search?query={searchTerms}`` be used instead?

PR #7512, opened 2021-04-21
Added OpenSearch support

YellowberryHN:
I saw that https://github.com/ppy/osu-web/issues/4728 has been open since 2019, and I figured I could do it, since it's a rather small thing that could be potentially helpful to many people.

I have only one concern, the OpenSearch spec says the `opensearch.xml` should be served with the Content-Type of `application/opensearchdescription+xml`, however in my testing it worked fine. It may be a problem for other browsers, so it's something to consider for later, but I personally have no idea how to do that.


alchzh:
I wonder if there should be a separate search spec for just beatmap search.

nanaya:
The problem with doing it this way is it hardcodes to `osu.ppy.sh` for all instances. This may get confusing for people testing on other places.

YellowberryHN:
> The problem with doing it this way is it hardcodes to `osu.ppy.sh` for all instances. This may get confusing for people testing on other places.

I had this concern as well, if you have a suggestion on how that could be solved, I'd love to hear it.

cl8n:
you'll need to serve it from application to do that, so make a route and blade template. see `testflight` route for a similar basic example. the `ext_view` function that sends blade view can also set content-type

nanaya:
Continued to #9916.

PR #7513, opened 2021-04-21
Add legacy score entry for solo play

nanaya:
Score mapping fun.

### DIFF
--- a/app/Http/Controllers/Solo/ScoresController.php
+++ b/app/Http/Controllers/Solo/ScoresController.php
@@ -68,8 +68,10 @@ class ScoresController extends BaseController
         $params['ended_at'] = now();
         $params['mods'] = Mod::parseInputArray($params['mods'] ?? [], $score->ruleset_id);
 
-        $score->complete($params);
-        $score->createLegacyEntry();
+        $score->getConnection()->transaction(function () use ($params, $score) {
+            $score->complete($params);
+            $score->createLegacyEntry();
+        });
 
         return json_item($score, 'Solo\Score');
     }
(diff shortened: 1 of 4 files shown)


app/Models/Solo/Score.php
```diff
@@ -65,4 +68,65 @@ public function complete(array $params): void
...
+        $statAttrs = [
+            'Good',
+            'Great',
+            'LargeTickHit',
+            'LargeTickMiss',
+            'Meh',
+            'Miss',
+            'Ok',
+            'Perfect',
+            'SmallTickMiss',
```
notbakaneko, line 83: are these attributes going to be pascal cased in the final? (and thus not match `snake_case` everywhere else for our json)

peppy: That's just our c# naming conventions coming through. We can probably change them if they are to be handled by `osu-web` going forward (although not sure we'd do the same if they were handled by a queue processor which was already .NET based..? ie. are the snake_casing conventions something we do in all cases for json, or when the receiving end is `osu-web`?)

app/Http/Controllers/Solo/ScoresController.php
```diff
@@ -69,6 +69,7 @@ public function update($beatmapId, $id)
         $params['mods'] = Mod::parseInputArray($params['mods'] ?? [], $score->ruleset_id);
 
         $score->complete($params);
+        $score->createLegacyEntry();
```
notbakaneko, line 72: transaction? (because can't `complete` a second time if something errors in the next line)

notbakaneko (changes requested):

notbakaneko (approved):

PR #7514, opened 2021-04-21
Fix documentation of UserAccountHistory

alchzh:
https://github.com/ppy/osu-web/blob/28bb42f994996bcf31c97e6d6ebb195a6a94c034/app/Transformers/UserAccountHistoryTransformer.php#L22-L30

Sorry if not the right way to do this -- not sure what the procedure is like for outside contributors.


peppy (approved):

resources/docs/source/includes/_structures/user_compact.md
```diff
@@ -93,7 +93,7 @@ image         | string      | |
 
 Field       | Type      | Description
 ------------|-----------|------------
-id          | number    | |
+description | string    |
```
nanaya, line 96: The trailing ` |` is required for correct table rendering.

```suggestion
description | string    | |
```

nanaya (changes requested):

peppy (approved):

resources/docs/source/includes/_structures/user_compact.md
```diff
@@ -108,4 +108,3 @@ awarded_at  | Timestamp | |
 description | string    | |
 image_url   | string    | |
 url         | string    | |
-
```
peppy, line 111: @nanaya is the missing final EOL going to be an issue?

alchzh: Hmm weird, looks like something caused by GitHub since I just approved the suggestion through the UI

alchzh: Oh there were originally two newlines at the end of that file for no reason (none of the other mds have it, at least), so removing one is probably better

nanaya (changes requested):

nanaya (approved):

PR #7517, opened 2021-04-22
Allow changing beatmap difficulty owner

### LINKED ISSUE #2165: Ability to assign person responsible for certain difficulty - beside the owner - for guest difficulties purposes.
Per title - I think it would be good idea to have owner of the map be able to assign person that can actually moderate only certain difficulty they own (i.e. respond to issues, resolve comments etc. like the original owner)

nanaya:
- [x] The submission server update for handling upload permission will need to be merged in first...

Resolves #2165.

Editing from discussion page

The dialog itself

The only indicator of the guest at the moment

Different badge for post from guest

Event log (no difference between reset and assignment)


peppy:
FWIW the changes from my end are minimal, so it's fine to get this merged in as soon as feasible.

notbakaneko (commented):
Doesn't renaming the migration cause it to try and run again?

resources/assets/lib/notification-maps/message.ts
```diff
@@ -5,12 +5,16 @@ import * as _ from 'lodash';
 import Notification from 'models/notification';
 
 export function formatMessage(item: Notification, compact = false) {
-  const replacements = {
+  const replacements: Partial<Record<string, string>> = {
```
notbakaneko, line 8: should intersect with `{ title: string }` since title should always be present.

resources/assets/lib/notification-maps/message.ts
```diff
@@ -5,12 +5,16 @@ import * as _ from 'lodash';
...
     content: item.details.content,
     title: item.title,
     username: item.details.username,
   };
 
+  if (item.name === 'beatmap_owner_change') {
+    replacements.beatmap = item.details.version;
```
notbakaneko, line 15: `NotificationDetails` interface needs to be updated

resources/assets/lib/notification-maps/url.ts
```diff
@@ -32,6 +36,8 @@ export function urlGroup(item: Notification) {
 
 export function urlSingular(item: Notification) {
   switch (item.name) {
+    case 'beatmap_owner_change':
+      return route('beatmapsets.discussion', { beatmapset: item.objectId, beatmap: item.details.betamapId });
```
notbakaneko, line 40: key order

notbakaneko (changes requested):

notbakaneko:
Also, there's now this situation on mobile
image

notbakaneko (approved):

PR #69, opened 2015-09-09
Build and run on PHP 5.6 and 7, and try to fix travis PR issues

ameliaikeda:
I contacted travis-ci, and they're fixing the pull request build issues (see https://github.com/travis-ci/travis-scheduler/commit/85688758392f362c4efc5f73b1fc59affda85b97).

This just adds 5.6 and 7.0 to the build matrix, which we should have parity with anyway.

PR #7518, opened 2021-04-22
Added avatars and rank summary to user profile cards

Xetera:
Before: 
image

After: 
image

Users without avatars: 
image

Not sure if the ranks need to be formatted like `123,000`. Totally open to changing the wording of the description if needed of course.

I wasn't able to test for cases where the user doesn't have a global rank since mysql doesn't let me set null values on user ranks but I know profiles that don't have global ranks like https://osu.ppy.sh/users/4198271 exist. If that needs to be tested before this is merged I might need a little help with it 🙏.

### DIFF
--- a/resources/lang/en/users.php
+++ b/resources/lang/en/users.php
@@ -141,6 +141,7 @@ return [
     ],
     'show' => [
         'age' => ':age years old',
+        'rank_summary' => ':username is rank :global worldwide and rank :local in :country.',
         'change_avatar' => 'change your avatar!',
         'first_members' => 'Here since the beginning',
         'is_developer' => 'osu!developer',
--- a/resources/views/users/show.blade.php
+++ b/resources/views/users/show.blade.php
@@ -12,7 +12,7 @@
       'country' => $data["country"]["name"],
     ]),
     'opengraph' => [
-        'title' => $user->username . "'s Profile",
+        'title' => trans('users.show.title', ["username" => $user->username]),
         'image' =>  str_replace('http://localhost:8080', 'https://ffvyiglk.tunnelto.dev', $data["avatar_url"])
         // 'image' =>  json_encode($data),
     ]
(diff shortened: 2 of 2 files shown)


peppy:
It doesn't let you set null because that's not how it works. You can check existing usages and display which you should share logic with (or at very least follow the same conditional and formatting paths).

The formatting should match existing displays everywhere, which it doesn't right now.

Xetera:
By matching formatting are you referring to displaying the stats as something like this?

> ### Peppy's profile
> **Global Ranking:** #69,420
> **Country Ranking:** \#1,337

On the frontend the number formatting is being done with `osu.formatNumber` but from what I'm seeing there's no logic for it on the php side. https://www.php.net/manual/en/function.number-format.php behaves the same way as the js function, I'm guessing it would be fine to use that instead of duplicating the same function logic?

peppy:
Seems better, for sure.

nanaya (changes requested):

peppy:
@Xetera are you still interested in applying the review for this pull request? it should be in a good state to get merged if you follow up on that.

Xetera:
Ah sorry, I got a little overwhelmed by the requested changes after I realized the scope of this is bigger than I thought. I'll try to see if I can address some of these over the weekend. Forgot osu had more than 1 game mode kekw

Xetera:
Ah sorry, I got a little overwhelmed by the requested changes after I realized the scope of this is bigger than I thought. I'll try to see if I can address some of these over the weekend. Forgot osu had more than 1 game mode kekw

Xetera:
Ah sorry, I got a little overwhelmed by the requested changes after I realized the scope of this is bigger than I thought. I'll try to see if I can address some of these over the weekend. Forgot osu had more than 1 game mode kekw

notbakaneko:
I'm taking over this (and the related ogp issue, I guess 👀 )

notbakaneko:
superseded by #10570 since I can't push to this branch

PR #7522, opened 2021-04-23
Include channel in response when joining channel

notbakaneko:
Some initial updates before adding UI support for joining/rejoining public and multiplayer channels on web.
Also adding support for preselecting a different channel on start.

This ended up doing a lot less than originally planned 🤔 
I was planning on also cleaning up the channel tests using `dataProvider`s but it ended up being in a weird state when some channels are actually supposed to be dependent on something else existing as a precondition.

Actually, I'm not sure multiplayer channels should be joinable via this endpoint since a precondition for them is the `Room` to exist and joining the room automatically puts you in the channel.
Also, multiplayer chat channels aren't supposed to be interacted with via the current web chat UI and will be removed in another PR (which makes the ability to join the channel in this PR pointless 🤔 )


app/Http/Controllers/Chat/ChannelsController.php
```diff
@@ -243,6 +244,10 @@ public function store()
             if ($channel->exists) {
                 $channel->addUser($sender);
             }
+        } else if ($params['type'] === Channel::TYPES['multiplayer']) {
+            $room = Room::findOrFail($params['target_id']);
+            $channel = $room->channel;
+            $channel->addUser($sender);
```
notbakaneko, line 250: I'm not sure this is necessary since `Room` includes the `channel_id` so you can use the other endpoint instead 🤔

app/Http/Controllers/Chat/ChannelsController.php
```diff
@@ -243,6 +244,10 @@ public function store()
             if ($channel->exists) {
                 $channel->addUser($sender);
             }
+        } else if ($params['type'] === Channel::TYPES['multiplayer']) {
+            $room = Room::findOrFail($params['target_id']);
```
nanaya, line 248: what's the conclusion on joining multiplayer room?

nanaya (changes requested):

notbakaneko:
This need a bit of rework 🙃

nanaya:
conflicts

resources/assets/lib/chat/conversation-list.tsx
```diff
@@ -1,33 +1,47 @@
...
-@inject('dataStore')
 @observer
-export default class ConversationList extends React.Component<any> {
+export default class ConversationList extends React.Component {
+  @computed
+  private get nonPmChannels() {
+    return core.dataStore.channelStore.nonPmChannels;
+  }
+
+  @computed
```
nanaya, line 17: are the `@computed` actually useful considering all the functions do is simply returning another `@computed` value?

resources/assets/lib/chat/conversation-list.tsx
```diff
@@ -1,33 +1,47 @@
...
+    return this.pmChannels.map((conversation) => <ConversationListItem key={conversation.channelId} channelId={conversation.channelId} />);
+  }
+
+  private renderSeparator() {
+    if (this.nonPmChannels.length === 0 || this.pmChannels.length === 0) {
+      return null;
+    }
 
-    return <div className='chat-conversation-list'>{conversationList}</div>;
+    return <div key='separator' className='chat-conversation-list-separator' />;
```
nanaya, line 45: key not needed

resources/assets/lib/chat/conversation-list.tsx
```diff
@@ -1,33 +1,47 @@
...
 
-    pmChannels.forEach((conversation) => {
-      conversationList.push(<ConversationListItem key={conversation.channelId} channelId={conversation.channelId} />);
-    });
+  private renderNonPmChannels() {
+    return this.nonPmChannels.map((conversation) => <ConversationListItem key={conversation.channelId} channelId={conversation.channelId} />);
+  }
+
+  private renderPmChannels() {
+    return this.pmChannels.map((conversation) => <ConversationListItem key={conversation.channelId} channelId={conversation.channelId} />);
```
nanaya, line 37: the actual functions seem rather same so `renderChannel` instead and map from main `render` function?

nanaya (changes requested):

nanaya (approved):

PR #7525, opened 2021-04-24
Add basic chat filters support

### LINKED ISSUE #7375: Bancho chat filters do not apply to messages sent via osu-web chat/lazer
Bancho-level filters (used extensively for basic chat moderation) currently do not apply to chat sent via osu-web. Similarly, the same applies to chat sent via osu!lazer, likely because they share the same frontend (?).

venix12:
resolves #7375

TODO:
- [x] caching (used same solution as groups because it seemed fitting in that case :thinking:)
- [x] tests

need to run ``app('chat-filters')->resetCache()`` in artisan when filters are changed

not really sure if the table/model should go to the regular database or to the chat one.

pretty simple thing, could add some support for channel-specific filters or something in the future.

also the message doesn't get filtered on sender side (until they refresh the page) - thought that's intended on osu! stable so that people won't try to omit the filter when they see their message filtered 🤔


peppy:
This is missing tests. In addition, we may want to consider caching this else it is going to be a huge query every message lookup. See how we are doing similar for languages etc.

venix12:
I see, will work on that.

app/Models/Chat/Channel.php
```diff
@@ -274,6 +274,14 @@ public function receiveMessage(User $sender, string $content, bool $isAction = f
             throw new API\ChatMessageEmptyException(trans('api.error.chat.empty'));
         }
 
+        if ($this->isPublic()) {
```
nanaya, line 277: apparently this should apply to all channels and not just public ones

database/migrations/2021_04_23_115603_create_chat_filters_table.php
```diff
@@ -0,0 +1,35 @@
+<?php
```
nanaya, line 1: please rename the file to current date

app/Libraries/ChatFilters.php
```diff
@@ -0,0 +1,48 @@
...
+            $this->fetch();
+        }
+
+        return $this->chatFilters;
+    }
+
+    public function fetch()
+    {
+        $localCacheVersion = cache()->get('chat_filters_local_cache_version');
+        $localStorage = config('cache.local');
```
nanaya, line 26: this is on my todo list for the groups one but this process can be slightly improved:

- set the entire cache repository to variable instead of just the name
  - `$localCache = cache()->store(config('cache.local'))`
- use static cache key with version as part of its value so cache directory doesn't get filled up with outdated cache data
  - `$localCache->forever('chat_filters', ['version' => $localCacheVersion, 'data' => ChatFilter::all()])`

database/migrations/2021_04_23_115603_create_chat_filters_table.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('chat_filters', function (Blueprint $table) {
+            $table->bigIncrements('id');
+            $table->string('match');
```
nanaya, line 21: can use unique index on this column

nanaya (changes requested):

app/Libraries/ChatFilters.php
```diff
@@ -0,0 +1,55 @@
...
+            && array_key_exists('version', $cachedFilters)
+            && $cachedFilters['version'] === $localCacheVersion
+        ) {
+                $this->chatFilters = $cachedFilters['data'];
+        }
+
+        if ($this->chatFilters === null) {
+            $this->chatFilters = ChatFilter::all();
+            $localCache->forever('chat_filters', ['version' => $localCacheVersion, 'data' => $this->chatFilters]);
+        }
```
nanaya, line 46: Is there reason for checking `data` key separately?

```suggestion
        $cachedFilters = $localCache->get('chat_filters');

        if ($cacheFilters === null || $cachedFilters['version'] !== $localCacheVersion)
        {
            $cachedFilters = ['version' => $localCacheVersion, 'data' => ChatFilter::all()];
            $localCache->forever('chat_filters', $cachedFilters);
        }

        $this->chatFilters = $cachedFilters['data'];
```

app/Models/Chat/Channel.php
```diff
@@ -304,6 +304,14 @@ public function receiveMessage(User $sender, string $content, bool $isAction = f
             throw new API\ChatMessageEmptyException(trans('api.error.chat.empty'));
         }
 
+        if (!$this->isPM()) {
```
nanaya, line 307: uh, by all channels I do mean all, including PMs

nanaya (changes requested):

nanaya (approved):

PR #7528, opened 2021-04-26
Add class member sort lint for eslint

nanaya:
Annoyingly there's no fixer for this on eslint.

Even more annoying it is case sensitive while tslint one isn't.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -62,7 +62,8 @@ module.exports = {
       ],
       '@typescript-eslint/member-delimiter-style': 'error',
       '@typescript-eslint/member-ordering': [
-        'error',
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -62,29 +62,33 @@ module.exports = {
       ],
       '@typescript-eslint/member-delimiter-style': 'error',
       '@typescript-eslint/member-ordering': [
-        'error',
+        // TODO: flip to error once tslint is turned off. The order is case sensitive while the tslint one isn't.
+        'warning',
```
notbakaneko, line 66:
```suggestion
        'warn',
```

notbakaneko (changes requested):

notbakaneko (approved):

PR #7529, opened 2021-04-26
More useful error on invalid route

### LINKED ISSUE #7516: Client credentials grant API endpoint returns HTTP 404 for GET requests.
It should probably return 405 Method Not Allowed instead to clarify mistake on user's end.

nanaya:
Resolves #7516. There's no way to do partial match check from what I can tell.

### DIFF
--- a/app/Http/Controllers/FallbackController.php
+++ b/app/Http/Controllers/FallbackController.php
@@ -22,10 +22,6 @@ class FallbackController extends Controller
 
     public function index()
     {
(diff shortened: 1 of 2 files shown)


peppy:
Is this going to be returned to users now? May want to consider only returning this message when it's a json request (as previous) as an end user visiting a url shouldn't ever care about invalid methods? Not sure..

notbakaneko (approved):

PR #7531, opened 2021-04-26
Update javascript dependencies

nanaya:
- update libraries in vulnerability list (except immer/slate)
- update latest karma (related to item above)
- remove webpack-dev-server (not used)


notbakaneko (changes requested):
should run `yarn-deduplicate` as well 🙃  (to flatten `yarn.lock`)

notbakaneko (approved):

PR #7532, opened 2021-04-27
Fix yarn lockfile

nanaya:
Not sure what's happened here but installing dependencies changes the lockfile (removes unused dependencies) for me (yarn 1.22.10) :thinking:


notbakaneko (approved):
must be the dedup part of yarn that actually works 🤔

PR #7544, opened 2021-05-02
Fix markdown TOC and header ID generation

cl8n:
- fixes bug where TOC is skipped entirely if `title_from_document` isn't set
- gives all headers IDs when TOC is enabled, even if they won't make it into the displayed TOC (3 levels max)


app/Libraries/Markdown/OsuMarkdownProcessor.php
```diff
@@ -202,12 +203,12 @@ public function loadToc()
         if (
             !$this->node instanceof Block\Heading ||
             !$this->event->isEntering() ||
-            $this->title === null ||
-            $this->node->getLevel() > 3
+            ($this->titleFromDocument && $this->title === null)
```
nanaya, line 206: this check should be changed to just `level == 1`

nanaya (changes requested):

app/Libraries/Markdown/OsuMarkdownProcessor.php
cl8n: there are some documents with multiple H1 so I didn't do that. but looking on osu-wiki now I see very few cases, so I'll apply this suggestion and open separate fix on osu-wiki

cl8n: I'll still fix on osu-wiki but I checked closer and nothing there is blocking this change, the only weird H1 stuff is in news posts where TOC/IDs have not been used so far.

nanaya (approved):

PR #7545, opened 2021-05-02
Use brighter beatmap status colors

### LINKED ISSUE #7463: Pending color on the new beatmaps card feels somehow bland
for comparison - wip and ranked and loved feel much more colorful, feel like making this yellow more alive would look nicer

cl8n:
resolves #7463 with suggestions from https://github.com/ppy/osu-web/issues/7463#issuecomment-817170150 and page-independent colors for WIP and Approved. see https://github.com/ppy/osu-web/pull/7545#issuecomment-842999263 for preview


geshii (changes requested):
I think you should just keep the same colors but brighten them. Ranked is now green, and it was originally blue.

cl8n:
imo changing colors doesn't hurt cuz these haven't been around that long and the proposed ones are closer to the well-known counterparts from osu!stable

geshii:
> 
> 
> imo changing colors doesn't hurt cuz these haven't been around that long and the proposed ones are closer to the well-known counterparts from osu!stable

why not then

geshii (approved):

nanaya:
hi @arflyte please check this one

arflyte:
Let's stick to what we have for now. Ranked has to stay prominent compared to the rest.

VINXIS:
maybe swap ranked and approved? i think green works for ranked better than it does for approved especially considering how qualified itself is also greenish

arflyte:
Even if we swap the rank and approved, blue is still kinda prominent (based on some beatmap listing test I did some time ago). If you think it's okay for approved and qualified to be slightly prominent like Ranked, then yeah.

VINXIS:
Ahhhhh true, i just found it odd as to how similar qualified and approved are in colour choice i guess while ranked is far more different but yea, approved is a dead category anyway so doesn't matter that much, it'd be weird in some very specific circumstances only

Ephemeralis:
We could possibly merge Approved into Ranked (as Approved no longer has a salient purpose in the modern day afaik) and utilize either the proposed bright green or blue for Ranked instead? A bit complicated, but the newer more vibrant set of colours do fit the overall palette better, I think.

peppy:
Also in favour of some combination of the new colours, as the old ones do seem too desaturated in contrast to everything else.

arflyte:
Figma_210518-181023

Probably this?

cl8n:
applied that, and made Approved same color as Ranked since it wasn't provided and what @Ephemeralis said

also, "Unranked" isn't a thing

nanaya (approved):

PR #7546, opened 2021-05-02
Add infobox support for wiki

### LINKED ISSUE #7031: add basic infobox support for wiki
the lack of infoboxes makes writing a lot of articles more difficult compared to mediawiki when you want to provide key points & a visual summary without disrupting the flow of the page. in lieu of having a more robust system to parse and render templates, I think it'd be nice to add styling for infoboxes so they can be included as normal markdown wrapped in an extra div. this implies allowing html in wiki page markdown, not sure if u want to do that

example on a nonsense page:

```markdown
<div class="osu-md__infobox">

#### Heading

![](https://a.ppy.sh/6616586?1434426025.png "Someone's avatar")

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.

##### Smaller heading

| | |
| :-- | :-- |
| Property | It's a cat |
| Color | Blue and also here is more text sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium |
| More prop | ... |

</div>
```

```css
.osu-md__infobox {
  width: 300px;
  padding: 10px;
  margin-left: 10px;
  float: right;
  border: 1px solid hsl(var(--hsl-b4));
  text-align: center;
}

.osu-md__infobox > :last-child {
  margin-bottom: 0;
}

.osu-md__infobox thead {
  display: none;
}

.osu-md__infobox tbody > tr:first-child > td {
  border-top: none;
}

.osu-md__infobox td {
  vertical-align: top;
}
```

image

cl8n:
resolves #7031. this adds a generic "StyleBlock" syntax that can be used to apply certain `osu-md` CSS classes, and specific styling for "infobox". usage on wiki would be

```
{{{ Infobox
<!-- markdown here -->
}}}
```

live example is at <https://github.com/cl8n/osu-wiki/blob/test/infobox/wiki/Infobox_test/en.md>


peppy:
Do you have an example of how this would be implemented from the wiki side? Are the html templates going to be something we make once and rarely edit?

cl8n:
it works the same way as the cards on Main Page -- `<div class="infobox"> markdown here </div>`. linked an example in OP. it's not a "template", just extra CSS, and won't require further changes from osu-web

app/Libraries/Markdown/HtmlBlockRenderer.php
```diff
@@ -0,0 +1,48 @@
...
+namespace App\Libraries\Markdown;
+
+use League\CommonMark\Block\Element\AbstractBlock;
+use League\CommonMark\Block\Renderer\BlockRendererInterface;
+use League\CommonMark\Block\Renderer\HtmlBlockRenderer as BaseHtmlBlockRenderer;
+use League\CommonMark\ElementRendererInterface;
+use League\CommonMark\Util\ConfigurationAwareInterface;
+use League\CommonMark\Util\ConfigurationInterface;
+
+class HtmlBlockRenderer implements BlockRendererInterface, ConfigurationAwareInterface
```
nanaya, line 15: a new block type is probably better?

Something like <code>```</code> for code block but different character.

nanaya (changes requested):

app/Libraries/Markdown/HtmlBlockRenderer.php
peppy: Definitely agree with something that isn't inline html.

app/Libraries/Markdown/StyleBlock/Element.php
```diff
@@ -0,0 +1,69 @@
...
+            }
+
+            $node = $event->getNode();
+
+            if ($node instanceof static && $node->isOpen()) {
+                return true;
+            }
+        }
+
+        return false;
```
nanaya, line 67: record instances of StyleBlock element when passed to `canContain` and check that instead.

nanaya (changes requested):

app/Libraries/Markdown/StyleBlock/Parser.php
```diff
@@ -0,0 +1,27 @@
...
+
+use League\CommonMark\Block\Parser\BlockParserInterface;
+use League\CommonMark\ContextInterface;
+use League\CommonMark\Cursor;
+
+class Parser implements BlockParserInterface
+{
+    public function parse(ContextInterface $context, Cursor $cursor): bool
+    {
+        if ($cursor->match('#^\{{3}[ \t]*#') === null) {
```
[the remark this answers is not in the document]
cl8n: in removing the space check I used `starts_with` instead of regex to test for `{{{`, so I didn't add a single-character check before it. lmk if you still want taht though. I don't know how to test performance at this level 🤔

app/Libraries/Markdown/StyleBlock/Element.php
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * @var static[]
+     */
+    protected $containedStyleBlocks;
+
+    public function __construct(string $class)
+    {
+        $this->class = $class;
+        $this->containedStyleBlocks = [];
```
nanaya, line 26: do it on the declaration above?

app/Libraries/Markdown/StyleBlock/Element.php
```diff
@@ -0,0 +1,71 @@
...
+
+    public function isCode(): bool
+    {
+        return false;
+    }
+
+    public function matchesNextLine(Cursor $cursor): bool
+    {
+        // Make sure the most nested open StyleBlock tries to handle this first
+        if ($cursor->getLine() === '}}}' && !$this->containsOpenStyleBlock()) {
```
nanaya, line 51: this should be `getRemainder` instead of `getLine` otherwise it won't close when inside a blockquote for example when the line look like `> }}}` and should still close (although practically it's not going to be used as such... yet?).

nanaya (changes requested):

app/Libraries/Markdown/StyleBlock/Element.php
cl8n: seen this style somewhere else, woops. setting in declaration is always preferred if possible?

nanaya: I think so unless the value can't be set there (function call etc).

nanaya (approved):

PR #7551, opened 2021-05-03
Fix check for restricted user when finding username availability

cl8n:
the comment below it says "when the user is restricted" but it was only checking for banned user 🤔


peppy (approved):

notbakaneko (commented):
Restrictions are more temporary and would make it weird as the reserved time shifts as they expire or get applied?

peppy:
We don't use the classic "ban" functionality full stop, so I think this is fine.

Temporary restrictions still allow the user to login, which would reset the cooldown for their username to become available, so it should not be an issue.

notbakaneko (approved):

PR #7568, opened 2021-05-08
Support playmodes in user group functions and log more user group events

cl8n:
one step closer to "staff log" pages

updated groups interop interface:

- add to group: `PUT /_lio/users/{user}/groups/{group}`
  - optional `playmodes` param, array of mode names
- remove from group: `DELETE /_lio/users/{user}/groups/{group}`
- set default group: `PUT /_lio/users/{user}/default-group`
  - required `group_id` param


nanaya (changes requested):

app/Models/UserGroupEvent.php
```diff
@@ -38,39 +40,84 @@ public static function logGroupRename(?User $actor, Group $group, string $oldNam
...
             'hidden' => true,
             'user_id' => $user->getKey(),
         ]);
     }
 
     private static function log(?User $actor, string $type, Group $group, array $attributes = []): self
     {
+        $details = [
+            'actor_name' => optional($actor)->username,
+            'group_name' => $group->group_name,
```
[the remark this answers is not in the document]
cl8n: since it's a historical logging thing I wanted to store the names of the users/groups at this point in time. in the view for group history I'm planning to add later, I'll show `group_name` and `user_name` instead of their current names

I don't know if/how `actor_name` would be used but thought more is better than less, just in case ¯\\\_(ツ)\_/¯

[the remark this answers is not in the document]
nanaya: Maybe move the assignment for `user_name` here then. Seems duplicated everywhere.

app/helpers.php
```diff
@@ -1655,3 +1655,13 @@ function unmix(string $resource)
 
     return new HtmlString($manifest[$resource]);
 }
+
+/**
+ * Checks if two arrays contain the same values in any order. Input arrays
+ * must not contain duplicate values.
```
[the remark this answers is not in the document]
cl8n: you could get incorrect result from this function by something like `['a', 'b', 'b'] ['a', 'b', 'c']` but I just changed this function to not have a precondition

app/Models/User.php
```diff
@@ -495,21 +496,64 @@ public static function lookupWithHistory($usernameOrId, $type = null, $findAll =
...
         }
 
-        $this->getConnection()->transaction(function () use ($actor, $group) {
-            $this
-                ->userGroups()
-                ->firstOrNew(['group_id' => $group->getKey()])
-                ->fill(['user_pending' => false])
+        $this->getConnection()->transaction(function () use ($actor, $group, $modes, $userGroup) {
+            if ($userGroup === null) {
+                UserGroupEvent::logUserAdd($actor, $this, $group, $modes);
```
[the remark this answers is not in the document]
cl8n: renamed it so it's more clear, also I removed the later comment as clearer variable name made it self explanatory

app/Models/User.php
```diff
@@ -495,21 +496,64 @@ public static function lookupWithHistory($usernameOrId, $type = null, $findAll =
...
     }
 
-    public function addToGroup(Group $group, ?self $actor = null): void
+    public function addToGroup(Group $group, ?array $modes = null, ?self $actor = null): void
     {
-        if ($this->findUserGroup($group, true) !== null) {
+        $this->addOrUpdateGroup($group, $modes, $actor, false);
+    }
+
+    public function addOrUpdateGroup(Group $group, ?array $modes = null, ?self $actor = null, bool $allowUpdate = true): void
```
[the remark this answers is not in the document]
cl8n: I made it just `addToGroup` again, now caller is responsible for checking if already in group if they don't want to accidentally update. I think it's actually more clear this way, cuz if I did `addToGroup(..., null)` I would be surprised if user now belonged to group with some modes

app/helpers.php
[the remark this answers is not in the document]
notbakaneko: use `Ds\Set::xor` instead, or require `playmodes` to always be in order without duplicates.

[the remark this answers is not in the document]
cl8n: used Set

app/Models/UserGroupEvent.php
```diff
@@ -36,44 +38,69 @@ class UserGroupEvent extends Model
 
     public static function logGroupRename(?User $actor, Group $group, string $oldName, string $newName): self
     {
-        return static::log($actor, static::GROUP_RENAME, $group, [
+        return static::log($actor, static::GROUP_RENAME, null, $group, [
             'details' => [
-                'old_name' => $oldName,
+                'group_name' => null,
```
nanaya, line 43: remove `new_name` instead? No entry for this yet anyway.

nanaya (changes requested):

database/migrations/2021_05_08_093103_add_mode_changes_to_user_group_events.php
```diff
@@ -0,0 +1,45 @@
...
+     * @return void
+     */
+    public function up()
+    {
+        DB::statement("ALTER TABLE user_group_events CHANGE type type ENUM(
+            'group_add',
+            'group_remove',
+            'group_rename',
+            'user_add',
+            'user_add_modes',
```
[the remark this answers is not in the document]
cl8n: misunderstood your earlier comment sorry. is the whole codebase switching to "playmodes"?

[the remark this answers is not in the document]
nanaya: it's playmodes when in context of user group

app/Models/User.php
```diff
@@ -495,21 +496,57 @@ public static function lookupWithHistory($usernameOrId, $type = null, $findAll =
...
+                if ($playmodesRemoved !== []) {
+                    UserGroupEvent::logUserRemovePlaymodes($actor, $this, $group, $playmodesRemoved);
+                }
+            }
+
+            ($activeUserGroup ??
+                $this->userGroups()->firstOrNew(['group_id' => $group->getKey()])
+            )
+                ->fill([
+                    'playmodes' => $playmodes === [] ? null : $playmodes,
```
nanaya, line 543: this probably should check if the group has playmodes instead and assign empty array anyway if has playmodes (or do this earlier to also allow to tell whether the group has playmodes or not in the event log instead of having them all contain an array value)

nanaya (changes requested):

app/Http/Controllers/InterOp/UserGroupsController.php
```diff
@@ -39,16 +40,16 @@ public function setDefault()
...
-            'group_id:int',
-            'user_id:int',
-        ]);
+        $params = array_merge(
+            ['group' => get_int(request()->input('group_id'))],
+            request()->route()->parameters(),
+        );
 
-        $group = app('groups')->byId($params['group_id'] ?? null);
+        $group = app('groups')->byId($params['group'] ?? null);
```
nanaya, line 48: `$params['group']` may be string (also `?? null` isn't needed; similarly for `user` below)

Though looking again, `byId` shouldn't be restricted to `?int` in the first place to match `find()` on models (or it should cast to int inside the function).

app/Http/Controllers/InterOp/UserGroupsController.php
```diff
@@ -39,16 +40,16 @@ public function setDefault()
 
     private function getUserAndGroupModels()
```
nanaya, line 41: I'm not sure if this function is still useful or can be
```
private function findGroupOrFail($groupId)
{
    ...
}
```

And in the controller actions
```
$group = $this->...($groupId);
$user = User::findOrFail($userId);
```

Considering both are accessible directly from `update($userId, $groupId)` etc (with only exception `setDefault`).

And maybe even add `byIdOrFail` to `Groups` class itself and throw `ModelNotFoundException` accordingly. Could also be useful for `GroupsController#show`.

database/migrations/2021_05_08_093103_add_playmode_changes_to_user_group_events.php
```diff
@@ -0,0 +1,45 @@
+<?php
```
nanaya, line 1: rename file to current date

app/Models/UserGroupEvent.php
```diff
@@ -34,46 +36,69 @@ class UserGroupEvent extends Model
...
     }
 
-    public static function logUserAdd(?User $actor, User $user, Group $group): self
+    public static function logUserAdd(?User $actor, User $user, Group $group, ?array $playmodes = null): self
     {
-        return static::log($actor, static::USER_ADD, $group, [
-            'user_id' => $user->getKey(),
+        return static::log($actor, static::USER_ADD, $user, $group, [
+            'details' => [
+                'playmodes' => $group->has_playmodes ? ($playmodes ?? []) : null,
```
nanaya, line 53: if the function receives `$userGroup` instead, this check can be removed.

app/Models/User.php
```diff
@@ -495,21 +496,57 @@ public static function lookupWithHistory($usernameOrId, $type = null, $findAll =
...
-                ->userGroups()
-                ->firstOrNew(['group_id' => $group->getKey()])
-                ->fill(['user_pending' => false])
+        $this->getConnection()->transaction(function () use ($actor, $group, $playmodes, $activeUserGroup) {
+            if ($activeUserGroup === null) {
+                UserGroupEvent::logUserAdd($actor, $this, $group, $playmodes);
+            } else {
+                $previousPlaymodes = $activeUserGroup->playmodes ?? [];
+                $playmodesAdded = array_values(array_diff($playmodes, $previousPlaymodes));
+                $playmodesRemoved = array_values(array_diff($previousPlaymodes, $playmodes));
```
nanaya, line 528: just noticed these checks could've been used in place of Set/xor above

app/Models/User.php
```diff
@@ -495,21 +496,57 @@ public static function lookupWithHistory($usernameOrId, $type = null, $findAll =
         }
     }
 
-    public function addToGroup(Group $group, ?self $actor = null): void
+    public function addToGroup(Group $group, ?array $playmodes = null, ?self $actor = null): void
```
nanaya, line 499: not really important but looking at the length of this file, this function could be in `UserGroup` instead.

nanaya (changes requested):

app/Models/User.php
```diff
@@ -539,7 +576,9 @@ public function removeFromGroup(Group $group, ?self $actor = null): void
 
     public function setDefaultGroup(Group $group, ?self $actor = null): void
     {
-        $this->addToGroup($group, $actor);
+        if (!$this->isGroup($group)) {
```
nanaya, line 579: forgot about this but `isGroup` shouldn't be used here as it's only for authorization check (value depends on auth method), not actual state of the user.

nanaya (changes requested):

peppy:
@cl8n another lingering one which needs your feedback

app/Models/User.php
cl8n: i think fits better in this class, or at least to me the mental model is more like "add user to group" than "make a new usergroup"

app/Models/UserGroupEvent.php
cl8n: i think this is fine as-is just to keep all the methods having similar signature & requirements

database/migrations/2022_01_27_203000_add_playmode_changes_to_user_group_events.php
```diff
@@ -0,0 +1,45 @@
+<?php
```
nanaya, line 1: update file date

nanaya (changes requested):

nanaya (approved):

PR #7581, opened 2021-05-11
Add beatmap playcount apidoc

### LINKED ISSUE #7550: `Get User Beatmaps` with type "most_played" does not match documented return type
`GET https://osu.ppy.sh/api/v2/users/12092800/beatmapsets/most_played` returns:

```json
[
   {
      "beatmap_id":1626537,
      "count":192,
      "beatmap":{
         "difficulty_rating":6.45,
         "id":1626537,
         "mode":"osu",
         "status":"ranked",
         "total_length":130,
         "version":"Sotarks' Peace of Mind"
      },
    ...
]
```

while the documentation at https://osu.ppy.sh/docs/index.html#get-user-beatmaps says it always returns an array of `BeatmapSet`s regardless of `type`.  The actual returned json doesn't match any model I've seen so far; probably a model specific to the most_played param on this endpoint if I had to guess.

notbakaneko:
`most_played` returns a different type 🙃 

closes #7550


nanaya (approved):
this endpoint is weird... (snake case path and `most_played` being mainly related to beatmap, not beatmapset)

PR #7585, opened 2021-05-12
Fix fail when no StyleBlock classes are allowed

cl8n:
I just learned in other PR that `new Set(null)` does not work...


cl8n:
as nanaya pointed out this can't actually cause fail where it's used. should prob still fix though, in case future thing registers this renderer and might not set any classes

nanaya (approved):
Just noticed the extensions/parsers used for main and indexable should probably be the same...

(a different issue)

PR #7592, opened 2021-05-13
Centralised user preferences handling

nanaya:
Still need more checking, I probably missed something :thinking:


resources/assets/lib/user-preferences.ts
```diff
@@ -0,0 +1,58 @@
...
+  private updatingOptions = false;
+  private user?: CurrentUser;
+
+  constructor() {
+    this.current = Object.assign({}, defaultUserPreferencesJson, this.fromStorage());
+  }
+
+  get = <T extends keyof UserPreferencesJson>(key: T) => this.current[key];
+
+  set = <T extends keyof UserPreferencesJson>(key: T, value: UserPreferencesJson[T]) => {
```
notbakaneko, line 21: I don't think these been to be bound since they're always invoked directly instead of their reference being passed elsewhere? 🤔

notbakaneko (changes requested):
beatmapset-filter.coffee
beatmapset-page/main.coffee
profile-page/detail.coffee
beatmapset-panel.tsx
notification.ts
beatmap-helper.ts
beatmapset-helper.ts

should call `userPreferences.get`?

notbakaneko (approved):

PR #7609, opened 2021-05-16
Improve featured artist beatmaps visibility

### LINKED ISSUE #3777: Improve Featured Artist beatmap visibility
Beatmapsets that use Featured Artist works should by their very nature, be of greater visibility to the general user than ordinary tracks. They're licensed content.

At the moment, there are no obvious markers that denote a track as inherently belonging to a Featured Artist game wise - they appear on both the website and the client as indistinguishable from normal beatmaps outside of perhaps the inclusion of 'featured artist' in their tags.

To fix this, I suggest:

- Add a new field to beatmapsets that allows them to be linked to an artist id.
- Create a design addition for Featured Artist listings that allows them to display ranked beatmaps linked to their artist's id.
- Add an element to the beatmapset listing page that clearly denotes a linked set as belonging to a particular Featured Artist's body of works, complete with a linkback to their FA listing
- Add a category to beatmap search that purely returns linked Featured Artist beatmaps - possibly with a special border or other visual marker to make them 'stand out' from ordinary content
- Consider exposing a feed of newly linked Featured Artist beatmaps for possible inclusion in stable/lazer as a 'feed' of licensed content

venix12:
Resolves #3777 

- adds a golden featured artist search filter

- adds featured artist badge on beatmapset panel and listing (silimar to explicit one), one on beatmapset page points to the artist page

- adds link to beatmapsets of featured artist on artist's page

to-do:
- run ``php artisan es:index-documents --types=beatmapsets --yes`` when it's merged
- ~possibly some admin page to attach featured artist to beatmapset or can be added to database staright?~ (https://github.com/ppy/osu-web/pull/7609#issuecomment-907915530)

manual migrations:
- [x] add new column (`track_id`, `bigint unsigned default null`) and index (`osu_beatmapsets_track_id_index` on `track_id`) on `osu_beatmapsets` table
- [x] add relevant migration entry `2021_08_30_153214_add_track_id_to_beatmapsets`


pishifat:
some songs are available on multiple artist listings, so it'd be good to allow assigning multiple artist ids to a single song

peppy:
If that's going to be a common thing, sounds like something we should fix at the other end (and reference the `track_id` rather than the artist).

Nivalyx:
> 
> 
> If that's going to be a common thing, sounds like something we should fix at the other end (and reference the `track_id` rather than the artist).

ah, the occurence is not that common actually but cases like these -do- happen from time to time yeah. here are all the maps that fell into such scenario iirc :

- Culprate & Dictate - Pencilina [(1)](https://osu.ppy.sh/beatmapsets/755651#osu/1590460) [(2)](https://osu.ppy.sh/beatmapsets/722634#osu/1564457), from both Culprate and Dictate's FA catalogue
- Culprate & Joe Ford - Gaucho [(1)](https://osu.ppy.sh/beatmapsets/847776#osu/1772923) [(2)](https://osu.ppy.sh/beatmapsets/937645#taiko/1960216) [(3)](https://osu.ppy.sh/beatmapsets/936234#taiko/1963887), from both Culprate and Joe Ford's FA catalogue
- DJ'TEKINA//SOMETHING - Hidamari no Uta [(1)](https://osu.ppy.sh/beatmapsets/831322#osu/1741594) [(2)](https://osu.ppy.sh/beatmapsets/966324#osu/2022692), from both MOtOLOiD and Yuyoyuppe's FA catalogue
- ginkiha - Winter Night Highway [(1)](https://osu.ppy.sh/beatmapsets/998565#taiko/2090565), from both Rising Sun Traxx and ginkiha's FA catalogue
- Panda Eyes & Teminite - Highscore [(1)](https://osu.ppy.sh/beatmapsets/615326#fruits/1297908) [(2)](https://osu.ppy.sh/beatmapsets/332532#osu/736214), from both Rising Sun Traxx and ginkiha's FA catalogue
- Panda Eyes & Teminite - Immortal Flame (feat. Anna Yvette) [(1)](https://osu.ppy.sh/beatmapsets/720024#osu/1520231) [(2)](https://osu.ppy.sh/beatmapsets/703957#osu/1511760), from both Rising Sun Traxx and ginkiha's FA catalogue
- Shawn Wasabi + YDG feat. YUNG GEMMY - Burnt Rice (Aiobahn & Jh-Anu Remix) [(1)](https://osu.ppy.sh/beatmapsets/519256#osu/1103732), from both Shawn Wasabi and Aiobahn's FA catalogue
- Yuyoyuppe - Emerald Galaxy [(1)](https://osu.ppy.sh/beatmapsets/682286#osu/1442757), from both MOtOLOiD and Yuyoyuppe's FA catalogue

aside from that to my knowledge there are also some other tracks that, although not yet mapped, are also featured in two or more FA pages (such as Powerless feat. Sennzai - Todestrieb for example) so yeah referencing the `track_id` intstead might be the best thing to do going forward here ;~;

cl8n:
associating by track is the more specific relation so I also think that's the way to go. that's technically a many-to-many bc some maps compile multiple songs together

venix12:
``track_id`` solution seems to make sense, where should the badge point then? to artist if there is one and just display "featured artists" with no url for ones assigned to multiple ones? or perhaps some more tricky solution 🤔

venix12:
@peppy Tracks in the current infrastructure can belong to only one artist so having it as ``track_id`` wouldn't change much at the very moment. Correct me if I'm wrong but I see two possible solutions of this problem:
- creating a pivot table with that would allow a track to be assigned to multiple artists
- creating a pivot table which would allow multiple artists to be assigned to a single beatmapset

Which one would you go with?

nanaya:
Add the `track_id` to beatmapsets table for now. We'll figure out how to deal with tracks with multiple artists later. Similarly for the link, link to the referenced track's artist.

Nivalyx:
*i'm not clayton but...*

...@nanaya as of now (limiting the scope to the Ranked/Loved beatmaps only) iirc there are 15 beatmaps that fall into such case -- the 14 that i mentioned earlier in [this reply](http://github.com/ppy/osu-web/pull/7609#issuecomment-843891837) + [Andromedik - Invasion](http://osu.ppy.sh/s/789544) which is now featured in both [High Tea Music](http://osu.ppy.sh/beatmaps/artists/26)'s and [Andromedik](http://osu.ppy.sh/beatmaps/artists/177)'s Featured Artist listing

this is in addition to a few more that will be so in the near future once certain Featured Artists are announced (that i know of)

peppy:
Just to reiterate, `track_id` at a beatmapset level is the best approach for now. We can make this system more complex if/when required, but let's get the basics out for now.

venix12:
Yeah, changed to the ``track_id``, should be pretty much ready for review now i guess :eyes:

Also, should there be an admin panel thingy added for that or is it going to be set on the database level? (@peppy)

peppy:
> Yeah, changed to the `track_id`, should be pretty much ready for review now i guess 👀
> 
> Also, should there be an admin panel thingy added for that or is it going to be set on the database level? (@peppy)

fine without for now

nanaya (changes requested):
Without interface on osu-web, whoever updating the beatmapset table will need to also queue reindex job for the updated beatmapsets...

venix12:
> Without interface on osu-web, whoever updating the beatmapset table will need to also queue reindex job for the updated beatmapsets...

so should it be added or not? or in another pr perhaps?

peppy:
For now it's fine to assume it will be handled elsewhere, until we have a process setup for that. It's out of scope for this PR.

resources/assets/lib/beatmapset-panel.tsx
```diff
@@ -415,6 +421,7 @@ export default class BeatmapsetPanel extends React.Component<Props> {
           <a className='beatmapset-panel__main-link u-ellipsis-overflow' href={this.url}>
             {osu.trans('beatmapsets.show.details.by_artist', { artist: getArtist(this.props.beatmapset) })}
           </a>
+          {(this.props.beatmapset.track !== null) && <FeaturedArtistBadge />}
```
nanaya, line 424: this check doesn't work because `!== null` is false when `track === undefined` (unless there's specific reason, `!= null` is generally preferred)

nanaya (changes requested):

resources/assets/lib/beatmapset-panel.tsx
nanaya: actually, this won't even work without `track` being included. Use `track_id` instead

resources/assets/coffee/react/beatmapset-page/header.coffee
```diff
@@ -142,13 +142,18 @@ export class Header extends React.Component
...
 
           span className: 'beatmapset-header__details-text beatmapset-header__details-text--artist',
             a
               className: 'beatmapset-header__details-text-link'
               href: laroute.route 'beatmapsets.index', q: getArtist(@props.beatmapset)
               getArtist(@props.beatmapset)
+            if @props.beatmapset.track?
+              a
+                className: 'beatmapset-badge beatmapset-badge--featured-artist'
+                href: laroute.route 'tracks.show', @props.beatmapset.track_id
```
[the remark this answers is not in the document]
venix12: i see, should i keep the include and the interface for future usage (?) or should it be just removed? 🤔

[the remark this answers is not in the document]
nanaya: remove it for now

nanaya (approved):

nanaya:
@peppy beatmapsets table need updating but may want to check if should be done through a different method because it may take a while (adding column and index; see the checklist in the first post)

peppy:
Not sure it needs to be a BIGINT, that's optimistic for sure.

database/migrations/2021_08_30_153214_add_track_id_to_beatmapsets.php
```diff
@@ -0,0 +1,37 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('osu_beatmapsets', function (Blueprint $table) {
+            $table->bigInteger('track_id')->unsigned()->nullable();
```
peppy, line 20: let's change to `INT` to match the `artist_tracks` table.

peppy (changes requested):

nanaya:
hm yeah, that could be changed as needed

peppy:
Migration has been applied

nanaya (approved):

PR #7613, opened 2021-05-17
Disallow nomination of beatmapset containing own beatmap

### LINKED ISSUE #7603: Disallow BNs/NAT from nominating sets that they're marked as GD owners on
Self-explanatory, it's a rule so might as well just grey out the nominate button for those sets now that GD support has been implemented.

nanaya:
Resolves #7603.


notbakaneko (changes requested):
I noticed that admins will skip the check in `OsuAuthorize` but not in `nominator.tsx` 🤔

notbakaneko (approved):

PR #7616, opened 2021-05-17
Option for larger beatmap card size

### LINKED ISSUE #7451: Include source on the new beatmapset card
Used to be a nice info which isn't available on the new design.

### LINKED ISSUE #7511: Always show play & favourites count on beatmap cards
Example gif showing issue

I don't see why this information is only visible on hover with the new beatmap card design. Before this change I would often use this info when looking at a large amount of pending/graved beatmaps in the search to see which maps had more favourites, in order to weed out maps that were more likely of a lower quality. Now, I have to manually move my cursor over every beatmap card in order to do this.

The space where this information appears is otherwise blank, so I don't see the issue with making this always display.

nanaya:
Resolves #7511, resolves #7451.

This also affects the display on other pages but the only place to change the setting is the listing page :eyes: 

- [x] depends on #7592


notbakaneko (changes requested):
This may be an issue as well, a bit more noticeable than previously (iPhone X width):
image

notbakaneko (approved):

PR #7619, opened 2021-05-18
Fix hiding mark as read button nearby "show more" button

nanaya:
Saw this when checking for other stuff :thinking:


notbakaneko (approved):
now the show less button moves 🙃

PR #7620, opened 2021-05-18
Don't show multiplayer channels in web chat

notbakaneko:
These aren't intended to be used with normal chat.
It's still in the presence response since that's still how the game client gets updates to those channels (and will be fixed  when #6805 is done).


resources/assets/lib/stores/channel-store.ts
```diff
@@ -248,7 +248,9 @@ export default class ChannelStore {
   @action
   updateWithPresence(presence: PresenceJson) {
     presence.forEach((json) => {
-      this.getOrCreate(json.channel_id).updatePresence(json);
+      if (json.type !== 'MULTIPLAYER') {
```
nanaya, line 251: how about the legacy multiplayer which has type `TEMPORARY`?

nanaya (changes requested):

nanaya (changes requested):

nanaya (approved):

PR #7628, opened 2021-05-20
Filter out restricted users on country statistics aggregate

### LINKED ISSUE #6469: empty pages on country ranking
void

nanaya:
Resolves #6469.

Also updated command in schedule to the full command name. It seems to still work regardless but confusing.


notbakaneko (approved):

notbakaneko (commented):
not sure how well these queries are actually going to run, though

PR #7632, opened 2021-05-20
Save post editor content

### LINKED ISSUE #7415: Prompt user whether they really want to navigate away from page if they are editing forum BBcode
I was editing a lengthy post, making formatting all pretty. I accidently clicked on this and now all changes are lost. 

In fact it happens for any link you press. There should be a confirmation prompting the user if they really want to navigate away from page.

nanaya:
Resolves #7415 by saving the content locally until it's saved or cancelled (or when the user logs out).

Also moved the storage for reply box and additionally saves new topic content (just the content; title/cover/poll/etc aren't saved at the moment).


resources/assets/lib/forum-post-input.ts
```diff
@@ -0,0 +1,93 @@
...
+    return input;
+  }
+}
+
+export default class ForumPostInput {
+  constructor() {
+    $(document).on('input change', '.js-forum-post-input', this.onInput);
+    $(document).on('turbolinks:load', this.handlePageLoad);
+    $.subscribe('forum-post-input:restore', this.handleRestore);
+    $.subscribe('forum-post-input:clear', this.handleClear);
```
notbakaneko, line 23: should also clear localstorage after posting

notbakaneko (changes requested):

PR #7634, opened 2021-05-21
Simplify additional script loading

nanaya:
Use jquery's existing functionality (and Promise) to load the scripts and use set instead of object for recording loaded scripts.

Discovered during exploring way to preload react components ahead of time.


notbakaneko (commented):
Should it abort loading on navigation?

resources/assets/lib/turbolinks-reload.ts
```diff
@@ -1,49 +1,36 @@
...
   constructor() {
-    document.addEventListener('turbolinks:before-cache', this.cleanup);
+    $(document).on('turbolinks:before-cache', this.abortLoading);
   }
 
-  cleanup = () => {
-    $(`.${className}`).remove();
+  abortLoading = () => {
+    for (const xhr of this.loading.values()) {
+      xhr.abort();
```
notbakaneko, line 14: Probably should also should remove from the set after aborting

resources/assets/lib/turbolinks-reload.ts
```diff
@@ -1,49 +1,36 @@
...
+    if (this.loaded.has(src)) return;
 
-    const el = document.createElement('script');
-    el.classList.add(className);
-    el.onload = () => {
-      // abort if the element has been removed (on navigation etc)
-      if (el.parentElement == null) {
-        return;
-      }
+    const xhr = $.ajax(src, { dataType: 'script' }) as JQuery.jqXHR<void>;
```
notbakaneko: add `cache: true` to avoid weird `?_=number` querystring

notbakaneko (changes requested):
Probably should also handle/prevent/abort multiple calls to load same script (original only loads once)

resources/assets/lib/turbolinks-reload.ts
nanaya: should already be done by the always callback

nanaya: ...and shouldn't be in loaded set if not done

resources/assets/lib/turbolinks-reload.ts
```diff
@@ -1,49 +1,36 @@
...
+    this.loaded.delete(src);
+    this.loading.get(src)?.abort();
   };
 
-  load = (src: string, onload?: () => void) => {
-    if (this.loaded[src]) {
-      return;
-    }
+  load(src: string) {
+    if (this.loaded.has(src) || this.loading.has(src)) return;
```
notbakaneko, line 24: return the existing xhr request?

resources/assets/coffee/_classes/twitch-player.coffee
```diff
@@ -9,7 +9,9 @@ class @TwitchPlayer
 
 
   initializeEmbed: =>
-    @turbolinksReload.load 'https://player.twitch.tv/js/embed/v1.js', @startAll
+    @turbolinksReload
+      .load 'https://player.twitch.tv/js/embed/v1.js'
+      .then @startAll
```
notbakaneko, line 14: `load` can be `undefined` by definition even though checking if `Twitch` is set should prevent it. Coming to think of it, wouldn't this loop infinitely if `Twitch` isn't set by the script? 🤔

notbakaneko (changes requested):

resources/assets/coffee/_classes/twitch-player.coffee
nanaya: the `load` wouldn't do anything if the script is already loaded

notbakaneko (approved):

PR #7636, opened 2021-05-21
Remove `beatmapset` terminology from the site localisation

TPGPL:
Removes the `beatmapset` terminology from the site based on [https://github.com/ppy/osu-wiki/issues/4723](https://github.com/ppy/osu-wiki/issues/4723)


TicClick:
would be awesome for `/beatmaps/ to be an alias for `/beatmapsets/`, too

peppy (approved):

PR #7639, opened 2021-05-23
Change Chinese "Country and Regions" to just "Countries" to be more in line with the other translations

myangelaku:
Context: https://discord.com/channels/188630481301012481/188630616286167041/846021060012015656


peppy:
The change needs to be made on crowdin. Please see the readme.

PR #7645, opened 2021-05-25
Observable window size state

nanaya:
Recompute whenever the screen is resized and use the computed value wherever needed.

To be used to skip rendering invisible beatmap background on mobile.


resources/assets/lib/beatmaps/virtual-list-meta.ts
```diff
@@ -21,21 +21,6 @@ export default class VirtualListMeta {
...
   }
-
-  @observable private isDesktop = true;
-
-  constructor() {
-    this.checkIsDesktop();
-    $(window).on('resize', this.checkIsDesktop);
-  }
-
-  destroy() {
```
notbakaneko, line 34: `search-content` is still calling this on unmount

notbakaneko (changes requested):

notbakaneko (approved):

PR #7651, opened 2021-05-26
Include is_restricted in User API responses

### LINKED ISSUE #7526: Missing Restricted Details on API Endpoint /api/v2/me
Hello.

As the title mentions, this issue is about the /api/v2/me endpoint does not show is the user is restricted or not. Day by day, the API v2 usage is increasing; especially for tournaments with OAuth login. The issue is that restricted users can also use OAuth with their osu! account and that leads to problems on systems which has integration with osu! OAuth login.

For instance, we are using osu! API v2 for the Turkish community in osu! - helping link their osu! and Discord account to enter the community Discord server for now, and also planned to extend with tournaments, news, and events sections which make everything about the community organized and easy to reach. However, we do not want restricted users to get authenticated on the system, so we need to check if they are restricted or not on callback. If we try using the `/api/v2/me` scope, which we already use to gather information about users, we cannot reach information about restriction detail. As a solution, we can try using the `/api/v2/users/:user` endpoint but in this way, we are actually sending more requests to osu! and also now need to request public scope too to use that endpoint.

Also, I am sure that tournaments will need to check if the user is restricted or not for more accurate results on registrations, and reducing the load on the screening team.

cl8n:
and update docs. resolves #7526


nanaya:
Conflicts

notbakaneko (changes requested):
`is_restricted` should only be present on `/api/v2/me` or on the user the token is for.

cl8n:
I wasn't sure whether to make the API inconsistent with included fields like that or not, but alright.

also had trouble deciding if another `OsuAuthorize` check for transformer was the right way to do this...

app/Libraries/OsuAuthorize.php
```diff
@@ -1723,6 +1723,24 @@ public function checkUserShow(?User $user, User $owner): string
...
+     * @param User $owner
+     * @return string
+     */
+    public function checkUserShowRestrictedStatus(?User $user, User $owner): string
+    {
+        if ($this->doCheckUser($user, 'IsNotOAuth')->can()) {
+            return 'ok';
+        }
+
+        if ($user !== null && $user->user_id === $owner->user_id) {
```
notbakaneko, line 1737: prefer using `getKey()` when possible.

notbakaneko (changes requested):

notbakaneko (approved):

PR #7655, opened 2021-05-26
Render code markdown in changelog titles

### LINKED ISSUE #7594: Add support for `grave` `escapes` in changelog titles
We've been doing this a lot more since github now supports this natively (see https://osu.ppy.sh/home/changelog/lazer/2021.514.0).

Styling could be something similar to how github does it - a rounded gray background around the phrase.

cl8n:
resolves #7594


resources/assets/coffee/react/_components/changelog-entry.coffee
```diff
@@ -24,9 +24,12 @@ export ChangelogEntry = ({entry}) =>
           a
             href: entry.url
             className: 'changelog-entry__link'
-            entry.title
+            dangerouslySetInnerHTML:
```
nanaya, line 27: this is all too much code for what's essentially just
```
_.escape(entry.title).replace(/`([^`]+)`/g, '<code>$1</code>')
```
as this is apparently what github is doing according to some basic testing (and it doesn't match commonmark spec either)

maybe if they updated it to allow proper markdown inline code then we'll adjust accordingly

nanaya (changes requested):

resources/assets/coffee/react/_components/changelog-entry.coffee
cl8n: alright, I'll just keep it to this component then
I noticed it can do balanced backticks as well, but it never allows backticks inside so it's still a simple regex almost identical to what you wrote.

nanaya (approved):

PR #7660, opened 2021-05-28
Update Apidoc (now Scribe)

nanaya:
The original project is barely updated (and still doesn't support php 8) and the previous maintainer (but not owner) forked the project (called Scribe).

`/resources/docs` now only contains autogenerated file and won't result in crappy `git grep` output anymore.

The new docs also changed the url/query/body params table into list.

Also custom styles are now inlined in `index.blade.php` because it seems to be the easiest way to do it :eyes:


app/Http/Controllers/Chat/Channels/MessagesController.php
```diff
@@ -26,7 +26,7 @@ class MessagesController extends BaseController
      *
      * Returns an array of [ChatMessage](#chatmessage)
      *
-     * @urlParam channel_id required The ID of the channel to retrieve messages for
+     * @urlParam channel integer required The ID of the channel to retrieve messages for
      * @queryParam limit number of messages to return (max of 50)
      * @queryParam since messages after the specified message id will be returned
      * @queryParam until messages up to but not including the specified message id will be returned
```
notbakaneko, line 32: missing types (also it thinks `number` is the type)

notbakaneko (changes requested):

notbakaneko (approved):

notbakaneko (changes requested):
conflicts 😕

notbakaneko (approved):

PR #7671, opened 2021-05-31
More useful browser test progress

nanaya:
Use `dataProvider` instead of running everything as single test and handling errors manually.

- [x] depends on #7670


phpunit.dusk.xml
```diff
@@ -6,7 +6,7 @@
          convertErrorsToExceptions="true"
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
-         processIsolation="true"
```
nanaya, line 9: @nekodex, do you remember why this is set to `true`? Checking [the dusk template](https://github.com/laravel/dusk/blame/6.x/stubs/phpunit.xml) looks like it's always `false`. Is it some leftover from experiments?

tests/Browser/SanityTest.php
```diff
@@ -180,76 +58,276 @@ public function cleanup()
...
+    }
+
+    private static function filterLog(array $log)
+    {
+        $return = [];
+
+        foreach ($log as $line) {
+            if ($line['source'] === 'network') {
+                $matches = [];
+                $count = preg_match_all("/^([^ ]+) - Failed to load resource: the server responded with a status of ([0-9]{3}) \(([^\)]*)\)$/i", $line['message'], $matches);
```
notbakaneko, line 203: why is this `$count` even here 🤔

phpunit.dusk.xml
nekodex: I have a feeling it was to try and work-around something weird but I don't remember exactly what, sorry.

If setting it to false doesn't cause any issues now, it's probably fine to change?

notbakaneko: I'm actually getting some weirdness where the test sometimes runs as the wrong user causing the expectations to fail but I haven't been able to isolate it yet 😕

notbakaneko (changes requested):
- Would be useful to add `app('groups')->resetCache()` before the browser tests start (or have the cache stored in different namespace)
- `run_dusk.sh` should be updated to pass flags to `phpunit`

notbakaneko (approved):

PR #7679, opened 2021-06-02
Add all available locales field in wiki page API

gagahpangeran:
This changes is add other available locales of the wiki article for the API.

Also add the documentation.


cl8n (changes requested):
style nitpick, the fields should be listed in alphabetical order, since they were before your change

notbakaneko (changes requested):
for API usage, it should really list all `available_locales` instead

gagahpangeran:
> for API usage, it should really list all `available_locales` instead

I just need to change the field name `other_locales` to `available_locales` but still using the function `otherLocales()` for it, am I right?

notbakaneko:
`otherLocales()` doesn't include the current locale

gagahpangeran:
Ah, I see. I just have to merge it with current locale.

notbakaneko (approved):
Would be nice if the locales were returned in order but that can be a separate issue.

PR #7680, opened 2021-06-02
Show description on group pages

### LINKED ISSUE #7640: Add descriptions on group pages
I don't remember when it was, but group pages used to have short descriptions about their purpose. not sure why that was removed. having a description on-page can help ppl learn about it quickly

currently this isn't very important since there aren't prominent links to the group listings in the first place (if you're on the page, you likely already know what the group is), but after #6936 it would have good use.

cl8n:
resolves #7640 with what hopefully is a basic enough design that it doesn't need special consideration. just a markdown document with smaller margins so that it sticks closer to the group title

- [x] depends on #7642


app/Transformers/GroupTransformer.php
```diff
@@ -23,4 +24,12 @@ public function transform(Group $group)
             'short_name' => $group->short_name,
         ];
     }
+
+    public function includeDescription(Group $group)
+    {
+        return $this->primitive([
```
nanaya, line 30: the typing is simpler if it returns null instead of array with null values `$this->primitive($group->group_desc === null ? null : [...])`

resources/assets/less/bem/osu-md.less
```diff
@@ -307,4 +309,8 @@
       text-align: right;
     }
   }
+
+  > *:last-child {
+    margin-bottom: 0;
+  }
```
nanaya, line 315: this breaks styling for store page (and not needed to be moved if using variable).

resources/assets/less/bem/osu-md.less
```diff
@@ -262,6 +263,7 @@
     margin-bottom: @md-paragraph-space;
     line-height: 1.5;
 
+    .@{_top}--group &,
```
nanaya, line 266: change `@md-paragraph-space` to use variable instead (`--paragraph-space: @md-paragraph-space`) and set the override at block level modifier style

resources/views/docs/_structures/group.md
```diff
@@ -5,7 +5,6 @@ This object is not returned by any endpoints yet. It is here only as a reference
 Field           | Type    | Description
 ----------------|---------|------------------------------------------------------------
 colour          | string? | |
-description     | string? | |
```
nanaya, line 8: Where's the updated doc for this

nanaya (changes requested):

resources/views/docs/_structures/group.md
cl8n: well there's no way to access it from api anymore, since there is no group endpoint and it's not included to usergroup

nanaya (changes requested):

nanaya (approved):

PR #7683, opened 2021-06-03
Absolute url for default path of user avatar

### LINKED ISSUE #7390: User avatar URL is returned as relative path for guests
In API responses, the avatar URL is usually returned as an absolute url:

`https://a.ppy.sh/9958822?1615128896.jpeg`

But for guests (I guess whenever a user doesn't have an avatar set), it's returned as a relative URL:

`/images/layout/avatar-guest.png`

This is a bit confusing, so I'd propose we change this to also/always be a full absolute URI.

As pointed out in https://github.com/ppy/osu/pull/12204#discussion_r602893245.

nanaya:
Resolves #7390.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -74,6 +74,7 @@ SLACK_ENDPOINT=https://myconan.net/null/
 # S3_AVATAR_BASE_URL=
 # AVATAR_STORAGE=local-avatar
 # AVATAR_CACHE_PURGE_PREFIX=
+# DEFAULT_AVATAR=http://localhost/images/layout/avatar-guest.png
 
 # Either "s3" or "local".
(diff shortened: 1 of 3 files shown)


notbakaneko (approved):

notbakaneko (changes requested):
Actually, the one in `user.ts` needs to be updated too? Then there's `avatar.less`...

nanaya:
Neither are supposed to be accessed from outside the web. The one in `avatar.less` isn't related as guest avatar may be different to user default avatar (which could be useful for browser test).

peppy (approved):

notbakaneko (approved):

PR #7688, opened 2021-06-03
Show participated lazer multiplayer games in user profile

notbakaneko:
Show games the user has participated in a new tab on the user profile

- [x] need to fix some text
- [x] some of the styles look like they can be consolidated with other components? (pretty sure some of the existing coffeescript can also be refactored into components but I'm not touching that yet 👀 )


nanaya (changes requested):
Not actually done reviewing but can use update for at least the user lookup part

nanaya (changes requested):

resources/assets/less/bem/multiplayer-room.less
```diff
@@ -0,0 +1,119 @@
...
+
+  &__cover {
+    .full-size();
+    object-fit: cover;
+    border-radius: var(--border-radius);
+
+    &--default {
+      .at2x('~@images/layout/beatmaps/default-bg.png');
+      background-position: center;
+      background-size: cover;
```
nanaya, line 39: Should be repeating pattern and not oversized icons...

resources/assets/less/bem/multiplayer-room.less
```diff
@@ -0,0 +1,119 @@
...
+    flex-wrap: wrap;
+    justify-content: space-between;
+  }
+
+  &__badge-container {
+    display: flex;
+    align-items: center;
+
+    &--bottom {
+      margin-top: 6px;
```
nanaya, line 60: seems a bit too much now

resources/assets/lib/user-multiplayer-index/room.tsx
```diff
@@ -0,0 +1,135 @@
...
+    if (beatmapset == null) return undefined;
+
+    return beatmapset.covers.cover;
+  }
+
+  render() {
+    const endsAt = moment(this.props.room.ends_at);
+
+    return (
+      <div className='multiplayer-room'>
```
nanaya, line 65: apparently I forgot about it but a link to the room detail could be useful? (`/multiplayer/rooms/:room`)

nanaya (changes requested):

resources/assets/less/bem/multiplayer-room.less
nanaya: so apparently it's fine in card because the max size in there is about the same as the image itself :upside_down_face: while in here it goes all the way to 700+px

resources/assets/lib/user-multiplayer-index/room.tsx
notbakaneko: I thought it's supposed to go to lazer instead 🤔

nanaya: well, any link is better than nothing (and can be updated later as needed)

peppy: Won't that endpoint not work due to the rooms being in a completely different table structure?

nanaya: that's the new multiplayer room leaderboard endpoint

resources/lang/en/multiplayer.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+return [
+    'empty' => 'No osu!lazer multiplayer games played yet!',
```
peppy, line 7:
```suggestion
    'empty' => 'No osu!(lazer) multiplayer games played yet!',
```

Trying not to make the "osu!lazer" term stick since it's only a codename and will eventually get dropped.

I guess if we end up having this in more areas, it may be better to add it as a blurb text (or tooltip) instead? Something like "This data is currently only populated from newer releases of osu!".

resources/assets/lib/user-multiplayer-index/room.tsx
peppy: Damn, totally forgot this was a thing!

nanaya (changes requested):

nanaya (approved):

PR #7694, opened 2021-06-05
Decode uri in ziggy route

### LINKED ISSUE #7693: Prefix link `/wiki/` is disappeared on BN General Information
currently, the `/wiki/` on [EN] [BN General Information](https://osu.ppy.sh/wiki/en/People%2FThe_Team%2FBeatmap_Nominators%2FGeneral_Information) that should be stated to the site is disappeared 
but in the GitHub osu-wiki is already stated.

gagahpangeran:
Closes #7693 

The real problem in that issue is the link of suggestion result is not decoded.

Not sure if I want to decode inside the `route` function itself or do it locally at this code.

https://github.com/ppy/osu-web/blob/e43a7eca9384e2ab32b0ed45d0de1a61fc967b8e/resources/assets/lib/wiki-search.tsx#L109

### DIFF
--- a/resources/assets/lib/laroute.ts
+++ b/resources/assets/lib/laroute.ts
@@ -20,7 +20,7 @@ export function route(name: string, params?: Attributes | null, absolute?: boole
     params = {};
   }
 
-  return ziggyRoute(name, params, absolute, Ziggy).toString();
+  return decodeURIComponent(ziggyRoute(name, params, absolute, Ziggy).toString());
 }
 
 export function link_to_route(name: string, text: string, params?: Attributes | null, attrs?: Attributes | null) {


nanaya:
This breaks query string (`decodeURIComponent(laroute.route('beatmapsets.search',{q:'hello&'}))` see how the last `&` sign becomes query string separator instead of `%26`).

ziggy is actually doing the right thing; it's laravel route function that's being incompetent of not escaping passed parameter and I incorrectly relied on that. (for example, `route('users.show',['user'=>'hello?world', 'mode'=>'osu'])` generates completely wrong url)

Wiki URL generation should probably not rely on normal url generator but do plain string append instead and the controller should redirect accesses on page which path contains `%2F`.

gagahpangeran:
Hmm, you're right.

For now should I move the decode uri locally in the code I mention above or just close this PR and work on general fix as you mentioned?

nanaya:
I wouldn't even pass the path to route but instead instead append (if passing empty string) or replace (if passing placeholder) the generated url.

notbakaneko (changes requested):
Is this abandoned?

gagahpangeran:
Oh, I even forgot about this PR.

gagahpangeran:
Closing because I'm not interested anymore to fix the issue. Anyone can pick it up.

PR #7696, opened 2021-06-07
Better error handling when downloading beatmap archive

notbakaneko:
`file_get_contents()` can fail with `false` and `fwrite` will still gladly write a 0 byte file which `ZipArchive` may or may not open depending on `libzip` version.


peppy (approved):

app/Models/BeatmapsetArchive.php
```diff
@@ -5,13 +5,18 @@
...
+
 class BeatmapsetArchive
 {
     public function __construct(string $osz)
     {
         $this->osz = $osz;
         $this->zip = new \ZipArchive();
-        $this->zip->open($this->osz);
+        $code = $this->zip->open($this->osz);
+        if ($code !== true) {
```
nanaya, line 17: `__destruct`'s `close` call can use a check if it's open before trying to close otherwise it'll throw another error (at least in console).

```
>>> $b = new App\Models\BeatmapsetArchive('/tmp/junk')
App\Exceptions\BeatmapProcessorException with message 'Failed to open archive'
>>>

Exit:  Ctrl+D
PHP Fatal error:  Uncaught ValueError: Invalid or uninitialized Zip object in /home/edho/osu/osu-web/app/Models/BeatmapsetArchive.php:24
Stack trace:
#0 /home/edho/osu/osu-web/app/Models/BeatmapsetArchive.php(24): ZipArchive->close()
#1 [internal function]: App\Models\BeatmapsetArchive->__destruct()
#2 {main}
  thrown in /home/edho/osu/osu-web/app/Models/BeatmapsetArchive.php on line 24
```

app/Models/BeatmapsetArchive.php
```diff
@@ -5,13 +5,18 @@
...
+use App\Exceptions\BeatmapProcessorException;
+
 class BeatmapsetArchive
 {
     public function __construct(string $osz)
     {
         $this->osz = $osz;
         $this->zip = new \ZipArchive();
-        $this->zip->open($this->osz);
+        $code = $this->zip->open($this->osz);
```
nanaya, line 16: not actually relevant to this pr but this class can use some property declarations

nanaya (changes requested):

nanaya (approved):

PR #7701, opened 2021-06-08
Remove duplicate route names

nanaya:
This causes error on laravel 8.

A bit hacky for redirect as there are two endpoints pointing to same route.


notbakaneko (approved):
"it works"™️

PR #7703, opened 2021-06-08
Support permalinking to beatmap discussion replies

### LINKED ISSUE #3150: Allow linking of replies in mod v2
Replies to posts I mean
If a thread for a particular issue gets large, it is currently impossible to link to a specific point or two, making it more difficult to reference other replies if there is an issue large enough to get so much discussion.

notbakaneko:
closes #3150

Just the linking for now, the highlighting needs a bit of separate fixing


resources/assets/coffee/_classes/beatmap-discussion-helper.coffee
```diff
@@ -188,8 +189,17 @@ class @BeatmapDiscussionHelper
         params.beatmap = discussionState.beatmapId
         params.mode = discussionState.mode
 
+      if post?
+        postId = post.id
```
nanaya, line 193: is this actually needed or can be `hash = ...#{post.id}... if post?`

nanaya (changes requested):

nanaya (approved):

PR #7704, opened 2021-06-08
Fix UserGroup group attribute return type

notbakaneko:
`Groups::byId` returns `?Group`, `group` should return the same.


notbakaneko:
or maybe it doesn't 🤔

PR #7715, opened 2021-06-10
Memoize user group lookup

nanaya:
Looks like could be useful :thinking: 

I haven't measured it (esp for single group user) but considering accessing relation as attribute already involves extra processing it probably balances out :shrug:


app/Models/User.php
```diff
@@ -951,12 +951,12 @@ public function groupIds()
...
-        $groupId = $group->getKey();
+        $byGroupId = $this->memoize(__FUNCTION__.':byGroupId', fn () => $this->userGroups->keyBy('group_id'));
 
-        foreach ($this->userGroups as $userGroup) {
-            if ($userGroup->group_id === $groupId) {
-                return $activeOnly && $userGroup->user_pending ? null : $userGroup;
-            }
+        $userGroup = $byGroupId->get($group->getKey());
+
+        if ($userGroup !== null && (!$activeOnly || !$userGroup->user_pending)) {
```
notbakaneko, line 958: I wonder if inverting the condition to `!($activeOnly && $userGroup->user_pending)` or 
```php
if ($userGroup === null || ($activeOnly && $userGroup->user_pending)) {
    return null;
}
```
would be easier to read?

notbakaneko (approved):

PR #7721, opened 2021-06-11
Unify user lookup logic for profile page

nanaya:
The one in modding page was doing an always-true check before ( ﾟ ヮﾟ)

Pushed to wrong branch so I figured might as well ignore the fix which is being removed anyway ( ﾟ ヮﾟ)


app/Http/Controllers/UsersController.php
```diff
@@ -71,7 +73,11 @@ public function __construct()
 
     public function card($id)
     {
-        $user = $this->lookupUser($id) ?? UserNotFound::instance();
+        try {
+            $user = FindForProfilePage::find($id);
+        } catch (UserProfilePageLookupException $e) {
+            $user = UserNotFound::instance();
+        }
```
notbakaneko, line 80: This is now weird as it goes through `lookupWithHistory`, finds a matching user, and then renders not found if not explicitly requested with json, even if it always returns json.

app/Exceptions/UserProfilePageLookupException.php
```diff
@@ -0,0 +1,21 @@
...
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace App\Exceptions;
+
+class UserProfilePageLookupException extends SilencedException
+{
+    private $render;
+
+    public function __construct(callable $render)
```
notbakaneko, line 12: I wonder if renderable exception instead of passing a `callable` would actually be useful 🤔 (or make even less sense 🤷 )

notbakaneko (changes requested):

app/Exceptions/UserProfilePageLookupException.php
nanaya: maybe if we find another thing needing that...

notbakaneko (approved):

PR #7725, opened 2021-06-12
Add report score menu in score info page

### LINKED ISSUE #7541: [Feature Request] Report a score from their score page
Score pages look like this: https://osu.ppy.sh/scores/osu/3626112929
image

It can be put on the left of "Download Replay" in the red color that is used for "Send Report"
image

Examples:
image
image
And it can have the same exclamation triangle as the profile report button:
image

gagahpangeran:
closes #7541 

I decided to implement the menu locally because `PlayDetailMenu` has menu item 'View Details' and 'Download Replay' and I thought it's pretty redundant in this page.


notbakaneko (changes requested):
The button should have a hover effect like the user card one on the page.

resources/assets/less/bem/score-buttons.less
```diff
@@ -0,0 +1,18 @@
...
+  display: flex;
+
+  &__menu {
+    .circle(35px);
+    margin-left: 10px;
+    background-color: hsla(var(--hsl-b6), 0.75);
+    position: relative;
+    transition: background-color 120ms ease-in-out;
+
+    &:hover {
```
notbakaneko, line 14: should use `.link-hover`

notbakaneko (changes requested):

notbakaneko (approved):

PR #7726, opened 2021-06-12
Add link and playmode tooltip to group badges

### LINKED ISSUE #6936: Make usergroup badges redirect to respective group listings
Think it would be very handy. Can code an implementation if deemed appropriate.

### LINKED ISSUE #7706: Add gamemode to BN and Loved Group tooltips
The icons, especially standard and taiko are hard to differentiate on small scales, so having the gamemode in the tooltip would help

cl8n:
closes #7706 and closes #6936


resources/assets/lib/user-group-badge.tsx
```diff
@@ -15,28 +16,35 @@ export default function UserGroupBadge({group, modifiers}: Props) {
...
+  }
+
+  const props = {
+    'children': group.playmodes != null && group.playmodes.length > 0 && (
+      <div className={'user-group-badge__modes'}>
+        {group.playmodes.map((playmode) => (
+          <i key={playmode} className={`fal fa-extra-mode-${playmode}`} />
+        ))}
+      </div>
+    ),
```
notbakaneko, line 36: can just assign to an optional variable in above block

resources/assets/lib/user-group-badge.tsx
```diff
@@ -15,28 +16,35 @@ export default function UserGroupBadge({group, modifiers}: Props) {
...
+      probationary: group.is_probationary,
+      [group.identifier]: true,
+      ...modifiers,
+    }),
+    'data-label': group.short_name,
+    'style': osu.groupColour(group),
+    title,
+  };
+
+  return group.id < 0
```
notbakaneko, line 47: where does this negative group id come from?

notbakaneko (changes requested):

resources/assets/lib/user-group-badge.tsx
cl8n: mapper and guest group

cl8n: ok actually I just remembered this was the whole reason I added `has_listing` to groups a while ago... will use that instead and  update blade component too

resources/views/objects/_user_badge.blade.php
```diff
@@ -0,0 +1,44 @@
...
+    @endif
+</{{ $tag }}>
+
+@if ($forum && $hasPlaymodes)
+    <div class="forum-post-info__row forum-post-info__row--group-badge-playmodes">
+        @foreach ($playmodes as $playmode)
+            <i class="fal fa-extra-mode-{{$playmode}}"></i>
+        @endforeach
+    </div>
+@endif
```
notbakaneko: The forum-specific parts should stay on the forum blades.

resources/views/objects/_user_badge.blade.php
```diff
@@ -0,0 +1,44 @@
...
+    $title = $group->group_name;
+
+    if ($hasPlaymodes) {
+        $playmodeNames = implode(', ', array_map(fn ($mode) => trans("beatmaps.mode.{$mode}"), $playmodes));
+        $title .= " ({$playmodeNames})";
+    }
+@endphp
+
+<{{ $tag }}
+    class="{{ class_with_modifiers('user-group-badge', $forum ? ['t-forum'] : []) }}"
```
notbakaneko: should pass in the list of modifiers instead.

notbakaneko (changes requested):

resources/assets/lib/user-group-badge.tsx
```diff
@@ -15,28 +16,38 @@ export default function UserGroupBadge({group, modifiers}: Props) {
...
+    const playmodeNames = group.playmodes
+      .map((playmode) => osu.trans(`beatmaps.mode.${playmode}`))
+      .join(', ');
+
+    title += ` (${playmodeNames})`;
+  }
+
+  const props = {
+    children,
+    'className': classWithModifiers('user-group-badge', {
```
notbakaneko, line 40: also we just changed the lint rule so some of these quotes aren't needed anymore >_>

notbakaneko (approved):

PR #7727, opened 2021-06-12
Show mod filter in scoreboard for non supporters

### LINKED ISSUE #7556: Beatmap info leaderboards don't show mod filter for non-supporters
Screen Shot 2021-05-04 at 4 36 32 PM

The rule of thumb is to show every supporter feature but just disable them. The options are:
- disable the selector with a tooltip
- show the leaderboard notice + "and mod filter" after a user trys to filter w/o supporter
Screen Shot 2021-05-04 at 4 39 45 PM

gagahpangeran:
Closes #7556


resources/lang/en/beatmapsets.php
```diff
@@ -142,7 +142,7 @@
             'friend' => 'Friend Ranking',
             'global' => 'Global Ranking',
             'supporter-link' => 'Click <a href=":link">here</a> to see all the fancy features that you get!',
-            'supporter-only' => 'You need to be an osu!supporter to access the friend and country rankings!',
+            'supporter-only' => 'You need to be an osu!supporter to access the friend and country rankings or mod spesific rankings!',
```
cl8n, line 145:
```suggestion
            'supporter-only' => 'You need to be an osu!supporter to access the friend, country, or mod-specific rankings!',
```

resources/assets/coffee/react/beatmapset-page/main.coffee
```diff
@@ -90,6 +90,13 @@ export class Main extends React.Component
     else
       @state.enabledMods
 
+    @setState
+      enabledMods: enabledMods
```
cl8n, line 94: `enabledMods` and `currentScoreboardType` should be reset to previous if scoreboard request fails

cl8n, line 94: the check for not-supporter viewing not-global leaderboard (L80) needs to come after this, else the mods menu in that case looks accessible but does nothing on clicks

cl8n (changes requested):

resources/assets/coffee/react/beatmapset-page/main.coffee
gagahpangeran: Merge the conditional.

gagahpangeran: Looks like `currentScoreboardType` not reset to previous state when failed is also a bug in current master.

cl8n (dismissed):
i kinda liked the behavior (well, if resetting were fixed) of scoreboard&mods selector before 7e624162c39f44bc2d15dc550ec1dc57105bbd9a, where it would show your selection & let u edit selection before completing request, but not really in scope of this change I guess.

nanaya:
Conflicts.

gagahpangeran:
Resolved.

resources/assets/coffee/react/beatmapset-page/main.coffee
```diff
@@ -74,13 +74,6 @@ export class Main extends React.Component
   }) =>
     @scoreboardXhr?.abort()
 
-    @setState
-      currentScoreboardType: scoreboardType
```
nanaya, line 78: this should stay otherwise it's annoying not seeing the option change after clicking.

If anything, this should also set enabled mods; save previous type and mods to a variable beforehand and restore it on fail.

nanaya (changes requested):

resources/assets/coffee/react/beatmapset-page/main.coffee
gagahpangeran: This cause flickering when fail fast enough.

flickmod

nanaya: that seems like a very rare case or just dead internet

gagahpangeran: Okay, I just pushed it.

nanaya (changes requested):

resources/assets/coffee/react/beatmapset-page/main.coffee
```diff
@@ -90,6 +90,14 @@ export class Main extends React.Component
     else
       @state.enabledMods
 
+    @setState
+      enabledMods: enabledMods
```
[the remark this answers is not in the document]
gagahpangeran: > Then restoring it is just `@setState(prevState)`

Ah, so how to do it in coffeescript.

Okay, done.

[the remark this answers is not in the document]
nanaya: ...there's nothing coffeescript about that :raised_eyebrow:

nanaya (approved):

PR #7728, opened 2021-06-12
Add multi accounting in user report option

### LINKED ISSUE #7626: Add multiaccounting as a user report option
Currently, a lot of people are reporting for multiaccounts with reason "other", which is kinda spammy on #web-reports listing, would make sense to have a custom option for that and have it in separate channel that's not visible to mods (see like cheating reports) as they are not responsible for handling those cases.

gagahpangeran:
Closes #7626


resources/lang/en/users.php
```diff
@@ -132,6 +132,7 @@
             'spam' => 'Spamming',
             'unwanted_content' => 'Linking inappropriate content',
             'nonsense' => 'Nonsense',
+            'multi_accounting' => 'Multi-Accounting',
```
cl8n, line 135:
```suggestion
            'multi_accounting' => 'Using multiple accounts',
```
"multi-accounting" feels like a slang to me
not sure if important 🤔

cl8n (changes requested):

resources/lang/en/users.php
gagahpangeran: Yeah, open for suggestion for this one.

tybug: fwiw, there are multiple mentions to "multi-accounting" on the wiki: https://osu.ppy.sh/wiki/en/Help_Centre/Account_Restrictions

nanaya (dismissed):
Seems fine to me but people are saying this option may not actually be needed in the first place? :shrug:

gagahpangeran:
Ping @peppy for confirmation.

peppy:
If this is to be added then the whole foul play thing is pointless / wrong. I didn't realise that was already there as an option. Requires further discussion, especially with people that actually work with the system, so figure if this split is actually required in the first place.

peppy:
Based on the original reason for creating, I'd suggest that we either remove "Other" as an option completely, or update the "Foul Play / Cheating" to just read "Cheating".

gagahpangeran:
> Based on the original reason for creating, I'd suggest that we either remove "Other" as an option completely, or update the "Foul Play / Cheating" to just read "Cheating".

Is this the final decision?

peppy:
I'm open to more discussion, especially from moderators/admins that actually handle these reports.

Ephemeralis:
More delineation in the reporting options like this is not a bad thing at all and will help the people processing things to do so more efficiently.

Agree completely that "Foul Play / Cheating" should be renamed to just "Cheating". I suspect a lot of ESL users are not really familiar with the term foul play anyway.

gagahpangeran:
> More delineation in the reporting options like this is not a bad thing at all and will help the people processing things to do so more efficiently.
> 
> Agree completely that "Foul Play / Cheating" should be renamed to just "Cheating". I suspect a lot of ESL users are not really familiar with the term foul play anyway.

So, how about 'Other' options, should we keep it or removed it?

Also, while you at it, please give suggestion for https://github.com/ppy/osu-web/pull/7728#discussion_r650464264

RockRoller01:
Adding a multi account option and filtering them to the cheating endpoint seems like a good step to me. Multi accounting isnt really handled by the GMT afterall but rather the support team, however we still get all the reports in, which arent relevant for us.

From a quick search it looks like most of the reports using the reason "Other" are for multi accounts, while nearlly the whole rest of them is for profile pictures and usernames, which should be using UnwantedContent anyways, but other seems to be occasionally used for content that doesnt fit any other category (the last 3 days had like 2 reports or so where even I dont know of a better category).

Leaving other shouldn't hurt and is occasionally useful, and everyone that would want to report a multi account would probably select the new option directly anyways.

gagahpangeran:
I rename "Foul Play / Cheating" to just "Cheating" based on discussion above.

database/migrations/2021_06_12_145807_add_multi_accounting_report.php
```diff
@@ -0,0 +1,44 @@
...
+     */
+    public function up()
+    {
+        DB::statement("ALTER TABLE osu_user_reports CHANGE reason reason ENUM(
+            'Insults',
+            'Spam',
+            'Cheating',
+            'UnwantedContent',
+            'Nonsense',
+            'MultiAccounting',
```
peppy, line 23: Let's call this `MultipleAccounts` instead.

Also I'm not sure if you've checked whether it matters about order of adding into an enum in mysql, but I think putting it last would probably feel safer to me.

peppy (changes requested):

database/migrations/2021_06_12_145807_add_multi_accounting_report.php
gagahpangeran: Got it!

peppy (approved):

app/Models/UserReport.php
```diff
@@ -35,7 +35,7 @@ class UserReport extends Model
     use RoutesNotifications, Validatable;
 
     const POST_TYPE_REASONS = ['Insults', 'Spam', 'UnwantedContent', 'Nonsense', 'Other'];
-    const SCORE_TYPE_REASONS = ['Cheating', 'Other'];
+    const SCORE_TYPE_REASONS = ['Cheating', 'MultipleAccounts', 'Other'];
```
nanaya, line 38: Reporting multi account on score is fine but also need to be added to `availableOptions` in `report-reportable.tsx`

nanaya (changes requested):

nanaya (approved):

PR #7729, opened 2021-06-13
Show usercard in supporter store cart

### LINKED ISSUE #7530: Ability to view usercards for osu!supporter product in store cart
I want to confirm the username is right by seeing their avatar/cover. You can see their usercard before putting in cart, but if you are buying at a later date, you may want to check again without copying and pasting name in search. Some users can have like a one letter difference so this is a UX improvement.

Screen Shot 2021-04-25 at 11 20 47 PM

gagahpangeran:
Closes #7530


resources/views/store/cart/_item.blade.php
```diff
@@ -13,7 +13,7 @@
     <input type="hidden" name="item[id]" value="{{ $item->id }}">
 
     <span class="store-order-item__name">
-        {{ $item->getDisplayName() }}
+        {!! $item->getDisplayName() !!}
```
nanaya, line 16: call `blade_safe(...)` for supporter tag display name instead

nanaya (changes requested):

resources/views/store/cart/_item.blade.php
gagahpangeran: Done

nanaya (approved):

PR #7736, opened 2021-06-14

peppy:


peppy:
@nanaya not sure we wanted to nuke the pending/ranked beatmap text? it looks like the translations already had the correct string. did you have a method of restoring these easily (i think it happened once or twice before)

nanaya:
I've renamed existing translations.  The end result is "Ranked & Approved" ones stay as well though (instead of just "Ranked").

peppy:
I think that's fine; those can be updated as we go. Will try to re-export.

PR #7740, opened 2021-06-14
Explain type assumption for `UserGroup::$group` in comment

cl8n:
just a short note, should have left it when I wrote this because the type is technically wrong (#7704)


app/Models/UserGroup.php
```diff
@@ -36,6 +36,8 @@ public function user()
 
     public function getGroupAttribute(): Group
     {
+        // Assumes that `group_id` refers to an existing group that is present
+        // in `app('groups')` cache
```
nanaya, line 40: I think it's more of "group_id of a user should always be valid"

nanaya (changes requested):

notbakaneko (approved):

PR #7741, opened 2021-06-14
Fix beatmap discussion mode switcher not auto scrolling in mobile when first load

### LINKED ISSUE #7207: Can't distinguish issue types on modding pages on mobile
Here's an example: what's usually on the side is not visible at all on mobile, so when I get a new notification about my map I can't tell whether it's a problem, suggestion, praise, or hype except by reading the text and guessing.

If it helps, firefox on Pixel 3xl

gagahpangeran:
The bottom bar is actually there, but hidden.

Also don't forget about [react lifecycle](https://reactjs.org/docs/state-and-lifecycle.html).


notbakaneko (changes requested):
How does this solve the linked issue and what does this do besides breaking the autofocus on the mode switcher? 😕

gagahpangeran:
The exact issue is not missing bottom bar, it always be there.

But when you open, for example https://osu.ppy.sh/beatmapsets/1113149/discussion, in mobile, the mode switcher does not scroll to it.

nanaya:
isn't the referenced issue about discussion type not visible in mobile?

gagahpangeran:
Hoo, I thought it's different issue, my bad.

So the linked issue is not relevant. But this is a fix for another issue.

resources/assets/coffee/react/beatmap-discussions/mode-switcher.coffee
```diff
@@ -16,7 +16,7 @@ export class ModeSwitcher extends React.PureComponent
     @scrollerRef = React.createRef()
 
 
-  componentDidUpdate: =>
+  componentDidMount: =>
```
notbakaneko, line 19: `componentDidUpdate` still needs to call `scrollTo` to bring the clicked item completely into view when it's only partially visible

https://user-images.githubusercontent.com/1694544/121883720-0da49d80-cd4d-11eb-877e-9a82a05c5452.mov

notbakaneko (changes requested):

resources/assets/coffee/react/beatmap-discussions/mode-switcher.coffee
gagahpangeran: Yep, just notice about that.

notbakaneko (approved):

PR #7752, opened 2021-06-16
Add osu! "Approach Different" mod

smoogipoo:
See: https://github.com/ppy/osu/pull/13514


peppy (approved):

notbakaneko (commented):
The order these are declared in needs to be fixed...

peppy:
@notbakaneko can you elaborate? is there some immediate issue?

notbakaneko:
The keys should be in alphabetical order but it's already an issue with the existing keys, so it doesn't necessarily have to be fixed here.

PR #7753, opened 2021-06-17
Allow profile page for bot

nanaya:
Not sure if should put back the check for modding profile because it's kind of visible already from the overall listing page.


notbakaneko (approved):
I don't think bots putting automated messages should have a modding section on the profile, though 🤔

PR #7758, opened 2021-06-17
Unify local cache functions and reset accordingly on job worker

nanaya:
Will need to add another reset when adding octane. Probably the swoole ticker thing every minute or so.

Also renamed OsuAuthorize's `cacheReset` to match the others.

<del>Wait no, somewhere along the way I ended up with wrong design ಠ_ಠ The version check isn't supposed to happen on validation (it shouldn't even be called validation).</del> This sure seems as convoluted as it was in my mind...


app/Libraries/Groups.php
```diff
@@ -6,41 +6,35 @@
...
 
     /**
      * Get all groups.
      */
     public function all(): Collection
     {
-        return $this->memoize(__FUNCTION__, function () {
-            return $this->fetch();
-        });
+        return $this->memoize(__FUNCTION__, fn () => $this->cachedFetch());
```
notbakaneko, line 21: I can't help but think this should be a different function that instead that accepts what would go in `fetch()` as the callable instead of `fetch()` being an abstract method.

`cachedFetch()` seems to exist just to call `fetch()` (which needs to be implemented) and be memoized

notbakaneko (changes requested):

notbakaneko (approved):

PR #7762, opened 2021-06-18
Change beatmap discussion user report to link directly to post

### LINKED ISSUE #7759: Beatmap discussions reports should use new permalinks
Currently reports we moderators receive still just link to the main post of a discussion and not to the reported subpost, leading to tidious searching of the reported post on longer discussions. Would be good if those reports could use the new permalinks instead.

notbakaneko:
closes #7759

Also tweaks the jump to behaviour to go the discussion instead of the post if it's known to be the first post.


app/Http/Controllers/BeatmapDiscussionPostsController.php
```diff
@@ -26,7 +26,7 @@ class BeatmapDiscussionPostsController extends Controller
 {
     public function __construct()
     {
-        $this->middleware('auth', ['except' => 'index']);
+        $this->middleware('auth', ['except' => 'index', 'show']);
```
nanaya, line 29: there are missing braces there resulting in auth middleware still being used for `show` action :upside_down_face:

nanaya (changes requested):

nanaya (approved):

PR #7765, opened 2021-06-18
Add query function to return result with pagination check

nanaya:
Looks like it could be useful? :thinking:


app/Models/Model.php
```diff
@@ -33,11 +34,13 @@ public function getForeignKey()
...
     public function getMacros()
     {
-        $macros = $this->macros ?? [];
-        $macros[] = 'realCount';
-        $macros[] = 'last';
+        static $baseMacros = [
+            'getWithHasMore',
+            'last',
+            'realCount',
+        ];
```
notbakaneko, line 41: could also have declared it has `protected const` and used `ReflectionClass::getConstants(ReflectionClassConstant::IS_PUBLIC)` since it's php8 now 🤔

nanaya: that's just adding even more hacks

notbakaneko (approved):

PR #7768, opened 2021-06-21
Return UserRelation instead of UserCompact for friends API

### LINKED ISSUE #7637: Change /api/v2/friends to return collection of UserRelation
Currently, the `/api/v2/friends` enpoint returns a collection of `UserCompact`. I think this should be changed to return a collection of `UserRelation` with the `target` included, this way you can get the relation and the user as well. This will be useful for applications that want to display friends along with their mutual status.

cl8n:
resolves #7637

### DIFF
--- a/app/Http/Controllers/FriendsController.php
+++ b/app/Http/Controllers/FriendsController.php
@@ -42,7 +42,7 @@ class FriendsController extends Controller
      *
      * ### Response Format
      *
-     * A collection of [UserRelation](#userrelation) objects with `target` included. `target`s include `cover`, `country`, `groups`, and `support_level`.
+     * A collection of [UserRelation](#userrelation) objects with `target` included. `target`s include `country`, `cover`, `groups`, and `support_level`.
      *
      * @response [
      *   {
@@ -115,8 +115,8 @@ class FriendsController extends Controller
                 'UserRelation',
                 [
                     'target',
-                    'target.cover',
                     'target.country',
+                    'target.cover',
                     'target.groups',
                     'target.support_level',
                 ],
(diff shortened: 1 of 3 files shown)


nanaya:
It looks like this will need to be coordinated with lazer.

cl8n:
yep. I'll wait for review on this first to get the API decided

I also removed the `statistics` include from the users in this response because I don't think it made sense to use the auth'd user's default mode (`friends.read` doesn't imply knowledge of that)... but lazer does use that info currently. should I add it back and require `identify` scope too or something?

peppy:
Fine to go ahead with this at your discretion (assuming this seems like a sane change - I think it does but haven't looked into it with too much thought).

Will coordinate the lazer side changes if/when this is merged.

app/Http/Controllers/FriendsController.php
```diff
@@ -30,8 +33,96 @@ public function __construct()
...
+     * ]
+     */
     public function index()
     {
+        if (is_api_request()) {
+            return json_collection(
+                auth()->user()
+                    ->relations()
+                    ->friends()
+                    ->withMutual()
```
nanaya, line 112: Make its own relation and load that instead.

```php
// in User model
public function relationFriends()
{
  return $this->relations()->friends()->withMutual();
}

// in this controller action
// the loaded reloaded data will be used regardless (in json for api, in current user default json for html)
$relations = $currentUser->relationFriends;
$relations->load(['target' => fn ($q) => $q->...]);

if (is_api...) {
    ...
}

$user = $relations->pluck('target');

// in Transformers etc
$user->relationFriends

nanaya (changes requested):

peppy:
@cl8n still willing to apply the review to this one?

nanaya:
The main thing has been merged through a different PR. Feel free to open another PR for the documentation (or maybe wait for the documentation system update PR (#11433) first?).

PR #7772, opened 2021-06-22
Add vote progress in art contest gallery

### LINKED ISSUE #6355: Add number of remaining votes when looking at contest entry in lightbox view
When viewing all the entries, I usually cycle through all of them in lightbox view. Seeing the number of remaining votes would help to decide whether an entry should receive a vote or rather not. It could look something like this:

gagahpangeran:
Closes #6355

votevote

### DIFF
--- a/resources/assets/less/photoswipe-overrides.less
+++ b/resources/assets/less/photoswipe-overrides.less
@@ -32,6 +32,8 @@
       display: flex;
       align-items: center;
       justify-content: center;
+      opacity: 1;
+      cursor: default;
 
       && {
         background-image: none;
--- a/resources/assets/less/photoswipe-overrides.less
+++ b/resources/assets/less/photoswipe-overrides.less
@@ -12,7 +12,7 @@
     opacity: 0.75;
     color: #fff;
 
-    &--contest-vote, &--vote-progress {
+    &--contest-vote {
       /*
        * Overrides original sytle.
        * The second selector is needed to override retina version that is slightly
@@ -26,6 +26,16 @@
     &--contest-vote-active {
       opacity: 0.9;
       color: @contest-selected-colour;
+    }
+
+    &--vote-progress {
+      display: flex;
+      align-items: center;
+      justify-content: center;
+
+      && {
+        background-image: none;
+      }
     }
 
     // last so it overrides everything else
--- a/resources/assets/less/photoswipe-overrides.less
+++ b/resources/assets/less/photoswipe-overrides.less
@@ -12,7 +12,7 @@
     opacity: 0.75;
     color: #fff;
 
-    &--contest-vote {
+    &--contest-vote, &--vote-progress {
       /*
        * Overrides original sytle.
        * The second selector is needed to override retina version that is slightly
--- a/resources/assets/lib/circular-progress.tsx
+++ b/resources/assets/lib/circular-progress.tsx
@@ -5,7 +5,6 @@ import * as React from 'react';
 import { classWithModifiers } from './utils/css';
 
 interface Props {
-  countUp: boolean;
   current: number;
   ignoreProgress: boolean;
   max: number;
@@ -16,7 +15,6 @@ interface Props {
 
 export class CircularProgress extends React.PureComponent<Props, any> {
   static defaultProps = {
-    countUp: false,
     ignoreProgress: false,
     onlyShowAsWarning: false,
     theme: '',
@@ -33,10 +31,6 @@ export class CircularProgress extends React.PureComponent<Props, any> {
       return null;
     }
 
-    const displayNumber = this.props.countUp
-      ? this.props.current
-      : this.props.max - this.props.current;
-
     const transform = this.props.ignoreProgress
       ? undefined
       : { transform: `rotate(${percentage}turn)` };
@@ -51,7 +45,7 @@ export class CircularProgress extends React.PureComponent<Props, any> {
         })}
         title={this.props.tooltip || `${this.props.current} / ${this.props.max}`}
       >
-        <div className={`${bn}__label`}>{displayNumber}</div>
+        <div className={`${bn}__label`}>{this.props.max - this.props.current}</div>
         <div className={`${bn}__slice`}>
           <div className={`${bn}__circle`} style={transform} />
           <div className={`${bn}__circle ${bn}__circle--fill`} />
--- a/resources/assets/lib/circular-progress.tsx
+++ b/resources/assets/lib/circular-progress.tsx
@@ -5,6 +5,7 @@ import * as React from 'react';
 import { classWithModifiers } from './utils/css';
 
 interface Props {
+  countUp: boolean;
   current: number;
   ignoreProgress: boolean;
   max: number;
@@ -15,6 +16,7 @@ interface Props {
 
 export class CircularProgress extends React.PureComponent<Props, any> {
   static defaultProps = {
+    countUp: false,
     ignoreProgress: false,
     onlyShowAsWarning: false,
     theme: '',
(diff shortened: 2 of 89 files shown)


peppy:
Not sure if it's just me, but I feel like it should be counting down, not up..

Looks pretty good otherwise. @arflyte can you check that you're okay with this design?

arflyte:
Yeah, looks good to me. I think counting up is fine. Counting down might be a lil bit confusing and requires explanation.

nanaya:
Isn't knowing how many votes available more useful than how many votes already placed?

arflyte:
Oh, I wasn't aware there's a vote limit. I thought it was just progress overall, not votes left. For vote availability then yes, it makes more sense for it to go down.

gagahpangeran:
Okay, I'll change it to count down.

gagahpangeran:
Already change to count down and update gif in OP.

peppy:
I'm so confused now. Why does the tooltip show it going up (and the progress fill)? There's so much contradicting information. I think the tooltip should read something like "Remaining votes: 9". The fill can probably stay as it is, since it's mostly just flair at this point.

gagahpangeran:
Hmm, I just use the existing API of `CircularProgress`. Maybe need a bit change in that component.

notbakaneko (changes requested):
use votes left instead of vote count....

gagahpangeran:
Ah, that's what I need. Thanks!

resources/assets/lib/gallery-contest-vote-progress.tsx
```diff
@@ -0,0 +1,63 @@
...
+  componentWillUnmount() {
+    $.unsubscribe('.count');
+  }
+
+  render() {
+    return (
+      <div className='pswp__button pswp__button--vote-progress'>
+        <CircularProgress
+          current={this.state.voteLeft}
+          max={this.getMaxVotes()}
```
nanaya, line 35: make it part of state (and computed together in syncState) otherwise might as well just get both votes left and this during render.

And I just noticed max votes is already fetched in getVoteLeft anyway.

resources/assets/lib/gallery-contest.ts
```diff
@@ -1,17 +1,22 @@
...
   private root: HTMLElement;
 
   constructor(container: HTMLElement, pswp: any) {
     this.root = container.querySelector('.js-pswp-buttons') as HTMLElement;
-    render(createElement(GalleryContestVoteButton, {pswp}), this.root);
+
+    const voteButton = createElement(GalleryContestVoteButton, { pswp });
+    const voteProgress = createElement(GalleryContestVoteProgress);
+
+    render(createElement(Fragment, null, [voteButton, voteProgress]), this.root);
```
nanaya, line 19: rename the file and use tsx syntax (and `<>` instead of `<Fragment>`)

resources/assets/lib/gallery-contest-vote-progress.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const voteCount = parseInt(voteSummary.dataset.contestVoteCount ?? '0', 10);
+
+    return maxVotes - voteCount;
+  };
+
+  private syncState = () => {
+    this.setState({ voteLeft: this.getVoteLeft() });
+  };
+
+  private voteSummary = () => document.querySelector('.js-contest-vote-summary') as HTMLElement;
```
nanaya, line 62: use proper instanceof check instead of casting (and just throw on error to simplify things)

resources/assets/lib/gallery-contest-vote-progress.tsx
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import { CircularProgress } from 'circular-progress';
+import * as React from 'react';
+
+type Props = Record<string, never>;
+
+interface State {
+  voteLeft: number;
```
nanaya, line 10: is it `voteLeft` or `votesLeft` :thinking:

resources/assets/lib/gallery-contest-vote-progress.tsx
```diff
@@ -0,0 +1,63 @@
...
+  };
+
+  private getVoteLeft = () => {
+    const maxVotes = this.getMaxVotes();
+
+    const voteSummary = this.voteSummary();
+    const voteCount = parseInt(voteSummary.dataset.contestVoteCount ?? '0', 10);
+
+    return maxVotes - voteCount;
+  };
```
nanaya, line 56: looks like it'll be simpler if the whole data is put in single dataset attribute passed around as json and use mobx for handling changes and computing votes left.

nanaya:
> I'm so confused now. Why does the tooltip show it going up (and the progress fill)? There's so much contradicting information. I think the tooltip should read something like "Remaining votes: 9". The circular fill can probably stay as it is, since it's mostly just flair at this point.

There's no contradicting information. Number in the circle is remaining votes available. The fill and the tooltip are showing how much are used (and total available).

It should be clearer if the tooltip says `:used / :max votes used` similar to beatmap discussion reviews.

peppy:
Can give that a try, but I think having two numbers going in opposite directions may still feel weird.

resources/assets/lib/circular-progress.tsx
```diff
@@ -31,6 +33,10 @@ export class CircularProgress extends React.PureComponent<Props, any> {
       return null;
     }
 
+    const displayNumber = this.props.reverse
+      ? this.props.current
+      : this.props.max - this.props.current;
```
nanaya, line 38: See https://github.com/ppy/osu-web/pull/7772#issuecomment-866544855 on fixing the tooltip instead.

nanaya (changes requested):

resources/assets/lib/circular-progress.tsx
gagahpangeran: Wait, without this changes, the number will count up from `0` to `max` if we use `votesLeft`, because of `max - votesLeft`.

nanaya: The number indicates remaining votes, the fill indicates how many votes are placed. It's the same behavior across other CircularProgress uses.

gagahpangeran: Hmm okay.

resources/assets/lib/gallery-contest-vote-progress.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const voteSummary = this.voteSummary();
+    const data = voteSummary.dataset.contestVoteSummary;
+
+    if (data === undefined) {
+      return { maxVotes: 0, voteCount: 0 };
+    }
+
+    const voteState: State = JSON.parse(data);
+    return voteState;
+  };
```
nanaya, line 55: The following is probably fine for now if we assume the element always has the expected dataset with the expected format.

The only difference being this also throws error if dataset is undefined instead of returning a blank "default" which is already kind of unexpected. And if we want to really account for such cases, the current one is missing check whether the dataset is in expected format...

```suggestion
  private getVoteState = (): State => JSON.parse(this.voteSummary().dataset.contestVoteSummary ?? '');
```

nanaya (changes requested):

resources/assets/lib/gallery-contest-vote-progress.tsx
gagahpangeran: Okay, I get it.

resources/assets/less/photoswipe-overrides.less
```diff
@@ -28,6 +28,16 @@
       color: @contest-selected-colour;
     }
 
+    &--vote-progress {
+      display: flex;
+      align-items: center;
+      justify-content: center;
+
+      && {
```
nanaya, line 36: sorry, I just noticed the element has 0.75 opacity on normal state and has pointer cursor on hover which is weird because it's not clickable and shouldn't have hover effect.

nanaya (changes requested):

resources/assets/less/photoswipe-overrides.less
gagahpangeran: No problem.

nanaya (approved):

PR #7773, opened 2021-06-22
Match user page with its preview

### LINKED ISSUE #7771: Profile page doesn't match its preview
Noticed today when I was rearranging my profile that there's a glitch that may happen when one of the images is dynamically updated (svg), where the dynamic image will shift itself up after saving all changes. This doesn't happen in the "Preview" mode, only when saving 
How its looking: (how it looks after saving)
image

How it should look like: (How it shows up in the preview mode)
image

nanaya:
Hopefully this doesn't break anything :eyes: 

An alternative is to match `withoutImageDimensions` option between preview and real.

Or just remove `withoutImageDimensions` option considering it's calculated during preview and everything goes through the proxy anyway...? &larr; now this (previously was matching `.bbcode img` and `.proportional-container` vertical-align css)

Resolves #7771.

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -630,7 +630,7 @@ class UsersController extends Controller
                 ]);
             }
 
-            return ['html' => $user->userPage->bodyHTML(['withoutImageDimensions' => true, 'modifiers' => ['profile-page']])];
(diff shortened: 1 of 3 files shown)


peppy:
Seems like removing may be best if it simplifies things. Since it's probably not going to save any edge cases (ie. when the user updates the image at the same URL while previewing... since at that point it's already cached at camo or cloudflare).

notbakaneko (approved):

PR #7777, opened 2021-06-23
Document breaking API change from removing group description

cl8n:
missed earlier

> [9:02 PM] tybug: was removing description from UserGroup intentional?
> [9:02 PM] tybug: if so it should probably be documented as a breaking change

and document where it got moved to, even though it's not used in API for now.


resources/views/docs/info.blade.php
```diff
@@ -304,6 +304,9 @@
 
 ## Breaking Changes
 
+### 2021-06-14
+- Removed `description` from [Group](#group) and [UserGroup](#usergroup). It has been moved to an optional attribute on [Group](#group).
```
nanaya, line 308: More like removed from UserGroup and changed to optional attribute with different structure on Group.

nanaya (changes requested):

nanaya (approved):

PR #7779, opened 2021-06-23
Add osu!catch "hard rock offsets" to DA mod

smoogipoo:
See: https://github.com/ppy/osu/pull/13629

Mod settings should really be validated per-ruleset. At the same time we'll need to figure out a path forward for acronym conflicts in the first place (which, I guess this could be treated as).

But the taiko setting is already in there so...


nanaya:
I guess we could add something like `SETTINGS_RULESET` and adjust the check accordingly.

peppy (approved):

PR #7798, opened 2021-06-25
Adjust timing for rendering react components

### LINKED ISSUE #7776: Render React before Turbolinks render new page
Currently, when navigating to page which rendered with react, there is a flash of empty page before the actual content is rendered. This is because React is rendered after the new page (which only has empty placeholder) is rendered (when `turbolinks:load` is fired).

Moving the render to when `turbolinks:before-render` is fired, where the new page is in `event.data.newBody`, should fix the issue.

Updates are needed to make sure anything that rely on data on new body is passed properly. And maybe some other checks.

nanaya:
This is mildly sketchy by using synchronous request because there doesn't seem a way to pause turbolinks render until the additional scripts on the page finishes downloading.

Not quite tested yet :eyes: 

Resolves #7776.

As it turned out I need to fix components event listeners to make sure they're all unique :eyes: 

Splitting out the event listener thing to a separate PR.

- [x] depends on #7804


notbakaneko (approved):
seems to work...so far 🤔

PR #7804, opened 2021-06-28
Unique namespace for React event

nanaya:
Split off from #7798. Also changed the generator to use sequential number instead of uuid.

And fixed one component to actually use the generated value instead of stringifying the function :eyes:


notbakaneko (approved):
seems to work

PR #7807, opened 2021-06-28
Remove comments count from current user json

nanaya:
Seems like one pointless extra query ಠ_ಠ


notbakaneko (changes requested):
comments_count should be excluded from the `CurrentUser` type (or `UserJsonExtended` needs some updating; the type itself is only used for user profile page components)

nanaya:
The typing could be better but I'm not tracking down those and just mark it optional in UserJson for now...

notbakaneko (approved):

PR #7809, opened 2021-06-29
Only store completed scores in main scores table

nanaya:
The route is getting weird :thinking: 

This allows score processor or whatever to process completed scores sequentially.

- [x] lazer submitting fails and retries (for playcount)


app/Http/Controllers/Solo/ScoresController.php
```diff
@@ -25,36 +26,13 @@ public function __construct()
...
-        return json_item($score, 'Solo\Score');
-    }
-
-    public function update($beatmapId, $id)
-    {
-        static::getBeatmapOrFail($beatmapId);
-
         $user = auth()->user();
-        $score = Score::where('user_id', $user->getKey())->findOrFail($id);
+        $scoreToken = ScoreToken::where([
```
peppy, line 32: Should probably have a check that `score_id` is `null`.

routes/web.php
```diff
@@ -380,7 +380,10 @@
                 Route::get('scores', 'BeatmapsController@scores')->name('scores');
 
                 Route::group(['as' => 'solo.', 'prefix' => 'solo'], function () {
-                    Route::resource('scores', 'Solo\ScoresController', ['only' => ['store', 'update']]);
+                    Route::group(['namespace' => 'Solo'], function () {
+                        Route::post('scores', 'ScoreTokensController@store')->name('score-tokens.store');
+                        Route::put('scores/{token}', 'ScoresController@store')->name('scores.store');
```
notbakaneko, line 385: shouldn't this be `store.update` 🤔

app/Http/Controllers/Solo/ScoresController.php
```diff
@@ -5,74 +5,59 @@
...
-        } catch (PDOException $e) {
-            // TODO: move this to be a validation inside Score model
-            throw new InvariantException('failed updating score');
+        $scoreToken = ScoreToken::where([
+            'beatmap_id' => $beatmapId,
+            'user_id' => $user->getKey(),
+        ])->findOrFail($tokenId);
+
+        // return existing score otherwise (assuming duplicated submission)
+        if ($scoreToken->score_id === null) {
```
notbakaneko, line 29: `$scoreToken` should be locked for update if doing this

notbakaneko (changes requested):

routes/web.php
nanaya: no because it's creating new score. The method stayed because that's what the client is using

notbakaneko (approved):

PR #7819, opened 2021-07-01
Add active user indicator in multiplayer rooms high table

nanaya:
Also only include users currently in room for recent participants of open room.


peppy (approved):

notbakaneko (approved):
If a room ends, players who aren't `in_room` will start appearing in the response? I suppose that isn't an issue since the user list isn't pageable, anyway.

peppy:
Yep. From the multiplayer server end we will also set the participant count to match, so the final results of all users can be viewed.

PR #7820, opened 2021-07-01
Perform full page load on logout

notbakaneko:
- Changes logout to do a full page reload instead; this saves us the problem of trying to clear and reinit js in the right order with the right timing - logging out should give clean state anyway.
- Moves the logout handlers so there's less places to look for logout handlers


resources/views/layout/_popup_user.blade.php
```diff
@@ -40,7 +40,7 @@ class="js-logout-link simple-menu__item"
         type="button"
         data-url="{{ route('logout') }}"
         data-confirm="{{ osu_trans('users.logout_confirm') }}"
-        data-method="delete"
+        data-method="delete
```
nanaya, line 43: 🤔

notbakaneko: 🤔

resources/assets/lib/user-login-observer.ts
```diff
@@ -15,7 +16,14 @@ export default class UserLoginObserver {
     dispatch(new UserLoginAction());
   };
 
-  userLogout = () => {
```
nanaya, line 18: `isNotificationEventLogoutJson` notification event handler in SocketWorker needs an update as well

nanaya (changes requested):

nanaya (approved):

PR #7822, opened 2021-07-01
Add topic logs

### LINKED ISSUE #7786: Forum posts should show who deleted it
Comments and beatmapdiscussion already show this.. forum posts probably should as well.

This also would be a step towards replacing quickmod logs

venix12:
partially resolves #7786.

based on how quickmod works/looks like in the legacy forum

did that thing in blade as the entire forum is made with it and it's also pretty much one static table 🤔


RockRoller01:
Would probably be better to display "Moved topic from X to Y" than just "Moved topic from X"

These entries are probably the most annoying ones in quickmod currently, because you need to remember where a post currently is while also keeping an overview of other move actions to fully understand a topics history.

Also not sure if IPs really need to be displayed, as they really do not matter and some maybe wouldnt want theirs to be openly displayed

resources/lang/en/forum.php
```diff
@@ -127,6 +127,49 @@
...
         ],
 
+        'logs' => [
+            '_' => 'Topic logs',
+            'title' => 'topic logs',
+
+            'columns' => [
+                'action' => 'Action',
+                'date' => 'Date',
+                'ip' => 'IP Adress',
```
zeusminus, line 137:
```suggestion
                'ip' => 'IP Address',
```
typo

venix12:
> Would probably be better to display "Moved topic from X to Y" than just "Moved topic from X"
> 
> These entries are probably the most annoying ones in quickmod currently, because you need to remember where a post currently is while also keeping an overview of other move actions to fully understand a topics history.

Well, this info isn't stored currently so that would require an infrastructure change, if deemed so, probably a thing for a different PR.

app/Models/Log.php
```diff
@@ -52,6 +52,51 @@ public function setLogDataAttribute($value)
...
+            $data = $this->log_data[1]; // username instead of topic title
+        }
+
+        return trans("forum.topic.logs.data.{$translationKey}", ['data' => $data]);
+    }
+
+    public function translationKey()
+    {
+        return strtolower(str_replace('LOG_', '', $this->log_operation));
+    }
```
cl8n, line 98: doesn't seem necessary to have function for it, unless you were planning to use this in later PR

app/Models/Log.php
```diff
@@ -52,6 +52,51 @@ public function setLogDataAttribute($value)
         $this->attributes['log_data'] = serialize($value);
     }
 
+    public function dataForDisplay()
```
cl8n, line 55:
```suggestion
    public function dataForDisplay(): string
```

cl8n (changes requested):
haven't tested yet, just minor things I noticed reading through.

app/Models/Log.php
cl8n: also just to be safe, only remove the `LOG_` at the beginning (currently removes all)

app/Http/Controllers/Forum/TopicLogsController.php
```diff
@@ -0,0 +1,32 @@
...
+    public function __construct()
+    {
+        return parent::__construct();
+    }
+
+    public function show($id)
+    {
+        if (priv_check('ForumTopicLogsView')->can() === false) {
+            abort(403);
+        }
```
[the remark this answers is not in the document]
venix12: well, it used to be a mod-only thing, while technically one can have mod perms for certain forums without being a mod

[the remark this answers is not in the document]
cl8n: ... "anyone" being the forum's moderators, which is what that permission checks for 🤔

app/Http/Controllers/Forum/TopicLogsController.php
```diff
@@ -0,0 +1,32 @@
...
+use App\Models\Log;
+
+class TopicLogsController extends Controller
+{
+    public function __construct()
+    {
+        return parent::__construct();
+    }
+
+    public function show($id)
```
nanaya, line 18: `index`, not `show`. `$topicId`, not `$id`

app/Http/Controllers/Forum/TopicLogsController.php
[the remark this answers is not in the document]
nanaya: yeah, unless there's specific reason the new per-forum moderator isn't allowed to, this should use `ForumModerate` (and `ensureCan`)

app/Models/Log.php
```diff
@@ -52,6 +52,51 @@ public function setLogDataAttribute($value)
...
+                    break;
+                case 2:
+                    $translationString = 'announcement';
+                    break;
+            }
+
+            return trans("forum.topic.logs.data.{$translationString}");
+        }
+
+        $data = $this->log_data[0];
```
nanaya, line 73: displaying topic title isn't useful in most cases because the page is already per-topic

app/Models/Log.php
```diff
@@ -52,6 +52,51 @@ public function setLogDataAttribute($value)
...
+            $state = $this->log_data['state'] ? 'add' : 'remove';
+            $data = $this->log_data['issueTag'];
+            $translationKey = "{$state}_tag";
+        }
+
+        if ($this->log_operation === 'LOG_POST_EDITED') {
+            $data = $this->log_data[1]; // username instead of topic title
+        }
+
+        return trans("forum.topic.logs.data.{$translationKey}", ['data' => $data]);
```
nanaya, line 92: `data` key is so vague. Please use actual name (username, topic_title, etc).

app/Models/Log.php
nanaya: yes, both of above

app/Models/Log.php
```diff
@@ -52,6 +52,51 @@ public function setLogDataAttribute($value)
...
+                    break;
+            }
+
+            return trans("forum.topic.logs.data.{$translationString}");
+        }
+
+        $data = $this->log_data[0];
+        $translationKey = $this->translationKey();
+
+        $noTranslation = ['LOG_DELETE_TOPIC', 'LOG_LOCK', 'LOG_RESTORE_TOPIC', 'LOG_UNLOCK'];
```
nanaya, line 76: more like `noDetail`. Also `static` (and considering its usage, probably should be a `Set` instead of plain array).

app/Models/Log.php
```diff
@@ -52,6 +52,51 @@ public function setLogDataAttribute($value)
         $this->attributes['log_data'] = serialize($value);
     }
 
+    public function dataForDisplay()
+    {
+        if ($this->log_operation === 'LOG_TOPIC_TYPE') {
+            switch ($this->log_data['type']) {
```
nanaya, line 58: can use `$translationString = match ($this->log_data['type']) { ... }`

...and why is it called `translationString` while the rest of function calls it `translationKey`

resources/views/forum/topics/logs/show.blade.php
```diff
@@ -0,0 +1,59 @@
...
+    @include('forum.topics._floating_header')
+    @include('forum._header', [
+        'isTopicLogs' => true,
+        'forum' => $topic->forum,
+        'topic' => $topic,
+    ])
+
+    <div class="osu-page osu-page--generic osu-page--full">
+        @if ($logs->count() > 0)
+            <table class="forum-topic-logs-table table">
```
nanaya, line 19: bootstrap table isn't all that useful here (and using bootstrap components should be generally avoided)

app/Http/Controllers/Forum/TopicLogsController.php
```diff
@@ -0,0 +1,32 @@
...
+    {
+        if (priv_check('ForumTopicLogsView')->can() === false) {
+            abort(403);
+        }
+
+        $topic = Topic::withTrashed()->findOrFail($id);
+
+        $logs = $topic->logs()
+            ->where('log_type', Log::LOG_FORUM_MOD)
+            ->paginate(30);
```
nanaya, line 28: a descending order seems more useful?

resources/lang/en/forum.php
```diff
@@ -127,6 +127,49 @@
...
+                'ip' => 'IP Address',
+                'user' => 'User',
+            ],
+
+            'data' => [
+                'add_tag' => 'added ":data" tag',
+                'announcement' => 'pinned topic and marked as announcement',
+                'delete_post' => 'from topic :data',
+                'edit_topic' => 'to :data',
+                'edit_poll' => 'for :data',
```
nanaya, line 146: poll title is in bbcode format so it should be converted accordingly for display

resources/lang/en/forum.php
```diff
@@ -127,6 +127,49 @@
...
+                'action' => 'Action',
+                'date' => 'Date',
+                'ip' => 'IP Address',
+                'user' => 'User',
+            ],
+
+            'data' => [
+                'add_tag' => 'added ":data" tag',
+                'announcement' => 'pinned topic and marked as announcement',
+                'delete_post' => 'from topic :data',
```
nanaya, line 144: an actual post id (and link to it) would be more useful. Same with `post_edited` and `restore_post`.

resources/views/forum/topics/show.blade.php
```diff
@@ -74,6 +78,14 @@ class="forum-topic-title__title forum-topic-title__title--edit js-forum-topic-ti
 
             @if (count($toolbarItems) > 0)
                 <div class="forum-topic-toolbar js-forum-topic-title--toggleable">
+                    @if (isset($toolbarItems['browse_logs']))
```
nanaya, line 81: I think this is better as part of bottom navigation bar otherwise it's rather difficult to access if you're not at the top of the topic already

routes/web.php
```diff
@@ -146,6 +146,8 @@
                 Route::post('topics/{topic}/edit-poll', 'TopicsController@editPollPost')->name('topics.edit-poll.store');
                 Route::get('topics/{topic}/edit-poll', 'TopicsController@editPollGet')->name('topics.edit-poll');
 
+                Route::get('topics/{topic}/logs', 'TopicLogsController@show')->name('topics.logs');
```
nanaya, line 149: shouldn't be in between a bunch of TopicsController routes. Also
```
Route::group(['as' => 'topics.', 'prefix' => 'topics/{topic}'], function () {
    Route::resource('logs', ...);

    ...other TopicsController actions...
});
...topics/preview...
...topics resource...
```

resources/views/forum/topics/logs/show.blade.php
```diff
@@ -0,0 +1,59 @@
+{{--
+    Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+    See the LICENCE file in the repository root for full licence text.
+--}}
+@extends('master', [
+    'titlePrepend' => trans('forum.topic.logs.title'),
```
nanaya, line 6: include topic title

resources/lang/en/forum.php
```diff
@@ -127,6 +127,49 @@
...
+                'pin' => 'pinned topic',
+                'post_edited' => 'posted by :data',
+                'remove_tag' => 'removed ":data" tag',
+                'restore_post' => 'from topic :data',
+                'unpin' => 'unpinned topic',
+            ],
+
+            'no_results' => 'no logs found...',
+
+            'operations' => [
```
nanaya, line 157: There are a few other operations not supported in osu-web but exist from the old system

Here's the full list of operations in the table

```
+------------------------+
| log_operation          |
+------------------------+
| LOG_DELETE_POST        |
| LOG_POST_EDITED        |
| LOG_MOVE               |
| LOG_LOCK               |
| LOG_MERGE              |
| LOG_USER_LOCK          |
| LOG_SPLIT_DESTINATION  |
| LOG_SPLIT_SOURCE       |
| LOG_UNLOCK             |
| LOG_DELETE_TOPIC       |
| LOG_TOPIC_TYPE_CHANGED |
| LOG_USER_UNLOCK        |
| LOG_FORK               |
| LOG_PIN                |
| LOG_UNPIN              |
| LOG_ISSUE_TAG          |
| LOG_RESTORE_POST       |
| LOG_TOPIC_TYPE         |
| LOG_EDIT_TOPIC         |
| LOG_EDIT_POLL          |
| LOG_RESTORE_TOPIC      |
+------------------------+
```

nanaya (changes requested):

app/Models/Log.php
```diff
@@ -52,6 +52,51 @@ public function setLogDataAttribute($value)
...
+                    break;
+                case 1:
+                    $translationString = 'pin';
+                    break;
+                case 2:
+                    $translationString = 'announcement';
+                    break;
+            }
+
+            return trans("forum.topic.logs.data.{$translationString}");
```
nanaya, line 70: forgot about this but `trans` shouldn't be used directly. Use `osu_trans` instead.

nanaya (changes requested):

resources/lang/en/forum.php
venix12: post id isn't stored though :thinking:

should i make it a thing?

nanaya: wtf the log function in the controller is a lie :raised_eyebrow: 

see if it's easy to add

venix12: yeah was confused about that as well 😔, probably an easy change though

resources/lang/en/forum.php
venix12: is there any documentation on those not used by osu-web? (see: LOG_MERGE, LOG_USER_LOCK, LOG_SPLIT_DESTINATION, LOG_SPLIT_SOURCE, LOG_TOPIC_TYPE_CHANGED, LOG_USER_UNLOCK, LOG_FORK), how is ``log_data`` stored on them etc.

app/Models/Log.php
venix12: re: no function, it's used in both ``dataForDisplay()`` function and view for the log title so using a function makes sense i guess

resources/lang/en/forum.php
cl8n: not "documentation" but english texts with log_data placeholders are at <https://github.com/phpbb/phpbb/blob/master/phpBB/language/en/acp/common.php>. if it's not clear what the log_data is from that, searching for the operation in this repo will probably find u the relevant `log::add` call

some of those might be specific to osu though.

venix12: ok, from delving into the phpbb code i figured out that pretty much all of them (excluding fork) store topic title only, didn't find anything regarding ``LOG_PIN`` and ``LOG_UNPIN`` though, guess it's safe to assume they are just pin and unpin without further information and just display them with no description 🤔

resources/lang/en/forum.php
venix12: since it can be multi-line somehow, think it's better to just not display details for this kinda log entry

resources/assets/less/bem/forum-topic-logs-table.less
```diff
@@ -0,0 +1,36 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+.forum-topic-logs-table {
+    width: 100%;
+
+    &__body {
+        font-size: @font-size--title-small;
+    }
+
+    &__date {
```
nanaya, line 11: the action column can probably have 100% width with other columns having some padding and `white-space: nowrap`

resources/views/forum/topics/logs/show.blade.php
```diff
@@ -0,0 +1,78 @@
...
+        'forum' => $topic->forum,
+        'topic' => $topic,
+    ])
+
+    <div class="osu-page osu-page--generic osu-page--full">
+        @if ($logs->count() > 0)
+            <table class="forum-topic-logs-table">
+                <thead class="forum-topic-logs-table__header">
+                    <tr>
+                        <th>{{ trans('forum.topic.logs.columns.date') }}</th>
```
nanaya, line 30: `osu_trans` (also everywhere else)

app/Models/Log.php
```diff
@@ -14,6 +14,8 @@
  * @property string $log_operation
  * @property int $log_time
  * @property int $log_type
+ * @property Forum\Post $post
```
nanaya, line 17: `|null`

nanaya (changes requested):

nanaya (approved):

PR #7834, opened 2021-07-06
Update Paypal SDK

notbakaneko:
Updates the Paypal SDK to an octane/swoole compatible version.
This version of the SDK uses the Paypal v2 API instead of the v1 API.

The most notable changes are the resource IDs on the Paypal side are different (again) and approval links from v1 won't work on v2.

This PR explicitly _does not_ include additional cleanups of stuffs and workarounds that seem to no longer be required


nanaya (changes requested):

nanaya (approved):
Seems working

PR #7837, opened 2021-07-07
Multiplayer realtime room password

nanaya:
Nothing fancy, just storing the passed parameter to database so the multiplayer server can then take care of it. It's not supposed to be secure either (and may be reset once the room ends).


notbakaneko (changes requested):
password to be stored is post-hash (already hashed)?

nanaya:
nope.jpg. Kinda needed for generating invitation link etc.

notbakaneko:
oh wait, I was reading it wrong 🙃

notbakaneko (approved):

Issue #7828, opened 2021-07-03
Implement new design of beatmapset info page


gagahpangeran:
Design : https://www.figma.com/file/HDSXf4LPPV7hnD5wrWYG4jix/Web%2FBeatmap-Info?node-id=0%3A1

---

I will implement it and post the update and question here if there's something not clear with the design.

gagahpangeran:
Found first problem, how to handle very large diff like this?

![Screenshot from 2021-07-04 21-35-18](https://user-images.githubusercontent.com/4964985/124389147-17983b80-dd10-11eb-89ba-578c1cfe2460.png)

peppy:
Can you explain what the issue is there?

gagahpangeran:
In figma, the difficulty icons are like this

![Screenshot from 2021-07-04 22-40-46](https://user-images.githubusercontent.com/4964985/124391059-ee2fdd80-dd18-11eb-9226-135037106323.png)

They are at the bottom of the section before the difficulty name in one line only.

But when the difficulties are very large enough, the icons become two lines.

![Screenshot from 2021-07-04 21-35-18](https://user-images.githubusercontent.com/4964985/124389147-17983b80-dd10-11eb-89ba-578c1cfe2460.png)

I want to confirm is this expected to become two lines or there should be something to do to handle such a large diffs.

peppy:
It looks fine, although I have no idea what those four weird dots are to the left. You may want to clarify with @arflyte what they are before proceeding.

frenzibyte:
It appears to be a button showing a pop-up next to it containing all the difficulties in an extended form:

![CleanShot 2021-07-04 at 22 09 45](https://user-images.githubusercontent.com/22781491/124396813-8d9ea180-dd14-11eb-82f9-a307ace7ca4e.png)

Phippe:
Is there a reason that the density and points of failure section stretches the whole width and is generally quite large? To me, this makes the impression that it is one of the most important things, while I don't think it's any more important than the graphs on the right.

So if I could suggest one thing, it would be to put those two graphs to right as well. You could leave the current height of the container unchanged and have some additional space for the other stats and metadata (maybe to display more tags or simply to make it less dense there).

peppy:
It's fine where it is. It's intentionally giving breathing space to the pieces above and below it.

nanaya:
[...]

PR #7849, opened 2021-07-09
Convert beatmap list component to typescript and make it more reusable

gagahpangeran:
Based on [new beatmapset page design](https://www.figma.com/file/HDSXf4LPPV7hnD5wrWYG4jix/Web%2FBeatmap-Info?node-id=0%3A1), there will be beatmap list dropdown in both beatmapset show and discussion. Currently #7828 is still WIP in my local branch and this PR is one of the pre-req for that.

This PR is convert the existing beatmap list component to typescript and make it more reusable without changing its current behaviour and design.

List of component that converted to typescript:
- beatmap list
- beatmap list item
- blackout

In beatmap list itself, I change some props to make it reusable.
- remove `beatmapset` props (this is not used anywhere in beatmap list component).
- remove `currentDiscussions` props (this is only used for count number).
- add `createLink` props to replace `BeatmapDiscussionHelper.url`, so it's more generic.
- add `getCount` as optional props to get count number of beatmap list item.
- add `onSelectBeatmap` props to handle when dropdown item is selected.


resources/assets/lib/beatmap-discussions/beatmap-list.tsx
```diff
@@ -0,0 +1,119 @@
...
+  constructor(props: Props) {
+    super(props);
+
+    this.state = {
+      showingSelector: false,
+    };
+  }
+
+  componentDidMount() {
+    $(document).on('click.beatmapList', this.onDocumentClick);
```
nanaya, line 32: conflicts (the event listeners are missing unique id)

nanaya (changes requested):

resources/assets/lib/beatmap-discussions/beatmap-list.tsx
gagahpangeran: Done

nanaya (changes requested):

nanaya (approved):

PR #7793, opened 2021-06-24
Show when a user's nomination gets reset

notbakaneko:
Adds a new `nomination_reset_received` event for when a beatmap that a user nominated gets its nomination reset.
Additionally records the ids of the users that had their nominations reset in the `disqualify` event.
As part of this, the disqualification and nomination logic is refactored out of the controllers (the part that handles the posts is still there), as well as the review create/update logic getting combined.

closes #7329

- [x] record user nominations separately (might as well, instead of some events having the related data and some not)
- [x] post migration updates (will go in a separate PR but need to check if anything needs to be back-ported to this one)

PR #7861, opened 2021-07-13
Use BeatmapsetNominations instead of events since reset

notbakaneko:
Uses the new `beatmapset_nominations` table added by #7793 to work out nominations

- [x] continues off #7793
- [x] #7983
- [x] #7985


app/Models/Beatmapset.php
```diff
@@ -1033,13 +1030,13 @@ public function nominationsSummaryMeta()
     public function isLegacyNominationMode()
     {
         return $this->memoize(__FUNCTION__, function () {
-            return $this->nominationsSinceReset()->whereNull('comment')->exists();
+            return $this->beatmapsetNominations()->current()->whereNull('modes')->exists();
```
nanaya, line 1033: apparently the playmode data weren't migrated :upside_down_face:

nanaya (changes requested):

nanaya (approved):

PR #7875, opened 2021-07-15
Use country flags from twemoji project

### LINKED ISSUE #3760: Several flags have incorrect details
Migrated from ppy/osu-resources#43. Since April, a few of the issues have been fixed. Here are the important ones which remain:

* Cambodia (`KH`): The central band should be red and the outer bands should be blue, not vice versa.
* Comoros (`KM`): As a symbol of Islam, the crescent (and stars) should probably not be distorted as they are.
* Sri Lanka (`LK`): The golden lion on the left side of the flag is completely missing.
* Mauritania (`MR`): The red stripes on the top and bottom are completely missing.
* Maldives (`MV`): The center is completely incorrect. It should not be blue, it should be green.

nanaya:
Should resolve #3760 once and for all (and any further complaints shall go to twemoji project instead :eyes: )


peppy:
In brief experimentation, if we're looking to bring the colours back in line with how they used to be (because the new ones, while correct, look completely out of place IMO) applying `filter:saturate(90%) hue-rotate(-25deg)` or so feels *slightly* better?

Before:
20210720 183337 (Safari)

20210720 183323 (Safari)

After:

20210720 183823 (Safari)

20210720 183832 (Safari)

Probably need more opinions.

notbakaneko:
While I agree the new ones have a bit too much contrast against the rest of the design, I don't think the hues should be colour shifted.

peppy:
Any other suggestions? As it stands, I'm pretty against this change as a whole. We use these in so many places that this becomes an eyesore.

nanaya:
Maybe a combination of white overlay and saturation adjustment? This is with 20% white overlay and 110% saturation

peppy:
That seems closer to expectations. Potentially taking the whiteness a touch further even?

nanaya:
@arflyte can you adjust the flag colour... thing (see comments above)? Live demo at https://osunext.dev.myconan.net/rankings/osu/performance

peppy:
Looks pretty good at that URL (25% white overlay). I can live with that.

peppy (approved):

peppy:
Dropping an approval on this one, but let's wait for a response from @arflyte before merging.

arflyte:
Oh, I like ones with white overlay. Hopefully, no one complains that their flag colour is slightly inaccurate >_<

PR #7883, opened 2021-07-16
Fix shopify debug typing

nanaya:
More like working around it because a proper fix of maybe importing `@types/shopify-buy` seems quite involved.

### DIFF
--- a/resources/assets/lib/shopify-debug.ts
+++ b/resources/assets/lib/shopify-debug.ts
@@ -36,7 +36,7 @@ export async function fetchAllVariants(): Promise<any[]> {
   return variants;
 }
 
-export async function fetchAllVariantIds(): Promise {
+export async function fetchAllVariantIds(): Promise<unknown[]> {
   const variants = await fetchAllVariants();
 
   return variants.map((x: any) => ({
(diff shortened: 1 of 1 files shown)


notbakaneko:
`@types/shopify-buy` doesn't even have the correct types 🙃

nanaya:
maybe should look into fixing it upstream :thinking: the `checkout.create` is very wrong at the moment.

resources/assets/lib/shopify-debug.ts
```diff
@@ -36,7 +36,7 @@ export async function fetchAllVariants(): Promise<any[]> {
   return variants;
 }
 
-export async function fetchAllVariantIds(): Promise<{ }> {
+export async function fetchAllVariantIds(): Promise<unknown[]> {
```
notbakaneko, line 39: The explicit return type isn't needed (it's `{ gid: string, name: string }`,  anyway)

notbakaneko (changes requested):

resources/assets/lib/shopify-debug.ts
nanaya: it is required...

```
TS2314: Generic type 'Promise<T>' requires 1 type argument(s).
```

nanaya: oh I forgot about just removing the typing entirely :raised_eyebrow:

notbakaneko (approved):

PR #7888, opened 2021-07-19
Add language hint on wiki pages

### LINKED ISSUE #7787: Google links to random language version of wiki articles
When I search for a wiki page via google to share with someone, I end up sharing some stupid localisation version of it.

Not sure was the best fix for this is but it needs to be addressed somehow.

20210624 103623 (Safari app)

nanaya:
Should resolve #7787.


notbakaneko (changes requested):
I think we could do with an `availableLocales()` function; currently, if you go to a page which doesn't have a translation, it includes a link for the requested language (which is missing the actual article), but the link for the fallback language isn't included.

nanaya:
config's `available_locales` sure look like can use a conversion to `Set` :thinking:

notbakaneko (approved):

PR #7898, opened 2021-07-20
Use button circle for page toggle button

gagahpangeran:
The styling of `profile-page-toggle` can be reused on another page, so I renamed it to `page-toggle` for more generic.


resources/assets/less/bem-index.less
```diff
@@ -263,6 +263,7 @@
 @import "bem/page-nav-fancy";
 @import "bem/page-tabs";
 @import "bem/page-title";
+@import "bem/page-toggle";
```
nanaya, line 266: changing this to be `btn-circle` modifier seems like a better plan

nanaya (changes requested):

nanaya (approved):

PR #7902, opened 2021-07-21
Add ability to choose certain difficulty for loved

### LINKED ISSUE #3839: Allow beatmaps to be excluded when moving a beatmapset to the Loved category
It's a common case for Project Loved to move most of the difficulties of a mapset to Loved, but not all (some maps are invalid, incredibly poor quality, or etc.). Toy has to contact an admin every time this occurs since there's no way to do it on the site

I think the best way to implement this would be with a modal that shows when clicking the "Love" button, listing checkboxes for every beatmap in the set and asking the user which ones they would like to exclude, if any

gagahpangeran:
Closes #3839

QoL changes for project loved people.

**Example for choosing one difficulty only**

Peek 2021-07-21 09-44

**Example for choosing one game mode only**

Peek 2021-07-21 09-41


peppy:
Seems fine to add this in, but given the very small scope of this I wouldn't spend too long on the review process. It's pretty isolated so as long as it works it should be fine to get in with a minimal review.

nanaya (changes requested):

app/Models/Beatmapset.php
```diff
@@ -547,9 +547,13 @@ public function setApproved($state, $user)
 
         $this->save();
 
-        $this
-            ->beatmaps()
-            ->update(['approved' => $this->approved]);
+        $beatmaps = $this->beatmaps();
+
+        if (isset($beatmapIds)) {
+            $beatmaps = $beatmaps->whereIn('beatmap_id', $beatmapIds);
```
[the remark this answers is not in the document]
gagahpangeran: Does `$this->beatmaps()` also return deleted beatmap?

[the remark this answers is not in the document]
nanaya: no it doesn't

resources/assets/lib/beatmap-discussions/love-confirmation.tsx
```diff
@@ -0,0 +1,181 @@
...
+              <label className='osu-switch-v2'>
+                <input
+                  checked={this.state.selectedBeatmapIds.has(beatmap.id)}
+                  className='osu-switch-v2__input'
+                  onChange={this.handleCheckboxDifficulty}
+                  type='checkbox'
+                  value={beatmap.id}
+                />
+                <span className='osu-switch-v2__content' />
+                <div className='love-confirmation__switch-text'>
```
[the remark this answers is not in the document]
gagahpangeran: Main reason is I want to avoid awkward break line like this.

Screenshot from 2021-07-30 00-10-00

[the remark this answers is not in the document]
nanaya:
```
<label> {
  display: flex;
  align-items: baseline;
}
```

resources/assets/lib/beatmap-discussions/love-confirmation.tsx
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    return (
+      <div key={mode} className='love-confirmation__diff-mode'>
+        <div className='love-confirmation__diff-mode-title'>
+          <label className='osu-switch-v2'>
+            <input
+              checked={this.checkIsModeSelected(mode) !== false}
+              className='osu-switch-v2__input'
+              data-indeterminate={this.checkIsModeSelected(mode) === 'indeterminate'}
```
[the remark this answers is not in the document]
gagahpangeran: Sorry, I don't get it(?)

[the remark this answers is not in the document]
nanaya: `this.checkIsModeSelected(mode)` is being called twice

huuishuu:
While this is still being worked on, using this how would we handle adding other modes to Loved after the previous one had been moved to Loved? Say in a taiko+mania set we move the osu!mania difficulties to Loved and then 2 months later we want to move the osu!taiko ones as those were nominated and passed voting. Should this be addressed here or would this be a separate issue entirely?

gagahpangeran:
With this current feature, you have to unloved (and causing scoreboard wipe) and add it again to love.

I don't follow project loved project from the start, but did such case ever happen?

cl8n:
it hasn't, and I doubt there ever will be many cases of that-- there are only so many loved mapsets with other not-loved gamemodes viable for loved to begin with

the usual "ask Ephemeral to do it" works fine for something so infrequent

app/Models/Beatmapset.php
```diff
@@ -547,9 +547,13 @@ public function setApproved($state, $user)
 
         $this->save();
 
-        $this
-            ->beatmaps()
-            ->update(['approved' => $this->approved]);
+        $beatmaps = $this->beatmaps();
+
+        if (isset($beatmapIds) && empty(array_diff($beatmapIds, $beatmaps->get()->modelKeys()))) {
```
nanaya, line 552: `$beatmaps->whereKey($beatmapIds)->count() === count($beatmapIds)`. And probably should be before saving the beatmapset state and throw error if doesn't match? Seems weird completely ignoring away the parameter if it contains invalid value.

resources/assets/lib/beatmap-discussions/love-beatmap-modal.tsx
```diff
@@ -0,0 +1,175 @@
...
+import * as React from 'react';
+import { group as groupBeatmaps } from 'utils/beatmap-helper';
+
+interface Props {
+  beatmapset: BeatmapsetJson;
+  onClose: () => void;
+}
+
+@observer
+export default class LoveConfirmation extends React.PureComponent<Props> {
```
nanaya, line 19: `@observer` can't be `PureComponent`

resources/assets/lib/beatmap-discussions/love-beatmap-modal.tsx
```diff
@@ -0,0 +1,175 @@
...
+    const params = {
+      data: { beatmap_ids: [...this.selectedBeatmapIds] },
+      method: 'PUT',
+    };
+
+    $.ajax(url, params).done((response) => {
+      $.publish('beatmapsetDiscussions:update', { beatmapset: response });
+      this.props.onClose();
+    }).fail(osu.ajaxError)
+      .always(LoadingOverlay.hide());
```
nanaya, line 122: don't call the function when setting it for callback

nanaya (changes requested):

app/Models/Beatmapset.php
gagahpangeran: Sure, what error should be thrown?

nanaya: `InvariantException` should be fine.

app/Models/Beatmapset.php
```diff
@@ -545,11 +545,19 @@ public function setApproved($state, $user)
             $this->approvedby_id = null;
         }
 
+        $beatmaps = $this->beatmaps();
```
nanaya, line 548: I did only say "before saving" but there's no point setting all the attributes just to throw them away in a semi-unrelated parameter validation :thinking:

app/Models/Beatmapset.php
```diff
@@ -545,11 +545,19 @@ public function setApproved($state, $user)
...
+
+        if (isset($beatmapIds)) {
+            if ($beatmaps->whereKey($beatmapIds)->count() === count($beatmapIds)) {
+                $beatmaps = $beatmaps->whereIn('beatmap_id', $beatmapIds);
+            } else {
+                throw new InvariantException('beatmap id contains invalid id');
```
nanaya, line 554: beatmap_ids

resources/assets/lib/beatmap-discussions/love-beatmap-modal.tsx
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  render() {
+    return (
+      <div className='love-beatmap-modal'>
+        <div className='love-beatmap-modal__row love-beatmap-modal__row--title'>
+          {osu.trans('beatmaps.nominations.love_choose')}
+        </div>
+
+        <div className='love-beatmap-modal__row love-beatmap-modal__row--content u-fancy-scrollbar'>
```
nanaya, line 45: not actually visible in mobile but the scrollbar class need to be on parent for it to be applied correctly

nanaya (changes requested):

app/Models/Beatmapset.php
gagahpangeran: Yeah, I'm not sure what was I thinking when code this in the night :thinking:

gagahpangeran: Should `$beatmaps->update(...)` before or after `$this->save()`?

resources/assets/lib/beatmap-discussions/love-beatmap-modal.tsx
gagahpangeran: I forgot to change this when changed the overflow component in the style.

app/Models/Beatmapset.php
nanaya: either should be fine

nanaya (approved):

PR #7904, opened 2021-07-21
Update Sentry to support Octane

nanaya:
Current version doesn't have correct support for octane and generates slightly useless report. The [updated version is still in review](https://github.com/getsentry/sentry-laravel/pull/495) but seems to work, so, yay? :shrug:


notbakaneko (approved):
looks ok 🤔

PR #7909, opened 2021-07-22
Fix beatmapset card border overlap with user profile section switcher

### LINKED ISSUE #7908: Beatmapset card border can overlap with user profile section switcher
Peek 2021-07-22 09-32

gagahpangeran:
Closes #7908


resources/assets/less/bem/page-extra-tabs.less
```diff
@@ -4,7 +4,7 @@
 .page-extra-tabs {
   position: sticky;
   top: @nav2-height--pinned;
-  z-index: 1;
+  z-index: 999;
```
nanaya, line 7: z-index should be listed in `variables.less` instead so it's easier to check the order site-wide

nanaya (changes requested):

resources/assets/less/bem/page-extra-tabs.less
gagahpangeran: Moved and change it to 99. Seems working 🤔

resources/assets/less/variables.less
```diff
@@ -113,6 +113,7 @@
 // modal-backdrop: 1040
 // modal: 1050
 @z-index--overlay: 9990;
+@z-index--page-extra-tabs: 99;
```
nanaya, line 116: this section is sorted by the value

nanaya (changes requested):

nanaya (approved):

PR #7913, opened 2021-07-23
Add more tests for osu-common

gagahpangeran:
Not testing all functions yet, but at least add more coverage.


tests/karma/osu_common.spec.ts
```diff
@@ -16,53 +16,227 @@
...
+      expect(osu.transArray(['Me', 'My Self'])).toBe('Me and My Self');
+    })
+
+    it('should return correct translation for three items array', () => {
+      expect(osu.transArray(['Me', 'My Self', 'I'])).toBe('Me, My Self, and I');
+    })
+  })
+
+  describe('test transExists', () => {
+    it('should return true for existed translation', () => {
```
notbakaneko, line 187: `returns true when the translation exists`

tests/karma/osu_common.spec.ts
```diff
@@ -16,53 +16,227 @@
...
+      expect(osu.transExists('common.confirmation')).toBe(true);
+    })
+
+    it('should return false for non existed translation', () => {
```
notbakaneko, line 191: `returns false when the translation does not exist`

tests/karma/osu_common.spec.ts
```diff
@@ -16,53 +16,227 @@
...
+      )
+    })
+  })
+
+  describe('test formatBytes', () => {
+    it('should return same value in bytes', () => {
+      expect(osu.formatBytes(100)).toBe('100 B');
+    })
+
+    it('should return correct value in KB', () => {
```
notbakaneko, line 95: `correct` is a bit redundant seeing as it should be expected already...

tests/karma/osu_common.spec.ts
```diff
@@ -16,53 +16,227 @@
...
+  describe('test formatNumber', () => {
+    it('should return null', () => {
+      expect(osu.formatNumber(null)).toBe(null);
+    })
+
+    it('should return number with correct precision', () => {
+      expect(osu.formatNumber(12.345, 2)).toBe('12.35');
+    })
+
+    it('shoudl return integer number', () => {
```
notbakaneko, line 117: `should return the number as an integer` or better describing the test, `formats the number without precision`

notbakaneko (changes requested):
A lot of the test naming is redundant; `describe` does not need `test ` in front of everything.
Considering the its methods being tested directly, they are probably better described with `describe('.methodName')`.

`it('...')` preferably should be describing what is happening instead of just repeating the expected return value, where possible.

tests/karma/osu_common.spec.ts
```diff
@@ -16,53 +16,227 @@
...
+
 describe('osu_common', () => {
-  it('test locale file loaded in test runner', () => {
-    expect(Lang.has('common.confirmation')).toBe(true);
-  });
+  describe('test locale file loaded in test runner', () => {
+    it('should be loaded', () => {
+      expect(Lang.has('common.confirmation')).toBe(true);
+    })
+  })
```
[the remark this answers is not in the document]
gagahpangeran: The linter settings (both tslint and eslint) does not include test folder, so my formatter does not auto format this files(?) ¯\\_ (ツ)_/¯

And lint script in CI also does not include test folder, so I'm not aware of this.

notbakaneko (approved):
some descriptions can be fixed later

PR #7914, opened 2021-07-23
Convert timeout to typescript

gagahpangeran:
A small file to convert.


nanaya:
This class shouldn't even be used in typescript. it only exists because the original setTimeout syntax is awkward in coffeescript

gagahpangeran:
Ah, I see. Will close it then.

PR #7915, opened 2021-07-23
Convert osu-common to typescript

gagahpangeran:
Most of file changes are import header.

Depends on:
- [x] #7913
- [x] #7917 

I hope I didn't miss something.

And there's some issue when I was testing this, see comment review.


resources/assets/lib/models/user.ts
```diff
@@ -92,7 +93,8 @@ export default class User {
 
 const deletedUser = new User(-1);
 deletedUser.isDeleted = true;
-deletedUser.username = osu.trans('users.deleted');
+// deletedUser.username = osu.trans('users.deleted');
```
gagahpangeran, line 96: Somehow this line causing webpack error, I don't know why.

Screenshot from 2021-07-23 15-57-03

resources/assets/lib/osu-common.ts
```diff
@@ -0,0 +1,372 @@
...
+      if (Object.prototype.hasOwnProperty.call(params, key)) {
+        const value = params[key];
+
+        if (value != null) {
+          urlObj.searchParams.set(key, value);
+        } else {
+          urlObj.searchParams.delete(key);
+        }
+      }
+    }
```
gagahpangeran, line 316: Original coffeescript is 

```coffeescript
    for own key, value of params
      if value?
        urlObj.searchParams.set(key, value)
      else
        urlObj.searchParams.delete(key)
```

I'm not sure enough about conversion `for own` to plain javascript.

resources/assets/lib/osu-common.ts
```diff
@@ -0,0 +1,372 @@
...
+    if (validationMessage != null) {
+      let allErrors: string[] = [];
+
+      // FIXME: not sure about this conversion `for own` from coffeescript
+      for (const field in validationMessage) {
+        if (Object.prototype.hasOwnProperty.call(validationMessage, field)) {
+          const errors = validationMessage[field];
+          allErrors = allErrors.concat(errors);
+        }
+      }
```
gagahpangeran, line 343: Same here about `for own`.

Original coffeescript:

```coffeescript
      for own _field, errors of validationMessage
        allErrors = allErrors.concat(errors)
```

nanaya:
classWithModifiers in osu_common should be removed and use the one in utils/css

gagahpangeran:
> classWithModifiers in osu_common should be removed and use the one in utils/css

Will remove in separated PR then. This PR is already too big.

resources/assets/lib/osu-common.ts
```diff
@@ -0,0 +1,359 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+import GroupJson from 'interfaces/group-json';
+import { each, find, unescape } from 'lodash';
+import core from 'osu-core-singleton';
+import * as React from 'react';
+import { TurbolinksAction } from 'turbolinks';
+import { currentUrl as getCurrentUrl } from 'utils/turbolinks';
+
+const osu = {
```
notbakaneko, line 11: export the functions instead and `import * as osu`

notbakaneko (changes requested):

resources/assets/lib/models/user.ts
notbakaneko: `osu` hasn't been initialized yet when this module is loaded because there's now an implicit circular dependency between `osu-core` and `osu-common` modules.

resources/assets/lib/osu-common.ts
notbakaneko: `validation_error` should just be a plain object so you can use `for (const [a, b] of Object.entries(obj))` or `Object.values` if only the values are needed.

resources/assets/lib/models/user.ts
gagahpangeran: You're right.

`ajaxError` in `osu-common` is calling `osu-core`. Looks like I should move that function to `utils/ajax`.

nanaya:
I'm not reviewing this but just want to comment that there's no reason this must be one big PR.

Most of the functions aren't related to each other so technically one can make PR to split off just the json-related functions (`jsonClone`, `parseJson`, `storeJson`) to its own utils file at `utils/json.ts` for example.

(and still on json, `jsonClone` should've been typed `jsonClone<T>(obj: T): T`)

gagahpangeran:
Yeah, I just have some spare time when code this and just go straight forward.

Perhaps the goal is to separate all functions into its own utils file and finally nuked osu-common(?)

Should I just close this PR for now?

notbakaneko:
Just leave it for now

notbakaneko:
marking as draft until it can be split into smaller PRs

notbakaneko:
closing this as we've moved/are moving the functions into separate modules

PR #7923, opened 2021-07-26
Fix session duration for non-guest user on octane

nanaya:
Session handler instance is persistent so the duration needs to be set before saving.


peppy:
Kind of glad it was something this shallow.

notbakaneko (approved):

PR #7925, opened 2021-07-26
More uniform title for page links

### LINKED ISSUE #7910: Update all page titles to match menu labels
As discussed in https://github.com/ppy/osu/pull/13918

`beatmap` section:
`listing` -> `beatmap listing`
`packs` -> `beatmap packs`

`osu!store` -> `store`

`ranking` -> `rankings`

`knowledge base` -> `wiki`

nanaya:
Since there's no mapping between route and link title, things are rather verbose ( ﾟ ヮﾟ)

Resolves #7910.


peppy:
Hmm, those footer changes look very out of place... (most prominently, the capitalisation differences between the first two columns and the last column)

nanaya:
Updated to keep existing footer titles and only update menu to use page title translations.

notbakaneko (approved):

PR #7931, opened 2021-07-27
Stop notification reconnect after logged out and add exponential delay

### LINKED ISSUE #7912: Requests to /notifications don't bail on 4xx error, causing a denial of service
Due to some weirdness with session expiry yesterday (cause still undetermined, but likely related to octane deploy), a lot of users got logged out. One day later, the req/s experience by the server is still up around 33% due to 401 requests to `/notifications`.

In discussion, it seems that the retry logic doesn't bail in cases it should be (like 401 unauthorised). So user with osu! tabs in the background that haven't realised they got logged out are collectively DDoSing the server.

nanaya:
A bunch of states are probably broken in this case and may as well just do full page reload.

...but that's for another day :eyes: 

Resolves #7912


resources/assets/lib/utils/retry-delay.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import { random } from 'lodash';
+
+const initialDelay = 500;
+const maxDelay = 1000; // 30 minutes
```
notbakaneko, line 7: this is just 1 second? 🤔

resources/assets/lib/notifications/worker.ts
```diff
@@ -34,17 +35,18 @@ const isNotificationEventReadJson = (arg: SocketEventData): arg is NotificationE
...
   get hasData() {
     return this.firstLoadedAt != null;
   }
+  @observable waitingVerification = false;
+  @observable private firstLoadedAt?: Date;
+
+  private retryDelay = new RetryDelay();
+  private timeout: Partial<Record<string, number>> = {};
+  private xhr: Partial<Record<string, JQueryXHR>> = {};
+  private xhrLoadingState: Partial<Record<string, boolean>> = {};
```
notbakaneko, line 49: fields before `getters`? and there's a blank line above

resources/assets/lib/socket-worker.ts
```diff
@@ -20,21 +21,22 @@ interface NotificationFeedMetaJson {
 type ConnectionStatus = 'disconnected' | 'disconnecting' | 'connecting' | 'connected';
 
 export default class SocketWorker {
+
+  @computed
+  get isConnected() {
+    return this.connectionStatus === 'connected';
+  }
```
notbakaneko, line 28: same as above

notbakaneko (changes requested):

notbakaneko (approved):

PR #7945, opened 2021-07-29
Add muted mod

bdach:
Reference: https://github.com/ppy/osu/pull/14041

Valid for all rulesets, no incompatibility rules (yet?). Metronome setting being persisted server-side makes _some_ sense as it makes things a bit easier.

<personal_opinion>Playtesting that this works with multi was quite dreadful gameplay-wise. Never touching this mod with a ten foot pole again...</personal_opinion>


bdach:
Have tentatively pushed in new settings for the mod as introduced in https://github.com/ppy/osu/pull/14064.

peppy (approved):

PR #7949, opened 2021-07-30
Rename translation key to avoid case-insensitive collision

### LINKED ISSUE #7738: Common localisation has duplicate scoreboard_time keys
osu! is about to consume osu!web localisations. ResX, our localisation storage format, is case-insensitive for keys.

```
Processing /home/smgi/Repos/osu-web/resources/lang/en/common.php...
  -> WARNING: Skipping duplicate key "scoreboard_time.m" (scoreboard_time.M, scoreboard_time.m) in /home/smgi/Repos/osu-web/resources/lang/en/common.php.
  -> WARNING: Skipping duplicate key "scoreboard_time.mm" (scoreboard_time.MM, scoreboard_time.mm) in /home/smgi/Repos/osu-web/resources/lang/en/common.php.
  -> /home/smgi/Repos/osu/osu.Game/Localisation/Web/CommonStrings.cs
  -> /home/smgi/Repos/osu/osu.Game/Localisation/Web/Common.resx
```
Is this something that could be fixed easily? Only for the purposes of removing warnings output from the tool.

This is not a priority - it just means that we can't consume this localisation. But we're unlikely to use it anyway since the humanisation library we're using should already have a suitable localisation for this.

nanaya:
I'll update the translation for other languages after this is merged.

@smoogipoo, is there any other key need renaming?

Resolves #7738.


smoogipoo (approved):

smoogipoo:
These are the only conflicts :+1:

PR #7951, opened 2021-07-30
Allow chat.write scope without Chat bot group

notbakaneko:
~~Allows the use of `chat.write` with `client_credentials` - the precondition is delegation has to be requested (`bot`) and it can't be mixed with other scopes to give them user permissions.~~

After further discussion, `chat.write` with `client_credentials` will still be limited to [chat bots](https://osu.ppy.sh/wiki/en/Bot_account), but `chat.write` will be allowed for authorization code grants where the owner of the  OAuth client is yourself. 

The argument presented was authorization code grants still have an explicit intervention step where the user can reject the request for the scope, while client credentials do not.

- [x] update api docs


resources/views/docs/info.blade.php
```diff
@@ -259,7 +259,18 @@
...
+[Client Credentials Grant](#client-credentials-grant) tokens do not have a Resource Owner (i.e. is a guest user), unless they have been granted the {{ ApidocRouteHelper::scopeBadge('bot') }} scope. The Resource Owner of tokens with the {{ ApidocRouteHelper::scopeBadge('bot') }} scope is the owner of the OAuth Application that was granted the token.
+
+
+## Client Credentials Delegation
+
+While Client Credentials Grant tokens normally do not have an associated user by additionally requesting the {{ ApidocRouteHelper::scopeBadge('bot') }} scope, they may be allowed to act on behalf of the owner of the OAuth client (delegation). When using delegation, scopes that support delegation cannot be used together with scopes that do not support delegation.
```
peppy, line 267:
```suggestion
While Client Credentials Grant tokens normally do not have an associated user, by additionally requesting the {{ ApidocRouteHelper::scopeBadge('bot') }} scope, they may be allowed to act on behalf of the owner of the OAuth client (delegation). When using delegation, scopes that support delegation cannot be used together with scopes that do not support delegation.
```

I *think* this comma is intended? The meaning is unclear without it.

peppy (changes requested):

resources/views/docs/info.blade.php
notbakaneko: Must have been deleted while rearranging sentences 🤔

resources/views/docs/info.blade.php
```diff
@@ -259,7 +259,18 @@
...
+
+## Client Credentials Delegation
+
+While Client Credentials Grant tokens normally do not have an associated user, by additionally requesting the {{ ApidocRouteHelper::scopeBadge('bot') }} scope, they may be allowed to act on behalf of the owner of the OAuth client (delegation). When using delegation, scopes that support delegation cannot be used together with scopes that do not support delegation.
+
+The following scopes currently support delegation:
+
+Name   |
+-------|
+{{ ApidocRouteHelper::scopeBadge('chat.write') }}
```
nanaya, line 273: This ended up not being part of the table

nanaya (changes requested):

app/Models/OAuth/Token.php
```diff
@@ -85,30 +91,37 @@ public function validate()
...
+                throw new InvalidScopeException('* is not allowed with Client Credentials');
             }
 
-            // in the case of client credentials, delegation must be enabled on the token.
-            if ($this->isClientCredentials() && !$this->canDelegate()) {
-                throw new InvalidScopeException('bot scope required.');
-            }
-        }
+            if (!$scopes->intersect($scopesRequireDelegation)->isEmpty()) {
+                if (!$this->canDelegate()) {
```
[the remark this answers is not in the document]
notbakaneko: the option is to not request the scope 🤔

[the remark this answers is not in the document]
nanaya: Yeah but if the token has the scope, it's not "canDelegate" anymore but "isDelegating" (or something else with better wording).

nanaya:
I'm not sure if I understand the PR title btw ಠ_ಠ

nanaya (approved):

PR #7953, opened 2021-07-31
Fix wording for mapping follow tooltips

yfxu:
- change "beatmap" to "beatmaps"

pretty simple change that no one _really_ needs but thought I'd fix it up anyway since I noticed it

### DIFF
--- a/resources/lang/en/follows.php
+++ b/resources/lang/en/follows.php
@@ -28,8 +28,8 @@ return [
         'followers' => 'mapping subscribers',
         'page_title' => 'mapper watchlist',
         'title' => 'mapper',
-        'to_0' => 'stop notifying me when this user uploads new beatmap',
-        'to_1' => 'notify me when this user uploads new beatmap',
+        'to_0' => 'stop notifying me when this user uploads new beatmaps',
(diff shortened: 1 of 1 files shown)


resources/lang/en/follows.php
```diff
@@ -28,8 +28,8 @@
         'followers' => 'mapping subscribers',
         'page_title' => 'mapper watchlist',
         'title' => 'mapper',
-        'to_0' => 'stop notifying me when this user uploads new beatmap',
-        'to_1' => 'notify me when this user uploads new beatmap',
+        'to_0' => 'stop notifying me when this user uploads new beatmaps',
+        'to_1' => 'notify me when this user uploads new beatmaps',
```
peppy, line 32:
```suggestion
        'to_0' => 'stop notifying me when this user uploads a new beatmap',
        'to_1' => 'notify me when this user uploads a new beatmap',
```

I think this reads better.

peppy (changes requested):

resources/lang/en/follows.php
yfxu: Yea either is fine in my opinion. Let's go with your suggestion.

peppy (approved):

PR #7960, opened 2021-08-02
Remove country flags drop shadow

nanaya:
Apparently the shadow causes lag for some people. It was done that way because some flags are shaped differently.

People affected should check if this test page is actually smoother: https://osunext.dev.myconan.net/rankings/osu/performance

### DIFF
--- a/resources/assets/less/bem/flag-country.less
+++ b/resources/assets/less/bem/flag-country.less
@@ -12,7 +12,7 @@
   background-repeat: no-repeat;
   border-radius: 3px;
   position: relative;
-  filter: saturate(1.1) drop-shadow(0 1px @box-shadow-radius @box-shadow-color);
+  filter: saturate(1.1);
 
   &::after {
     .full-size();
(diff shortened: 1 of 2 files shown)


peppy:
Still a bit weird when scrolling (I don't see this kind of behaviour on any other page) but by far a lot better:

https://user-images.githubusercontent.com/191335/127856054-7067a971-8f10-4f64-9ec1-b34067faecbd.MP4

nanaya:
uh that was a wrong url :eyes:

peppy:
Oops, iOS is weird with opening background tabs:

https://user-images.githubusercontent.com/191335/127856844-257dc703-5148-474d-a751-8d677a00714b.MP4

notbakaneko:
With latest update:
https://user-images.githubusercontent.com/1694544/127861300-1fb557bc-48ba-47af-8e0e-fbf53c290197.mp4

notbakaneko (approved):

peppy (approved):

PR #7965, opened 2021-08-03
Change multiplayer realtime category into types

### LINKED ISSUE #7962: Add new parameter to room table and creation API request
We are expanding osu!(lazer) to offer Team VS next, for multiplayer games. This will only be supported for realtime rooms. In order to allow this to happen, a new parameter should be added to `/rooms POST` endpoint:

`type { playlists, headtohead, teamvs }`

Should also be added to `multiplayer_rooms` as an `enum` type I guess.

Note that we will probably want to reconsider how the existing `category` type is used. I believe the `realtime` setting should not need to exist with this new type enum. It may be that we want to have "spotlighted" realtime rooms in the future, so probably fine to leave it existing for that purpose.

One could alternatively argue that we could add these new types to the existing `category` enum, and if that's deemed as a better direction then it can likely still work.

So either we would end up with option 1:

```csharp
category { normal, spotlight } // realtime is removed either immediately or as a second step
type {playlists, headtohead, teamvs }
```

or option 2:

```csharp
category { normal, spotlight, realtimevs, realtimeteamvs } // not sure about naming
```

This is semi high priority as it's a blocker for making initial game creation work (currently the type is only able to be changed post-creation). @nanaya @notbakaneko @smoogipoo

nanaya:
<del>Fake category value in response for now until client is updated....</del> or is it not needed?

Resolves #7962.

### DIFF
--- a/app/Models/Multiplayer/Room.php
+++ b/app/Models/Multiplayer/Room.php
@@ -12,6 +12,7 @@ use App\Models\Model;
 use App\Models\User;
 use App\Traits\WithDbCursorHelper;
 use Carbon\Carbon;
+use Ds\Set;
 use Illuminate\Database\Eloquent\SoftDeletes;
 
 /**
@@ -48,6 +49,10 @@ class Room extends Model
 
     const DEFAULT_SORT = 'created';
 
+    const PLAYLIST_TYPE = 'playlists';
+    const REALTIME_DEFAULT_TYPE = 'headtohead';
+    const REALTIME_TYPES = ['headtohead', 'teamvs'];
+
     protected $casts = [
         'password' => PresentString::class,
     ];
@@ -66,10 +71,18 @@ class Room extends Model
         $sort = $params['sort'] ?? null;
 
         $category = presence(get_string($params['category'] ?? null)) ?? 'any';
-        if ($category === 'any') {
-            $query->where('category', '<>', 'realtime');
-        } else {
-            $query->where('category', $category);
+        switch ($category) {
+            case 'any':
+                $query->where('type', static::PLAYLIST_TYPE);
+                break;
+            case 'realtime':
+                $query->whereIn('type', static::REALTIME_TYPES);
+                break;
+            default:
+                $query->where([
+                    'type' => static::PLAYLIST_TYPE,
+                    'category' => $category,
+                ]);
         }
 
         switch ($mode) {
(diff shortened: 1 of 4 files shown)


peppy:
I should have probably mentioned that client-side we use camel case for these. If that suits better then it's probably best to use that in the API portion (my original issue used all lower only to match the mysql enums).

nanaya:
Newer enum for osu-web are `snake_case`d so the current one is kinda wrong as well (should be `head_to_head` and `team_vs`?).

peppy:
That's going to be an eternal struggle. I think we do have the ability to automatically parse snake to camel, might be worth investigating from our end..

app/Models/Multiplayer/Room.php
```diff
@@ -48,6 +49,10 @@ class Room extends Model
 
     const DEFAULT_SORT = 'created';
 
+    const PLAYLIST_TYPE = 'playlists';
+    const REALTIME_DEFAULT_TYPE = 'headtohead';
+    const REALTIME_TYPES = ['headtohead', 'teamvs'];
```
peppy, line 54: `teamvs` -> `teamversus` (we changed it in a naming pass earlier today)

frenzibyte:
I believe we do have that via JSON converters:
 - https://github.com/ppy/osu/blob/master/osu.Game/IO/Serialization/Converters/SnakeCaseStringEnumConverter.cs
 - https://github.com/ppy/osu/blob/b43f5f755d536873ba2e598b027dafb07eef4377/osu.Game/Online/Rooms/Room.cs#L44-L50

peppy:
> I believe we do have that via JSON converters:
> 
> * https://github.com/ppy/osu/blob/master/osu.Game/IO/Serialization/Converters/SnakeCaseStringEnumConverter.cs
> * https://github.com/ppy/osu/blob/b43f5f755d536873ba2e598b027dafb07eef4377/osu.Game/Online/Rooms/Room.cs#L44-L50

@nanaya should be fine to use snake case here then, while you're at it.

peppy:
> Fake category value in response for now until client is updated.... or is it not needed?

What is this referring to?

nanaya:
> > Fake category value in response for now until client is updated.... or is it not needed?
> 
> What is this referring to?

The json response for rooms.

peppy:
Right. I'll check on that.

app/Transformers/Multiplayer/RoomTransformer.php
```diff
@@ -26,7 +26,8 @@ public function transform(Room $room)
         return [
             'id' => $room->id,
             'name' => $room->name,
-            'category' => $room->category,
+            'category' => $room->isRealtime() ? 'realtime' : $room->category,
```
peppy, line 29: Looks like this isn't required. We only ever use this response to check whether it is `Spotlight` or not.

peppy (dismissed):

notbakaneko (approved):

PR #7978, opened 2021-08-05
Include recent participants in rooms index

### LINKED ISSUE #7972: Multiplayer/playlist room indexing requests should return recent participants
Flyte's new designs list recent participants in the lounge before the room is selected/entered:
image

But recent participants are currently only returned in `RoomsController.show()`:
https://github.com/ppy/osu-web/blob/f51a514e911e08b75e31bb1f10179adaa93e5cb9/app/Http/Controllers/Multiplayer/RoomsController.php#L110-L124

And not in `RoomsController.index()`:
https://github.com/ppy/osu-web/blob/f51a514e911e08b75e31bb1f10179adaa93e5cb9/app/Http/Controllers/Multiplayer/RoomsController.php#L29-L34

nanaya:
The query involved is wunderbar ( ﾟ ヮﾟ) (also the way preloading is done)

Haven't checked if index is required but doesn't seem too slow.

Resolves #7972.


smoogipoo:
If performance is an issue, this can be reduced to 8 users or so.

notbakaneko (approved):

PR #7980, opened 2021-08-06
Change `fragment-target` to `scroll-padding-top`

gagahpangeran:
- Simplify the implementation.
- Now it works for inline element.


resources/assets/less/functions.less
```diff
@@ -59,16 +59,11 @@
...
   position: static; // relative will still block pointer-events
 
-  &[id]::before {
-    content: '';
-    display: block;
-    height: @nav2-height--pinned + @offset;
-    margin: -(@nav2-height--pinned + @offset) 0 0;
-    pointer-events: none;
+  &[id] {
+    scroll-margin-top: @nav2-height--pinned + @offset;
```
nanaya, line 63: see if removing this function entirely and setting the scroll-margin-top globally instead is better

(also it's only supported in safari since last april so I'm not sure if we can migrate just yet)

nanaya (changes requested):

resources/assets/less/functions.less
gagahpangeran: By you mean 'globally' is using wildcard `*` or what?

Hmm, only recent safari version support it :thinking:

nanaya: yes, with wildcard

gagahpangeran: Yeah, it works correctly using wildcard.

But we still need offset support for something like this.

https://github.com/ppy/osu-web/blob/d3f74531023a46126490c4eac03f8679f1dbc207/resources/assets/less/bem/ranking-page.less#L11

I guess add variable `--scroll-margin-top-offset` and set it locally should works?

notbakaneko (changes requested):
set `scroll-padding-top` at the top level and `scroll-margin-top` on any elements that need additional margin.

The issue with `scroll-*-top` has always been Safari not supporting it properly until the current version.

gagahpangeran:
Already test this changes with firefox and chrome (desktop and mobile), seems to work.

Can't testing in safari (desktop and mobile) because I don't have any apple device.

resources/assets/coffee/react/_components/comments.coffee
```diff
@@ -27,7 +27,7 @@ export class Comments extends React.PureComponent
 
       div className: classWithModifiers('comments', @props.modifiers),
         div className: 'u-has-anchor u-has-anchor--no-event',
-          div(className: 'fragment-target', id: 'comments')
+          div(id: 'comments')
```
nanaya, line 30: `u-has-anchor` and the standalone id element can probably be just removed and the `id` moved to the actual target element.

nanaya (changes requested):

resources/assets/less/base.less
```diff
@@ -4,6 +4,14 @@
 // main: compiled.less
 @import "app";
 
+* {
+  scroll-padding-top: @nav2-height--pinned;
```
notbakaneko, line 8: `scroll-padding-top` only needs to be applied to `html` (and still doesn't work properly in safari)

notbakaneko (changes requested):

resources/assets/less/base.less
gagahpangeran: Hmm, I thought safari already full support it in version 14.1 (MacOS) and 14.5 (iOS) because I saw it in MDN :thinking:

resources/assets/coffee/react/_components/comments.coffee
gagahpangeran: Done.

resources/assets/less/base.less
notbakaneko: `scroll-padding` and `scroll-margin` are getting fixed in different versions for some reason 👀

gagahpangeran: Hmm, right, Only `scroll-margin-top` currently have full support.

Should we use `scroll-margin-top` instead in `html` selector and override it locally if needs offset? Currently only one place that need offset.

gagahpangeran:
This is the only place that needs 100px offset when using `fragment-target`.

Screenshot from 2021-09-09 17-50-52

Maybe we can move the `#scores` id to somewhere else.

Also with current 100px offset the position looks awkward :thinking:

resources/assets/lib/oauth/own-clients.tsx
```diff
@@ -33,14 +33,15 @@ export class OwnClients extends React.Component {
...
-        <BigButton
-          icon={'fas fa-plus'}
-          props={{
-            onClick: this.handleNewClientClicked,
-          }}
-          text={osu.trans('oauth.own_clients.new')}
-        />
+        <div id='new-oauth-application'>
+          <BigButton
+            icon={'fas fa-plus'}
```
nanaya, line 38: remove the brackets while at it

nanaya (changes requested):

resources/assets/less/layout.less
```diff
@@ -54,6 +54,14 @@ html, body {
...
 }
 
+html {
+  scroll-padding-top: @nav2-height--pinned;
+
+  @media @mobile {
+    scroll-padding-top: @navbar-height;
+  }
+}
+
```
gagahpangeran, line 64: I *think* for this to work in safari is change `scroll-padding-top` to `scroll-margin-top`, because currently safari only full support `scroll-margin` based on mdn.

Also if we use `scroll-margin`, we have to use `*` instead of `html`.

gagahpangeran:
@nanaya did you already check this on safari?

I doubt it works because MDN said safari does not support `scroll-padding` yet, it only `scroll-margin`.

See https://github.com/ppy/osu-web/pull/7980#discussion_r712901769 for proposed solution.

nanaya:
it works on ios. Also the table is probably outdated (mdn/browser-compat-data#12483).

gagahpangeran:
Hmm, okay.

notbakaneko (approved):

PR #7983, opened 2021-08-06
Fix beatmapset nomination sync command

notbakaneko:
Missing nomination modes from the migration in #7793
Also fixes nominations being marked as reset when they shouldn't be if the command is run multiple times.

Should be able to just truncate `beatmapset_nominations` and run `php artisan beatmapset:nomination-sync` again.
Tried to use `upsert` but it seems `upsert` doesn't like `mode` value being an array 🤔


nanaya (approved):
we need to actually run this :upside_down_face:

PR #7988, opened 2021-08-09
Fix chat input autosize height

### LINKED ISSUE #7918: osu!web chat in Android refused to scroll down when typing.
https://user-images.githubusercontent.com/79167465/126791802-4b99f71f-c04a-48a6-957b-b0dfb16d0260.mp4

notbakaneko:
closes #7918

Manually apply `max-height` via css.
Likely due to #7798 causing the line height calculation to be `NaN` on first render when arriving via turbolinks navigation.
Can't just call `updateLineHeight` on `TextareaAutosize` after mounting as it's a `forwardRef`.

### DIFF
--- a/resources/assets/less/bem/chat-input.less
+++ b/resources/assets/less/bem/chat-input.less
@@ -2,7 +2,9 @@
 // See the LICENCE file in the repository root for full licence text.
 
 .chat-input {
-  padding: 10px 30px 10px 50px;
+  @_height: 80px;
+  @_vertical-padding: 10px;
+
   color: white;
   position: absolute;
   bottom: 0;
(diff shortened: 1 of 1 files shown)


resources/assets/less/bem/chat-input.less
```diff
@@ -2,29 +2,34 @@
...
-    padding: 10px 0px 10px 20px;
+    padding-right: 0;
+    padding-left: 20px
   }
 
   &__box {
     .reset-input();
     .default-border-radius();
     .default-box-shadow();
+    max-height: @_height - @_vertical-padding * 2;
```
nanaya, line 32: 100% seem to work as well...? :thinking: Less variable math etc

nanaya (changes requested):

nanaya (approved):

PR #7991, opened 2021-08-10
Fix callback props for React components

nanaya:
I made an attempt to also remove `any` typing in `stateful-activation-context` but then it's more involved than I hoped ( ﾟ ヮﾟ)


resources/assets/lib/quick-search/main.tsx
```diff
@@ -103,7 +103,14 @@ const otherModes: ResultMode[] = ['forum_post', 'wiki_page'];
     }
   };
 
-  private onMouseLeave = (event: React.MouseEvent<HTMLInputElement>) => {
+  private onMouseEnter = (event: React.MouseEvent<HTMLDivElement>) => {
+    const section = event.currentTarget.dataset.section as Section;
+    const index = parseInt(event.currentTarget.dataset.index ?? '0', 10);
```
notbakaneko, line 108: kind of moot since these values are set privately, but the function should bail out if they're not defined/valid instead of coalesced 🤔

notbakaneko (changes requested):

resources/assets/lib/quick-search/main.tsx
nanaya: The `_others` don't actually have index...

notbakaneko: 🤔

notbakaneko (approved):

PR #7995, opened 2021-08-12
Convert BBCode Editor component to TypeScript

nanaya:
One component at a time :snail: 

The props don't quite match with how it's used but it's the other side that needs fixing (disable on submit).


resources/assets/lib/bbcode-editor.tsx
```diff
@@ -0,0 +1,210 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+import { snakeCase } from 'lodash';
+import * as React from 'react';
+import { classWithModifiers, Modifiers } from 'utils/css';
+
+interface OnChangeProps {
+  event?: React.SyntheticEvent;
+  hasChanged: boolean;
+  type: 'cancel' | 'save';
```
gagahpangeran, line 11: This type can be extracted and reused on `sendOnChange` below.

resources/assets/lib/bbcode-editor.tsx
```diff
@@ -0,0 +1,212 @@
...
+interface OnChangeProps {
+  event?: React.SyntheticEvent;
+  hasChanged: boolean;
+  type: ChangeType;
+  value: string | undefined;
+}
+
+interface Props {
+  disabled: boolean;
+  modifiers: Modifiers;
```
gagahpangeran, line 19: Should modifiers be optional? :thinking:

notbakaneko (approved):

PR #7996, opened 2021-08-12
Only decrement user stats when deleting personal best score

### LINKED ISSUE #7152: Negative rank count
image
[https://osu.ppy.sh/users/10690090](url)

nanaya:
Resolves #7152.


app/Jobs/RemoveBeatmapsetBestScores.php
```diff
@@ -64,7 +64,12 @@ public function handle()
 
             $class = static::scoreClass($mode);
```
notbakaneko, line 65: let's fix the non-static method while we're at it 👀

notbakaneko (changes requested):

app/Jobs/RemoveBeatmapsetBestScores.php
nanaya: how does it even work ಠ_ಠ

notbakaneko (approved):

notbakaneko (approved):

PR #8014, opened 2021-08-16
Refactor chat presence response

notbakaneko:
Now just uses `ChannelTransformer`.
Some additional moderator checks should be fixed now as well.

preloaded users stored in `request()->attributes` for now; this should be moved to a container object if this kind of lookup is useful for other stuff.


app/Models/Chat/Channel.php
```diff
@@ -387,6 +426,16 @@ public function hasUser(User $user)
         ])->exists();
     }
 
+    public function setUserChannelFor(User $user, UserChannel $userChannel)
+    {
+        if ($userChannel->user_id !== $user->getKey()) {
+            throw new InvariantException('userChannel does not belong to the same user.');
+        }
+
+        $userChannel->setRelation('user', $user);
```
nanaya, line 435: this can be done as part of `userChannelFor`?

app/Models/Chat/Channel.php
```diff
@@ -387,6 +426,16 @@ public function hasUser(User $user)
...
+        $this->userChannel = $userChannel;
```
nanaya, line 436: store in an array keyed by user id? Can then also be set by the getter function

app/Models/Chat/Channel.php
```diff
@@ -387,6 +426,16 @@ public function hasUser(User $user)
         ])->exists();
     }
 
+    public function setUserChannelFor(User $user, UserChannel $userChannel)
+    {
+        if ($userChannel->user_id !== $user->getKey()) {
```
nanaya, line 431: probably not going to make much sense but can use check of `channel_id`

app/Http/Controllers/ChatController.php
```diff
@@ -26,19 +26,17 @@ public function index()
...
-            $canMessage = priv_check('ChatStart', $targetUser)->can();
-
             $channel = Channel::findPM($targetUser, $user);
-            optional($channel)->addUser($user);
+            $channel?->addUser($user);
 
             $json['send_to'] = [
-                'can_message' => $canMessage,
-                'channel_id' => optional($channel)->getKey(),
+                'can_message' => $channel?->canMessage($user) ?? false,
```
nanaya, line 38: haven't checked but this seem to be always false for new channels? (`?sendto=user` where there's no pm channel with the `user` yet)

nanaya (changes requested):

app/Transformers/Chat/ChannelTransformer.php
```diff
@@ -50,6 +52,11 @@ public function includeLastMessageId(Channel $channel)
         return $this->primitive($channel->last_message_id);
     }
 
+    public function includeLastReadId(Channel $channel)
+    {
+        return $this->primitive($channel->lastReadIdFor($this->user));
+    }
```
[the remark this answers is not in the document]
notbakaneko: I was wonder whether it'd be better to change the whole response format to `UserChannel[]` instead 👀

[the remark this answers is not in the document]
nanaya: An alternative is `current_user_attributes` like currently in BeatmapDiscussion and BeatmapsetCompact :thinking:

nanaya (changes requested):

nanaya (approved):

PR #8022, opened 2021-08-18
Implement new notification header design

venix12:
based on the figma designs
@arflyte not sure about the hover design for the buttons (assumed less transparent background-color for the notification one) 🤔


peppy:
@arflyte can you confirm intention here?

This is how it looks with this change:

20210820 121530 (Safari)

- Not sure about the chat icon having a background while the notifications one doesn't
- The twitter/support/flag feel very weirdly padded, and floating a bit.

notbakaneko:
notification icons having a stadium border and the other not having any seems odd, to be honest.

notbakaneko:
Rewinding at bit, the original reason for the chat and notification icons being turned into a single "button" was because all notifications went into a single widget - this isn't the case any more and chat notifications are in a separate widget.

peppy:
@arflyte can you advise of whether this is still correct?

arflyte:
This is still the latest design. Social media and language has no border to keep the header less busy. The transparent is actually for hover and when it's actually activated.

peppy:
@arflyte and the twitter etc. icons don't have any kind of hover circle? is this the direction for buttons going forward? i'm not sure i like it all that much.

arflyte:
Not in the design, but a hover transparent circle can be easily added to twitter etc. Just add a lil bit more padding to the icon.

nanaya:
There has been no update here so I updated it applying the latest feedback.

notbakaneko (approved):

PR #8025, opened 2021-08-19
Move url helpers to utils

nanaya:
Better place for this file. Also fewer static class methods and more specific function imports.

And one time regexp generation.


notbakaneko (changes requested):
conflicts

notbakaneko (approved):

PR #8036, opened 2021-08-20
Update MobX

### LINKED ISSUE #6790: Update to MobX 6
["It's quite different from MobX 5"](https://mobx.js.org/migrating-from-4-or-5.html) (or maybe not if all the decorator options are enabled and maybe adjust some of the config as needed)

Also the docs for MobX 5 is now just a pile of [md files](https://github.com/mobxjs/mobx/tree/mobx4and5/docs).

nanaya:
Let's just get this over with. Still need more testing but it at least compiles and chat and comments and beatmapset search seem to work.

Resolves #6790.


notbakaneko (changes requested):
missing `makeObservable`
- beatmapset-panel.tsx
- beatmap-owner-editor.tsx
- beatmaps-owner-editor.tsx
- love-beatmap-modal.tsx
- multiplayer-history.tsx

needs `@action`
- `OsuCore.setCurrentUser`
- `UserPreferences.setUser`
- `UserPreferences.set`
- `BeatmapsetPanel.toggleFavourite` (because beatmapset prop comes from observable map)

notbakaneko (changes requested):
conflicts 👀

notbakaneko (approved):

PR #8043, opened 2021-08-21
[DE] Fix typo

Pasi4K5:
Just a minor typo.


frenzibyte:
As stated in the [readme](https://github.com/ppy/osu-web#contributing), any changes to localised strings should happen at the crowdin projects instead rather than in here.

PR #8053, opened 2021-08-25
Convert user profile account standing page to TypeScript

nanaya:
- [x] depends on #8052


resources/assets/lib/profile-page/account-standing.tsx
```diff
@@ -0,0 +1,145 @@
...
+            </tbody>
+          </table>
+        </div>
+      </div>
+    );
+  }
+
+  private readonly renderColumn = (column: Column, history: UserAccountHistoryJson) => (
+    <td key={column} className={classWithModifiers(`${bn}__table-cell`, column)}>
+      {React.createElement(content[column], { history })}
```
notbakaneko, line 136:
```
const Elem = content[column];
<Elem .... />
```
also works

notbakaneko (approved):
should be fine

resources/assets/lib/profile-page/account-standing.tsx
nanaya: Well if you see the last comment :upside_down_face:

notbakaneko: 🙃

PR #8063, opened 2021-08-27
Clean up `BeatmapsetExtendedJson` interface

nanaya:
- move to `interfaces` directory
- rename to follow `ThingExtendedJson` name
- export default
- remove redundant `BeatmapsetEvent` interface
- remove nonexistent fields from `BeatmapsetEventJson`


resources/assets/lib/utils/beatmap-helper.ts
```diff
@@ -1,10 +1,10 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
 // See the LICENCE file in the repository root for full licence text.
 
-import { BeatmapsetJson } from 'beatmapsets/beatmapset-json';
 import * as d3 from 'd3';
+import { isValid as isBeatmapJsonExtended } from 'interfaces/beatmap-extended-json';
```
notbakaneko, line 5: should rename the alias too 👀

notbakaneko (changes requested):

resources/assets/lib/utils/beatmap-helper.ts
nanaya: great (ノ･_･)ノ︵ ┻━┻

notbakaneko (approved):

PR #8087, opened 2021-09-06
Update commonmark to v2.0

gagahpangeran:
- [x] depends on #8091

In this PR:

- updating commonmark from v1.6 to v2.0.
- Change almost all namespace and class because so many breaking changes.
- add more markdown test.
- separate config in `OsuMarkdown`. Because in v2.0 config schema needs to be declared, I decided to separate the config.
- use `DefaultAttributesExtension` to add class name and other simple attributes. This reduce complexity in `DocumentProcessor` by removing method `addClass` and `addListStartAsVariable`, and also support type hint when accessing the node via its callable. (can be revert if unnecessary in this PR)

After back and forth reading breaking changes documentation and also dive into commonmark code base, finally it's here!

There is so much breaking changes in this version, from API changes to namespace changes.

Reference to read:
- https://commonmark.thephpleague.com/2.0/upgrading/developers/
- https://commonmark.thephpleague.com/2.0/changelog/

I hope I didn't miss something.

PR #8088, opened 2021-09-06
Add back style block with new syntax

gagahpangeran:
- [ ] Depends on #8087

After some discussion in #osu-wiki channel in discord. We decided to change the styleblock syntax to maintain compatibility with [markdig custom container syntax](https://github.com/xoofx/markdig/blob/master/src/Markdig.Tests/Specs/CustomContainerSpecs.md), so we don't have to implement custom extension to parse style block in osu!-side.

The implementation of parser is similar with `FencedCode` parser from commonmark.

Example syntax:

```
::: Class Name
Some content
:::
```

```
::: Class Name
Some content
:::: Class Name
Nested content
::::
:::
```


gagahpangeran:
Closing as part of #8087

PR #8090, opened 2021-09-06
New featured artist track listing design

nanaya:
Not-so-minimal changes for the track listing design update to be used in search page.

The rows are quite a lot bigger compared to current one even after I reduce cover image size.

Actual final styling may differ.

And @arflyte, design for mobile is missing.

depends on:
- [ ] #8079
- [ ] #8080


resources/assets/lib/tracklist-track.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+        <div className='artist-track__col artist-track__col--buttons'>
+          <button
+            className='artist-track__button js-audio--play js-audio--player'
+            data-audio-url={this.props.track.preview}
+          >
+            <span className='fa-fw play-button' />
+          </button>
+
+          {this.props.track.osz == null ? (
```
notbakaneko, line 79: `osz` is non-nullable.
The icon should also probably have something to differentiate it from one that's available 🤔

resources/assets/lib/tracklist-track.tsx
```diff
@@ -0,0 +1,99 @@
...
+            >
+              {osu.trans('artist.songs.original_badge')}
+            </span>
+          )}
+          {this.props.track.is_new && (
+            <span className='pill-badge pill-badge--yellow pill-badge--with-shadow'>
+              {osu.trans('common.badges.new')}
+            </span>
+          )}
+        </div>
```
notbakaneko, line 60: could do with some spacing when more than one badge shows

notbakaneko (changes requested):
really needs a mobile design 👀

PR #8099, opened 2021-09-08
Split chat-related json exports into separate modules

notbakaneko:
Extracting some changes out of the conversion of chat to use websockets. 

Shared stuff goes into separate modules; local interfaces can just go into the same module that's using it.


resources/assets/lib/interfaces/message-json.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import UserJson from 'interfaces/user-json';
+
+export default interface MessageJson {
```
nanaya, line 6: seems better either inside `interfaces/chat/` directory or prefixed with Chat (also ChannelJson and maybe ChatUpdatesJson)

nanaya (changes requested):

nanaya (approved):

PR #8100, opened 2021-09-08
Fix MobX `@computed` usage

nanaya:
It turned out some of the `@computed`s aren't actually memoized ( ﾟ ヮﾟ)a


resources/assets/lib/app-deps.ts
```diff
@@ -38,3 +39,7 @@ window.LangMessages ??= {};
 window.Lang = new Lang({ messages: window.LangMessages });
 window.moment = moment;
 window.Turbolinks = Turbolinks;
+
+mobxConfigure({
```
nanaya, line 43: is this even the correct place to put this :thinking:

notbakaneko: as long as it runs before anything else mobx does 👀

resources/assets/lib/chat/conversation-list.tsx
```diff
@@ -31,6 +32,7 @@ export default class ConversationList extends React.Component {
     return channels.map((conversation) => <ConversationListItem key={conversation.channelId} channelId={conversation.channelId} />);
   }
 
+  @action
```
notbakaneko, line 35: This `action` doesn't seem to do anything?; also looks like `nonPmChannels` and `pmChannels` in `ChannelStore` should be `@computed.struct` instead of just `@computed`.

notbakaneko (changes requested):

notbakaneko (approved):

PR #8103, opened 2021-09-09
Tidy up some chat-related classes

notbakaneko:
- use `get_params`
- add typings to function arguments
- do the content length check before the rate check
- instantiate transformer object instead of magic string


app/Http/Controllers/Chat/Channels/MessagesController.php
```diff
@@ -74,30 +74,30 @@ class MessagesController extends BaseController
      */
     public function index($channelId)
     {
-        $request = request()->all();
-        $userId = Auth::user()->user_id;
-        $since = get_int($request['since'] ?? null);
-        $until = get_int($request['until'] ?? null);
-        $limit = clamp(get_int($request['limit'] ?? null) ?? 50, 1, 50);
+        $params = get_params(request()->all(), null, [
```
nanaya, line 77: either destructure here or use `$params[...]` as is?

nanaya (changes requested):

nanaya (approved):

PR #8105, opened 2021-09-09
Home button cleanup

nanaya:
Use updated colour... except I'm not sure if better. At least the css is probably better with css variables and stuff.

(was cleaning up different thing but ended up doing this as well)

depends on:
- [x] #8104


notbakaneko (approved):
not sure about the colour...I guess we'll see 🤔 

image
image

PR #8110, opened 2021-09-10
Support selectively excluding data from chat updates

notbakaneko:
Allows specifying includes to skip returning data from the response.
Defaults to `true` for compatibility.

The web socket version of web chat doesn't need any of the messages, nor does it need `presence` except when reconnecting. `silences` will still be polled on a schedule for the moment.


nanaya:
Conflicts

app/Http/Controllers/Chat/ChatController.php
```diff
@@ -123,65 +124,88 @@ public function __construct()
...
+
+            $messages = $messages->get()->reverse();
+
+            $response['messages'] = json_collection($messages, new MessageTransformer(), ['sender']);
         }
 
-        $messages = $messages->get()->reverse();
+        if ($includePresence) {
+            // to match old behaviour
+            $response['presence'] = $includeMessages && ($messages ?? collect())->isEmpty()
```
nanaya, line 180: `$includeMessages` already means `$messages` will never be null?

app/Http/Controllers/Chat/ChatController.php
```diff
@@ -123,65 +124,88 @@ public function __construct()
...
+            $messages = Message
+                ::with('sender')
+                ->whereIn('channel_id', $channelIds)
+                ->since($since)
+                ->limit($limit)
+                ->orderBy('message_id', 'DESC');
+
+            if ($params['channel_id'] !== null) {
+                $messages->where('channel_id', $params['channel_id']);
+            }
```
nanaya, line 171: looks like this can be combined with `$channelIds` above (`$channelIds = $params[...] ... ? [$params[...]] : array_...`)

and then the query & fetch can be done at once

app/Http/Controllers/Chat/ChatController.php
```diff
@@ -123,65 +124,88 @@ public function __construct()
...
-        return [
-            'presence' => $presence,
-            'messages' => json_collection(
-                $messages,
-                new MessageTransformer(),
-                ['sender']
-            ),
-            'silences' => json_collection($silences, 'Chat\UserSilence'),
-        ];
+        $isEmpty = array_reduce($response, fn ($acc, $val) => $acc + count($val), 0) === 0;
```
nanaya, line 206: Flip the check with `array_first`?

(also is it `[]` or `null` for empty response?)

```php
$isAny = array_first($response, fn ($val) => count($val) > 0) !== null;

return $isAny ? $response : response(null, 204);
```

app/Http/Controllers/Chat/ChatController.php
```diff
@@ -123,65 +124,88 @@ public function __construct()
...
-        $params = request()->all();
+        $request = request()->all();
+        $params = get_params($request, null, [
+            'channel_id:int',
+            'history_since:int',
+            'limit:int',
+            'since:int',
+        ], ['null_missing' => true]);
+
+        $includes = get_params($request, 'includes', [
```
nanaya, line 135: should this be this or `includes[]=messages&includes[]=presence` parsed above with `string[]` and use `Set` (with default everything when null/not specified/empty(?)) :thinking:

nanaya (changes requested):

app/Http/Controllers/Chat/ChatController.php
notbakaneko: or we can use `response()->noContent()` and not think about it 🤔

app/Http/Controllers/Chat/ChatController.php
```diff
@@ -123,65 +125,88 @@ public function __construct()
      */
     public function updates()
     {
-        $params = request()->all();
+        static $availableIncludes;
+        $availableIncludes ??= new Set(['messages', 'presence', 'silences']);
 
-        if (!present($params['since'] ?? null)) {
+        $request = request()->all();
+        $params = get_params($request, null, [
```
nanaya, line 132:
```suggestion
        $params = get_params(request()->all(), null, [
```

app/Http/Controllers/Chat/ChatController.php
```diff
@@ -123,65 +125,88 @@ public function __construct()
      */
     public function updates()
```
nanaya, line 126: just noticed the docs can use some updates

nanaya (changes requested):

nanaya (approved):

PR #8115, opened 2021-09-14
Fix cover-editor being below page-extra-tabs

RockRoller01:
#7909 increased the z-index to prevent beatmap-card borders to be above the extra tabs, however this lead to the cover selector being below them now.

image

Increased the editors index and left a note that it must be at least 1 higher than extra-tabs

### DIFF
--- a/resources/assets/less/variables.less
+++ b/resources/assets/less/variables.less
@@ -92,10 +92,10 @@
 @z-index--wiki-search: 1;
 
 @z-index--beatmaps-panel: 4;
-@z-index--profile-page-cover-selector: 100; // must be at least 1 more than @z-index--page-extra-tabs
 @z-index--profile-previous-usernames: 10;
 @z-index--profile-tournament-banner: 11;
 @z-index--page-extra-tabs: 99;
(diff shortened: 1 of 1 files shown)


nanaya:
It's actually kinda expected as the floating bar at the bottom is considered the same as menu bar at the top

I have no problem either way but if we're going with this, at least the line should be moved according to the actual value which should've been obvious if you look at the whole section.

RockRoller01:
ah I didn't see that these are listed in increasing order. Will change in a bit. imo this must be changed as the tabs otherwise cover up the information on the cover, leaving people clueless as to size and supporter requirement

nanaya:
You can scroll the page down to see it...

RockRoller01:
> You can scroll the page down to see it...

no you actually can not, the cover uploader is stuck in position

nanaya:
> no you actually can not, the cover uploader is stuck in position

:thinking:

RockRoller01:
ah i see why you dont see the issue. Apparently a very specific setup is needed:

- ranking folded in
- 1 row of badges
- 
image

nanaya (approved):

PR #8117, opened 2021-09-15
Show custom notice on stub pages of the wiki

TicClick:
while we have the `stub: true` marker, the wiki doesn't use it at all and shows no notice

<details>
<summary>after the change</summary>

image

</details>


resources/lang/en/wiki.php
```diff
@@ -12,6 +12,7 @@
         'missing_translation' => 'Requested page could not be found for currently selected language.',
         'needs_cleanup_or_rewrite' => 'This page does not meet the standards of the osu! wiki and needs to be cleaned up or rewritten. If you are able to help out, please consider updating the article!',
         'search' => 'Search existing pages for :link.',
+        'stub' => 'This article is a stub. You can help osu! wiki by expanding it.',
```
peppy, line 15: Reads pretty weirdly.

- If it's going to directly say "you can help" then it needs to be a link
- The term "stub" doesn't mean much to a lot of people I think. "Incomplete" feels better to me.

Maybe something like "This article is incomplete and waiting on someone to expand it"?

TicClick: making it a link was my original idea, as well as adding it to the `outdated`/`needs_cleanup` articles. I couldn't decide what to reference, the article on GitHub, or `CONTRIBUTING. md`, or https://osu.ppy.sh/wiki/OWCG (the latter would be more fair, but it's not in the best shape right now)

TicClick:
taking your suggestion for now; also, I figured that we have articles with more than two notices, so I made the stub one [separate](https://user-images.githubusercontent.com/4686101/133483617-59c2de1e-745b-44f7-acde-fb9f12024efa.png)

```sh
user@cube:~/stuff/osu-wiki$ find wiki/ -type f -name "*.md" -print0 | xargs -0 grep -l "stub: true" | xargs grep "outdated: true" | wc -l
12
```

nanaya (approved):

PR #8120, opened 2021-09-15
Update Solo Score table structure

nanaya:
This drops existing data ( ﾟ ヮﾟ)a


app/Models/Solo/Score.php
```diff
@@ -39,45 +31,77 @@ class Score extends Model
...
-        return $this->ended_at !== null;
-    }
-
-    public function complete(array $params): void
-    {
-        $this->fill($params);
+        $score = new static([
+            'beatmap_id' => $params['beatmap_id'],
+            'ruleset_id' => $params['ruleset_id'],
+            'user_id' => $params['user_id'],
```
notbakaneko, line 45: need to handle missing field or set default

notbakaneko (changes requested):

app/Models/Solo/Score.php
notbakaneko: I guess they're in the ScoreToken 🤔

notbakaneko (approved):

nanaya:
@peppy please make sure to adjust the tables (migration entry, manual data migration etc) before deploying this.

peppy:
yep, saw this

PR #8124, opened 2021-09-16
Fix ChannelTest namespace

notbakaneko:


tests/Models/Chat/ChannelTest.php
```diff
@@ -3,7 +3,7 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
 // See the LICENCE file in the repository root for full licence text.
 
-namespace Tests\Models;
+namespace Tests\Models\Chat;
```
nanaya, line 6: can use addition of [this `SlevomatCodingStandard.Files.TypeNameMatchesFileName` sniff](https://github.com/slevomat/coding-standard#slevomatcodingstandardfilestypenamematchesfilename) (or [this](https://github.com/suin/phpcs-psr4-sniff) for config-free but more deps)

```xml
    <rule ref="SlevomatCodingStandard.Files.TypeNameMatchesFileName">
        <exclude-pattern>database/</exclude-pattern>
        <properties>
            <property name="rootNamespaces" type="array">
                <element key="app" value="App"/>
                <element key="tests" value="Tests"/>
            </property>
        </properties>
    </rule>
```

nanaya (changes requested):

nanaya (approved):

PR #8125, opened 2021-09-16
Add footnote support for wiki markdown

### LINKED ISSUE #7316: Superscript support for osu! wiki (footnotes)
the title says it all, but in case it does not: we'd *love* support of superscript formatting (ᵗʰᵃᵗ ᶠˡᵒᵃᵗᶦⁿᵍ ᵗᵉˣᵗ ᵗʰᶦⁿᵍ ᵖᵉᵒᵖˡᵉ ᵘˢᵉ ᶠᵒʳ ᶠᵒᵒᵗⁿᵒᵗᵉˢ) in Markdown for osu! wiki to have it look like Wikipedia ([quick example: search for `[1]` here](https://en.wikipedia.org/wiki/Map_symbol#Representing_spatial_phenomena)).

could be anything really, but I've seen this in one of Markdown parsers:

```markdown
blah blah^[1] blah^[anything, not only numbers] ...

...

[1]: http://bl.ah?
[anything, not only numbers]: https://bl.ah/blah
```

gagahpangeran:
Depends on:
- [ ] #7980 
- [ ] #8087 

- closes #7316

Now without any custom renderers and with simple approach.

## How to use

Basic syntax:

```markdown
Footnote[^ref]

[^ref]: Reference here
```

And will automatically assign number and render something like this:

Footnote<sup><a href="#">1</a></sup>

1. Reference here <a href="#">↑</a>

[1] link will target to the reference list item and ↑ link will go back to [1].

### Some caveats

**Be careful with exclamation**

Valid:

```markdown
osu! [^ref]

[^ref]: https://osu.ppy.sh
```

Invalid:

```markdown
osu![^ref]

[^ref]: https://osu.ppy.sh
```

**Be careful with duplicate reference key**

Valid:

```markdown
First[^ref]

Second[^ref]

[^ref]: Valid shared reference
```

Invalid:

```markdown
First[^ref]

Second[^ref]

[^ref]: This become invalid
[^ref]: Valid shared reference
```

## Demo

*Use [this branch](https://github.com/ppy/osu-wiki/compare/master...gagahpangeran:footnote-test?expand=1) for this example*

https://user-images.githubusercontent.com/4964985/133560550-23024f0f-7c6a-42d7-a6ae-52c1a19fdb40.mp4


resources/assets/lib/core/wiki/reference-link-tooltip.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export default class ReferenceLinkTooltip {
+  constructor() {
+    $(document).on('mouseover', '.js-reference-link', this.showTooltip);
+  }
+
+  private createTooltip = (element: HTMLElement, content: HTMLElement) => {
+    $(element).qtip({
+      content: {
+        text: content,
```
nanaya, line 12: this only looks correct by accident as `--paragraph-space` isn't set on any of the parents.

nanaya (changes requested):

tests/Libraries/Markdown/html_markdown_examples/footnote_invalid_duplicate.html
```diff
@@ -0,0 +1,32 @@
...
+</p>
+
+<p class="osu-md__paragraph">
+Second
+<sup id="osu-fnref-test__2">
+  <a class="osu-md__link osu-md__link--footnote-ref js-reference-link" href="#osu-fn-test" role="doc-noteref">1</a>
+</sup>
+</p>
+
+<li class="osu-md__list-item osu-md__list-item--footnote" id="osu-fn-test" role="doc-endnote">
```
nanaya, line 15: this extension sure likes producing weird output. This invalid list item aside, it also accepts `[^1]: > aaa` and produces blockquote. Similarly with `[^1]: 1. aaa` :neutral_face:

gagahpangeran: Just do some *moderation* from wiki-side to not using weird syntax inside footnote list :upside_down_face: 

I think they only need to use emphasis and link for footnote list in wiki, cc: @TicClick for confirmation.

TicClick: I think so, yeah. maybe some other basic formatting like bold/underlined, but that's going to be very rare. typical footnote usage is link references or remarks like the one under ¹ at https://osu.ppy.sh/wiki/VCC#criteria

PR #8128, opened 2021-09-17
Fix oauth error message

nanaya:
Original renderer is already doing the right thing™


notbakaneko (changes requested):
need to render correct page for `oauth/authorize` failures

notbakaneko (approved):

PR #8098, opened 2021-09-08
Index Featured Artist tracks to ElasticSearch

nanaya:
No frontend in this PR so here's how to query

```php
(new App\Libraries\Search\ArtistTrackSearch(App\Libraries\Search\ArtistTrackSearchParamsFromRequest::fromArray(['query'=>'hello'])))->records()
```

Instead of `ArtistTrackSearchParamsRequest`, I made `ArtistTrackSearchParamsFromRequest` accepting array, returning `ArtistTrackSearchParams` (and the other way, exporting to generate request query).

PR #8129, opened 2021-09-17
Add Featured Artist tracks search

nanaya:
...I just noticed there's no auth check in place. Is it needed here?

Demo: https://osunext.dev.myconan.net/beatmaps/artists/tracks

Depends on:
- [ ] #8098
- [ ] #8090


peppy:
Fine to have no auth requirement here, I think. We can adjust if it turns out to be a performance / other issue.

gagahpangeran:
Just checking the demo and looks like it's overflow :thinking: 

Screenshot from 2021-09-21 11-46-39

Gamingzone123:
Demo link throws 502
image

PR #8130, opened 2021-09-17

peppy:


peppy:
Note that I'm pushing a heap of new files, including languages we don't list yet. The reason the languages are not listed is due to incomplete translation status, but I don't think there's reason to not sync all files for future usage. Just removing the amount of manual steps in this process.

PR #8143, opened 2021-09-22
Use correct transaction id for logging paypal discrepencies

notbakaneko:


app/Libraries/Payments/PaypalPaymentProcessor.php
```diff
@@ -132,14 +137,20 @@ public function validateTransaction()
         $this->validatePendingStatus();
 
         // just check if IPN transaction id is as expected with the Paypal v2 API.
-        if ($this->getPaymentTransactionId() !== $order->getProviderReference()) {
+        $capturedId = $this->getOrder()->getProviderReference();
+        $transactionIdMatches = $this->getNotificationType() === NotificationType::REFUND
```
nanaya, line 141: Set `$transactionId` instead and then compare the value in the `if` below? (for simpler ternary)

nanaya (changes requested):

nanaya (approved):

PR #8149, opened 2021-09-26
Increase max blocked users to double the previous amount

peppy:
Received a request to increase this via email, and it does seem like it could do with a bit of a bump. I've also added a hard minimum because it is feasible that a user with no friends still wants to block at least a few users.

Previously, non-supporter could block up to 25 users, supporter up to 50 (based on friend count div 10). Numbers are now double that, with a minimum of 5 entries.

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -1325,7 +1325,7 @@ class User extends Model implements AfterCommit, AuthenticatableContract, HasLoc
 
     public function maxBlocks()
     {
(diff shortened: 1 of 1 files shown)


nanaya:
That's an interesting story but `maxFriends` isn't affected by current friend count (and thus nor is `maxBlocks`).

peppy:
Right, that sounds more sane. Will remove the `max` portion then as that is not required.

nanaya (approved):

PR #8150, opened 2021-09-27
Allow chat bots to send messages to users regardless of block or friend settings

### LINKED ISSUE #8147: Some chat bots (using API) need ability to send messages to users regardless of settings
As reported by @pishifat, some bots rely on being able to contact users for administrative purposes, regardless of the user's "Don't allow non-friend PMs" setting.

For any bot in the "Chat Bots" group (generally recognised as legitimate bots), this setting should be bypassed when sending chat messages.

notbakaneko:
closes #8147

### DIFF
--- a/app/Libraries/OsuAuthorize.php
+++ b/app/Libraries/OsuAuthorize.php
@@ -895,7 +895,7 @@ class OsuAuthorize
             $this->ensureHasPlayed($user);
         }
 
-        if ($user->isModerator()) {
+        if ($user->isModerator() || $user->isBot()) {
             return 'ok';
         }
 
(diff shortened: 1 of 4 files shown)


app/Libraries/OsuAuthorize.php
```diff
@@ -895,7 +895,7 @@ public function checkChatPmStart(?User $user, User $target): string
             $this->ensureHasPlayed($user);
         }
 
-        if ($user->isModerator()) {
+        if ($user->isModerator() || $user->isBot()) {
```
nanaya, line 898: why is this one missing `isAdmin` compared to the one in `App\Models\Chat\Channel#isVisibleFor` :thinking: 

is it time for a new `isThing` check

nanaya (changes requested):

app/Libraries/OsuAuthorize.php
notbakaneko: because it's `OsuAuthorize` 👀

nanaya (approved):

PR #8169, opened 2021-10-01
Remove underline in beatmap discussion history thumbnail

### LINKED ISSUE #7782: Text decoration on image links in discussion history
Not a big problem really, just kinda odd to have that short pink line on hovers, which comes from the anchors containing the images, which doesnt match the image size 
grafik
grafik

gagahpangeran:
Closes #7782

Another solution to fix this:
- Create `u-link-plain` and use it for the thumbnail link, or
- Change `beatmap-discussion-timestamp__icons` to flex box, so the link box has same size with thumbnail image.


resources/views/beatmap_discussion_posts/_item.blade.php
```diff
@@ -11,7 +11,7 @@
     <div class="beatmap-discussion beatmap-discussion--single beatmapset-activities__post-grow">
         <div class="beatmap-discussion-timestamp__icons-container">
             <div class="beatmap-discussion-timestamp__icons">
-                <a href="{{ route('beatmapsets.discussion', $beatmapset) }}#/{{ $post->beatmap_discussion_id }}">
+                <a class="beatmap-discussion-timestamp__link-plain" href="{{ route('beatmapsets.discussion', $beatmapset) }}#/{{ $post->beatmap_discussion_id }}">
```
nanaya, line 14: there's `link link--no-underline`

nanaya (changes requested):

resources/views/beatmap_discussion_posts/_item.blade.php
gagahpangeran: Ah, that's what I need.

nanaya (approved):

Issue #5529, opened 2020-01-21
Suggestions for wiki page headers


cl8n:
- the article title should be displayed in the article itself. right now, the section headers are larger than the title at the top, and the title feels a little disconnected from the article below
  - the subtitle and title don't need to be written in the yellow part anymore, if that happens. I think the yellow text should just be "article", similar to how a news post says "post" there
- breadcrumbs are missing grandparents, it only ever shows index > parent > page (see [Beatmap Veto](https://osu.ppy.sh/help/wiki/People/The_Team/Beatmap_Nominators/Beatmap_Veto) for example)
- h1 and h2 should probably have a bottom border to make sections more clear. plus I think this is just common styling for a wiki that's proven to work well. [example with border and some padding](https://user-images.githubusercontent.com/9954349/72808812-359ed880-3c0f-11ea-944c-cf06104040bc.png)

peppy:
Can agree with all of these, though we'll probably want to limit the depth of pages to not have runaway breadcrumbs. May be best to split these out into separate issues for clarity?

PR #8171, opened 2021-10-01
Add bottom line for wiki and news header

gagahpangeran:
Part of #5529

Just small changes compare to other wiki PR.

After asking in `osu-wiki` discord channel, looks like this is an okay changes for both news and wiki. Only applied for h1 and h2 like in markdown github.


nanaya:
@arflyte can you check this?

PR #8173, opened 2021-10-01
Remove store delayed shipping notice

notbakaneko:
Physical items are delegated elsewhere now, so this doesn't so anything (aside from the condition also being wrong now)


resources/views/store/checkout/show.blade.php
```diff
@@ -70,10 +70,6 @@
             <div class="store-page store-page--footer">
                 <h1 class="store-text store-text--title">Select Payment Method</h1>
 
-                @if ($checkout->isShippingDelayed() && $order->requiresShipping())
-                    @include('store._shipping_delay_warning')
```
nanaya, line 74: remove the view as well?

nanaya (changes requested):

nanaya (approved):

PR #8176, opened 2021-10-03

peppy:


resources/lang/vi/common.php
```diff
@@ -73,11 +74,12 @@
         'minute_short_unit' => 'phút',
         'minutes' => ':count phút',
         'months' => ':count tháng',
-        'notifications' => '',
-        'plus_others' => '',
+        'notifications' => ':count_delimited thông báo|:count_delimited thông báo',
+        'plus_others' => '+ :count_delimited khác!+ :count_delimited khác!',
```
nanaya, line 78: missing separator (`|`). I've updated it in crowdin so maybe run another update.

nanaya (changes requested):

resources/lang/vi/common.php
nanaya: actually Vietnamese don't even need this as there's no change for plural things :thinking:

peppy: Does that actually cause a runtime issue?

nanaya: the extra versions are ignored. Although without the separator it'll just look weird...

nanaya (approved):

PR #8183, opened 2021-10-04
Add development deploy footer

### LINKED ISSUE #8054: Implement dev.ppy.sh footer
We have the following patch applied on dev.ppy.sh that would be great to have upstreamed for future kubernetes dev.ppy.sh deployment, probably with its dedicated env var as a toggle.

```patch
diff --git a/resources/views/master.blade.php b/resources/views/master.blade.php
index 7cffa04f5..a2f75a888 100644
--- a/resources/views/master.blade.php
+++ b/resources/views/master.blade.php
@@ -131,6 +131,11 @@
         @include('layout._loading_overlay')
         @include('layout.popup-container')
 
+        <!-- stolen from smoogi https://github.com/ppy/osu-web/commit/095b34a8644dbe81630ba94877e82404b82223cb -->
+        <div class="not-official-footer" style="width: 100%; height: 50px; background: #ff0000; position: fixed; bottom: 0; left: 0; text-align: center; color: #fff; line-height: 50px;">
+          <b>This is a development instance of the <a href="https://osu.ppy.sh" style="color: #fff; text-decoration: underline;">osu! website</a>. Please do not login with your osu! credentials.</b>
+        </div>
+
         <script id="json-route-section" type="application/json">
             {!! json_encode($currentRoute) !!}
         </script>
```

I am not a regular contributor to this project, so I'm leaving this up for contribution :)

gagahpangeran:
Closes #8054

Not sure the text should be localisable or not (currently it's not) :thinking:


peppy:
Important to note that we need to update the production env to set this to `FALSE` before deploying.

resources/assets/less/bem/not-official-footer.less
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+.not-official-footer {
+  width: 100%;
+  padding: 10px;
+  background: @red;
```
nanaya, line 7: old colour variables shouldn't be used anymore

resources/assets/less/bem/not-official-footer.less
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+.not-official-footer {
```
nanaya, line 4: what does this even mean. Just call it `development-deploy-footer`

resources/views/master.blade.php
```diff
@@ -111,6 +111,12 @@ class="js-permanent-fixed-footer
                 data-sync-height-target="permanent-fixed-footer"
             >
                 @yield('permanent-fixed-footer')
+
+                @if (config('osu.is_development_deploy'))
+                    <div class="not-official-footer">
+                        <strong>This is a development instance of the <a href="https://osu.ppy.sh" class="not-official-footer__link">osu! website</a>. Please do not login with your osu! credentials.</strong>
```
nanaya, line 117: put font weight in css instead

resources/assets/less/bem/not-official-footer.less
```diff
@@ -0,0 +1,18 @@
...
+  background: @red;
+  text-align: center;
+
+  @media @mobile {
+    font-size: @font-size--title-small;
+  }
+
+  &__link {
+    .link-white();
+    text-decoration: underline;
```
nanaya, line 16: Maybe use `--hsl-c2` color instead. I don't think there's anywhere else where link is underlined

config/osu.php
```diff
@@ -111,6 +111,7 @@
     'git-sha' => presence(env('GIT_SHA'))
         ?? (file_exists(__DIR__.'/../version') ? trim(file_get_contents(__DIR__.'/../version')) : null)
         ?? 'unknown-version',
+    'is_development_deploy' => env('IS_DEVELOPMENT_DEPLOY', true),
```
nanaya, line 114: use `get_bool`

nanaya (changes requested):

resources/assets/less/bem/not-official-footer.less
gagahpangeran: I just copy this class name from the issue. Will change.

resources/assets/less/bem/not-official-footer.less
gagahpangeran: image

This is the combination of `--hsl-red-3` background colour and `--hsl-c2` link colour. Is it okay?

nanaya: yes

nanaya (approved):

nanaya:
@peppy please remember to set the env variable

PR #8186, opened 2021-10-06
Handle legal wiki page API

gagahpangeran:
See https://github.com/ppy/osu/issues/14974 for related issue.


notbakaneko:
`/api/v2/legal` isn't needed for the same reasons stated here https://github.com/ppy/osu-web/pull/8186#pullrequestreview-777276697

PR #8189, opened 2021-10-06
improve FR translation for tournament period

Biokax:
In order to fix a sentence that sound really weird in french

current
image

PR
image


Biokax:
> translations are periodically pulled from crowdin. unless you fix it there, at https://crowdin.com/translate/osu-web/45/en-fr?filter=basic&value=0#1205, this change will be overwritten

I updated crowdin for this, thanks

nanaya:
This will be included as part of next translation update.

PR #8282, opened 2021-10-31
Convert `Comment` factory to class

cl8n:
and `NewsPost` so it works with any commentable type.


database/factories/CommentFactory.php
```diff
@@ -3,25 +3,35 @@
...
+
+class CommentFactory extends Factory
+{
+    protected $model = Comment::class;
+
+    public function definition(): array
+    {
+        return [
+            'commentable_type' => fn () => $this->faker->randomElement(Comment::COMMENTABLES),
+            'created_at' => fn () => $this->faker->dateTime(),
```
nanaya, line 21: nothing else use random date for `created_at` and `updated_at` and I don't think it's needed either (also for NewsPost)

tests/Browser/SanityTest.php
```diff
@@ -172,9 +172,10 @@ private static function createScaffolding()
         self::$scaffolding['group'] = factory(\App\Models\Group::class)->create();
 
         // factory for comments
-        self::$scaffolding['comment'] = factory(\App\Models\Comment::class)->create([
-            'user_id' => self::$scaffolding['user']->user_id,
+        self::$scaffolding['comment'] = \App\Models\Comment::factory()->create([
             'commentable_id' => self::$scaffolding['build'],
+            'commentable_type' => 'build',
+            'user_id' => self::$scaffolding['user']->user_id,
```
nanaya, line 178: `->user_id` not needed

nanaya (changes requested):

nanaya (approved):

PR #8283, opened 2021-10-31
Allow mapper and moderators to pin comments on beatmapsets

### LINKED ISSUE #6574: The ability for mapper to pin a comment.
Basically letting the mapper pin a comment like how Ephemeral did on the Aspire final results.
chrome_raliqIgI3k

It would be nice as some people tend to read comments and ignore map descriptions. Sometimes a mapper may want to link alternate mp3s, oszs, or even needs to provide a update on the status of the mapset.

Right now the only real way to kinda pin a comment is to make sure its the most upvoted one which doesn't ensure people will even see it as some will inevitably sort by "Recent" instead of "Top".

chrome_I4NRCcydTW

vs

chrome_1sG2UGR0cY

cl8n:
resolves #6574, if it ends up being wanted.

- [x] depends on #8282


peppy:
At *very least* this would need to be limited to one pinned comment total.

resources/assets/lib/stores/commentable-meta-store.ts
```diff
@@ -16,7 +16,7 @@ export default class CommentableMetaStore {
     this.meta.clear();
   }
 
-  get(type: string, id: number) {
+  get(type: string, id: number): CommentableMetaJson | undefined {
```
cl8n, line 19: the implicit return type is the intersection of the `CommentableMetaJson` union, `{ title: string; }`. I don't really understand why...

nanaya:
> At _very least_ this would need to be limited to one pinned comment total.

is this one pinned comment total for the entire map? If moderator pinned a comment, can the mapper pin another comment?

peppy:
> > At _very least_ this would need to be limited to one pinned comment total.
> 
> is this one pinned comment total for the entire map? If moderator pinned a comment, can the mapper pin another comment?

I think that scenario is quite unlikely, so whatever is easiest. Optimally the limit of one would be for the mapper specifically - moderators should be able to pin any number of comments alongside it. But it's not a dealbreaker is this isn't possible.

nanaya:
what's the rule on pinning? anyone can pin and unpin whatever they want even if not their own comment (or the comments they pinned)? Or should it be limited to own comments for non-moderators? I noticed there's no `pinned_by` data at the moment.

peppy:
Own comment only for mappers.

cl8n:
> Own comment only for mappers.

seems like a large part of the reason ppl want this is to pin other comments though. i've seen the sentiment of https://github.com/ppy/osu-web/issues/6574#issuecomment-955632130 shared around, and it's what prompted this PR too

for permissions i kept it simple because we don't track who pinned comments atm, like nanaya said. this branch currently allows mapper/moderators pin 1 comment per mapset overall, and any top-level comment can be pinned. moderators are included only cuz i thought it'd be weird if they couldn't unpin something inappropriate or w/e. and to be clear, admin can still pin as many as they want

peppy:
Pinning other user comments is out of scope. It's going to lead to contention when we unpin stupid shit (or delete it due to moderation). I'd rather not go there.

Kagetsu24:
>  It's going to lead to contention when we unpin stupid shit (or delete it due to moderation). I'd rather not go there.

is pinning "stupid shit" really that bad though? i think pinning comments isn't really that big of a deal, if people want to pin nonsense, you should let them imo. at the end of the day it's their map and if they want to look stupid it's up to them. it's not like people can't achieve similar results by using the description or upvotes, for example.

about deleting, yeah... i agree, but if a pinned comment were to be removed due to moderation, it would probably not be allowed even if wasn't pinned in the first place.

i think that restricting it to your own comments only really limits the social part of this feature and would also make it (presumably) more difficult to implement just for the restriction to be potentially dropped in the future? 

in any case, personally i would rather have the feature implemented with restrictions if that means moving forward, but i don't think it makes much sense.

peppy:
I've already state my opinion here, but welcome those from other team members (cc @Ephemeralis). Just want to avoid more contention between users expecting "freedom of speech", where that usually means shitting on the system.

Ephemeralis:
Can wholeheartedly agree that adding contention between moderators and mappers over dumb user comment pins is 100% guaranteed to happen, which is counter to the intention of keeping beatmap discussions as productive spaces. This *will* be abused extensively in opposition to that fact if opened up to users.

Allowing moderators to do it should be sufficient, and having mappers contact them to pin in circumstances where they feel appropriate, especially with a fair handful of them being closely accessible to the mapping scene at large these days.

NevarrenF:
> This will be abused extensively in opposition to that fact if opened up to mappers.

Well considering pinned comments are effectively just like a maps description shouldn't we look at them in a similar way. Currently mappers can put whatever random meme garbage they want in the description of a map and they would get either deleted or edited by moderators so the same would be said for pinned comments? Having to go find a moderator to post something on your map seems kinda unneeded since at the end of the day it is the mappers mapset so if they want to be dumb then whatever. The only thing I would see a mapper requiring to go to a moderator for is pinning someone else's comment but honestly I can't imagine that happening often.

peppy:
@cl8n Rather than having this issue come to a standstill, can we go with the minimal permissions and expand at a later stage? As proposed by @Ephemeralis as a first step.

cl8n:
did https://github.com/ppy/osu-web/pull/8283#issuecomment-970074669 , & it's still one pin total for mapper+moderator

the test is kinda silly, idk how else to make an effective test for a permission set though
it was easy to write but now is hard to read...

nanaya (approved):

PR #8286, opened 2021-11-02
Handle replay download with missing beatmap or user

nanaya:
`headerChunk` throwing error is too late to abort once streamDownload runs.

### DIFF
--- a/app/Http/Controllers/ScoresController.php
+++ b/app/Http/Controllers/ScoresController.php
@@ -42,19 +42,25 @@ class ScoresController extends Controller
 
         try {
             $filename = "replay-{$mode}_{$score->beatmap_id}_{$score->getKey()}.osr";
-            $content = $replayFile->get();
-
-            return response()->streamDownload(function () use ($replayFile, $content) {
-                echo $replayFile->headerChunk();
-                echo pack('i', strlen($content));
-                echo $content;
-                echo $replayFile->endChunk();
-            }, $filename, ['Content-Type' => 'application/x-osu-replay']);
+            $body = $replayFile->get();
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/ScoresController.php
```diff
@@ -42,19 +42,25 @@ public function download($mode, $id)
...
+
+        $file = [
+            $replayFile->headerChunk(),
+            pack('i', strlen($body)),
+            $body,
+            $replayFile->endChunk(),
+        ];
+
+        return response()->streamDownload(function () use ($file) {
+            foreach ($file as $f) {
```
notbakaneko, line 60: `foreach ($chunks as ...`? `foreach ($file` sounds weird

nanaya:
asset build fail on pr which doesn't even touch it is quite something :ok_hand:

app/Http/Controllers/ScoresController.php
```diff
@@ -42,19 +42,19 @@ public function download($mode, $id)
...
             log_error($e);
             abort(404);
         }
+
+        $file = $replayFile->headerChunk()
+            .pack('i', strlen($body))
+            .$body
+            .$replayFile->endChunk();
+
+        return response()->streamDownload(fn () => echo $file, $filename, ['Content-Type' => 'application/x-osu-replay']);
```
notbakaneko, line 57: can't `fn () => echo ...`

besides, which are the params for `echo`, the closure and `streamDownload`

notbakaneko (changes requested):

notbakaneko (approved):

Issue #8196, opened 2021-10-07
Display Player's Guest Difficulties in the User page


nicholauscjj:
With the addition of setting "Difficulty Owner", or better known as "Guest Difficulty", it would be better if all their guest diff to be displayed under their respective mapper. This will not entirely cover the pre-moddingv2.0 maps, so the mapper has to do it themselves (unfortunately).

![image](https://user-images.githubusercontent.com/36277652/136359770-d913cad4-c43b-4a6d-980a-808bce7d4d9b.png)
Figure 1. Current state of Beatmaps part of the user page

![image](https://user-images.githubusercontent.com/36277652/136359942-f4a01eb7-b51e-48ee-8af2-ac05eaf5dfd6.png)
Figure 2. Proposed state of Beatmaps part of the user page. Notice that there's "Guest Difficulty" subpart, although I only rename it from existing layout. It can be inserted anywhere, though.

cl8n:
you can use `artisan tinker` to manually update models (e.g. `Beatmap::find($beatmapId)->update(['user_id' => $userId]);`) or make new models using the provided factories

can also use the menu provided on the beatmap page to update creators

Diduhless:
What if the host of a beatmap
set falsely adds someone as a gder into their set? This should require some sort of confirmation system to prevent this new feature from being abused

I also believe there should be an option where you can remove yourself as the guest difficulty creator from the set (only if it's not ranked) in case you don't want to display some of your pending guest difficulties for the same reason as you being able to delete your graveyarded maps in your profile

cl8n:
@Diduhless open a separate issue for each of those, they're both problems with the existing gd settings and not this proposed display. can be worked on independently too

Milkitic:
Here is my proposal: For accuracy of statistics I think we need a solution for old ranked beatmaps which hasn't been marked as guest difficulty. Anyway it's another topic

Hiviexd:
should be solved by addressing https://github.com/ppy/osu-web/issues/8243 alongside the current issue

nanaya:
Instead of having all this "guest difficulty" badge thing, can it be just a new section which lists all non-wip/graveyard guest diff maps?

peppy:
I think that's fine as a starting point. Should probably be beneath "Ranked Beatmaps" and called "Ranked Beatmaps (with guest difficulty)"

Hiviexd:
[...]

PR #8287, opened 2021-11-03
added a 'guest apperance notice' to user profiles #8196'

TalhaAbid:
just need some pointers on the visual look now since rn its very basic, also my built in formatter auto formatted but otherwise this is a working feature and tested.

### DIFF
--- a/resources/assets/lib/beatmapset-panel.tsx
+++ b/resources/assets/lib/beatmapset-panel.tsx
@@ -1,94 +1,128 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
 // See the LICENCE file in the repository root for full licence text.
 
-import BeatmapsPopup from 'beatmapset-panel/beatmaps-popup';
-import { CircularProgress } from 'circular-progress';
-import Img2x from 'img2x';
-import BeatmapJson from 'interfaces/beatmap-json';
-import BeatmapsetExtendedJson from 'interfaces/beatmapset-extended-json';
-import BeatmapsetJson, { BeatmapsetStatus } from 'interfaces/beatmapset-json';
-import GameMode from 'interfaces/game-mode';
-import { route } from 'laroute';
-import { sum, values } from 'lodash';
-import { action, computed, makeObservable, observable } from 'mobx';
-import { observer } from 'mobx-react';
-import core from 'osu-core-singleton';
-import * as React from 'react';
-import { Transition } from 'react-transition-group';
-import StringWithComponent from 'string-with-component';
-import TimeWithTooltip from 'time-with-tooltip';
-import { UserLink } from 'user-link';
-import { getArtist, getDiffColour, getTitle, group as groupBeatmaps } from 'utils/beatmap-helper';
-import { showVisual, toggleFavourite } from 'utils/beatmapset-helper';
-import { classWithModifiers } from 'utils/css';
-import { formatNumberSuffixed } from 'utils/html';
-import { beatmapsetDownloadDirect } from 'utils/url';
+import BeatmapsPopup from "beatmapset-panel/beatmaps-popup";
+import { CircularProgress } from "circular-progress";
+import Img2x from "img2x";
+import BeatmapJson from "interfaces/beatmap-json";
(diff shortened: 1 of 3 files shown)


resources/assets/lib/profile-page/beatmapsets.tsx
```diff
@@ -28,49 +28,58 @@ type Props = {
...
               <div
                 key={beatmapset.id}
-                className='osu-layout__col osu-layout__col--sm-6'
+                className="osu-layout__col osu-layout__col--sm-6"
               >
-                <BeatmapsetPanel beatmapset={observable(beatmapset)} />
+                <BeatmapsetPanel
+                  beatmapset={observable(beatmapset)}
+                  currentUserName={this.props.user.username}
+                  section={section}
```
cl8n, line 65: the logic of deciding to display guest badge should be local to the profile page, so figure it out here instead of in `BeatmapsetPanel` and then only pass a flag into `BeatmapsetPanel`

resources/assets/lib/beatmapset-panel.tsx
```diff
@@ -410,151 +467,176 @@ export default class BeatmapsetPanel extends React.Component<Props> {
...
+          <div className="u-ellipsis-overflow">
             <StringWithComponent
-              mappings={{ mapper: <MapperLink beatmapset={this.props.beatmapset} /> }}
-              pattern={osu.trans('beatmapsets.show.details.mapped_by')}
+              mappings={{
+                mapper: <MapperLink beatmapset={this.props.beatmapset} />,
+              }}
+              pattern={osu.trans("beatmapsets.show.details.mapped_by")}
             />
+            {this.props.beatmapset.creator !== this.props.currentUserName &&
```
cl8n, line 506: compare the profile page user's ID to the beatmapset creator ID. usernames can change and this isn't displaying what you want

... or, you probably won't have to make this check at all if you send the mapsets to browsers in a different field than the other mapsets (which you should)

resources/assets/lib/beatmapset-panel.tsx
```diff
@@ -410,151 +467,176 @@ export default class BeatmapsetPanel extends React.Component<Props> {
...
+            this.props.section !== "favouriteBeatmapsets" ? (
+              <div>Guest Apperance</div>
```
cl8n, line 508: needs to be localisable, look into usage of `osu.trans` around this

resources/assets/lib/beatmapset-panel.tsx
```diff
@@ -410,151 +467,176 @@ export default class BeatmapsetPanel extends React.Component<Props> {
...
             <StringWithComponent
-              mappings={{ mapper: <MapperLink beatmapset={this.props.beatmapset} /> }}
-              pattern={osu.trans('beatmapsets.show.details.mapped_by')}
+              mappings={{
+                mapper: <MapperLink beatmapset={this.props.beatmapset} />,
+              }}
+              pattern={osu.trans("beatmapsets.show.details.mapped_by")}
             />
+            {this.props.beatmapset.creator !== this.props.currentUserName &&
+            this.props.section !== "favouriteBeatmapsets" ? (
```
cl8n, line 507: `false` is a special value to React so u can do `condition && element` instead of `condition ? element : null`

resources/assets/lib/beatmapset-panel.tsx
cl8n, line 508: also I can't imagine this looks great with no extra spacing or anything :thinking: 

even then it's probably not prominent enough of a label here. I'd either wait for flyte's response to the issue or go with the badge thing similar to FA and NSFW (seems like you were going to do that but scrapped it? you left `GuestMapperBadge` unused)

cl8n (changes requested):
the server needs to send the guest diff mapsets to the browser in the first place... start at `UsersController::show()` and follow how data gets from there to these react components. you'll probably need to touch `User` and `UserTransformer`

use the formatting standards for the project (given by phpcs and eslint) and undo the formatting changes you committed

resources/assets/lib/beatmapset-panel.tsx
TalhaAbid: Just wanted a review on the logic first I’ll make the changes then work on visual

peppy:
Converting this to a draft PR as that seems to be what it is.

PR #8291, opened 2021-11-04
Throw validation error when failing to save factory user

notbakaneko:
Sometimes the user generated by the factory fails to save but we can't see what the exact error is.


nanaya (approved):
I hope this is temporary :thinking:

notbakaneko:
It should be if we find which part is causing the validation to fail 👀

PR #8295, opened 2021-11-05
Typescript UserPageEditor component

nanaya:
Less trivial than the other one.

- [x] depends on #8294 (for no particular reason apart of maybe fewer potential conflicts)


resources/assets/lib/interfaces/user-json.ts
```diff
@@ -26,6 +26,10 @@ export default interface UserJson {
   is_online: boolean;
   is_supporter: boolean;
   last_visit: string | null;
+  page?: {
```
notbakaneko, line 29: this should probably be in `UserExtendedJson` but `CurrentUser` currently extends it and doesn't have `page` 🤔

notbakaneko (approved):

resources/assets/lib/interfaces/user-json.ts
nanaya: it's a `UserCompact` transformer include so it's part of `UserJson` (and it's not default include in `User` transformer either).

notbakaneko: it's only in the profile page which is what `UserExtendedJson` is (was) supposed to be for.

PR #8296, opened 2021-11-05
Typescript Loading Overlay utility

nanaya:
This also reduces lint warning of dealing with `any` :thinking:


notbakaneko (changes requested):
`user-page-editor` needs updating

notbakaneko (approved):

PR #8298, opened 2021-11-05
Typescript for pp display component

nanaya:
- [x] depends on #8295


resources/assets/lib/interfaces/user-statistics-json.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
 // See the LICENCE file in the repository root for full licence text.
 
-import GameMode from 'interfaces/game-mode';
+import GameMode from './game-mode';
```
notbakaneko, line 4: can revert this change

notbakaneko (changes requested):

resources/assets/lib/interfaces/user-statistics-json.ts
nanaya: should it though :thinking:

notbakaneko (approved):

PR #8305, opened 2021-11-08
Typescript profile page section of user profile

nanaya:
Finally something that's used in `main` :neutral_face: 

- [x] depends on #8295


resources/assets/lib/profile-page/beatmapsets.tsx
```diff
@@ -9,7 +9,7 @@ import * as React from 'react';
 import ShowMoreLink from 'show-more-link';
 import ExtraHeader from './extra-header';
 import { BeatmapsetSection } from './extra-page-props';
-import ExtraPageProps from './extra-page-props';
+import ExtraPageProps, { ProfilePagePaginationData } from './extra-page-props';
```
notbakaneko, line 12: these should be together 🤔

notbakaneko (changes requested):

resources/assets/lib/profile-page/beatmapsets.tsx
notbakaneko: as it unselects the other line 🤨

notbakaneko (approved):

PR #8307, opened 2021-11-08
Typescript top scores section of profile page

nanaya:
`SectionMap` seems a bit shady :thinking: 

- [x] depends on #8305


resources/assets/lib/profile-page/top-scores.tsx
```diff
@@ -0,0 +1,108 @@
...
+              className='wiki-notice__markdown-inline-content'
+              dangerouslySetInnerHTML={{ __html: this.props.scoresNotice }}
+            />
+          </div>
+        )}
+
+        {sectionMaps.map((section) => (
+          <div key={section.key}>
+            <h3 className='title title--page-extra-small'>
+              {osu.trans('users.show.extra.top_ranks.best.title')}
```
notbakaneko, line 62: Needs some variable.
Also the param is `firsts` but the translation key is `first` (and other general mismatches) 🤔

notbakaneko (changes requested):

resources/assets/lib/profile-page/top-scores.tsx
nanaya: fun ಠ_ಠ

I'm not fixing that now

notbakaneko (commented):
conflicts 💥

notbakaneko (approved):

PR #8310, opened 2021-11-09
Refactor Friend Button

nanaya:
Now with more mobx and less unused options.

Also I'm placing it inside `components` because there are a bit too many various things in the top `lib` directory.


notbakaneko (commented):
conflicts 💥

notbakaneko (changes requested):
I wonder if `friendButton:refresh` event can the removed now 🤔

notbakaneko (approved):
`UserCardBrick` can probably be updated to not need `friendButton:refresh` later

PR #8313, opened 2021-11-10
Preserve whitespace in difficulty names

### LINKED ISSUE #8288: Difficulty names with multiple consecutive spaces get compressed
Related: #5942

This is visible on the recently ranked beatmap [Toby Fox - Dialtone](https://osu.ppy.sh/beatmapsets/1584926#osu/3237398). The second difficulty is named

    HARD                                     !!

with 37 spaces after the word `HARD`. It appears correctly in osu!stable and lazer, but displays as simply "HARD !!" in various locations on the website.

image
image
image

## Possible fix

Setting `white-space: pre-wrap;` causes the difficulty name to appear correctly, at least on the main difficulty page (i haven't tried other locations). `white-space: pre` also works, but doesn't introduce line breaks for really really long diff names.

Might be kind of a pain because difficulty names can appear on the website in lots of contexts though.

image

cl8n:
resolves #8288

feels messy to have different solutions in different places but I'm not sure how to do it better. this PR is like #6092 but for diff names.

### DIFF
--- a/resources/assets/coffee/_classes/tooltip-default.coffee
+++ b/resources/assets/coffee/_classes/tooltip-default.coffee
@@ -44,6 +44,8 @@ class window.TooltipDefault
       classes += ' tooltip-default--fixed'
     if el.dataset.tooltipModifiers?
       classes += " tooltip-default--#{el.dataset.tooltipModifiers}"
+    if el.classList.contains 'u-ellipsis-pre-overflow'
+      classes += ' tooltip-default--pre'
 
     options =
       overwrite: false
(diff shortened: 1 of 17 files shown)


nanaya:
Is this actually needed? This looks like adding complexity just for some pointless use case.

peppy:
I think multiple spaces should really be blocked at a point of beatmap submission, maybe something to consider in the future. I also agree that not showing them seems fine.

cl8n:
guess I jumped the gun on this one, thought it'd warrant the same treatment as legacy usernames even if it shouldn't be allowed anymore (which I agree with as well). the map the issue used as an example is already ranked, fwiw

cl8n:
this can be closed along with #8288, then

PR #8316, opened 2021-11-10
Fix discussion review selector menu popup direction

notbakaneko:
Just recycles the existing clipping logic for now.
Doesn't test for clipping in the other direction since that isn't needed.....yet 👀


nanaya:
I wonder if the default position should just be automatically determined by whether it's on left or right half of the screen instead :thinking:

notbakaneko:
all the menus hang to the left even on the left side of the screen (makes sense since they're usually after wide buttons?) except this one, so 🤷

nanaya (approved):

PR #8319, opened 2021-11-10
Convert user profile rank component to typescript

notbakaneko:
Needed some type signatures for something else - might as well convert it now instead of adding more declarations to `coffee-modules`


resources/assets/lib/profile-page/rank.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  const variantTooltip: string[] = [];
+  for (const variant of stats.variants ?? []) {
+    const variantRank = variant[key];
+    if (variantRank == null) continue;
+
+    const name = osu.trans(`beatmaps.variant.${variant.mode}.${variant.variant}`);
+    const value = `#${osu.formatNumber(variantRank)}`;
+
+    variantTooltip.push(`<div>${name} ${value}</div>`);
```
nanaya, line 26: the original has `:` separator

nanaya (changes requested):

nanaya (approved):

PR #8325, opened 2021-11-11
Handle per-mode rank history offset

nanaya:
Matches recent score system update.


peppy (approved):

notbakaneko (changes requested):
`$this->mode` is `int` but the functions that they're passed into expect the `string` version of the mode 🤔

notbakaneko:
or is it 🤔

notbakaneko (approved):
missed the attribute getter

notbakaneko (approved):

PR #8331, opened 2021-11-12
Reorganise traits for model

nanaya:
I was looking into other stuff and noticed the traits are in multiple places for reason not quite clear to me. One of them is just blending together with models in `App\Models`.

I figured traits specific for model can be put in their own directory (and maybe moved out whenever it's to be used by anything else).

Also `EsIndexableModel` has been renamed to `BaseDbIndexable` because `Wiki\Page` is a model but it uses `EsIndexable` as it's not backed by database.


app/Models/Traits/Macroable.php
```diff
@@ -3,13 +3,13 @@
...
-namespace App\Traits;
+namespace App\Models\Traits;
 
 use App\Scopes\MacroableModelScope;
 
-trait MacroableModel
+trait Macroable
 {
-    public static function bootMacroableModel()
+    public static function bootMacroable()
```
notbakaneko, line 12: apparently this can be just `booted()` 🤔

notbakaneko (approved):

PR #8334, opened 2021-11-15
Convert RankChart to typescript

notbakaneko:
~~also removed the chart update in `componentDidUpdate` since the data shouldn't have changed without a full remount in this case (I think).~~
After some fiddling, updated props can just be passed directly to the chart, so I put it back in. (And `loadRankChart` can be put in `autorun` if the `props` become `observable`)


nanaya:
but if the component is static, is it still react component or just convenient html renderer :thinking:

nanaya:
and conflicts :eyes:

resources/assets/lib/profile-page/rank-chart.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+function formatX(d: number) {
+  return d === 0 ? osu.trans('common.time.now') : osu.transChoice('common.time.days_ago', -d);
+}
+
+function formatY(d: number) {
+  return `<strong>${osu.trans('users.show.rank.global_simple')}</strong> #${osu.formatNumber(-d)}`;
+}
+
+export default class RankChart extends React.Component<Props> {
```
nanaya, line 45: can use a warning somewhere that updates to props are ignored

nanaya (changes requested):

nanaya (approved):

PR #8336, opened 2021-11-15
Add web support for multiplayer queue modes

smoogipoo:
This is an MVP implementation from my side. Requesting help from others on the osu-web team for tests / migrations / anything that I've missed.

- The `owner_id` column is used to give visibility for who created the playlist item. It can be any user id from the room, and doesn't relate to the host.
- The `queue_mode` column is a new room property.

Database changes should be something like the following:
```sql
ALTER TABLE `multiplayer_rooms`
  ADD COLUMN `queue_mode` enum('host_only', 'free_for_all', 'fair_rotate') NOT NULL DEFAULT 'host_only'
  AFTER `category`;

ALTER TABLE `multiplayer_playlist_items`
  ADD COLUMN `owner_id` int unsigned NOT NULL DEFAULT '0'
  AFTER `id`;
```


peppy (approved):

database/migrations/2021_11_17_074221_add_owner_id_to_multiplayer_playlist_items.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('multiplayer_playlist_items', function (Blueprint $table) {
+            $table->unsignedInteger('owner_id')->after('room_id')->default(0);
```
nanaya, line 20: a nullable would make more sense than a magic default

tests/Controllers/Multiplayer/RoomsControllerTest.php
```diff
@@ -109,6 +110,28 @@ public function testStoreRealtimeByType()
         $this->assertSame($type, $room->type);
     }
 
+    public function testStoreRealtimeByQueueMode()
+    {
+        $token = Token::factory()->create(['scopes' => ['*']]);
+        $queue_mode = array_rand_val(Room::REALTIME_QUEUE_MODES);
```
nanaya, line 116: variables are camelCased

tests/Controllers/Multiplayer/RoomsControllerTest.php
```diff
@@ -88,6 +88,7 @@ public function testStoreRealtime()
         $this->assertNotNull($room);
         $this->assertTrue($room->isRealtime());
         $this->assertSame($type, $room->type);
+        $this->assertSame($token->user->user_id, $room->playlist()->first()->owner_id);
```
nanaya, line 91: `getKey()` is preferred to get model's primary key

```suggestion
        $this->assertSame($token->user->getKey(), $room->playlist()->first()->owner_id);
```

database/migrations/2021_11_17_074221_add_owner_id_to_multiplayer_playlist_items.php
smoogipoo: How do I make it not nullable with no default? I was getting test errors: https://github.com/ppy/osu-web/runs/4236637200?check_suite_focus=true

nanaya: add a user in the factory (`database/factories/Multiplayer/PlaylistItemFactory.php`)

peppy: Not sure how I missed this one 💦

peppy:
Looks like this should be deployable without a new osu! update, right? If so we can probably try and get this merged today (pending finalisation of enum naming).

smoogipoo:
Deployable without an osu! update, yes.

nanaya (approved):

PR #8346, opened 2021-11-18
simplify beatmap pack installation instructions

Samaoo:
- simplify installation instructions 
- rephrase a few things

### DIFF
--- a/resources/lang/en/beatmappacks.php
+++ b/resources/lang/en/beatmappacks.php
@@ -12,7 +12,7 @@ return [
         'blurb' => [
             'important' => 'READ THIS BEFORE DOWNLOADING',
             'instruction' => [
-                '_' => 'Installation: Once a pack has been downloaded, extract the contents of the pack into your osu! Songs directory.',
+                '_' => 'Installation: Once a pack has been downloaded, extract the contents of the pack into your osu! Songs directory and osu! will do the rest.',
             ],
             'note' => [
                 '_' => 'Also note that it is highly recommended to :scary, since older maps are generally of much lower quality than more recent maps.',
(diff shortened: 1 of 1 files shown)


resources/lang/en/beatmappacks.php
```diff
@@ -12,14 +12,10 @@
         'blurb' => [
             'important' => 'READ THIS BEFORE DOWNLOADING',
             'instruction' => [
-                '_' => "Installation: Once a pack has been downloaded, extract the .rar into your osu! Songs directory.
-                    All songs are still .zip'd and/or .osz'd inside the pack, so osu! will need to extract the beatmaps itself the next time you go into Play mode.
-                    :scary extract the zip's/osz's yourself,
-                    or the beatmaps will display incorrectly in osu! and will not function properly.",
-                'scary' => 'Do NOT',
```
peppy, line 19: I don't think this part should change. The process is still the same.

resources/lang/en/beatmappacks.php
```diff
@@ -12,14 +12,10 @@
...
+                '_' => 'Installation: Once a pack has been downloaded, extract the contents of the pack into your osu! Songs directory.',
             ],
             'note' => [
-                '_' => 'Also note that it is highly recommended to :scary, since the oldest maps are of much lower quality than most recent maps.',
```
peppy, line 22: Not against this line change though.

peppy (changes requested):

resources/lang/en/beatmappacks.php
Samaoo: I think all the user should know is that they should extract the beatmappack to the osu songs folder. If you don't want users to extract osz files manually, I think the best way to do that is just to not tell them that they can be extracted manually, the way it's currently phrased just overcomplicates things in my opinion.

peppy: Will wait for more opinions on this

TicClick: agree with Samaoo (not to prevent people from extracting `.osz`s, but because that's irrelevant here)

peppy (approved):
Let's go with this.

PR #8347, opened 2021-11-19
Support svg in wiki

nanaya:
To be used by ppy/osu-wiki#6390


peppy:
That said, if we start to support this it's also going to need to be supported in lazer... which I.. can't guarantee.

peppy:
Let's close this one for now. We can revisit if we ever decide to support svg in the game client.

PR #8349, opened 2021-11-19
Show only public and pm channels in web chat

notbakaneko:
The other channel types apparently aren't a part of web chat.
- ~~`nonPmChannels` is now `publicChannels` because that's what it is (and another grouping is coming)~~
- channel list now renders channels according to their grouping (`GROUP` channels don't exist yet, so using default public channel sort)
- `channelList` moved to `ChatStateStore` because it's only used by that
- remove last usages of `inject`


resources/assets/lib/stores/channel-store.ts
```diff
@@ -18,7 +18,7 @@ import Message from 'models/chat/message';
 import core from 'osu-core-singleton';
 import UserStore from './user-store';
 
-const skippedChannelTypes = new Set<ChannelType>(['MULTIPLAYER', 'TEMPORARY']);
+const visibleChannelTypes = new Set<ChannelType>(['PM', 'PUBLIC']);
```
peppy, line 21: I don't think we use it yet, but `GROUP` type should probably also be displayed.

nanaya:
Conflicts. And `publicChannels` now also includes `group` channels...? Maybe the previous `nonPmChannels` name is still better?

nanaya:
conflicts :eyes:

notbakaneko:
probably should make an alias for `Record<string, never>` at some point 👀

nanaya (changes requested):

nanaya (approved):

PR #8358, opened 2021-11-22
Clean up User Json interface

nanaya:
The `ProfilePageUserJson` will be eventually moved somewhere else...


notbakaneko (changes requested):
might as well add the properties from `CurrentUser` to `UserJson` as well and `Require<Pick<` them 🤔 
(and also rename that `User::defaultJson` method)

notbakaneko (approved):

PR #8365, opened 2021-11-24
Remove global current user from typescript

nanaya:
Added `core.currentUserOrFail` for places which depends on its existence.

- [x] depends on #8358


notbakaneko (commented):
conflicts 💥

notbakaneko (approved):
seems fine 🤔

PR #8359, opened 2021-11-22
Typescript Cover Uploader

nanaya:
And then I still split the branches ( ﾟ ヮﾟ)

depends on:
- [x] #8356
- [x] #8358

PR #8366, opened 2021-11-24
Typescript Cover Selector

nanaya:
Now with mobx for state.

Only six more remaining after this one ( ﾟ ヮﾟ)

- [x] depends on #8359


notbakaneko (changes requested):

notbakaneko (commented):
and conflicts now ( ﾟ ヮﾟ)

notbakaneko (approved):

PR #8367, opened 2021-11-25
Documentation fixes

tybug:
Mostly involves making attributes optional, but a few other miscellaneous changes in here too. Explanations for the potentially questionable changes follow.

### beatmapset discussion changes

I don't 100% know what I'm talking about here because I don't do modding, so these might warrant actual api behavior changes instead of documentation changes, not sure.

* `beatmapset_discussion.beatmap_id` optional
  * https://osu.ppy.sh/beatmapsets/1113445/discussion#/2725011 has a null `beatmap_id`, among many others
* `starting_post` optional
  * calling https://osu.ppy.sh/docs/index.html#get-beatmapset-discussion-votes returns discussions with no `starting_post` attribute at all (not even a null one). Seems to be the only endpoint with this issue, everywhere else properly returns a nonnull `starting_post`

### other changes

* `beatmap.bpm` optional
  * some beatmaps like https://osu.ppy.sh/beatmapsets/1090330#osu/2279323 return null bpm
* `user_statistics.playtime` optional
  * some users like https://osu.ppy.sh/users/17906393 return null play time

A rather large caveat that I haven't tested these changes myself (didn't want to set up docker just for some docs fixes), but should :tm: be fine since it's just markdown.


peppy (dismissed):

resources/views/docs/_structures/kudosu_history.md
```diff
@@ -3,7 +3,7 @@
 Field      | Type       | Description
 -----------|------------|-----------------------------
 id         | number     | |
-action     | string     | Either `give`, `reset`, or `revoke`.
+action     | string     | Either `vote.give`, `vote.reset`, or `vote.revoke`.
```
nanaya, line 6: the non prefixed ones still exist

resources/views/docs/_structures/user_statistics.md
```diff
@@ -64,7 +64,7 @@ level.current             | number                      | Current level.
 level.progress            | number                      | Progress to next level.
 maximum_combo             | number                      | Highest maximum combo.
 play_count                | number                      | Number of maps played.
-play_time                 | number                      | Cumulative time played.
+play_time                 | number?                     | Cumulative time played.
```
nanaya, line 67: ...that's actually a bug. There's a missing default value assignment.

nanaya (changes requested):

resources/views/docs/_structures/kudosu_history.md
tybug: have changed it to reference all 6 possibilities in that case, reads a bit wonky but not sure there's a better wording

nanaya (approved):

PR #8378, opened 2021-11-26
Typescript Line Chart

nanaya:
That was _fun_. No idea if works with the Historical component when it's typed later ( ﾟ ヮﾟ)a

Versioning for d3 typing is a bit weird (put nicely).


resources/assets/lib/charts/line-chart.ts
```diff
@@ -0,0 +1,378 @@
...
+    marginTop: options.marginTop ?? 20,
+    modifiers: options.modifiers,
+    scaleY: options.scaleY ?? scaleLinear(),
+    ticksX: options.ticksX,
+    tickValuesX: options.tickValuesX,
+  };
+}
+
+export default class LineChart<X extends Date | number> {
+  data: { x: X; y: number}[] = [];
```
notbakaneko, line 87:
```suggestion
  data: { x: X; y: number }[] = [];
```

notbakaneko (changes requested):

notbakaneko (approved):

PR #8385, opened 2021-11-29
Typescript profile page header

nanaya:
This may need further clean ups once the dependencies are merged.

- [x] #8378
- [x] #8376
- [x] #8372
- [x] #8368
- [x] #8366
- [x] #8359
- [x] #8356


resources/assets/lib/profile-page/header.tsx
```diff
@@ -0,0 +1,202 @@
...
+    this.debouncedCoverSet(this.props.user.cover.url);
+  };
+
+  private coverSet(url: string | null) {
+    if (this.isCoverUpdating) return;
+
+    this.coverUrl = url;
+  }
+
+  private readonly onCloseEdit = (e?: JQuery.Event | JQuery.ClickEvent) => {
```
notbakaneko, line 144: `e` shouldn't be undefined anymore?

resources/assets/lib/profile-page/header.tsx
```diff
@@ -0,0 +1,202 @@
...
+    if (!this.editing) return;
+
+    if (e != null) {
+      if (e.button !== 0) return;
```
notbakaneko, line 148: doesn't this mean the `key:esc` won't work anyway 🤔

notbakaneko (changes requested):

notbakaneko (approved):

notbakaneko (approved):

PR #8386, opened 2021-11-29
Fix displayed channel messages being reset when opening new PM

notbakaneko:
Opening chat with `?sendto=$userId` causes other tabs with chat open to have that channel's display reset.

Also fixes another issue with the user preloading order when rejoining channels (causes other chat tabs to die because details in the socket event are missing)

### DIFF
--- a/app/Http/Controllers/ChatController.php
+++ b/app/Http/Controllers/ChatController.php
@@ -31,9 +31,8 @@ class ChatController extends Controller
         $user = auth()->user();
         Chat::ack($user);
 
-        $targetId = get_params(request()->all(), null, ['sendto:int'], ['null_missing' => true])['sendto'];
         // rejoin any existing channel first, otherwise it'll be missing from the preload later.
-        $targetUser = $targetId !== null ? User::lookup($targetId, 'id') : null;
+        $targetUser = User::lookup(get_int(request('sendto')), 'id');
         if ($targetUser !== null) {
             $channel = Channel::findPM($targetUser, $user);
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/ChatController.php
```diff
@@ -32,16 +31,26 @@ public function index()
         $user = auth()->user();
         Chat::ack($user);
 
+        $targetId = get_params(request()->all(), null, ['sendto:int'], ['null_missing' => true])['sendto'];
+        // rejoin any existing channel first, otherwise it'll be missing from the preload later.
+        $targetUser = $targetId !== null ? User::lookup($targetId, 'id') : null;
+        if ($targetUser !== null) {
+            $channel = Channel::findPM($targetUser, $user);
+            $channel?->addUser($user);
```
nanaya, line 39: looks like the channel loaded inside `$userChannel` need to call `resetMemoized` as well in the `addUser` function.

And maybe move the `$this->resetMemoized()` (a few lines up) outside.

https://github.com/ppy/osu-web/blob/756e37cc4ea7b78fc92160f1cffc4043943e2ef5/app/Models/Chat/Channel.php#L421

nanaya (changes requested):

nanaya (approved):
The other chat tab dying on rejoin is caused by `users()` returning incorrect data when `PRELOADED_USERS_KEY` is incomplete/outdated and will need to be fixed separately. (the fix in this pr only fixes one specific usage)

PR #8389, opened 2021-11-30
Typescript bottomPage function

nanaya:
Another thing being used by profile page and not currently typed.

(and then I learned I messed up the `osu_common.coffee` history)


notbakaneko (commented):
conflicts 💥

notbakaneko (approved):

PR #8391, opened 2021-11-30
Increment user count cache on registration

### LINKED ISSUE #8380: Registered players counter on the home page is not up to date
Seems to have been frozen at 15,210,892 since late 2019 (according to the [wayback machine](https://web.archive.org/web/20191103215834/http://osu.ppy.sh/home)).

nanaya:
Not sure about the accuracy esp with user restrictions and old user renames and deletes (and maybe other) but at least it increments on registration? :shrug: 

Resolves #8380?


app/Models/Count.php
```diff
@@ -11,23 +11,24 @@
...
-    protected $table = 'osu_counts';
-    protected $primaryKey = 'name';
-
+    public $incrementing = false;
     public $timestamps = false;
 
+    protected $primaryKey = 'name';
+    protected $table = 'osu_counts';
+
     public static function currentRankStart(string $mode): int
```
notbakaneko, line 20: wrong return type

notbakaneko (changes requested):

notbakaneko (approved):

PR #8392, opened 2021-11-30
API for batched beatmap lookup

nanaya:
~~Preloading beatmaps for `playmodes` is also possible but assuming the `difficulty_names` can be parsed with some split by comma and then grab the number right to rightmost `@`, it's one less query.~~ Nope, nontrivial changes to test needed.

Similar to `/users` but `/beatmaps`.

- [x] ~~is the beatmapset state check needed here? (the `active` flag). There's currently none for existing lookup either though.~~ yes, and now filtered out


app/Http/Controllers/BeatmapsController.php
```diff
@@ -10,22 +10,144 @@
...
+     *     {
+     *       "id": 2,
+     *       "other": "attributes..."
+     *     }
+     *   ]
+     * }
+     */
+    public function index()
+    {
+        $ids = array_slice(get_arr(request('ids'), 'get_int'), 0, 50);
```
notbakaneko, line 59: needs some handling when `ids` isn't array and `get_arr` returns null

notbakaneko (changes requested):

notbakaneko (approved):

PR #8394, opened 2021-12-01
Add support for sending public channel messages through LIO

### LINKED ISSUE #8393: Web chat not showing new messages automatically
The page is not showing new messages, making it out of sync with the in-game chat (even when switching channels or pm)
image

Refreshing the page makes them appear, but it still doesn't update automatically

notbakaneko:
Quick update splicing in support for public channels into LIO

will fix #8393 after other relevant parts are updated to send to `_lio`

### DIFF
--- a/app/Http/Controllers/LegacyInterOpController.php
+++ b/app/Http/Controllers/LegacyInterOpController.php
@@ -196,7 +196,8 @@ class LegacyInterOpController extends Controller
      *   - too many messages has been sent by the sender
      *
      * @bodyParam messages[<id>][sender_id] integer required id of user sending the message
-     * @bodyParam messages[<id>][target_id] integer required id of user receiving the message. Must not be restricted
+     * @bodyParam messages[<id>][target_id] integer required id of user receiving the message if `type` is `pm`; channel, otherwise. Must not be restricted
+     * @bodyParam messages[<id>][type] string required type of the target of the message, `pm` or `public`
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/LegacyInterOpController.php
```diff
@@ -222,35 +225,50 @@ public function userBatchSendMessage()
             $messageParams = get_params($messageParams, null, [
                 'sender_id:int',
                 'target_id:int',
+                'type:string',
```
nanaya, line 228: can use docs update

app/Http/Controllers/LegacyInterOpController.php
```diff
@@ -222,35 +225,50 @@ public function userBatchSendMessage()
...
+            ->keyBy('channel_id');
+
         foreach ($params as $id => $messageParams) {
             try {
                 if (!is_array($messageParams)) {
                     abort(422);
                 }
 
-                if (!isset($messageParams['sender_id']) || !isset($messageParams['target_id'])) {
+                if (!isset($messageParams['type']) || !isset($messageParams['sender_id']) || !isset($messageParams['target_id'])) {
```
nanaya, line 271: the `pm` type default is still not set here

nanaya (changes requested):

nanaya (approved):

PR #8400, opened 2021-12-02
Support sending message to any channel type via LIO

notbakaneko:
In hindsight, `type` is just `pm` or `not pm` 👀

### DIFF
--- a/app/Http/Controllers/LegacyInterOpController.php
+++ b/app/Http/Controllers/LegacyInterOpController.php
@@ -233,18 +233,16 @@ class LegacyInterOpController extends Controller
 
             // TODO: default to null later
             $messageParams['type'] ??= 'pm';
-            // ignore if type missing (and return error?)
-            if (in_array($messageParams['type'], ['pm', 'public'], true)) {
-                if (isset($messageParams['sender_id'])) {
-                    $userIds->add($messageParams['sender_id']);
-                }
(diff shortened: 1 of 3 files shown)


app/Libraries/OsuAuthorize.php
```diff
@@ -32,6 +32,7 @@
 class OsuAuthorize
 {
     const REQUEST_ATTRIBUTE_KEY = 'auth_map';
+    const REQUEST_LIO_KEY = 'is_lio_request';
```
nanaya, line 35: it hasn't been "legacy" anymore and the newer endpoints are in `InterOp` (also the old controller should eventually be split and moved there)

`lio` name should only remain in the url and nowhere else (or maybe even add a copy route and eventually rename everything).

app/Http/Controllers/LegacyInterOpController.php
```diff
@@ -233,18 +233,16 @@ public function userBatchSendMessage()
 
             // TODO: default to null later
             $messageParams['type'] ??= 'pm';
-            // ignore if type missing (and return error?)
-            if (in_array($messageParams['type'], ['pm', 'public'], true)) {
```
nanaya, line 237: another docs update needed

nanaya (changes requested):

nanaya (approved):

PR #8405, opened 2021-12-02
Add gameplay_order column to playlist items

smoogipoo:
Updated by osu-server-spectator to contain the order in which the items will be played in the room. Changes based on the room's queue mode.


smoogipoo:
I just realised that there's already a `playlist_order` column :sweat_smile: Closing for now.

PR #8408, opened 2021-12-03
Expose playlist item's playlist_order in JSON responses

smoogipoo:
I think we probably want to upgrade this column to a mediumint / int, but that can be done in the future.


peppy:
missed the comma again

peppy:
> I think we probably want to upgrade this column to a mediumint / int, but that can be done in the future.

`smallint` will allow some 32768 items, is there something i'm missing with how this is going to be used? if we get that high in a room, there are going to be other issues that arise first.

smoogipoo:
The case I foresee is a large round robin room which lasts for a very long time, with many people joining. If I'm storing the gameplay order of all items _historically_, then I'm worried that we can overflow this value.

So one of two things has to happen: either the limit should be increased, or we cannot rely on this for historical ordering and need to use the `updated_at` column as I initially proposed.

peppy:
I don't think that's important to think about right now. Like, before that, the room is going to get so large in request size that things will fall over. Far before we get anywhere near the 30k limit.

peppy (approved):

PR #8409, opened 2021-12-03
Fix looking up beatmap with multiple parameters

nanaya:
It should stop on not null, not the other way... :raised_eyebrow: 

Closes https://github.com/ppy/osu/issues/15912.

### DIFF
--- a/tests/Controllers/BeatmapsControllerTest.php
+++ b/tests/Controllers/BeatmapsControllerTest.php
@@ -71,6 +71,25 @@ class BeatmapsControllerTest extends TestCase
     }
 
     /**
+     * Make sure the lookup stops when finding beatmap from one of the parameters
+     */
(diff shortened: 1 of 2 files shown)


peppy (approved):

tests/Controllers/BeatmapsControllerTest.php
```diff
@@ -70,6 +70,25 @@ public function testLookupForApi(string $key, callable $valueFn): void
             ->assertJsonPath('id', $beatmap->getKey());
     }
 
+    /**
+     * Make sure the lookup stops when finding beatmap from one of the parameters
```
notbakaneko, line 74: shouldn't it be `doesn't stop until`?

nanaya: ...same thing? :thinking:

PR #8413, opened 2021-12-03
Show reason when sending to channel is disabled

notbakaneko:
Makes the display a bit more compact (and possibly moved to be rendered with the input box).
The error message shows up anyway in the http response if the UI state is ignored.

- [x] #8412 but not really


nanaya (changes requested):

nanaya:
conflicts :pensive:

nanaya (approved):

PR #8415, opened 2021-12-05
Implement a GraphQL API

LavaDesu:
This PR adds support for a GraphQL API in addition to the REST API.

The GraphQL endpoint is currently under `/graphql`, might have to be changed to `/api/graphql` later
A web interface to test queries (with docs and autocompletion) can be accessed under `/graphql-playground`

Some things that might need discussion:
- Currently, the minimal schema is modelled closely after the internal property names, and is cased with camelCase. Is this fine or should it more closely resemble the REST API?
- There's a custom query complexity calculator in `app\Providers\GraphQLServiceProvider`, which assigns cost 2 to object-types (fields that return an object) and cost 1 to leaf-types (fields that return a simple value). It is quite complicated and might not be necessary, as we can just assign cost 1 to everything instead. I figured that by assigning cost 2 to object-types, it would make more sense due to the (usually) extra database query.
- I've tried to make most of the implementation exist under the `app\GraphQL` namespace, as some of them don't make much sense in normal HTTP context (e.g. the middlewares under `app\GraphQL\Middleware` which wraps around existing middleware and returns a spec-compliant response on error)
- There's some questionable class names, such as the exceptions in `app\GraphQL\Exceptions`
- Should mutations exist? It would be similar to HTTP PATCH/DELETE/etc.., where data can be modified by a client

Aside from the schema, most of the implementation code should be ready to be reviewed (excluding some TODO notes).

TODO:

- [ ] Tests
- [ ] Documentation (for both clients and in code)
- [ ] The rest of the schema


peppy:
> * Currently, the minimal schema is modelled closely after the internal property names, and is cased with camelCase. Is this fine or should it more closely resemble the REST API?

One would hope they match the already documented models on https://docs.ppy.sh

notbakaneko:
If Graphql is to even be considered, it shouldn't result in us having to maintain another version of the API; the schema should match the existing one as much as possible.

peppy:
Can also agree with that.

nanaya (changes requested):
I've only done quick check but 

- having to manually input authorization header in playground isn't very convenient
  - and if the api is to be used by web frontend, it'll need to allow cookies authentication as well
- pagination shouldn't go beyond `config('osu.pagination.max_count')` (or probably less, and preferably with extra cost as offset goes higher)
  - either that or somehow add the custom cursor pagination support in (the one by lighthouse is fake; it still uses offset pagination internally)
- `orderBy` should only be on indexed columns
- beatmap.best_scores is a bit more complicated than simple hasMany (check `macroForListing` in `app/Models/Score/Best/Model.php`)
- I'm not sure which part failed but I got this error when querying best scores: `"debugMessage": "Expected a value of type \"Int\" but received: 2215896885"` (something somewhere can't handle large int?)
  - looks like graphql just doesn't support ints larger than 32-bit because javascript and so they should be casted to string instead (or add extension but probably safer to just make them string)

LavaDesu:
@nanaya Thanks for the review!
> * having to manually input authorization header in playground isn't very convenient
>   
>   * and if the api is to be used by web frontend, it'll need to allow cookies authentication as well

I've been playing around with session auth, and it seems to just work™ by adding some of the web middleware in. However, that probably isn't a very good idea as I already spot some problems like CSRF, updating user online status, user verification, etc..
I'm thinking that we can just have a new route for that purpose? Maybe `/graphql-web` or something, then set different middlewares for them.

> * pagination shouldn't go beyond `config('osu.pagination.max_count')` (or probably less, and preferably with extra cost as offset goes higher)
>   
>   * either that or somehow add the custom cursor pagination support in (the one by lighthouse is fake; it still uses offset pagination internally)

I'm aware of the fake cursor. I had looked into it and there doesn't seem to be any easy way to extend it other than creating a new directive (`@connection` would look nice). Might not be too hard to implement since laravel has support for it already.

> * `orderBy` should only be on indexed columns

Yep. I was using it to test how the ordering stuff works. Will remove it.
> * beatmap.best_scores is a bit more complicated than simple hasMany (check `macroForListing` in `app/Models/Score/Best/Model.php`)

Will look into it.

> * I'm not sure which part failed but I got this error when querying best scores: `"debugMessage": "Expected a value of type \"Int\" but received: 2215896885"` (something somewhere can't handle large int?)
>   
>   * looks like graphql just doesn't support ints larger than 32-bit because javascript and so they should be casted to string instead (or add extension but probably safer to just make them string)

:thinking: Maybe that's a score that's larger than signed 32-bit
I think we can have a new `Long` scalar that's just a renamed `String`.

nanaya:
That was an interesting experiment but still needs quite a bit more work.

PR #8417, opened 2021-12-06
Remove default scribe generated authentication docs

notbakaneko:
and just replace the template with our own since their text is embedded in the `Writer` itself.


resources/views/docs/info.blade.php
```diff
@@ -45,389 +33,6 @@
...
-## Managing OAuth applications
-
-Your [account settings]({{ route('account.edit').'#oauth' }}) page will show your registered OAuth applications, and all the OAuth applications you have granted permissions to.
-
-### Reset Client Secret
-
-You can generate a new `Client Secret` by choosing to "Reset client secret", however, this will disable all access tokens issued for the application.
-
 
 # Changelog
```
nanaya, line 37: I think moving this changelog section to before endpoint would make more sense.

nanaya (changes requested):

nanaya (approved):

Issue #6404, opened 2020-08-01
Create an "announcement" style message in new chat system


cl8n:
the old forum pms have been in an end-of-life stage for a while (as said in multiple issues here, and on discord), but I don't see anything tracking ideas for what could properly replace it. #5774 requests for "group chat" to be exposed, but that isn't really in the same spirit as the mail-like forum pms. #6359 asks for api access to chat to get some workaround going due to forum pms having awkward bugs sometimes (#6334), but it would be a nonissue if there were a new alternative to forum pms. I and probably anyone else trying to do important communication with osu! projects would benefit a lot from having something similar to the old forum pms, but without all of the bloat that phpbb brought along with it.

I feel like a good enough "replacement" could be closely related to chat, so these are what I see as necessary features that *aren't* available via chat already:

- multi-paragraph & formatted messages
- "to" and "bcc" fields with same function as email, specifically *not* just opening a group chat
- subject line that can be used in notifications and emails

and these are more features I use in forum pms that are nice but probably not necessary:

- read receipts
- message icons
- drafts
- basic message sorting or "folders"

peppy:
Going to milestone this so we can finally turn off the old system.

Route forward is adjusting chat to have a special message type which gets more prominence and can be seen as "important" compared to normal chat.

huuishuu:
Apologies for bumping, but it's been since August 2020 when the last mention on this happened, so I'd like to chime in on why this would be beneficial for stuff like Project Loved. As part of Project Loved, with the way things are - old forum PMs are just about the only way we have of mass-messaging multiple people when it comes to:
- letting them know their map is being put up for vote
- letting them know that changes are necessary (as moddingv2 doesn't automatically subscribe the mapper to make them receive notifications).

Old forum PMs are basically a necessity for us, yet it's in such a state of disrepair that it's not just buggy but it also no longer sends an on-website notification that someone has received a message, making the entire process of communicating with mappers a real drag.

Is there any update on this or anything to expect anytime soon?

peppy:
No updates on this just yet.

If you don't mind using the API, "mass messaging" via the new chat system is already feasible.

cl8n:
if nobody else starts on this by June or so (a guess when I'll have my other PRs + WIP stuff in good states) I'd like to try it-- for Project Loved we rly need this functionality soon, I'm sure other projects would prefer it over raw chat messages, and it also looks like a blocker for turning off the old site.

[...]

PR #8418, opened 2021-12-06
Add support for announcement channels in chat

notbakaneko:
Basic support for creating the relevant channels for #6404
Uses a new `announce` group that is intended to be less privileged so that its groups permissions can be granted when using OAuth.
Also requires channels to have `name` and `description` set instead of magically filled by mysql.
Sends a different notification type so they can't be simply ignored by turning off chat notifications since the announcements are supposed to be important.

Does not include the actual presentation and formatting of the messages, those will be separate from this.

- [x] #8349
- [x] #8413 (for displaying reason later)
- [x] #8447


app/Libraries/Chat.php
```diff
@@ -28,6 +30,48 @@ public static function ack(User $user)
...
+
+        $users = $users->push($sender)->uniqueStrict('user_id');
+
+        $channel = (new Channel())->getConnection()->transaction(function () use ($sender, $params, $users) {
+            $channel = Channel::createAnnouncement($users, $params['channel']);
+
+            foreach ($users as $user) {
+                event(new ChatChannelEvent($channel, $user, 'join'));
+            }
+
```
nanaya, line 64: also applies for sending pm but failing sending the actual message result in empty channel on the receiver side...

nanaya (changes requested):
Handling replies will be fun with this. Also rip people who accidentally closes the channel.

app/Libraries/Chat.php
notbakaneko: ironically because the `event`s are being called in the function instead of after the transaction completes...

nanaya:
Conflicts...

database/migrations/2021_11_26_090035_add_announce_type_to_channels.php
```diff
@@ -0,0 +1,47 @@
+<?php
```
nanaya, line 1: redate the file

nanaya (changes requested):

nanaya:
Conflicts 🤔

app/Libraries/Chat.php
```diff
@@ -28,6 +29,43 @@ public static function ack(User $user)
...
+            'channel:any',
+            'message:string',
+            'target_ids:int[]',
+        ], ['null_missing' => true]);
+
+        if (!isset($params['target_ids'])) {
+            throw new InvariantException('missing target_ids parameter');
+        }
+
+        if (!isset($rawParams['channel'])) {
```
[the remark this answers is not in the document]
nanaya: should be `$params` (otherwise error 500 if not an array)

database/migrations/2022_01_20_090035_add_announce_type_to_channels.php
```diff
@@ -0,0 +1,47 @@
+<?php
```
nanaya, line 1: another redate 👀

nanaya (changes requested):

nanaya (approved):

PR #8425, opened 2021-12-08
Per-channel scroll behaviour

notbakaneko:
Handles scroll position on a per-channel basis.
Also fixes the initial scroll not working properly when navigating to the page.

- window focus changed to an `observable`
- mark as read now only happens when scroll is at the bottom of a channel when:
  - switching away from the channel to another channel (so scrolling up and then switching will preserve the marker)
  - window comes into focus
  - sending a message (which still auto scrolls to the bottom)

More accurate handling of marking as read needs a bit more complicated tracking of the position of the unread marker which I really don't want to do here 👀


nanaya:
conflicts...

resources/assets/lib/chat/conversation-view.tsx
```diff
@@ -116,23 +127,33 @@ export default class ConversationView extends React.Component<Props> {
...
     if (this.didSwitchChannel) {
-      if (this.unreadMarkerRef.current) {
-        this.scrollToUnread();
-      } else {
-        this.scrollToBottom();
-      }
-      this.didSwitchChannel = false;
+      // This can happen after a turbolinks navigation,
+      // so we have to wait for the elements to be on the current document before scrolling.
+      core.reactTurbolinks.runAfterPageLoad(this.id, action(() => {
```
nanaya, line 135: also need listener removal for this

Looking again the `runAfterPageLoad` function should just return a disposer :thinking: then the id wouldn't be needed anymore. And looking even closer I noticed one is called synchronously and the other is not - that may cause confusing behavior difference.

resources/assets/lib/chat/conversation-view.tsx
```diff
@@ -94,6 +96,15 @@ export default class ConversationView extends React.Component<Props> {
 
     makeObservable(this);
 
+    disposeOnUnmount(
+      this,
+      observe(core.windowFocusObserver, 'hasFocus', (change) => {
```
nanaya, line 101: just a minor thing but according to the docs [it's preferred to use reaction over observe](https://mobx.js.org/intercept-and-observe.html#intercept--observe-) due to how observe works (ignores action and just run right away)

resources/assets/lib/chat/input-box.tsx
```diff
@@ -40,10 +37,17 @@ export default class InputBox extends React.Component<Props> implements Dispatch
...
 
-    dispatcher.register(this);
-
     makeObservable(this);
 
+    disposeOnUnmount(
+      this,
+      autorun(() => {
+        if (core.windowFocusObserver.hasFocus) {
+          this.focusInput();
```
nanaya, line 46: it has always been the case but I just noticed opening quick search when on chat page and then switching focus out and back in results in being able to type in chat while quick search is still open :upside_down_face:

nanaya (changes requested):
conflicts again...

resources/assets/lib/chat/conversation-view.tsx
notbakaneko: doesn't `.one()` unbind automatically after it's done? 🤔

nanaya: that's assuming the component is still mounted when it runs

notbakaneko: sounds like `runAfterPageLoad` should change; `$(document).off` in the component without `$(document).on` seems counter intuitive 🤔

resources/assets/lib/chat/input-box.tsx
```diff
@@ -73,11 +77,12 @@ export default class InputBox extends React.Component<Props> implements Dispatch
     this.focusInput();
   }
 
-  componentWillUnmount() {
-    dispatcher.unregister(this);
-  }
+  focusInput(always = true) {
+    const activeElement = document.activeElement;
+    if (!always && (activeElement instanceof HTMLInputElement || activeElement instanceof HTMLTextAreaElement)) {
```
nanaya, line 82: that's an attempt but there's nothing forcing focus on quick search input at the moment :eyes: 

(search thing -> open result in new tab with mouse (middle click etc) -> switch to the tab -> go back to chat tab)

nanaya (changes requested):

nanaya (approved):

PR #8430, opened 2021-12-09
Separate chat message component rendering

notbakaneko:
message render split off into its own component and wrapped in an error boundary...although the boundary seems not so useful since there's not much to fail when rendering the message 👀


resources/assets/lib/chat/message-group.tsx
```diff
@@ -44,46 +41,20 @@ export default class MessageGroup extends React.Component<Props, any> {
...
-                      <i className='fas fa-times'/>
-                    </div>
-                  }
-                </div>
-                { showTimestamp &&
-                  <div className='chat-message-group__message-timestamp'>{moment(message.timestamp).format('LT')}</div>
-                }
-              </div>
+              <MessageErrorBoundary key={message.uuid}>
+                <MessageItem message={message} showTimestamp={showTimestamp} />
```
nanaya, line 56: `MessageItem` doesn't throw anything and it doesn't contain anything that's obviously going to throw on error so I don't think the boundary is useful here?

Not to mention the component is also sometimes responsible rendering timestamp (though this part could be separated by pre-grouping the messages by timestamp and always render timestamp separate at the end).

resources/assets/lib/chat/message-error-boundary.tsx
```diff
@@ -0,0 +1,30 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+import { observer } from 'mobx-react';
+import * as React from 'react';
+
+interface State {
+  hasError: boolean;
+}
+
+@observer
```
nanaya, line 11: this component doesn't seem to access anything observable?

nanaya (changes requested):

resources/lang/en/chat.php
```diff
@@ -22,12 +22,18 @@
...
+
     'input' => [
         'disabled' => 'unable to send message...',
         'disconnected' => 'Disconnected',
         'placeholder' => 'type message...',
         'send' => 'Send',
     ],
+
+    'message' => [
+        'render_failed' => 'This message could not be displayed.',
```
nanaya, line 34: this isn't used anymore...

nanaya (changes requested):

nanaya (approved):

PR #8431, opened 2021-12-09
Put all javascript entrypoints in one directory

nanaya:
css file is excluded because a bit more work is needed for it.

Webpack restart is still needed when adding/removing entrypoint but at least no need to update the config anymore. And less clutter.


notbakaneko (approved):
looks ok so far 🤔

PR #8438, opened 2021-12-13
Combine score helper files

nanaya:
Having separate files seem rather pointless. Also moved definition for score statistics attribute type to score-json.


notbakaneko (changes requested):
https://github.com/ppy/osu-web/blob/513a05d50e58a8e4d7e1933aaccb02336f51ed74/resources/assets/coffee/react/beatmapset-page/scoreboard.coffee#L20 comment can use update (or just update the function 🤔 )

notbakaneko (approved):

PR #8439, opened 2021-12-13
Add support for rendering markdown in announcement messages

notbakaneko:
Rendering the message uses `dangerouslySetInnerHTML` already anyway.
Only allows basic formatting types; normal links are allowed, pretty links are intentionally disabled, i.e. `https://osu.ppy.sh` will create a link, `[https://osu.ppy.sh](https://www.youtube.com/watch?v=dQw4w9WgXcQ)` will not

- [x] #8418
- [x] #8430


resources/assets/less/bem/chat-message-item.less
```diff
@@ -38,5 +38,16 @@
       color: red;
     }
   }
+
+  // TODO: css class instead?
+  code {
+    background: @osu-colour-b4;
```
nanaya, line 44: use variable

resources/assets/lib/chat/message-error-boundary.tsx
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
```
nanaya, line 1: not used

tests/Libraries/Markdown/ChatMarkdownTest.php
```diff
@@ -0,0 +1,40 @@
...
+
+    private function loadOutputTest(string $name, string $path)
+    {
+        $mdFilePath = "{$path}/{$name}.md";
+        $htmlFilePath = "{$path}/{$name}.html";
+
+        return [
+            file_get_contents($mdFilePath),
+            file_get_contents($htmlFilePath),
+        ];
```
nanaya, line 38: looks like can use map

resources/assets/lib/models/chat/message.ts
```diff
@@ -39,6 +41,7 @@ export default class Message {
     return Object.assign(message, {
       channelId: json.channel_id,
       content: json.content,
+      contentHtml: json.content_html,
```
nanaya, line 44: not related but is there specific reason using assign instead of just usual assignment (in a `runInAction`)?

`Object.assign` isn't type checked.

nanaya (changes requested):

resources/assets/lib/models/chat/message.ts
notbakaneko: it is a mystery 🤔

notbakaneko: it only seems to avoid `message.` on every line and isn't even less lines 🤔

PR #8441, opened 2021-12-14
Add multiplayer room duration limits

### LINKED ISSUE #8426: Add check to limit playlist room duration
New options:

2 weeks
1 month (supporter only)
3 month (supporter only) <- not 100% sure about this but let's give it a try i guess

Reference: https://github.com/ppy/osu/discussions/14251

Would appreciate if this can be added as a relative priority, as the user in the linked thread would like to use it in January 2022.

notbakaneko:
The maximum duration is `3 * 31` days following the option here: https://github.com/ppy/osu/pull/16024/files#diff-d5f8d0eb0eac5cfd6d2f486d34c4168e036b83b622c7a3c5bfce5205d67bf52bR330

closes #8426


app/Models/Multiplayer/Room.php
```diff
@@ -478,6 +478,16 @@ private function assertValidStartGame()
...
 
+        // TODO: configurable duration (same for min and max durations)?
+        if (!$this->user->isSupporter() && $this->starts_at->addDays(14)->lt($this->ends_at)) {
+            throw new InvariantException(osu_trans('multiplayer.room.errors.duration_require_supporter'));
+        }
+
+        // see https://github.com/ppy/osu/pull/16024/files#diff-d5f8d0eb0eac5cfd6d2f486d34c4168e036b83b622c7a3c5bfce5205d67bf52bR327-R330
+        if ($this->starts_at->adddays(63)->lt($this->ends_at)) {
+            throw new InvariantException(osu_trans('multiplayer.room.errors.duration_too_long'));
+        }
```
nanaya, line 489: - almost none of other supporter/non-supporter checks do separate check like this at the moment (they all simply call relevant `maxThing()` function in User model)
- the message for non-supporter is kind of a lie if it's more than 3 months
- the second `addDays` has lowercase `d` for `Days`
- probably micro optimization but all of those date attribute accesses re-create the object every time

tests/Models/Multiplayer/RoomTest.php
```diff
@@ -137,4 +170,18 @@ public function testCannotStartPlayedItem()
         $this->expectException(InvariantException::class);
         (new Room())->startGame($user, $params);
     }
+
+    public function startGameDurationDataProvider()
+    {
+        return [
+            '2 weeks' => [20160, true, null],
```
nanaya, line 177: all those numbers are rather hard to read (and update). Put them in variables with original calculation instead...

tests/Models/Multiplayer/RoomTest.php
```diff
@@ -15,6 +15,39 @@
...
+            ],
+        ];
+
+        if ($errorMessageKey !== null) {
+            $this->expectException(InvariantException::class);
+            $this->expectExceptionMessage(osu_trans($errorMessageKey));
+        }
+
+        $room = (new Room())->startGame($user, $params);
+        $this->assertSame($errorMessageKey === null, $room->exists);
```
nanaya, line 48: that's an interesting combination between `$this->assertTrue($room->exists)` assertion and `if ($errorMessageKey === null) { ... }` condition check (which is the opposite of the previous check above).

At a glance it looks like it's checking `false` if `$errorMessageKey` is not null except it'll never happen (as the `startGame` call throws)

nanaya (changes requested):

app/Models/Multiplayer/Room.php
notbakaneko: `addDays` mutates the instance

nanaya: ends_at doesn't seem to be mutated anywhere though

resources/lang/en/multiplayer.php
```diff
@@ -13,6 +13,10 @@
         'player_count' => ':count_delimited player|:count_delimited players',
         'time_left' => ':time left',
 
+        'errors' => [
+            'duration_too_long' => 'Duration cannot be more than :duration days.',
```
nanaya, line 17: That's a nice error message (no `:duration` variable set)

nanaya (changes requested):

nanaya (approved):

PR #8442, opened 2021-12-14
Add pinning scores

### LINKED ISSUE #2664: Add user option to pin plays to their profile
https://www.reddit.com/r/osugame/comments/835xwo/feature_idea_being_able_to_choose_what_plays/

Has been brought up before as well, wanted to bring it to github in case you weren't aware that it could be a neat little addition to optionally show off cool scores that aren't high pp plays.

nanaya:
Reordering will come later :tm: 

No design for score-top component and scores-show page :eyes: 

- [x] depends on #8433

Resolves #2664. Note though only best play is currently available for pinning and if pinned score is replaced by another score, it'll be gone.


resources/assets/lib/profile-page/play-detail-list.tsx
```diff
@@ -24,6 +24,10 @@ const sectionMaps = {
     countKey: 'scores_first_count',
     translationKey: 'top_ranks.first',
   },
+  scoresPinned: {
+    countKey: 'scores_pinned_count',
```
notbakaneko, line 28: also need to update the count somehow after pin/unpin 🤔

notbakaneko (changes requested):
first pass...

notbakaneko:
> if pinned score is replaced by another score, it'll be gone.
what happens to the existing entry in the database?

nanaya:
> what happens to the existing entry in the database?

to be cleaned up with some scheduled job task later™

resources/assets/lib/components/score-pin.tsx
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  constructor(props: Props) {
+    super(props);
+
+    makeObservable(this);
+  }
+
+  componentWillUnmount() {
+    this.onUpdateCallbacks = null;
```
notbakaneko, line 43: the active menu key still needs to be `null`ed; `PopupMenuPersistent` unmount seems to be the appropriate place for this, although `useEffect` seems to trigger too many times for each change 🤔

resources/assets/lib/core/user/score-pins.ts
```diff
@@ -0,0 +1,71 @@
...
+import OsuCore from 'osu-core';
+
+export default class ScorePins {
+  @observable pins = new Map<string, boolean>();
+
+  constructor(private core: OsuCore) {
+    makeObservable(this);
+  }
+
+  apiPin(score: ScoreJson, toPin: boolean): JQuery.jqXHR<void> {
```
notbakaneko, line 16: return type of of the `ajax` call should be casted to be not `<any>`

notbakaneko (changes requested):
weighted score disappearing after reload but newly pinned score having weight seems strange
image

resources/assets/lib/core/user/score-pins.ts
nanaya: Doesn't seem to matter? 🤔

notbakaneko: returning `jqXHR<any>` for a function typed `jqXHR<void>` is a lint warning

notbakaneko (approved):

Loreos7:
Will it be available to translate?

peppy:
yes, please wait until we sync localisations.

PR #8444, opened 2021-12-15

peppy:


resources/lang/fi/beatmappacks.php
```diff
@@ -12,11 +12,10 @@
         'blurb' => [
             'important' => 'LUE TÄMÄ ENNEN LATAAMISTA',
             'instruction' => [
-                '_' => "Asennus: Kun paketti on ladattu, pura .rar tiedosto osu! Songs kansioon.
+                '_' => 'Asennus: Kun paketti on ladattu, pura .rar tiedosto osu! Songs kansioon.
```
peppy, line 15: We're probably going to want to clear these translations, unfortunately.. due to `:scary`.

PR #8446, opened 2021-12-15
Don't return Builder from UserChannel::last_read_id

notbakaneko:
extracted from #8418

While looking at making the chat event broadcast dispatch after the transactions are committed instead of during the transactions, came across this issue again where `last_read_id` is a query `Expression` instance when it gets read after it's updated

### DIFF
--- a/app/Models/Chat/UserChannel.php
+++ b/app/Models/Chat/UserChannel.php
@@ -9,6 +9,7 @@ use App\Libraries\Notification\BatchIdentities;
 use App\Models\User;
 use App\Models\UserNotification;
 use DB;
+use Illuminate\Database\Query\Expression;
 
 /**
  * @property Channel $channel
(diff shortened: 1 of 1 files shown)


app/Models/Chat/UserChannel.php
```diff
@@ -23,6 +23,14 @@ class UserChannel extends Model
 {
     protected $primaryKeys = ['user_id', 'channel_id'];
 
+    private ?int $lastReadIdToSet;
+
+    public function getLastReadIdAttribute($value): ?int
+    {
+        // return the value we tried to set it to, not the builder.
+        return $this->lastReadIdToSet ?? $value;
```
nanaya, line 31: only return `lastReadIdToSet` if `$value` is a ~~builder~~ query expression instead?

nanaya (changes requested):

nanaya (approved):

PR #8448, opened 2021-12-15
Remove listener on ConversationView unmount

notbakaneko:
Before I forget about it again.
This seems to have been missing forever 👀


nanaya (approved):
should probably use a unique id but that can be part of other pr which already adds an id, I guess :thinking:

PR #8449, opened 2021-12-15
Change Homepage Button Colours

### LINKED ISSUE #8174: Change the "support osu!" and "osu!store" button colours
i noticed the "support osu!" button changed colours a bit and i thought that the pink colour in the "osu!store" button fits better with the "support osu!" button.

i made a little edit just to see how it would look like.
image
this is the current one (obviously)
Untitled-1
and this is how it would look like.

(also the store button is 1px taller than the rest)

RockRoller01:
closes #8174 

should now look like this:
image

edit: I think the buttons are used nowhere else, but didn't dare to remove the green type since i am not 100% sure


resources/assets/less/bem/btn-home.less
```diff
@@ -52,6 +52,10 @@
     --move: 2px;
   }
 
+  &--c-darkorange {
+    --colour-h: var(--colour-darkorange-hue)
```
nanaya, line 56: the css (less) file is invalid due to missing semicolon
```suggestion
    --colour-h: var(--colour-darkorange-hue);
```

resources/assets/less/bem/btn-home.less
```diff
@@ -52,6 +52,10 @@
...
+  }
+
   &--c-green {
```
nanaya, line 59: yeah it's not used anywhere else

nanaya (changes requested):

nanaya (approved):

PR #8453, opened 2021-12-16
Move translation key of updated beatmap pack instruction

nanaya:
Prevents old one with removed `:scary` key from being used on other languages.


peppy (approved):

notbakaneko (approved):

YDu8VchU:
this mean new string will be added in crowdin and is useless to translate the old one, right?

PR #8454, opened 2021-12-16
Fix preset relation name in Room

notbakaneko:
There's no `$user` in `Room` (and everything that calls it is amazingly after `setRelation`)

### DIFF
--- a/app/Models/Multiplayer/Room.php
+++ b/app/Models/Multiplayer/Room.php
@@ -343,7 +343,7 @@ class Room extends Model
             'user_id' => $owner->getKey(),
         ]);
 
-        $this->setRelation('user', $owner);
+        $this->setRelation('host', $owner);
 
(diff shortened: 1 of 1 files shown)


app/Models/Multiplayer/Room.php
```diff
@@ -343,7 +343,7 @@ public function startGame(User $owner, array $rawParams)
             'user_id' => $owner->getKey(),
         ]);
 
-        $this->setRelation('user', $owner);
+        $this->setRelation('host', $owner);
```
nanaya, line 346: renaming `$owner` and `assertUserRoomAllowance` to "host" seems useful

(and why is it called host when the column is `user_id` :expressionless: )

nanaya (changes requested):

nanaya (approved):

PR #8463, opened 2021-12-22
Class-based chat test factories

notbakaneko:


tests/Models/Chat/ChannelTest.php
```diff
@@ -49,7 +49,7 @@ public function testChannelCanMessageModeratedPmChannel(?string $group, bool $ca
     {
         $user = User::factory()->withGroup($group)->create();
         $otherUser = User::factory()->create();
-        $channel = $this->createChannel([$user, $otherUser], 'moderated', 'pm');
+        $channel = $this->createChannel([$user, $otherUser], 'pm', 'moderated');
```
nanaya, line 52: the new signature for the third parameter is boolean, not string (not enforced here because no strict is declared and can be internally casted)

nanaya (changes requested):

nanaya (approved):

PR #8465, opened 2021-12-23
Clean up PlayDetailList component

nanaya:
Moved it and the related component inside profile page directory as it's its only consumer.

Also added some observables and computeds.

And section title is also now common component.


resources/assets/lib/profile-page/play-detail-list.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { classWithModifiers } from 'utils/css';
+import Controller from './controller';
+import { TopScoreSection } from './extra-page-props';
+import PlayDetail from './play-detail';
+
+type ScoreSections = TopScoreSection | 'scoresRecent';
+
+const sectionMaps = {
+  scoresBest: {
+    count: 'scores_best_count',
```
notbakaneko, line 20: `countKey`

notbakaneko (changes requested):

notbakaneko (approved):

PR #8466, opened 2021-12-24
Support joining public channels from web

notbakaneko:
Eventually to be used for creating new announcements after #8439 and #8418

- [ ] missing standalone close button
- [ ] text / tooltips needs some work

Thinking maybe the title of the modal becomes options for creating new announcement or something 🤔 
image


notbakaneko:
This will be redone with the newer design

Artrix9095:
updates?

PR #8469, opened 2021-12-24
Pt-br localization enhancements

shinobu-uwu:
These changes are closer to how we talk in the Brazilian community. We use English terms such as mapper and mapping so I think it'd make sense to use them in the localization.


frenzibyte:
Localisation enhancements should be applied in the crowdin project, see https://github.com/ppy/osu-web#contributing.

PR #8475, opened 2022-01-03
Limit title length for beatmapset model factory

nanaya:
Too long title causing test failures 😐


peppy:
Was this a CI run failure? Do you happen to still have the link? It's hard to find previous runs if they've already been reattempted on github :(.

nanaya:
https://github.com/ppy/osu-web/runs/4687924858?check_suite_focus=true

peppy (approved):

Issue #8471, opened 2021-12-27
Displaying loved #1's on userpage with a heart instead of a dash


uhuhuhh:
Loved #1s look a bit weird on the userpage and it would look a lot better with just a heart rather than a dash, kinda like this:
before:
![loved_dash](https://user-images.githubusercontent.com/60112032/147512685-38cde18f-d05b-4661-9fce-fa7a41ab41d6.png)

after:
![loved_heart](https://user-images.githubusercontent.com/60112032/147512690-efb6d6a2-5554-4630-9f18-e7d9a128fd4a.png)

apart from the aesthetic side it would also allow to differenciate loved #1s from qualified #1s
edit: would be the same thing for pp display on recent scores etc

peppy:
I can agree with this change.

Walavouchey:
does the above pr close this?

nanaya:
Indeed.

PR #8478, opened 2022-01-04
replaced dash for unranked & loved maps with heart

DrOpossum:
Implemented #8471 

The current solution displays a heart everywhere, not just for loved maps. I think this would be best as having dash, score or heart would be needlessly cluttered.


Walavouchey:
It wouldn't make sense to have a heart on a non-loved map

DrOpossum:
Alright, seems like that's what more people think. I'll adjust it accordingly then.

nanaya (changes requested):
the code doesn't compile...

nanaya (approved):

PR #8479, opened 2022-01-05
Ignore elasticsearch cursor parameter error

nanaya:
Don't report invalid cursor error and return different status code.

Also return 500 (server error) instead of 504 (gateway timeout) because the latter doesn't really make sense.


app/Libraries/Elasticsearch/Search.php
```diff
@@ -310,8 +311,15 @@ private function runQuery(string $operation, callable $callable)
                 $this->getDatadogTags()
             );
         } catch (ElasticsearchException $e) {
-            $this->error = $e;
-            $this->handleError($e, $operation);
+            $err = json_decode($e->getMessage(), true);
+
+            // handle cursor parsing error differently
+            if (is_array($err) && str_starts_with($err['error']['caused_by']['reason'] ?? '', 'Failed to parse search_after value for field ')) {
+                $this->error = new InvalidCursorException();
```
notbakaneko, line 318: probably better to move to `handleError` and have it return error to use (and set appropriate tag for datadog).

notbakaneko (changes requested):

app/Libraries/Elasticsearch/Search.php
```diff
@@ -257,14 +259,21 @@ private function getDatadogTags()
 
     private function handleError(ElasticsearchException $e, string $operation)
     {
+        $err = json_decode($e->getMessage(), true);
+
+        if (is_array($err) && str_starts_with($err['error']['caused_by']['reason'] ?? '', 'Failed to parse search_after value for field ')) {
+            $e = new InvalidCursorException();
```
nanaya, line 265: function signature says `ElasticsearchException` but php doesn't complain if it's set to something else 🤔

app/Libraries/Elasticsearch/Search.php
```diff
@@ -257,14 +259,21 @@ private function getDatadogTags()
...
+        }
+
         $tags = $this->getDatadogTags();
         $tags['class'] = get_class($e);
 
-        // Only report non query timeout errors to Sentry.
         // Printing the entire exception to log makes the breadcrumb too large to be sent to Sentry (16kb limit)
         // so we're only printing the message.
```
nanaya, line 272: Looking again, I'm not sure I understand this comment ಠ_ಠ The Sentry logging function only gets raw `$e`...

notbakaneko: Pretty sure this was for debugging timeouts when they were more common and having all relevant errors in the log file instead of bouncing between the log and sentry to find if a different exception class got thrown.

Logging after capture just causes the breadcrumb to go into the next captured error in the request.

This can probably be removed now, though 🤔

notbakaneko (approved):

PR #8484, opened 2022-01-05
Fix error trying to observe guest user's nonexistent cover

nanaya:
Whoops 👀


notbakaneko (approved):
I was expecting this not to run when guest for some reason 🤔

PR #8489, opened 2022-01-07
Add endpoint to reindex artist tracks

nanaya:
This should be hit whenever artist related tables (`artists`, `artist_albums`, `artist_tracks`) are updated. This does full reindex but at current number of tracks it's simpler to do that instead of trying to figure out which tracks to update.

Inplace reindex doesn't delete deleted tracks so a new index will need to be created (by passing `inplace=false` parameter) although I don't think it's a common case at the moment? Old index (the one previously used) will be removed by default unless `cleanup=false` is specified.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -526,7 +526,7 @@ Route::group(['prefix' => '_lio', 'middleware' => 'lio', 'as' => 'interop.'], fu
     Route::apiResource('users', 'InterOp\UsersController', ['only' => ['store']]);
 
     Route::group(['namespace' => 'InterOp'], function () {
-        route::post('artist-tracks/reindex-all', 'ArtistTracksController@reindexAll');
+        Route::post('artist-tracks/reindex-all', 'ArtistTracksController@reindexAll');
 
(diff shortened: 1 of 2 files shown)


peppy:
Does this mean the previous indices won't be automatically cleaned up?

I will be running this after each new featured artist release, so if that's the case I'd probably want them to be cleaned up automatically.

nanaya:
Unless there's track deletion involved an inplace reindex should sufficient.

routes/web.php
```diff
@@ -526,6 +526,8 @@
     Route::apiResource('users', 'InterOp\UsersController', ['only' => ['store']]);
 
     Route::group(['namespace' => 'InterOp'], function () {
+        route::post('artist-tracks/reindex-all', 'ArtistTracksController@reindexAll');
```
notbakaneko, line 529: `Route::post`

notbakaneko (changes requested):

notbakaneko (approved):
Will do for now

PR #8490, opened 2022-01-07
Update laravel

nanaya:
Seems like a good thing to do 👀 

May or may not break something.


notbakaneko (commented):
seems fine so far 🤔

notbakaneko (approved):
:shipit:

PR #8491, opened 2022-01-07
Add string cursor for forum topics

nanaya:
This gets interesting...

Part 2/?

- [x] depends on #8480


notbakaneko (changes requested):
conflicts 💥

resources/views/forum/topics/_moderate_toggle_deleted.blade.php
```diff
@@ -9,10 +9,8 @@ class="
         btn-circle--topic-nav
         btn-circle--yellow
         {{ $showDeleted ? '' : 'btn-circle--activated' }}
-        js-forum-topic-moderate--toggle-deleted
```
notbakaneko, line 12: how does the delete toggle work now? 👀

notbakaneko (changes requested):

resources/assets/lib/core-legacy/forum.coffee
```diff
@@ -70,15 +70,6 @@ class window.Forum
     parseInt @_totalPostsDiv[0].dataset.total, 10
 
 
-  # null if option not available (not moderator), false/true accordingly otherwise
-  showDeleted: =>
```
notbakaneko, line 74: `forum.topics.delete` blade is still using this; display currently doesn't update properly after deleting/restoring a post

notbakaneko (changes requested):
noticed another issue 👀

notbakaneko (approved):

PR #8500, opened 2022-01-11
Fix comment footer overlapping over "read more" button

### LINKED ISSUE #8497: "Read More" link is covered by like button on mobile comments
This makes it very hard to press (only the corners work)

https://user-images.githubusercontent.com/34715248/148758904-4149a06e-f22c-4431-9b61-fd7db8e38922.mp4

notbakaneko:
closes #8497


notbakaneko:
wonder if the button should be slighter larger for mobile, though 🤔

nanaya:
I'm not sure about fixing something that only happens in dev tool so I'm waiting for more details on where else it happens.

nanaya (changes requested):
looks like it's better to add margin on the `read more` button instead with even more margin for mobile view

resources/assets/less/bem/comment.less
```diff
@@ -197,7 +197,7 @@
   }
 
   &__row {
-    margin: -2px -5px;
+    gap: 2px;
```
nanaya, line 200: should be 4px

resources/assets/less/bem/comment.less
```diff
@@ -243,6 +243,10 @@
     .link-hover({
       color: @osu-colour-l1;
     });
+
+    @media @mobile {
+      padding: 5px 0;
```
nanaya, line 248: can use some margin for desktop as well as currently it's the only row with missing margin

nanaya (changes requested):

nanaya (approved):

PR #8501, opened 2022-01-11
Add string cursor for beatmapset discussion

nanaya:
I can't find reference where the cursors are used 🤔 The controllers/bundles don't seem to accept `cursor` parameter either.

- [x] depends on #8480


notbakaneko:
It's part of moving pagination, in particular for api responses, to using a cursor param. The bundle query should be updated to prioritize using a cursor, if anything.

nanaya:
Cursor support will need to be removed from html requests if anything other than offset pagination is going to be supported ಠ_ಠ The library probably shouldn't even be dealing with laravel offset paginator in the first place.

app/Http/Controllers/BeatmapDiscussionPostsController.php
```diff
@@ -57,12 +57,12 @@ public function destroy($id)
...
-     * Field       | Type                                                    | Description
-     * ----------- | ------------------------------------------------------- | -----------
-     * beatmapsets | [BeatmapsetCompact](#beatmapsetcompact)                 | |
-     * cursor      | [Cursor](#cursor)                                       | |
-     * posts       | [BeatmapsetDiscussionPost](#beatmapsetdiscussionpost)[] | |
-     * users       | [UserCompact](#usercompact)                             | |
+     * Field         | Type                                                    | Description
+     * ------------- | ------------------------------------------------------- | -----------
+     * beatmapsets   | [BeatmapsetCompact](#beatmapsetcompact)                 | |
+     * cursor_string | string?                                                 | |
```
notbakaneko, line 63: I think this should reference what `cursor_string` is for instead of just being `string?`

notbakaneko (changes requested):

notbakaneko (approved):

PR #8504, opened 2022-01-12
Directly import route function

nanaya:
Additionally ensure no conflicting variable and remove use of `link_to_route`.

- [x] depends on #8503


notbakaneko (changes requested):
`link_to_route` function also needs removing 🤔

notbakaneko (approved):

PR #8507, opened 2022-01-12
Move standalone React components to one directory

nanaya:
Less crowded root lib directory and less confusing `coffee/react/_components` directory being in search path.

Also renamed one file (`tooltip-context.tsx`) which doesn't contain jsx. One tsx is excluded (`gallery-contest.tsx`) because it's not a component but just a renderer class.

`user-card-tooltip` technically contains two different things: a "singleton" containing logic for tooltip display and `UserCardTooltip` component.

- [x] depends on #8506


notbakaneko (changes requested):
conflicts 👀

notbakaneko (approved):

PR #8520, opened 2022-01-17
Update default beatmaps bpm

nanaya:
Recently added to live system...

...except it's currently missing `NOT NULL` constraint 🤔 

@peppy please confirm whether or not the constraint is needed.

Also the migration entry will be `2022_01_17_104933_update_default_beatmaps_bpm`.


peppy:
I've added the not null spec. Didn't originally as it requires an actual expensive migrate operation.

peppy (approved):

PR #8522, opened 2022-01-17
Prevent sending messages to non-friends if only accepting messages from friends

### LINKED ISSUE #7395: Notify users with non-friend dms blocked when messaging a BN
Surely someone has made an issue for this before but I couldn't find anything so here goes. Apologies if I missed something.

Nomination requests over the osu chat are quite common. Users will message BNs and usually expect a reply (deny or acceptance). Especially in the acceptance case they really want to be able to read the answer.

Current Behavior:
User A is blocking dms from non friends
User A messages User B asking for a nomination (or advice, or whatever)
User B receives the message and wants to accept their request but cannot reply because they aren't friended
User A has no way of knowing there was a reply

Proposal: 
Notify users when messaging someone that can't message them back, similar to the message you get right now when replying to someone that you can't message.
Either only for certain user groups (BN, NAT, GMT...) but I think it could make sense for all users.
chrome_ReDChb9KK4

"You have messaged this user, but they can't reply to you. You might be blocking private messages from users not on your friends list."

notbakaneko:
closes #7395

Also changed the priority of the moderator permission check, otherwise there'd be the same issue with moderators sending messages but not accepting them.


resources/assets/lib/chat/chat-state-store.ts
```diff
@@ -157,6 +160,19 @@ export default class ChatStateStore implements DispatchListener {
...
+  private handleFriendUpdated(event: FriendUpdated) {
+    if (!this.isChatMounted) return;
+
+    // TODO: update other tabs...somehow
+    const channel = this.channelStore.groupedChannels.PM.find((value) => value.pmTarget === event.userId);
+    if (channel != null) {
+      getChannel(channel?.channelId).done((response) => {
+        channel.updateWithJson(response.channel);
+      });
+    }
```
nanaya, line 173: make a `refresh` function in `Channel` class and call it here? (`channel?.refresh()`)

resources/assets/lib/chat/chat-api.ts
```diff
@@ -11,6 +11,11 @@ import { action } from 'mobx';
 import Message from 'models/chat/message';
 import core from 'osu-core-singleton';
 
+interface GetChannelResponse {
+  channel: ChannelJson;
+  users: UserJson;
```
nanaya, line 16: array of UserJson

resources/assets/lib/chat/chat-state-store.ts
```diff
@@ -157,6 +160,19 @@ export default class ChatStateStore implements DispatchListener {
     this.selectChannel(event.channelId);
   }
 
+  @action
+  private handleFriendUpdated(event: FriendUpdated) {
+    if (!this.isChatMounted) return;
+
+    // TODO: update other tabs...somehow
```
nanaya, line 167: doesn't relevant for non-pm channels? 🤔 what to update here

nanaya (changes requested):

resources/assets/lib/chat/chat-state-store.ts
notbakaneko: Other tabs with chat open currently don't get updated with the new state

nanaya (approved):

PR #8527, opened 2022-01-19
Add participated multiplayer games to user profile page

### LINKED ISSUE #7320: History of multiplayer lobbies you joined
If it's possible, it would be nice to have a page that displays what multiplayer matches I joined so I can access the mp links after I accidentally leave and forget grabbing it from the mp chat.

notbakaneko:
closes #7320

Old `multiplayer` is now `playlists` and `realtime` (which is split into other types) is `multiplayer`
`category` is also not actually `category`...this naming is getting confusing 🤔 

I decided to split the lists into tabs instead of having them on the same tab at the last moment, as that seems to be more useful so I might be missing something 🤔 .


peppy:
Is "realtime" being shown to the user? Not sure we should be doing that if so.

notbakaneko:
`realtime` is in the url to match the existing category param naming; the display the for "realtime" says "multiplayer" instead

notbakaneko:
we can change the `realtime` param value to `multiplayer` if lazer updates to using `multiplayer` instead of `realtime` for the `category` param value

peppy:
If it's just in the URL I think it may be okay..

nanaya:
Category is either spotlight or normal. Previously it was normal/spotlight/realtime and then got moved out to type in #7965. Anything that still refer to "category" for those separation is just migration leftover and should be deleted/updated.

type is one of those playlists/head-to-head/team-versus. The grouping for playlists (just playlists) and realtime (head-to-head/team-versus) should then be type group or something.

app/Http/Controllers/Users/MultiplayerController.php
```diff
@@ -19,39 +20,52 @@ public function index()
     {
         $request = request();
         $user = FindForProfilePage::find($request->route('user'));
```
nanaya, line 22: not really related but just noticed this can be just `$userId` if the function signature is `index($userId)`...

resources/lang/en/users.php
```diff
@@ -301,8 +301,8 @@
                 'recent' => 'Latest',
                 'title' => 'Medals',
             ],
-            'multiplayer' => [
-                'title' => 'Multiplayer Games',
+            'playlists' => [
+                'title' => 'Online Playlist Games',
```
nanaya, line 305: where did the "Online" come from 🤔

nanaya (changes requested):

nanaya (approved):

Issue #8532, opened 2022-01-21
Popup's elements are not removed from the DOM


nerlihmax:
Initially the popups should be closed with a click handler call, but this approach does not remove the popup's element from the DOM.

[https://github.com/ppy/osu-web/blob/master/resources/assets/coffee/osu_common.coffee](url)

```coffeescript
popup: (message, type = 'info') ->
    $popup = $('#popup-container')
    $alert = $('.popup-clone').clone()

    closeAlert = -> $alert.click()

    # handle types of alerts by changing the colour
    $alert
      .addClass "alert-#{type} popup-active"
      .removeClass 'popup-clone'

    $alert.find('.popup-text').html message

    # warning/danger messages stay forever until clicked
    if type in ['warning', 'danger']
      $('#overlay')
        .off('click.close-alert')
        .one('click.close-alert', closeAlert)
        .fadeIn()
    else
      Timeout.set 5000, closeAlert

    document.activeElement.blur?()
    $alert.appendTo($popup).fadeIn()
```

I found this on score pinning information popup. But since popups, as I understand it, are using the popup function from resources/assets/coffee/osu_common.coffee, this error is common to other popups as well.

[https://getbootstrap.com/docs/3.4/javascript/#alerts](url)
If I understand correctly, referring to the documentation, I can correctly remove the popup, for example by calling $(...).alert('close')

I'm also not sure if I need to fix the behavior of the if branch for warning/danger popups, because I don't know how I can check this in practice.

P.S: I hope I made it clear what I mean, I'm not good at English :-(

PR #8533, opened 2022-01-21
Popup's elements are not removed from the DOM

nerlihmax:
https://github.com/ppy/osu-web/issues/8532

### DIFF
--- a/resources/assets/coffee/osu_common.coffee
+++ b/resources/assets/coffee/osu_common.coffee
@@ -116,6 +116,9 @@ window.osu =
     $alert = $('.popup-clone').clone()
 
     closeAlert = -> $alert.click()
+    
+    # https://getbootstrap.com/docs/4.0/components/alerts/#methods
+    disposeAlertWithFadeOut = -> $alert.fadeOut().alert('close')
 
     # handle types of alerts by changing the colour
     $alert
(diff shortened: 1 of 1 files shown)


resources/assets/coffee/osu_common.coffee
```diff
@@ -131,7 +134,7 @@ window.osu =
         .one('click.close-alert', closeAlert)
         .fadeIn()
     else
-      Timeout.set 5000, closeAlert
+      Timeout.set 5000, -> disposeAlertWithFadeOut
```
nanaya, line 137: that's not how callback works

resources/assets/coffee/osu_common.coffee
```diff
@@ -116,6 +116,9 @@ window.osu =
     $alert = $('.popup-clone').clone()
 
     closeAlert = -> $alert.click()
+    
+    # https://getbootstrap.com/docs/3.4/javascript/#alerts
+    disposeAlertWithFadeOut = -> $alert.fadeOut().alert('close')
```
nanaya, line 121: fade out being ignored aside, there's already logic for removing the element when clicked (the original callback) except it's not quite working right and that's the one that needs fixing (in `shared.coffee`)

The leaking problem also applies to warning/danger popups when clicking anywhere but close button for same reason.

nanaya (changes requested):
- do

PR #8544, opened 2022-01-26
Remove clipboard polyfill

nanaya:
No longer needed since around 2020.

### DIFF
--- a/resources/assets/lib/click-to-copy.tsx
+++ b/resources/assets/lib/click-to-copy.tsx
@@ -1,7 +1,6 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
 // See the LICENCE file in the repository root for full licence text.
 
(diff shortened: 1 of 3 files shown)


resources/assets/lib/click-to-copy.tsx
```diff
@@ -34,7 +33,7 @@ export default class ClickToCopy extends React.Component<Props> {
     }
 
     // copy url to clipboard
-    clipboard.writeText(this.props.value);
+    navigator.clipboard.writeText(this.props.value);
```
notbakaneko, line 36: `navigator.clipboard` requires https (otherwise it becomes `undefined`, I think).
Also noticed all the stuff that comes after this should be running after the promise from `writeText` completes instead.

notbakaneko (changes requested):

nanaya:
D: that's annoying nvm then (also my dev are all https which is why I didn't see anything broken)

PR #8555, opened 2022-01-28
Fix some mod names not being consistent with lazer / wiki

Joehuu:
https://osu.ppy.sh/wiki/en/Game_modifier/Autopilot
https://osu.ppy.sh/wiki/en/Game_modifier/Easy
https://github.com/ppy/osu/blob/d1158acb828e0e0b5c4fe97621f0c5f3b7010bf6/osu.Game/Rulesets/Mods/ModNoMod.cs#L13

Not sure if I should expand the `xK` to `x Keys`, see https://github.com/ppy/osu/blob/d1158acb828e0e0b5c4fe97621f0c5f3b7010bf6/osu.Game.Rulesets.Mania/Mods/ManiaModKey4.cs#L9.

I'm in the process of normalising strings in lazer to use most of web localisation and noticed this. Will make a proper discussion in the lazer repo later to ensure no one is wasting time doing the same thing.


peppy:
Keep in mind that by default, this will also purge all translations on crowdin for the altered strings. Is that what we want in this case?

Joehuu:
It seems that none of the other languages translated these strings, so go ahead. Not sure of the reason though.

notbakaneko:
Translations for mod names are disabled on crowdin.

peppy:
That's awesome.

peppy (approved):

PR #8557, opened 2022-01-28
New page extra style

nanaya:
Based on new profile page style. Some cards background have been adjusted to work on `b4` page background.

- [x] should probably check #8590 first


notbakaneko (changes requested):
background colour of the modding posts and recent votes panels need to be adjusted; they blend into the background now.

notbakaneko (approved):

PR #8561, opened 2022-01-28
Add support for country flags in markdown

### LINKED ISSUE #8083: osu! wiki: support inline country flags
quote from https://github.com/ppy/osu-wiki/issues/6017:

> While osu-web has moved over to using twemoji for flags, the wiki still relies on using the old pastel designs. The problem here is that they still contain many of the mistakes that are mentioned in [ppy/osu-web#3760](https://github.com/ppy/osu-web/issues/3760). While some of them are not easily spotted due to the small nature of the flags, there are still some obvious errors, as in the case of [Albania [`AL`]](https://github.com/ppy/osu-wiki/blob/master/wiki/shared/flag/AL.gif?rgh-link-date=2021-08-25T22%3A52%3A36Z) [(correct design)](https://commons.wikimedia.org/wiki/File:Flag_of_Albania.svg) and [Trinidad and Tobago [`TT`]](https://github.com/ppy/osu-wiki/blob/master/wiki/shared/flag/TT.gif?rgh-link-date=2021-08-25T22%3A52%3A36Z) [(correct design)](https://commons.wikimedia.org/wiki/File:Flag_of_Trinidad_and_Tobago.svg).
> 
> I'm not sure what the best solution to this problem would be, maybe there is some easy way to use with twemoji, but my familiarity with using it is next to nothing. Hopefully, it isn't as big of an issue to fix as I worry it could be.

twemoji flags were added in ppy/osu-web#7875. if this is implemented on the website rather than the wiki itself, we could throw away a lot of [manually copied, carried, and pasted flag references](https://github.com/ppy/osu-wiki/issues/328), ([example](https://github.com/ppy/osu-wiki/blob/master/wiki/Tournaments/OWC/2020/en.md?plain=1#L570-L618))

notbakaneko:
closes #8083

Renders the appropriate country flag when the syntax for it is used in the wiki; `::{ flag="AU" }::` renders the Australian flag.

image

The lazer markdown renderer for wiki articles will probably have to be updated to support it before the existing usages can be replaced @peppy .

A side-effect is inline container support is also added since it uses the `::` delimiter syntax for it.
Also fixes the parser for indexing not using the same options as the html version.

The configuration for the markdown extensions probably needs some cleaning up and the processor-specific options should probably have their own namespace but that's a separate issue (also, tests samples could probably do with some deduplicating).


nanaya (changes requested):
Haven't done checking but at least one function can use existing helper ( ﾟ ヮﾟ)

app/Libraries/Markdown/OsuMarkdown.php
```diff
@@ -287,7 +287,10 @@ private function getHtmlConverterAndExtension(): array
     private function getIndexableConverter(): MarkdownConverter
     {
         if ($this->indexableConverter === null) {
-            $environment = $this->createEnvironment();
+            $extraConfig = $this->osuExtensionConfig;
+            // remove attributes.
+            $extraConfig['attributes_allowed'] = [];
```
nanaya, line 292: shouldn't this only be when it's set? And does it even matter what's being allowed in this context? Removing this line doesn't seem to change test result (also the comment can be more descriptive on the why...)

nanaya (changes requested):

app/Libraries/Markdown/OsuMarkdown.php
notbakaneko: seems like I had some weird combination of settings when I did this part 🤔

nanaya (approved):

PR #8563, opened 2022-01-29
Throw an error when a factory fails to create models

### LINKED ISSUE #8214: Factory should error when failing to create model
Currently, it silently fails which can lead to tests passing or failing when they shouldn't, or test cases not testing the right thing. This is especially an issue for tests when the setup silently fails to create dependent models.

(on further testing it does seem to fail on the top level model but not the related ones? 🤔 )

cl8n:
resolves #8214 (i think, i ran into a similar thing when i messed up a test myself earlier). for class factory only

makes new base Factory class that runs a check before `afterCreating` callbacks, throwing if the models weren't saved. can opt out of it if the test is intentionally creating invalid model or something, like in `testNumberOfClientsIsLimited`

this is a more general version of what `UserFactory` was already doing.


nanaya:
Conflicts

nanaya (approved):

PR #8571, opened 2022-01-31
Force wrap room name

### LINKED ISSUE #8564: Participated multiplayer games card overflows
image
image

i suspect that this happens because some lobbys names are too long.

https://osu.ppy.sh/users/7188652/realtime

notbakaneko:
closes #8564

Normally it would have wrapped long names even with CJK text:
image

The issue is only if a single word is longer than the line

- [x] Safari


notbakaneko:
and then it turns out Safari _still_ doesn't support this

nanaya (dismissed):

nanaya (approved):

Issue #8360, opened 2021-11-22
User search sometimes misses results


Wieku:
I tried to find a similar issue, but couldn't find one.

![image](https://user-images.githubusercontent.com/5307179/142867056-19b9fcca-14dc-4ee1-b924-63a83a25b076.png)

As a second result we get WubWoofWolfM8 and WubWoofWolf is at the 3rd page:

![image](https://user-images.githubusercontent.com/5307179/142868015-f08bf4be-de13-4a2e-ad25-4f0a12bf1d81.png)

Although this is probably #2156

Here I get no results:
![image](https://user-images.githubusercontent.com/5307179/142867120-307fab7b-489e-4872-bcd8-9d207b700f6f.png)

Searching for `wubwoofwol` returns only 1 person but misses other "wubwoofwol*":
![image](https://user-images.githubusercontent.com/5307179/142867152-1cddc69e-74de-4e96-aa8f-0d93d1a7fb43.png)

When I search for `wubwoofwolf` I get him properly as expected:
![image](https://user-images.githubusercontent.com/5307179/142867209-72a6eab1-4532-44a6-9d37-2b9c2d632903.png)

but:
![image](https://user-images.githubusercontent.com/5307179/142867227-04daa3df-7be1-4c31-90f5-bbad4b88d962.png)

WubWoofWolfM8 (mentioned under the 1st screenshot) is missing here.

it's also missing when searching for `wubwoofwolfm`
![image](https://user-images.githubusercontent.com/5307179/142868968-1cf04875-9ea9-469d-a9c4-8e855827f90e.png)

peppy:
What is the perceived issue here? All the results match your search.

Wieku:
I think number of screenshots is overwhelming so:
- `WubWoofWolfM8` is searchable by `WubWoof` , but not by `WubWoofWolf` or `WubWoofWolfM`
- Searching by `WubWoofWo` yields no results
- Searching `WubWoofWol` returns only 1 user

peppy:
Definitely looks broken.

peppy:
Scary that we didn't notice it for a year if that's the case o_O. I'm running reindexing now.

Wieku:
Seems to be working now:
![image](https://user-images.githubusercontent.com/5307179/142912036-f7209fb3-ce7b-4448-be28-75edad331b83.png)
[...]

PR #8574, opened 2022-02-01
Improve user search ordering

nanaya:
Add tie-breaker based on last visit so online users (who are not invisible) are more visible when using prefix search.

Maybe? 🤔 

Related: #8360


notbakaneko (changes requested):
shouldn't it always be `('user_lastvisit', 'desc')`? it seems weird to prioritize by online and then flip it when the ordering changes 🤔

nanaya:
the idea is when you sort it by the other way the order should come up in the other way and not "the other way except when tie breaking then it'll always be the same order"

notbakaneko:
using `username.raw` as the tiebreaker for prefix search makes more sense, imo.

notbakaneko:
prefix search with username sort will return something like 
```
abc prefix
prefixa abc
prefixb abc
prefixz abc
zzz prefix
```

while `_score, username.raw` should return something more like
```
prefixa abc
prefixb abc
prefixz abc
abc prefix
zzz prefix
```

and with just `_score` it's free-for-all order 🤔

nanaya:
There seems to be some weirder cases like with three words name etc but it's already pretty weird anyway so ┐( ﾟ ヮﾟ)┌

nanaya:
...this `username.raw` sort is case sensitive isn't it

notbakaneko (approved):

PR #8575, opened 2022-02-01
Use dedicated plugin for TypeScript type check

nanaya:
Compilation seems to be a bit faster and `--watch` correctly watches updates to interface-only files.


notbakaneko (approved):
output seems a bit more useful 🤔

PR #8578, opened 2022-02-02
Fix 24h osu!catch scores from lazer missing count50 values

smoogipoo:
Found this as I'm reading through the new infrastructure planning.

### DIFF
--- a/app/Models/Solo/Score.php
+++ b/app/Models/Solo/Score.php
@@ -150,6 +150,7 @@ class Score extends Model
                 $score['count300'] = $statistics->Great;
                 $score['count100'] = $statistics->LargeTickHit;
                 $score['countkatu'] = $statistics->SmallTickMiss;
+                $score['count50'] = $statistics->SmallTickHit;
                 break;
             case 'mania':
                 $score['countgeki'] = $statistics->Perfect;
--- a/app/Models/Solo/Score.php
+++ b/app/Models/Solo/Score.php
@@ -112,6 +112,7 @@ class Score extends Model
             'Miss',
             'Ok',
             'Perfect',
+            'SmallTickHit',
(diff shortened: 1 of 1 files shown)


nanaya:
I think it was skipped because it's not shown anywhere on web (and doesn't exist in old score?).

smoogipoo:
Oh, hang on. This goes the other way around and I derped hard.

Right, I see. So this is basically only relevant for the 24h rolling scores, if I'm understanding the code correctly? I suppose it's fine to lose data here because these aren't stored forever there's no display?

smoogipoo:
> and doesn't exist in old score?

As for this, I believe that scores from stable do populate count50 for osu!catch in this table.

nanaya:
Hm yeah so it's not currently shown anywhere for other reason then...?

smoogipoo:
Today I learnt :smile: Fine to close this if you think it's okay as is.

peppy:
I think there's no harm in fixing this as long as the mapping is correct?

nanaya (approved):

nanaya:
I assume it means the same thing in stable...

smoogipoo:
I've double confirmed that it does.

nanaya:
Should we show it in score table on web? 🤔

smoogipoo:
For highscores (which this doesn't apply to, so it's a separate issue altogether), lazer shows the leaderboard in the form `hit/total`:
image
(Ignore the lack of spacing on the columns)

Neither lazer nor osu-web display scores in the `osu-scores` 24hr rolling table anywhere other than as a detail-less panel on the user page:
image

That being said, I believe there are third party website addons that show each judgement specifically (and I suppose they could be queried via the API).

PR #8583, opened 2022-02-03
Clean up polyfills

nanaya:
~~Three~~ All of the functions aren't needed anymore. <del>and typescript the one remaining function. That said, making it importable instead of replacing `window.localStorage` is also an option 🤔 </del>


resources/assets/lib/core-legacy/polyfills.ts
```diff
@@ -0,0 +1,28 @@
...
+    localStorage: Storage;
+  }
+}
+
+export default class Polyfills {
+  constructor() {
+    this.localStorage();
+  }
+
+  // Mainly for Safari Private Mode.
```
notbakaneko, line 17: I don't think Safari throws an error anymore in private mode 🤔

nanaya: wow they fixed things 🤔

notbakaneko (approved):

PR #8584, opened 2022-02-03
Send notification when unresolved discussion is replied to

### LINKED ISSUE #8258: Notifications to modders about unresolved mods with replies
I suggest to make it default for users to receive notifications when their mods are replied to but not marked as resolved. Sometimes a mapper might ask for clarification about a mod, but if the modder receives no notification about this, I think the modder is less likely to respond.
Example: https://osu.ppy.sh/beatmapsets/1506042/discussion/3085024/timeline#/2667379/7312429

notbakaneko:
closes #8258

Additional tests to be added in a separate PR as I try to figure out how to cleanup /separate the existing controller tests and update existing factories for them.


nanaya:
Should this have option just like comments or all replies to this are considered important?

Now I wrote it, should it ignore replies to hype...

notbakaneko:
It should be currently using the existing beatmap modding notification option 🤔

nanaya:
But that also affects beatmap watch 🤔 and while one can toggle watch on/off this one is global.

Also I feel the notification should still be sent even resolved (it's still updates). Or at least when the post resolves the discussion.

And the notifications may not get sent if there's resolving post a bit later and the posts before it are checked as replies on resolved discussion due to delays in queue...?

notbakaneko:
Extra setting options seem unnecessary, although, a different appearance and grouping to differentiate the notification would be useful 🤔

nanaya (approved):

PR #8590, opened 2022-02-04
Clean up user list page styling

nanaya:
margin-bottom of `osu-page--users` removed in #8557 accidentally also affects user listing page (friends, groups) and removed the space between content and footer. Fixed this by using different modifier and adding specific background through user list container class. Also removed some wrapper classes while at it.

Also made the background `b4` for added bonus of making hover effect of sort buttons visible.


notbakaneko (changes requested):
now the brick background is not visible 🤔

notbakaneko (approved):

PR #8591, opened 2022-02-04
Additional class-based factories for discussion-related tests

notbakaneko:
Convert `BeatmapDiscussionPost` to use class-based factory and add some extra conveniences when factorying `User`s


database/factories/UserFactory.php
```diff
@@ -115,4 +115,13 @@ public function withNote()
     {
         return $this->has(UserAccountHistory::factory(), 'accountHistories');
     }
+
+    public function withPlays(?int $count = null, ?string $mode = 'osu')
+    {
+        $sMode = studly_case($mode);
+
+        return $this->has("App\\Models\\UserStatistics\\{$sMode}"::factory()->state([
```
nanaya, line 123: `App\Models\UserStatistics\Model::getClass($mode)` is a thing

tests/BeatmapsetDisqualifyNotificationsTest.php
```diff
@@ -144,10 +144,7 @@ protected function setUp(): void
         Event::fake();
 
         $owner = User::factory()->create();
```
nanaya, line 146: not used

database/factories/BeatmapsetFactory.php
```diff
@@ -57,6 +59,12 @@ public function noDiscussion()
         return $this->state(['discussion_enabled' => false]);
     }
 
+    public function owner(?User $user = null)
+    {
+        return $this->state(['user_id' => $user ?? User::factory()])
+            ->afterMaking(fn (Beatmapset $beatmapset) => $beatmapset->creator = $beatmapset->user->username);
```
nanaya, line 65: update `creator` attribute function in definition instead (put at the bottom so `user_id` is set first).

nanaya (changes requested):

database/factories/BeatmapsetFactory.php
notbakaneko: `user` isn't part of attributes available in `definition`?

nanaya: it's not but can be looked up `User::find($attr['user_id'])->.. ?? ...`

notbakaneko: or it can be set without another lookup....

nanaya: Even then this should be part of configuration and not just one specific state (and remove default `creator` in that case).

database/factories/BeatmapsetFactory.php
```diff
@@ -10,16 +10,21 @@
...
+use App\Models\User;
 use Illuminate\Database\Eloquent\Factories\Factory;
 
 class BeatmapsetFactory extends Factory
 {
     protected $model = Beatmapset::class;
 
+    public function configure()
+    {
+        return $this->afterMaking(fn (Beatmapset $beatmapset) => $beatmapset->creator = $beatmapset->user?->username ?? $this->faker->userName());
```
nanaya, line 22: I think setting with `??=` is more appropriate here

(also the line is getting kinda long)

nanaya (changes requested):

nanaya (approved):

PR #8595, opened 2022-02-04
Fix infinite inline container parsing

notbakaneko:
fixes #8594

`getDelimiterUse` isn't quite clear on implementation and apparently the processor can go into negative lengths and keep going ಠ_ಠ


nanaya (changes requested):
Can use some tests

nanaya (approved):

PR #8599, opened 2022-02-06
Show difficulty rating with mods applied on score pages

cl8n:
random thing my friend mentioned would be nice. not sure if this fits as an include for scoretransformer, was just the easiest way to get it to that page

### DIFF
--- a/app/Models/Score/Model.php
+++ b/app/Models/Score/Model.php
@@ -54,6 +54,18 @@ abstract class Model extends BaseModel
         return snake_case(get_class_basename(static::class));
     }
 
+    public function difficultyRating(): ?float
+    {
+        $value = $this->beatmap
+            ->difficulty()
+            ->where('mode', Beatmap::modeInt(static::getMode()))
+            ->where('mods', ModsHelper::toBitset($this->enabled_mods))
+            ->first()
+            ?->diff_unified;
+
+        return $value === null ? null : round($value, 2);
+    }
+
     public function scopeDefault($query)
     {
         return $query
--- a/resources/assets/lib/scores-show/main.tsx
+++ b/resources/assets/lib/scores-show/main.tsx
@@ -20,6 +20,10 @@ export default function Main(props: Props) {
     throw new Error('score json is missing beatmap or beatmapset details');
   }
 
+  if (props.score.difficulty_rating != null) {
+    beatmap.difficulty_rating = props.score.difficulty_rating;
+  }
+
   return (
     <>
       <HeaderV4 />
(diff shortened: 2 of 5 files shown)


app/Models/Score/Model.php
```diff
@@ -54,6 +54,18 @@ public static function getMode(): string
         return snake_case(get_class_basename(static::class));
     }
 
+    public function difficultyRating(): ?float
```
peppy, line 57: If this is to actually exist here, it at very least needs a better name. A score can't have a "difficulty rating", so it would need a beatmap prefix.

Optimally I don't think it should be in here at all, but will defer to the web team to ascertain whether a structure to make that happen is possible.

nanaya: the closest thing I can think of is making it a relation called `beatmapDifficulty` instead.

Similarly for the transformer although considering difficulty rating is the only relevant value `beatmap_difficulty_rating` could be sufficient...

resources/assets/lib/scores-show/main.tsx
```diff
@@ -20,6 +20,10 @@ export default function Main(props: Props) {
     throw new Error('score json is missing beatmap or beatmapset details');
   }
 
+  if (props.score.difficulty_rating != null) {
+    beatmap.difficulty_rating = props.score.difficulty_rating;
+  }
```
nanaya, line 25: There should be indicator that it's a modified difficulty rating instead of simply overwriting it.

nanaya (changes requested):

app/Models/Score/Model.php
nanaya: looking again, relation is quite difficult because the difficulty table uses composite primary key. Short of adding virtual column and indexing it, it's probably better being an extra select column with subquery similar to [this](https://github.com/ppy/osu-web/blob/e50b1a4a16efc11fd3f491c332222a1326ac5788/app/Models/Multiplayer/Room.php#L219-L244) (although with simpler query).

nanaya: And if adding skipping difficulty query, then both mode and mods should be checked in case of convert maps (which would have different difficulty rating even when no other mods).

Oh and don't bother going through Beatmap model.

nanaya:
After some discussion, the attribute should be fetched through [external api server](https://github.com/ppy/osu-beatmap-difficulty-lookup-cache) instead. Similar to [score rank](https://github.com/ppy/osu-web/blob/e50b1a4a16efc11fd3f491c332222a1326ac5788/app/Models/Score/Best/Model.php#L185-L197) but on its own library class (with the data either fetched through score model/transformer itself or its own json).

peppy:
To elaborate, if we just want to get this in, maybe it's okay to go with it as-is rather than draw things out, but eventually we're going to want to use the lookup microservice as it will be required in the future.

PR #8604, opened 2022-02-07
More discussion-related test cleanups

notbakaneko:
continuing from #8591, more discussion-related test cleaning up, with more usage of data providers.

Also fix `TestCase::actAsUser` incorrectly preserving the previous verified status (we shouldn't be calling as multiple users in the same test case, anyway).


nanaya (changes requested):
Also conflicts

nanaya (approved):

PR #8605, opened 2022-02-07
Typescript Scoreboard Table Row component

nanaya:
With bonus of using computed current user friends mapping for friend button.

Also fixes friend row highlight not updating when adding/removing friend through user card.

- [x] depends on #8603


resources/assets/lib/beatmapsets-show/scoreboard-table-row.tsx
```diff
@@ -0,0 +1,160 @@
...
+  href: string;
+  modifiers?: Modifiers;
+}
+
+function TdLink(linkProps: TdLinkProps) {
+  return (
+    <td className={`${bn}__cell`}>
+      <a
+        className={classWithModifiers(`${bn}__cell-content`, linkProps.modifiers)}
+        href={linkProps.href}
```
notbakaneko, line 32: This is always going to be the score url; I think I originally planned to use the function for other urls, but the extra stuff in those cells made it too complicated to share them 🤔

notbakaneko (changes requested):

resources/assets/lib/beatmapsets-show/scoreboard-table-row.tsx
nanaya: There's no good way to share it short of using context or being part of the parent component (which means the convenient children syntax can't be used).

I don't want to do the previous "its own component but shares props with parent" thing. Creating new component class every time also seem equally shady.

notbakaneko: oh, it's a component 🤔

notbakaneko (approved):

PR #8611, opened 2022-02-09
Move beatmapset events listing to components directory

nanaya:
Not only conflicts with the one in coffeescript but the component isn't used by the beatmap-discussions itself.


resources/assets/lib/components/beatmapset-events.tsx
```diff
@@ -1,18 +1,18 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
 // See the LICENCE file in the repository root for full licence text.
 
+import Event, { EventViewMode } from 'beatmap-discussions/event';
```
notbakaneko, line 4: `Event` should be moved too

notbakaneko (changes requested):

notbakaneko (approved):

PR #8613, opened 2022-02-09
Merge coffee react directory with lib

nanaya:
Bye `coffee/react` directory.

Depends on:
- [x] #8609
- [x] #8611


notbakaneko (approved):

notbakaneko (changes requested):
conflict 👀

Issue #7019, opened 2020-12-22
Replace sorting by _id in elasticsearch queries


notbakaneko:
Sorting by `_id` has been discouraged since Elasticsearch 6 since the `_id` field does not have `doc_values` making it expensive to sort. This is something rarely mentioned in previous iterations despite it being the most obvious choice :D

PR #8616, opened 2022-02-10
Add model keys to elasticsearch index fields

notbakaneko:
This adds doc_value fields for the model keys to elasticsearch which is needed to do #7019. Sorting by `_id` isn't recommended and hasn't been suggested for a while now, and will probably be disabled by default in the future.

Indices will need to be updated with the new fields


config/schemas/beatmaps.json
```diff
@@ -93,6 +93,9 @@
             }
           }
         },
+        "beatmapset_id": {
```
nanaya, line 96: I think primary key should be just `id` everywhere.

nanaya (changes requested):

config/schemas/beatmaps.json
```diff
@@ -150,6 +150,9 @@
         "hype": {
           "type": "long"
         },
+        "id": {
```
nanaya, line 153: this needs mapping in the relevant function 👀 (also for users)

nanaya (changes requested):

nanaya (approved):

PR #8624, opened 2022-02-14
Update footer server status link to page rather than twitter

peppy:
Seems like a better target these days.


config/osu.php
```diff
@@ -194,7 +194,7 @@
         'installer' => 'https://m1.ppy.sh/r/osu!install.exe',
         'installer-mirror' => 'https://m2.ppy.sh/r/osu!install.exe',
         'osx' => 'https://osx.ppy.sh',
-        'server_status' => 'https://twitter.com/osustatus',
+        'server_status' => 'https://status.ppy.sh,
```
nanaya, line 197: ...missing closing quote

nanaya (changes requested):

nanaya (approved):

Issue #6572, opened 2020-08-30
API v2 beatmaps endpoint: mods and convert param.


TheEZIC:
Please add "a", and "mods" (like in api v1) query params on "beatmaps" endpoint. I know, I can get converts on "beatmapsets" endpoint. But anyway I cat't get beatmap data with mods. I need it for the beatmap stats (like SR, OD, CS etc) server side calculation.

PR #8631, opened 2022-02-15
Add query parameter to get converted beatmaps from API v2

Alexmal007:
As mentioned in #6572, we can't get converts from Get Beatmap endpoint, so I added a `mode` query parameter to convert the beatmap to specified gamemode


nanaya:
There's an endpoint in the work which will return difficulty attributes for requested combination of mods and modes so I'm not sure if this specific case is worth it.

Alexmal007:
I dont think so.. Is there any information on when it will become available?

nanaya:
Should be this or next week.

PR #8635, opened 2022-02-16
Restructure rooms listing

nanaya:
Removes beatmap include and adds `difficulty_range.min` and `difficulty_range.max` 🤔 

...when `x-api-version: 20220215` is specified on the header.

### DIFF
--- a/app/Models/Multiplayer/Room.php
+++ b/app/Models/Multiplayer/Room.php
@@ -257,12 +257,12 @@ class Room extends Model
 
                 $this->playlist->each(function ($item) use (&$ret) {
                     $rating = $item->beatmap->difficultyrating;
+                    $ret['max'] ??= $rating;
+                    $ret['min'] ??= $rating;
 
-                    if ($ret['max'] === null || $ret['max'] < $rating) {
+                    if ($ret['max'] < $rating) {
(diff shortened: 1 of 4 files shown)


notbakaneko (commented):
The user playlist and multiplayer history pages could probably make use of `difficulty_range` too since all it uses the beatmaps for is the difficulty and cover image of the first one 🤔

nanaya:
Cover url is still needed for those pages so it's not as useful there (and more changes are needed).

app/Models/Multiplayer/Room.php
```diff
@@ -243,6 +244,44 @@ public function scopeWithRecentParticipantIds($query, ?int $limit = null)
...
+        if ($this->relationLoaded('playlist')) {
+            $firstItem = $this->playlist[0];
+
+            if ($firstItem->relationLoaded('beatmap')) {
+                $ret = [
+                    'max' => null,
+                    'min' => null,
+                ];
+
+                $this->playlist->each(function ($item) use ($ret) {
```
notbakaneko, line 258: `&$ret`

notbakaneko (changes requested):

notbakaneko (approved):

PR #8639, opened 2022-02-17
Fix api version check

nanaya:
So I trolled myself 👀 

Not sure why I thought `>` is a good idea yesterday.


notbakaneko (commented):
what does this solve 🤔

nanaya:
Being able to look at specified version string and use it right away instead of having to +1

peppy (approved):
Much of a muchness at this point but sure

PR #8641, opened 2022-02-17
Update forum-related factories, seeder, and tests

cl8n:
was originally going to be just the factories but I couldn't help tidying up everything else I saw along the way of replacing factory usage

`ForumSeeder` is completely rewritten. tests are just cleaned up a little and using the new factories, but make all the same assertions as before. except I also removed `ForumTopicsControllerTest::testDestroy` because it doesn't do what it says it does and duplicates `testReply` :thinking: 

- [x] update after #8634 and see about solving the TODOs there. I haven't tried to merge that one into this branch yet.
  - [x] depends on #8634


notbakaneko (changes requested):
conflicts 💥

tests/Controllers/ForumPostsControllerTest.php
```diff
@@ -87,36 +78,25 @@ public function testDestroyNotLastPost()
...
         $this
-            ->actingAsVerified($user)
+            ->actingAsVerified($moderator)
             ->post(route('forum.posts.restore', $post))
             ->assertSuccessful();
 
         $topic->refresh();
 
-        $this->assertSame($initialPostCount + 1, Forum\Post::count());
         $this->assertSame(2, $topic->postCount());
```
notbakaneko, line 100: or `$this->expectCountChange(fn () => $topic->fresh()->postCount(), 1);`

tests/Controllers/ForumTopicsControllerTest.php
```diff
@@ -3,310 +3,192 @@
...
+use App\Models\Forum\Post;
+use App\Models\Forum\Topic;
+use App\Models\Forum\TopicTrack;
 use App\Models\User;
 use App\Models\UserStatistics\Osu as StatisticsOsu;
 use Tests\TestCase;
 
 class ForumTopicsControllerTest extends TestCase
 {
-    public function testDestroy()
```
notbakaneko, line 15: where did this test go?

tests/Controllers/ForumTopicsControllerTest.php
```diff
@@ -3,310 +3,192 @@
...
 
         $this
             ->be($user)
-            ->get(route('forum.topics.show', $topic->topic_id))
+            ->get(route('forum.topics.show', $topic))
             ->assertSuccessful();
     }
 
-    public function testStore()
+    public function testStore(): void
```
notbakaneko, line 140: now is a good time to split the test into user `withPlays` and without plays

notbakaneko (changes requested):

tests/Controllers/ForumTopicsControllerTest.php
cl8n: it was a duplicate of testReply with different name, so I removed

...but since this PR is just general cleanup/improvement of forum tests now, I'll write a new one

database/factories/Forum/PollOptionFactory.php
```diff
@@ -16,6 +19,7 @@ public function definition(): array
     {
         return [
             'poll_option_text' => fn () => $this->faker->sentence(),
+            'topic_id' => Topic::factory()->poll(0),
```
notbakaneko, line 22: set to `null` so `PollOption` can't be created without `Topic`, it doesn't make sense to factory `PollOption` by itself.

database/factories/Forum/TopicFactory.php
```diff
@@ -0,0 +1,82 @@
...
+                $this->faker->numberBetween(86400, 604800),
+                0,
+            ]),
+            'poll_max_options' => 1,
+            'poll_start' => fn (array $attr) => $attr['topic_time'],
+            'poll_title' => fn () => $this->faker->sentence(),
+            'poll_vote_change' => fn () => $this->faker->boolean(),
+        ]);
+
+        if ($optionCount <= 0) {
```
notbakaneko, line 50: throw if < 1

database/factories/BeatmapsetFactory.php
```diff
@@ -31,7 +32,7 @@ public function definition(): array
             'genre_id' => Genre::factory(),
             'language_id' => Language::factory(),
             'submit_date' => fn () => $this->faker->dateTime(),
-            'thread_id' => 0,
+            'thread_id' => Topic::factory()->withPost(),
```
notbakaneko, line 35: Beatmapset with description should be a separate state, we don't want to create an extra post and related stuff with the default factory when they're not needed.

tests/Controllers/ForumTopicsControllerTest.php
```diff
@@ -3,331 +3,234 @@
...
+use App\Models\Forum\Topic;
+use App\Models\Forum\TopicTrack;
 use App\Models\User;
-use App\Models\UserStatistics\Osu as StatisticsOsu;
 use Tests\TestCase;
 
 class ForumTopicsControllerTest extends TestCase
 {
-    public function testDestroy()
+    public function testDestroyAsModerator(): void
```
notbakaneko, line 20: missing the unauthorized users tests

notbakaneko (changes requested):

database/factories/Forum/TopicFactory.php
cl8n: instead of <1 i matched TopicPoll validation

notbakaneko (approved):

PR #8642, opened 2022-02-17
More restructure room listing json

nanaya:
Fun hacks™ inside.

Now I wonder if I can use it for `recent_participants` as well 🤔 

New JSON details can be checked [here](https://github.com/nanaya/osu-web/blob/e28389256d67d712c514df24215c2e95ce500792/resources/assets/lib/interfaces/room-json.ts). Following are included for `/rooms` endpoint:

- `current_playlist_item.beatmap.beatmapset`
- `difficulty_range`
- `host.country`
- `playlist_item_stats`
- `recent_participants`


app/Models/Multiplayer/Room.php
```diff
@@ -338,6 +356,28 @@ public function completePlay(Score $score, array $params)
...
+                return null;
+            }
+
+            $groupedItems = $this->playlist->groupBy('expired');
+
+            // the key is casted to int
+            if (isset($groupedItems[0])) {
+                $ret = $groupedItems[0]->reduce(fn (?PlaylistItem $ret, PlaylistItem $i) => $ret === null ? $i : ($i->playlist_order < $ret->playlist_order ? $i : $ret));
+            } else {
+                $ret = $groupedItems[1]->reduce(fn (?PlaylistItem $ret, PlaylistItem $i) => $ret === null ? $i : ($i->playlist_order > $ret->playlist_order ? $i : $ret));
```
notbakaneko, line 372: these lines look kind of long 👀

smoogipoo (changes requested):

app/Models/Multiplayer/Room.php
```diff
@@ -338,6 +356,34 @@ public function completePlay(Score $score, array $params)
...
+                    }
+
+                    return $i->playlist_order < $currentItem->playlist_order ? $i : $currentItem;
+                })
+                : $groupedItems[1]->reduce(function (?PlaylistItem $currentItem, PlaylistItem $i) {
+                    if ($currentItem === null) {
+                        return $i;
+                    }
+
+                    return $i->playlist_order > $currentItem->playlist_order ? $i : $currentItem;
```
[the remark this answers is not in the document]
nanaya: While at it, how should the playlist room be handled? (both playlist_order/played_at should be null, I assume?)

Should it always use first playlist item instead?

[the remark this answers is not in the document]
peppy: I think that's adequate for now (always first item).

app/Models/Multiplayer/Room.php
```diff
@@ -338,6 +356,42 @@ public function completePlay(Score $score, array $params)
         });
     }
 
+    public function findAndSetCurrentPlaylistItem(): PlaylistItem
```
notbakaneko, line 359: this will error if a `Room` has no playlist items (whose expected behaviour is currently being worked out 👀 )

notbakaneko (changes requested):

resources/assets/lib/user-multiplayer-index/room.tsx
```diff
@@ -36,21 +35,8 @@ export default class Room extends React.Component<Props> {
...
-  @computed
-  private get minDifficulty() {
-    const min = minBy(this.props.room.playlist, (playlist) => this.props.store.beatmaps.get(playlist.beatmap_id)?.difficulty_rating);
-    return this.props.store.beatmaps.get(min?.beatmap_id ?? 0)?.difficulty_rating ?? 0;
-  }
-
   private get background() {
-    const beatmap = this.props.store.beatmaps.get(this.props.room.playlist[0].beatmap_id);
-    const beatmapset = this.props.store.beatmapsets.get(beatmap?.beatmapset_id ?? 0);
+    const beatmapset = this.props.room.current_playlist_item.beatmap?.beatmapset;
```
notbakaneko, line 39: not really an issue but this can return directly by continuing with `beatmapset?.covers.cover`

notbakaneko (changes requested):

peppy:
Have tested to work well from our end 👍 .

peppy (approved):

app/Models/Multiplayer/Room.php
```diff
@@ -338,6 +360,46 @@ public function completePlay(Score $score, array $params)
...
+        return $this->memoize(__FUNCTION__, function () {
+            if ($this->playlist->count() === 0) {
+                $ret = null;
+            } else {
+                if ($this->isRealtime()) {
+                    $groupedItems = $this->playlist->groupBy('expired');
+
+                    // the key is casted to int
+                    $ret = isset($groupedItems[0])
+                        ? $groupedItems[0]->reduce(function (?PlaylistItem $currentItem, PlaylistItem $i) {
```
notbakaneko, line 374: isn't the first param the carry and the second the current? 🤔

notbakaneko (changes requested):

notbakaneko (approved):

PR #8647, opened 2022-02-18
Add endpoint to fetch beatmap difficulty attributes

### LINKED ISSUE #6572: API v2 beatmaps endpoint: mods and convert param.

nanaya:
Todo:
- [x] more tests
- [x] ~~error handler~~
- [x] docs

Resolves #6572.


notbakaneko (changes requested):
How is the test going to run without the lookup server with it not being mentioned anywhere? 🤔

notbakaneko (approved):

PR #8649, opened 2022-02-18
Refactor new discussion post and reply handling

notbakaneko:
Move the handling into its own classes with their context instead of doing it in the controller.
Removes the weird `BeatmapsetEvent::getBeatmapsetEventType` which only existed to help single posts and reviews have similar logic when handling problem discussions; instead reviews will run the same code path as standalone posts now.
Actually check if discussion is resolvable if `resolved` parameter is used instead of silently ignoring it (and making resolve/reopen explicit param instead of magic `fill`)

part 1/2; this is mostly having the updated handling work with the existing tests, part 2 is updated tests to be less dependent on pass through the controller.


app/Libraries/BeatmapsetDiscussionPostHandlesProblem.php
```diff
@@ -0,0 +1,60 @@
...
+
+namespace App\Libraries;
+
+use App\Jobs\Notifications;
+use App\Models\BeatmapDiscussion;
+use App\Models\User;
+
+abstract class BeatmapsetDiscussionPostHandlesProblem
+{
+    protected int $priorOpenProblemCount = 0;
```
nanaya, line 16: this probably doesn't need to be a count but a boolean of `exists()` check instead

app/Libraries/BeatmapsetDiscussionPostHandlesProblem.php
```diff
@@ -0,0 +1,60 @@
...
+    protected ?BeatmapDiscussion $problemDiscussion = null;
```
nanaya, line 17: this feels like should be called `newProblemDiscussion` as the other problem discussion handling (resolved/reopen) isn't using this

...and maybe rename the class to `...HandlesNewProblem`? And remove the `wasRecentlyCreated`/`exists` check

there may be further simplification possible (like moving the prior open problems check to right before assigning new problem discussion and removing the count variable entirely - or changing it as a cache variable)

```php
// both in review and non-review assignment
// app/Libraries/BeatmapsetDiscussionReview.php#148
// app/Libraries/BeatmapsetDiscussionPostNew.php#51
$this->shouldRecordNewProblem() { // or maybe roll this and next line into single function
  $this->recordNewProblem($discussion);
}

function shouldRecordNewProblem()
{
  return $this->hasPriorOpenProblems ??= ...openProblems()->exists();
}

function recordNewProblem($discussion)
{
    if ($this->newProblemDiscussion === null && ($discussion->message_type !== 'problem' || !$this->shouldRecordNewProblem())) {
        return;
    }
    $this->newProblemDiscussion = $discussion;
}
```

app/Libraries/BeatmapsetDiscussionPostHandlesProblem.php
```diff
@@ -0,0 +1,60 @@
...
+    protected function handleProblemDiscussion()
+    {
+        if ($this->problemDiscussion === null) {
+            return;
+        }
+
+        $beatmapset = $this->problemDiscussion->beatmapset;
+
+        if ($this->shouldDisqualifyOrResetNominations()) {
+            return $beatmapset->disqualifyOrResetNominations($this->user, $this->problemDiscussion);
```
nanaya, line 29: split and return void instead?

nanaya (changes requested):

nanaya:
I wonder if it's better to further split handling new discussion and replies since they barely share anything.

app/Libraries/BeatmapsetDiscussionPostHandlesProblem.php
notbakaneko: I don't see how this helps; disqualification and nomination reset depend on whether the discussion is new and the user has the right permissions, regardless of whether there are existing problems; the problem notification depends on whether there are existing problems regardless of it being a reply or new discussion but only if the action didn't result in it being disqualified or reset.

nanaya: oh huh. Also apparently I missed the confirmation thing for reopening issue ಠ_ಠ

app/Libraries/BeatmapsetDiscussionNew.php
```diff
@@ -0,0 +1,69 @@
...
+            $this->discussion->saveOrExplode();
+
+            $post = $this->discussion->beatmapDiscussionPosts()->make(['message' => $this->message]);
+            $post->user()->associate($this->user);
+            $post->saveOrExplode();
+
+            $this->handleProblemDiscussion();
+
+            // TODO: make transactional
+            (new BeatmapsetDiscussionPostNew($post, $this->user))->dispatch();
```
nanaya, line 59: put this outside the transaction maybe?

app/Libraries/BeatmapsetDiscussionNew.php
```diff
@@ -0,0 +1,69 @@
...
+
+            return [$this->discussion, [$post]];
+        });
+    }
+
+    private function getUser(): User
```
nanaya, line 65: what's the plan for this? All of the `getUser`s are exact same thing

app/Libraries/HandlesProblemBeatmapsetDiscussionPost.php
```diff
@@ -0,0 +1,59 @@
...
+
+declare(strict_types=1);
+
+namespace App\Libraries;
+
+use App\Jobs\Notifications;
+use App\Models\BeatmapDiscussion;
+use App\Models\User;
+
+// TODO: should go into namespace
```
nanaya, line 14: uh, yeah, how about now instead. Also for all other BeatmapsetDiscussion libraries (`Libraries/BeatmapsetDiscussion`, `Libraries/BeatmapsetDiscussion/Traits`)

app/Http/Controllers/BeatmapDiscussionPostsController.php
```diff
@@ -112,60 +111,31 @@ public function store()
...
-
-                case BeatmapsetEvent::DISQUALIFY:
-                case BeatmapsetEvent::NOMINATION_RESET:
-                    $discussion->beatmapset->disqualifyOrResetNominations($user, $discussion);
-                    break;
-            }
+        if ($discussionId !== null) {
+            $discussion = BeatmapDiscussion::findOrFail($discussionId);
+            $resolve = get_bool($request['beatmap_discussion']['resolved'] ?? null);
+            $message = presence(get_string($request['beatmap_discussion_post']['message'] ?? null));
```
nanaya, line 122: inside the class just like `BeatmapsetDiscussionNew`? Seems weird one is dealt here and the other is in the class.

app/Libraries/BeatmapsetDiscussionReply.php
```diff
@@ -0,0 +1,110 @@
...
+            return;
+        }
+
+        // if a resolved state change was requested, check if someone else got to it first.
+        $discussion = $this->discussion->lockSelf();
+        if ($discussion->resolved !== $this->discussion->resolved) {
+            throw new InvariantException('resolved state of the discussion has changed');
+        }
+
+        if ($discussion->resolved !== $this->resolve) {
```
nanaya, line 96: already established in the constructor and combined with the check above should sufficient?

app/Libraries/BeatmapsetDiscussionReply.php
```diff
@@ -0,0 +1,110 @@
...
+        if ($resolve !== null) {
+            if (!$discussion->canBeResolved()) {
+                throw new InvariantException("{$discussion->message_type} does not support resolving.");
+            }
+
+            if ($discussion->resolved !== $resolve) {
+                if ($resolve) {
+                    priv_check_user($user, 'BeatmapDiscussionResolve', $discussion)->ensureCan();
+                } else {
+                    priv_check_user($user, 'BeatmapDiscussionReopen', $discussion)->ensureCan();
```
nanaya, line 46:
```php
$priv = $resolve ? '...' : '...';
priv_check_user(..., $priv, ...)->...;
```

app/Libraries/BeatmapsetDiscussionNew.php
```diff
@@ -0,0 +1,69 @@
...
+class BeatmapsetDiscussionNew
+{
+    use HandlesProblemBeatmapsetDiscussionPost;
+
+    private BeatmapDiscussion $discussion;
+    private ?string $message;
+
+    public function __construct(private User $user, private Beatmapset $beatmapset, array $request)
+    {
+        priv_check_user($this->user, 'BeatmapDiscussionPostStore', $beatmapset)->ensureCan();
```
nanaya, line 24: move to be part of `BeatmapDiscussionStore`?

app/Libraries/BeatmapsetDiscussionNew.php
```diff
@@ -0,0 +1,69 @@
...
+
+        $params = get_params($request, 'beatmap_discussion', [
+            'beatmap_id:int',
+            'message_type',
+            'timestamp:int',
+        ], ['null_missing' => true]);
+
+        $this->discussion = new BeatmapDiscussion(['resolved' => false]);
+        $this->discussion->fill($params);
+        $this->discussion->beatmapset()->associate($beatmapset);
```
nanaya, line 34: `$this->discussion = $beatmapset->beatmapDiscussions()->make(...)`?

nanaya (changes requested):

nanaya (approved):

PR #8651, opened 2022-02-20
Filter out blocked user's comments on beatmap pages. 

### LINKED ISSUE #8462: Allow users to block/hide other users' comments
I think that the comments sections of maps are long overdue for a block/hide function that allows users to prevent themselves from seeing comments from other users. Reporting doesn't work when there is outside context to the message or simply annoying individuals who aren't annoying enough for GMT to do anything, so a block/hide function would be great.

jacobvm04:
Fixes #8462 by filtering out comments written by users that are blocked by the logged in user when querying comments.


notbakaneko (changes requested):
This doesn't do anything for replies.

We don't want to be being it this way, either; for query performance reasons and for how it breaks comment threads

jacobvm04:
> This doesn't do anything for replies.
> 
> We don't want to be being it this way, either; for query performance reasons and for how it breaks comment threads

Ok, I've revised the PR to fit this criteria.

Now, comments from blocked users display
- In place of the message content, a message saying that the comment is not shown because it was from a blocked user.
- In place of the comment's author's name, "Blocked user".
- In place of the comment's author's avatar, the default avatar.
- An action button to display the comment anyway.

notbakaneko (changes requested):
There's been some discussion about reducing the display to a smaller and simpler `This comment is hidden` type thing (still undecided), but at the very least the blocked comment should be dimmed like the deleted ones.

jacobvm04:
> There's been some discussion about reducing the display to a smaller and simpler `This comment is hidden` type thing (still undecided), but at the very least the blocked comment should be dimmed like the deleted ones.

I've reduced the display message to your suggested one for now, the initial longer message was basically a placeholder and the suggested one sounds like a good replacement for now. 

Additionally, I've added dimming to the blocked comments like that of the deleted comments, and have set blocked comment's replies to be collapsed by default.

resources/assets/lib/components/comment.coffee
```diff
@@ -592,3 +622,6 @@ export class Comment extends React.PureComponent
 
   toggleClip: =>
     @setState clipped: !@state.clipped
+
+  toggleForceShow: =>
```
notbakaneko, line 626: 2 lines between methods for coffeescript

resources/assets/less/bem/comment.less
```diff
@@ -140,7 +141,11 @@
     position: relative;
 
     &--deleted {
-      opacity: 0.5;
+      opacity: @_dim_opacity;
+    }
+
+    &--blocked {
```
notbakaneko, line 147: use and rename the existing `deleted` modifier instead.

notbakaneko (changes requested):
I mean something smaller like
Screen Shot 2022-03-11 at 16 59 20
; no icon, no username, no votes, just a single line.
Showing a blocked comment should remain dimmed, just like deleted comments

resources/assets/lib/components/comment.coffee
```diff
@@ -211,6 +210,26 @@ export class Comment extends React.PureComponent
...
+    div className: blockClass,
+      div className: mainClass,
+        span 
+          className: if @props.depth > 0 then 'comment__avatar' else ''
+          style:
+            height: 'auto'
+
+        div className: 'comment__container',
+          div className: 'comment__main blocked comment__message osu-md osu-md--comment',
+            el 'p',
```
notbakaneko, line 227: use `p` from react-dom-factories

resources/assets/lib/components/comment.coffee
```diff
@@ -211,6 +210,26 @@ export class Comment extends React.PureComponent
...
+
+    div className: blockClass,
+      div className: mainClass,
+        span 
+          className: if @props.depth > 0 then 'comment__avatar' else ''
+          style:
+            height: 'auto'
+
+        div className: 'comment__container',
+          div className: 'comment__main blocked comment__message osu-md osu-md--comment',
```
notbakaneko, line 226: `osu-md osu-md--comment` should be on the `p` element with `osu-md__paragraph`.
`comment__main blocked` shouldn't be here.

resources/assets/lib/components/comment.coffee
```diff
@@ -90,6 +97,10 @@ export class Comment extends React.PureComponent
       parent = store.comments.get(@props.comment.parentId)
       user = @userFor(@props.comment)
       meta = commentableMetaStore.get(@props.comment.commentableType, @props.comment.commentableId)
+      @isBlocked = isBlocked(user)
+
+      if !@props.comment.isDeleted && @isBlocked && !@state.forceShow
+        return @renderBlocked()
```
notbakaneko, line 103: The entire comment should be dimmed if the user is blocked, not just the placeholder.

resources/assets/lib/components/comment.coffee
```diff
@@ -198,6 +210,26 @@ export class Comment extends React.PureComponent
               label: osu.trans('comments.load_replies') if @children.length == 0
               ref: @loadMoreRef
 
+  
+  renderBlocked: =>
```
notbakaneko, line 214: should take `mainClass` and `blockClass` param so the values from `render` can just be passed in.

notbakaneko (changes requested):

notbakaneko (approved):

PR #8652, opened 2022-02-20
Add voter participation sticker for Project Loved

cl8n:
> [11:18 PM] DigitalHypno: unrelated to my last question, is it possible to create some sort of "I Voted" sticker on people's profiles, similar to the world cup banners that people can buy? clicking on the banner would then link to the current Loved voting post, and the sticker would go away when voting closes

didn't take very long to try so here it is, it appears in the same place as tournament banners but a bit smaller. comes after the tournament banner if both are present

not complete yet but it Kinda Works:tm:, figured I should PR now and save my time if it's not wanted or needs to be very different.
- design and image is just random thing I put together with loved assets, also there's no 2x version rn
- no mobile design at all. can't figure anything out that doesn't look dumb (css is currently a mess because I was trying things for mobile)
- poll cache might not be necessary :thinking: idk if there is an easier way to select for open polls


peppy:
This will definitely require @arflyte input.

arflyte:
Hmm, will look into this since I'm fixing the profile page nau. I can allocate temporary participation stuff somewhere.

Hiviexd:
@arflyte bump about this, lots of things have changed with the profile redesign so we'd love to have your input on where to fit this mini banner 🙏

Hiviexd:
Any chance we could revisit this idea in the near future? would be very handy to repurpose in a more arbitrary fashion so it's not Loved-specific.

Discussion about a temporary participation badge display came up internally for Locus, and it would be great to have it for that + any other ongoing events like Loved, fanart contests, Community Choice or any other official events that could benefit from it.

peppy:
This looks really low effort to get in and would definitely be a cool detail.

peppy:
@cl8n any interest in rebasing this and preparing for review?

PR #8660, opened 2022-02-23
Updated tests for refactored discussion post handling

notbakaneko:
Less going through the controller for tests, easier to check what's actually making a test pass or fail (which was really annoying while doing #8649).

Includes the extra test for #8584

- [x] #8649


nanaya:
Conflicts 💥

database/factories/BeatmapDiscussionFactory.php
```diff
@@ -26,6 +27,13 @@ class BeatmapDiscussionFactory extends Factory
 
     protected $model = BeatmapDiscussion::class;
 
+    public function configure()
+    {
+        return $this->afterCreating(function (BeatmapDiscussion $discussion) {
+            $discussion->wasRecentlyCreated = false;
```
nanaya, line 33: can use a comment on why this is needed (and seems like something that should be its own state instead of always applied just for this one model?)

(or maybe just `->fresh()` right after create)

tests/Libraries/BeatmapsetDiscussion/DiscussionTest.php
```diff
@@ -0,0 +1,420 @@
...
+        $beatmapset->watches()->create(['user_id' => $watcher->getKey()]);
+
+        Queue::fake();
+
+        (new Discussion($user, $beatmapset, $this->makeParams('praise'), static::TEST_MESSAGE))->handle();
+
+        Queue::assertPushed(BeatmapsetDiscussionPostNew::class, function (BeatmapsetDiscussionPostNew $job) use ($user, $watcher) {
+            return in_array($watcher->getKey(), $job->getReceiverIds(), true)
+                && !in_array($user->getKey(), $job->getReceiverIds(), true);
+        });
```
nanaya, line 123: kinda long line. Maybe arguments on their own line and use the short anonymous function syntax

also similarly for some others

nanaya (changes requested):

database/factories/BeatmapDiscussionFactory.php
notbakaneko: or just stop relying on `wasRecentlyCreated` as it kept resulting in obtuse behaviour when I was moving things around ಠ_ಠ

nanaya (approved):

PR #8672, opened 2022-02-28
More detailed typing of BeatmapsetEvent

nanaya:
Removes the need of explicit type predicates. The comment field typing for some of them is a bit questionable though 👀


resources/assets/lib/interfaces/beatmapset-event-json.ts
```diff
@@ -4,23 +4,260 @@
...
+
+interface LoveEvent extends BaseBeatmapsetEvent {
+  comment: null;
+  type: 'love';
+}
+
+interface RemoveFromLovedEvent extends BaseBeatmapsetEvent {
+  comment: {
+    beatmap_discussion_id?: number;
+    beatmap_discussion_post_id?: number;
```
notbakaneko, line 35: event doesn't have a related discussion, though?

resources/assets/lib/interfaces/beatmapset-event-json.ts
```diff
@@ -4,23 +4,260 @@
...
+    beatmap_discussion_id: number;
+    beatmap_discussion_post_id: number;
+  };
+  type: 'kudosu_deny';
+}
+
+interface KudosuGainEvent extends BaseBeatmapsetEvent {
+  comment: {
+    beatmap_discussion_id?: number;
+    beatmap_discussion_post_id?: number;
```
notbakaneko, line 84: `beatmap_discussion_id` shouldn't be nullable and `beatmap_discussion_post_id` isn't relevant; same for `KudosuLostEvent` and `KudosuRecalculateEvent`

resources/assets/lib/interfaces/beatmapset-event-json.ts
```diff
@@ -4,23 +4,260 @@
...
+    new_vote: VoteEventJson;
+    votes: VoteEventJson[];
```
notbakaneko, line 86: old events don't have these

resources/assets/lib/interfaces/beatmapset-event-json.ts
```diff
@@ -4,23 +4,260 @@
...
+  comment: {
+    beatmap_discussion_id?: number;
+    beatmap_discussion_post_id?: number;
+  };
+  type: 'kudosu_recalculate';
+}
+
+interface IssueResolveEvent extends BaseBeatmapsetEvent {
+  comment: {
+    beatmap_discussion_id?: number;
```
notbakaneko, line 111: discussion_id for the issue-related events aren't nullable (but post_id is for oldest events)

resources/assets/lib/interfaces/beatmapset-event-json.ts
```diff
@@ -4,23 +4,260 @@
...
+    beatmap_discussion_post_id?: number;
+  };
+  type: 'issue_reopen';
+}
+
+interface DiscussionLockEvent extends BaseBeatmapsetEvent {
+  comment: {
+    beatmap_discussion_id: number;
+    beatmap_discussion_post_id: number;
```
notbakaneko, line 128: lock and unlock don't use these

resources/assets/lib/interfaces/beatmapset-event-json.ts
```diff
@@ -4,23 +4,260 @@
...
+  comment: {
+    beatmap_discussion_id?: number;
+    beatmap_discussion_post_id?: number;
+  };
+  type: 'discussion_unlock';
+}
+
+interface DiscussionDeleteEvent extends BaseBeatmapsetEvent {
+  comment: {
+    beatmap_discussion_id?: number;
```
notbakaneko, line 144: pretty sure deleting discussion has `beatmap_discussion_id` 🤔

resources/assets/lib/interfaces/beatmapset-event-json.ts
```diff
@@ -4,23 +4,260 @@
...
+    beatmap_discussion_post_id?: number;
+  };
+  type: 'discussion_delete';
+}
+
+interface DiscussionRestoreEvent extends BaseBeatmapsetEvent {
+  comment: {
+    beatmap_discussion_id: number;
+    beatmap_discussion_post_id: number;
```
notbakaneko, line 153: doesn't have / always null `beatmap_discussion_post_id`

notbakaneko (changes requested):
Wonder if we should consider making the `$extraData` param of `BeatmapsetEvent::log` private to the class as well so it's easier to trace the actual contents of `comment` instead of having them scattered all over 🤔

notbakaneko (approved):

PR #8675, opened 2022-02-28
Remove tournament banner padding

nanaya:
~~Moved to be inside profile cover.~~ back out but with no paddings

- [x] depends on #8674 because they update same part of code


notbakaneko (commented):
are there going to be more changes 👀

nanaya:
...no? 👀 @arflyte

arflyte:
I'm not sure which one this refers to D:

notbakaneko (approved):

PR #8677, opened 2022-03-01
Update styling of username change history

nanaya:
Now with way more codes to hopefully correctly position the box.

Originally planned to reuse PopupMenu but the needed behavior is beyond what it's designed for. Or maybe it can be eventually updated to deal with it as well (positioning, transition, hover+click active state).


nanaya:
Not happening, apparently ( ﾟ ヮﾟ)

And on semi-related note, I noticed there's positioning bug when there are too many usernames.

PR #8689, opened 2022-03-03
Adjust mobile version of profile page details

nanaya:
Equal width stats column (to align with ranking display) and fix font size of ranking display.


notbakaneko (commented):
this makes the desktop rank text smaller too?

notbakaneko (approved):

PR #8690, opened 2022-03-03
Fix ranks section notice background

nanaya:
Current `b4` is the same as section background...


notbakaneko (changes requested):
one says `profile-page-extra` the other says `profile-extra-page` 🤔

notbakaneko (approved):

PR #8694, opened 2022-03-04
Match BeatmapIcon tooltip difficulty colour

### LINKED ISSUE #8662: SR text is using wrong color spectrum when hovering on diff icons in discussions
some few obvious examples:
image
image
image

map: https://osu.ppy.sh/beatmapsets/1487999/discussion

notbakaneko:
closes #8662

Now with badge instead
image
image

Now the problem is the icon itself not being very visible in the dropdown 👀 

- [x] #8726


nanaya:
Something needs to be done about this...

nanaya (changes requested):
Something needs to be done about this...

nanaya:
maybe update the style to be like the one on beatmaps popup. The background would be rather invisible on that but the content would be readable? @arflyte

nanaya (changes requested):
looks like the css `--diff-*` and `@diff-*` variables aren't used anymore (and the `beatmaps-popup-item__col--difficulty` isn't used either).

nanaya (approved):

PR #8706, opened 2022-03-08
Fix active rooms end time display

nanaya:
Turns out end time is nullable.


notbakaneko (changes requested):
Actually, what's the display ordering supposed to be if the room doesn't end? 🤔

nanaya:
There's a default `id` tie breaker.

notbakaneko:
`order by ends_at desc` puts null at the bottom, is this what we want, though? 👀

nanaya:
That can be fixed/discussed separately I think

nanaya:
oh cool mysql doesn't even have `nulls [first | last]` operator 😐

PR #8708, opened 2022-03-08
Hide maps without scores for multiplayer display page

nanaya:
...maybe? 🤔 Only for realtime/multiplayer rooms.


notbakaneko (commented):
Should this affect `Room::playlistItemStats` as well? Since the panel in the multiplayer history page will say `x` maps and the one on the rankings page will say `y` maps

nanaya:
Those aren't quite related because one is showing active room state (`/users/id/realtime`) and the other is current room result (`/multiplayer/rooms/id`)

PR #8712, opened 2022-03-08
Use custom cast for user preferences

nanaya:
When trying to clean up score model, I found that returning empty array on `CastAttributes#set` results in noop. It can then be used to assign specific field instead of the casting field. This is what I originally intend #8550 to be 😐 

`extras_order` stays because it should be updated to get rid of the column first... (or the other way around by getting rid of it from `options`)

Also I don't know why I didn't make `user_id` the table's primary key 🤔 

It looks like I need to dig a bit deeper on this CastAttributes thing because it behaves in a rather unexpected way.


nanaya:
Let's just pretend this never happened.

PR #8714, opened 2022-03-09
Add "Adaptive Speed" mod

bdach:
Reference: https://github.com/ppy/osu/pull/17054

### DIFF
--- a/app/Libraries/Multiplayer/Mod.php
+++ b/app/Libraries/Multiplayer/Mod.php
@@ -29,6 +29,7 @@ class Mod
     const MIRROR = 'MR';
     const MUTED = 'MU';
     const NO_SCOPE = 'NS';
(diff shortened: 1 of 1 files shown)


app/Libraries/Multiplayer/Mod.php
```diff
@@ -118,10 +119,9 @@ class Mod
             self::HALFTIME,
             self::DAYCORE,
             self::NIGHTCORE,
-        ],
-        [
```
bdach, line 122: The removal of these brackets is intentional because as far as I can tell this was always wrong. {Half,Double} Time, {Day,Night}core and Wind {Up,Down} were always mutually exclusive and should have been in one incompatibility group all along. Adaptive Speed is incompatible with them all, so this both needed for correctness and sort of killing two birds with one stone.

peppy (approved):

PR #8716, opened 2022-03-11
Combine isReview check

nanaya:
That's a lot of same things.


notbakaneko (commented):
is it even possible for `@props.type` to not be `discussion` when `message_type` is `review` 🤔

nanaya:
a reply...?

notbakaneko (approved):
I forgot you could even reply to reviews 👀

PR #8719, opened 2022-03-11
Report improvements

### LINKED ISSUE #8123: Changes to the report system
Recent talks inside the GMT regarding our response rate to web-reports have lead to the following 2 suggestions for the report system:

-   **All score reports should be routed to the cheating endpoint.** Currently only ones marked as "cheating" land on the support teams feed and ones marked as "other" land in our. There is not really any use case for "other" for score reports, they always are cheating reports.
-   **Reports should not have an empty reason/description.** Many reports, especially user profile ones, come in wihtout any context and no obvious case. Having context would make it easier to figure out whats going on.

iirc @venix12 he would implement these if there are no issues with these suggestions

venix12:
resolves #8123
- redirects all score related reports to the cheating endpoint
- requires comments for report to be submittable

### DIFF
--- a/app/Models/UserReport.php
+++ b/app/Models/UserReport.php
@@ -70,7 +70,7 @@ class UserReport extends Model
 
     public function routeNotificationForSlack(?Notification $_notification): ?string
     {
-        if ($this->reason === 'Cheating' || $this->reason === 'MultipleAccounts') {
+        if ($this->reason === 'Cheating' || $this->reason === 'MultipleAccounts' || str_starts_with($this->reportable_type, 'score_best')) {
             return config('osu.user_report_notification.endpoint_cheating');
         } else {
             return config('osu.user_report_notification.endpoint_moderation');
@@ -95,6 +95,12 @@ class UserReport extends Model
     public function isValid()
     {
         $this->validationErrors()->reset();
+
+        $commentsLength = mb_strlen(trim($this->comments));
+
+        if ($commentsLength === 0) {
+            $this->validationErrors()->add('comments', 'required');
+        }
 
         if ($this->user_id === $this->reporter_id) {
             $this->validationErrors()->add(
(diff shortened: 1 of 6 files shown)


app/Models/UserReport.php
```diff
@@ -70,7 +70,7 @@ public function reporter()
 
     public function routeNotificationForSlack(?Notification $_notification): ?string
     {
-        if ($this->reason === 'Cheating' || $this->reason === 'MultipleAccounts') {
+        if ($this->reason === 'Cheating' || $this->reason === 'MultipleAccounts' || str_starts_with($this->reportable_type, 'score_best')) {
```
nanaya, line 73: `$this->reportable()->getModel() instanceof BestModel`

app/Models/UserReport.php
```diff
@@ -96,6 +96,12 @@ public function isValid()
     {
         $this->validationErrors()->reset();
 
+        $commentsLength = mb_strlen(trim($this->comments));
+
+        if ($commentsLength === 0) {
```
nanaya, line 101: use `present(trim(...))`

resources/assets/lib/components/report-form.coffee
```diff
@@ -95,15 +100,17 @@ export class ReportForm extends PureComponent
         className: "#{bn}__row"
         textarea
           className: "#{bn}__textarea"
+          onChange: @handleCommentsChange
```
nanaya, line 103: the label should be changed to just "Comments" (or some other better wordings).

resources/assets/lib/components/report-form.coffee
```diff
@@ -95,15 +100,17 @@ export class ReportForm extends PureComponent
...
           placeholder: osu.trans 'users.report.placeholder'
           ref: @textarea
```
nanaya, line 105: `sendReport` should be updated to use state and this ref variable can be removed

nanaya (changes requested):
This should've been two separate PRs...

nanaya (approved):

PR #8726, opened 2022-03-15
Convert BeatmapIcon tooltip to typescript

notbakaneko:
plus some typing fixes and behaviour tweaks.

The `declare global` extension in `user-card-tooltip` actually affects other modules, so that's been moved.


nanaya (approved):

nanaya (changes requested):
...I mean this ಠ_ಠ

nanaya (approved):

PR #8727, opened 2022-03-16
Update to latest React

nanaya:
Virtual List project seems dead and will probably stop working on next react update...


notbakaneko (approved):
seems working

PR #8734, opened 2022-03-19
Display guest difficulties on the user profile

### LINKED ISSUE #8196: Display Player's Guest Difficulties in the User page

venix12:
resolves #8196

Adds a new beatmap section in the user profile, showing beatmapsets with user guest participation (in leaderboard (ranked, loved, approved) -> pending -> wip -> graveyard order).


apollo-dw:
this should only be ranked and loved, since there's nothing stopping people from (falsely) assigning someone as the GDer

app/Models/User.php
```diff
@@ -2043,6 +2043,14 @@ public function profileBeatmapsetsLoved()
             ->with('beatmaps');
     }
 
+    public function profileBeatmapsetsGuest()
+    {
+        return Beatmapset::whereNotIn('beatmapset_id', $this->beatmapsets->pluck('beatmapset_id'))
+            ->whereHas('beatmaps', fn ($q) => $q->where('user_id', $this->getKey()))
+            ->active()
+            ->with('beatmaps');
```
nanaya, line 2051: as mentioned in the comment, this should hide pending and lower state maps. Also I haven't tested it but this probably should be `where('user_id', '<>', $userId)` instead of `whereNotIn(...)` (or at least use subquery instead of separate query).

nanaya (changes requested):

nanaya (approved):

PR #8740, opened 2022-03-22
Add support for multiplayer auto countdown timers

smoogipoo:
Duration provided in seconds as I don't believe there's a time-span compatible MYSQL data type. Capped at SMALLINT as we're likely never going to support auto-start countdowns longer than 5 minutes.

### DIFF
--- a/app/Models/Multiplayer/Room.php
+++ b/app/Models/Multiplayer/Room.php
@@ -462,6 +462,7 @@ class Room extends Model
             'playlist:array',
             'type',
             'queue_mode',
+            'auto_start_duration:int',
         ], ['null_missing' => true]);
 
         $this->fill([
@@ -470,6 +471,7 @@ class Room extends Model
             'starts_at' => now(),
             'type' => $params['type'],
             'queue_mode' => $params['queue_mode'],
+            'auto_start_duration' => $params['auto_start_duration'],
             'user_id' => $host->getKey(),
         ]);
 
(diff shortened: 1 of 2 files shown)


database/migrations/2022_03_18_125356_add_auto_start_duration_to_multiplayer_rooms.php
```diff
@@ -0,0 +1,32 @@
...
+class AddAutoStartDurationToMultiplayerRooms extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('multiplayer_rooms', function (Blueprint $table) {
```
peppy, line 16: Hesitant to keep adding columns for each setting like this (we've moved to `json` storage for other similar cases). Fine this time to keep the changes simple, but just noting my concerns.

smoogipoo: Absolutely. And this column is being used for both playlists and multiplayer which have different requirements that are being bunched up together here.

That sort of change is a little too involved for me to make in quick order so I'll let someone else handle the web-side of things whenever that happens.

peppy (approved):

peppy (approved):

app/Models/Multiplayer/Room.php
```diff
@@ -462,6 +462,7 @@ public function startGame(User $host, array $rawParams)
             'playlist:array',
             'type',
             'queue_mode',
+            'auto_start_duration:int',
```
notbakaneko, line 465: Is this field going to be optional?
Does `0` mean "not set" or "start immediately"?

peppy: 0 means no auto

PR #8742, opened 2022-03-22
Update location bar when switching chat channels

notbakaneko:
...although this is really about no longer returning uninitialized channels from the state store and removing the `selectedBoxed` boxed primitive so there'll be less layers to go through when updating the channel selection logic for create/join new channels/announcements.

- [x] don't remove chat stuff during component unmount if still on chat
- [x] initial channel selection needs fixing; previous assumptions don't quite work

~~The `u-chat`/`osu-layout--mobile-app` class is moved to the body since it _seems_ to work now on current mobile Safari, except there are new issues (that happen even without the change) with mobile Safari that need fixing separately.~~
Turns out it's needed on `html` thanks to safari's scrolling but the layout shift navigating through history seems fixed by applying `fancy-scrollbar` separately to the body...

### DIFF
--- a/resources/assets/lib/chat/chat-state-store.ts
+++ b/resources/assets/lib/chat/chat-state-store.ts
@@ -17,18 +17,24 @@ import { updateQueryString } from 'utils/url';
 import ChannelJoinEvent from './channel-join-event';
 import ChannelPartEvent from './channel-part-event';
 import { getUpdates } from './chat-api';
+import MainView from './main-view';
 import PingService from './ping-service';
 
 @dispatchListener
 export default class ChatStateStore implements DispatchListener {
-  @observable isChatMounted = false;
   @observable isReady = false;
   skipRefresh = false;
+  @observable viewsMounted = new Set<MainView>();
   @observable private isConnected = false;
   private lastHistoryId: number | null = null;
   private pingService: PingService;
   @observable private selected = 0;
   private selectedIndex = 0;
+
+  @computed
+  get isChatMounted() {
+    return this.viewsMounted.size > 0;
+  }
 
   @computed
   get selectedChannel() {
@@ -49,6 +55,14 @@ export default class ChatStateStore implements DispatchListener {
       // refocus channels if any gets removed
       if (changes.type === 'delete') {
         this.refocusSelectedChannel();
+      }
+    });
+
+    autorun(() => {
+      if (this.isChatMounted) {
+        document.querySelector('html')?.classList.add('u-chat');
+      } else {
+        document.querySelector('html')?.classList.remove('u-chat');
       }
     });
 
--- a/resources/assets/lib/chat/chat-state-store.ts
+++ b/resources/assets/lib/chat/chat-state-store.ts
@@ -93,7 +93,7 @@ export default class ChatStateStore implements DispatchListener {
   }
 
   @action
-  selectChannel(channelId: number) {
+  selectChannel(channelId: number, updateUrl = true) {
     if (this.selected === channelId) return;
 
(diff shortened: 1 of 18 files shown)


nanaya (changes requested):

resources/assets/lib/chat/chat-state-store.ts
```diff
@@ -62,6 +58,14 @@ export default class ChatStateStore implements DispatchListener {
       }
     });
 
+    autorun(() => {
+      if (this.isChatMounted) {
+        document.querySelector('html')?.classList.add('u-chat');
+      } else {
+        document.querySelector('html')?.classList.remove('u-chat');
```
notbakaneko, line 65: unmount of the previous page happening after the current page mounts does cause the layout to flash when going through history

nanaya: maybe at least the scrollbar thing should just be done in the chat component class instead of global ಠ_ಠ

nanaya: or just move the `u-chat` to be part of `chat/index.blade.php`'s `bodyAdditionalClasses` if that works...

notbakaneko: as far as I know this was on `html` because of iOS...except it turns out there's new problems with with the layout on the latest iOS ಠ_ಠ

resources/assets/lib/chat/chat-state-store.ts
```diff
@@ -103,7 +99,9 @@ export default class ChatStateStore implements DispatchListener {
   }
 
   @action
-  selectChannel(channelId: number) {
+  selectChannel(channelId: number, mode: 'advanceHistory' | 'replaceHistory' = 'advanceHistory', updateHistory = true) {
```
nanaya, line 102: updateHistory seem redundant if null can be passed for mode (and mode is already a noop when updateHistory is false).

resources/assets/lib/stores/channel-store.ts
```diff
@@ -117,14 +116,16 @@ export default class ChannelStore implements DispatchListener {
   }
 
   @action
-  getOrCreate(channelId: number): Channel {
+  getOrCreate(json: ChannelJson): Channel {
```
nanaya, line 119: similar to the one in user; "get" also update with this change.

nanaya (changes requested):

notbakaneko:
Seems like it make more sense to have a `channel doesn't exist view` when navigating to a non-existent channel instead of trying to maintain the current behaviour of defaulting to the first channel...

resources/assets/lib/chat/conversation-panel.tsx
```diff
@@ -0,0 +1,43 @@
...
+            {core.dataStore.channelStore.channels.size > 0 ? (
+              <>
+                <div className='chat-conversation-panel__title'>{osu.trans('chat.not_found.title')}</div>
+                <div className='chat-conversation-panel__instructions'>{osu.trans('chat.not_found.message')}</div>
+              </>
+            ) : (
+              <>
+                <div className='chat-conversation-panel__title'>{osu.trans('chat.no-conversations.title')}</div>
+                <div className='chat-conversation-panel__instructions'>{osu.trans('chat.no-conversations.howto')}</div>
+                <div dangerouslySetInnerHTML={{__html: osu.trans('chat.no-conversations.lazer', { link: lazerLink })}} />
```
notbakaneko, line 35: I'll be updating this and all the translations separately

nanaya (approved):

PR #8747, opened 2022-03-24
Add UI for sending chat announcements

notbakaneko:
Join/create channel UI retrofitted into the existing design.
Initially limited to sending announcements to help simplify the component for now.
The component will eventually be expanded into a more general join channel + new conversation feature.

- [ ] #8742
- [ ] Announcement channels should show the list of users

### DIFF
--- a/resources/assets/lib/chat/chat-state-store.ts
+++ b/resources/assets/lib/chat/chat-state-store.ts
@@ -93,7 +93,7 @@ export default class ChatStateStore implements DispatchListener {
   }
 
   @action
-  selectChannel(channelId: number) {
+  selectChannel(channelId: number, updateUrl = true) {
     if (this.selected === channelId) return;
 
     // mark the channel being switched away from as read.
@@ -110,8 +110,10 @@ export default class ChatStateStore implements DispatchListener {
     this.selected = channelId;
     this.selectedIndex = this.channelList.indexOf(channel);
 
-    const url = updateQueryString(null, { channel_id: channelId.toString() });
-    Turbolinks.controller.advanceHistory(url);
+    if (updateUrl) {
+      const url = updateQueryString(null, { channel_id: channelId.toString() });
+      Turbolinks.controller.advanceHistory(url);
+    }
 
     // TODO: should this be here or have something else figure out if channel needs to be loaded?
     this.channelStore.loadChannel(channelId);
@@ -121,7 +123,7 @@ export default class ChatStateStore implements DispatchListener {
   selectFirst() {
     if (this.channelList.length === 0) return;
 
-    this.selectChannel(this.channelList[0].channelId);
+    this.selectChannel(this.channelList[0].channelId, false);
   }
 
   @action
--- a/resources/assets/lib/chat/chat-state-store.ts
+++ b/resources/assets/lib/chat/chat-state-store.ts
@@ -13,6 +13,7 @@ import { clamp, maxBy } from 'lodash';
 import { action, autorun, computed, makeObservable, observable, observe, runInAction } from 'mobx';
 import Channel from 'models/chat/channel';
 import ChannelStore from 'stores/channel-store';
+import { updateQueryString } from 'utils/url';
 import ChannelJoinEvent from './channel-join-event';
 import ChannelPartEvent from './channel-part-event';
 import { getUpdates } from './chat-api';
@@ -108,6 +109,9 @@ export default class ChatStateStore implements DispatchListener {
 
     this.selected = channelId;
     this.selectedIndex = this.channelList.indexOf(channel);
+
+    const url = updateQueryString(null, { channel_id: channelId.toString() });
+    Turbolinks.controller.advanceHistory(url);
 
     // TODO: should this be here or have something else figure out if channel needs to be loaded?
     this.channelStore.loadChannel(channelId);
--- a/resources/assets/lib/chat/chat-state-store.ts
+++ b/resources/assets/lib/chat/chat-state-store.ts
@@ -93,7 +93,7 @@ export default class ChatStateStore implements DispatchListener {
   }
 
   @action
-  selectChannel(channelId: number, updateUrl = true) {
+  selectChannel(channelId: number, replaceUrl = false) {
     if (this.selected === channelId) return;
 
     // mark the channel being switched away from as read.
@@ -110,9 +110,12 @@ export default class ChatStateStore implements DispatchListener {
     this.selected = channelId;
     this.selectedIndex = this.channelList.indexOf(channel);
 
-    if (updateUrl) {
-      const url = updateQueryString(null, { channel_id: channelId.toString() });
-      Turbolinks.controller.advanceHistory(url);
+    if (replaceUrl) {
+      // Remove channel_id from location on selectFirst();
+      // also handles the case when history goes back to a channel that was removed.
+      Turbolinks.controller.replaceHistory(updateQueryString(null, { channel_id: null }));
+    } else {
+      Turbolinks.controller.advanceHistory(updateQueryString(null, { channel_id: channelId.toString() }));
     }
 
     // TODO: should this be here or have something else figure out if channel needs to be loaded?
@@ -123,7 +126,7 @@ export default class ChatStateStore implements DispatchListener {
   selectFirst() {
     if (this.channelList.length === 0) return;
 
-    this.selectChannel(this.channelList[0].channelId, false);
+    this.selectChannel(this.channelList[0].channelId, true);
   }
 
   @action
--- a/resources/assets/lib/chat/chat-state-store.ts
+++ b/resources/assets/lib/chat/chat-state-store.ts
@@ -110,12 +110,22 @@ export default class ChatStateStore implements DispatchListener {
     this.selected = channelId;
     this.selectedIndex = this.channelList.indexOf(channel);
 
-    if (replaceUrl) {
+    if (channel.newPmChannel) {
+      Turbolinks.controller.replaceHistory(updateQueryString(null, {
+        channel_id: null,
+      }));
+    } else if (replaceUrl) {
       // Remove channel_id from location on selectFirst();
       // also handles the case when history goes back to a channel that was removed.
-      Turbolinks.controller.replaceHistory(updateQueryString(null, { channel_id: null }));
+      Turbolinks.controller.replaceHistory(updateQueryString(null, {
+        channel_id: null,
+        sendto: null,
+      }));
     } else {
-      Turbolinks.controller.advanceHistory(updateQueryString(null, { channel_id: channelId.toString() }));
+      Turbolinks.controller.advanceHistory(updateQueryString(null, {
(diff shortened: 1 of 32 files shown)


nanaya:
Conflicts 🥹

resources/assets/lib/models/chat/create-announcement.ts
```diff
@@ -0,0 +1,172 @@
...
+      message,
+      target_ids: [...this.validUsers.keys()],
+      type: 'ANNOUNCE' as const,
+      uuid: this.uuid,
+    };
+  }
+
+  @action
+  updateUsers(text: string, immediate: boolean) {
+    this.busy.lookupUsers = true;
```
nanaya, line 98: emptying the input box shouldn't show the spinner. And maybe also when the change is only on whitespace.

resources/assets/lib/components/user-card-brick.tsx
```diff
@@ -70,7 +78,17 @@ export default class UserCardBrick extends React.Component<Props> {
         <div className='user-card-brick__username u-ellipsis-overflow'>
           {this.props.user.username}
         </div>
+        {this.props.onRemoveClick != null && (
+          <div className='user-card-brick__remove'>
+            <span className='fas fa-times' onClick={this.handleRemoveClick}/>
```
nanaya, line 83: onClick on the div? Or the hover effect not be part of `div` which has extra padding (which highlights but doesn't actually remove)

resources/assets/lib/components/user-card-brick.tsx
```diff
@@ -70,7 +78,17 @@ export default class UserCardBrick extends React.Component<Props> {
...
+          </div>
```
nanaya, line 84: First this should be a button (otherwise can't be targeted with keyboard), and then this shouldn't be inside an `a` element.

resources/assets/lib/models/chat/create-announcement.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  private validUsersContains(userId?: string | null) {
+    if (userId == null) return false;
+
+    const userIdNumber = Number(userId);
+    if (Number.isInteger(userIdNumber) && this.validUsers.has(userIdNumber)) return true;
+
+    // maybe it's a username
+    for (const user of this.validUsers.values()) {
```
nanaya, line 166:
```suggestion
    return (Number.isInteger(userIdNumber) && this.validUsers.has(userIdNumber))
      // maybe it's a username
      || [...this.validUsers.values()].some(...);
```

resources/lang/en/chat.php
```diff
@@ -13,18 +14,43 @@
         'user' => 'You cannot message this user at this time.',
     ],
 
+    'channels' => [
+        'join' => 'join channel',
```
nanaya, line 18: as nothing seem to be set in stone yet regarding all this join channel and create channel/group/announcement thing, this should say new announcement for now (or something in that effect). Especially if there will be different interface for each action

resources/assets/lib/chat/join-channel.tsx
```diff
@@ -0,0 +1,170 @@
...
+  <div className='chat-join-channel__spinner'>
+    {busy && <Spinner />}
+  </div>
+);
+
+// TODO: look at combining with ValidatingInput
+const InputContainer = observer((props: React.PropsWithChildren<InputContainerProps>) => (
+  <div className={classWithModifiers('chat-join-channel__input-container', { error: props.error })}>
+    {props.labelKey && (
+      <label className='chat-join-channel__input-label'>{osu.trans(props.labelKey)}</label>
```
nanaya, line 32: the `div` and `label` should be swapped so clicking anywhere on the area will focus on the input. Or at least give the label appropriate name property

resources/assets/lib/chat/join-channel.tsx
```diff
@@ -0,0 +1,170 @@
...
+    )}
+    {props.children}
+  </div>
```
nanaya, line 35: also could be combined with `artist-track-search-form__input-group` (which has wrong border-radius 👀 )

resources/assets/lib/models/chat/channel.ts
```diff
@@ -30,10 +30,18 @@ export default class Channel {
...
+  @observable userIds: number[] = [];
 
   @observable private messagesMap = new Map<number | string, Message>();
   private serverLastMessageId?: number;
+  @observable private usersLoaded = false;
+
+  @computed
+  get announcementUsers() {
+    return this.usersLoaded
+      ? this.userIds.map((userId) => core.dataStore.userStore.get(userId))
```
nanaya, line 42: should probably be sorted (by name case insensitive seems good)

resources/assets/lib/models/chat/create-announcement.ts
```diff
@@ -0,0 +1,172 @@
...
+   */
+  @action
+  private extractValidUsers() {
+    const userIds = this.inputs.users.split(',');
+    if (userIds.length === 0) return false;
+
+    const invalidUsers: string[] = [];
+
+    for (const userId of userIds) {
+      const trimmedUserId = osu.presence(userId.trim());
```
nanaya, line 119: I don't know if adding your own id makes sense. Or if it should always be the first item and not removable instead.

nanaya (changes requested):

resources/assets/lib/models/chat/create-announcement.ts
```diff
@@ -0,0 +1,172 @@
...
+  return (inputKeys as Readonly<string[]>).includes(key);
+}
+
+export default class CreateAnnouncement {
+  @observable busy = {
+    create: false,
+    lookupUsers: false,
+  };
+
+  debouncedLookupUsers = debounce(action(() => this.lookupUsers()), 1000);
```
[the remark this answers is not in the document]
notbakaneko: `debounce` doesn't use the decorator 👀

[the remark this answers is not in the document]
nanaya: it doesn't but debounce isn't exactly debouncing the exact function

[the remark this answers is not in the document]
nanaya: I guess my initial wording was bad 🤔

resources/assets/lib/models/chat/create-announcement.ts
```diff
@@ -0,0 +1,231 @@
...
+      return;
+    }
+
+    try {
+      this.xhrLookupUsers = this.fetchUsers(userIds);
+      const response = await this.xhrLookupUsers;
+
+      this.extractValidUsers(response.users);
+    } catch (error) {
+      osu.ajaxError(error);
```
nanaya, line 217: this shouldn't be called if the error is from `.abort()` call.

nanaya (changes requested):

resources/assets/lib/chat/chat-state-store.ts
```diff
@@ -166,6 +185,13 @@ export default class ChatStateStore implements DispatchListener {
   @action
   private handleChatChannelJoinEvent(event: ChannelJoinEvent) {
     this.channelStore.update(event.json);
+
+    if (this.waitJoinChannelUuid != null && this.waitJoinChannelUuid === event.json.uuid) {
+      this.selectChannel(event.json.channel_id);
+      if (event.json.type === 'ANNOUNCE') {
+        this.createAnnoucement.clear();
+      }
```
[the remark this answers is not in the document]
notbakaneko: This already only affects the tab that sent the message, conflicts with tab contents not updating with local storage changes from other tabs is a different kind of issue .

resources/assets/lib/models/chat/create-announcement.ts
```diff
@@ -0,0 +1,252 @@
...
+        if (userIds.length > 0) {
+          this.updateUsers(userIds.join(','), true);
+        }
+      } catch (error) {
+        console.error('invalid json in localStorage');
+        localStorage.removeItem(localStorageKey);
+      }
+    }
+
+    autorun(() => {
```
nanaya, line 119: missing disposer

nanaya (changes requested):

resources/assets/lib/components/user-card-brick.tsx
```diff
@@ -60,21 +67,31 @@ export default class UserCardBrick extends React.Component<Props> {
...
-          className='user-card-brick__group-bar'
-          style={osu.groupColour(group)}
-          title={group?.name}
-        />
-
-        <div className='user-card-brick__username u-ellipsis-overflow'>
-          {this.props.user.username}
-        </div>
-      </a>
+      <div ref={this.ref} className={`js-usercard ${blockClass}`} data-user-id={this.props.user.id}>
```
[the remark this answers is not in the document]
nanaya: this hasn't quite been fixed. The padding is now correct but the border is still part of the container and not a link.

resources/assets/lib/models/chat/create-announcement.ts
```diff
@@ -0,0 +1,252 @@
...
+    this.validUsers.clear();
+    // localStorage key not removed because the currently the autorun will fill it again with empty values.
+  }
+
+  @action
+  create() {
+    if (!this.isValid || this.busy.has('create')) return;
+
+    // busy state should remain active so the same model can't be used to send multiple requests
+    // until the entire workflow is complete, including switching the channel.
```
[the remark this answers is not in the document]
notbakaneko: no, it's literally what it says; only one create announcement is allowed to be active at any time and it's only considered complete when the web socket message arrives and the UI switches the channel to the new channel.

resources/assets/lib/models/chat/create-announcement.ts
notbakaneko: this class is created once and doesn't get unreferenced ( ﾟ ヮﾟ)

resources/assets/lib/models/chat/create-announcement.ts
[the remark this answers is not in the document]
nanaya: should probably be mentioned in this comment then. And maybe listen for the event in this model instead.

resources/assets/lib/models/chat/create-announcement.ts
```diff
@@ -0,0 +1,252 @@
...
+      type: 'ANNOUNCE' as const,
+      uuid: this.uuid,
+    };
+  }
+
+  private validUsersContains(userId?: string | null) {
+    if (userId == null) return false;
+
+    const userIdNumber = Number(userId);
+    return (Number.isInteger(userIdNumber) && this.validUsers.has(userIdNumber))
```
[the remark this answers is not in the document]
notbakaneko: I think this is more assuming the API doesn't return user with `id` like `1.1` 🤔

nanaya:
conflicts. Also maybe at least reply on the comments I posted 🤔

notbakaneko:
I haven't looked at this in a while 👀

nanaya:
conflicts

PR #8750, opened 2022-03-28
Use new score structure for frontend

nanaya:
There are more things remain to be done after this:

- figuring out perfect state
- pp and replay state
- linking to solo scores url (`/scores/<id>` instead of/in addition to current `/scores/mode/<id>`)
- etc

Currently only for non-api request. Updating for API can be done with API versioning at which point osu-web (for frontend) will be something like version `99999999` to always get the latest version.

Depends on

- [ ] #8732 
- [ ] #8729 (not quite but the typing does)


app/Http/Controllers/BeatmapsController.php
```diff
@@ -298,8 +300,9 @@ public function scores($id)
 
             static $scoreIncludes = ['user', 'user.country', 'user.cover'];
 
+            $scoreTransformer = is_api_request() ? new ScoreTransformer() : new SoloScoreTransformer();
             $results = [
-                'scores' => json_collection($query->visibleUsers()->forListing(), 'Score', $scoreIncludes),
+                'scores' => json_collection($query->visibleUsers()->forListing(), $scoreTransformer, $scoreIncludes),
```
notbakaneko, line 305: transformer for `$userScore` also needs to be updated, same for `userScores` method.

notbakaneko (changes requested):

notbakaneko (approved):

PR #8752, opened 2022-03-29
Remove unused User.load() feature

notbakaneko:
We never ended up making use of it and probably won't with the current direction

### DIFF
--- a/resources/assets/lib/models/user.ts
+++ b/resources/assets/lib/models/user.ts
@@ -30,29 +30,15 @@ export default class User {
 
   static fromJson(json: UserJson): User {
     const user = new User(json.id);
-    return Object.assign(user, {
-      avatarUrl: json.avatar_url,
-      countryCode: json.country_code,
-      defaultGroup: json.default_group,
-      groups: json.groups,
-      isActive: json.is_active,
-      isBot: json.is_bot,
-      isOnline: json.is_online,
-      isSupporter: json.is_supporter,
-      loaded: true,
-      pmFriendsOnly: json.pm_friends_only,
-      profileColour: json.profile_colour,
-      username: json.username,
-    });
+
+    user.updateFromJson(json);
+
+    return user;
   }
 
   is(user?: User | UserJson | null) {
     if (user == null) return false;
     return user.id === this.id;
-  }
-
-  load() {
-    // TODO: do automagic loading stuff?
   }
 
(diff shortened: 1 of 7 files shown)


resources/assets/lib/models/user.ts
```diff
@@ -28,33 +28,11 @@ export default class User {
     makeObservable(this);
   }
 
-  static fromJson(json: UserJson): User {
```
nanaya, line 31: looks like `loaded` property can be removed

resources/assets/lib/stores/user-store.ts
```diff
@@ -27,25 +27,16 @@ export default class UserStore implements DispatchListener {
   }
 
   @action
-  getOrCreate(userId: number, props?: UserJson): User {
+  getOrCreate(json: UserJson): User {
```
nanaya, line 30: this looks more like `updateOrCreate` 🤔 (or maybe just `update` with implied create if missing)

resources/assets/lib/models/user.ts
```diff
@@ -28,33 +28,11 @@ export default class User {
...
-      loaded: true,
-      pmFriendsOnly: json.pm_friends_only,
-      profileColour: json.profile_colour,
-      username: json.username,
-    });
-  }
-
   is(user?: User | UserJson | null) {
     if (user == null) return false;
     return user.id === this.id;
```
nanaya, line 33: rather unrelated but this can be just `return user != null && ...`

nanaya (changes requested):

tests/karma/osu-core.spec.ts
```diff
@@ -26,7 +26,8 @@ describe('OsuCore user:update subscriber testing thing', () => {
       username: 'foo',
     };
 
-    const user = User.fromJson(json);
+    const user = new User(json.id);
+    user.updateWithJson(json);
```
nanaya, line 30: just noticed this but the listener actually expects json...

nanaya (changes requested):

tests/karma/osu-core.spec.ts
nanaya: ... 🤔 

this test is weird (hardcoded `1` and trying to shallow compare User instance)

nanaya (approved):

PR #8758, opened 2022-03-30
Fix null return for replay field in scores

smoogipoo:
I noticed the client was receiving responses with `"replay": null` values, which causes client-side exceptions.

It was changed in https://github.com/ppy/osu-web/pull/8731, and I believe it to be an unintentional change since the sequence `?? null` doesn't make sense.


notbakaneko (approved):
somehow I had a brain fart thinking this was going to be a file when it just a bool 🤔

nanaya:
This was originally meant so the value is only available for scores with best score id. Also the `?? null` is so php doesn't fail on undefined variable.

That said setting it to false is fine anyway I guess 🤷

PR #8759, opened 2022-03-30
Tweak chat input box position on mobile safari

notbakaneko:
`position: fixed` with `bottom: 0` doesn't track with the bouncy scroll at the end. (noticed it only happens if we move the chat full screen class to `body` 🤔 )

There's still the stupid gap when the address bar is minimized but mobile safari doesn't appear to include that in any sizing (i.e. it returns all heights assuming the full bar is visible) and gets even worse when the keyboard is visible.

### DIFF
--- a/resources/assets/less/bem/chat-input.less
+++ b/resources/assets/less/bem/chat-input.less
@@ -16,7 +16,6 @@
   display: flex;
 
   @media @mobile {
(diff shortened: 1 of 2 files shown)


notbakaneko:
not needed since the we're still leaving the class on `html`

PR #8760, opened 2022-03-30
Blur the redundant user profile header image

nanaya:
Hopefully this doesn't kill people browser 👀

### DIFF
--- a/resources/assets/less/bem/header-v4.less
+++ b/resources/assets/less/bem/header-v4.less
@@ -114,6 +114,7 @@
 
     .@{_top}--users & {
       opacity: 0.25;
+      filter: blur(50px);
     }
   }
 


peppy:
This seems like a very bad idea.

That is all I'll say.

notbakaneko (commented):
This seems like the equivalent of putting png with random glow gradient...

image
image

nanaya:
@arflyte 🤷

arflyte:
Gah. Fine, I'll look into other ways to get the desired effect.

peppy:
I mean we can try it, but also be ready to revert if people complain about performance on some odd browser/platform combination? As long as the cached blur isn't invalidated all should be fine, but I don't think we can guarantee that...

peppy:
Let's deploy and see how many performance complaints arrive.

PR #8761, opened 2022-03-31
Clean up BeatmapsetJson typing

nanaya:
I wonder if BeatmapsetExtendedJson and BeatmapsetJson should be just in one file instead...


notbakaneko (approved):
Probably can combine the interfaces when combining the transformers and make the extra fields just includes 🤔

PR #8763, opened 2022-04-02
Add additional icons in recent events

### LINKED ISSUE #8410: Add an icon statements for events in "recent"
For example, it may have this:

image

This is necessary to eliminate the "voids" in this graph

KeziahMoselle:
## Events

- [x] beatmapPlaycount
- [x] beatmapsetApprove
- [x] beatmapsetDelete
- [x] beatmapsetRevive
- [x] beatmapsetUpdate
- [x] beatmapsetUpload
- [x] rankLost
- [x] userSupportAgain
- [x] userSupportFirst
- [x] userSupportGift
- [x] usernameChange

### Colors

Should we color them as proposed initially? If so how?
With additionnal classes based on event type in `resources/assets/less/bem/page-extra.less` ?

closes #8410

### DIFF
--- a/resources/assets/less/bem/profile-extra-entries.less
+++ b/resources/assets/less/bem/profile-extra-entries.less
@@ -94,6 +94,38 @@
     height: @item-height;
     margin-right: 5px;
     font-size: 14px; // icon size
+
+    &--ranked {
+      color: var(--beatmapset-ranked-bg);
+    }
+
+    &--qualified {
+      color: var(--beatmapset-qualified-bg);
+    }
+
+    &--loved {
+      color: var(--beatmapset-loved-bg);
+    }
+
+    &--approved {
+      color: var(--beatmapset-approved-bg);
+    }
+
+    &--danger {
+      color: hsl(var(--hsl-red-3));
+    }
+
+    &--green {
+      color: @green;
+    }
+
+    &--yellow {
+      color: @yellow;
+    }
+
+    &--pink {
+      color: @pink;
+    }
   }
 
   &__time {
--- a/resources/assets/less/bem/profile-extra-entries.less
+++ b/resources/assets/less/bem/profile-extra-entries.less
@@ -112,19 +112,19 @@
     }
 
     &--danger {
-      color: hsl(var(--hsl-red-3));
+      color: hsl(var(--hsl-red-1));
     }
 
     &--green {
-      color: @green;
+      color: hsl(var(--hsl-green-1));
     }
 
-    &--yellow {
-      color: @yellow;
+    &--orange {
+      color: hsl(var(--hsl-orange-1));
     }
 
     &--pink {
-      color: @pink;
+      color: hsl(var(--hsl-pink-1));
     }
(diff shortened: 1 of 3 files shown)


peppy:
Are you able to provide a screenshot of how this looks so far?

KeziahMoselle:
Not yet, I need to seed the user profile with all possible types of events but I didn't see how yet

Loreos7:
Perhaps this down below would sute better "has been revived from eternal slumber"
image
Although it could be smth like osu circle with the wings

And dont forget about the colors, because, at least, there are several statements for the beatmap: ranked, qualified, loved, approved

KeziahMoselle:
> Perhaps this down below would sute better for "has been revived from eternal slumber" image Although it could be smth like osu circle with the wings
> 
> And dont forget about the colors, because, at least, there are several statements for the beatmap: ranked, qualified, loved, approved

afaik the icon wave pulse isn't available in font awesome v5

RockRoller01:
I believe a play icon for "map has been played x times" would be better, not to mention thats also what the play count medal uses

https://fontawesome.com/icons/play?s=solid

resources/assets/less/bem/profile-extra-entries.less
```diff
@@ -94,6 +94,38 @@
...
+      color: @green;
+    }
+
+    &--yellow {
+      color: @yellow;
+    }
+
+    &--pink {
+      color: @pink;
+    }
```
KeziahMoselle, line 128: I am unsure about this class organization 🤔

KeziahMoselle:
Here is a first version with colors: 
image

resources/assets/lib/profile-page/parse-event.tsx
```diff
@@ -60,6 +64,7 @@ export default function parseEvent(event: EventJson, modifiers: Modifiers): { ba
 
     case 'beatmapsetUpdate':
       return {
+        badge: <span className='fas fa-sync-alt profile-extra-entries__icon--green' />,
```
nanaya, line 67:
```suggestion
        badge: <span className='fas fa-sync-alt profile-extra-entries__icon--green' />,
        iconModifiers: 'green',
```

And then adjust the render function to use the modifiers and apply it correctly to the `__icon` element.

resources/assets/less/bem/profile-extra-entries.less
```diff
@@ -94,6 +94,38 @@
...
+    &--approved {
+      color: var(--beatmapset-approved-bg);
+    }
+
+    &--danger {
+      color: hsl(var(--hsl-red-3));
+    }
+
+    &--green {
+      color: @green;
```
nanaya, line 119: those `@$colour`s shouldn't be used anymore. Use `--hsl-$colour-$ver` instead.

nanaya (changes requested):

KeziahMoselle:
Updated @nanaya 👍

nanaya (approved):

Issue #8736, opened 2022-03-19
Basic implementation of spotlights systems


venix12:
This issue is proposal of implementation of basic spotlights sytems into the osu-web, I'm willing to work on it, putting a draft in here, open for improvements.

### Basic infrastructure
- Add spotlights marker on spotlighted beatmapsets, and allow to filter by it (similar to how FA tags work). Those would be linked to URL from spotlights table mentioned below, or remain unclickable for legacy spotlights/charts implementations.
- Create ``spotlights`` table, which would contain fields such as title, gamemode, URL (such as newspost, wiki or something), and start/end dates beatmappack id (?) for modern (playlist-based) spotlights seasons along with ``spotlights_playlists`` pivot table where the lazer playlists would be attached to certain spotlights season.
- Create frontend for ``spotlights`` where the data would be displayed along with a list of atttached playlist lobbies.

This kind of implementation could make a base for creating a spotlights leaderboard in the future.

### Some ideas (would be nice to have but not really high priority)
- Some form of "subscribtion" - a person could follow spotlights and get a notification when new lobbies open (are attached to spotlights season).
- Admin panel of some sort to manage that all - probably can be just done via inserts/commands, would still be more handy to use, i guess.

peppy:
I think a lot of what is asked for here should potentially be added to the playlist system itself, ie. allowing any playlist to have an attached description, URL, etc.

Maybe we could also implement a "seasons" system where multiple playlists could be grouped together (implement the backend for spotlights but eventually let others use this?)

nanaya:
The title says "Basic implementation" so is it a different system from the lazer playlist's spotlight flag?

peppy:
The goal here is to give spotlights more of an official presence as they expand in scope (ie. more users are interacting with the system). Making it visible on a beatmap's page (and potentially.. the beatmap listing as proposed), as well as allowing better linking of metadata so players can get more info on the spotlight/playlist they are playing on.

To get some context you can check the latest news post regarding the system usage (https://osu.ppy.sh/home/news/2022-02-15-beatmap-spotlights-season-5-winter-2022). At very least I'd want to see the ability for us to add descriptive text (maybe full markdown?) so spotlight ranking pages (and in-game display) could give users the same amount of detail as the news post. Or just link to the news post if that's simpler.

[...]

PR #8768, opened 2022-04-04
Add support for adding spotlight label to beatmaps

venix12:
green one :eyes: (maybe pink would make more sense to match the spotlights marker in lazer?)

- needs to queue reindex when setting spotlight value, should there be an on-site UI for that perhaps?
- should run ``php artisan es:index-documents --types=beatmapsets --yes`` after merging

part of [#8736](https://github.com/ppy/osu-web/issues/8736)


peppy:
I think this should be fine to get merged in to see how it works?

@venix12 One question I have is whether the intention here is to mark any beatmap that was ever included in a spotlight, or whether to only mark active spotlight beatmaps?

Ephemeralis:
Marking any map ever featured in a spotlight has been the intention historically.

venix12:
could also have them be a different colour or so for the active ones in the future (when the seasons system is in place, would just take playlist items of ongoing ones)

database/migrations/2022_04_04_152502_add_spotlight_to_osu_beatmapsets.php
```diff
@@ -0,0 +1,35 @@
+<?php
```
nanaya, line 1: update the date in filename

also the `osu_`/`Osu` prefix can be removed

(optionally change the class into anonymous class like in `2022_04_12_...` migration)

resources/assets/lib/components/beatmapset-panel.tsx
```diff
@@ -67,9 +67,9 @@ const BeatmapDots = observer(({ compact, beatmaps, mode }: { beatmaps: BeatmapJs
   </div>
 ));
 
-const FeaturedArtistBadge = () => (
-  <span className='beatmapset-badge beatmapset-badge--featured-artist beatmapset-badge--panel'>
-    {osu.trans('beatmapsets.featured_artist_badge.label')}
+const BeatmapsetBadge = ({ type }: { type: string }) => (
+  <span className={`beatmapset-badge beatmapset-badge--${type} beatmapset-badge--panel`}>
+    {osu.trans(`beatmapsets.${type.replace('-', '_')}_badge.label`)}
```
nanaya, line 72: `replaceAll` instead of `replace`

nanaya (changes requested):

nanaya (changes requested):

nanaya:
Conflicts 👀

nanaya (approved):
this involves adding a boolean column `spotlight` to `osu_beatmapsets`. It takes a bit on my system so if to be done manually, the migration entry will be `2022_04_30_152502_add_spotlight_to_beatmapsets`

peppy (approved):

PR #8771, opened 2022-04-05
Use plain html for rating and fail chart

nanaya:
No need to use d3 to generate bar charts.


notbakaneko (changes requested):
old graph seems taller...sometimes

`beatmapset-info` seems to be only at `min-height` with this but `max-height` with d3 unless there was no graph originally 🤔 

Maybe the graph height should be fixed?

nanaya:
It looks like the old one has buggy behavior where svg has default height of 150, incorrectly pushing the height of its container element (and the measurement happens after the svg is added).

notbakaneko (commented):
seems faster 🤔

notbakaneko (approved):

PR #8775, opened 2022-04-05
Typescript formatNumber

nanaya:
And add a default formatter to speed things up a bit for basic usage.


resources/assets/lib/utils/html.ts
```diff
@@ -26,6 +26,30 @@ export function createClickCallback(target: unknown) {
...
 
+const defaultNumberFormatter = new Intl.NumberFormat(window.currentLocale);
+
+// Duplicated definition with implementation not having explicit return type
+// so the conditional return type behaves as expected.
+// Note that actual implementation return type isn't properly checked in this case.
+// Reference: https://github.com/microsoft/TypeScript/issues/33912
+export function formatNumber<T extends number | null>(num?: T, precision?: number, options?: Intl.NumberFormatOptions, locale?: string): T extends number ? string : null;
+export function formatNumber<T extends number | null>(num?: T, precision?: number, options?: Intl.NumberFormatOptions, locale?: string) {
+  if (num == null) return null;
```
notbakaneko, line 37: I wonder if it's better to return an empty string instead, then interpolated strings would show as ` bar` instead of `null bar` (and component rendering already just renders nothing).

...or make `num` not nullable; none of the uses on the typescript side, at least, appear to be nullable

notbakaneko (changes requested):

resources/assets/lib/utils/html.ts
nanaya: The original idea is to allow `formatNumber(maybeNull) ?? 'some fallback text'` but I'm not sure if it's ever used like that 🤔

notbakaneko (approved):

PR #8784, opened 2022-04-08
New ruleset switcher style

nanaya:
Taken out from new beatmap page PR as it also applies for profile page. ~~Design is very not final~~ flyte says it's fine.

Note that discussion button on the beatmapset page cover area has been moved to the header.

Profile page

Beatmap page

Beatmap discussion (with disabled rulesets)

Mobile (hopefully there's always enough space)


notbakaneko (changes requested):
`titleAppend` can be removed from `HeaderV4`, maybe 🤔 

Sort of tangential but I wonder if this should use `currentUser.playmode` instead of page's user; also, updating default currently doesn't update `currentUser` https://github.com/ppy/osu-web/pull/8784/files#diff-a73b8c5f136ee1ff2ba7e1530e78080c0417ed116d65240e44385a78b01e3e15R50

Flade228:
no please

notbakaneko (approved):

PR #8790, opened 2022-04-11
Add channel type group headers to channel list 

notbakaneko:
And remove the group separators as part of the newer chat channel list design

### DIFF
--- a/resources/assets/less/bem/chat-conversation-list.less
+++ b/resources/assets/less/bem/chat-conversation-list.less
@@ -15,4 +15,16 @@
     width: 100vw;
     overflow-x: auto;
   }
+
+  &__header {
+    color: white;
+    margin: 10px 30px;
+    font-size: @font-size--normal;
+    text-transform: uppercase;
+
+    @media @mobile {
+      // FIXME: no idea how to display on mobile...
+      display: none;
+    }
+  }
 }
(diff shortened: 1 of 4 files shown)


resources/assets/lib/chat/conversation-list.tsx
```diff
@@ -15,6 +15,7 @@ export default class ConversationList extends React.Component {
       <div className='chat-conversation-list'>
         {supportedChannelTypes.map((type) => (
           <React.Fragment key={type}>
+            {this.renderHeader(type)}
             {this.renderChannels(core.dataStore.channelStore.groupedChannels[type])}
             {this.renderSeparator()}
```
nanaya, line 20: looks like those can be combined into one function? And do the length check first thing and return null if empty (also removes need for most of separator hiding logic css except the last-child).

resources/assets/less/bem/chat-conversation-list.less
```diff
@@ -15,4 +15,16 @@
     width: 100vw;
     overflow-x: auto;
   }
+
+  &__header {
+    color: white;
```
nanaya, line 20: use variable

nanaya (changes requested):

nanaya (approved):

PR #8792, opened 2022-04-12
Fix Typos on Downloads Page

RockRoller01:
someone i know found an typo and i ended up finding another two. I believe its now correct, but someone please double check as i am not a native english speaker

could use "any problems" instead of "a problem", no clue which one is preferred


peppy:
How has this gone unnoticed all this time?

peppy (approved):

PR #8794, opened 2022-04-13
Fixes for container-based Kubernetes deployments

ThePooN:
- Some Dockerfile folder permissions-related fixes
- Add support for container-based Octane deployments
- Add config_reloader deployment entrypoint for dynamic .env reloading
- Add schedule deployment entrypoint


docker/deployment/entrypoint.sh
```diff
@@ -6,15 +6,56 @@ if [ "$#" -gt 0 ]; then
...
 fi
 
+_octane() {
+  # exit on config update fail
+  set -e
+
+  /app/artisan config:cache
+  /app/artisan route:cache
+
+  /app/docker/deployment/entrypoint.sh config_reloader &
```
notbakaneko, line 16: I'm not sure the deployment version should be trying to watch and reload by default 🤔

nanaya: production probably shouldn't even use `.env` file to begin with...

(similarly for oauth key but notification server will need to be updated to be able to load from env variable as well)

ThePooN: why not? i'm using it to enable faster reloads in production. i can apply envs to containers directly instead but that will require container restarts for any env change.

@peppy what are your thoughts on the matter?

peppy: We don't change env often enough in production to require watching, I don't think. I would expect to have to restart the containers when changing env settings.

Keep it simple for now I guess (no watching required)

docker/deployment/entrypoint.sh
```diff
@@ -6,15 +6,56 @@ if [ "$#" -gt 0 ]; then
...
+  done
+}
+
 _php() {
     /app/artisan config:cache
     /app/artisan route:cache
     exec php-fpm8.0 -y /app/docker/deployment/php-fpm.conf
 }
 
+_schedule() {
```
nanaya, line 46: is there no existing schedule function on kubernetes or related management tool? This will have the box doing nothing but `sleep` most of the time.

docker/deployment/entrypoint.sh
```diff
@@ -6,15 +6,56 @@ if [ "$#" -gt 0 ]; then
...
+    exec /app/artisan schedule:run &
+    echo 'Sleeping for 5 minutes'
+  done
+}
+
 case "$command" in
     artisan) exec /app/artisan "$@";;
     assets) exec nginx -c /app/docker/deployment/nginx-assets.conf "$@";;
+    octane) _octane;;
     php) _php;;
```
nanaya, line 57: this the related fpm package/files should just be removed.

Dockerfile.deployment
```diff
@@ -7,6 +7,7 @@ RUN add-apt-repository ppa:ondrej/php
 
 RUN apt-get update
```
nanaya, line 8: `DEPLOYMENT.md` needs update

nanaya (changes requested):

docker/deployment/entrypoint.sh
ThePooN: There is, I am not actually using this in Kubernetes deploys. This exists because I went for a secondary docker-compose that was more production-like at the beginning, as I did not want to dive in blindly.

I will remove the loop and just call schedule:run.

Dockerfile.deployment
```diff
@@ -44,12 +43,8 @@ RUN composer install --no-autoloader --no-dev
 COPY package.json yarn.lock ./
 RUN yarnpkg
 
-RUN rm -f /var/log/nginx/access.log /var/log/nginx/error.log && \
```
nanaya, line 47: this and wrong `/var/lib/nginx` owner seem to cause this error? 🤔 

(the original lines were made so the logs are passed to docker, taken from [nginx official docker image](https://github.com/nginxinc/docker-nginx/blob/master/Dockerfile-debian.template#L88))
```
~/osu/osu-web-docker$ docker run nanaya/osu-web assets
nginx: [alert] could not open error log file: open() "/var/log/nginx/error.log" failed (13: Permission denied)
2022/04/27 03:58:06 [emerg] 1#1: mkdir() "/var/lib/nginx/body" failed (13: Permission denied)
```

nanaya (changes requested):

Dockerfile.deployment
ThePooN: the commit in which i removed that was completely unrelated :thinking:

ThePooN:
@nanaya @notbakaneko can either of you check this again?

Dockerfile.deployment
```diff
@@ -65,7 +64,7 @@ ARG GIT_SHA
 RUN printf "%s" "$GIT_SHA" > version
 
 RUN useradd -m osuweb
-RUN chown -R osuweb bootstrap/cache storage /var/lib/nginx
```
nanaya, line 68: `/var/lib/nginx` still has wrong permission

Dockerfile.deployment
```diff
@@ -65,7 +64,7 @@ ARG GIT_SHA
...
+RUN chown -R osuweb bootstrap/cache storage public/uploads public/uploads-avatar public/uploads-replay
```
nanaya, line 67: I don't think the upload directories should be writable at all. Those will be gone if the docker instance is stopped. And mounted volumes don't use the existing permission.

nanaya (changes requested):

Dockerfile.deployment
ThePooN: what purpose does chowning it to `osuweb` serves?

Dockerfile.deployment
ThePooN: i probably did it for a good reason. will double-check

docker/deployment/nginx-assets.conf
```diff
@@ -14,7 +14,7 @@ http {
     sendfile on;
 
     server {
-        listen 8000;
+        listen 8080;
```
nanaya, line 17: I just remembered, this should match the `EXPOSE` set in dockerfile.

Dockerfile.deployment
ThePooN: i focused on octane (i do not run nginx assets at all) so i may not have caught everything re: nginx and did not see any obvious sign of nginx needing to write to /var/lib/nginx

Dockerfile.deployment
ThePooN: was definitely a testing left-over, it's not needed. at least not under Kubernetes. can re-add later if needed. good catch

nanaya (approved):

PR #8799, opened 2022-04-14
👷 Add Docker push GitHub workflow

ThePooN:
Pushes to Docker Registry everytime a tag is added, pushing both `latest` and current git tag.

Credentials are already added in the GitHub organization.

### DIFF
--- a/.github/workflows/pack.yml
+++ b/.github/workflows/pack.yml
@@ -1,0 +1,32 @@
+name: Build and publish
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  push_to_registry:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      -
+        name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      -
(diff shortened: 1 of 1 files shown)


peppy (approved):

nanaya:
The required env variables seem missing. Check `DEPLOYMENT.md` for details.

.github/workflows/pack.yml
```diff
@@ -0,0 +1,32 @@
...
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      -
+        name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./Dockerfile.deployment
+          platforms: linux/amd64
+          push: true
+          tags: pppy/osu-web:latest,pppy/osu-web:${{ github.event.release.tag_name }}
```
nanaya, line 32: looks like this `github.event.release.tag_name` should be something else. It's currently blank:

```
/usr/bin/docker buildx build --file ./Dockerfile.deployment --iidfile /tmp/docker-build-push-XHNwHg/iidfile --platform linux/amd64 --tag pppy/osu-web:latest --tag pppy/osu-web: --metadata-file /tmp/docker-build-push-XHNwHg/metadata-file --push .
error: invalid tag "pppy/osu-web:": invalid reference format
Error: buildx failed with: error: invalid tag "pppy/osu-web:": invalid reference format
```

https://github.com/ppy/osu-web/runs/6077299095?check_suite_focus=true

notbakaneko: I think `release.tag_name` depends on the release workflow so this is always going to be empty if the workflow runs on every tag push.

PR #8814, opened 2022-04-18
Accept int for mods parameter

### LINKED ISSUE #8807: Beatmap attributes endpoint doesn't work with bitset number.
I get HTTP 422 {"error":"invalid mods specified"}, it does work if you wrap number in quotes making it a json string.

nanaya:
Fixes #8807.

### DIFF
--- a/app/Http/Controllers/BeatmapsController.php
+++ b/app/Http/Controllers/BeatmapsController.php
@@ -87,7 +87,7 @@ class BeatmapsController extends Controller
         }
 
         if (isset($params['mods'])) {
-            if (ctype_digit($params['mods'])) {
+            if (is_numeric($params['mods'])) {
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/BeatmapsController.php
```diff
@@ -87,7 +87,7 @@ public function attributes($id)
         }
 
         if (isset($params['mods'])) {
-            if (ctype_digit($params['mods'])) {
+            if (is_numeric($params['mods'])) {
```
notbakaneko, line 90: `1.1` is numeric too (but then gets truncated by the cast) 🤔

nanaya: weird people deserve weird response. I don't think the function throws error for extraneous bits either so negatives and invalid bits are still accepted regardless

notbakaneko (approved):

PR #8821, opened 2022-04-19
Fix viewing user modding history as guest

notbakaneko:


nanaya (approved):
`$user !== null && ($user->.. ?? $user->...)->...` is one less ternary but this works as well I guess

I've been seeing a lot of this pattern recently for some reason 🤔

PR #8822, opened 2022-04-19
Restrict removing beatmaps from loved to admin only

venix12:


Ephemeralis:
For context, this came up as a result of a discussion about current Project Loved privileges where it was identified that even a single account being compromised in the entire usergroup could conceivably cause a hostile actor to cause permanent score loss in Loved across dozens or even hundreds of mapsets.

It may have been appropriate to allow at-will unloving earlier in the project's life, but given the scope and scale it has reached these days, it isn't anymore.

cl8n:
should be no problem, but this doesn't seem necessary since the Project Loved group already forces tighter verification (always verify + password resets not allowed). iirc it was *because* of the Remove permission that the group got added to the extra verification thing

MChecaH:
Seems like when asked in the dev server, most people involved in the project are fine with the chance given the rationale.

However I do have to say something about the last paragraph:
> It may have been appropriate to allow **at-will unloving** earlier in the project's life, but given the scope and scale it has reached these days, it isn't anymore.

What is your definition of _"at-will"_? I might not be understanding it properly and I believe you were not trying to completely remove access for a mapper to unlove their map, but that's how I read it and want to clarify this. Mappers should still be given the right to unlove their maps at any time given a reason no matter the scale, in my honest opinion.

The reasons can just range from unloving a map to rank it, to just a player dissatisfied with their work not wanting it to be featured anymore. But given a reasonable reason, maps should still be able to be unloved.

peppy:
conceptually this is fine. loved team doesn't have the more stringent internal checks that other groups to, so this makes sense (and is a rare usage scenario anyway).

peppy (approved):

nanaya (approved):

PR #8829, opened 2022-04-21
Typescript beatmap header component

nanaya:
More or less straight conversion after refactoring that favourite popup thing.

- [x] #8827 
- [x] #8802
- [x] #8835


resources/assets/lib/beatmapsets-show/header.tsx
```diff
@@ -0,0 +1,357 @@
...
+        text={{
+          bottom: bottomTextKey == null ? undefined : osu.trans(`beatmapsets.show.details.download.${bottomTextKey}`),
+          top: osu.trans(`beatmapsets.show.details.download.${topTextKey}`),
+        }}
+      />
+    );
+  }
+
+  private readonly onClickFavourite = () => {
+    if (core.userLogin.showIfGuest(this.onClickFavourite)) return;
```
notbakaneko, line 238: does this even do anything if the button isn't even shown when guest?

notbakaneko (changes requested):
It seems the page now renderers before going to verification after signing in instead of just going to verification 🤔

nanaya:
looks like the rerender is a general existing bug of user data being set despite the page is getting a full reload (so the set is very much pointless).

notbakaneko (approved):

PR #8853, opened 2022-04-26
Change default discussion page to general all

nanaya:
Mainly so the <del>upcoming</del> banchobot post on older maps will always be the first visible thing when visiting the page but also probably makes more sense as mapper or moderators usually put note in there (maybe).


peppy:
Let's see what people say once this hits live.

peppy (approved):

PR #8862, opened 2022-04-27
Fix Docker build tagging

ThePooN:
Runs build on Git tag events. Tagging the Docker image using the Git SHA, the Git tag, and `latest` if the tag doesn't contain a hyphen.

### DIFF
--- a/.github/workflows/pack.yml
+++ b/.github/workflows/pack.yml
@@ -24,7 +24,6 @@ jobs:
           tags: |
             type=raw,value=latest,enable=${{ !contains(github.ref_name, '-') }}
             type=raw,value=${{ github.ref_name }}
-            type=sha
           flavor: |
             latest=false
       -
--- a/.github/workflows/pack.yml
+++ b/.github/workflows/pack.yml
@@ -22,7 +22,7 @@ jobs:
             pppy/osu-web
           # generate Docker tags based on the following events/attributes
           tags: |
-            type=raw,value=latest,enable=${{ !contains(github.ref_name, "-") }}
+            type=raw,value=latest,enable=${{ !contains(github.ref_name, '-') }}
             type=raw,pattern=${{ github.ref_name }}
             type=sha
           flavor: |
--- a/.github/workflows/pack.yml
+++ b/.github/workflows/pack.yml
@@ -23,7 +23,7 @@ jobs:
           # generate Docker tags based on the following events/attributes
           tags: |
             type=raw,value=latest,enable=${{ !contains(github.ref_name, '-') }}
-            type=raw,pattern=${{ github.ref_name }}
+            type=raw,value=${{ github.ref_name }}
             type=sha
           flavor: |
             latest=false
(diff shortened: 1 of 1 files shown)


peppy (approved):

nanaya (changes requested):
um,

```
Invalid workflow file

The workflow is not valid. .github/workflows/pack.yml (Line: 24, Col: 17): Unexpected symbol: '"-"'. Located at position 28 within expression: !contains(github.ref_name, "-")
```

https://github.com/nanaya/osu-web/actions/runs/2237529750/workflow

ThePooN:
fixed. settings page of my docker hub account return a blank page so i can't properly test, and a linter had no trouble parsing the file... :roll_eyes:

nanaya:
```
Error: Missing value attribute for type=raw,pattern=2022.428.2
```

https://github.com/nanaya/osu-web/runs/6207403801

.github/workflows/pack.yml
```diff
@@ -12,6 +12,21 @@ jobs:
...
+        uses: docker/metadata-action@v3
+        with:
+          # list of Docker images to use as base name for tags
+          images: |
+            pppy/osu-web
+          # generate Docker tags based on the following events/attributes
+          tags: |
+            type=raw,value=latest,enable=${{ !contains(github.ref_name, '-') }}
+            type=raw,value=${{ github.ref_name }}
+            type=sha
```
nanaya, line 27: do we need this when there's already the one with tag name?

ThePooN: not really, but also why not? seems like a good practice

nanaya: there's already another one whenever this is run that's unique (the tag/version) so this will just be image of same thing. Alternatively just append the hash to the version name

ThePooN: it's not pushing twice, just a tag as another way to reference the image (depending on the user's use case, don't have one atm).

not willing to append hash, trying to find the image for a specific version will be a pain otherwise.

nanaya: > it's not pushing twice, just a tag as another way to reference the image (depending on the user's use case, don't have one atm).

but it still fills up the listing with redundant images in no meaningful way since version and hash corresponds one to one - that's what git tag is after all (and we're not overriding tags afaik).

> not willing to append hash, referencing the image for a specific version will be a pain otherwise.

can use more specifics on this. At least it doesn't make any difference for looking up on docker hub.

ThePooN: docker image tag is automatically referenced by the Helm chart's "appVersion" field, would rather not add the git SHA to it.

nanaya: I'm not sure how that matter but fine, I guess. Also I just noticed the versioning (`GIT_SHA`) used for current deployment is `version-buildid?-hash` (something like `20220427-3411-74c50df`)

nanaya: The sentry release need some clean ups (namely sending actual commit hash) so it's fine for now.

nanaya (approved):

PR #8878, opened 2022-05-09
Update source for perfect combo score highlight

nanaya:
Comparing `combo` attribute with beatmap's `max_combo` attribute seems to be the most correct method but there's currently nowhere to put it in the json. Even score's `beatmap.max_combo` in score display page could be wrong because that attribute ignores the score mods combination.

### DIFF
--- a/resources/assets/lib/utils/score-helper.ts
+++ b/resources/assets/lib/utils/score-helper.ts
@@ -26,6 +26,19 @@ export function hasShow(score: SoloScoreJson): score is SoloScoreJson & Required
   return score.best_id != null;
 }
 
+export function isPerfect(score: SoloScoreJson) {
+  // TODO: Check against beatmap max_combo instead.
+  //       It's currently done this way because the beatmap data is sometimes
+  //       missing max_combo attribute for the score's mods combination.
+  return score.legacy_perfect ?? (
+    ([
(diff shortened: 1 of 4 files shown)


peppy:
Discussed this one IRL, but would appreciate a second check from @smoogipoo.

resources/assets/lib/utils/score-helper.ts
```diff
@@ -26,6 +26,19 @@ export function hasShow(score: SoloScoreJson): score is SoloScoreJson & Required
...
 
+export function isPerfect(score: SoloScoreJson) {
+  // TODO: Check against beatmap max_combo instead.
+  //       It's currently done this way because the beatmap data is sometimes
+  //       missing max_combo attribute for the score's mods combination.
+  return score.legacy_perfect ?? (
+    ([
+      'miss',
+      'large_tick_miss',
+      'small_tick_miss',
```
smoogipoo, line 37: Small ticks don't contribute to combo. Only miss + large ticks.

smoogipoo (changes requested):

smoogipoo (dismissed):

smoogipoo (approved):

notbakaneko (approved):

Issue #8866, opened 2022-04-29
Disable editing of beatmap descriptions after beatmap is ranked


peppy:
Permissions should be similar to guest mapper assignment. It should not be editable by the mapper after rank.

MkGuh:
I disagree. I don't know what cases of abuse ever happened from this, but instead of disallowing mappers to do that maybe you should improve the tools available for staff members to monitor description changes made in Ranked beatmaps by the mapper.

peppy:
Ah, this was not intended to be a discussion sorry, this is a work task and required from my end.

We have pinned comments these days which can be used to provide updates to maps (mappers can use these already).

Dada-jodel:
Surely the convenience and uses of editing a description while the map is ranked (updating art, tracking previous and future ranked maps in order, giving props to good plays, changing GD info such as names or icons when SR is updated) outweighs the extremely rare cases of abuse of ranked map descriptions which are already dealt with as soon as they are found by GMTs / NATs.

I see no reason to dismiss a discussion around this, given that anyone who's remotely familiar with the system as it stands figures it's a change that would make the system in question actively worse. Pinned comments are not looked at in the same way as descriptions as they are not visible upon opening the page unless scrolled to - and if the intention here is that if someone has to ask a moderator to change their description for them every time they wish to do so, then surely the burden of doing that is much, much greater than removing abuse cases (of which there were only a very limited amount). An analogous case to this would be something like having to parse every beatmap submission through a moderator to avoid the issue that pops up much more often, in comparison, of racist or political or otherwise innapropriate submissions being published. We can both agree that this isn't practical or, y'know, sensical in the slightest.

Please do not dismiss discussion when making a change like this in the future as it's bound to lead to cases like this - where pretty much everyone involved with this system immediately saw how bad this change was when it was brought up.

peppy:
In osu!, we aim for preserving all content. Allowing content to be edited after in a final state goes against this. I understand your opinions on the matter, but disagree with them from a fundamental perspective of how the ranking system and game works.

We've had cases of
- People replacing their ranked beatmap description with nothing, and having it lost forever (if they decide to quit or destroy content for whatever reason)
- People linking off-site downloads which contain differing content (which is generally not condoned in that description field)
- People bypassing DMCA takedowns and causing further headaches.

Your listed use cases of "updating mapper names" can be fixed by using user ID links (the standrad these days). Previous names are shown on profiles and the links will still work.

[...]

PR #8881, opened 2022-05-09
Disable description edit and comment on certain beatmaps

nanaya:
Related to #8866.

### DIFF
--- a/app/Libraries/OsuAuthorize.php
+++ b/app/Libraries/OsuAuthorize.php
@@ -1155,7 +1155,7 @@ class OsuAuthorize
         $commentable = $comment->commentable ?? $comment->parent?->commentable;
 
         if ($commentable instanceof Beatmapset && $commentable->downloadLimited()) {
-            return 'unauthorized';
+            return 'comment.store.disabled';
         }
 
         return 'ok';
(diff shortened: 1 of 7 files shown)


app/Libraries/OsuAuthorize.php
```diff
@@ -722,7 +722,7 @@ public function checkBeatmapsetDescriptionEdit(?User $user, Beatmapset $beatmaps
     {
         $this->ensureLoggedIn($user);
 
-        if ($user->user_id === $beatmapset->user_id || $user->isModerator()) {
+        if ((!$beatmapset->download_disabled && $user->user_id === $beatmapset->user_id) || $user->isModerator()) {
```
peppy, line 725: does this handle the case where `download_disabled` is `false` but `download_disabled_url` is non-null? if not, it should be (along with any similar usages). rationale is that most takedown requests these days are focused on specific files, not the whole beatmap content, and we leave the pieces that are still okay to host.

nanaya:
There's no indicator when comments are disabled (until submitted; which then it will show error message "Comments are disabled") but adding one requires quite a lot more changes so maybe later...

peppy (dismissed):

notbakaneko (changes requested):
Could do with some tests

notbakaneko (approved):

PR #8889, opened 2022-05-10
Immediately reload on login if no callback is set

nanaya:
What's the worst that can happen 🤷


notbakaneko (approved):
reloading early just skips everything that gets reset when reloading, anyway 🤔

PR #8906, opened 2022-05-16
Pass XSolla sandbox option when fetching token

nanaya:
Seems to make sense this way? 🤔 

Mainly to reduce need of passing env vars to js during build step.


notbakaneko (approved):
looks correct, sandbox seems to work 👀

PR #8907, opened 2022-05-16
Move shopify options to store page layout

nanaya:
Another one removing env vars from js.

...this should work, I think? 🤔


notbakaneko (approved):
seems to work 🤔

PR #8908, opened 2022-05-16
Fix wrong beatmap popup position on resize

nanaya:
Before I forget about this again.


notbakaneko (approved):
I feel like the width shouldn't have a transition, though 🤔

PR #8928, opened 2022-05-18
🐛 Docker: Fix missing htmlpurifier cache directory

ThePooN:


nanaya (dismissed):

nanaya:
Actually, I'll just fix the htmlpurifier itself instead. And I have a feeling only the top `storage` directory need to be created after this one is fixed.

nanaya:
Fixed in #8928.

PR #8931, opened 2022-05-19
Reuse html purifier instance and always create cache directory

nanaya:
Seems like good idea.

### DIFF
--- a/app/Libraries/CleanHTML.php
+++ b/app/Libraries/CleanHTML.php
@@ -5,6 +5,7 @@
 
 namespace App\Libraries;
 
+use ErrorException;
 use HTMLPurifier;
 use HTMLPurifier_Config;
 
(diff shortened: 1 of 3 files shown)


app/Libraries/CleanHTML.php
```diff
@@ -10,10 +10,17 @@
...
 class CleanHTML
 {
-    public static function purify($text)
+    private HTMLPurifier $purifier;
+
+    public function __construct()
     {
+        $cachePath = storage_path().'/htmlpurifier';
+        if (!is_dir($cachePath)) {
+            mkdir($cachePath, 0700, true);
```
notbakaneko, line 19: this isn't atomic and has a chance of the directory not existing during the check and existing at the `mkdir` call when starting octane

```
   ErrorException

  mkdir(): File exists

  at app/Libraries/CleanHTML.php:19
     15▕     public function __construct()
     16▕     {
     17▕         $cachePath = storage_path().'/htmlpurifier';
     18▕         if (!is_dir($cachePath)) {
  ➜  19▕             mkdir($cachePath, 0700, true);
     20▕         }
```

notbakaneko (changes requested):

notbakaneko (approved):

PR #8943, opened 2022-05-23
Clean up polymorphic class name handling

nanaya:
Flip map once, skip creating new object instance just to get the type, and simplify type lookup.

Also added a throw when trying to morph without mapping. I don't think there's still anything using full class name anymore?


notbakaneko (approved):
I thought I pressed the button already 🤔

PR #8946, opened 2022-05-24
Combine score ruleset function

nanaya:
The `gameModeString` cache is removed but hopefully it doesn't affect much...? Functions where it's repeatedly accessed have been updated to only do it once.

And it's no longer static function because it's only useful for the main legacy score classes under `App\Models\Score`; solo/multiplayer/legacy match score classes aren't split by ruleset.


notbakaneko (approved):
I guess this is fine since it's already in #8947 👀

PR #8948, opened 2022-05-24
Add score page for solo scores

nanaya:
There's nothing linking to this at the moment except maybe client. A part of this will also be needed to support pinning solo scores.

Depends on:

- [x] #8947 
- [x] #8946 
- [x] #8936


app/Http/Controllers/Solo/ScoresController.php
```diff
@@ -19,6 +19,35 @@ public function __construct()
         $this->middleware('auth');
     }
 
+    public function show($modeOrId, $legacyId = null)
```
notbakaneko, line 22: which route maps to this?

notbakaneko (changes requested):

notbakaneko (approved):

PR #8952, opened 2022-05-25
Fix pinned score visibility check

nanaya:
I don't remember why it was filtered by user visibility before when it's already specific to the user being viewed (or themselves).


notbakaneko (changes requested):
@peppy is there any case where a high score has `hidden` set that isn't a result of the user being restricted?

peppy:
There are not.

notbakaneko (approved):

PR #8957, opened 2022-05-26
Adjust spotlight badge colour

### LINKED ISSUE #8924: The "Spotlights" label should be coloured green to follow the colour scheme of the Rankings page
this is more of a suggestion rather that an "issue" i reckon, but

image

since the "Featured Artist" label is coloured blue (which matches the color of http://osu.ppy.sh/beatmaps and http://osu.ppy.sh/beatmaps/artists) shouldn't the "Spotlights" label be coloured green to match the color of http://osu.ppy.sh/rankings and http://osu.ppy.sh/rankings/osu/charts?

i feel like -- aside from the consistency aspect -- the pink colour should be reserved for tournament maps tho (if [#8436](http://github.com/ppy/osu-web/issues/8436) goes through) since tournaments are currently placed under http://osu.ppy.sh/community...

nanaya:
Resolves #8924.

Also adjusted featured artist badge colour and unify hover styling.


peppy (approved):

resources/assets/less/bem/beatmapset-badge.less
```diff
@@ -12,23 +12,29 @@
   white-space: nowrap;
   .link-plain();
 
+  --colour: hsl(var(--hsl-c1));
+  --colour-hover: var(--colour);
+  color: var(--colour);
+
+  .link-hover({
+    color: var(--colour-hover);
+  });
```
notbakaneko, line 21: this doesn't seem to do anything (and apparently isn't working before the change, either?) 🤔

notbakaneko (changes requested):

notbakaneko (approved):
hover thing was already broken, anyway 👀

nanaya:
it should work in beatmapset page. it's not linked in beatmapset panel so there's no hover effect there (pointer event is none iirc)

PR #8958, opened 2022-05-26
Show user list in announcement channel

### LINKED ISSUE #8950: Add view of chat channel participants
would be most useful for smaller types of channels like announcements or private groups(when that's a thing)

just noticed since I sent a lot of automated announcements that I now have a bunch of nearly identical channels with no way to see who the recipients were

notbakaneko:
(mostly) The relevant portions for rendering the list of users in announcement channels extracted from #8747.

closes #8950

- [x] confirming stuff

### DIFF
--- a/resources/assets/lib/chat/conversation-view.tsx
+++ b/resources/assets/lib/chat/conversation-view.tsx
@@ -5,11 +5,13 @@ import ShowMoreLink from 'components/show-more-link';
 import { Spinner } from 'components/spinner';
 import StringWithComponent from 'components/string-with-component';
 import UserAvatar from 'components/user-avatar';
+import UserCardBrick from 'components/user-card-brick';
 import { route } from 'laroute';
 import { each, isEmpty, last, throttle } from 'lodash';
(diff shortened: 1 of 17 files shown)


resources/assets/lib/chat/conversation-view.tsx
```diff
@@ -236,6 +239,24 @@ export default class ConversationView extends React.Component<Props> {
     );
   }
 
+  renderUsers() {
```
nanaya, line 242: this thing gets rather long if the channel contains a lot of people like in one of the most recent announcement (35 users).

also I wonder if it's supposed to be visible to everyone in the channel or only for announcers...

nanaya (changes requested):

resources/assets/lib/chat/conversation-view.tsx
notbakaneko: it's visible to everyone and this design is supposed to be temporary

nanaya (approved):

Issue #8972, opened 2022-05-30
Replace react-virtual-list


notbakaneko:
Needs to be replaced as it's using deprecated react functions that are being removed in the next version, and the project no longer seems to be updated.

PR #8980, opened 2022-06-02
Convert beatmapset SearchFilter to typescript

notbakaneko:
Another part of #8972

This one was fun ಠ_ಠ

Also normalizes the option keys to `string` for now to make it simpler to handle the querystring updates for now. 
There's still the bug where the display urls (but navigation is fine) on the options don't update correctly when the query changes, but that will probably need a separate PR to handle changes to the search initialization so the change can be observed by the component correctly.

The bug with the param not unsetting when the same option is clicked should be fixed, though. 🤔


resources/assets/lib/beatmapset-search-filters.ts
```diff
@@ -12,6 +50,7 @@ export type BeatmapsetSearchParams = {
 export type FilterKey = (typeof keyNames)[number];
 type filterValueType = string | null;
 
+
 export class BeatmapsetSearchFilters implements BeatmapsetSearchParams {
```
nanaya, line 54: the `BeatmapsetSearchParams` can probably be removed...

resources/assets/lib/beatmapset-search-filters.ts
```diff
@@ -40,65 +110,47 @@ export class BeatmapsetSearchFilters implements BeatmapsetSearchParams {
...
-
-    return BeatmapsetFilter.queryParamsFromFilters(values);
-  }
-
-  @computed
-  get searchSort() {
-    const [field, order] = this.displaySort.split('_');
-    return { field, order };
+  href(key: FilterKey, value: string | null) {
+    return updateQueryString(null, { [keyToChar[key]]: value });
```
nanaya, line 135: maybe add check to remove value if it's the default value? "Has Leaderboard" links to `s=leaderboard` but doesn't exist in the url when actually selected

resources/assets/lib/beatmaps/search-filter.tsx
```diff
@@ -0,0 +1,113 @@
...
+    if (this.props.name === 'general' && option.id === 'recommended' && this.controller.recommendedDifficulty != null) {
+      text += ` (${formatNumber(this.controller.recommendedDifficulty, 2)})`;
+    }
+    // FIXME: do an actual navigation
+    return (
+      <a
+        key={option.id}
+        className={cssClasses}
+        data-filter-value={option.id}
+        href={this.href(option.id)}
```
nanaya, line 98: it looks like `q` parameter sometimes not updated correctly? `q=...` is stuck with old value until the filter is selected

nanaya (changes requested):

resources/assets/lib/beatmaps/search-filter.tsx
notbakaneko: This is an existing bug that won't be fixed here as mention in the PR description

nanaya: looks like generating href using actual filters' `queryParams` is enough to fix it? Depending on the document href doesn't seem quite right anyway.

```typescript
// resources/assets/lib/beatmapset-search-filters.ts#135
return updateQueryString(null, { ...this.queryParams, [keyToChar[key]]: value });
```

resources/assets/lib/beatmaps/search-filter.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+interface Props {
+  multiselect: boolean;
+  name: FilterKey;
+  options: FilterOption[];
+  title?: string;
+}
+
+@observer
+export class SearchFilter extends React.PureComponent<Props> {
```
nanaya, line 21: forgot about this one but this probably shouldn't inherit PureComponent

nanaya: still need update on this, I think 👀

notbakaneko: weird, I though I did; maybe it was a different branch 🤔

nanaya (approved):

PR #8982, opened 2022-06-03
Use a different twemoji package

nanaya:
The svg is included this one so it should be more reliable compared to additional download in postinstall script.


notbakaneko (approved):
seems to work, essentially a rebase of twemoji and adding the svgs back into the distribution

PR #8986, opened 2022-06-05
Update email recovery process link in account verification dialog

### LINKED ISSUE #8969: 'email recovery process' link in account verification dialog is outdated
image

The current link redirects multiple times:

image

Should probably be changed to https://osu.ppy.sh/wiki/en/Help_centre/Account#sign-in

jhdcruz:
Fixes #8969

### DIFF
--- a/config/osu.php
+++ b/config/osu.php
@@ -207,7 +207,7 @@ return [
             'twitter' => '/wiki/Twitter',
         ],
         'user' => [
-            'recover' => '/p/forgot-email',
+            'recover' => '/wiki/Help_centre/Account#sign-in',
             'rules' => '/wiki/Osu!:Rules',
         ],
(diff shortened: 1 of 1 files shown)


peppy (approved):

The-Last-Cookie:
Will this also work for localised articles? German for example has [`#anmeldung`](https://osu.ppy.sh/wiki/de/Help_centre/Account#anmeldung) as the title for the `sign-in` section.

peppy:
I'd hope so.

nanaya:
The wiki page for other languages should be updated to have same id. The english one should probably also be hardcoded with the id so the link won't break if the title changes.

TicClick:
> The wiki page for other languages should be updated to have same id

I have that on my radar

PR #8989, opened 2022-06-06
Include per ruleset statistics for own profile endpoint

nanaya:
This actually only result in one more query because it removes two queries after adding three 🤔 

Related to ppy/osu#18517

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -454,7 +454,26 @@ class UsersController extends Controller
      */
     public function me($mode = null)
     {
-        return static::show(auth()->user()->user_id, $mode);
+        $user = auth()->user();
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/UsersController.php
```diff
@@ -448,13 +448,34 @@ public function scores($_userId, $type)
      *
      * See [Get User](#get-user).
      *
+     * Additionally, `statistics_fruits`, `statistics_mania`, `statistics_osu`, and `statistics_taiko` are also included.
```
frenzibyte, line 451: I think it's included in the form of `statistics_rulesets`? So something like:
```suggestion
     * Additionally, `statistics_rulesets` is included, containing statistics for all rulesets.
```

notbakaneko (approved):

PR #8997, opened 2022-06-08
Allow disabling comment on specific item

### LINKED ISSUE #8985: Allow moderators to lock beatmap comments sections
Some beatmap comments sections get out of hand very quickly. In some cases, they can be saved by aggressive moderation, but in some circumstances that is simply not possible nor productive to the volume of incoming messages, the scope of the provoking situation, or any other myriad of reasons that crop up when handing a multinational community.

Beatmap comments sections need to have the same lock functionality as modding discussions and forum threads do - toggle it active and no new comments are permitted until it is toggled off.

nanaya:
Resolves #8985 but only for admins with database access (or whatever other system) to actually disable commenting.

```sql
update osu_beatmapsets set comment_locked = 1 where beatmapset_id = $beatmapsetId;
```

This involves adding `comment_locked` column of type boolean/tinyint with default value of 0. Migration entry is

```
2022_06_08_065812_add_comment_locked_to_beatmapsets
```

in case manual migration is needed @peppy.

### DIFF
--- a/app/Http/Controllers/CommentsController.php
+++ b/app/Http/Controllers/CommentsController.php
@@ -194,7 +194,7 @@ class CommentsController extends Controller
         $comment = new Comment($params);
         $comment->setCommentable();
 
-        priv_check('CommentStore', $comment)->ensureCan();
+        priv_check('CommentStore', $comment->commentable)->ensureCan();
 
         try {
             $comment->saveOrExplode();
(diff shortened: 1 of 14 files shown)


peppy (approved):

notbakaneko (changes requested):
Replying to comment needs fixing, I think

notbakaneko (changes requested):
Also comments are disabled when there are no comments yet 🤔

notbakaneko (approved):

peppy:
I've run the migration.

peppy (approved):

PR #8999, opened 2022-06-08
Allow leaving announcement channel

### LINKED ISSUE #9046: Rejoin chat channel from notification
Seems useful 🤔

notbakaneko:
Receiving a message on the channel will cause the user to rejoin (unhide) the channel.

This doesn't cover the case where the channel creator leaves the channel and can't send from the web UI 👀 
~~Also, doesn't do anything about rejoining a channel from notification.~~ Now rejoins channel from notification; closes #9046.

Also tries to keep the selected channel in focus 👀 
...and makes the list header sticky so you can tell what group you're looking at.

- [x] confirmation prompt
- [x] #9050


peppy:
I think there should be confirmation when attempting to leave a notification channel, letting the user know they won't receive further updates on it. Some people may close it thinking it's a temporary close and will come back when further updates come on the same announcement thread (ie. existing behaviour).

peppy:
Interested in @cl8n / @Ephemeralis opinions on this one (and anyone else involved in the announcement flow). Should users be allowed to leave announcement threads?

notbakaneko:
Leaving an announcement channel still allows you to receive messages including making the channel visible again.

There's just no way to manually rejoin one without receiving a new message (problem if you're the announcer) or completely removing a user from the channel.

peppy:
Oh I see. So it will reappear on the next message arriving?

notbakaneko:
Yeah, it works the same way as pms

peppy:
All good then, I misunderstood the OP.

nanaya:
I think a confirmation before leaving announcement channel would be useful since unlike other channels one can't just rejoin after accidentally closing it

Ephemeralis:
I am very late to this, but yes, users should be allowed to leave announcement channels. Extended use of this system (and it *will* be used extensively) are going to cause spam hell over the course of a few years.

If anything, you may want to consider automatic expiry in rooms with no chat a la Discord threads? Something like 14 or 30 days.

app/Http/Controllers/ChatController.php
```diff
@@ -31,17 +31,29 @@ public function index()
...
+                $sendToJson = [
+                    'can_message_error' => ($channel?->checkCanMessage($user) ?? priv_check('ChatPmStart', $targetUser))->message(),
+                    'channel_id' => $channel?->getKey(),
+                    'target' => json_item($targetUser, 'UserCompact'),
+                ];
+            }
+        } elseif ($params['channel_id'] !== null) {
+            $channel = Channel::find($params['channel_id']);
+            if ($channel?->isHideable() ?? false) {
+                $channel->addUser($user);
```
nanaya, line 55: I think this is missing permission check? If anything, this should be queried from UserChannel instead so the `addUser` action will be limited to unhiding.

app/Models/Chat/Channel.php
```diff
@@ -555,18 +557,26 @@ public function validationErrorsTranslationPrefix()
...
 
-        return UserChannel::where([
-            'channel_id' => $this->channel_id,
-            'hidden' => true,
-        ])->update([
-            'hidden' => false,
-        ]);
+        $params = ['channel_id' => $this->channel_id, 'hidden' => true];
+
+        $users = User::whereIn('user_id', UserChannel::where($params)->select('user_id')->pluck('user_id'))->get();
```
nanaya, line 573: remove `pluck` so it's a subquery instead?

nanaya (changes requested):

app/Models/Chat/Channel.php
notbakaneko: Turns out I originally added the `pluck` because the `whereIn` didn't match anything with only subquery when running the tests 🤔

app/Models/Chat/Channel.php
```diff
@@ -454,12 +459,9 @@ public function receiveMessage(User $sender, ?string $content, bool $isAction =
                 $userChannel->markAsRead($message->message_id);
             }
 
-            if ($this->isPM()) {
-                if ($this->unhide()) {
-                    // assume a join event has to be sent if any channels need to need to be unhidden.
-                    (new ChatChannelEvent($this, $this->pmTargetFor($sender), 'join'))->broadcast();
-                }
+            $this->unhide();
```
nanaya, line 462: I just noticed the changes in `unhide` function may cause extra queries especially when doing batch sending.

nanaya (changes requested):

notbakaneko:
I'll do a final update on this after the other PRs get merged 👀

nanaya:
> I'll do a final update on this after the other PRs get merged 👀

And then it's merged. Also, conflicts.

nanaya (changes requested):

notbakaneko:
I forgot to push this, nice 🦆

nanaya (approved):

PR #9007, opened 2022-06-09
👷 Trigger automated staging deploy on tags push

ThePooN:
Added a job to trigger our staging deployment workflow on new builds.

Also split Sentry releasing into its own job, to avoid the whole workflow to fail even if build succeeds.

The deploy job requires the `KUBERNETES_CONFIG_REPO_ACCESS_TOKEN` secret; see https://github.com/marketplace/actions/repository-dispatch#token

Pending on further modifications depending on workflow validation in the internal repository.


ThePooN:
Merged and tested in internal repository, ready for review/merge now.

peppy (approved):

PR #9019, opened 2022-06-13
Add support for tournament ban infringement

### LINKED ISSUE #9018: Add profile infringement for tournament bans
Currently, tournament bans are tracked via internal notes which require support operators to manually disclose their status to tournament organizers (TOs) during screening. Out of an interest to not pass the responsibility of informing the general public of such a status to TOs, present practices involve informing the public via the osu! support twitter in a manner that can be easily construed as targeted.

Making active/permanent tournament ban status visible on user profiles as per the mockup below would help to alleviate this problem:

This would allow TO's to make quicker staffing decisions when making on-the-fly replacements and would also negate some need for non-supported tournaments to request screening at all.

Adding this into internal tool workflow for standard restriction appeal returns (aka, for noting the typical 1 year tournament ban that accompanies most standard appeals) will also be helpful.

venix12:
resolves #9018 (at least the web-side part)

temporary:

permanent:

also adds ``pernament`` bool column to ``osu_user_banhistory``, which can be also used for other infringement types :thinking:
- migration entry: ``2022_08_16_195413_add_permanent_to_user_account_history``


database/migrations/2022_06_13_195413_add_permanent_to_user_account_history.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('osu_user_banhistory', function (Blueprint $table) {
+            $table->boolean('permanent')->default(0);
```
cl8n, line 20: could use period=0 to mean permanent, instead of a new column

resources/assets/lib/interfaces/user-account-history-json.ts
```diff
@@ -8,7 +8,8 @@ export default interface UserAccountHistoryJson {
   description: string;
   id: number;
   length: number;
+  permanent: boolean;
   supporting_url?: string;
   timestamp: string;
-  type: 'note' | 'restriction' | 'silence';
+  type: 'note' | 'restriction' | 'silence' | 'tournament_ban';
```
cl8n, line 14: also update usercompact-useraccounthistory in api docs

resources/assets/lib/profile-page/account-standing.tsx
```diff
@@ -52,7 +52,7 @@ const ColumnDescription = ({ history }: ColumnProps) => (
 );
 
 const ColumnLength = ({ history }: ColumnProps) => {
-  if (history.type === 'restriction') {
+  if (history.type === 'restriction' || history.permanent === true) {
```
cl8n, line 55:
```suggestion
  if (history.type === 'restriction' || history.permanent) {
```

resources/assets/lib/profile-page/account-standing.tsx
```diff
@@ -24,7 +24,7 @@ interface ColumnProps {
 }
 
 const ColumnAction = ({ history }: ColumnProps) => (
-  <div className={`${bn}__action ${bn}__action--${history.type}`}>
+  <div className={`${bn}__action ${bn}__action--${history.type.replace(/_/g, '-')}`}>
```
cl8n, line 27: use `kebabCase` from lodash

cl8n (changes requested):

resources/assets/lib/profile-page/account-standing.tsx
nanaya: probably fine as is

app/Models/UserAccountHistory.php
```diff
@@ -68,18 +71,19 @@ public static function logUserUpdateEmail($user, $previousEmail)
...
     {
-        return $query->where('ban_status', static::TYPES['silence']);
+        return $query->whereIn('ban_status', [static::TYPES['silence'], static::TYPES['tournament_ban']]);
     }
 
     public function scopeRecent($query)
     {
         return $query
             ->where('timestamp', '>', Carbon::now()->subDays(config('osu.user.ban_persist_days')))
+            ->orWhere('permanent', true)
```
nanaya, line 86: this may cause rare problem when someone has hundreds of account history entries because their whole history will have to be checked. I'm not quite sure how much impact it'll have but it's probably/hopefully low enough it may be fine.

Otherwise some fancy indexing setup will be required.

database/migrations/2022_06_13_195413_add_permanent_to_user_account_history.php
```diff
@@ -0,0 +1,35 @@
+<?php
```
nanaya, line 1: filename need to be updated to more recent date

nanaya (changes requested):

database/migrations/2022_06_13_195413_add_permanent_to_user_account_history.php
peppy: The question is: what goes in the `period` column when `permanent` is set to `1`?

app/Models/UserAccountHistory.php
peppy:
```sql
mysql> select count(ban_id) from osu_user_banhistory group by user_id order by count(ban_id) desc limit 10;
+---------------+
| count(ban_id) |
+---------------+
|          3324 |
|          1650 |
|          1512 |
|           889 |
|           885 |
|           804 |
|           661 |
|           551 |
|           444 |
|           395 |
+---------------+
```

Possibly a valid concern? Not sure. Could likely be enough to add a new index on `(user_id,permanent)` if we go with that.

database/migrations/2022_06_13_195413_add_permanent_to_user_account_history.php
nanaya: 0 just like note? Thing is `note`s have period of 0 and they're not visible after `BAN_PERSIST_DAYS`... Unless we're changing it to be always visible as well.

peppy: Aha, if there's already a standard for it then that's fine 👍

app/Models/UserAccountHistory.php
nanaya: That doesn't work because the query is an `or` between `permanent` and `timestamp` (and then ordered by `timestamp`).

peppy: Hmm.. usually mysql will do a collate index usage, but I guess it doesn't bother due to the final `ORDER BY`? (it might also change depending on the number of rows..)

If that's the case it's probably not worth worrying about until it becomes an issue... although it's one of those issue we will likely never notice until it's so bad it takes things down.

nanaya (approved):

PR #9021, opened 2022-06-14
Pin important wiki notices to the top of the page

cl8n:
so ppl will still see them when they open a wiki page from a section link, or just scroll past the notices at the beginning

looks a little strange when other not-pinned notices scroll under it, but I'm not sure if anything should be done about that

### DIFF
--- a/resources/assets/less/bem/wiki-notice.less
+++ b/resources/assets/less/bem/wiki-notice.less
@@ -27,15 +27,7 @@
   }
 
   &__pinned-container {
+    .sticky-below-navbar();
     background-color: hsl(var(--hsl-b5));
-    position: sticky;
-    --navbar-height: @navbar-height;
-    // 1px overlap with navbar to prevent possible gap in some cases.
-    // Reference: https://bugs.chromium.org/p/chromium/issues/detail?id=1261852
-    top: calc(var(--navbar-height) - 1px);
-
-    @media @desktop {
-      --navbar-height: @nav2-height--pinned;
-    }
   }
 }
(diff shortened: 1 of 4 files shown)


nanaya:
This covers section title (and some part of the content) when using table of content or section link...

cl8n:
that's true... im not sure how to fix, would the best way be to get the height of this container and add it to scroll-padding/scroll-margin? or is there some way i can make this work without js :thinking:

resources/assets/less/bem/wiki-notice.less
```diff
@@ -25,4 +25,9 @@
       margin: 0;
     }
   }
+
+  &__pinned-container {
```
nanaya, line 29: I feel in some cases this may end up annoying people with this being permanently visible and reducing actual content area when reading the page (especially on mobile)

nanaya (changes requested):

resources/assets/less/bem/wiki-notice.less
cl8n: yeah... talked about it with TicClick, who originally suggested idea of this PR, and we both aren't sure how to make it better. I was thinking it could be good to add a dismiss option that puts it back into its normal position, but I don't know if that's actually less "annoying", might feel like a cookie popup where ppl instantly click+close by habit

Walavouchey: i'm not sure if this should be necessary at all... maybe the least intrusive solution could be a dismissable cookie popup only for when arriving from a section link and the original notice isn't visible (and it auto-dismisses if you scroll up to the notice)

nanaya:
I wonder if a floating `(!)` icon at a corner with relevant tooltip would work

PR #9022, opened 2022-06-14
Make entire linkable portion of notification a link 

notbakaneko:
Entire area now a clickable link instead of being this weird thing that doesn't show the link and clicks on the link portion for you.

Also fixes this hard to see `border-radius` issue on the cover image (right side should not be rounded)
image


resources/assets/lib/notification-widget/item.tsx
```diff
@@ -8,8 +8,7 @@ import { NotificationContext } from 'notifications-context';
 import NotificationDeleteButton from 'notifications/notification-delete-button';
```
nanaya, line 8: timestamp is missing hover tooltip

resources/assets/less/bem/notification-popup-item.less
```diff
@@ -82,11 +82,16 @@
...
+  &__link {
+    .full-size();
+  }
+
   &__main {
     border-radius: 0 @border-radius-large @border-radius-large 0;
     flex: 1;
     display: flex;
     min-width: 0;
+    pointer-events: none;
```
nanaya, line 94: `__unread-stripe` needs one as well (the one that appears on notifications history page). Or maybe wrap the whole thing (cover, main, stripe) in another div

nanaya (changes requested):

resources/assets/lib/notification-widget/item.tsx
notbakaneko: side note, I think the timestamp hover on user cards is also missing 👀

resources/assets/less/bem/notification-popup-item.less
```diff
@@ -110,15 +117,16 @@
     }
 
     &--time {
+      pointer-events: initial;
```
nanaya, line 120: This should be just on the span, not the whole div (as it covers the whole width). Maybe with `u-hover` on the span instead.

(also it can still be a link if it's an `<a>` 👀 although that'll require additional styling

nanaya (changes requested):

nanaya (approved):

PR #9023, opened 2022-06-14
Use docker for elasticsearch and notification servers for testing

nanaya:
Seems working? 🤔 Should also eventually look into using octane instead of `artisan serve`.

The yaml file gets a bit noisy with the private key but it's not too long (only 1024 bits) so hopefully doesn't cause too much annoyance.


.github/workflows/tests.yml
```diff
@@ -48,6 +66,32 @@ jobs:
           -e MYSQL_ALLOW_EMPTY_PASSWORD=1
           --entrypoint sh mysql:8.0 -c "exec docker-entrypoint.sh mysqld --default-authentication-plugin=mysql_native_password"
 
+      elasticsearch:
+        env:
+          action.auto_create_index: "false"
+          discovery.type: single-node
+        image: elasticsearch:7.17.4
```
notbakaneko, line 73: This should be `6.8.x` to match production or be a matrix of versions

notbakaneko (changes requested):

notbakaneko (approved):

PR #9025, opened 2022-06-15
Use exported mods data from client

### LINKED ISSUE #8945: Required and allowed mods should be validated for multiplayability
Currently, the only validation performed is a check to make sure the requested mods are generally valid:
https://github.com/ppy/osu-web/blob/ac94d7516f0978266188f8e731ea98cea3d7132b/app/Models/Multiplayer/PlaylistItem.php#L132-L145

This needs to be expanded to also check whether the mods are valid for multiplayer at all. To summarise:
- Mods that are invalid as "required" (i.e. room-wide) mods, are those which have a per-player speed effect. This is as opposed to `DoubleTime`, which equally speeds up all players in the room.
  - Example: https://github.com/ppy/osu/blob/master/osu.Game/Rulesets/Mods/ModAdaptiveSpeed.cs
- Mods that are invalid as "allowed" (i.e. per-player) mods, are those which increase speed. E.g. `DoubleTime`

For implementation, this should be implemented per-ruleset, and the validity of mods can be cross-referenced via the properties in mods from the osu! repo:
```
public override bool ValidForMultiplayer => false;
public override bool ValidForMultiplayerAsFreeMod => false;
```

nanaya:
Also resolves #8945.


app/Libraries/Mods.php
```diff
@@ -0,0 +1,211 @@
...
+    public array $idToBitsetMap;
+    public Set $difficultyReductionIds;
+    public array $mods = [];
+
+    public function __construct()
+    {
+        $this->idToBitsetMap = array_flip(static::LEGACY_BITSET);
+        $this->difficultyReductionIds = new Set();
+
+        $metadata = json_decode(file_get_contents(database_path('mods.json')), true);
```
peppy, line 60: Does this somehow get cached?

nanaya: yes, the class is initialised as a singleton

nanaya: yeah

notbakaneko (approved):

PR #9028, opened 2022-06-16
Replace react-virtual-list

### LINKED ISSUE #8972: Replace react-virtual-list

notbakaneko:
Updated virtual list that should work with newer version of React.

requires because it just continues off from #8995 and I don't want to extract the commits right now:
- [ ] #8995

closes #8972


resources/assets/lib/virtual-list/virtual-list.tsx
```diff
@@ -0,0 +1,108 @@
...
+  componentWillUnmount() {
+    this.scrollContainer.removeEventListener('scroll', this.throttledSetScroll);
+  }
+
+  render() {
+    const visibleItemBounds = this.visibleItemBounds;
+    const items = visibleItemBounds.lastItemIndex > 0 ? this.props.items.slice(visibleItemBounds.firstItemIndex, visibleItemBounds.lastItemIndex) : [];
+
+    const style = {
+      boxSizing: 'border-box' as React.CSSProperties['boxSizing'],
```
nanaya, line 92:
```suggestion
    const style: React.CSSProperties = {
      boxSizing: 'border-box',
```

resources/assets/lib/virtual-list/virtual-list.tsx
```diff
@@ -0,0 +1,108 @@
...
+@observer
+export default class VirtualList<T> extends React.Component<Props<T>> {
+  static readonly defaultProps = {
+    itemBuffer: 0,
+  };
+
+  private readonly ref = React.createRef<HTMLDivElement>();
+  private readonly scrollContainer = window; // we only care about window for now.
+  @observable private scrollTop = 0;
+  private readonly throttledSetScroll = throttle(() => this.setScroll(), 10);
```
nanaya, line 44: the 10ms throttle seems to be only useful on systems with high refresh rate 🤔 With enough item buffer, even 33ms is probably fine...?

nanaya (changes requested):

resources/assets/lib/virtual-list/virtual-list.tsx
notbakaneko: 60fps is 16.6ms; it's also more noticeable when scrolling fast with the scroll wheel.
Larger buffer means more elements for React to render and if you scroll fast enough, everything gets repainted, anyway, so the buffer has minimal effect beyond scrolling a certain speed.
`scroll` events in current browsers are already supposed to be throttled at the same rate as requestAnimationFrame, anyway

nanaya: well, what I'm saying is the throttle isn't doing much (if at all) on 60hz displays. Even on 120hz it's still only throttling very little (10ms vs 8ms).

notbakaneko: The throttle is there to cap unlimited fps scenarios, otherwise it works fine without any throttle at all.

nanaya (approved):

PR #9039, opened 2022-06-20
Convert beatmap discussion post user card to typescript

notbakaneko:
Starting with the simpler beatmap discussion components...


resources/assets/lib/beatmap-discussions/user-card.tsx
```diff
@@ -0,0 +1,76 @@
...
+            {this.props.user.is_deleted ? (
+              <span className={`${bn}__user-text u-ellipsis-overflow`}>
+                {this.props.user.username}
+              </span>
+            ) : (
+              <a
+                className={`${bn}__user-link`}
+                href={route('users.show', { user: this.props.user.id })}
+              >
+                {this.props.user.username}
```
nanaya, line 53: missing `u-ellipsis-overflow` which apparently stopped working sometime when the parent stopped being a flex container...

also missing `__user-text` so the styling is rather wrong (although now I looked closer the text and hover underline have different colour 👀 maybe fix that as well while at it).

nanaya (changes requested):

resources/assets/less/bem/beatmap-discussion-user-card.less
```diff
@@ -50,7 +50,7 @@
   }
 
   &__user-link {
-    color: white;
```
nanaya, line 53: more like this isn't thorough enough - it should have correct hover color as well otherwise it ends up with white text and colored underline on hover

also I don't know what purpose it serves for the avatar link.

And then might as well move the whole `__user-text` styling here and remove it (leaving just the `u-ellipsis-overflow` for the `span`)

nanaya (changes requested):

resources/assets/less/bem/beatmap-discussion-user-card.less
nanaya: ...or if the link isn't flex, it can have the `u-ellipsis-overflow` instead? I don't remember why it's structured like that (is it because the link originally covered the whole card?)

notbakaneko: > white text and colored underline on hover
which isn't a change from the current styling?

Then there's all the different combinations currently:
- this; white text, coloured underline, no card
- `mapped by x`; white text, no change on hover, has card
- links in the nomination box; coloured text, underline on hover, has card
- scoreboard; white text, colour change + underline on hover, has card
👀

nanaya: >  which isn't a change from the current styling?

I did say "maybe fix that as well while at it" in the previous comment... Text and underline having different colour doesn't make much sense.

nanaya (approved):

PR #9047, opened 2022-06-22
Don't requery when showing profile for logged in user

nanaya:


app/Libraries/User/FindForProfilePage.php
```diff
@@ -36,4 +36,27 @@ public static function find($id, ?string $type = null, ?bool $assertCanonicalId
...
+        $userId = (string) $user->getKey();
+        switch ($type) {
+            case 'id':
+                $isSelf = $id === $userId;
+                break;
+            case 'username':
+                $isSelf = $id === $user->username;
+                break;
+            default:
+                $isSelf = $id === $userId || $id === $user->username;
```
notbakaneko, line 57: Wouldn't this match if someone else's username is the same as your id or something equally silly?

notbakaneko (changes requested):

notbakaneko (approved):

PR #9050, opened 2022-06-22
Remove need for client to be in a channel when receiving messages targeted at the user

notbakaneko:
Since PMs and announcement messages are targeted at specific users, it shouldn't be required for them to be specifically "in" the channel (since they're always in it, just hidden from the list). Eventually, PM and announcement messages will have no `join` event sent at all.

This should remove the need to account for varying latencies with the responses from php and the socket server.
Messages sent to public channels after leaving are still ignored (otherwise messages that were in-flight after leaving would cause the channel to reappear in the list).

- [x] #9108


resources/assets/lib/stores/channel-store.ts
```diff
@@ -232,12 +232,16 @@ export default class ChannelStore implements DispatchListener {
...
-  private handleChatMessageNewEvent(event: MessageNewEvent) {
+  private async handleChatMessageNewEvent(event: MessageNewEvent) {
     for (const message of event.json.messages) {
       const channel = this.channels.get(message.channel_id);
-      if (channel == null) continue;
 
-      channel.addMessage(message);
+      if (channel != null) {
+        channel.addMessage(message);
+      } else if ((event.json.type === 'PM' || event.json.type === 'ANNOUNCE')) {
```
nanaya, line 241: extra parentheses aside, with `getChannel` returning 403 for unjoined public channel, the type check can probably be skipped (and check status code instead)? This will also allow the channel to appear if the user joined it in game without having to refresh the page

notbakaneko: The check is to avoid making multiple `getChannel` requests when receiving messages from public channels after leaving them

nanaya: That sounds like a same problem if multiple pm/announcement messages arrive after leaving them 🤔

resources/assets/lib/stores/channel-store.ts
```diff
@@ -196,13 +214,26 @@ export default class ChannelStore implements DispatchListener {
     }
   }
 
+  private cleanupIgnoredChannels() {
+    const now = new Date();
```
nanaya, line 218: wouldn't it be better to remove the entry on `apiPartChannel` done callback instead of having this clean up?

otherwise at least this should check for map length first before creating new date object.

resources/assets/lib/stores/channel-store.ts
```diff
@@ -49,11 +49,20 @@ const channelSorts = {
   PUBLIC: alphabeticalSort,
 };
 
+const hideableChannelTypes: Set<ChannelType> = new Set(['ANNOUNCE', 'PM']);
```
nanaya, line 52: this can be part of the Channel model itself?

nanaya (changes requested):

resources/assets/lib/stores/channel-store.ts
```diff
@@ -196,13 +214,26 @@ export default class ChannelStore implements DispatchListener {
...
+  private async handleChatMessageNewEvent(event: MessageNewEvent) {
     for (const message of event.json.messages) {
       const channel = this.channels.get(message.channel_id);
-      if (channel == null) continue;
 
-      channel.addMessage(message);
+      if (channel != null) {
+        channel.addMessage(message);
+      } else if (!this.ignoredChannels.has(message.channel_id)) {
+        const json = await getChannel(message.channel_id);
```
nanaya, line 234: this can use an error handler

resources/assets/lib/stores/channel-store.ts
notbakaneko: Messages sent to public channels sent after the you leave but before the request finishes processing and arrive after the request finishes processing due to latency, should not cause the channel to reappear in the channel list. 

The whole temporarily tracking channels to ignore messages from, is essentially a workaround for things that aren't currently sending the necessary `chat.channel.join` event. The value of `ignoreInterval` is such that it's beyond any reasonable latency.

nanaya: > Messages sent to public channels sent after the you leave but before the request finishes processing and arrive after the request finishes processing due to latency, should not cause the channel to reappear in the channel list.

This should fail on the `getChannel` request though?

notbakaneko: Also, only the tab where the leaving is initiated calls `apiPartChannel` 👀

nanaya: Well, no other tab is getting the ignored channel list anyway either

notbakaneko: Other tabs will get the part channel event relayed to them and ignore the channels.
Sending the join event is the most reliable method of determining whether a message should be dropped or not

nanaya: I'm not sure I follow here. I'm only suggesting instead of `cleanupIgnoredChannels()`, do `apiPartChannel(channelId, core.currentUserOrFail.id).done(() => this.ignoredChannels.delete(channelId))`.

Any messages coming through when it's in process should still be ignored and any excess messages after the request is completed should also still be ignored by 403 error when getting the channel info.

resources/assets/lib/stores/channel-store.ts
```diff
@@ -200,9 +218,15 @@ export default class ChannelStore implements DispatchListener {
   private handleChatMessageNewEvent(event: MessageNewEvent) {
     for (const message of event.json.messages) {
       const channel = this.channels.get(message.channel_id);
-      if (channel == null) continue;
 
-      channel.addMessage(message);
+      if (channel != null) {
+        channel.addMessage(message);
+      } else if (!this.ignoredChannels.has(message.channel_id)) {
+        getChannel(message.channel_id)
```
nanaya, line 225: because the check for this endpoint is using auth check instead of userChannel existence, this causes weird behavior for admin account 🫠

resources/assets/lib/stores/channel-store.ts
```diff
@@ -200,9 +218,15 @@ export default class ChannelStore implements DispatchListener {
...
+          .done((json) => this.update(json))
+          // FIXME: this seems like the wrong place to tigger an error popup.
```
nanaya, line 227:
```suggestion
          // FIXME: this seems like the wrong place to trigger an error popup.
```

also probably should ignore 404/403 errors for when the user isn't in the channel anymore

nanaya (changes requested):

nanaya (approved):

PR #9055, opened 2022-06-23
Fix forum navigation

### LINKED ISSUE #9053: Opening 'go to last post' breaks pagination if opened in new tab
x09KqrW24r

You can no longer go up, any posts but the last 20 disappear as well as the "load more" button thingy

nanaya:
Resolves #9053.


notbakaneko (changes requested):
This almost always goes to something like `50/68` for me if opening the link in a new tab; if I just load the page with the link itself, it goes to `n=51` first before going to the end 🤔

notbakaneko (approved):

PR #9056, opened 2022-06-23
Fix review editor deletion behaviour when empty

notbakaneko:
Prevents the last node in the editor from being deleted, preventing the internal cursor from going out of bounds and causing weird behaviour (which gets worse with IMEs)

This doesn't fix the desync between the DOM cursor and slate cursor when using an IME or any other weird IME-related behaviours


nanaya:
Can use the exact reproduction steps this pr is fixing...

notbakaneko:
correct non-buggy method in #9062

PR #9061, opened 2022-06-27
Create legacy score id mapping table

### LINKED ISSUE #9060: `solo_scores_legacy_id_map` table migration is missing
osu-queue-score-statistics is untestable on the `solo_scores` tables without this (can't import high scores).

nanaya:
Resolves #9060 (nothing is using it at the moment).

@peppy the existing table has wrong type for `score_id`. It should be `unsigned bigint` but currently it's just `bigint`. That said, practically it shouldn't be a problem? And migration name for this is `2022_06_27_073817_create_solo_scores_legacy_id_map`.


peppy:
I'll update it on production.

peppy (approved):

PR #9062, opened 2022-06-27
Update slatejs

### LINKED ISSUE #6097: In modding review, autocorrect mobile keyboards may duplicate text, make text input weird, and make the entire modding page blank.
Autocorrect mobile keyboards may duplicate text, make text input weird, and make the entire modding page blank. This issue will occur when using Google Keyboard for Android. Other similar autocorrect keyboards may be affected by these too.

I am using the Chinese Pinyin variant of Google Keyboard in the below videos attached. This is reproducible with Gboard - the Google Keyboard.

[Video: Weird text input when editing in-between numbers.](https://cdn.discordapp.com/attachments/423292757038071809/716920546151104554/Screenrecorder-2020-06-01-15-42-27-492.mp4)

[Video: Any kind of keyboard input mid-text will cause texts to duplicate.](https://cdn.discordapp.com/attachments/451253981616930816/717362463485591582/Screenrecorder-2020-06-01-15-52-33-312.mp4)

[Video: Holding backspace will cause the page disappear.](https://cdn.discordapp.com/attachments/451253981616930816/717358555153301604/Screenrecorder-2020-06-02-20-46-09-386.mp4)

### LINKED ISSUE #9057: Update slatejs
New version seems to have some really needed input and deletion fixes

notbakaneko:
- Updates to latest slate 
- fixes (most of) review editor typings
- corrects the method of resetting document and deleting last node.
- `Slate.value` only used for initial value now; slate is supposed to only be modified using `transforms` and `operations` or setting `children` and triggering `onChange()` directly
- typescript updated

depends on
- [x] #9064

Largely doesn't change declarations (e.g. `private`, `readonly`, etc) or cleanup logic, since this is mostly about updating slate and its usage.

closes #9057
maybe fixes #6097

Known issue where `scrollIntoView()` still causes the focus to still be under the tabs header; native `scrollIntoView()` makes use of `scroll-padding` but I'm not sure it there's a better way than using an arbitrarily large number (that isn't overly convoluted) 🤔 
Also seems like Safari will centre the input if slate doesn't trigger its `scrollSelectionIntoView`....


notbakaneko:
....or I could just have not pushed the commit 🥇

package.json
```diff
@@ -76,9 +76,9 @@
     "react-transition-group": "^4.4.2",
     "remark-parse": "^7.0.1",
     "shopify-buy": "^2.6.1",
-    "slate": "^0.58.4",
-    "slate-history": "^0.58.4",
-    "slate-react": "^0.58.4",
+    "slate": "^0.81.1",
```
nanaya, line 79: I don't know which thing causes it but entering 0 (or any number) scrolls the page to some amount until the input area is approximately at the center of the window ?_?

nanaya (changes requested):
not done but just found something weird

package.json
notbakaneko: For some inexplicable reason, slate appears to be updating the scroll position when the cursor selection is updated ಠ_ಠ
...and for some reason numbers make it offset more than letters ಠ_ಠ (you can see the scrolling if `position: sticky` is removed)

nanaya:
Conflicts...

resources/assets/lib/beatmap-discussions/editor.tsx
```diff
@@ -394,8 +414,17 @@ export default class Editor extends React.Component<Props, State> {
...
-    Transforms.deselect(this.slateEditor);
-    this.onChange(this.emptyDocTemplate);
+    this.resetEditorValue();
+  };
+
+  // Overrides slate Editable's defaultScrollSelectionIntoView and always use browser's version
+  // There's something wrong with slate's range detection in the prior calling function that causes
+  // letters and numbers to be differently.
+  scrollSelectionIntoView = (editor: ReactEditor, domRange: DOMRange) => {
+    domRange.startContainer.parentElement?.scrollIntoView({
```
nanaya, line 424: how does this work in safari?

nanaya (changes requested):

resources/assets/lib/beatmap-discussions/editor.tsx
notbakaneko: Safari supports `scrollIntoView(ScrollOptions)` since 15.4
https://developer.apple.com/documentation/safari-release-notes/safari-15_4-release-notes#Web-API

package.json
```diff
@@ -76,9 +76,9 @@
     "react-transition-group": "^4.4.2",
     "remark-parse": "^7.0.1",
     "shopify-buy": "^2.6.1",
-    "slate": "^0.58.4",
-    "slate-history": "^0.58.4",
-    "slate-react": "^0.58.4",
+    "slate": "^0.81.3",
```
nanaya, line 79: the latest is 0.82.1 btw 👀 and the `-react` one is 0.83.0

nanaya (approved):

PR #9065, opened 2022-06-29
Convert MessageLengthCounter to typescript

notbakaneko:
Should have done this one first 👀


resources/assets/lib/beatmap-discussions/message-length-counter.tsx
```diff
@@ -0,0 +1,27 @@
...
+  message: string;
+}
+
+export default function MessageLengthCounter({ message, isTimeline }: Props) {
+  if (!isTimeline) return null;
+
+  const maxLength = BeatmapDiscussionHelper.MAX_LENGTH_TIMELINE;
+
+  const modifiers = {
+    'almost-over': message.length > maxLength * 0.95,
```
nanaya, line 18: this only works (text being red when `>750` despite having this modifier) because the css is in correct order...

nanaya (changes requested):

nanaya (approved):

PR #9067, opened 2022-06-30
Cleanup global typescript declarations

notbakaneko:
What started out as moving the `BeatmapsetDiscussionJson` declarations turned into cleaning up the remaining global declarations


notbakaneko:
Beatmapset discussions response really need cleaning up ಠ_ಠ

nanaya (approved):

PR #9070, opened 2022-06-30
Combine elasticsearch config

nanaya:
Better than adding more files for each new connection. Also actually allow setting multiple hosts (space separated).

Looks like the file structure is a leftover from the now-unused laravel-elasticsearch package.


config/elasticsearch.php
```diff
@@ -1,12 +1,27 @@
...
     'connectionParams' => [
         'client' => [
             'timeout' => get_float(env('ES_CLIENT_TIMEOUT')) ?? 5,
             'connect_timeout' => get_float(env('ES_CLIENT_CONNECT_TIMEOUT')) ?? 0.5,
         ],
     ],
-    'connectionPool' => ['\Elasticsearch\ConnectionPool\SimpleConnectionPool'],
+    'connectionPool' => [SimpleConnectionPool::class],
+];
+$parseHosts = fn ($envName) => explode(' ', env($envName, 'localhost:9200'));
```
notbakaneko, line 16: `presence`?

`foo=` still returns `""` for `env`

notbakaneko (changes requested):

notbakaneko (approved):

PR #9072, opened 2022-06-30
Convert beatmapset discussions NewReply component to typescript

notbakaneko:
- [x] #9067


nanaya:
conflicts

nanaya (changes requested):

resources/assets/lib/beatmap-discussions/new-reply.tsx
```diff
@@ -0,0 +1,273 @@
...
+
+export class NewReply extends React.PureComponent<Props, State> {
+  state: Readonly<State> = {
+    editing: osu.present(this.storedMessage),
+    message: this.storedMessage,
+    posting: null,
+  };
+
+  private readonly box = React.createRef<HTMLTextAreaElement>();
+  private readonly handleKeyDown;
```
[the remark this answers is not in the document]
notbakaneko: turns out type doesn't need to be declared if assigned in constructor 🤔

resources/assets/lib/beatmap-discussions/new-reply.tsx
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  @action
+  private readonly editStart = () => {
+    if (core.userLogin.showIfGuest(this.editStart)) return;
+    this.editing = true;
+  };
+
+  private handleKeyDownCallback: TextAreaCallback = (type, event) => {
+    runInAction(() => {
```
notbakaneko, line 111: declaring `@action` isn't enough

nanaya: I think you just need to set `handleKeyDown` after `makeObservable` since that's when decorations are applied

notbakaneko: makes sense 🤔

nanaya: also missing readonly 👀

nanaya (approved):

PR #9076, opened 2022-07-01
Fix some mods being disallowed for playlist

nanaya:
I completely forgot about it when actually writing the code 👀


notbakaneko (commented):
should this handle the `UserPlayable` `false` cases?

nanaya:
Neither seem a breaking problem so maybe later. There's also configuration requirement which isn't checked at the moment either.

notbakaneko:
As far as I can tell, `ValidForMultiplayer: false` always has an associated `UserPlayable: false`, so non-realtime cases that were legitimately failing before would be passing now?

nanaya:
Anyone weird enough going through effort allowing cinema/autoplay mods probably already expect weird result in the first place.

peppy:
Let's deploy this early next week in case there's some new missed cases.

app/Libraries/Mods.php
```diff
@@ -88,6 +88,17 @@ public function __construct()
         }
     }
 
+    public function checkAllModsMeta(int $rulesetId, array $mods, callable $check, callable $failItemCallback): void
```
notbakaneko, line 91: Everything in this function is external to the class other than the call to a `public $this->mods`, it looks like it should go in `PlaylistItem`

notbakaneko (changes requested):

PR #9078, opened 2022-07-04
Move BeatmapsetTransformer includes

notbakaneko:
No need to `array_merge` now that we know `primitive` is a thing..or is there 🤔 
Seems a bit slower:

`array_merge`
```
>>> $s=microtime(true); $beatmapsets = Beatmapset::limit(100)->get(); for ($i = 0; $i < 100; $i++) { $r = json_collection($beatmapsets, new BeatmapsetTransformer()); };  microtime(true) - $s;
=> 1.0002069473267
=> 0.99510312080383
=> 1.002110004425
```

`includes`
```
>>> $s=microtime(true); $beatmapsets = Beatmapset::limit(100)->get(); for ($i = 0; $i < 100; $i++) { $r = json_collection($beatmapsets, new BeatmapsetTransformer()); };  microtime(true) - $s;
=> 1.2349479198456
=> 1.2817931175232
=> 1.2342507839203
```

### DIFF
--- a/app/Transformers/BeatmapCompactTransformer.php
+++ b/app/Transformers/BeatmapCompactTransformer.php
@@ -11,10 +11,28 @@ use App\Models\BeatmapFailtimes;
 class BeatmapCompactTransformer extends TransformerAbstract
 {
     protected $availableIncludes = [
+        'accuracy',
+        'ar',
         'beatmapset',
+        'bpm',
         'checksum',
+        'convert',
+        'count_circles',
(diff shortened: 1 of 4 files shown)


nanaya:
Maybe only move as needed? 🤔 And the typing will need updating if we're going ahead with this.

notbakaneko:
I was originally doing this to update the interfaces, before noticing `beatmapset-extended-json` still exists 🤔

PR #9082, opened 2022-07-05
Fix handling deleted model for indexing

nanaya:
Avoid error being thrown when trying to index thing that's been deleted from database. This assumes deleting only needs primary key.

Also there's no good way to handle composite primary key at the moment so those models are logged and then ignored (there shouldn't be any of such case).


notbakaneko (changes requested):
This is going to error on existing items in the queue 👀

nanaya:
ಠ_ಠ I guess a new job is in order

notbakaneko (approved):

PR #9084, opened 2022-07-05
Convert beatmapset discussions NewDiscussion component to typescript

### LINKED ISSUE #5105: Cannot praise directly with suggestions nearby
So if you intend to write a praise and link a timestamp where suggestions and issues are nearby you first have to confirm that "none of the posts address my concern" before praising whatever you wanted to praise

img

I dont think that makes a whole lot of sense because you didn't have a concern to begin with in that case

notbakaneko:
Largely skips logic fixes (`props.currentUser` looks like it could be removed, `beatmapset` probably isn't changing between renders, etc) - those will be done after other components are converted.

Split the hype and timestamp label renders so they don't try to shoehorn into each other ಠ_ಠ

closes #5105

- [x] #9072


resources/assets/lib/utils/beatmapset-discussion-helper.ts
```diff
@@ -54,7 +55,14 @@ export function badgeGroup({ beatmapset, currentBeatmap, discussion, user }: Bad
...
 
-function discussionLinkify(text: string) {
+export function canModeratePosts(user?: UserJson) {
+  user ??= core.currentUser;
+  if (user == null) return false;
+
+  return user.is_admin || user.is_moderator;
+}
+
+export function discussionLinkify(text: string) {
```
nanaya, line 65: ...doesn't need to be exported?

resources/assets/lib/beatmap-discussions/new-discussion.tsx
```diff
@@ -0,0 +1,517 @@
...
+  discussions: Discussion[];
+  timestamp: number | null;
+}
+
+interface Props {
+  autoFocus: boolean;
+  beatmapset: BeatmapsetExtendedJson & BeatmapsetWithDiscussionsJson;
+  currentBeatmap: BeatmapExtendedJson;
+  currentDiscussions: CurrentDiscussions;
+  currentUser: CurrentUserJson;
```
nanaya, line 77: there's a weird mixture of `this.props.currentUser` and `core.currentUser` going on here ಠ_ಠ use the latter for everything instead?

resources/assets/lib/beatmap-discussions/new-discussion.tsx
```diff
@@ -0,0 +1,517 @@
...
+      ? BeatmapDiscussionHelper.parseTimestamp(this.message)
+      : null;
+  }
+
+  private get validPost() {
+    if (!validMessageLength(this.message, this.isTimeline)) return false;
+
+    if (this.isTimeline) {
+      return this.timestamp != null && (this.nearbyDiscussions.length === 0 || this.timestampConfirmed);
+    }
```
nanaya, line 153: This is indeed keeping the current behavior but there's suggestion for different behavior for praise and note buttons (#5105)

resources/assets/lib/beatmap-discussions/new-discussion.tsx
```diff
@@ -0,0 +1,517 @@
...
+          </div>
+          <div className={`${bn}__content`}>
+            <div className={`${bn}__avatar`}>
+              <UserAvatar modifiers='full-rounded' user={this.props.currentUser} />
+            </div>
+            <div className={`${bn}__message`} id='new'>
+              {this.props.currentUser?.id != null ? (
+                <>
+                  <TextareaAutosize
+                    key='input'
```
nanaya, line 334: key not needed anymore (also for the other one)

nanaya (changes requested):

nanaya (approved):

PR #9086, opened 2022-07-06
Remove unused EsDeleteDocument job

nanaya:
Looking again apparently it was never used in the first place?


notbakaneko (approved):
I think this was added but then turned out that we wanted the deletes to be synchronous or they were already called inside jobs, anyway 🤔

PR #9097, opened 2022-07-08
Add `allow_performance` column to `osu_builds`

smoogipoo:
Not sure what the correct default value should be - I've gone with `false` for now but maybe it'll be `true` in the future when lazer is more "stable"?

Kinda required for https://github.com/ppy/osu-queue-score-statistics/pull/44.

### DIFF
--- a/database/migrations/2022_07_08_085310_add_allow_performance_to_osu_builds.php
+++ b/database/migrations/2022_07_08_085310_add_allow_performance_to_osu_builds.php
@@ -18,6 +18,7 @@ class AddAllowPerformanceToOsuBuilds extends Migration
     {
         Schema::table('osu_builds', function (Blueprint $table) {
             $table->boolean('allow_performance')->default(false);
+            $table->index('allow_performance');
         });
     }
 
@@ -30,6 +31,7 @@ class AddAllowPerformanceToOsuBuilds extends Migration
     {
         Schema::table('osu_builds', function (Blueprint $table) {
             $table->dropColumn('allow_performance');
+            $table->dropIndex('allow_performance');
         });
     }
 }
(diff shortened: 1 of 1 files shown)


peppy:
I think `false` is fine. I'd want to ensure that all legacy systems creating builds don't accidentally set it to `true`.

database/migrations/2022_07_08_085310_add_allow_performance_to_osu_builds.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('osu_builds', function (Blueprint $table) {
+            $table->boolean('allow_performance')->default(false);
```
peppy, line 20:
```suggestion
            $table->boolean('allow_performance')->default(false)->after('allow_bancho');
```

Just for my own sanity.

Also as discussed, let's add an index on this column.

peppy (changes requested):

database/migrations/2022_07_08_085310_add_allow_performance_to_osu_builds.php
```diff
@@ -30,6 +31,7 @@ public function down()
     {
         Schema::table('osu_builds', function (Blueprint $table) {
             $table->dropColumn('allow_performance');
+            $table->dropIndex('allow_performance');
```
peppy, line 34: Not sure if order matters but this feels backwards 🤔 
I guess laravel is smart enough to run correctly?

smoogipoo: :| I looked at other migrations in reference to this, but yes this does indeed fail. As should e.g. https://github.com/ppy/osu-web/blob/2ffc17cc6268898f1f9e19aa304436fcbaa9490c/database/migrations/2021_08_30_153214_add_track_id_to_beatmapsets.php#L30-L36

peppy: Good thing we don't rollback migrations.. ever 🙈

nanaya (approved):

PR #9104, opened 2022-07-09
Translate "Ranked" and "Loved"

CleverShadowx:
translate "ranked" and "loved" to Arabic in all(?) of their instances.


venix12:
Translations should be done via [crowdin](https://crowdin.com/project/osu-web).

PR #9110, opened 2022-07-13
Add endpoint to fetch solo scores from elasticsearch

nanaya:
Tests to be added and on production this depends on the scores to magically appear on elasticsearch 🪄


peppy:
Working flawlessly, although I have not checked the code. Would appreciate if this can be merged as a priority since I'll be wanting to keep it deployed (currently deployed from a temporary branch).

peppy (dismissed):

app/Libraries/Score/BeatmapScores.php
```diff
@@ -0,0 +1,122 @@
...
+            $nextCursor = $search->getSortCursor();
+            $hasNext = $nextCursor !== null;
+        }
+
+        return new Collection(array_values($this->result));
+    }
+
+    public function rank(?SoloScore $score): ?int
+    {
+        if ($score === null && $totalScore === null) {
```
notbakaneko, line 60: `$totalScore` from where 👀

notbakaneko (changes requested):
(still checking)

app/Libraries/Score/UserRank.php
```diff
@@ -0,0 +1,31 @@
...
+
+use App\Libraries\Search\ScoreSearch;
+use App\Libraries\Search\ScoreSearchParams;
+
+class UserRank
+{
+    public static function getRank(ScoreSearchParams $params): ?int
+    {
+        if ($params->beforeTotalScore === null && $params->beforeScore === null) {
+            return null;
```
notbakaneko, line 16: I can't help but think this should throw instead of returning `null` if not set.

app/Libraries/Score/BeatmapScores.php
```diff
@@ -0,0 +1,122 @@
...
+                break;
+            }
+            $nextCursor = $search->getSortCursor();
+            $hasNext = $nextCursor !== null;
+        }
+
+        return new Collection(array_values($this->result));
+    }
+
+    public function rank(?SoloScore $score): ?int
```
notbakaneko, line 58: Should this even be nullable? The only call currently does `isset` before calling

notbakaneko (changes requested):

notbakaneko (approved):

PR #9120, opened 2022-07-19
Set COMPRESSED row format on solo_scores

smoogipoo:
Something that's eventually going to come up as data.ppy.sh dumps start being generated for the new table. This table is a bit chonky.


peppy:
I've added a `migrations` entry for this.

peppy (approved):

PR #9122, opened 2022-07-19
Use InvariantException for score request parameter errors

nanaya:
Seems like a good thing to unify to ( ﾟ ヮﾟ)


notbakaneko (approved):
it occurs me the supporter only error message doesn't actually show 🤔

PR #9127, opened 2022-07-20
Add beatmap leader for solo scores

nanaya:
This only creates the table and a command to fill it up. `refresh` name is a bit of lie because it doesn't delete scores which beatmap have been unranked/disqualified...


notbakaneko (approved):

database/migrations/2022_07_27_000000_create_beatmap_leaders.php
```diff
@@ -0,0 +1,41 @@
+<?php
```
notbakaneko, line 1: wait, that date looks awfully far into the future 🤔

notbakaneko (changes requested):

notbakaneko (approved):

PR #9128, opened 2022-07-21
Fix solo scores nomod filter support

nanaya:
Selecting nomod should work as an "or": the score must be either the other selected mods or has no mods (ignoring preferences mods).

This matches the existing behaviour of legacy scoreboard.


app/Libraries/Search/ScoreSearch.php
```diff
@@ -95,4 +70,50 @@ public function getQuery(): BoolQuery
...
+
+    private function addModsFilter(BoolQuery $query): void
+    {
+        $mods = $this->params->mods;
+        if ($mods === null || count($mods) === 0) {
+            return;
+        }
+
+        $modsHelper = app('mods');
+        $allMods = new Set(array_keys($modsHelper->mods[$this->params->rulesetId]));
```
notbakaneko, line 82: just noticed 'rulesetId' should default to something or there should be some `null` check.

notbakaneko (changes requested):

app/Libraries/Search/ScoreSearch.php
nanaya: To be fixed by #9129

notbakaneko (approved):

PR #9129, opened 2022-07-21
Fix mods filter when no ruleset is set

nanaya:
Either this or require `rulesetId` to be set. Its main use case is for beatmap pack completion check. That said, the no-ruleset query is only possible if the convert state is indexed otherwise per-ruleset queries will be needed.

- [x] depends on #9128 (because things are moved around)


app/Libraries/Search/ScoreSearch.php
```diff
@@ -79,7 +79,9 @@ private function addModsFilter(BoolQuery $query): void
         }
 
         $modsHelper = app('mods');
-        $allMods = new Set(array_keys($modsHelper->mods[$this->params->rulesetId]));
+        $allMods = new Set(isset($this->params->rulesetId)
+            ? array_keys($modsHelper->mods[$this->params->rulesetId])
+            : array_merge(...array_map('array_keys', array_values($modsHelper->mods))));
```
notbakaneko, line 84: seems like something that can be memoized if only to give it a name instead of `array_merge(...array_map....array.....` (or have an `allModsForRuleset` function or something)

notbakaneko (changes requested):

notbakaneko (approved):

PR #9130, opened 2022-07-21
Fix setHash triggering on wrong page after navigation.

### LINKED ISSUE #9114: when returning to user page from beatmap page, wrong parameter are given
*notice: this was translated by DeepL*

open userpage (ex. [my user page](https://osu.ppy.sh/users/22136262)) and click the beatmap with least 2 diff from favrite beatmap,
next, make sure you have selected the highest diff for the beatmap you have chosen, then press the page back button,
then `#osu/beatmap_id` will be given to the url of the user page.

video: https://youtu.be/OC08V_Xjkwc

- worked with Microsoft Edge, Google Chrome
- no userscript available
- doesnt work without logging in to site

I think this bug is recently happened because it didnt happen until yesterday
is this my problem? does anyone else this working?

my user agent:
Microsoft Edge
`Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/103.0.5060.114 Safari/537.36 Edg/103.0.1264.62`
Google Chrome
`Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/103.0.0.0 Safari/537.36`

notbakaneko:
Seems like this isn't as new as it looks but the exact condition to trigger it might not be so common 🤔 

Because the component of the next page mounts before the current out unmount, `autorun` ends up triggering `setHash` if any of the observables it's watching gets updated.

It just so happens that current user `playmode` ends up being observed if via `findDefault()` through `currentBeatmap()`
if a beatmap hasn't explicitly been selected yet, which causes `autorun` to trigger when `core.setCurrentUser` is triggered on navigation.

A more complete fix would to be set the default beatmap when initializing `beatmapsets-show/controller` so that currentUser doesn't set sometimes observed.

closes #9114

### DIFF
--- a/resources/assets/lib/beatmapsets-show/main.tsx
+++ b/resources/assets/lib/beatmapsets-show/main.tsx
@@ -6,7 +6,7 @@ import { CommentsManager } from 'components/comments-manager';
 import HeaderV4 from 'components/header-v4';
 import PlaymodeTabs from 'components/playmode-tabs';
 import GameMode, { gameModes } from 'interfaces/game-mode';
-import { action, autorun, computed, makeObservable, observable } from 'mobx';
+import { action, autorun, computed, IReactionDisposer, makeObservable, observable } from 'mobx';
 import { observer } from 'mobx-react';
 import * as React from 'react';
 import { generate, setHash } from 'utils/beatmapset-page-hash';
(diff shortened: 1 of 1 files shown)


nanaya:
shouldn't this just dispose the listeners on `turbolinks:before-cache`? (in addition to on unmount although technically it only unmounts in that one case)

nanaya (approved):

PR #9133, opened 2022-07-22
Set max length on channel message input box

notbakaneko:
Limit max length so it doesn't get truncated by mysql.

Also increases the maximum height of the textbox in announcement channels to the same size as those in creating announcements, and anchors the textarea to the bottom.

Was going to add a message length counter but I figure that should be separate 🤔

### DIFF
--- a/resources/assets/lib/models/chat/message.ts
+++ b/resources/assets/lib/models/chat/message.ts
@@ -9,6 +9,8 @@ import * as moment from 'moment';
 import core from 'osu-core-singleton';
 import { linkify } from 'utils/url';
 
+export const maxLength = 1024;
+
 export default class Message {
   @observable channelId = -1;
   @observable content = '';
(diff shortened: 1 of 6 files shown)


resources/assets/lib/models/chat/message.ts
```diff
@@ -9,6 +9,8 @@ import * as moment from 'moment';
 import core from 'osu-core-singleton';
 import { linkify } from 'utils/url';
 
+export const maxLength = 1024;
```
nanaya, line 12: as it turned out there are two different max length on the backend; one is for announcement (no limit) and the other is limited to `config('osu.chat.message_length_limit')`:

https://github.com/ppy/osu-web/blob/master/app/Models/Chat/Channel.php#L400

nanaya (changes requested):

resources/assets/lib/models/chat/message.ts
nanaya: probably should also update the check there to account for column size. Something like `$lengthLimit = $this->isAnnouncement() ? 1024 : config(...)`.

nanaya (approved):

PR #9145, opened 2022-07-26
Truncate report text content sent to webhook

### LINKED ISSUE #9141: User report slack/discord integration needs description length limit
Either that or the report itself shouldn't allow message that long in the first place.

Sentry Issue: 5103

```
GuzzleHttp\Exception\ClientException: Client error: `POST <redacted>` resulted in a `400 Bad Request` response:
{"description": ["Must be 4096 or fewer in length."]}

  File "/artisan", line 37
    new Symfony\Component\Console\Output\ConsoleOutput
...
(48 additional frame(s) were not displayed)
```

notbakaneko:
1000 is still pretty long 🤔 

Also adds a hard limit to the comment text length...4000 is still pretty long 🤔 

closes #9141


nanaya:
Can it even be viewed through other means? Otherwise no one can view the rest of it. Might as well just set lower limit to begin with?

notbakaneko:
That's a good point 🤔

nanaya (approved):

PR #9146, opened 2022-07-27
Add tests for beatmap solo scores listing

nanaya:
The commands should probably be split up to their own PR but the tests also kind of implicitly (and only partially) test those commands.

A default command for the docker image would be nice so the workaround for github action service can be removed 👀


app/Console/Commands/EsIndexScoresQueue.php
```diff
@@ -0,0 +1,98 @@
...
+
+        $query = Score::select('id');
+
+        $validIds = false;
+        if ($this->option('all')) {
+            $validIds = true;
+        } else {
+            $ids = array_filter(array_map('get_int', explode(',', get_string($this->option('ids')) ?? '')), 'present');
+            if (count($ids) > 0) {
+                $query->orWhereIn('id', $ids);
```
notbakaneko, line 59: querying for scores when just pushing the id seems unnecessary 🤔

tests/Controllers/BeatmapsControllerSoloScoresTest.php
```diff
@@ -0,0 +1,233 @@
...
+            '--no-interaction' => true,
+        ]);
+        sleep(3);
+        (new ScoreSearch())->refresh();
+    }
+
+    public static function tearDownAfterClass(): void
+    {
+        (new static())->refreshApplication();
+        Beatmap::truncate();
```
notbakaneko, line 159: not really keen on `truncate` without some kind of sanity check 🤔

notbakaneko (changes requested):
Tests that need something else to be running probably should have some `@group` set so they can be excluded

nanaya:
The test suite already requires beatmap difficulty attribute server running to pass 👀

tests/Controllers/BeatmapsControllerSoloScoresTest.php
```diff
@@ -0,0 +1,233 @@
...
+                'settings' => [],
+            ],
+            $modNames,
+        );
+    }
+
+    /**
+     * @dataProvider dataProviderForTestQuery
+     */
+    public function testQuery(array $scoreKeys, array $params)
```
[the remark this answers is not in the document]
nanaya: I think this is the same requirement as having database/redis/etc running and setup.

tests/Controllers/BeatmapsControllerSoloScoresTest.php
```diff
@@ -0,0 +1,233 @@
...
+            ]),
+        ];
+
+        UserRelation::create([
+            'friend' => true,
+            'user_id' => static::$user->getKey(),
+            'zebra_id' => static::$scores['friend']->user_id,
+        ]);
+
+        Artisan::call('es:index-scores:queue', [
```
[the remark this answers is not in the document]
nanaya: it does give stack trace in case of coding failure at least

```
1) Tests\Controllers\BeatmapsControllerSoloScoresTest::testQuery
Error: Call to undefined function App\Console\Commands\asb() in /home/edho/osu/osu-web/app/Console/Commands/EsIndexScoresQueue.php:116
Stack trace:
#0 /home/edho/osu/osu-web/app/Console/Commands/EsIndexScoresQueue.php(83): App\Console\Commands\EsIndexScoresQueue->queueIds(Array)
...
```

tests/Controllers/BeatmapsControllerSoloScoresTest.php
nanaya: it's fine ( ﾟ ヮﾟ)

notbakaneko (approved):

PR #9147, opened 2022-07-27
Elasticsearch 8 support

notbakaneko:
Adds support for elasticsearch 8, compatibility added where needed to work with the existing 6.x servers.

- there is no longer a `_doc` type in the mapping schema.
- `number_of_replicas` in the schema file changed to 0; `number_of_replicas` can be changed online anyway.
- Bulk indexing requests in es8 no longer accepts `type`, 6 requires `type`, neither seem to need `type` when querying so there shouldn't need to be any changes here, except;
- the cross-index `terms` blacklist filter still needs `type` to function on es6. 

ES version | `ES_COMPATIBILITY`
--|--
6 | `true` (default)
7 | `true` or `false`
8 | `false`

The `high_scores` indices are not going to be upgraded from 6.x, so no changes there on this side.

This _doesn't_ use `ELASTIC_CLIENT_APIVERSIONING` as that changes the request header which 6.x doesn't support.

---
update: es6 support removed


nanaya:
What's the es upgrade plan here? Are we going to create a new es instance, reindex everything, and then redeploy pointing to the new instance?

Or if we're going to upgrade existing instance, we can just upgrade current one to 7, update the codes, and then finally upgrade to 8 (and there's no direct upgrade path to 8 anyway).

(and should we finally use clustered es so rolling upgrade is possible?)

peppy:
I think updating existing instances to 7 is probably the easiest plan (you're welcome to look at that if you beat me).

nanaya:
Production main es instance has been upgraded to 7 so the compat thing can be removed.

nanaya (approved):

PR #9160, opened 2022-08-02
Update mods data

nanaya:


database/mods.json
```diff
@@ -2631,4 +2650,3 @@
     ]
   }
 ]
-
```
notbakaneko, line 2634: this line keep coming and going ಠ_ಠ

peppy (approved):

database/mods.json
nanaya: ( ﾟ ヮﾟ)

PR #9162, opened 2022-08-02
Convert Beatmapset discussion system post to typescript

notbakaneko:
`message` returning `string` or `object` sounds like not such a good idea now 🤔 
I just picked the simplest option to handle the first post typing (skip if it's a system post)


resources/assets/lib/components/beatmapset-event.tsx
```diff
@@ -43,7 +43,8 @@ export default class BeatmapsetEvent extends React.PureComponent<Props> {
   }
 
   private get firstPost() {
-    return this.discussion?.starting_post ?? this.discussion?.posts?.[0];
+    const post = this.discussion?.starting_post ?? this.discussion?.posts?.[0];
+    return post?.system ? null : post;
```
nanaya, line 47: I wonder if `startingPost` should just always return `BeatmapsetDiscussionMessagePostJson` (but nullable?)

resources/assets/lib/interfaces/beatmapset-discussion-post-json.ts
```diff
@@ -14,11 +14,22 @@ interface BeatmapsetDiscussionPostDefaultAttributes {
...
-type BeatmapsetDiscussionPostJson = BeatmapsetDiscussionPostDefaultAttributes & Partial<BeatmapsetDiscussionPostAvailableIncludes>;
+type BeatmapsetDiscussionMessagePostJson = BeatmapsetDiscussionPostDefaultAttributes & {
+  message: string;
+  system: false;
+};
+
+type BeatmapsetDiscussionSystemPostJson = BeatmapsetDiscussionPostDefaultAttributes & {
+  message: {
+    type: string;
+    value: boolean;
```
nanaya, line 29: continuing the stronger typing for `type` field above, `boolean` is probably only applicable for `type: 'resolved'` anyway...

nanaya (changes requested):

resources/assets/lib/beatmap-discussions/system-post.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+import StringWithComponent from 'components/string-with-component';
+import BeatmapsetDiscussionPostJson from 'interfaces/beatmapset-discussion-post-json';
+import UserJson from 'interfaces/user-json';
+import { route } from 'laroute';
+import * as React from 'react';
+import { classWithModifiers } from 'utils/css';
+
+interface Props {
+  post: BeatmapsetDiscussionPostJson;
```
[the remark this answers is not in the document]
notbakaneko: none of these typing changes are going to be done until the `discussion` component is converted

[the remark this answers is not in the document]
nanaya: but why. The current component rendering this is already doing the check in the first place.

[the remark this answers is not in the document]
nanaya: also do you mean for other comments as well? The `BeatmapsetDiscussionSystemPostJson` typing is wrong regardless

resources/assets/lib/interfaces/beatmapset-discussion-post-json.ts
nanaya: there's not but there's no telling what type the value would be in other cases 🤔

notbakaneko: turns out `{ type: 'resolved'; value: boolean } | { type: string }` is not a thing (always resolves to `string`)

nanaya: ...of course it's not. The typing wouldn't make sense either (like, what is it typing?). I think it should just be `{ type: 'resolved'; value: boolean }` for now.

notbakaneko: that barely changes anything

nanaya: that's because there's only one at the moment but it will allow making sure nothing is missed if more type is added, similar to [user event](https://github.com/ppy/osu-web/blob/d56f0371595d07daf953d2929335017132b162b2/resources/assets/lib/profile-page/parse-event.tsx)

nanaya (approved):

PR #9163, opened 2022-08-03
Queue new scores for processing

nanaya:


app/Models/Solo/Score.php
```diff
@@ -127,6 +131,32 @@ public function getMode(): string
...
+
+        LaravelRedis::lpush(static::PROCESSING_QUEUE, json_encode([
+            'ProcessHistory' => null,
+            'Score' => [
+                'beatmap_id' => $dataJson['beatmap_id'],
+                'data' => json_encode($dataJson),
+                'id' => $dataJson['id'],
+                'user_id' => $dataJson['user_id'],
+            ],
+            'TotalRetries' => 0,
```
peppy, line 156: Not that it really matters, but I think this (and `ProcessHistory`) can be fully omitted.

peppy:
From my end, this can be merged whenever it's ready. Have done very preliminary testing and looks to work as expected.

app/Models/Solo/Score.php
```diff
@@ -127,6 +131,30 @@ public function getMode(): string
         return Beatmap::modeStr($this->ruleset_id);
     }
 
+    /**
+     * Queue the item for score processing
+     *
+     * @param array $dataJson Pre-generated json of the score so it doesn't need to be redundantly generated
+     * @param bool $force By default only newly created scores are queued. Pass true to force queuing the score
+     */
+    public function queueForProcessing(?array $dataJson = null, bool $force = false): void
```
notbakaneko, line 140: Is the nullable or flag being used somewhere?

notbakaneko (changes requested):

app/Models/Solo/Score.php
nanaya: this is a weird function and those are originally so its main function (queuing score for processing) can be accessed regardless of state and existence of pre-generated json...

notbakaneko (approved):

PR #9166, opened 2022-08-04
Reduce wait time for score indexing in test

nanaya:
The check is rather fragile (it doesn't check if the index content is actually correct) but I think it's good enough for testing purpose. Better than waiting 3s especially as more tests are added.


notbakaneko (changes requested):
conflicts 💥

notbakaneko (approved):

PR #9170, opened 2022-08-05
Remove unused soft delete for scores

nanaya:
Nothing is using it and there's `preseve` flag already for indexing...


peppy (approved):

peppy:
I've removed this on production and added the migration in place.

PR #9173, opened 2022-08-08
 Support reporting solo scores 

nanaya:
- [x] depends on #9172


notbakaneko (approved):
I feel like we should have a helper for this `alter table enum` thing with the number of times it appears 🤔

PR #9180, opened 2022-08-09
Convert Beatmapset discussion post to typescript

notbakaneko:
Now with more mobx and less data that doesn't exist.

The `post.system` checks are just there for typing checks, the class itself shouldn't getting any system posts but that's a typing fix for another time.

There's more changes to be done involving `Editor` that I'm not doing without #9062.
Known existing issue where after making an edit and then navigating away and back, the unedited version shows.

- [x] #9162
- [x] #9039


nanaya:
Conflicts...

resources/assets/lib/beatmap-discussions/post.tsx
```diff
@@ -0,0 +1,457 @@
...
+            editor: <UserLink className={`${bn}__info-user`} user={this.props.lastEditor} />,
+            update_time: <TimeWithTooltip dateTime={this.props.post.updated_at} relative />,
+          }}
+          pattern={osu.trans('beatmaps.discussions.edited')}
+        />
+      </span>
+    );
+  }
+
+  private renderKudosu() {
```
nanaya, line 218: Missing check `if @props.type == 'discussion' && @props.discussion.current_user_attributes?.can_moderate_kudosu`

Also there's another kudosu thing in this component (kudosu denied message) but this function name doesn't tell which one it is.

nanaya (changes requested):

resources/assets/lib/beatmap-discussions/post.tsx
```diff
@@ -0,0 +1,445 @@
...
+    }
+
+    this.xhr = $.ajax(route('beatmapsets.discussions.posts.update', { post: this.props.post.id }), {
+      data: {
+        beatmap_discussion_post: {
+          message: this.message,
+        },
+      },
+      method: 'PUT',
+    }).done((beatmapset) => runInAction(() => {
```
nanaya, line 438: `beatmapset` is still `any` here

resources/assets/lib/beatmap-discussions/post.tsx
```diff
@@ -0,0 +1,445 @@
...
+  @computed
+  private get isTimeline() {
+    return this.props.discussion.timestamp != null;
+  }
+
+  constructor(props: Props) {
+    super(props);
+    makeObservable(this);
+
+    this.handleTextareaKeyDown = makeTextAreaHandler(this.handleKeyDownCallback);
```
nanaya, line 92: the original callback can use a rename as well I think

nanaya (changes requested):

nanaya (approved):

PR #9190, opened 2022-08-12
QOL change: Opens the download link in a new tab.

IvanDuran02:
As someone who would download 10s of map packs at a time, this would be a life saver.


Walavouchey:
ctrl-clicking does the same thing

peppy:
This is a UX anti-pattern. Please figure a solution locally such as the ctrl-click proposal.

PR #9193, opened 2022-08-15
Add beatmap id index on solo scores

nanaya:
To be used for deleting scores when its beatmap(set) state is changed.

migration entry is `2022_08_15_074123_index_beatmap_id_on_solo_scores`
index name is `solo_scores_beatmap_id_index`

- [x] add to production


notbakaneko (commented):
Should this be run separately on production (based on how long the previous one took)?

peppy:
Yep, leave this one open, I'll run and merge once done.

peppy:
I may have missed the index name for this one, but should be okay. I don't think we'll ever be referencing it directly?

peppy (approved):

PR #9206, opened 2022-08-18
Add pp weight support for solo scores

nanaya:
There might be a better way to do this 🤔 Also nothing is assigning it to the model at the moment (it's supposed to replace the legacy user best scores).

Removed `$position` attribute since it doesn't seem to be used anywhere.


app/Models/Solo/Score.php
```diff
@@ -28,14 +30,16 @@
  */
 class Score extends Model implements Traits\ReportableInterface
 {
-    use Traits\Reportable;
+    use Traits\Reportable, Traits\WithWeightedPp;
 
     const PROCESSING_QUEUE = 'osu-queue:score-statistics';
 
+    public ?float $weight = null;
```
notbakaneko, line 37: unneeded?

notbakaneko (changes requested):

notbakaneko (approved):

PR #9209, opened 2022-08-18
Remove unused current email translation key

ThePooN:
Presumably introduced by this PR https://github.com/ppy/osu-web/pull/1236 5 years ago... which I think tried fixing the wrong string. That's leaving me very confused o_O


nanaya:
Pretty sure that field is for current password 🤔 And if anything that translation entry should probably be deleted instead.

notbakaneko (changes requested):
The field is for current password
image

ThePooN:
Ah! got too easily convinced by that string presence.  
Updated the PR to remove it. I assume a crowdin update will remove it from all the lang files.

nanaya (approved):

PR #9212, opened 2022-08-19
Fix test route url

nanaya:
🤨


notbakaneko (approved):
that's an....interesting file 🤨

PR #9214, opened 2022-08-19
Update RouteScopesHelper

notbakaneko:
update to use `Route::gatherRouteMiddleware`; the main difference between the current `gatherMiddleware` and `Route::gatherRouteMiddleware` is `Route::gatherRouteMiddleware` calls `gatherMiddleware` and also resolves the middleware groups and class names, as well as removing middleware excluded by `withoutMiddleware()`


app/Libraries/RouteScopesHelper.php
```diff
@@ -32,16 +37,16 @@ public function fillMissingMiddlewares()
             }
 
             // add missing middleware if necessary; exact order might be wrong.
```
nanaya, line 39: not part of this PR scope - unless you count "missing route-scopes:" should be updated to "missing RouteScopes" - but it would be nice if a comment is added to `fillMissingMiddlewares` function header which says something along the line it's to ensure the RouteScope is in the middleware (during export/convert) to be later used for testing. (that seems backward - as in the test itself should be the one checking for RouteScopes - but whatever)

nanaya (changes requested):

nanaya (approved):

PR #9215, opened 2022-08-19
Add users lookup to public api

### LINKED ISSUE #8800: Make `/users?ids=[]` endpoint public
We're developing a website that requires frequent batch-updating players' statistics (potentially on all rulesets as well) and `/users` endpoint seems to be a perfect fit for that, except it's locked to be lazer-only. Would it be possible to make it public? 

Another good option would be adding `all` ruleset for either `/me` or `/users/{id}` endpoint to cut down api calls from potentially 4 calls per user to just one.

notbakaneko:
I just picked an arbitrary cost of 1 per user to lookup 👀 

closes #8800

requires:
- [x] #9207


nanaya:
Conflicts 💥

nanaya (approved):

PR #9220, opened 2022-08-22
Add queue scores by user id

nanaya:
To be used by user reindex process (LIO userIndex).


notbakaneko (changes requested):
Having to use `--user=1 --all` seems unintuitive especially if `user` is being ignored for `ids`. Then there's `--user=1 --from=100` which is actually `user = 1 or id > 100`...

nanaya:
I'm fixing this mixing parameter nonsense ಠ_ಠ

app/Console/Commands/EsIndexScoresQueue.php
```diff
@@ -104,11 +90,66 @@ public function handle()
...
+        } else {
+            $this->schemas = [$schema];
+        }
+    }
+
     private function parseOptionIds(): Set
     {
         $ret = new Set();
+        $ids = $this->option('ids');
+        $ids = is_array($ids) ? $ids : explode(',', get_string($ids) ?? '');
```
notbakaneko, line 150: seems strange that `$ids` allows array without `--ids=*` as option signature 🤔  (and weirder that laravel doesn't enforce `*` when passed as direct argument 🤔 )

tests/Commands/EsIndexScoresQueueTest.php
```diff
@@ -0,0 +1,107 @@
...
+            array_merge(is_callable($params) ? $params() : $params, [
+                '--no-interaction' => true,
+                '--schema' => static::SCHEMA,
+            ]),
+        );
+    }
+
+    public function dataProviderForTestParameterValidity(): array
+    {
+        return [
```
notbakaneko, line 64: seems missing `--user=0` case

notbakaneko (changes requested):

app/Console/Commands/EsIndexScoresQueue.php
nanaya: I don't think supporting that for cli is a good idea... The resulting option would then always be in an array and if we're going to still support comma separated value it also means additional parsing inside the array.

Laravel doesn't seem to do anything with the param values when passed using `Artisan::call`.

notbakaneko (approved):

PR #9222, opened 2022-08-23
Use solo scores

nanaya:
Might as well get this out for more testing. Some parts may be less tested than the others 👀 

- beatmap leaderboard
- user beatmap scores
- user profile scores (first, best, recent)
  - recent plays show all user plays, potentially more than 24h. Also the count only goes up to 100 (not displayed by frontend)
- beatmap pack completion state
- beatmap state change (use solo scores instead of legacy for adjusting user rank stats)
- queue reindex when reindexing user using lio (`/user-index/{id}`)

Pre-merge tasks (incomplete):
- [x] depends on #9220 
- [x] finalise `ScoreInfo` (`ScoreData`) structure
- [x] indexer update based on ^
- [x] legacy importer update based on ^^
  - [x] legacy_total_score
  - [x] legacy replay (ppy/osu-queue-score-statistics#70)
- [ ] final legacy score imports
- [ ] push legacy scores (`osu_scores` and `osu_scores_high`) to solo scores table
- [ ] use solo scores to calculate user rank stats (`osu_user_stats{$mode}.{$rank}_rank_count`)
  - only `is_legacy` will be considered for now
- [x] ~update legacy score processor to also set `beatmap_leaders`~ not necessary for now, only displaying legacy firsts, sourced from `osu_leaders{mode}`
- [x] split `Beatmaps@scores` and `Beatmaps@soloScores` again so one returns legacy scores (`legacy_total_score != null` / `is_legacy is true`) and the other returns lazer scores ~sorted by new total score~ (everything sorted the same now)
- [x] ~add total score calculation to frontend for displaying lazer total score (when `legacy_total_scores` is missing/zero)~ `total_score` is always included for now
- [ ] add config to disable `is_legacy` filter (for displaying lazer score on dev web leaderboard)
- [ ] anything I missed and forgot about

Immediate task after merging:
- ~fill up beatmap leaders table (`artisan beatmap-leaders:refresh`)~ can be done whenever as nothing will be using it until lazer scores are shown on the relevant section

Todo after this got merged and deployed and stuff:
- stop converting new solo scores to legacy score table
- use solo scores for user total score reset
- update pinned scores to point solo scores
- redirect legacy scores (`/scores/{mode}/{id}`) to solo scores (`/scores/{id}`)
- handle new solo score id-based replay storage
- restructure multiplayer score to use solo score


smoogipoo:
Not sure what I'm supposed to be reviewing here, so I'm going to take myself off review.

Just one question: I added https://github.com/ppy/osu-elastic-indexer/pull/126 so the es indexer calculates total score/accuracy. Is osu-web able to use this, or will you need to port the lazer/es indexer calculations to the frontend?

nanaya:
It needs to be ported here as well, yes.

smoogipoo:
Okay. So to calculate total score and accuracy, you need to:
- Refer to https://github.com/ppy/osu/wiki/Scoring#hit-results and create a mapping for hit results to the "Score" value, and a mapping for which result affects combo/accuracy/bonus. If you'd rather read code, https://github.com/ppy/osu/blob/82b9e1f0eb218a35c2c3314fb4b0456de36069da/osu.Game/Rulesets/Judgements/Judgement.cs#L139-L173 and https://github.com/ppy/osu/blob/82b9e1f0eb218a35c2c3314fb4b0456de36069da/osu.Game/Rulesets/Scoring/HitResult.cs#L136-L285 should guide you.
- Add a way to store "proportions" of score (`accuracy_portion`, `combo_portion`). Scores are normalised to 1M score, but:
    - **osu!**: Accuracy = 30% of score, Combo = 70% of score.
    - **osu!taiko**: Accuracy = 75% of score, Combo = 25% of score.
    - **osu!catch**: Accuracy = 30% of score, Combo = 70% of score.
    - **osu!mania**: Accuracy = 99% of score, Combo = 1% of score.
- The calculation follows:
    - Get the count of each ["basic"](https://github.com/ppy/osu/blob/82b9e1f0eb218a35c2c3314fb4b0456de36069da/osu.Game/Rulesets/Scoring/HitResult.cs#L185-L195) judgement in `statistics`, multiplied by its score value. This is `base_score`.
    - Get count of each ["basic"](https://github.com/ppy/osu/blob/82b9e1f0eb218a35c2c3314fb4b0456de36069da/osu.Game/Rulesets/Scoring/HitResult.cs#L185-L195) judgement in `maximum_statistics`, multiplied by its score value. This is `max_base_score`.
    - Get count of each ["affects combo"](https://github.com/ppy/osu/blob/82b9e1f0eb218a35c2c3314fb4b0456de36069da/osu.Game/Rulesets/Scoring/HitResult.cs#L150-L171) judgement in `maximum_statistics`. This is `max_achievable_combo`.
    - Get count of each ["bonus"](https://github.com/ppy/osu/blob/82b9e1f0eb218a35c2c3314fb4b0456de36069da/osu.Game/Rulesets/Scoring/HitResult.cs#L215-L229) judgement in `statistics`. This is `bonus_score`.
    - `accuracy_ratio = max_base_score == 0 ? 1 : base_score / max_base_score`
    - `combo_ratio = max_achievable_combo == 0 ? 1 : score_max_combo / max_achievable_combo`.
    - `total_score = (1_000_000 * (accuracy_portion * accuracy_ratio + combo_portion * combo_ratio) + bonus_score) * mod_multiplier`.
    - The accuracy % is just `accuracy_ratio`.
    - If you'd rather read code: https://github.com/ppy/osu/blob/82b9e1f0eb218a35c2c3314fb4b0456de36069da/osu.Game/Rulesets/Scoring/ScoreProcessor.cs#L301-L335

We don't actually have a `mod_multiplier` right now. Since scores are immutable, I'd say the best way is probably to export the multipliers via osu!tools ([mods.json](https://github.com/ppy/osu-web/blob/master/database/mods.json)), and have osu!web consume that?

Classic scoring mode isn't necessary for initial implementation, but can be done easily later too. The calculation just needs to extract the count of basic judgements and multiply it in along with yet another factor provided by each ruleset.

app/Jobs/RemoveBeatmapsetScores.php
```diff
@@ -0,0 +1,107 @@
...
+            }
+        });
+    }
+
+    private function deleteScore(Score $score): void
+    {
+        $this->resetUserRankStatsFor($score);
+        $score->delete();
+        $score->performance?->delete();
+        $this->scoreSearch->queueForIndex($this->schemas, [$score->getKey()]);
```
notbakaneko, line 68: Can do with a transaction, otherwise errors can cause the score to be already deleted from the database but not be queued yet; they won't be picked up in `handle` if the job is run again since they're already gone from the database. (or delete the score from index first like the other job)

app/Jobs/RemoveBeatmapsetScores.php
```diff
@@ -0,0 +1,107 @@
...
+                $this->deleteScore($score);
+            }
+        });
+    }
+
+    private function deleteScore(Score $score): void
+    {
+        $this->resetUserRankStatsFor($score);
+        $score->delete();
+        $score->performance?->delete();
```
notbakaneko, line 67: Should this be part of `Score`'s `delete()`?

app/Jobs/RemoveBeatmapsetScores.php
notbakaneko: Also, this is a lot of delete queries 👀 
`RemoveBeatmapsetBestScores` triggered `delete` on each score to run the extra delete logic on them, this doesn't do any yet...

Also not sure if the default job timeout is going to be an issue with all the delete queries and es requests in `resetUserRankStatsFor`; all the jobs during my testing with 10k scores per beatmap have all managed to exceed the default 60s timeout and failed.

notbakaneko (changes requested):
Still checking...

app/Jobs/RemoveBeatmapsetScores.php
nanaya: timeout should be 3600 as set in the class 🤔  are you observing that it's not the case?

also the `resetUserRankStatsFor` kind of made it difficult to have complete `delete` function so I figured everything should just go here instead for now.

notbakaneko: It seems that `queue:listen` has its own timeout that overrides the job's timeout but `queue:work` prioritizes the correct one psy🦆

app/Http/Controllers/ScoresController.php
```diff
@@ -117,14 +119,12 @@ public function userRankLookup()
...
-                'score' => $params['score'],
-            ])->firstOrFail();
+        $rank = UserRank::getRank(ScoreSearchParams::fromArray([
+            'beatmap_ids' => [$params['beatmapId']],
+            'ruleset_id' => $params['rulesetId'],
+            'before_total_score' => $params['score'],
+        ]));
 
-        return response()->json($score->userRank(['cached' => false]) - 1);
+        return response()->json(($rank ?? 1) - 1);
```
notbakaneko, line 128: actually, why does this return 1 less than the position? 🤔

notbakaneko (changes requested):
I noticed `Score::userRank()` doesn't have a fallback like best scores for when the cutoff in `UserRankCache::fetch` isn't met, which mostly affects the `global_rank` includes on `/score/{id}` being missing 🤔

app/Models/Score/Model.php
```diff
@@ -112,6 +112,12 @@ public function best()
         return $this->belongsTo("App\\Models\\Score\\Best\\{$basename}", 'high_score_id', 'score_id');
     }
 
+    public function performance()
+    {
+        // Fake relationship for when included as part of polymorph together with Solo\Score
+        return $this->beatmap()->none();
```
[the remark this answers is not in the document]
nanaya: It needs to be an actual relationship...?

```
>>> App\Models\Score\Osu::first()->performance
LogicException with message 'App\Models\Score\Osu::performance must return a relationship instance.'
>>> App\Models\Score\Osu::with('performance')->first()
BadMethodCallException with message 'Call to undefined method Illuminate\Database\Eloquent\Builder::addEagerConstraints()'
```

app/Http/Controllers/ScoresController.php
nanaya: because that's what the original api returns

app/Jobs/RemoveBeatmapsetScores.php
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    private function deleteScore(Score $score): void
+    {
+        DB::transaction(function () use ($score): void {
+            $this->resetUserRankStatsFor($score);
+            $score->performance?->delete();
+            $score->delete();
+        });
+        $this->scoreSearch->queueForIndex($this->schemas, [$score->getKey()]);
```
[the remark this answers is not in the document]
nanaya: Updated... but that caused the scores to be updated to have its preserve flag false first before queuing as otherwise it won't be deleted by indexer process ಠ_ಠ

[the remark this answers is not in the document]
notbakaneko: 🤔

app/Models/BeatmapPack.php
```diff
@@ -74,65 +76,56 @@ public function userCompletionData($user)
...
+                    ],
+                ],
+            ]]);
+            $response = $search->response();
+            $search->assertNoError();
+            $scoreIds = array_map(
+                fn (array $hit): string => $hit[$subAggScore]['hits']['hits'][0]['_id'],
+                $response->aggregations($aggName)['buckets'],
+            );
+            $completedBeatmapIds = Solo\Score::whereKey($scoreIds)->pluck('beatmap_id')->all();
```
notbakaneko, line 123: Can't this use the `key` values from `$response->aggregations($aggName)['buckets']` instead?

notbakaneko (changes requested):

nanaya:
With legacy score tables staying for a bit longer, this PR can be split into more parts, slowly changing components to use solo scores instead of in one go like here.

nanaya:
As mentioned above, this PR will be split into parts.

PR #9223, opened 2022-08-24
Add new statistics attributes for score data json

nanaya:
Up next will be replacing `total_score` with `legacy_total_score` and calculation function in js for display.


smoogipoo (approved):

smoogipoo:
Have tested this one to work :+1:

PR #9224, opened 2022-08-24
Convert more beatmapset discussion helpers to typescript

notbakaneko:
Let's just get this out of the way now ಠ_ಠ


nanaya:
conflicts...

resources/assets/lib/utils/beatmapset-discussion-helper.ts
```diff
@@ -102,6 +133,56 @@ export function linkTimestamp(text: string, classNames: string[] = []) {
   );
 }
 
+export function nearbyDiscussions<T extends BeatmapsetDiscussionJson>(discussions: T[], timestamp: number) {
+  const nearby: Partial<Record<NearbyDiscussionsCategory, T[]>> = {};
+
+  for (const discussion of discussions) {
+    if (discussion.timestamp == null || !nearbyDiscussionsMessageTypes.has(discussion.message_type)) {
```
nanaya, line 140: same user post of same day check seems missing?

resources/assets/lib/beatmap-discussions/editor-discussion-component.tsx
```diff
@@ -194,17 +195,17 @@ export default class EditorDiscussionComponent extends React.Component<Props> {
   };
 
   nearbyIndicator = (drafts: SlateElement[]) => {
-    if (this.timestamp() == null || this.discussionType() === 'praise') {
+    if (!this.editable() ||  this.timestamp() == null || this.discussionType() === 'praise') {
```
nanaya, line 198: this change removes this message

nanaya (changes requested):

resources/assets/lib/beatmap-discussions/editor-discussion-component.tsx
notbakaneko: it's there? ?_?
image

nanaya: that's weird 🤔 maybe I was imagining things. Also I just saw there's an extra space there after `||`.

resources/assets/lib/beatmap-discussions/editor-discussion-component.tsx
```diff
@@ -35,6 +38,7 @@ interface Props extends RenderElementProps {
   readOnly?: boolean;
 }
 
+@observer
```
nanaya, line 41: looks like this isn't actually needed because `this.nearbyDiscussions` is called by child component `nearbyIndicator` as part of the `DraftsContext.Consumer` (and apparently lost the reactive context from this).

Alternatively, keeping this `@observer` but removing `@action` and `makeObservable`, and then calling `this.nearbyDiscussions()` right before `const nearbyIndicator =` also removes the warning as the cache would then be filled within reactive context (this component being observer) and the subsequent `nearbyDiscussions` will use the cached value.

nanaya (changes requested):

resources/assets/lib/beatmap-discussions/editor-discussion-component.tsx
notbakaneko: better would be to make the whole thing observable and use `@computed` but I'm not fixing anything using slate without #9062 👀

nanaya: oh yeah, I forgot about that one 🫠

nanaya (approved):

PR #9226, opened 2022-08-26
Ensure http kernel and router middleware groups are set

notbakaneko:
Fixes apidocs not being generated on production builds.

It turns out that `laravel-query-detector` causes the `Http\Kernel` boot via its `QueryDetectorServiceProvider` registration, which gets excluded if dev dependencies are not installed, causing the middleware groups to not be set on the router when run from the console.

Or maybe we should always have the http kernel and middleware groups set even in console... 🤔


nanaya:
but why is it not set in docker build?

nanaya (approved):
ಠ_ಠ

PR #9230, opened 2022-08-29
Better new beatmap discussions footer alignment

### LINKED ISSUE #9228: Spelling error/responsiveness in beatmap discussion pages (Hype)
Needs a space between "hype" and "in" as per this screenshot: 
image

Also might want to check the responsiveness on mobile:
image

notbakaneko:
closes #9228

React and flexbox fun 🤨

### DIFF
--- a/resources/assets/less/bem/beatmap-discussion-new.less
+++ b/resources/assets/less/bem/beatmap-discussion-new.less
@@ -24,13 +24,12 @@
   &__footer {
     display: flex;
     gap: 5px;
-    flex-wrap: wrap;
     background-color: @osu-colour-b5;
     padding: 10px @_side-padding;
     justify-content: space-between;
 
     @media @mobile {
-      flex-direction: column;
+      flex-wrap: wrap;
     }
   }
 
--- a/resources/assets/less/bem/beatmap-discussion-new.less
+++ b/resources/assets/less/bem/beatmap-discussion-new.less
@@ -36,7 +36,6 @@
   &__footer-content {
     display: flex;
     min-height: 2em; // prevent squashing when wrapping.
-    .default-border-radius();
 
     &--right {
       gap: 5px;
--- a/resources/assets/less/bem/beatmap-discussion-new.less
+++ b/resources/assets/less/bem/beatmap-discussion-new.less
@@ -23,9 +23,10 @@
 
   &__footer {
     display: flex;
+    gap: 5px;
     flex-wrap: wrap;
     background-color: @osu-colour-b5;
-    padding: 5px @_side-padding;
+    padding: 10px @_side-padding;
     justify-content: space-between;
 
     @media @mobile {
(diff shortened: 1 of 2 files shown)


resources/assets/lib/beatmap-discussions/new-discussion.tsx
```diff
@@ -356,16 +356,19 @@ export class NewDiscussion extends React.Component<Props> {
...
-                      }}
-                      pattern={` ${osu.trans('beatmaps.hype.new_time')}`}
-                    />
+                    <>
+                      {' '}
+                      <StringWithComponent
+                        mappings={{
+                          new_time: <TimeWithTooltip dateTime={this.props.beatmapset.current_user_attributes.new_hype_time} relative />,
+                        }}
+                        pattern={` ${osu.trans('beatmaps.hype.new_time')}`}
```
nanaya, line 365: one of the extra spaces isn't needed

resources/assets/less/bem/beatmap-discussion-new.less
```diff
@@ -23,21 +23,18 @@
 
   &__footer {
     display: flex;
-    flex-wrap: wrap;
```
nanaya, line 26: The buttons look rather wrong. The box message could also use some padding (this one is an existing problem though).

resources/assets/less/bem/beatmap-discussion-new.less
```diff
@@ -23,21 +23,18 @@
...
+    gap: 5px;
     background-color: @osu-colour-b5;
-    padding: 5px @_side-padding;
+    padding: 10px @_side-padding;
     justify-content: space-between;
 
     @media @mobile {
-      width: 100%;
-      flex-direction: column;
```
nanaya, line 33: Button overflow due to this change

nanaya (changes requested):

resources/assets/less/bem/beatmap-discussion-new.less
notbakaneko: More like text is too long and should be wrapping anyway.

I intentionally changed this so the timestamp message would not shift the buttons:
image
image

Either the text should be wrapping or the elements should always be stacking (or buttons should not grow which then causes elements to not align visually)

nanaya: I don't think the tall button is a good compromise

resources/assets/less/bem/beatmap-discussion-new.less
```diff
@@ -49,6 +46,23 @@
     }
   }
 
+  &__footer-column {
```
nanaya, line 49: the commit message says less confusing name but there are two footer columns and this only for one specific column, not both 🤨

resources/assets/less/bem/beatmap-discussion-new.less
peppy: Would also prefer the buttons to not get taller.

nanaya (approved):

PR #9232, opened 2022-08-31
Support displaying legacy total score

nanaya:
Prioritise displaying it if available. There's a different problem that there's no indication whether the score is legacy or lazer though 🤷


smoogipoo:
Also, scores submitted from lazer must not have this set, if that isn't already checked (can't see such a check at a glance).

nanaya:
Only fields specified in the [controller](https://github.com/ppy/osu-web/blob/0fc2b21cbaa3b7729c4013ade2ddfdc5584ad1c8/app/Http/Controllers/Solo/ScoresController.php#L32-L50) are stored.

smoogipoo:
Oh, so that doesn't include `maximum_statistics`? That's an issue.

nanaya:
That is indeed. And should be fixed by #9233.

notbakaneko (approved):

PR #9233, opened 2022-08-31
Fix maximum statistics attribute not being stored

nanaya:
Uh ( ﾟ ヮﾟ)

@peppy, if client has been sending it, it's never stored in the db.


smoogipoo:
Does this also need to be added to the multiplayer controller?

smoogipoo (approved):

peppy:
Not a huge deal, we can just re-run the migration script (was planning to anyway to make sure).

Regarding multiplayer scores, I believe the end goal is still to combine the multiplayer and solo scores tablet into one.

notbakaneko (approved):
Is this fine to merge now?

smoogipoo:
It's fine by me.

PR #9234, opened 2022-08-31
Add perfect combo check for solo scores

nanaya:
`legacy_perfect` only exists on scores from `osu_scores{mode}_high`.


nanaya:
yes, according to @smoogipoo . Although now I think about it I don't know if that also applies to lazer scores (or if it's decided at all)

smoogipoo:
Mania in lazer doesn't have the same quirk as stable, so yes you can just ensure there's no `miss` or `large_tick_miss`.

Also I don't know how osu!web decides perfect for stable scores, because it doesn't seem to be just absence of misses but rather absence of anything < `hit_200`? Notice score #20 for example:
image

nanaya:
Currently it's based on the `perfect` flag in the score itself.

smoogipoo:
Jesus.
```csharp
if ((CurrentScore.CountMiss + CurrentScore.Count50 + CurrentScore.Count100) == 0 && ((float)CurrentScore.CountKatu / CurrentScore.TotalHits) < 0.1)
    CurrentScore.Perfect = true;
else
    CurrentScore.Perfect = false;
```
So no misses/50/100s and less than 10% 200s. Yeah just go via statistics for lazer :sweat_smile:

notbakaneko (approved):
ok then 👀

PR #9235, opened 2022-08-31
Add auto skip to multiplayer rooms

smoogipoo:
A bit unfortunate to add more columns to this table. As a future effort I'll probably look into a better structure - perhaps a json column?


peppy (approved):

app/Models/Multiplayer/Room.php
```diff
@@ -38,6 +38,7 @@
  * @property int $user_id
  * @property string $type
  * @property string $queue_mode
+ * @property bool $auto_skip
```
nanaya, line 41: also need to add the column to `$casts` variable (see [this](https://github.com/ppy/osu-web/blob/444721f0812540be670af1d580a7e2eada4421bf/app/Models/User.php#L192) for example)

nanaya (changes requested):

nanaya (approved):

PR #9247, opened 2022-09-04
Update country names for CZ and TR

### LINKED ISSUE #8650: Update country name for Czechia
Hello,
player from Czechia here.

According to the [ISO-3166 standard](https://www.iso.org/obp/ui/#iso:code:3166:CZ), that I found out the website uses to label countries, Czech Republic's short country name is defined as `Czechia` while the osu!web still uses the full name `Czech Republic`. 
We adopted the short name in 2016 (should been in 1993 but failed) and prefer to be called that way (like French Republic is called France). So please update this.

Thank you

ubinici:
As per [ISO 3166](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes), the country names for the following countries were changed:
"Czech Republic" -> "Czechia" [[1]](https://www.iso.org/obp/ui/#iso:code:3166:CZ)
"Turkey" -> "Türkiye" [[2]](https://www.iso.org/obp/ui/#iso:code:3166:TR)

resolves #8650 

@peppy these changes are probably needed to be carried out within the database as well.


database/seeders/ModelSeeders/CountrySeeder.php
```diff
@@ -270,7 +270,7 @@ private function getCountryData()
             ['acronym' => 'TM', 'name' => 'Turkmenistan', 'display' => 0],
             ['acronym' => 'TN', 'name' => 'Tunisia', 'display' => 1],
             ['acronym' => 'TO', 'name' => 'Tonga', 'display' => 0],
-            ['acronym' => 'TR', 'name' => 'Turkey', 'display' => 1],
+            ['acronym' => 'TR', 'name' => 'Türkiye', 'display' => 1],
```
cl8n, line 273: rest of this table has diacritics removed, idk if it matters but may as well be consistent.

ex. Saint Barthélemy -> Saint Barthelemy, Åland Islands -> Aland Islands

cl8n (dismissed):

nanaya:
I think with either have updated their name committees like IOC ([CZE](https://olympics.com/ioc/czechia), [TUR](https://olympics.com/ioc/turkiye)) and FIFA ([CZE](https://www.fifa.com/fifa-world-ranking/CZE), [TUR](https://www.fifa.com/fifa-world-ranking/TUR)), this change should be fine.

Although I wonder if there will be problem with handling `ü` character in stable client. @peppy

ubinici:
> Although I wonder if there will be problem with handling `ü` character in stable client. @peppy 

I don't think `ü` needs to be handled as clayton pointed out. I have already replaced it with `u` to keep things uniform, should be fine to keep it that way in web or any other platforms to be altered to reflect this change.

nanaya (approved):

PR #9250, opened 2022-09-05
Queue beatmapset for bss process when ranked

### LINKED ISSUE #9246: Add newly ranked beatmaps to `bss_process_queue`
Currently, the osu!(lazer) client is made aware of any changes in beatmaps statuses on startup. This happens via a mechanism that tracks `bss_process_queue`.

Right now, this doesn't correctly handle changes in ranked status (from qualified to ranked) as entries are not added at this boundary (tracking at https://github.com/ppy/osu/issues/19488). The easiest solution is to insert an entry during the qualified -> ranked operation.

```sql
INSERT INTO bss_process_queue (beatmapset_id) VALUES (1520317);
```

Default values for the table work fine.

nanaya:
The model name and relation seem rather questionable but otherwise should work 🤷‍♀️

Resolves #9246.

### DIFF
--- a/app/Models/Beatmapset.php
+++ b/app/Models/Beatmapset.php
@@ -222,6 +222,11 @@ class Beatmapset extends Model implements AfterCommit, Commentable, Indexable, T
         return $this->hasMany(BeatmapDiscussion::class);
     }
 
+    public function bssProcessQueues()
+    {
+        return $this->hasMany(BssProcessQueue::class);
+    }
+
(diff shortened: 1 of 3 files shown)


notbakaneko (changes requested):
`testRankFromWrongState` should have the corresponding expects `0` change assertion, shouldn't it? 🤔

notbakaneko (approved):

PR #9253, opened 2022-09-06
Add support for `is_legacy` solo score index field

nanaya:
This also updates the BeatmapLeader to only consider lazer scores. User rank on score page (`Solo\Score#userRank`) will be updated separately.

- [x] production index needs to be recreated with updated indexer


app/Models/BeatmapLeader.php
```diff
@@ -26,6 +26,7 @@ public static function sync(int $beatmapId, int $rulesetId): void
     {
         $searchParams = ScoreSearchParams::fromArray([
             'beatmap_ids' => [$beatmapId],
+            'is_legacy' => false,
```
notbakaneko, line 29: Do solo scores from stable have a `build_id`?

nanaya: no, it's what's being checked for is_legacy in the indexer itself as well.

notbakaneko: nvm, this is a different leader from what I was thinking of 🤨

notbakaneko (changes requested):

notbakaneko (approved):
I was thinking could just make use of the `array_merge` already in `makeData` but this works.

nanaya:
This has been approved but not merged 🤔

notbakaneko:
is the index updated yet 👀

nanaya:
it sure has since end of last month

nanaya:
it means client is currently showing mixed scores for some rulesets until this is deployed 👀

PR #9256, opened 2022-09-07
Add contest vote requirement

nanaya:
Add `requirement` object inside `extra_options`:

```json5
{
  "requirement": {
    "name": "playlist_beatmapsets",
    "room_ids": [1, 2, 3],
    "must_pass": false // optional (default true)
  },
  // ...other options...
}
```

Contest with child contest(s) is currently not supported quite right. Namely the requirement must be duplicated and they all must be the same.

### DIFF
--- a/app/Models/Contest.php
+++ b/app/Models/Contest.php
@@ -5,11 +5,13 @@
 
 namespace App\Models;
 
+use App\Exceptions\InvariantException;
 use App\Traits\Memoizes;
 use App\Transformers\ContestEntryTransformer;
 use App\Transformers\ContestTransformer;
 use App\Transformers\UserContestEntryTransformer;
 use Cache;
+use Exception;
 
 /**
  * @property \Carbon\Carbon|null $created_at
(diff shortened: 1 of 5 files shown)


app/Models/Contest.php
```diff
@@ -61,6 +63,39 @@ public function votes()
...
+        }
+
+        switch ($requirement['name']) {
+            // requires playing (and optionally passing) all the beatmapsets in the specified room ids
+            case 'playlist_beatmapsets':
+                $roomIds = $requirement['room_ids'];
+                $mustPass = $requirement['must_pass'] ?? true;
+                $beatmapIdsQuery = Multiplayer\PlaylistItem::whereIn('room_id', $roomIds)->select('beatmap_id');
+                $requiredBeatmapsetCount = Beatmap::whereIn('beatmap_id', $beatmapIdsQuery)->distinct('beatmapset_id')->count();
+                $playedBeatmapIdsQuery = Multiplayer\Score
```
notbakaneko, line 81: limit to completed plays?

notbakaneko (changes requested):

PR #9261, opened 2022-09-08
Remove unused broken `UserReport#score` relation

nanaya:
The fix is a bit long and involves overriding function (see [this](/staudenmeir/eloquent-json-relations/issues/52) for example) and nothing is using it at the moment anyway (otherwise errors would've been thrown a long time ago).


notbakaneko (approved):
I think we'd be using `reportable` anyway since score wouldn't make sense for any of the new types added since then...

PR #9264, opened 2022-09-09
Add migration changing all `score_id`s to `BIGINT`

smoogipoo:
- [x] migration entry `2022_09_09_011035_upgrade_legacy_scoreids_to_bigint`

This PR is just to get the ball rolling to make sure the change isn't forgotten about. I have not tested if osu!web doesn't actually fall over in some way with `score_id > int_max`.

So @nanaya / @notbakaneko please feel free to take over this PR and/or build on top of it if any additional changes are necessary.

For comparison:

[Before](https://github.com/ppy/osu-web/files/9531736/before.txt)
[After](https://github.com/ppy/osu-web/files/9531737/after.txt)
[Rollback](https://github.com/ppy/osu-web/files/9531780/rollback.txt)


peppy:
Have confirmed all these changes are applied (or applying) on production, and inserted the migration entry.

peppy (approved):

PR #9265, opened 2022-09-09
Add score indexer to docker-compose

notbakaneko:
Also increases the migrator's timeout waiting for db as it can easily timeout if nothing has been initialized yet in the db's volume.

- [x] https://github.com/ppy/osu-elastic-indexer/pull/129
- [x] new docker image uploaded


docker-compose.yml
```diff
@@ -111,6 +112,18 @@ services:
...
+  score-indexer:
+    image: pppy/osu-elastic-indexer
+    command: ["queue", "--force-version", "--wait"]
+    depends_on:
+      - db
+      - elasticsearch
+      - redis
+    environment:
+      <<: *x-env
+      ES_HOST: http://elasticsearch:9200
```
nanaya, line 124: osu-web can use the  `ES_HOST` value in this form so update that instead

docker-compose.yml
```diff
@@ -111,6 +112,18 @@ services:
...
+      DB_CONNECTION_STRING: Server=db;Database=osu;Uid=osuweb;
```
nanaya, line 125: probably fine in default env

docker-compose.yml
```diff
@@ -12,6 +12,7 @@ x-env: &x-env
   NOTIFICATION_REDIS_HOST: redis
   REDIS_HOST: redis
   SESSION_DRIVER: redis
+  SCHEMA: "${SCHEMA:-1}"
```
nanaya, line 15: this should be in the service itself as the name is too generic for a very specific thing.

nanaya (changes requested):

nanaya (approved):

PR #9267, opened 2022-09-09
Use mod names from data JSON

nanaya:
Noticed this when checking own score with new mods on score detail page. The icons will still be missing but that's for another day...


resources/assets/lib/globals.d.ts
```diff
@@ -1,6 +1,10 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
 // See the LICENCE file in the repository root for full licence text.
 
+declare module 'mod-names.json' {
+  export default Partial<Record<string, string>>;
```
notbakaneko, line 5: type can't be `export default` directly, it needs to be declared on a value first. The `import` from this is still `any`

notbakaneko (changes requested):

notbakaneko (approved):

PR #9274, opened 2022-09-12
Add scope to remove order in a query

nanaya:
Seems like a good thing to have.


app/Models/Score/Best/Model.php
```diff
@@ -186,7 +186,9 @@ public function macroAccurateRankCounts()
         return function ($query) {
             $newQuery = clone $query;
             // FIXME: mysql 5.6 compat
-            $newQuery->getQuery()->orders = null;
+            // FIXME: figure out what exactly the above is. Also it's probably
```
notbakaneko, line 189: I know with `group by` with `limit`, mysql 5.5 ordered the results before grouping and 5.6+ doesn't which caused duplicates if paging by offset 🤔

notbakaneko (approved):

PR #9280, opened 2022-09-13
Allow organising news posts by year

### LINKED ISSUE #9279: Support per-year news posts directory
Current `/news/` in osu-wiki repo is getting a bit too large and can't be browsed in full from the web anymore (truncated listing). Moving them to subdirectories by year should help but it needs to be supported here first.

nanaya:
Maybe? 🤷 

Resolves #9279


notbakaneko (approved):
Seems fine, need to arrange for the wiki to be updated for the existing articles to link to the correct entry.

nanaya:
this should work with either directory structure so the actual structure can be updated whenever after this is deployed

TicClick:
perhaps I am late to the party, but we also had an idea of grouping news by whatever project they belong to -- see https://github.com/ppy/osu-wiki/issues/2247. I don't have a strong opinion here (nor this should block the PR), but maybe @cl8n or @Joehuu do

cl8n:
that issue seems like tangential idea that can still happen later if it's wanted
this pr looks good to me

notbakaneko:
The link back to the github back is always `yyyy/slug-name` even if the page isn't actually in the directory

nanaya:
the slug isn't available to anyone but admin and wiki people can't actually change the structure yet as it won't work without this...

PR #9285, opened 2022-09-15
Fix user rank on solo score model

nanaya:
I think this should work 🤔 

- [x] depends on #9253
- [x] adding test is probably a good idea


notbakaneko (changes requested):

notbakaneko (changes requested):
should `ScoreSearch` be using `legacyTotalScore` instead of `totalScore` for legacy scores?

PR #9298, opened 2022-09-22
Faster and less blurry flags

### LINKED ISSUE #9299: Some wiki articles freeze on iPhone 12 (iOS 16)
An issue similar to https://github.com/ppy/osu-web/issues/9289

When trying to access some of the wiki articles (likely the ones which have country flags in tables) Safari won't load them properly, while working perfectly fine from PC and Android devices. Additionally, if I try to switch between tabs, the page just freezes and I cannot press on any link as seen on the last video, the only option is to close the browser tab.

Bellow I'm adding three examples of the pages where I could notice such problem:

https://youtube.com/shorts/TEqTD15SkPU?feature=share
https://youtube.com/shorts/BgjXLJmVqUg?feature=share
https://youtube.com/shorts/A8QGB2Kf3jY?feature=share

https://osu.ppy.sh/wiki/en/Contests/triangles_Beatmapping_Contest
https://osu.ppy.sh/wiki/en/Contests/A_Labour_of_Love_Contest
https://osu.ppy.sh/wiki/en/People/The_Team/Beatmap_Nominators

notbakaneko:
Always render flag in a separate layer; speeds up rendering on Safari especially on pages with a lot of flags (particularly mobile safari - desktop Safari doesn't seem to have the issue to the same extent unless in a scrollable element).
This was generally avoided previously due it turning the flags into a blurry mess on Chrome, except a recent update seems to always turn it into a blurry mess anyway.

The flag ratio is changed to the original 100/72 of the svg instead of 4/3; this stops the flag from turning into a blurry mess on Chrome.

fixes #9299

### DIFF
--- a/resources/assets/less/bem/flag-country.less
+++ b/resources/assets/less/bem/flag-country.less
@@ -3,7 +3,7 @@
 
 .flag-country {
   --height: 1em;
-  --width-height-ratio: calc(100 / 75);
+  --width-height-ratio: calc(100 / 72); // Original flag svg ratio; using a different ratio turns the flag into a blurry mess on Chrome.
 
   height: var(--height);
   width: calc(var(--height) * var(--width-height-ratio));
--- a/resources/assets/less/bem/flag-country.less
+++ b/resources/assets/less/bem/flag-country.less
@@ -14,6 +14,8 @@
   position: relative;
   filter: saturate(1.1);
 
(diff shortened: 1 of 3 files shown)


peppy (approved):

notbakaneko:
Random 2024/10/11 update because I noticed it: The flag in the menu bar is blurry on Chrome on the first load but fine on subsequent navigations.

PR #9309, opened 2022-09-26
Skip refetching same user from database

nanaya:
The other preload is harder to do. Also this will still rerun the user transformer. Looking again it seems a bit overkill passing the whole thing repeatedly for each item just to display the username 🤷

Also who knows what else it's currently being used for in term of api.


notbakaneko (commented):
Should probably mark for removal and deprecate instead; API requests hitting this should already have the user or be getting it separately, anyway..

nanaya:
What's supposed to be simple PR ended up way larger than expected. Deprecation and thing will be done separately ಠ_ಠ

notbakaneko (approved):

PR #9310, opened 2022-09-26
Missing No Mod tooltip

### LINKED ISSUE #9307: Global Leaderboard shows "unknown" in place of "No Mod".
NlBkQ

Seems to happen for every gamemode. Tried multiple maps starting from different gamemodes as well. Tried different browsers.

notbakaneko:
Forgot about this extra case 👀 

closes #9307

### DIFF
--- a/resources/assets/lib/utils/score-helper.ts
+++ b/resources/assets/lib/utils/score-helper.ts
@@ -95,12 +95,6 @@ export const modeAttributesMap: Record<GameMode, AttributeData[]> = {
 };
 
 export function modName(acronym: string): string {
-  if (acronym === 'V2') {
-    return 'Score V2';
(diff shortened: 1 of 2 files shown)


resources/assets/lib/utils/score-helper.ts
```diff
@@ -97,6 +97,8 @@ export const modeAttributesMap: Record<GameMode, AttributeData[]> = {
 export function modName(acronym: string): string {
   if (acronym === 'V2') {
     return 'Score V2';
+  } else if (acronym === 'NM') {
```
nanaya, line 100: looking again, those additional names should just be added when generating the `mod-names.json` file...

nanaya (changes requested):

nanaya (approved):

PR #9311, opened 2022-09-26
Deprecate user include in profile page scores

nanaya:
Not sure whether or not explicit user parameter for `hasMenu`/`canBeReported` is better. Making it optional is slightly shorter but in places where score may not contain user (like in profile page after the include removed), it may cause unexpected result when the user isn't actually passed.

The alternative is to just not require user for reporting scores. Reportable requirement and form title will need adjustment though.

Or maybe just don't check `user.is_deleted`...


notbakaneko:
Maybe just letting them be reportable and discarding it server side if the user doesn't exist makes more sense 🤔

notbakaneko (changes requested):
Score by deleted user can be made reportable especially considering nothing else seems to check against it.

notbakaneko (approved):

notbakaneko (changes requested):
conflicts 💥

notbakaneko (approved):

PR #9323, opened 2022-09-29
Fix beatmapset 'download disabled' boolean casting

S1n1st3rn3ss:
Addition to #9321 

Along with `nsfw` and `spotlight` flags on beatmapsets `download_disabled` was also affected


notbakaneko (approved):

riesha:
when is this getting pushed to production because it's still broken atm

PR #9325, opened 2022-09-30
Simpler model `getKey` function

nanaya:
The original `getKey` uses `getAttribute` which is not the fastest thing. By "not the fastest" thing I mean it does a lot of things to deal with casting magic, *especially* when there's no casting involved for the specified field. Speed improvement is minimal but it's definitely there.

### DIFF
--- a/app/Models/Model.php
+++ b/app/Models/Model.php
@@ -30,6 +30,11 @@ abstract class Model extends BaseModel
         static::addGlobalScope(new MacroableModelScope());
     }
 
+    public function getKey()
+    {
+        return $this->attributes[$this->primaryKey] ?? null;
+    }
+
     public function getForeignKey()
     {
         if ($this->primaryKey === null || $this->primaryKey === 'id') {


notbakaneko:
I'd hope we don't have single column primary keys that aren't integer or string..... ಠ_ಠ

peppy:
Here's a few that you may want to double-check:

```csharp
| weak_passwords                    | binary(16)             |
| osu_countries                     | char(2)                |
| phpbb_forums_access               | char(32)               |
| osu_user_month_playcount          | char(4)                |
| osu_user_replayswatched           | char(4)                |
| test_solo_scores                  | datetime               |
| osu_beatmap_failtimes             | enum('fail','exit')    |

| osu_events                        | timestamp              |
| osu_registrations                 | timestamp              |
| osu_scores                        | timestamp              |
| osu_scores_fruits                 | timestamp              |
| osu_scores_mania                  | timestamp              |
| osu_scores_taiko                  | timestamp              |
| osu_stats                         | timestamp              |
| score_process_queue               | timestamp              |
```

All the rest are `varchar` or `int` types.

peppy:
(some of those are multi-column and can be disregarded)

peppy:
Let's deploy this and see how much goes wrong.

notbakaneko:
I wonder if we should override `getAttribute` to remove all the stuff we don't use; this is suspiciously a lot of overhead for a bunch of method calls which which loop over empty arrays, and then a couple of `method_exists` and `array_key_exists` that return `false` to return the original value 🤔 

Not to mention it seems to need to do less work if an attribute getter is defined since it can just skip the last few extra checks...

PR #9327, opened 2022-09-30
Lazy load profile page extra sections

notbakaneko:
Avoid loading all data on page load.
~25 queries on first load instead of ~70

Part 2 of ?(3?)

Scroll position update if a lazy loaded area above the current scroll position loads in will be handled in another PR

- [x] ~~better placeholder layout~~ just going with the blank space for now unless there's a better option that's simple to implement
- [x] #9324
- [x] #9363 scroll after load + scroll to section


nanaya:
> Scroll position update if a lazy loaded area above the current scroll position loads in will be handled in another PR

Clicking on any of the page section link before anything loads also results in nonsense scroll position...

Also the intersection margin should be a lot larger than current 0.

<del>there's also the part where if someone opens multiple tabs of profile and they actually check the tabs they'll still be faced with spinners if they don't scroll each tabs down beforehand but that's just the result of lazy loading</del>

nanaya:
I'm reviewing this in #9363 because otherwise the profile page will be half broken if this gets deployed before the other one.

nanaya:
As mentioned above, this will be reviewed as part of #9363.

PR #9332, opened 2022-10-03
Allow beatmap model attribute casts to be usable without argument

nanaya:
I was about to replace the functions with just `getThing` (or `thing`?) instead of current `getThingAttribute` but that would require a lot more changes for the places which are still using it through magic accessor. That'll be for another time.


app/Models/Beatmap.php
```diff
@@ -188,8 +191,9 @@ public function getDiffSizeAttribute($value)
         return $value;
     }
 
-    public function getVersionAttribute($value)
+    public function getVersionAttribute()
     {
+        $value = $this->attributes['version'] ?? null;
         if ($this->mode === 'mania') {
             $keys = $this->getDiffSizeAttribute($this->attributes['diff_size'] ?? null);
```
notbakaneko, line 198: still calling with param

notbakaneko (changes requested):

notbakaneko (approved):

PR #9333, opened 2022-10-03
Simplify getting model raw attribute

nanaya:
I don't know why I thought it made things slower last time but now I checked it again it doesn't seem make much difference.

Depends on:
- [ ] #9325


peppy:
Needs minor conflict resolution.

notbakaneko (approved):
merges fine outside of github, oddly enough 🤔

notbakaneko:
does this just get overridden by #9335 👀

nanaya:
Kind of. The function added here is still used there.

nanaya:
Superseded by aforementioned PR.

PR #9338, opened 2022-10-04
Use correct key for beatmaps in discussion post API

nanaya:
Wow it already found a bug of accessing invalid attribute <_<


notbakaneko (approved):
👀

PR #9349, opened 2022-10-06
Update User phpdoc

notbakaneko:
Fill in/fix some missing types and add more specific `read`/`write` accessor levels (although not all tools differentiate read/write yet).


nanaya (changes requested):

app/Models/User.php
```diff
@@ -166,8 +165,8 @@
  * @property int $user_warnings
  * @property string $user_website
  * @property string $username
- * @property \Illuminate\Database\Eloquent\Collection $usernameChangeHistory UsernameChangeHistory
- * @property \Illuminate\Database\Eloquent\Collection $usernameChangeHistoryPublic publically visible UsernameChangeHistory containing only user_id and username_last
+ * @property-read \Illuminate\Database\Eloquent\Collection $usernameChangeHistory UsernameChangeHistory
+ * @property-read \Illuminate\Database\Eloquent\Collection $usernameChangeHistoryPublic publically visible UsernameChangeHistory containing only user_id and username_last
```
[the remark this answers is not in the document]
notbakaneko: I'm pretty sure a bunch of the `score*` property relations were never added to the doc header 🤔

app/Models/User.php
```diff
@@ -110,10 +109,10 @@
  * @property string $user_birthday
  * @property string $user_colour
  * @property string $user_dateformat
- * @property mixed $user_discord
+ * @property string $user_discord
```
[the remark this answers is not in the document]
notbakaneko: Some of these are a bit weird because the backing column is not nullable and the setter doesn't handle the null 👀

app/Models/User.php
```diff
@@ -38,66 +38,65 @@
...
+ * @property-read string|null $playmode
+ * @property bool $pm_friends_only
+ * @property-read \Illuminate\Database\Eloquent\Collection $profileBanners ProfileBanner
+ * @property-read Rank $rank
+ * @property-read \Illuminate\Database\Eloquent\Collection $rankHistories RankHistory
+ * @property-read \Illuminate\Database\Eloquent\Collection $receivedKudosu KudosuHistory
+ * @property-read \Illuminate\Database\Eloquent\Collection $relations UserRelation
  * @property string|null $remember_token
- * @property \Illuminate\Database\Eloquent\Collection $replaysWatchedCounts UserReplaysWatchedCount
- * @property UserReport $reportedIn
```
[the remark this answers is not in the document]
notbakaneko: `$reportedIn` is moved to the trait

app/Models/User.php
```diff
@@ -32,73 +32,102 @@
...
- * @property \Illuminate\Database\Eloquent\Collection $friends static
- * @property \Illuminate\Database\Eloquent\Collection $githubUsers GithubUser
- * @property \Illuminate\Database\Eloquent\Collection $givenKudosu KudosuHistory
+ * @property-read Collection<UserAccountHistory> $accountHistories
+ * @property-read ApiKey|null $apiKey
+ * @property-read Collection<UserBadge> $badges
+ * @property-read Collection<BeatmapDiscussionVote> $beatmapDiscussionVotes
+ * @property-read Collection<BeatmapDiscussion> $beatmapDiscussions
+ * @property-read Collection<BeatmapPlaycount> $beatmapPlaycounts
+ * @property-read Collection<Beatmap> $beatmaps Beatmap
```
nanaya, line 50: redundant `Beatmap`

app/Models/User.php
```diff
@@ -32,73 +32,102 @@
...
+ * @property-read Collection<BeatmapsetNomination> $beatmapsetNominations
```
nanaya, line 51: there's also `beatmapsetNominationsToday`

app/Models/User.php
```diff
@@ -32,73 +32,102 @@
...
+ * @property-read Collection<Chat\Channel> $channels
+ * @property-read Collection<UserClient> $clients
+ * @property-read Collection<Comment> $comments
+ * @property-read Country|null $country
+ * @property string|null $country_acronym
+ * @property-write string|null $current_password
+ * @property-read Carbon|null $displayed_last_visit
+ * @property-read string|null $email
+ * @property-read Collection<Event> $events
+ * @property-read Collection<FavouriteBeatmapset> $favourites
```
nanaya, line 66: there's `$follows`

app/Models/User.php
```diff
@@ -32,73 +32,102 @@
...
+ * @property-read Collection<Score\Osu> $scoresOsu
+ * @property-read Collection<Score\Taiko> $scoresTaiko
+ * @property-read UserStatistics\Fruits|null $statisticsFruits
+ * @property-read UserStatistics\Mania|null $statisticsMania
+ * @property-read UserStatistics\Osu|null $statisticsOsu
+ * @property-read UserStatistics\Taiko|null $statisticsTaiko
+ * @property-read Collection<Store\Address> $storeAddresses
+ * @property-read Collection<UserDonation> $supporterTagPurchases
+ * @property-read Collection<UserDonation> $supporterTags
+ * @property-read Collection<OAuth\Token> $tokens
```
nanaya, line 124: this is missing in the `getAttribute` 👀

app/Models/User.php
```diff
@@ -32,73 +32,102 @@
...
+ * @property-read Collection<Forum\TopicWatch> $topicWatches
+ * @property-read Collection<UserAchievement> $userAchievements
+ * @property-read Collection<UserGroup> $userGroups UserGroup
```
nanaya, line 127: more redundant class name

app/Models/User.php
```diff
@@ -109,30 +138,30 @@
...
  * @property int $user_last_privmsg
  * @property int $user_last_search
  * @property int $user_last_warning
- * @property int $user_lastmark
+ * @property Carbon|null $user_lastmark
  * @property string $user_lastpage
- * @property int $user_lastpost_time
- * @property int $user_lastvisit
+ * @property Carbon|null $user_lastpost_time
+ * @property Carbon|null $user_lastvisit
```
nanaya, line 164: `user_lastfm` and `user_lastfm_session` are missing...

Although I wonder if those unused/unwanted attributes should be made inaccessible instead (and commented out in `getAttribute` instead of deleted so it's not accidentally added back unless actually needed)

app/Models/DeletedUser.php
```diff
@@ -6,88 +6,7 @@
...
- * @property string $user_twitter
- * @property int $user_type
- * @property int $user_unread_privmsg
- * @property int $user_warnings
- * @property string $user_website
- * @property string $username
- * @property string $username_clean
- * @property string|null $username_previous
- * @property int|null $userpage_post_id
+ * {@inheritdoc}
```
nanaya, line 9: what's the conclusion on `@inheritdoc` usage? (also for UserNotFound)

nanaya (changes requested):
I hope I didn't miss anything 🤪

app/Models/User.php
notbakaneko: `beatmapsetNominationsToday()` returns `int` 👀

app/Models/User.php
notbakaneko: I put them in for now but I think we should consider removing unused attributes since it just adds clutter if they're unused (and also avoid accidentally using them 👀 )

nanaya (approved):

PR #9355, opened 2022-10-07
Clean up models with composite primary keys

nanaya:
I don't even know what to do with the `$primaryKey` so it gets some random value. It needs to survive `Str::studly` (`_` doesn't) just to make sure it won't match any casting function ever (and also column name).

Also reverted the nullable string getRawAttribute as primary key names are now all string (also some of them were boolean before which is how it passes model deletion null primary key check).


app/Models/Forum/Authorize.php
```diff
@@ -16,9 +16,12 @@
  */
 class Authorize extends Model
 {
-    protected $table = 'phpbb_acl_groups';
+    public $incrementing = false;
     public $timestamps = false;
+
+    protected $primaryKey = ':composite';
     protected $primaryKeys = ['group_id', 'forum_id', 'auth_option_id', 'auth_role_id', 'auth_setting'];
```
notbakaneko, line 23: wonder if it's worth fixing this not actually being a unique key in the schema 🤔

notbakaneko (changes requested):

app/Models/Forum/Authorize.php
nanaya: maybe another time ಠ_ಠ

notbakaneko (approved):

PR #9360, opened 2022-10-11
Move user achievement interop controller action and adjust test

nanaya:
Use hardcoded names to avoid #9350. Not exactly the right place but should be fine for now...?


app/Http/Controllers/InterOp/UsersController.php
```diff
@@ -7,12 +7,40 @@
...
+    {
+        $user = User::findOrFail($id);
+        $achievement = Achievement::findOrFail($achievementId);
+
+        try {
+            $userAchievement = $user->userAchievements()->create([
+                'achievement_id' => $achievement->getKey(),
+                'beatmap_id' => $beatmapId,
+            ]);
+        } catch (Exception $e) {
```
notbakaneko, line 29: missing `using`

notbakaneko (changes requested):

PR #9363, opened 2022-10-11
Lazy load profile page extra sections

notbakaneko:
Tries to maintain scroll position after a section is lazy loaded. This was really annoying due to Chrome handling off-screen layout changes differently from Safari and Firefox.

- [x] #9327
- [x] chart ref
- [x] scroll on lazy load (more like special case for page bottom breaks everything)
  - [x] Chrome
  - [x] Safari
  - [x] Firefox
- [x] skip load if data already available
- [x] scroll position when at end of document and lazy load section _shrinks_
- [x] something weird going on with Safari sizing

~~Known issue with jumpTo marker not being unset; that probably needs some reworking of `pageJump` and `pageScan`~~
~~pageJump uses `scrollIntoVIew()` instead, also~~ previous `$(window).scrollTo` sometimes still had scroll event firing immediately after callback runs.
It turns out `window.scrollTo` has an options variant that supports smooth scrolling, so that's being used for now, to keep the scroll position at least `1px` from the bottom when not scrolling to the last section on the page; at least until a more appropriate heuristic for scanning the active tab can be used.

TODO: update this
The basic logic for handling layout shift is:
- if container wants element to be in focus, scroll until top of element is visible.
- if element was above visible area, maintain distance from bottom of the page.
- if element was split by the top of viewable area, maintain distance from bottom of the page 
(this is the simpler and easier to verify option for the moment, instead of trying to figure out if the new area should go above or below the cutoff line)
- if element's top was visible and the new size causes it to be not visible, shift to make it visible
- else ignore.

Generally nothing needs to be done if the area below the visible portion gets taller.

(also modding page doesn't get lazy loading until it's typescript'ed because I'm not porting the extra context stuff to coffeescript)


nanaya:
Is this supposed to handle url with page specified like https://osu.ppy.sh/users/2#kudosu ?

notbakaneko:
It works.....when 3 different things aren't trying to scroll the page in different directions at once ಠ_ಠ

resources/assets/lib/components/lazy-load.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    const element = this.ref.current;
+    if (element == null || this.beforeRenderedBounds == null) {
+      return;
+    }
+
+    this.hasUpdated = true;
+
+    // skip scroll update if context callback has handled it.
+    if (this.context.name != null && this.context.onWillUpdateScroll?.(this.context.name)) {
```
nanaya, line 60: the immediate jump when the page is still animating scroll is rather jarring.

app/Http/Controllers/UsersController.php
```diff
@@ -143,6 +143,7 @@ public function checkUsernameExists()
     public function extraPages($_id, $page)
     {
         // TODO: counts basically duplicated from UserCompactTransformer
+        // TOOD: switch to cursor pagination?
```
nanaya, line 146: TOO D 🤔

nanaya (changes requested):

resources/assets/lib/profile-page/main.tsx
```diff
@@ -251,6 +284,67 @@ export default class Main extends React.Component<Props> {
...
+    }
+
+    this.queueScroll();
+  };
+
+  private readonly handleLazyLoadGetSnapshot = (name: ProfileExtraPage) => {
+    const element = this.extraPages[name].current;
+    if (element == null) return;
+
+    return {
```
nanaya, line 328: probably want to check distance to bottom page so the page is positioned as expected when its initial scroll is the last section

...but that may cause weird behavior if the user scrolls around while loading 🤔 (always apply `diff` for `scrollBy` in that case instead of checking the `marginTop`?)

nanaya (changes requested):
warning lint should also be adjusted accordingly

resources/assets/lib/profile-page/main.tsx
notbakaneko: I found that snapshots could happen before the final DOM tree was fully rendered which resulted in it thinking it was at the bottom when it actually wasn't 😔 
It's one of the reasons maintaining distance from the bottom got abandoned

nanaya: mark something if `pageJumptingTo` the last page?

nanaya (approved):
let's just get this in and see how it works out ( ﾟ ヮﾟ)

(and pray no one discover weird edge case on normal use

PR #9369, opened 2022-10-13
Define attributes for score legacy, solo, and token model

nanaya:
These won't make actual difference until we migrate to solo scores. Replaced initial date storage with string so no need to initialise date object just to turn it into string again.

Before: 16.17s
After: 6.18s

```
$u = App\Models\Solo\Score::find(1); $t = new App\Transformers\ScoreTransformer; json_item($u, $t); $start = hrtime(true); for ($i=0;$i<50000;$i++) { json_item($u, $t); }; echo ((hrtime(true) - $start)/1000000000).PHP_EOL;
```


notbakaneko (approved):
> Replaced initial date storage with string so no need to initialise date object just to turn it into string again
I guess this is fine as long at the value only comes from our end which appears to be the case 🤔

PR #9379, opened 2022-10-16
Contest submission requirements

### LINKED ISSUE #9243: Contest system submission requirements
Currently, the contest system isn't set up to allow for sanity checking of submissions save for basic filename requirements linked to the contest type.

This poses a problem especially for fan art contests, which have extremely stringent image resolution requirements that have notoriously caused a great deal of problems for entrants who've exported their image incorrectly (sometimes by as few as a pixel or two). The processing scripts we use filter out these incorrect entries, but a manual pass is also needed for every fan art contest for obviously incorrect entries, since many users just upload random shit to a contest listing page if they are able to.

It would be ideal if there were ``extra_options`` available to enforce resolution requirements, like ``forced_height`` and ``forced_width``. A user attempting to upload something that does not fit the requirements would be informed clientside before the upload takes place.

shinobu-uwu:
Hello! This is my first contribution to an actual open source project, so any criticism is much appreciates :smile:

This PR adds the ability to set `forced_width` and `forced_height` in `extra_options` field for fanart contests

Resolves #9243

### DIFF
--- a/resources/assets/lib/contest-entry/uploader.coffee
+++ b/resources/assets/lib/contest-entry/uploader.coffee
@@ -78,6 +78,15 @@ export class Uploader extends React.Component
           osu.popup osu.trans('contest.entry.too_big', limit: osu.formatBytes(maxSize, 0)), 'danger'
           return
 
+        if @props.contest.type == 'art'
+          image = await @convertFileToImage file
+
+          if image.width != @props.contest.forced_width || image.height != @props.contest.forced_height
+            osu.popup osu.trans('contest.entry.wrong_dimensions',
+              width: @props.contest.forced_width,
+              height: @props.contest.forced_height), 'danger'
+            return
+
         data.submit()
 
       submit: ->
@@ -94,6 +103,16 @@ export class Uploader extends React.Component
     @$uploadButton()
       .fileupload 'destroy'
       .remove()
+
+  convertFileToImage: (file) =>
+    new Promise (resolve, reject) ->
+      image = new Image()
+      reader = new FileReader()
+      reader.onload = () -> image.src = reader.result
+      reader.onerror = (error) -> reject(error)
+      image.onload = () -> resolve(image)
+      reader.readAsDataURL(file)
+
 
   render: =>
     div
--- a/tests/Controllers/ContestEntriesControllerTest.php
+++ b/tests/Controllers/ContestEntriesControllerTest.php
@@ -1,0 +1,80 @@
+<?php
+
+namespace Tests\Controllers;
+
+use App\Models\Contest;
+use App\Models\User;
+use DateTime;
+use Illuminate\Http\Testing\File;
+use Tests\TestCase;
+
+class ContestEntriesControllerTest extends TestCase
+{
+    private Contest $contest;
+    private User $user;
+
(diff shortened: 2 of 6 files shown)


resources/assets/lib/contest-entry/uploader.coffee
```diff
@@ -78,6 +78,15 @@ export class Uploader extends React.Component
           osu.popup osu.trans('contest.entry.too_big', limit: osu.formatBytes(maxSize, 0)), 'danger'
           return
 
+        if @props.contest.type == 'art'
+          image = await @convertFileToImage file
+
+          if image.width != @props.contest.forced_width || image.height != @props.contest.forced_height
```
nanaya, line 84: even if the above works, this breaks when there's no requirement

tests/Controllers/ContestEntriesControllerTest.php
```diff
@@ -0,0 +1,80 @@
...
+    private User $user;
+
+    public function testStoreEntryWithWrongFileExtension()
+    {
+        $this->prepareForStore();
+        $file = File::fake()->create('entry.osz');
+
+        $this
+            ->actingAsVerified($this->user)
+            ->postJson('/community/contest-entries', ['entry' => $file, 'contest_id' => $this->contest->id])
```
nanaya, line 23: use named route

tests/Controllers/ContestEntriesControllerTest.php
```diff
@@ -0,0 +1,80 @@
...
+
+    private function prepareForStore()
+    {
+        $this->user = User::factory()->create();
+        $this->contest = Contest::create([
+            'name' => 'Test',
+            'description_enter' => 'This is just a test!',
+            'description_voting' => 'This is just a test!',
+            'entry_starts_at' => new DateTime(),
+            'entry_ends_at' => (new DateTime())->modify('+1 day'),
```
nanaya, line 68: use carbon's `now()` and `now()->addDay()` etc

or just use the existing factory

tests/Controllers/ContestEntriesControllerTest.php
```diff
@@ -0,0 +1,80 @@
...
+
+    public function testStoreEntryWithWrongFileExtension()
+    {
+        $this->prepareForStore();
+        $file = File::fake()->create('entry.osz');
+
+        $this
+            ->actingAsVerified($this->user)
+            ->postJson('/community/contest-entries', ['entry' => $file, 'contest_id' => $this->contest->id])
+            ->assertStatus(422);
```
nanaya, line 24: this should check error message as well (and add in controller since currently there's none)

nanaya (changes requested):

nanaya (approved):

shinobu-uwu:
Thanks for the review @nanaya! Despite working with php professionally it's my first time with laravel and I think it showed :sweat_smile:

Looking forward to contribute more to osu!

Ephemeralis:
Thanks a ton for putting this in! I previously got about 4 emails on average from distressed entrants wondering why their artworks weren't on the listing, and what you've done here is going to save them a lot of anguish, and make things a bit more straightforward for anyone else hosting contests in the future.

PR #9387, opened 2022-10-18
Fix incorrect state revert on already saved model

nanaya:
It was weird after a closer look 👀


notbakaneko (approved):
makes sense I guess 🤔

PR #9388, opened 2022-10-18
Show highest achieved rank on user profile

### LINKED ISSUE #5032: Add “All time highest rank” feature aside of current rank
Got this idea while playing Clash Royale, basically when you’re hitting a rank that is your all time high, there will be a record of your rank even if you decay into oblivion. Gives the player a sense of pride and accomplishment.
Basically Cookiezi would have rank 17 and (highest rank: 1) 
A scrub like me would finally show off that I once belonged to the 3 digits club too (inb4 peppy was actually rank 1)

cl8n:
resolves #5032 + add to user api response

for now it's just in the tooltip on global rank hover. I think mr designer should find a cooler place for it because ppl probably wanna show this off


database/migrations/2022_10_18_103120_create_osu_user_performance_rank_highest.php
```diff
@@ -0,0 +1,50 @@
...
+
+            $table->primary(['user_id', 'mode']);
+        });
+        DB::statement('
+            ALTER TABLE `osu_user_performance_rank_highest` PARTITION BY RANGE (`mode`) (
+                PARTITION p0 VALUES LESS THAN (1),
+                PARTITION p1 VALUES LESS THAN (2),
+                PARTITION p2 VALUES LESS THAN (3),
+                PARTITION p3 VALUES LESS THAN (4)
+            );
```
cl8n, line 37: @peppy are these the right partitions? I copied from another table, but they weren't listed in the schema u sent

peppy:
```sql
root@db-master.osu.io:osu> show create table osu.osu_user_performance_rank_highest\G;
Table        | osu_user_performance_rank_highest
Create Table | CREATE TABLE `osu_user_performance_rank_highest` (
  `user_id` int unsigned NOT NULL,
  `mode` tinyint NOT NULL,
  `rank` int NOT NULL DEFAULT '0',
  `updated_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  PRIMARY KEY (`user_id`,`mode`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci ROW_FORMAT=DYNAMIC
/*!50100 PARTITION BY RANGE (`mode`)
(PARTITION p0 VALUES LESS THAN (1) ENGINE = InnoDB,
 PARTITION p1 VALUES LESS THAN (2) ENGINE = InnoDB,
 PARTITION p2 VALUES LESS THAN (3) ENGINE = InnoDB,
 PARTITION p3 VALUES LESS THAN (4) ENGINE = InnoDB) */
```

nanaya (approved):

yuzupon1133:
according to one person, highest rank includes the rank before rework. Is this intentional? 🤔

peppy:
Yes. It's best ever, over all changes over time.

kafibu:
When does exactly the highest rank achieved started being recorded?

peppy:
You can find details [here](https://twitter.com/ppy/status/1585991185352790016)

heyngra:
not sure if this is a problem with API data or this implementation itself, or if it's intentional, but shouldn't highest rank be at the same level or better than current one?
image

peppy:
That may happen if you have not played recently. It will *not* track passive increases. You need to set a new high score for it to track.

6800cube:
love this update too much, look forward including earlier data
新建位图图像

Loreos7:
@cl8n Could you also add localisation for the string mentioned above?

PR #9391, opened 2022-10-20
Show guest mapper name on beatmap page

nanaya:
Adding a (tiny) mapper name next to difficulty name if it's a guest map. Not sure about the wording. There's "Guest difficulty by ..." for discussion page but the new design just says "mapped by ...". I feel if using the discussion page wording it should at least be lowercased here. The same wording will probably be used on other places as well (beatmap card, score info, etc; [example](https://s.myconan.tk/2022-10/Figma_2022-10-20_20-15-31.png)).

The info is gone when hovering other difficulty at the moment (replaced with star difficulty text).

This is preparing for updated beatmap page design as it'll incorporate the same info. Added earlier here just to reduce change on backend. <sub>and god knows when that one will actually happen</sub>


notbakaneko (approved):
thought I reviewed this already 🤔

PR #9394, opened 2022-10-20
Fix various code quirks

peppy:
These are some misc changes I had lying around for some months now. I was intending on creating a larger changeset and adding proper CI coverage, but never got around to doing that.

Figure there's no harm in fixing up these issues I came across?


app/Console/Commands/BuildsCreate.php
```diff
@@ -34,15 +34,17 @@ public function handle()
         $streamId = get_int($this->option('stream-id'));
 
         if ($streamId === null) {
-            return $this->error('Missing stream-id');
+            $this->error('Missing stream-id');
+            return;
```
notbakaneko, line 38: I'm not sure about these, we do this to have the `return` and call inline when the return types are the same (usually `void`)

peppy: Yeah, I did this following one of the inspections provided by whatever I was looking at (and because of my c# inclining - this would be a syntax error there) but feel free to reject any of the changes.

nanaya: The expected return type for `handle` function is int anyway (exit code) so this change is probably fine.

notbakaneko (changes requested):

app/Console/Commands/BuildsCreate.php
notbakaneko: There's still the others like `return $this->markRead(...)` where the enclosing function returns `void`. 

Also, `handle` doesn't have a return type, `Command::execute` casts it to `int`- if we're going to start treating the return type of `handle` itself as `int` then we should be using `return 0` everywhere instead.

If we want to start disallowing inline return voids, then we should enable the corresponding `@typescript-eslint/no-confusing-void-expression` for eslint as well.

notbakaneko (approved):

PR #9401, opened 2022-10-21
Declare missing instance variables

notbakaneko:
Explicitly declare used instance variables.

Some cleanups in the process of adding static analysis. 
This doesn't bother with fixing the magic laravel stuff.

### DIFF
--- a/app/Events/Fulfillments/PaymentEvent.php
+++ b/app/Events/Fulfillments/PaymentEvent.php
@@ -11,7 +11,6 @@ class PaymentEvent implements HasOrder
 {
     public function __construct(public Order $order)
     {
-        $this->order = $order;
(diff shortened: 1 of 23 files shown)


app/Events/Fulfillments/PaymentEvent.php
```diff
@@ -9,7 +9,7 @@
 
 class PaymentEvent implements HasOrder
 {
-    public function __construct(Order $order)
+    public function __construct(public Order $order)
     {
         $this->order = $order;
```
nanaya, line 14: the assignment seems redundant

nanaya (changes requested):

nanaya (approved):

PR #9409, opened 2022-10-24
Add default build id option

### LINKED ISSUE #9404: Add a flag to force a `build_id` on new scores for staging environments
In cases where we don't require hash validation, no `build_id` is associated with new scores. With the recent indexing changes, this results in the score not showing up, as they are treated as legacy scores.

We probably need a way to specify a `build_id` to use in these cases, for the staging environment.

See [discussion in #osu](https://discord.com/channels/188630481301012481/188630652340404224/1033370868639858708) for more context.

nanaya:
Resolves #9404. Although considering `build_id` is now required for solo score to not be considered legacy I wonder if there should always be a default value (0?).


peppy:
> Although considering `build_id` is now required for solo score to not be considered legacy I wonder if there should always be a default value (0?).

I think I'd feel more comfortable with the request erroring if a build can't be matched..

nanaya:
There's already an error if valid build is required.

notbakaneko (approved):
note that admins submitting invalid `build_id` will now have `default_build_id` applied instead

nanaya:
Actually in that case (and without the default id set), the score will be slightly invalid (considered legacy) 🤔

notbakaneko:
It's not like invalid builds would have had `build_id` set if the validation was disabled 👀

nanaya:
@peppy without default value, the `DEFAULT_BUILD_ID` will need to be set in production as well otherwise admins submitting score with invalid hash have the score marked as legacy as well

notbakaneko (approved):

Issue #9405, opened 2022-10-23
Replies on map discussion page duplicate themselves into the "Type Your Response Here" field


NevarrenF:
So for about the last month or so I and a few other mappers have noticed that when you reply to a post sometimes that reply duplicates itself into the new response field.

![image](https://user-images.githubusercontent.com/62331647/197370130-4089bcb5-726f-4907-b6d4-bec008a2ff99.png)
![image](https://user-images.githubusercontent.com/62331647/197370136-462fc801-0488-4828-847e-1e55859bb49b.png)

here are the threads where it effects me personally.
https://osu.ppy.sh/beatmapsets/1821864/discussion
https://osu.ppy.sh/beatmapsets/1352410/discussion

Sorry I don't have much more info because it doesn't happen on every thread, my browser is Chrome.

nanaya:
is this still happening?

NevarrenF:
I personally haven't noticed it happening recently.

nanaya:
Cool. Feel free to comment here or open a new one if happens again.

PR #9410, opened 2022-10-24
Clean up discussion reply handling

nanaya:
May or may not resolve issue mentioned in #9405. I couldn't reproduce it but updating the storage synchronously seem to make more sense.

Also only focus input box after pressing the reply button and nothing else (posting reply on another open box).

### DIFF
--- a/resources/assets/lib/beatmap-discussions/new-reply.tsx
+++ b/resources/assets/lib/beatmap-discussions/new-reply.tsx
@@ -79,6 +79,7 @@ export class NewReply extends React.Component<Props> {
     this.handleKeyDown = makeTextAreaHandler(this.handleKeyDownCallback);
   }
 
+  @action
   componentDidUpdate(prevProps: Readonly<Props>) {
     if (prevProps.discussion.id !== this.props.discussion.id) {
       this.message = this.storedMessage;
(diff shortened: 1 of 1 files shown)


resources/assets/lib/beatmap-discussions/new-reply.tsx
```diff
@@ -253,9 +259,10 @@ export class NewReply extends React.Component<Props> {
   }
 
   @action
-  private readonly setMessage = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
-    this.message = e.target.value;
-  };
+  private setMessage(message: string) {
+    this.message = message;
+    this.storeMessage();
+  }
```
notbakaneko, line 265: I've since discovered if a getter is decorated as `computed`, the corresponding setter automatically runs as an action without needing the decoration 👀

nanaya: but there's no getter here 🤔

notbakaneko: but there could be <_<

also, `storeMessage` should be rolled up into `setMessage`

notbakaneko (approved):

PR #9417, opened 2022-10-24
Fix error on duplicated notification options update

nanaya:
This one is using lock for update because it's already in a transaction and involves multiple rows.

Although I wonder if it's better to handle it per option instead of sending everything and updating in a transaction. That'd probably require a larger change though.


app/Http/Controllers/AccountController.php
```diff
@@ -177,15 +177,21 @@ public function updateNotificationOptions()
         }
 
         DB::transaction(function () use ($requestParams) {
-            // FIXME: less queries
+            $user = auth()->user();
+            UserNotificationOption
```
notbakaneko, line 181: `$user->notificationOptions()->....->lockForUpdate()->get()`?

notbakaneko (changes requested):

notbakaneko (approved):

PR #9418, opened 2022-10-24
Use correct number of params when invoking functions

notbakaneko:
`addHour()` etc doesn't accept params, `addHours()`, etc defaults to 1 so the param could optionally be removed; this PR doesn't bother with that and standardizes the calls to only use the plural version (to be more specific, both singular and plural versions actually do the same thing but only the plural version is documented with accepting options).


nanaya (changes requested):
There are still a few `subDay` and `addDay` calls (including a new one)

notbakaneko:
apparently I didn't do the days 🤔

nanaya (approved):

PR #9425, opened 2022-10-27
Preload solo score performance for profile page

nanaya:
If there's nothing to preload (the `whereIn` is empty), laravel will still do a query. A noop query. Built into the framework with [its explicit `0 = 1`](https://github.com/laravel/framework/blob/ec3958de050d14f15832aa7604fb6834efa807bf/src/Illuminate/Database/Query/Grammars/Grammar.php#L292) instead of not bothering doing the query at all.

And so this PR will cause a bunch of ``` SELECT * FROM `solo_scores_performance` WHERE 0 = 1 ``` calls until profile page is migrated to use solo score so the query actually does something.

...so I'm not sure if this PR is useful at all or should just be part of profile page update to use solo scores 🤔

- [x] depends on #9369


notbakaneko:
will be merged with #9427

PR #9427, opened 2022-10-28
Use solo score for profile page (except recent plays)

nanaya:
Just best and first place scores. Recent scores will come later... maybe.

Depends on
- [ ] #9284 (for the solo score deployment)
- [x] #9705


app/Models/Traits/UserScoreable.php
```diff
@@ -5,53 +5,26 @@
...
-                        'min_score_id' => ['min' => ['field' => 'score_id']],
-                    ],
-                ],
-            ]);
-
-        $response = $search->response();
-        $search->assertNoError();
-
-        return $response;
+        return (new FetchDedupedScores('beatmap_id', ScoreSearchParams::fromArray([
```
notbakaneko, line 21: keep the "aggregatedScoresBest_{$mode}" tag for metrics

notbakaneko (changes requested):

notbakaneko (dismissed):
reminder for myself that I've already looked at this

PR #9430, opened 2022-10-30
Show contest vote percentage based on count of users who voted

### LINKED ISSUE #9428: Percentage display for contest entries uses unusual value
As per the above, the display for total voting percentage (4.47% in this image) appears to use a percentage of all votes submitted instead of a percentage of users that actually voted for the map. I'm not actually sure if this is the case, but I hope it at least illustrates the issue in display. The ```contest_id``` for this particular instance is ``154``, but this issue applies to seemingly all contests.

Multiplying the percentage result by the number of votes given to each user (15, in this case) yields a more convincing (albeit probably wrong) answer (4.47 * 15 = 67.05%) which seems to better reflect the proportion of users that voted for something.

cl8n:
resolves #9428


app/Models/Contest.php
```diff
@@ -343,6 +347,11 @@ public function userEntries($user = null)
         );
     }
 
+    public function usersVotedCount(): int
+    {
+        return $this->votes()->distinct()->count('user_id');
```
nanaya, line 352: nonsense query aside as when read normally `distinct count(column)` is a noop (but still works somehow because laravel rewrites it; if anything it should be `distinct('user_id')->count()`), this should probably be cached like `entriesByType` otherwise it may take 40+ms on larger contests

nanaya (changes requested):

nanaya (approved):

PR #9437, opened 2022-10-31
Fix score query scope

nanaya:
...that was weird. Hopefully it doesn't break any tests 👀


notbakaneko (approved):
probably helps the query was empty from the beginning 👀

PR #9442, opened 2022-11-01
Update Channel phpdocs

notbakaneko:
- fix some doc typings
- return type-hint as `static` instead of `self`
- add scopes to phpdoc block
- remove unused method
- remove `hidden` attribute from `Channel`; there's an existing `protected $hidden` on the laravel model
- declare generic type for `Collection`s; this seems to be the syntax tools like phpstan and phpdocumentor recognize

### DIFF
--- a/app/Models/Chat/Channel.php
+++ b/app/Models/Chat/Channel.php
@@ -24,16 +24,20 @@ use Illuminate\Support\Str;
 use LaravelRedis as Redis;
 
 /**
- * @property string[] $allowed_groups
+ * @property int[] $allowed_groups
  * @property int $channel_id
  * @property \Carbon\Carbon $creation_time
+ * @property-read string $creation_time_json
  * @property string $description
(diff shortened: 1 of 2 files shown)


app/Models/Chat/Channel.php
```diff
@@ -24,16 +24,22 @@
...
- * @property \Illuminate\Database\Eloquent\Collection $messages Message
+ * @property-read \Illuminate\Database\Eloquent\Collection<Message> $messages
  * @property int|null $match_id
- * @property int $moderated
+ * @property bool $moderated
  * @property string $name
  * @property int|null $room_id
- * @property mixed $type
+ * @property string $type
+ * @property int|null $last_message_id
```
nanaya, line 38: if this is supposed to be somewhat sorted then this isn't quite sorted

nanaya (changes requested):

nanaya (approved):

PR #9444, opened 2022-11-01
Add basic support for nominated beatmaps profile display

### LINKED ISSUE #7555: Add a "nominated beatmaps" section in the "beatmaps" tab on user profiles
Would be nice if users that are/were BNs/NATs/QATs/BATs/MATs/e.t.c could have a section on the user profile showcasing the beatmaps they have/had nominated for ranked/approval

Something like this would be cool

Kind of ties in with https://github.com/ppy/osu-web/issues/7329 though that one is moreso for NATs/management while this is moreso just to showcase stuff nominated

venix12:
resolves #7555

Works with modding discussion system only for now... perhaps should scrap the nominations for older ones according to old forum icons system? Seems like a thing for different PR though.

Went with:
- ranked only
- only "final" (non-reset) nominations included

which is probably up to debate.


Hiviexd:
I think it makes sense to also include pending and qualified maps that are currently nominated by that user, they clearly fall under the category of "nominated maps".

venix12:
Could be, went with ranked only as a placeholder pretty much.

cl8n:
> I think it makes sense to also include pending and qualified maps that are currently nominated by that user, they clearly fall under the category of "nominated maps".

since this is showing final nominations, I think it's fine as-is -- you don't know if the nominations are "final" until the map is ranked

nanaya (changes requested):

database/migrations/2022_11_12_103015_index_user_id_on_beatmapset_nominations.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('beatmapset_nominations', function (Blueprint $table): void {
+            $table->index('user_id');
```
nanaya, line 20: adding reset as well is probably good idea. Also the rollback should have the columns in an array

nanaya (changes requested):

PR #9446, opened 2022-11-02
Increase username reuse limit

nanaya:
ಠ_ಠ


notbakaneko (approved):
should be fine? 🤔 @peppy

peppy (approved):

PR #9449, opened 2022-11-02
Misc. API token cleanup

cl8n:
testing "invalid" is same thing as testing star so I think that's just easier to read.

delegate scope needs no description because it can't be authorized, also that trans key didn't exist


app/Http/Middleware/RequireScopes.php
```diff
@@ -47,8 +47,7 @@ protected function validateScopes(?Token $token, $scopes)
         }
 
         if (!$this->requestHasScopedMiddleware(request())) {
-            // use a non-existent scope; only '*' should pass.
-            if (!$token->can('invalid')) {
+            if (!$token->can('*')) {
```
notbakaneko, line 50: Shouldn't be changed from intentionally checking a scope that cannot be assigned; it's specifically like this because `can` just treats the presence of `*` on the token as a complete bypass and not a permission check - if there's an underlying issue, `*` will still pass while `invalid` will start failing which is the more desirable scenario.

notbakaneko (changes requested):

app/Http/Middleware/RequireScopes.php
cl8n: I'm not really following, what "underlying issue" could make `*` pass but not `invalid`?

if you meant like a programming error that causes `*` to not be the "bypass" scope anymore, it would be caught by tests already

notbakaneko (approved):
already an newer existing case of `->can('*')` being used

PR #9450, opened 2022-11-02
Remove refresh from `es:index-scores:queue`

cl8n:
:broom:


nanaya:
This test doesn't actually involve es...

cl8n:
I thought it was failing in the same way as others (because I dont have anything for solo scores set up locally). maybe it was some other problem and I didnt read close enough :thinking: will recheck

nanaya:
It looks like the refresh request in the process caused it to require the es. There has been talk about that part isn't making sense there though...

nanaya (approved):

PR #9451, opened 2022-11-02
Show Loved polls on beatmapset pages

### LINKED ISSUE #8202: Improve visibility of polls for beatmaps nominated for Loved
Currently, a typical user needs to go through newspost -> beatmap page -> download + play (if unfamiliar) -> newspost again -> forum thread in order to vote upon just one of the beatmaps nominated for Loved. Given how layered and tedious to navigate the current system is, most users simply don't bother voting for any of these nominations. 

We've recently discussed ways to alleviate this visibility issue and reduce the layers for the better in the `#osu-loved` channel. As a result, we've come up with the following solutions:

- **Enable users to vote directly on the beatmap page.** This is the primary goal we would like to achieve, albeit the most complex one. 
   - Upon pushing the `Love` button the beatmap will be moved to an "In Voting" state, where the Hype Train section will be replaced with Yes and No buttons for users to vote.
   - Prior to the user casting their vote, there should also be a check on whether the user has played the map before.
   - The polls will be closed either manually by a Project Loved Team member or automatically after 2 weeks. Depending on the results, the beatmap will be moved to Loved or fall back to the Pending state.

- **Add a button/banner redirecting users to the already existing forum polls to the beatmap page.** This will keep the layers intact, but still improve the visibility of the polls to some extent.
   - Just a simple button to be manually adjusted by a Project Loved Team member will probably be enough.   
   - As much as it is easier to implement, I personally believe it creates more burdens on the shoulders of the team members rather than improving their lives. 

There is also #7641 that I believe can further enhance the visibility of Project Loved quite a bit, which can be perceived as an alternative to the aforementioned suggestions.

---

I think I've compiled all the ideas discussed in the channel, but pinging @huuishuu and @cl8n just in case I forgot something.

cl8n:
all working but draft cuz I should probably add some tests for creating/deleting polls. maybe tests for the transformer too

for management, this also adds special `loved` API scope that *can't* be implied by `*` and needs to be requested by a designated client

resolves #8202

### DIFF
--- a/app/Http/Controllers/BeatmapsetsController.php
+++ b/app/Http/Controllers/BeatmapsetsController.php
@@ -19,6 +19,7 @@ use App\Models\BeatmapsetEvent;
 use App\Models\BeatmapsetWatch;
 use App\Models\Genre;
 use App\Models\Language;
+use App\Models\LovedPoll;
 use App\Transformers\BeatmapsetTransformer;
 use Auth;
 use Carbon\Carbon;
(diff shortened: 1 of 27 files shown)


notbakaneko (changes requested):
What's this `loved` client for, where and how is it going to be used?

Also, all the OAuth-related parts should be separated into a different PR.

cl8n:
@notbakaneko I put the oauth changes in #9584 , that's ready for review.

I'm gonna mark this one as draft and revisit / probably split it into more manageable parts after figuring out the oauth part.

PR #9461, opened 2022-11-04
Fix inconsistent verification logic when bypassing

### LINKED ISSUE #9460: Sending chat messages with bypassed verification brings up verification page
In `.env`:

```
USER_BYPASS_VERIFICATION=true
USER_POST_ACTION_VERIFICATION=false
```
No verification screen anywhere (including user settings) until actually hitting send on a new message. Since there's no email server, I can't receive the verification code.

@nanaya suggested to grab the verification message from `laravel.log`, but inputting that in with the env variable still set to `true` leads to:
- The (previous) message failed to send (red cross/grayed out message).
- Sending a new message pops up the verification screen again.

The only way to make it work is by setting the .env variable to `false` and inputting the verification code. Which is okay now that I know all of the above, but quite a series of events for anyone new to the project to gather.

cl8n:
isSessionVerified fixes #9460

the middleware change lets the notifications server use this setting too (but it's only updated on user login when bypassing, didn't bother to fix for this PR)

### DIFF
--- a/app/Http/Middleware/VerifyUserAlways.php
+++ b/app/Http/Middleware/VerifyUserAlways.php
@@ -12,7 +12,9 @@ class VerifyUserAlways extends VerifyUser
 {
     public static function isRequired($user)
     {
-        return $user !== null && ($user->isPrivileged() || $user->isInactive());
+        return !config('osu.user.bypass_verification')
+            && $user !== null
+            && ($user->isPrivileged() || $user->isInactive());
     }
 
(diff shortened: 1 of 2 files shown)


nanaya:
The way verification is handled can use more fixes than this especially as `isSessionVerified` has been added to User model (by setting it early instead of repeatedly checking it from session and then skipping verification check altogether early when session is already verified).

cl8n:
#9467 :tada:

PR #9464, opened 2022-11-06
Add group history page

### LINKED ISSUE #6216: Record and show group changes
an on-site replacement for https://osu.ppy.sh/help/wiki/Staff_Log would be nice. it's a pain to keep these things updated when they could be done automatically, and filters for showing only some groups at a time would make it easier to find the info you're looking for

to keep it simple I don't think "moves" need to be recorded in the way they're presented on Staff Log. just additions and removal is fine

cl8n:
resolves #6216

### DIFF
--- a/database/migrations/2022_11_06_022852_require_user_group_events_details_and_type.php
+++ b/database/migrations/2022_11_06_022852_require_user_group_events_details_and_type.php
@@ -1,0 +1,61 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+use App\Models\UserGroupEvent;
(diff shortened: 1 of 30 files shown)


database/migrations/2022_11_06_022852_require_user_group_events_details_and_type.php
```diff
@@ -0,0 +1,61 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        // The format of `details` before this was not finalised, and on
+        // production nothing important was logged anyway. No going back!
```
cl8n, line 22: can someone check if this is true :eyes: 
basically I am expecting `select count(*) from user_group_events where group_id <> 2 or type <> 'user_add'` to be 0.

nanaya:
please split into smaller PRs

cl8n:
has been mostly replaced

PR #9465, opened 2022-11-06
Add account setting to link GitHub account

cl8n:
so open source contributors can link themselves on changelogs


app/Http/Controllers/Account/GithubUsersController.php
```diff
@@ -0,0 +1,88 @@
...
+        ]);
+
+        abort_unless(isset($params['code']), 422, 'Invalid code.');
+        abort_unless(
+            isset($params['state']) && $params['state'] === session()->pull('github_auth_state'),
+            403,
+            'Invalid state.',
+        );
+
+        $tokenResponseBody = (new HttpClient())
```
nanaya, line 39: just use oauth2-github library

app/Http/Controllers/Account/GithubUsersController.php
```diff
@@ -0,0 +1,88 @@
...
+
+        abort_if($token === null, 500, 'Invalid response from GitHub API.');
+
+        $githubClient = new GithubClient();
+        $githubClient->authenticate($token, null, GithubClient::AUTH_ACCESS_TOKEN);
+        $githubApiUser = $githubClient->currentUser()->show();
+
+        GithubUser::importFromGithub($githubApiUser, auth()->user());
+
+        return redirect(route('account.edit').'#github');
```
nanaya, line 60: I'm not sure about allowing random people to link their github account as opposed to just the ones with actual changelog entry (existing GithubUser from commit callback)

app/Models/GithubUser.php
```diff
@@ -3,75 +3,120 @@
...
+    public function user(): BelongsTo
     {
         return $this->belongsTo(User::class, 'user_id');
     }
 
-    public function changelogEntries()
+    public function scopeWithGithubInfo(Builder $query): void
     {
-        return $this->hasMany(ChangelogEntry::class);
+        $query->whereNotNull(['canonical_id', 'username']);
```
nanaya, line 75: looking at actual data, there's no entry with either set to null

app/Http/Controllers/AccountController.php
```diff
@@ -113,10 +114,20 @@ public function edit()
 
         $notificationOptions = $user->notificationOptions->keyBy('name');
 
+        if (GithubUser::canAuthenticate()) {
+            $githubUsers = json_collection(
+                $user->githubUsers()->withGithubInfo()->get(),
```
nanaya, line 119: I'm not sure why I thought multiple github users could be good idea but there's no such thing and I don't think it needs to be supported. It's especially weird when there's actual user interface to link the account(s?).

.env.example
```diff
@@ -104,6 +104,10 @@ PUSHER_SECRET=
 
 # GITHUB_TOKEN=
 
+# GitHub client for users to associate their GitHub accounts
```
nanaya, line 107: as I learned recently, this should probably also mention where to point the callback url (either here or in setup documentation)

nanaya (changes requested):

app/Models/GithubUser.php
cl8n: are there any with a blank username? there's some function that looks like it was trying to handle blank or null username

nanaya: nope

app/Http/Controllers/AccountController.php
cl8n: there are at least a few people that have used multiple GitHub accounts on osu-wiki for whatever reason. I agree it's not important but it was easy to support...

nanaya: how many are there? And what's the reason? That just seems confusing.

cl8n: very few. I don't know either lol. anyway looking at it again, yeah it does seem odd to be able to set multiple, I'll make the account page only support linking one

cl8n:
oh right I forgot I broke legacy changelog too. i will fix that tomorrow :zzz:

cl8n:
... I had a long day :sweat_smile: 

I reverted some of the changes I made to the `GithubUser` model so legacy changelogs can work again (by making fake `GithubUser` with most attributes missing), and added some notes about the potentially-confusing typing

nanaya (changes requested):

peppy:
We were discussing this IRL and not sure how much this makes sense. It's going to end up with a lot of non-developers linking accounts just because the option is there... and the number of contributors over the years is so low we can probably just handle this manually?

It feels like a bit of a waste since this is already implemented though, so curious to hear your thoughts @cl8n.

cl8n:
it only lets you link the github account if it has any changelog entries associated. but I could make that more clear in the display as well (https://github.com/ppy/osu-web/pull/9465#discussion_r1241772028)

you probably *could* handle it manually but I don't see why you'd want to... from user perspective I think people generally don't want to bother admins with requests like this either; there's nothing stopping people from asking you to link their account right now but nobody seems to have done that.

(and I will apply last review soon I have a lot of osu catch-up to get through ;_ ; )

peppy:
I guess it's okay if it's only for people with contributions (although I don't know how that's checked.. is it post auth?).

It should be shown under the dev section of the account settings so most users don't see it.

app/Http/Controllers/Account/GithubUsersController.php
```diff
@@ -0,0 +1,97 @@
...
+        $token = $provider->getAccessToken('authorization_code', [
+            'code' => $params['code'],
+        ]);
+        $client->authenticate($token->getToken(), null, GithubClient::AUTH_ACCESS_TOKEN);
+        $apiUser = $client->currentUser()->show();
+        $user = GithubUser::firstWhere('canonical_id', $apiUser['id']);
+
+        abort_if($user === null, 422, osu_trans('accounts.github_user.error.no_contribution'));
+
+        abort_if($user->user_id !== null, 422, osu_trans('accounts.github_user.error.already_linked'));
```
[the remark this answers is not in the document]
cl8n: that check happens in `create()`

peppy:
@cl8n is this ready for re-review?

nanaya (changes requested):

app/Http/Controllers/Account/GithubUsersController.php
```diff
@@ -0,0 +1,125 @@
...
+        return redirect(route('account.edit').'#github');
+    }
+
+    public function create()
+    {
+        abort_unless(GithubUser::canAuthenticate(), 404);
+        abort_if(
+            auth()->user()->githubUser()->exists(),
+            422,
+            'Cannot link more than one GitHub account.',
```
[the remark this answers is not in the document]
cl8n: for this one, I changed it so that you get redirected back to the account page (where you should see your existing github account if you didn't already for whatever reason)

for the other untranslated error messages, I left them like that because they should never be seen if not purposely messing with the request from github... seems kinda pointless to translate?

nanaya (approved):

PR #9483, opened 2022-11-08
Remove nulls before whereIn query in comment bundle

nanaya:
Also:
- combine parent fetching on null commentable and reply request
- skip querying on empty ids
- use Set for reject lookup
- don't reconcat `$allComments` with `$comments`

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -19,6 +19,18 @@ function api_version(): int
     return $version;
 }
 
+function array_filter_set(array|ArrayAccess $array): array
+{
+    $ret = [];
+    foreach ($array as $item) {
+        if ($item !== null) {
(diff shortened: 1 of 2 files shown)


app/helpers.php
```diff
@@ -19,6 +19,18 @@ function api_version(): int
     return $version;
 }
 
+function array_filter_set(array|ArrayAccess $array): array
```
notbakaneko, line 22: should probably call this `array_filter_isset` or `array_reject_null` or something else; currently it sounds like this does something with Sets or is assigning something

notbakaneko (changes requested):

notbakaneko (approved):

PR #9486, opened 2022-11-09
Clean up `UserGroupEvent` model

cl8n:
part 1 of #9464


app/Models/UserGroupEvent.php
```diff
@@ -102,18 +116,42 @@ private static function log(?User $actor, string $type, ?User $user, Group $grou
...
+        return match ($key) {
+            'actor_id',
+            'group_id',
+            'id',
+            'type',
+            'user_id' => $this->getRawAttribute($key),
+
+            'details' => json_decode($this->getRawAttribute($key), true),
+
+            'hidden' => (bool) $this->getRawAttribute($key),
```
nanaya, line 145: this should be its own function to avoid confusion because the class already have this attribute ([similar to this](https://github.com/ppy/osu-web/blob/930a4c8a8a9b8fdc0f21f8be5b39afb80b93e49f/app/Models/Chat/UserChannel.php#L50)).

(other models with this column should also be updated eventually - although a better check against this column is needed, especially for assignment which there's none at the moment)

app/Models/UserGroupEvent.php
```diff
@@ -102,18 +116,42 @@ private static function log(?User $actor, string $type, ?User $user, Group $grou
...
+    public function getAttribute($key)
+    {
+        return match ($key) {
+            'actor_id',
+            'group_id',
+            'id',
+            'type',
+            'user_id' => $this->getRawAttribute($key),
+
+            'details' => json_decode($this->getRawAttribute($key), true),
```
nanaya, line 143: the column is nullable and json_decode will throw error here because it requires string

nanaya (changes requested):

nanaya (approved):

PR #9494, opened 2022-11-10
Check beatmapset existence for score listing

nanaya:
Apparently recent scores aren't cleaned up when beatmapset is deleted.


notbakaneko (approved):
That's some really long query...

PR #9498, opened 2022-11-10
Show nominators on beatmap info page

### LINKED ISSUE #7008: Show nominators on qualified/ranked beatmapset listing page
The nominators for a given set should appear on the listing page when it hits qualified or ranked. For example, https://osu.ppy.sh/beatmapsets/1223737#osu/2545340 should show that Karen and Milan- are responsible for nominating it somewhere on the main page instead of needing to click through into discussions.

Why?

Currently, nominating a map is fairly transactional - there is little prestige or acclaim involved for the people putting the sets forward themselves, knowing full well that their names are essentially hidden behind a clickthrough that 90%+ of players will not ever use.

Associating nominators closely with a set allows them to use the types of work they push as a kind of social "badge of honor". Nominators that push consistently good sets will gain reputation along with the mappers they help elevate, while nominators that consistently push poor sets will gain the appropriate notoriety for their decisions as well. This kind of push-pull effect is very important to have in communities as a form of social self-regulation. 

This is otherwise a very minor design change that will likely have measurable (if slight) impact on map quality and the health of the modding scene going forward. Please consider it.

nanaya:
The placement seems a bit wrong (and makes the box taller) but that's the only itemised listing area.

Resolves #7008

### DIFF
--- a/app/Http/Controllers/BeatmapsetsController.php
+++ b/app/Http/Controllers/BeatmapsetsController.php
@@ -381,6 +381,7 @@ class BeatmapsetsController extends Controller
             'beatmaps.max_combo',
             'converts',
             'converts.failtimes',
+            'current_nominations',
             'current_user_attributes',
             'description',
             'genre',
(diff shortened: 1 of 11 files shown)


notbakaneko (approved):

AJT127:
In light of this being added, it might be nice to add a subtle search feature on the website that’s like =“nominatedby=username” in similar fashion to the other search functions, not sure if possible though?

PR #9502, opened 2022-11-11
Convert reloadPage and popup to typescript

notbakaneko:
Wasn't going to move these to modules originally as they're still used by blade views, but figured might as well do it and shim the functions for now 🤷

### DIFF
--- a/resources/assets/coffee/osu_common.coffee
+++ b/resources/assets/coffee/osu_common.coffee
@@ -4,7 +4,6 @@
 import { formatNumber } from 'utils/html'
 import { present } from 'utils/string'
 import { popup } from 'utils/popup'
-import { currentUrl, navigate, reloadPage } from 'utils/turbolinks'
 
 window.osu =
   trans: (key, replacements = {}, locale) ->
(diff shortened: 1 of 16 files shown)


nanaya:
maybe export them from either `main.coffee` or `app-deps.ts`

resources/assets/lib/utils/popup.ts
```diff
@@ -0,0 +1,36 @@
...
+      .one('click.close-alert', closeAlert)
+      .fadeIn();
+  } else {
+    setTimeout(closeAlert, 5000);
+  }
+
+  const activeElement = document.activeElement;
+  if (activeElement instanceof HTMLElement) {
+    activeElement.blur();
+  }
```
nanaya, line 33: can use `htmlElementOrNull` here

resources/assets/coffee/osu_common.coffee
```diff
@@ -3,42 +3,10 @@
 
 import { formatNumber } from 'utils/html'
 import { present } from 'utils/string'
-import { currentUrl, navigate } from 'utils/turbolinks'
+import { popup } from 'utils/popup'
+import { currentUrl, navigate, reloadPage } from 'utils/turbolinks'
 
 window.osu =
-  reloadPage: (keepScroll = true) ->
```
nanaya, line 9: `globals.d.ts` still contains the definitions

nanaya (changes requested):

nanaya (approved):

PR #9503, opened 2022-11-11
Mark recent_messages as deprecated

notbakaneko:
This was deprecated long ago; osu-web doesn't use it (it's not even in `ChannelJson`) and only kept around for compatibility with https://github.com/ppy/osu/blob/4bc26dbb487241e2bbae73751dbe9e93a4e427da/osu.Game/Online/API/Requests/Responses/APIChatChannel.cs#L17-L18 having a reference to it

### DIFF
--- a/resources/views/docs/info.blade.php
+++ b/resources/views/docs/info.blade.php
@@ -26,7 +26,7 @@ For a full list of changes, see the
 ## Breaking Changes
 
 ### 2022-11-11
-- `recent_messages` in [ChatChannel](#chatchannel) has been deprecated, it will be removed from [Create Channel] (#create-channel) response in the near future.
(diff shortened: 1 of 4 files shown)


resources/views/docs/info.blade.php
```diff
@@ -25,6 +25,9 @@
 
 ## Breaking Changes
 
+### 2022-11-11
+- `recent_messages` in [ChatChannel](#chatchannel) has been deprecated, it will be removed from [Create Channel] (#create-channel) response in the near future.
```
nanaya, line 29:
```suggestion
- `recent_messages` in [ChatChannel](#chatchannel) has been deprecated, it will be removed from [Create Channel](#create-channel) response in the near future.
```

nanaya (changes requested):

nanaya (approved):

PR #9510, opened 2022-11-14
Return recent silences with chat ack

notbakaneko:


nanaya:
Hmm, `recent` may potentially be more expensive in term of query due to `OR permanent = 1` 🤔

notbakaneko:
maybe that was why it wasn't added in the first place 🤔

app/Http/Controllers/Chat/ChatController.php
```diff
@@ -313,22 +313,7 @@ public function updates()
...
-                    $silenceQuery->none();
-                } else {
-                    $silenceQuery->where('timestamp', '>', $previousMessage->timestamp);
-                }
-            } else {
-                $silenceQuery->where('ban_id', '>', $lastHistoryId)->reorderBy('ban_id', 'DESC');
-            }
-
-            $silences = $silenceQuery->get();
+            $silences = $this->getSilences($params['history_since'], $params['since']);
```
nanaya, line 316: there's already `$since` which only used once after this change. Maybe use either and not both

nanaya (approved):

nanaya:
There's also another problem of this potentially still returning a lot of expired silences... Should be fine as long whatever doing the removal checks the length. It also seems impossible to filter them out using query without table structure change anyway.

PR #9522, opened 2022-11-17
Convert translation funcions to typescript

notbakaneko:
That's a lot of calls 🥵 

Includes retyping of lang.js calls to match our current usage of expecting to be able to pass in `number | null | undefined` and have them implicitly converted to `string`.


resources/assets/lib/utils/lang.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export function trans(key: string, replacements: Replacements = {}, locale?: string) {
+  if (!transExists) {
+    locale = fallbackLocale;
+  }
+
+  return window.Lang.get(key, replacements, locale);
+}
+
+export function transArray(array: Replacement[], key = 'common.array_and') {
```
notbakaneko, line 22: Come to think of it, `transArray` doesn't actually do any translation stuff 👀

resources/assets/lib/utils/lang.ts
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import Lang from 'lang.js';
+import { formatNumber } from 'utils/html';
+import { present } from 'utils/string';
+
+type Replacement = string | number | null | undefined;
+
+export interface Replacements {
```
nanaya, line 10: it seems possible to just use the original typing by not setting nulls/undefined in the first place and formatting numbers beforehand (formatting them seems like a good idea anyway).

The most annoying one seems to be `formatTimestamp` but the function itself can probably be split to version that only accept number (if only conditional generic return is still supported in typescript...).

resources/assets/lib/utils/lang.ts
nanaya, line 22: this should only accept `string|number` (or maybe even just `string`) in the first place...

nanaya (changes requested):

resources/assets/lib/utils/lang.ts
notbakaneko: actually, I think `formatTimestamp` might not need to handle `null` at all...

nanaya:
conflicts

PR #9523, opened 2022-11-18
Support joining component into string sentences

notbakaneko:
....or something...

A version of `transArray` that supports components so they can be used directly instead of stringifying them as html strings

- [x] #9522


nanaya:
Conflicts

resources/assets/lib/utils/lang.tsx
```diff
@@ -2,12 +2,33 @@
...
+    case 1:
+      return array[0];
+    case 2:
+      return <>{array[0]}{trans(`${key}.two_words_connector`)}{array[1]}</>;
+  }
+
+  const nodes = [array[0]];
+  const lastIndex = array.length - 1;
+  for (let i = 1; i < lastIndex - 1; i++) {
+    nodes.push(<React.Fragment key={i}>{trans(`${key}.words_connector`)} {array[i]}</React.Fragment>);
```
nanaya, line 25: this will cause key conflict if the first element (which has key specified by the caller) has key value of 1. Similarly for the last item below. It's probably better to just explicitly set all the keys in this function itself instead. 

Also may want to wrap the return with key-less fragment so the caller doesn't need to make sure adjacent components have the expected keys.

nanaya (changes requested):

resources/assets/lib/beatmap-discussions/new-discussion.tsx
```diff
@@ -376,26 +375,23 @@ export class NewDiscussion extends React.Component<Props> {
     if (this.nearbyDiscussions.length === 0) return;
     const currentTimestamp = this.timestamp != null ? formatTimestamp(this.timestamp) : '';
     const timestamps = this.nearbyDiscussions.map((discussion) => (
-      linkHtml(
-        BeatmapDiscussionHelper.url({ discussion }),
-        formatTimestamp(discussion.timestamp) ?? '',
-        { classNames: ['js-beatmap-discussion--jump'] },
-      )
+      <a
+        key={discussion.id}
```
nanaya, line 379: keys not needed anymore?

nanaya (changes requested):

resources/assets/lib/beatmap-discussions/new-discussion.tsx
notbakaneko: `key` is still needed to pass linting (also probably shouldn't be assuming another component is splitting up arrays to not need keys)

nanaya: I wonder if it's better to pass the key separately instead (passing as map or array) 🤔 that way the join function can use it and allow the react render optimization thing

resources/assets/lib/utils/lang.tsx
```diff
@@ -2,12 +2,32 @@
...
+  const nodes: React.ReactFragment[] = [];
+
+  if (array.length > 0) {
+    nodes.push(<React.Fragment key={0}>{array[0]}</React.Fragment>);
+
+    const lastIndex = array.length - 1;
+    const lastConnector = lastIndex === 1 ? trans(`${key}.two_words_connector`) : trans(`${key}.last_word_connector`);
+
+    for (let i = 1; i < lastIndex; i++) {
+      nodes.push(<React.Fragment key={i}>{trans(`${key}.words_connector`)}{array[i]}</React.Fragment>);
```
nanaya, line 22: shouldn't really matter much but the words connector can be stored in a variable before the loop

notbakaneko: I think getting rid of the `key` param altogether would be better since we don't use it at all 👀

nanaya: but that's not related to my comment 🤔 (`const wordsConnector = trans(...)` instead of calling the same function and parameter repeatedly)

resources/assets/lib/utils/lang.tsx
```diff
@@ -2,12 +2,33 @@
...
 import { present } from 'utils/string';
 
 type Replacement = string | number;
 export type Replacements = Partial<Record<string, Replacement>>;
 
+export function joinComponents(array: React.ReactElement[], key = 'common.array_and') {
+  const nodes: React.ReactFragment[] = [];
+
+  if (array.length > 0) {
+    nodes.push(<React.Fragment key={array[0].key}>{array[0]}</React.Fragment>);
```
nanaya, line 16:
```suggestion
    nodes.push(array[0]);
```

nanaya (approved):

PR #9530, opened 2022-11-22
Typescript Modal component

nanaya:
Relatively straightforward conversion. Added `syncVisible` to combine the two visible props dom sync.


notbakaneko (approved):
looks right 🤔

PR #9537, opened 2022-11-24
Hot fix solo scores leaderboard when extra scores need to be fetched

notbakaneko:
Change in #9527 means getting extra scores for deduping needs passing in a extra param in the cursor which is expected to be boolean by the value returned for the cursor is an integer....

Currently the part handling the cursor doesn't have enough info about the parameter to perform any casting.

### DIFF
--- a/app/Libraries/Score/FetchDedupedScores.php
+++ b/app/Libraries/Score/FetchDedupedScores.php
@@ -31,6 +31,8 @@ class FetchDedupedScores
 
         while ($hasNext) {
             if ($nextCursor !== null) {
+                // FIXME: cursor value returned is 0/1 but elasticsearch expects false/true...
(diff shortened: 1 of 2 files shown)


app/Libraries/Score/FetchDedupedScores.php
```diff
@@ -31,6 +31,8 @@ public function all(): array
 
         while ($hasNext) {
             if ($nextCursor !== null) {
+                // FIXME: cursor value returned is 0/1 but elasticsearch expects false/true...
```
nanaya, line 34: 0/1 works but only if it's actual int instead of casted to string. Maybe look into removing the casting in `getSortCursor` especially as `cursor_string` is now the recommended way to pass around cursors and thus isn't affected by javascript limitations

nanaya (approved):

PR #9549, opened 2022-11-28
Make deleted forum posts to searchable for moderation

### LINKED ISSUE #5978: Show deleted forum posts in user's post history
GMT/NAT should be able to see deleted forum posts in a user's post history as it is useful for moderation. 

Perhaps they should also be shown on the [search page](https://osu.ppy.sh/home/search?mode=forum_post) (similar to how it's done for modding posts), though I don't how many moderators would actually need that.

This is also similar to #4982, but I'm not sure how similar the fixes are.

notbakaneko:
Include deleted forum posts to be be searchable by moderators.

Needs a reindex with new schema for the deleted posts to be added; The search itself should still work fine without the update, however.

closes #5978


nanaya:
this probably should also extend the indexing to include all forums as sometimes deleting means moving to other forum (as mentioned in the references and related issues)

notbakaneko:
Forums where the posts should be searchable should just have `enable_indexing = 1` set; the query sent to elasticsearch should already be filtering based on the `ForumView` permission.

nanaya:
as the goal is to allow mods to search everything then might as well drop the check for indexing...?

notbakaneko:
some of the forums like the one for user profile bbcode aren't meant to be searchable and seems like they would be adding a lot of noise if they started showing up in forum post search results

There's also some public forums which aren't indexed for whatever reason 🤷

nanaya (changes requested):

nanaya:
conflicts 👀

nanaya (approved):

PR #9561, opened 2022-11-29
Separate report form from its button

nanaya:
This allows the menu (and user card if its on the card) to be hidden when the form shows up.

This makes showing user card on modal (e.g. beatmap owner editor) possible. That said, it still needs additional handling to hide the original modal before rendering the form.

The way it's done is probably non-react though 👀

Reopening the form after submitting now shows a new form instead of a thank you message. Not sure if it was a feature or bug 🤔 I could technically add a list of reported things to fix that (and as a bonus it'll carry over as long the page isn't reloaded).


notbakaneko (approved):
I feel like keeping the submitted state was a UX thing...even though it just goes away when you refresh the page, anyway 🤔 

This made #9598 issue a bit more obvious 👀

PR #9568, opened 2022-11-30
Remove visible props from modal

nanaya:
Everything using it always passes it true.


resources/assets/lib/beatmap-discussions/nominations.coffee
```diff
@@ -543,7 +543,7 @@ export class Nominations extends React.PureComponent
   renderLoveBeatmapModal: =>
     return if !@state.loveBeatmapModal
 
-    el Modal, visible: true, onClose: @handleLoveBeatmapModal,
+    el Modal, onClose: @handleLoveBeatmapModal,
```
notbakaneko, line 546: not related but it seems odd this one closes when clicking outside the modal but the beatmap owner one doesn't 🤔

notbakaneko (approved):

resources/assets/lib/beatmap-discussions/nominations.coffee
nanaya: I think it's to avoid accidentally close the modal. Although it can be fixed by adding prompt, I guess 🤷‍♀️

PR #9569, opened 2022-11-30
Clean up popup component

nanaya:
This component looks like something that can (or should) be used in Modal and others 🤔 

This PR just clean things up.


resources/assets/lib/components/portal.tsx
```diff
@@ -3,41 +3,33 @@
...
-    if (this.container.parentElement === document.body) {
-      document.body.removeChild(this.container);
-    }
-  };
+  render() {
+    return createPortal(this.props.children, this.container);
+  }
 
-  render = () => createPortal(this.props.children, this.container);
+  private readonly addPortal = () => document.body.appendChild(this.container);
```
notbakaneko, line 30: Can use update to `window.newBody` otherwise it mounts to the previous version instead of the rendered one when navigating backwards

notbakaneko (changes requested):

notbakaneko (approved):

PR #9573, opened 2022-12-01
Use Portal component to handle portal needs

nanaya:
Seems mostly working 🤔 This saves repeating the same portal mounting logic.

Depends on:
- [ ] #9569 
- [ ] #9568 (mostly because it updates the same file)


notbakaneko (changes requested):
image

That doesn't look right 👀  (beatmap search and go back)

Also it leaves a bunch of empty `div`s now when the modals aren't showing, not sure if this part is weird...

Issue #7351, opened 2021-03-21
[Proposal] A GraphQL API implementation


LavaDesu:
Endpoints like `/users/{user}` and `/beatmapsets/{beatmapset}` return huge amounts of data that most clients probably never need.

For instance, `beatmapsets.recent_favourites` can take up >50% of the total json response, but there are very rare cases where you actually need that as an API consumer.

A GraphQL API can alleviate issues like this on both the client and server by transferring and processing only the necessary data, while also allowing complex yet efficient queries that would otherwise require multiple REST requests.

LavaDesu:
If the devs are interested, I could try my hand at implementing this. I made a proof of concept on [my fork](/LavaDesu/osu-web) which allows a simple query of a user by id.

peppy:
@LavaDesu thanks for the proof of concept! are you able to do some benchmarks (just using `ab`) against a normal API request when only requesting a small subset of data? I'd be interested to know how much performance improvement this provides, as that would be the main reason we would consider taking this on.

LavaDesu:
I would like to say that the benefits of implementing a GraphQL API is not only performance, but also the flexibility that comes with it.
Theoretically, you would be able to do something like
- Get the names and scores of the 20 latest ranked beatmaps
- Get 3 different users' usernames
- Get a paginated list of every open multiplayer room

and all that can be done in one single request, which would be faster than multiple REST requests due to overhead.
It is also strongly-typed in an easy-to-read schema, and can throw errors if the server sends unconforming data (such as sending null for a non-nullable field).

But here are the "benchmarks", unfortunately there were a few problems with it:
- I'm not very well-versed in PHP so I wasn't really sure how to benchmark it. As such, I just did it with `curl --write-out '%{time_total}'`.
- The hardware running my local osu-web instance is very old, so everything is on a spinning disk.
- The GraphQL schema in the poc I linked is not a 1:1 implementation of the REST API. For instance, i've replaced a few fields that normally was "_count" in the REST API with a full, paginated list of objects.
- I have not taken a deep look into optimising performance just yet. For example, as shown in the database queries there seems to be auth checks that are done multiple times...

### A full GraphQL query of (mostly) everything in the poc
Response time: ~315 ms (avg over 5 requests)
<details>
<summary>Query</summary>

```gql
query {
    user(id: 9) {
        id
        username
        account_history {
            period
            reason
            supporting_url
            timestamp
        }
        achievements {
            achievement {
                id
              	description
            }
        }
[...]

PR #9584, opened 2022-12-03
Add API scope "loved" that can only be requested by a designated client

cl8n:
for #9451, splitting this part off.

this is my proposal for authorizing a single known client to have special permissions over the API. similar to the concept of `_lio` endpoints but using existing API setup bc it's easier to implement on both ends & document

the actual requests will come in later PR, but basically this will be used by https://loved.sh to associate Loved forum polls to their respective beatmaps/gamemodes/authors/etc. -- it will be done automatically after opening the polls each round

if this is merged as-is, LOVED_OAUTH_CLIENT_ID should be set to 7 on production.


grumd:
I wonder if osu-pps.com can also get some custom endpoints, because right now I do quite a lot of requests every day to recalculate all the data it needs. If it was an SQL query, that would be way faster I reckon.

peppy:
> I wonder if osu-pps.com can also get some custom endpoints, because right now I do quite a lot of requests every day to recalculate all the data it needs. If it was an SQL query, that would be way faster I reckon.

The only way that is going to happen is if graphql is implemented (see https://github.com/ppy/osu-web/issues/7351).

grumd:
Makes total sense, I'll make sure to follow that issue, thanks

notbakaneko (changes requested):
What's the issue with using the authorization code grant for this?

peppy:
@cl8n you're probably going to have to give an answer to https://github.com/ppy/osu-web/pull/9584#pullrequestreview-1446238643.

Bumping for visibility.

notbakaneko:
Additionally, we don't really want to be adding more special clients if possible, especially if we don't own both ends.

cl8n:
> What's the issue with using the authorization code grant for this?

no issue in the sense that I *can* make it work like that. but in the end it would just be myself authorizing once and then forever refreshing that one token for any automated process performed like this, so I thought client credentials was a more fitting scheme

> Additionally, we don't really want to be adding more special clients if possible, especially if we don't own both ends.

the endpoints I plan to add for this are not useful for, and should not be used by, anything except the one program they're designed to be an interface for. so that is why I chose to write in a specific client ID here

I'm open to other suggestions for how to authenticate for these endpoints, but I don't really have any better ideas myself.

peppy:
I'm not sure on the correct oauth way to structure this, but here's a few pointers:

- I think this is best suited to client credentials (which is already implemented) right? It's a single server component authorising, not users. So I'm confused about the [suggestion to use authorisation grant](https://github.com/ppy/osu-web/pull/9584#pullrequestreview-1446238643)
- I think we *will* want elevated API privileges for some user groups. If we don't want to be adding more oauth scopes, then is the better way forward to add group checks local to API endpoint implementations? ie. all forum poll endpoints would require use of client credentials, and the calling user to be in a specific `phpbb_group`.

nanaya:
Scope should be for limiting what the token can do from the client perspective, not for server side authorisation. Additionally, `*` here stops being actually `*` makes things confusing.

The permission should be checked on client level instead. Similar to user group but for oauth client.

As for client credential acting as a user, I don't think that should be a thing (although it already kind of is with `delegate` scope). There's authorization code for that. Creating the initial access and refresh token is a bit involved but not impossible and only need to be done once as long it keeps refreshed regularly. Or alternatively a way to create a [personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens) which is just authorization code without extra steps.

As for token with access to users' group, a new scope which gives authenticated user full group access would be appropriate.

cl8n:
> Scope should be for limiting what the token can do from the client perspective, not for server side authorisation.

fair enough, but to be clear I got this type of idea from the existing chat scope :thinking: 
I'll leave it as a normal scope then, and the endpoints will be responsible for authorisation.

> As for client credential acting as a user, I don't think that should be a thing

well me neither but point I was trying to make is I don't think this really can be attributed to one user, it's just implementing part of how Loved itself works, with the help of a third-party component. I think the client itself is what's being authorised and not *me*. my osu! account is not particularly relevant to anything that this will be doing.

anyway --- with the conclusion being to not have special behavior for requesting the scope itself, the majority of this PR will not be used anymore so I'm closing it. my first PR in this series will instead be for the main endpoint required, where I hope the context will help decide how the authorisation should be performed

PR #9596, opened 2022-12-06
Add elasticsearch health check and replace `--wait` flag usage in `docker-compose` flow

peppy:
I have a hope of removing the `--wait` parameter from the `queue` command exposed by `osu-elastic-indexer`, but in order to remove it we need to make sure that it can be replaced with docker compose health checks.

This is the first step in ensuring that's feasible. Of note, the parameter is still being used in the github actions flow. A next step would be to change github actions to consume this compose flow instead of doing its own thing.

Local testing shows that startup order does seem to be respected. Would appreciate a second test.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -57,8 +57,10 @@ services:
   notification-server:
     image: pppy/osu-notification-server
     depends_on:
-      - redis
-      - db
+      redis:
+        condition: service_healthy
+      db:
+        condition: service_healthy
     volumes:
       - .env:/app/.env
       - ./storage/oauth-public.key:/app/oauth-public.key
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -110,6 +119,11 @@ services:
       action.auto_create_index: "false"
       discovery.type: single-node
       ES_JAVA_OPTS: "-Xms512m -Xmx512m" # less OOM on default settings.
+    healthcheck:
+      test: curl -s http://localhost:9200 >/dev/null || exit 1
```
notbakaneko, line 123: This should be at least `/_cluster/health?wait_for_status=yellow` as elasticsearch can be http accessible before any of the shards are ready.

notbakaneko (approved):
seems to be working

PR #9598, opened 2022-12-06
Cleanup modal portal element on unmount

notbakaneko:
Side effect of removing the `Modal` element instead of toggling visibility was the portal `div` not being cleaned up


nanaya:
Is it not covered by #9573?

PR #9611, opened 2022-12-08
Add medal unlock components

cl8n:
- [ ] depends on #9610

this is a proposal for an interface to award medals from osu-web

each medal is a queueable event listener that extends `MedalUnlock`. basic flow is
- `$event` type-hints the events it will listen for
- `shouldHandle()` tests whether the received event is actually relevant to the medal unlock -- checks in here should be inexpensive bc it runs synchronously after the event is dispatched. if true, the handler will be queued as a job
- `getApplicableUsers()` gets the users that may unlock the medal, based on the event and any info saved at queue time (`getQueueableState()`)
- `shouldUnlockForUser()` tests whether a user will unlock the medal, if they didn't have it already

medal unlocks won't run without a corresponding entry in `osu_achievements` where `enabled`=1, so it's safe to either add the implementation before the db entry, or add a disabled db entry first and enable it after implementation.

the event-driven design is partly so medal logic is in one place, and partly so hush-hush medals can be easily dropped in from a closed-source repo and Just Work:tm:. kind of over-engineered lol but it should be easy to make lots of different kinds of medals with this. I have some random example medals & tests at https://github.com/cl8n/osu-web/tree/medal-examples and https://github.com/cl8n/osu-web-hush-hush-medals

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,6 @@ composer-installer
 /bootstrap/cache/
 /storage/htmlpurifier/
 /storage/paypal_auth.cache
+
+# Closed-source components
+/hush-hush-medals/
--- a/app/Http/Controllers/InterOp/UsersController.php
+++ b/app/Http/Controllers/InterOp/UsersController.php
@@ -7,37 +7,24 @@ namespace App\Http\Controllers\InterOp;
 
 use App\Exceptions\ValidationException;
 use App\Http\Controllers\Controller;
-use App\Jobs\Notifications\UserAchievementUnlock;
 use App\Libraries\UserRegistration;
-use App\Models\Achievement;
-use App\Models\Event;
+use App\Models\Beatmap;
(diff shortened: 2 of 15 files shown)


peppy:
For the hush-hush portion, is there already premise for this existing? Feel free to PM me example usages, because I can't immediately think of any which wouldn't be score related.

cl8n:
not really, I have some ideas for hush-hush but it's not the immediate reason I was working on this. will PM though

nanaya:
This looks like non trivial to deploy with docker?

Might as well create a separate service where osu-web just forwards all received requests to? It can then send back medal generation request to osu-web through interop interface.

peppy:
I think the modularity should be removed if we just want to get some new web-based medals implemented, since it's only there for hush-hush (which we don't even know if we need/want).

PR #9612, opened 2022-12-08
Use memory for local cache

nanaya:
There's no need for file cache when using octane. This will cause full query on cache reset but shouldn't be too bad? Also removes the need to manually trigger reset as the current cache doesn't expire at all until next deploy.

Also centralised handling of the cache reset counter.


peppy (dismissed):

app/Libraries/LocalCacheManager.php
```diff
@@ -0,0 +1,32 @@
...
+
+        if ($this->resetTicker > $this->resetTickerLimit) {
+            $this->resetTicker = 0;
+            $this->resetCache();
+        }
+    }
+
+    private function resetCache(): void
+    {
+        foreach (['chat-filters', 'groups', 'layout-cache'] as $memoizedSingleton) {
```
cl8n, line 28: would this ever be useful for `groups` or `layout-cache`? their only changes are followed by deploy or manual `resetMemoized()`

nanaya: there have been some updates to groups directly to db. also resetMemoized doesn't reset other instances

notbakaneko, line 28: const?

notbakaneko (changes requested):

notbakaneko (approved):

PR #9613, opened 2022-12-08
Use iterable for array-like typing

nanaya:
So that's a thing 👀

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -19,7 +19,7 @@ function api_version(): int
     return $version;
 }
 
-function array_reject_null(array|ArrayAccess $array): array
+function array_reject_null(iterable $array): array
 {
     $ret = [];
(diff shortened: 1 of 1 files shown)


app/helpers.php
```diff
@@ -1509,7 +1509,7 @@ function get_params($input, $namespace, $keys, $options = [])
 
     $params = [];
 
-    if (is_array($input) || ($input instanceof ArrayAccess)) {
+    if (is_iterable($input)) {
```
notbakaneko, line 1512: `Generator`s are `iterable` but can't be accessed by array syntax used by `array_get` (which uses `is_array` and `ArrayAccess` for the check, anyway)

notbakaneko (changes requested):

notbakaneko (approved):

PR #9616, opened 2022-12-09
Add support for redirect-only product

nanaya:
Please jam the url in description field "for now" 👀 

The header image should probably have some indicator of it being an external store link.


notbakaneko:
I think something has to exclude `/store/products/redirect` in `ProductsController::getProduct`, maybe? 🤔

nanaya:
I think it's fine as long it's not actually linked from anywhere 🤷‍♀️

notbakaneko (approved):
I feel like something should show when redirecting to external sites in the same tab but that's a different problem 👀

PR #9620, opened 2022-12-10
Use consistent order for FA tracks

### LINKED ISSUE #9519: Prioritize osu!originals in Featured Artist track listing sort order
This is a fairly minor cosmetic thing, but I feel like the logical expectation for sorting here is to have the osu!originals be first in the order and clustered together instead of spread out amongst the album/song grouping.

Expected order:
- New originals
- Existing originals
- New normal tracks
- Existing normal tracks

cl8n:
resolves #9519, and now display_order works outside of albums too if that's ever needed.


app/Http/Controllers/ArtistsController.php
```diff
@@ -40,7 +40,10 @@ public function show($id)
             ->where('visible', true)
             ->orderBy('id', 'desc')
             ->with(['tracks' => function ($query) {
-                $query->orderBy('display_order', 'ASC');
+                $query
+                    ->orderBy('display_order', 'asc')
+                    ->orderBy('exclusive', 'desc')
+                    ->orderBy('id', 'desc');
```
nanaya, line 46: this part hasn't been working after the `tracks.artist` below was added. Also this one probably should have id ascending instead.

nanaya (changes requested):

peppy:
Shall we un-stall this and get it merged?

notbakaneko (approved):

cl8n:
mb for missing this one, I'm not subscribed to this PRs notifications for some reason

PR #9623, opened 2022-12-10
Prevent incorrect error on simultaneous first wiki image request

nanaya:
Wait up to 10 seconds on first image load if other worker is already doing it and holding the lock.

### DIFF
--- a/app/Models/Wiki/Image.php
+++ b/app/Models/Wiki/Image.php
@@ -8,6 +8,7 @@ namespace App\Models\Wiki;
 use App\Exceptions\GitHubNotFoundException;
 use App\Libraries\OsuWiki;
 use Exception;
+use Illuminate\Contracts\Cache\LockTimeoutException;
 
 class Image implements WikiObject
 {
@@ -37,7 +38,7 @@ class Image implements WikiObject
     public function __construct($path)
     {
         $this->path = OsuWiki::cleanPath($path);
-        $this->cache = cache()->get($this->cacheKey());
+        $this->cache = $this->fetchFromCache();
     }
 
     public function cacheKey()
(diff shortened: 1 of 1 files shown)


peppy:
Do you think this is the reason for the full failure earlier? Or something you found along the way?

nanaya:
This probably is the reason.

peppy:
I guess it's gotten worse due to now having 30+ workers.

app/Models/Wiki/Image.php
```diff
@@ -67,10 +68,23 @@ public function sync($force = false)
...
 
         if (!$lock->get()) {
-            return $this;
+            if ($this->cache !== null) {
+                return $this;
+            }
+
+            try {
+                $lock->block(10);
+                $lock->release();
```
notbakaneko, line 81: should this be in a `finally`?

notbakaneko (changes requested):

app/Models/Wiki/Image.php
nanaya: I don't think so because in other cases the lock isn't obtained

notbakaneko: 🤔 
(assuming the follow lines don't explode)

notbakaneko (dismissed):

notbakaneko (approved):
apparently the whole fetching back can be done with `->lock(..)->block(...,fn () => ...) instead  🤔  (but I don't want to check everything again right now :eyes:)

app/Models/Wiki/Image.php
nanaya: actually this can't even be in finally as otherwise the lock would still be held when calling sync for return

notbakaneko (approved):

PR #9627, opened 2022-12-12
Update composer lockfile

notbakaneko:
This is to update `laravel/octane` to 1.3 to support Swoole 5.
There're  2 parts, the minimum to have `laravel/octane` 1.3 installed 
- [x] #9650

`composer update` and fix for breaking changes ಠ_ಠ

The jwt validation thing wants a longer key
Now with more overrides to fractal be able to return `null`s in the response ಠ_ಠ


notbakaneko:
set to draft because it turns out there're breaking changes

notbakaneko:
hope nothing else is broken 👀

nanaya (approved):

PR #9628, opened 2022-12-12
Add support for exp rank history

nanaya:
🤷‍♀️


peppy:
Have added migration entry.

notbakaneko (approved):

PR #9629, opened 2022-12-12
Fix rank history display when there's no history

nanaya:
Not sure how frequent this happens (or if still possible to happen at all).


notbakaneko (commented):
not sure if this happens after #9630 since `is_ranked` should now be now `false` those cases 🤔

nanaya:
It's more for when the rank history entry is missing despite having rank in user statistics.

notbakaneko (approved):

PR #9635, opened 2022-12-12
Add support for hiding supporter event on own recent activity

### LINKED ISSUE #9578: Supporter gifting improvements
This has been a recurring request to me, and I feel we should spend a short amount of time to make this work better.

### Ability to hide fact one gifted a tag

When gifting a supporter, the gifter should have a checkbox which decides whether it shows the gift on their user page via the following message:

Safari 2022-12-02 at 04 57 47

### Ability to attach a message when gifting

A textbox should be added which allows adding an optional message which can be sent along with the tag.

---

It would be great if we could get this implemented in the week or so, I think it would give people a nice way to personalise supporter gifts over the giving season.

@notbakaneko @nanaya can one of you take this (or a community contributor like @cl8n otherwise)?

notbakaneko:
- [ ] decide what actually adds events to activity

This is a bit less trivial than it sounds as it's possible to have multiple supporter tags in the same `Order`, so I'm not sure what should be the rule of when it can be hidden or not 🤔 
The current version hides only if everything in the same order is marked to be hidden.

closes #9578

### DIFF
--- a/app/Models/Store/Order.php
+++ b/app/Models/Store/Order.php
@@ -706,6 +706,7 @@ class Order extends Model
     private function extraDataSupporterTag(array $orderItemParams)
     {
         $params = get_params($orderItemParams, 'extra_data', [
+            'hidden:bool',
             'target_id:int',
         ]);
 
--- a/app/Libraries/Fulfillments/SupporterTagFulfillment.php
+++ b/app/Libraries/Fulfillments/SupporterTagFulfillment.php
@@ -85,10 +85,12 @@ class SupporterTagFulfillment extends OrderFulfiller
 
         $isGift = count($gifts) !== 0;
 
-        Event::generate(
-            $this->continued ? 'userSupportAgain' : 'userSupportFirst',
-            ['user' => $donor, 'date' => $this->order->paid_at]
-        );
+        if ($items->contains(fn (OrderItem $item) => !$item->extra_data->hidden)) {
+            Event::generate(
+                $this->continued ? 'userSupportAgain' : 'userSupportFirst',
+                ['user' => $donor, 'date' => $this->order->paid_at]
+            );
+        }
 
         if (present($donor->user_email)) {
             Mail::to($donor)
--- a/app/Models/Store/ExtraDataSupporterTag.php
+++ b/app/Models/Store/ExtraDataSupporterTag.php
@@ -14,12 +14,14 @@ class ExtraDataSupporterTag extends ExtraDataBase implements JsonSerializable
     const TYPE = 'supporter-tag';
 
     public int $duration;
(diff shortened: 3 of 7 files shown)


cl8n:
I think it should hide if *any* are marked to be hidden, otherwise the option is kind of lying and user won't get the expected result. also, I think this can be stated outright in the option for clarity -- "hide all gifts in this order from my activity" or similar

peppy:
Could work better as a final checkout option, agree.

notbakaneko:
I put it on the very last step before picking a payment method, although I wonder if it should be on the previous page instead 🤔

app/Models/Store/Order.php
```diff
@@ -368,6 +369,11 @@ public function canUserCancel(): bool
         return $this->status === static::STATUS_PAYMENT_REQUESTED;
     }
 
+    public function containsSupporterTag(): bool
+    {
+        return $this->items->first(fn (OrderItem $item) => $item->product->custom_class === Product::SUPPORTER_TAG_NAME) !== null;
```
nanaya, line 374: more like `->contains(...);`

resources/views/store/checkout/show.blade.php
```diff
@@ -34,6 +34,18 @@
...
+                    @include('objects._switch', ['locals' => [
+                        'additionalClass' => 'js-store-checkout-option',
+                        'checked' => $order->isGiftsHidden(),
+                        'name' => 'hide_from_activity',
+                    ]])
+                    {{ osu_trans('store.checkout.hide_from_activity') }}
+                </label>
+            @endif
+
+
```
nanaya, line 48: extra line

resources/views/store/checkout/show.blade.php
```diff
@@ -34,6 +34,18 @@
 
             @include("store.objects.order", ['order' => $order, 'modifiers' => ['checkout']])
 
+            @if ($order->containsSupporterTag())
```
nanaya, line 37: I think the final state should be shown in the invoice

resources/assets/lib/core-legacy/store-checkout.coffee
```diff
@@ -33,7 +33,8 @@ export class StoreCheckout
       showLoadingOverlay.flush()
 
       init[provider]?.then ->
-        $.post(route('store.checkout.store'), { provider, orderId })
+        hide_from_activity = document.querySelector('.js-store-checkout-option')?.checked
```
nanaya, line 36: the class name ("store checkout option") and what it does doesn't match ("hide from activity")

nanaya (changes requested):

nanaya (approved):

PR #9636, opened 2022-12-13
Create groups hidden by default in `Groups::byIdentifier()` fallback

cl8n:
to prevent https://osu.ppy.sh/groups/47 from happening again (and someone should fix that; either hide it or give it a real name/description)

also blank description by default instead of identifier name.


notbakaneko:
There also used to be a thing to stop groups from being auto-created on production because groups on production should exist first https://github.com/ppy/osu-web/pull/5617

app/Libraries/Groups.php
```diff
@@ -70,9 +70,10 @@ public function byIdentifier(string $id): Group
         if ($group === null) {
             try {
                 $group = Group::create([
-                    'identifier' => $id,
+                    'group_desc' => '',
                     'group_name' => $id,
-                    'group_desc' => $id,
+                    'group_type' => 0,
```
nanaya, line 75: should be 2 instead ([`GROUP_HIDDEN`](https://github.com/phpbb/phpbb/blob/eb1edd12a1bc6adbb496d043fd2c883f366f452c/phpBB/includes/constants.php#L78-L83))

nanaya (changes requested):

nanaya (approved):

PR #9638, opened 2022-12-13
New download page

### LINKED ISSUE #8246: Update design of download page
The current download page feels too weeb, dated, and doesn't have anywhere to put the lazer download link.

It should be redesigned to match closer to the modern/flat/technical look we have elsewhere. As it stands a lot of people probably get there and close their browser window immediately.

nanaya:
[Alternative version which keeps the super header style](https://s.myconan.net/2022-12/Screenshot%202022-12-13%20at%2022-17-02%20download%20osu%21.png)

[Same with screenshot below but with default image for background](https://s.myconan.tk/2022-12/Screenshot%202022-12-21%20at%2017-40-09%20download%20osu%21.png)

That's a thing. I don't have the download page icon so it's using the home icon.

10 minutes search doesn't indicate that it's possible to detect apple silicon correctly so it's just macos intel for now. Also there are some reports iPadOS pretending to be macos (desktop) by default so that may get presented wrong version as well.

@arflyte check the design maybe? 🤷 

Resolves #8246?


Phippe:
Not Flyte, but wouldn't it look better if the alternative platforms below the download buttons were centred or left-aligned?

peppy:
agree, centre aligned and separated by `・`　or `/` would work better.

Tanza3D:
also not flyte, but i feel it could possibly look nicer if aligned like this? 
image

peppy:
Definitely looks better

nanaya:
I think the download buttons should at least be aligned the same (but then the question is to what)

Tanza3D:
image
image

maybe like this?

nanaya:
how's it aligned horizontally? centered to the column or ~70px to the center separator line?

also the extra links are missing

and I just noticed is the button just plain color without background and shadow?

Tanza3D:
i just threw together a basic concept for positioning's sake, not visual sake - i'd centered to the column would make the most sense, as it is right now, and for the extra links they should probably be centered as @Phippe said

here's a bit more of a visual one
image

nanaya:
Updated as per suggestion. I haven't put any background in the actual code since there's no actual image for it yet.

Tanza3D:
is the design in the screenshot up to date? if so, just a suggestion but could you move the text up a bit on the buttons? they look vertically uncentered to me right now. a small thing but it bugs me every time i see it

app/Http/Controllers/HomeController.php
```diff
@@ -55,7 +56,40 @@ public function downloadQuotaCheck()
...
+            $platform = 'ios';
+        } elseif ($agent->is('OS X')) {
+            // FIXME: figure out a way to differentiate Intel and Apple Silicon
+            $platform = 'macos_as';
+        } elseif ($agent->is('Linux')) {
+            $platform = 'linux_x64';
+        } elseif ($agent->isAndroidOS()) {
+            $platform = 'android';
+        }
+        $platform ??= 'windows_x64';
```
notbakaneko, line 87: that's an interesting way of writing `else { $platform = 'windows_x64' }` (or use `match` since every case is 1 line 🤔 )

notbakaneko (changes requested):

app/Http/Controllers/HomeController.php
nanaya: it was originally a bunch of nested ifs but then it's not anymore

notbakaneko:
I'm pretty sure there's no way to identify apple silicon from http headers alone, with the whole everything identifying as "Mozilla/5.0 Intel" 👀 
If I'm not mistaken iPadOS always identifies as a `Macintosh; Intel Mac OS X` by default now unless explicitly requested or disabled.

Basically, User-Agent is a lie :trollface:

notbakaneko (approved):
seems fine other than missing image 👀

Issue #9316, opened 2022-09-28
Self-Mutual/Friend via Network + Console


WindowsMeosu:
I've felt like this is worth reporting as an issue, so apparently you can self-mutual yourself with your web browser's console (which is found with Inspect Element). To do this, one must first open Inspect Element (I assume you know how) and friend someone. Open "network" 
![image](https://user-images.githubusercontent.com/104236864/192891032-87add1b1-0334-4900-9708-543f28c1cb96.png)
and after friending someone, you'll see this;
![image](https://user-images.githubusercontent.com/104236864/192891607-89d2b5bb-f539-440c-9fa4-2241db9bf636.png)
right-click "friends?target=2" and hover over "Copy"
and these options will show up:
![image](https://user-images.githubusercontent.com/104236864/192891853-507716d7-0ce1-4c5f-a407-9ec8f183f02f.png)
Click on "copy as fetch" and move to Console 
![image](https://user-images.githubusercontent.com/104236864/192892204-ade919aa-9bf9-446c-8fad-3e65f2c94fd9.png)
PASTE this but DO NOT enter yet: "fetch("https://osu.ppy.sh/home/friends?target=2", {
  "headers": {
    "accept": "*/*",
    "accept-language": "en-US,en;q=0.9",
    "sec-ch-ua": "\"Microsoft Edge\";v=\"105\", \"Not)A;Brand\";v=\"8\", \"Chromium\";v=\"105\"",
    "sec-ch-ua-mobile": "?0",
    "sec-ch-ua-platform": "\"Windows\"",
    "sec-fetch-dest": "empty",
    "sec-fetch-mode": "cors",
    "sec-fetch-site": "same-origin",
    "x-csrf-token": "eROpGIwWkPOKYqJmxjIerR6dnAsmcp5lNvZf5uqc",
    "x-requested-with": "XMLHttpRequest"
  },
  "referrer": "https://osu.ppy.sh/users/2",
  "referrerPolicy": "strict-origin-when-cross-origin",
  "body": null,
  "method": "POST",
  "mode": "cors",
  "credentials": "include"
});"

replace "target=2" and /users/2 with your ID
```
fetch("https://osu.ppy.sh/home/friends?target=YOUR-ID", {
  "headers": {
    "accept": "*/*",
    "accept-language": "en-US,en;q=0.9",
[...]

PR #9639, opened 2022-12-13
prevent self-subscribing

WindowsMeosu:
- [ ] fix debugging errors 

image
although #9316 states that you can both mutual and subscribe to yourself; this was never resolved for mapping subscribers. I have copied the resulting commit from #9319 and pasted it into `app/Https/FollowersController.php`.

drafted because debugging purposes

debug failed, drafted until I have time to fix debugging problems


WindowsMeosu:
closing because debugging keeps resulting in major failure, can reopen if anyone wants me to retry

PR #9642, opened 2022-12-14
Simplify proxy trusting logic

nanaya:
There's no need to reset trusted proxy settings on each request anymore.


peppy (dismissed):

notbakaneko (changes requested):
`fideloper/proxy` not used anymore?

notbakaneko (approved):

PR #9645, opened 2022-12-16
Reduce count queries for notification bundle

nanaya:
This surprisingly works without need to change database at all? 🤔 The query is a bit questionable though. Both the way it being generated and the query itself.

- [x] depends on #9634 just because modifying same file


app/Libraries/NotificationsBundle.php
```diff
@@ -91,12 +92,71 @@ private function fillStacks(string $objectType, int $objectId, string $category)
...
+                {$groupColumns['id']},
+                {$groupColumns['name']}")
+            ->whereRaw($whereString, $bindValues)
+            ->groupBy(array_values($groupColumns));
+
+        if ($this->unreadOnly) {
+            $query->where('is_read', false);
+        }
+
+        foreach ($query->get() as $count) {
```
notbakaneko, line 151: `$row`

`$count->count` seems...not right

notbakaneko (changes requested):

notbakaneko (approved):

PR #9646, opened 2022-12-16
Add support for including messages in supporter tag gifts

notbakaneko:
Allows attaching a message to gifted supporter tags

The transition is kind of janky at the moment with the whole visibility + height shifting

to self (no message input):
image

gift to others (input visible):
image

- [x] need better transition for textbox
- [x] include note on cart/invoice for tags with messages


nanaya:
Maybe either make the box always visible (but disabled and have different placeholder when for self) or make it opacity 0 as needed instead?

notbakaneko:
The issue would be there's now a big blank space 🤷

nanaya:
better than moving content... I think.

nanaya (changes requested):

nanaya (approved):

PR #9648, opened 2022-12-16
Add seasons infrastructure and basic display

venix12:
(a big) part of https://github.com/ppy/osu-web/issues/8736 :sparkles:

image


venix12:
am i doing something wrong or is lint just having a stroke 🤨

notbakaneko:
It means the new code added more warnings; we set a max limit on warnings for existing code, new code should not be adding lint errors or new warnings.

If running `yarn lint` returns the same number of warnings, you may need to clear the lint cache.

venix12:
Well, no matter what I tried, I got 158 on both this branch and master, tried clearing cache too. 🤷

nanaya:
Is it intentional allowing room to be potentially part of multiple seasons?

Also conflicts.

venix12:
> Is it intentional allowing a room to be potentially part of multiple seasons?

It is, see https://github.com/ppy/osu-web/issues/8736#issuecomment-1155514126.

nanaya:
I don't think the comment has any conclusion on that, just wondering it could be a possibility

nanaya:
that looks more like on which way to do one to many (as per the question), not whether it should be one or many to many

venix12:
Right... guess we could go with one-to-many and extend to many-to-many if/when needed then?

venix12:
remember to never use git rebase irresponsibly! 

replaced by #9725 as it's probably way easier than fixing this one =D

PR #9649, opened 2022-12-17
Fix store line item closing tag

nanaya:
Randomly found this weird thing 🤨

It's immediately followed by another closing tag so browser automatically fixes it (and there's no matching open tag for what it's closing).


notbakaneko (approved):
🤔

PR #9652, opened 2022-12-19
Update docs on docker compose

nanaya:
Current compose file requires newer docker compose.


notbakaneko (changes requested):
should the instances in `DEPLOYMENT.md`, `docker_dev.sh` and `prepare.sh` be updated too? 👀

notbakaneko (approved):

PR #9653, opened 2022-12-19
Support downloading solo score replays

nanaya:
Seems working on local storage. No idea about s3 👀 although technically should just work...?


notbakaneko (commented):
Do solo scores need the same metadata embedded into the download like legacy scores or does the meta data come from elsewhere?

nanaya:
it's stored as full replay file

notbakaneko (approved):
seems to work on my local totally an s3 👀

PR #9661, opened 2022-12-20
Update Twitter colour and fix header item hover animation

nanaya:
...that said, I don't think it looks good. Maybe that's why it's like that in the first place? The colour has also been updated so I adjusted accordingly.

I also removed icon colour transition for the supporter because the background change is instant.


notbakaneko:
I'm pretty sure the clashing colours was why all the icons there never had highlight colours with only a special exception being made for the support icon.

nanaya:
yeah =/ reverted the hover colour but kept the transition and colour update.

notbakaneko (approved):

PR #9664, opened 2022-12-21
Update docker compose to use octane

notbakaneko:
I've been using `octane --watch` for dev via a custom `docker-compose.yml` for a while now and it seems to work fine.
The `restart_policy` is needed to auto restart octane if it crashes from a yarn build replacing the manifest file.

(I think documentation for running tests might need updating but that's an unrelated problem 👀 )

- [x] dusk


public/index.php
```diff
@@ -1,55 +0,0 @@
-<?php
```
nanaya, line 1: I guess this is still used by dusk. Maybe it can be updated to use octane as well. And maybe do something about `artisan serve`

nanaya (changes requested):

nanaya (approved):

PR #9667, opened 2022-12-22
Fix check for pluralised translation

### LINKED ISSUE #9666: The "forum posts" and "comments" counter on profile page always default to English regardless of chosen language
Noticed this happening after the recent Crowdin update (#9662). The "forum posts" and "comments" counter has long been localized as far as I can remember, but since the update they now default to English no matter what language you're currently in.

Here's the bug appearing in various languages :

image
image
image

nanaya:
🤨

Resolves #9666


notbakaneko (approved):
👀

PR #9671, opened 2022-12-23
Adjust beatmap difficulty item display

nanaya:
Combine mapper name display logic which supports guest mapper. The popup styling matches the new beatmap page with some adjustments - mainly of always displaying scroll bar because otherwise mapper name may get cut off when it shows instead.


resources/assets/lib/components/beatmap-list-item.tsx
```diff
@@ -2,17 +2,31 @@
...
+import StringWithComponent from './string-with-component';
 
-interface Props {
+interface BaseProps {
   beatmap: BeatmapJson | BeatmapExtendedJson;
+  beatmapUrl?: string;
   inline?: boolean;
   modifiers?: Modifiers;
 }
+type MapperProps = {
```
notbakaneko, line 21:
```suggestion

type MapperProps = {
```

notbakaneko (changes requested):

notbakaneko (approved):

PR #9672, opened 2022-12-23
Disable elasticsearch geoip downloader for docker compose

notbakaneko:
Not used, so no point having elasticsearch try to update it while starting or spitting out not enough shards errors


nanaya (commented):
should the same be applied to github action as well? 🤔

nanaya (approved):

PR #9673, opened 2022-12-23
🐛 Fix missing jpegtran binary

ThePooN:
Noticed the following error in production logs:  
image

Curious that this never reached Sentry.

No idea what's the impact -- maybe our produced jpegs are heavier now? We surely would have heard about it from support if it directly impacted users. Also no idea if that was installed on our previous droplets.

Both `libjpeg-progs` and `libjpeg-turbo-progs` provide the `jpegtran` binary. I couldn't figure which implementation is best through this codebase and Laravel docs, but overall `libjpeg-turbo-progs` seems best.

### DIFF
--- a/Dockerfile.deployment
+++ b/Dockerfile.deployment
@@ -26,7 +26,8 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get install -y \
     php8.0-xml \
     php8.0-zip \
     php8.0 \
-    zip
+    zip \
+    libjpeg-turbo-progs
 
 WORKDIR /app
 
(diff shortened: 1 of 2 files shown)


nanaya:
that doesn't seem right. It should already be part of `jhead` dependency.

nanaya (changes requested):
need a comment saying it's a missing dependency of jhead and shouldn't be needed in future ubuntu releases*

*although probably we'd have changed to something else by then as ubuntu also fucked up the packaging of chromium which is needed for testing in dev

nanaya (approved):

PR #9697, opened 2022-12-30
Throttle registration by ip network

nanaya:
/24 for ipv4 and /56 for ipv6 addresses. And here's hoping the cgnats don't have ip addresses too close together 🤷 (and that they don't host too many users with single ip)

Ipv6 /56 is _usually_ the largest prefix internet providers give out to users.

@peppy please check first if it's meaningful to do this check as it's probably not all that difficult to get new ip beyond the specified prefix.

### DIFF
--- a/config/osu.php
+++ b/config/osu.php
@@ -227,6 +227,7 @@ return [
     'user' => [
         'allow_email_login' => get_bool(env('USER_ALLOW_EMAIL_LOGIN')) ?? true,
         'allow_registration' => get_bool(env('ALLOW_REGISTRATION')) ?? true,
+        'registration_mode' => presence(env('REGISTRATION_MODE')) ?? 'client',
         'allowed_rename_groups' => explode(' ', env('USER_ALLOWED_RENAME_GROUPS', 'default')),
         'bypass_verification' => get_bool(env('USER_BYPASS_VERIFICATION')) ?? false,
         'hide_pinned_solo_scores' => get_bool(env('USER_HIDE_PINNED_SOLO_SCORES')) ?? true,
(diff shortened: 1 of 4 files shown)


ThePooN:
> @peppy please check first if it's meaningful to do this check as it's probably not all that difficult to get new ip beyond the specified prefix.

I suggest blocking (or rather, redirecting to the web-based registration flow with captcha) based on ASN instead.

I used a dump of 2985 accounts that registered within 24 minutes during one of the account registrations spam wave (filtered to only contain spam accounts), and determined we have 289 /24 subnets vs. 30 ASNs. Using ASNs, we could use much more sensible thresholds to avoid impacting legitimate users.

I have attached detailed (anonymized) results with registration counts per subnets and per ASN for comparison ([source code](https://gist.github.com/ThePooN/5ea95d5d92536b486ac3493811ad8fac)) using the GeoLite2 database.

[asn.csv](https://github.com/ppy/osu-web/files/10340784/asn.csv)
[subnets.csv](https://github.com/ppy/osu-web/files/10340786/subnets.csv)

config/osu.php
```diff
@@ -227,6 +227,7 @@
     'user' => [
         'allow_email_login' => get_bool(env('USER_ALLOW_EMAIL_LOGIN')) ?? true,
         'allow_registration' => get_bool(env('ALLOW_REGISTRATION')) ?? true,
+        'registration_mode' => presence(env('REGISTRATION_MODE')) ?? 'client',
```
notbakaneko, line 230: duplicate

notbakaneko (changes requested):

nanaya:
apparently we're going with asn way.

config/osu.php
nanaya: so I incorrectly amended commit adding the new registration page on this branch 👀

PR #9698, opened 2022-12-30
Add web interface for user registration

nanaya:
ಠ_ಠ

Also had to add support for rendering multiple captcha thing on a page.

```
# "client" for client registration
REGISTRATION_MODE=web
```


peppy:
> Dunno if/where to put the ToS agreement thing. One way is to put a "By creating account you agree to the [terms of service](https://github.com/ppy/osu-web/pull/9698#)." underneath the "create account" button.

I think this would more than suffice.

peppy:
I'm going to merge this one as to have it handy for future use.

Client side updates are pending, will add support for that over the coming days. Makes it easier if I have this deployed.

PR #9705, opened 2023-01-03
Only fetch scores with valid beatmap

### LINKED ISSUE #9686: Bugged Profile Page (lazer.ppy.sh)
I'm not totally sure how to debug this, but here's the broken page: https://lazer.ppy.sh/users/14613516 (example error code `098bf7af363d4ac29bc6b9e116966961`) and here's some potentially related information:

- About 10 hours ago, there was no error, but my "Best performance" section listed 0 scores and scores set on lazer did not add to total PP on lazer.ppy.sh. See my report here: https://github.com/ppy/osu/discussions/21830
- Subsequently, I tried opening stable up and setting one score. That did not change anything, although it did give me rank on osu.ppy.sh again, because mine was reset.
- I primarily play on lazer and have for nearly two years. I have under 10 hours in stable. I use the AppImage build for linux.

nanaya:
Resolves #9686.


notbakaneko (approved):
I feel like there should be an option to skip the extra query when it's fetching for a known valid beatmap 🤷

PR #9709, opened 2023-01-03
Pass elasticsearch sort cursor value as is

nanaya:
Beatmap search has been updated to use cursor string a while ago so there shouldn't be any more problem passing the value as is. And the client shouldn't have problem dealing with it in the first pl--

...or does it check for string type in the cursor json 🤔 

Although ideally everything should just be updated to cursor string...

- [x] confirm client doesn't break when getting int cursor json instead of string


nanaya:
Seems fine on client, at least for beatmap listing.

peppy (approved):

notbakaneko (approved):

PR #9711, opened 2023-01-04
Limit registration by ASN

nanaya:
Alternative to #9697. Can use input for updated throttle rule.

To be added <del>separately</del> if this makes sense:
- [x] composer command to update the database


ThePooN:
Hmm, yeah I also have a few concerns about storing that database in the repo.

I would make geoip optional so we can run without the database in development (and remove it from git, ideally amending your existing commit to avoid bloating the repo).  
For deployments, I would add an env var pointing to the download location (eg. `https://download.maxmind.com/app/geoip_download?edition_id=GeoLite2-ASN&license_key=YOUR_LICENSE_KEY&suffix=tar.gz` or a local cache per MaxMind recommendations) for download at startup. Ideally optional too.

Don't think we really need a composer command for that, just bash scripts?

notbakaneko (changes requested):
Given that the dbs are no longer downloadable without registering (https://blog.maxmind.com/2019/12/significant-changes-to-accessing-and-using-geolite2-databases/) we should probably look at the terms for including and using the database.

This looks like something that should download on `composer install`, but that's also adding another point of failure on deploys which I'm also not keen on 🤔

nanaya:
Fine, it's now using a free asn db provided by some dns person ಠ_ಠ

Initial lookup time increased by 1-2 second because the database is in text format and must be parsed. It could probably be sped up by caching it but 🤷‍♂️

notbakaneko (commented):
The main issue I see is it takes ~~350mb~~(now) ~35mb on boot for every worker process 👀

notbakaneko:
Is it worth some basic sanity tests for `lookup`? 🤔

notbakaneko (approved):

nanaya:
So apparently just recording the row positions on the file could also work 👀 not sure about the performance though

app/Libraries/Ip2AsnUpdater.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        return database_path('ip2asn.tsv');
+    }
+
+    public static function getIndexPath()
+    {
+        return database_path('ip2asn.idx');
+    }
+
+    private string $tsv;
```
notbakaneko, line 22: unused

app/Libraries/Ip2AsnUpdater.php
```diff
@@ -0,0 +1,56 @@
...
+            : file_get_contents($dbPath);
+
+        $currentLine = 0;
+        $index = pack('l', $currentLine);
+        while (($currentLine = strpos($tsv, "\n", $currentLine)) !== false) {
+            $currentLine++;
+            if (isset($tsv[$currentLine])) {
+                $index .= pack('l', $currentLine);
+            }
+        }
```
notbakaneko, line 50: can do with newline maybe 🤔

notbakaneko (changes requested):

notbakaneko (approved):

PR #9725, opened 2023-01-10
Add seasons infrastructure and basic display

venix12:
replaces #9648

(a big) part of https://github.com/ppy/osu-web/issues/8736 :sparkles:

image


app/Models/Season.php
```diff
@@ -0,0 +1,50 @@
...
+    ];
+
+    public function endDate(): ?Carbon
+    {
+        return $this->hasRooms() && $this->concluded
+            ? $this->rooms->sortByDesc('ends_at')->first()->ends_at
+            : null;
+    }
+
+    public function hasRooms(): bool
```
nanaya, line 34: this causes one query on every call

app/Models/Season.php
```diff
@@ -0,0 +1,50 @@
...
+
+    protected $casts = [
+        'concluded' => 'boolean',
+    ];
+
+    public function endDate(): ?Carbon
+    {
+        return $this->hasRooms() && $this->concluded
+            ? $this->rooms->sortByDesc('ends_at')->first()->ends_at
+            : null;
```
nanaya, line 31:
```suggestion
        return $this->concluded
            ? $this->rooms->max('ends_at')
            : null;
```

app/Models/Season.php
```diff
@@ -0,0 +1,50 @@
...
+    public function hasRooms(): bool
+    {
+        return $this->rooms()->count() > 0;
+    }
+
+    public function startDate(): ?Carbon
+    {
+        return $this->hasRooms()
+            ? $this->rooms->sortBy('starts_at')->first()->starts_at
+            : null;
```
nanaya, line 43:
```suggestion
        return $this->rooms->min('starts_at');
```

database/migrations/2022_12_16_151601_create_season_rooms_table.php
```diff
@@ -0,0 +1,37 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('season_rooms', function (Blueprint $table) {
+            $table->id();
+            $table->integer('room_id')->unsigned();
+            $table->integer('season_id')->unsigned();
```
nanaya, line 24: need unique index

app/Http/Controllers/SeasonsController.php
```diff
@@ -0,0 +1,69 @@
...
+                abort(404);
+            }
+        } else {
+            $season = Season::findOrFail($id);
+        }
+
+        $seasons = Season::orderByDesc('id')->get();
+
+        $params = get_params(request()->all(), null, [
+            'cursor:array',
```
nanaya, line 57: use `cursor_from_params`

app/Http/Controllers/SeasonsController.php
```diff
@@ -0,0 +1,69 @@
...
+        [$rooms, $hasMore] = $search['query']->with([
+            'playlist.beatmap',
+            'host',
+        ])->getWithHasMore();
+
+        $rooms->each->findAndSetCurrentPlaylistItem();
+        $rooms->loadMissing('currentPlaylistItem.beatmap.beatmapset');
+
+        return [
+            'cursor' => $hasMore ? $search['cursorHelper']->next($rooms) : null,
```
nanaya, line 35: use `cursor_for_response`

nanaya (changes requested):
also needs to fix merge conflicts

(and I haven't checked the frontend part much)

app/Http/Controllers/SeasonsController.php
```diff
@@ -0,0 +1,71 @@
...
+        ])->getWithHasMore();
+
+        $rooms->each->findAndSetCurrentPlaylistItem();
+        $rooms->loadMissing('currentPlaylistItem.beatmap.beatmapset');
+
+        $nextCursor = $hasMore ? $search['cursorHelper']->next($rooms) : null;
+
+        return array_merge([
+            'rooms' => json_collection($rooms, new RoomTransformer(), ['current_playlist_item.beatmap.beatmapset', 'difficulty_range', 'host', 'playlist_item_stats']),
+            'search' => $search['search'],
```
nanaya, line 38: I don't think this ever end up useful so might as well remove it

nanaya (changes requested):

app/Models/Season.php
```diff
@@ -0,0 +1,43 @@
...
+    use HasFactory;
+
+    protected $casts = [
+        'concluded' => 'boolean',
+    ];
+
+    public function endDate(): ?Carbon
+    {
+        return $this->concluded
+            ? $this->rooms->max('ends_at')
```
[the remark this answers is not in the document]
venix12: basically, spotlights seasons contain of few playlists created one after another so it would end up on on having end date changed with each playlist added. the concluded attribute would be set to true after the season conclude, so after last lobby is added.

resources/js/interfaces/multiplayer-list-json.ts
```diff
@@ -15,7 +15,7 @@ type RoomJsonIncludes =
 
 export type EndpointRoomJson = RoomJson & Required<Pick<RoomJson, RoomJsonIncludes>>;
 
-export default interface UserMultiplayerHistoryJson {
+export default interface MultiplayerListJson {
```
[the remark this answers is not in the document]
venix12: what would you propose for type_group thing then? making it an optional part of response and interface/store?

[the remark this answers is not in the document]
nanaya: separate json just on the user multiplayer page

app/Models/Multiplayer/Room.php
```diff
@@ -103,18 +107,58 @@ public static function preloadRecentParticipants(Collection $rooms)
...
+    {
+        $params = get_params($rawParams, null, [
+            'cursor:array',
+            'limit:int',
+            'mode',
+            'season_id:int',
+            'sort',
+            'type_group',
+            'user:any',
+        ], ['null_missing' => true]);
```
nanaya, line 120: not needed here?

app/Models/Multiplayer/Room.php
```diff
@@ -103,18 +107,58 @@ public static function preloadRecentParticipants(Collection $rooms)
...
+        $rooms->loadMissing('currentPlaylistItem.beatmap.beatmapset');
+
+        $response = [
+            'rooms' => json_collection($rooms, new RoomTransformer(), ['current_playlist_item.beatmap.beatmapset', 'difficulty_range', 'host', 'playlist_item_stats']),
+            'type_group' => $params['type_group'],
+        ];
+
+        if ($includeSearch) {
+            $response['search'] = $search['search'];
+        }
```
nanaya, line 141: I don't think the multiplayer page use this either

resources/views/seasons/show.blade.php
```diff
@@ -0,0 +1,39 @@
...
+    <script id="json-currentSeason" type="application/json">
+        {!! json_encode(json_item($season, 'Season')) !!}
+    </script>
+
+    <script id="json-rooms" type="application/json">
+        {!! json_encode($roomsJson) !!}
+    </script>
+
+    <script id="json-seasons" type="application/json">
+        {!! json_encode(json_collection($seasons, 'Season')) !!}
```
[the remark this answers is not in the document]
nanaya: I don't think this has been fixed. If anything the dropdown doesn't even need any of those data being queried.

database/migrations/2022_12_16_151601_create_season_rooms_table.php
```diff
@@ -0,0 +1,41 @@
...
+
```
nanaya, line 25: random extra line

resources/lang/en/rankings.php
```diff
@@ -23,6 +23,12 @@
         'multiplayer' => 'multiplayer',
         'performance' => 'performance',
         'score' => 'score',
+        'seasons' => 'seasons',
+    ],
+
+    'seasons' => [
+        'ongoing' => 'This season is still ongoing.',
```
nanaya, line 30: I wonder if the "ongoing" need further explanation (as in "there will be more playlists added") or it's already obvious to players. Because the other meaning of ongoing is there are still open playlists (similar to concluded, and also how I initially thought what they mean).

nanaya (changes requested):

resources/lang/en/rankings.php
venix12: The intent was to set concluded=1 after last playlist ends so pretty much "ongoing" would mean that.

nanaya (changes requested):

app/Transformers/SeasonTransformer.php
```diff
@@ -0,0 +1,37 @@
...
+        'end_date',
+        'start_date',
+    ];
+
+    public function transform(Season $season): array
+    {
+        return [
+            'id' => $season->getKey(),
+            'name' => $season->name,
+            'room_count' => $season->rooms()->count(),
```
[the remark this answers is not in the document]
venix12: so a separate transformer for room/season name and id only would work?

app/Http/Controllers/SeasonsController.php
```diff
@@ -0,0 +1,48 @@
...
+
+use App\Models\Multiplayer\Room;
+use App\Models\Season;
+use App\Transformers\SeasonTransformer;
+
+class SeasonsController extends Controller
+{
+    public function rooms($id)
+    {
+        $season = Season::latestOrId($id);
```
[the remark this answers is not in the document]
venix12: thought it would be reasonable for potential future api exposure, but can probably reduce it to just pass $id to params for now i guess :thinking:

resources/lang/en/rankings.php
venix12: I mean, the season concludes when all added playlists are finished, so pretty much "ongoing" means the last added playlist is still open, and that possibly more playlists will be added (not a case for the very last playlist of the season). 

The chain is as it goes (with each step being exectued when the following playlist lobby closes):
Playlist A -> Playlist B -> Playlist C -> set concluded=1

Could change the label for something like "This season is still ongoing. (more playlists might be added)" or so?

venix12:
Needs #9836 merged for tests to pass.

resources/lang/en/rankings.php
nanaya: > season concludes when all added playlists are finished

yes I got that already after you explained it. But to me usually something is concluded when it's over, not when "the final round is open".

Similarly for ongoing, if a season is not ongoing anymore (the tooltip is gone), doesn't that mean it's over? (although in this case the date would be in the future so it's then clear that it's still... ongoing? open?)

app/Transformers/SeasonTransformer.php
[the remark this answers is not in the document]
nanaya: yeah

nanaya:
the fail is just syntax error fwiw...

resources/lang/en/rankings.php
venix12: yes, the concluded is meant to be set to true when the season is over is what i mean.

venix12:
I'm blind. 🤨

resources/lang/en/rankings.php
nanaya: oh wait, I guess it's me who read it wrong 😵‍💫 I thought it'd be set to concluded when the final playlist is added
so there's no actual indication whether or not more playlists will be added? is it just going to be announced somewhere else or is it intentionally not public? and now I think about it, should the page contain some kind of description?

venix12: The playlists are announced in spotlights newsposts and wiki article (e.g. https://osu.ppy.sh/home/news/2022-12-06-beatmap-spotlights-season-8-winter-2023-precision#season-schedule-and-leaderboards and https://osu.ppy.sh/wiki/en/Beatmap_Spotlights/Seasons/2023_Winter#schedule).

Hmm, guess some description could be used however I'm not sure if it's going to be used for spotlights only in the end.

nanaya: wouldn't it be more useful to have flag on whether or not more playlists will be added as I initially thought? that way the end date is visible right on the page when the final playlist is up without having to check the relevant page (and now I wrote it, at least an info link maybe?)

venix12: i mean the room card already has this info so not really sure if that's needed, but we could go this route i guess (requires only column name / tooltip text cahnge i think 🤔)

nanaya: but it's not necessarily the end of the season (unless the room title says so)? also when the actual end is already known why not just show in the header

venix12: yeah, changed concluded to finalised and added tooltip notice as suggested before

nanaya (changes requested):

nanaya:
oh and being able to link to some url would probably be useful

app/Models/Multiplayer/Room.php
```diff
@@ -103,18 +107,48 @@ public static function preloadRecentParticipants(Collection $rooms)
...
+            'max_limit:int',
             'mode',
+            'season_id:int',
             'sort',
             'type_group',
             'user:any',
         ], ['null_missing' => true]);
 
+        $maxLimit ??= 250;
+        $maxLimit = clamp($maxLimit, 1, 250);
```
nanaya, line 149: not needed (it should only be set internally)

venix12: won't the issue mentioned in https://github.com/ppy/osu-web/pull/9725#discussion_r1095453624 stay then?

nanaya: only if it comes from external variable (rooms index endpoint passes all params by default)

venix12: oh right, didn't notice that

venix12:
Tried to do some barebones url implementation, guess that could work for now?

nanaya:
yeah probably fine for now

app/Models/Multiplayer/Room.php
```diff
@@ -103,18 +107,47 @@ public static function preloadRecentParticipants(Collection $rooms)
...
+
+        return array_merge($response, cursor_for_response($nextCursor));
+    }
+
+    public static function search(array $rawParams, ?int $maxLimit = null)
     {
         $params = get_params($rawParams, null, [
             'category',
             'limit:int',
+            'max_limit:int',
```
nanaya, line 140: uh, this should be removed as well

nanaya (changes requested):
uh, I forgot about list of internal urls in `resources/js/utils/url.ts`. `/seasons` needs to be added there as well

nanaya (dismissed):

venix12:
thought you'd tell me to redate migrations before merge :eyes:

PR #9729, opened 2023-01-11
Add classic mod icon

### LINKED ISSUE #9688: (lazer.ppy.sh) classic badge not showing next to scores with it
In both Firefox and Chrome, the classic badge is not showing up on my profile next to scores that have the mod applicable to it. However, the badge has a tip that shows up when you hover over the area in which it should be. These tags appear just fine when viewing my profile from within the game itself though. 

Website version
Screenshot 2022-12-28 at 9 57 52 AM

In game version
Screenshot 2022-12-28 at 9 57 04 AM

nanaya:
Please don't look too close at my hackjob 👀 The background doesn't quite exactly match with the one on lazer but then again there's no other mods on web currently using this background colour.

@arflyte maybe want to create the actual mod icon file?

Resolves #9688.


peppy (approved):
Close enough to my eye.

PR #9736, opened 2023-01-12
Split client and web registration endpoints

nanaya:
CSRF requirement with web registration makes it rather difficult to return correct error message.

Also add tests which discovered confirmation field check isn't actually working 👀

Although as there's already check on frontend it's probably not actually needed?

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -211,108 +211,56 @@ class UsersController extends Controller
 
     public function store()
     {
-        if (!config('osu.user.allow_registration')) {
-            return abort(403, 'User registration is currently disabled');
+        if (config('osu.user.registration_mode') !== 'client') {
+            return response([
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/UsersController.php
```diff
@@ -211,108 +211,56 @@ public function extraPages($_id, $page)
...
+        if (!starts_with(Request::header('User-Agent'), config('osu.client.user_agent'))) {
+            return error_popup(osu_trans('users.store.from_client'), 403);
+        }
 
-        if (IpBan::where('ip', '=', $ip)->exists()) {
-            return error_popup('Banned IP', 403);
+        return $this->storeUser(request()->all());
+    }
+
+    public function storeWeb()
```
notbakaneko, line 228: guest middleware?

notbakaneko (changes requested):

notbakaneko (approved):

PR #9750, opened 2023-01-16
Remove lazy-load class when loaded

### LINKED ISSUE #9748: In some cases, having a too long beatmap title in your profile breaks the page
Screenshot_2023-01-16-09-11-22-15_3aea4af51f236e4932235fdada7d1643

Sometimes when I'm looking to my profile on mobile, the page gets stretched, as the width of beatmap cards matches the longest title ("We Could Get More Machinegun Psystyle! (And More Genres Switches!)" in the capture). As the title is very long, the text should be cropped, but in this case it's not cropped.

Only all beatmap cards from the "Beatmap" section seems to be affected by this bug. Also, I can't see the left border of these cards, so that's why it's cropped.

It doesn't happens everytime. I got the bug twice and in this situation, the "beatmap" section of my profile (and other sections too) were loading, and once the information got loaded, the beatmap cards appeared and the page's width got broken.

Bug achieved the 2022/01/16 around 9am GMT+1 on Firefox Mobile 105.1.0.

My osu! profile: https://osu.ppy.sh/users/15646039

nanaya:
Resolves #9748.

### DIFF
--- a/resources/assets/less/bem/lazy-load.less
+++ b/resources/assets/less/bem/lazy-load.less
@@ -2,20 +2,18 @@
 // See the LICENCE file in the repository root for full licence text.
 
 .lazy-load {
-  display: flex;
   // TODO: need better min-height for different sections?
   // most extra pages will be larger than this unless they have less than the default number of items.
   min-height: 150px; // excludes the non-lazy-loaded part (e.g. headers)
-  align-items: center;
-  justify-content: center;
+
+  &--loading {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+  }
 
   &__button {
     text-align: center;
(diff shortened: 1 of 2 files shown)


notbakaneko (changes requested):
revert `7ad99cda507d4c54e142f3a1087a055beb3d6db1` instead

This is the specific issue it's working around; as far as I can tell it only happens on Safari
https://user-images.githubusercontent.com/1694544/212641327-e6bf24ad-5560-44d7-aab1-3b1ed06a2eda.mov

notbakaneko (approved):

Corne2Plum3:
Wow that was fast

PR #9753, opened 2023-01-16
Larger form for creating announcements

notbakaneko:


resources/assets/less/bem/chat-form.less
```diff
@@ -4,19 +4,20 @@
 .chat-form {
   --input-size: @font-size--title-small-3;
 
+  display: flex;
+  flex-direction: column;
   flex: 1;
   background: hsl(var(--hsl-b4));
 
   &__button-bar {
```
nanaya, line 12: the bar should probably be get rid of entirely so the margins are uniform

nanaya (changes requested):
I have a feeling I asked about this before and the reply was something around the line it was the design 🤔

nanaya (approved):

Issue #9741, opened 2023-01-14
Issues with Announcement Messages


RockRoller01:
Returning to using announcement messages after a while, the issue in #9002 still persists, and I have noticed a couple new ones

- descriptions feel useless. in my experience they are just a hassle to fill out and a duplicate of the title
- the GUI does not show the limits for lengths. These are title (50), description (255), message (1024)
- the API docs do not mention these limits either
- API docs have a mistake, they say that `target_ids` is required if PM, else ignored. In reality its required for ANNOUNCE, ignored for others.
- descriptions dont wrap if there is no spaces (very minor thing, noticed this while testing with random inputs, but i can totally see someone just holding down `-` because descriptions are annoying, maybe relevant on mobile with a very long word..?)
- there seems to be no check besides the GUI for title and description length, trying to send something longer via api (or console, or modify maxlength) results in 500 error
- the editor not filling the full screen height and just ending somewhere in the middle feels weird

![image](https://user-images.githubusercontent.com/35633093/212442365-f1ee5b0c-754d-458c-a65a-da951473b468.png)

- when you try to send another message in the announcement chat you are limited to the normal chat editor, which means you can not type a new line manually, only get it by copy-pasting into the chat box.
- there is no mention about what markdown features are supported, probably should be a wiki page and then get linked here
- headings are supported, but their styling seems weird to me. probably should be a different colour, like [heading] on the forums
- h2 is bigger than h1 for some reason, h1 and h3 very very similar

![image](https://user-images.githubusercontent.com/35633093/212443692-8743f8fb-5663-46f5-aa52-6520c5da916c.png)

- notification e-mails for announcement messages still seem to not be working (not sure if they are even implemented)

Overall a pretty long list of things that should be improved with announcement messages, many of them actually being relevant and not just "oh, you will find these if you do intense testing for edge cases" things. I can split this list into a bunch of seperate issues if wished for.

PR #9754, opened 2023-01-16
Validate channel property lengths

notbakaneko:
#9741

### DIFF
--- a/app/Models/Chat/Channel.php
+++ b/app/Models/Chat/Channel.php
@@ -52,6 +52,10 @@ class Channel extends Model
 
     const ANNOUNCE_MESSAGE_LENGTH_LIMIT = 1024; // limited by column length
     const CHAT_ACTIVITY_TIMEOUT = 60; // in seconds.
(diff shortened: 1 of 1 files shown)


app/Models/Chat/Channel.php
```diff
@@ -389,10 +393,22 @@ public function isValid()
 
         if ($this->name === null) {
             $this->validationErrors()->add('name', 'required');
+        } else if (mb_strlen($this->name) > static::MAX_LENGTHS['name']) {
```
nanaya, line 396: can use similar logic with existing `MAX_FIELD_LENGTHS` in other models. May even refactor the check to be shared between models.

nanaya (changes requested):

app/Models/Chat/Channel.php
notbakaneko: so that's the name I was looking for :dernk:

nanaya (approved):

PR #9760, opened 2023-01-18
Allow multi-line input in announcement channels

notbakaneko:
`shift-enter` for newline.

Also removes the input box from announcement channels if you're not allowed to send messages.

 #9741


resources/assets/lib/chat/conversation-view.tsx
```diff
@@ -192,53 +193,59 @@ export default class ConversationView extends React.Component<Props> {
       return <div className='chat-conversation' />;
     }
 
+    const renderInput = !(!channel.canMessage && channel.type === 'ANNOUNCE');
```
nanaya, line 196: this is a rather difficult to read conditional... (was it originally `hideInput` or something)

```suggestion
    const renderInput = channel.canMessage || channel.type !== 'ANNOUNCE';
```

nanaya (changes requested):

nanaya:
oh and maybe this message isn't needed if the input box isn't even visible? 🤔

nanaya (approved):

PR #9763, opened 2023-01-18
Move up resource directory structure

nanaya:
The whole assets directory is kind of silly being already in "resources". It's also how laravel is doing it nowadays. coffee directory is merged with js since there are only so few of them now. `build` is renamed to `builds` because that's what I've been using <del>in rails jsbundling</del> in other places.


notbakaneko (approved):
It still builds 👀

PR #9766, opened 2023-01-19
Fix mod background styling

### LINKED ISSUE #9765: Lazer mod icons cut off sometimes
Seen here https://lazer.ppy.sh/users/3771576/osu (pinned scores):

image

all the mods in the middle column have an extra bit at the bottom.

nanaya:
Resolves #9765. Kind of. The label will be too large in cases where the element size is in wrong proportion (due to flexbox sizing etc).

There are more problems with rendering many mods but those are for another time...

### DIFF
--- a/resources/assets/less/bem/mod.less
+++ b/resources/assets/less/bem/mod.less
@@ -61,6 +61,8 @@
     .full-size();
     background-image: var(--type-bg);
     background-size: contain;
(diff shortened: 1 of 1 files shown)


notbakaneko (approved):
wonder if there should be a function for this or something considering how often it appears 👀

PR #9775, opened 2023-01-22
Add coverage setting to Fade In and Hidden for Mania

mbnkwm:
support for https://github.com/ppy/osu/pull/21725

### DIFF
--- a/database/mods.json
+++ b/database/mods.json
@@ -286,7 +286,7 @@
       },
       {
         "Acronym": "TP",
-        "Name": "Target",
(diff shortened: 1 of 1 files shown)


peppy:
@mbnkwm for future reference, the `mods.json` file is automatically generated using https://github.com/ppy/osu-tools (`dotnet run --project PerformanceCalculator -- mods`)

I've run that command after updating game packages (https://github.com/ppy/osu-tools/pull/174) to ensure we're in sync.

peppy (approved):

PR #9777, opened 2023-01-23
Update scribe

nanaya:
Some large changes. Mostly from adding the full css (minus normalization) and javascript though so it's not really all that bad 👀

I didn't clean up the css beyond setting smaller font size.


notbakaneko (changes requested):
Other style fixes/improvements can be made later

notbakaneko (approved):

PR #9780, opened 2023-01-23
Add supported markdown guide to creating announcement UI

notbakaneko:
Adds a button 
image

that pops up a guide for the markdown that is supported
image

Not sure it should be a wiki page instead or pull markdown from somewhere else (or if this kind of input guide would be useful in other places) 🤔 
Draft because this is largely a PoC 👀 

Could use some ideas for the formatting of the display @arflyte


RockRoller01:
Some things are missing from this guide:
- bold emphasis
- underscore based syntax

PR #9783, opened 2023-01-24
Convert beatmapset discussions mode switcher to typescript

notbakaneko:


resources/js/beatmap-discussions/discussions.tsx
```diff
@@ -64,7 +64,7 @@ interface Props {
   currentBeatmap: BeatmapExtendedJson;
   currentDiscussions: CurrentDiscussions;
   currentFilter: Filter;
-  mode: DiscussionsMode;
+  mode: Exclude<DiscussionsMode, 'events'>;
```
nanaya, line 67: this type looks still common enough to have its own name

nanaya (changes requested):

nanaya (approved):

PR #9788, opened 2023-01-25
Fix beatmap discussions tab order

### LINKED ISSUE #9787: General (all difficulties) moved to the right somehow
firefox_2023-01-25_18-59-52
idk why this is moved. the last design seem makes much more sense in order of importance (idk the right word) from left to right. any reason why this was moved?

notbakaneko:
closes #9787


nanaya (changes requested):
should put a comment about it also being display order

nanaya (approved):

PR #9790, opened 2023-01-25
Restore missing report-form-select-options css

### LINKED ISSUE #9789: Report pop-up looks bad and unintuitive
Issue on latest Opera GX, chrome, but havent checked other browsers.

Asked in #osu-web and was told to go here :shrug:

Initially had an extension, then turned if off and both looked the same so no real extension issue on my hand.

CyNbGOG

notbakaneko:
this whole `${bn}` string interpolation for class names ಠ_ಠ

closes #9789


nanaya (approved):
it's not the bn interpolation but it's building block name off another block name 😐

PR #9793, opened 2023-01-26
Use modifiers to modify select options behaviour

nanaya:
That was weird 😐

Conversion of the component to typescript will be done separately.


notbakaneko (changes requested):
The `beatmap-discussions-user-filter` one also need to apply `--group-colour` to the text colour

notbakaneko (approved):

PR #9794, opened 2023-01-26
Convert SelectOptions to TypeScript

nanaya:
The generics don't seem quite right but it at least compiles 🫥

- [x] depends on #9793


resources/js/components/select-options.tsx
```diff
@@ -0,0 +1,160 @@
...
+  private renderOption({ children, onClick, option, selected = false }: ComponentOptionRenderProps<T>) {
+    const cssClasses = classWithModifiers(`${bn}__option`, { selected });
+
+    if (this.props.renderOption != null) {
+      return this.props.renderOption({ children, cssClasses, onClick, option });
+    }
+
+    return (
+      <a
+        key={typeof option.id === 'string' ? option.id : ''}
```
notbakaneko, line 128: can restrict T to `string | number` and not need the `typeof option.id` (and apparently `null` is a valid key but not `undefined` 🤔 )

resources/js/components/select-options.tsx
```diff
@@ -0,0 +1,160 @@
...
+  static readonly defaultProps = { blackout: true };
+
+  private readonly blackoutAutoToggleDisposer;
+  private readonly ref = React.createRef<HTMLDivElement>();
+  @observable private showingSelector = false;
+
+  constructor(props: Props<T>) {
+    super(props);
+    makeObservable(this);
+    this.blackoutAutoToggleDisposer = reaction(
```
notbakaneko, line 51: can use `disposeOnUnmount` but not sure if actually better 🤔

resources/js/components/select-options.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+  private renderOption({ children, onClick, option, selected = false }: ComponentOptionRenderProps<T>) {
+    const cssClasses = classWithModifiers(`${bn}__option`, { selected });
+
+    if (this.props.renderOption != null) {
+      return this.props.renderOption({ children, cssClasses, onClick, option });
+    }
+
+    return (
+      <a
```
notbakaneko, line 127: I can't remember the reason this wasn't changed to `button` 👀

notbakaneko (changes requested):

resources/js/components/select-options.tsx
nanaya: rendering difference aside, button can't contain div 🤔 (can probably also updated but that'll end up with quite a few unrelated updates)

notbakaneko (changes requested):
conflicts 💥

notbakaneko (approved):

PR #9797, opened 2023-01-26
Remove force-version flag when calling starting score indexer

notbakaneko:
We're looking at remove the `--force-version` flag from the score indexer ppy/osu-elastic-indexer#137; github actions can just set the version with `run` before the tests.

~~This still has the issue for with docker compose where the score index isn't aliased after the docker containers start and has to be done separately (which is what `force-version` got around)~~

Tries to set the score indexer schema version when a fresh migration is run so the score indexer can setup the alias for it.
Uses a loop in case the score indexer hasn't finished starting yet - the alternative is `healthcheck` on score-indexer which needs more things installed into the image as well as parsing the output, so I just went with adding it to the migration command.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -142,6 +142,7 @@ jobs:
 
       - name: Setup indices
         run: |
+          php artisan es:index-scores:set-schema --schema=1
           php artisan es:create-search-blacklist
           php artisan es:index-documents --yes
           php artisan es:index-wiki --create-only --yes
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,6 +13,7 @@ env:
   DB_HOST: 127.0.0.1
   ES_SOLO_SCORES_HOST: http://localhost:9200
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  SCHEMA: 1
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -141,6 +141,7 @@ jobs:
 
       - name: Setup indices
         run: |
+          php artisan es:index-scores:set-schema --schema=1
```
nanaya, line 144: the actual schema version should be defined using env instead so it always matches docker compose

nanaya (changes requested):
adding set schema command in docker dev prepare script is probably fine

peppy:
@notbakaneko was this intended to address the issue mentioned in https://github.com/ppy/osu-elastic-indexer/pull/137#issuecomment-1404973857? If so can we revisit it? Would like to cross that PR off the list.

notbakaneko:
I remember running into an issue on the indexer side when testing the changes and didn't get further than that yet

notbakaneko:
`set-schema` needs to be after the indexer starts in order to alias any newly created index correctly

nanaya:
So what to do with dev env here?

nanaya:
is it just me or keeping `--force-version` would be way simpler?

notbakaneko:
All the additional stuff isn't needed
#10477

PR #9800, opened 2023-01-27
Add option to send notification when webpack finishes building

nanaya:
Put notification config in `.webpack-build-notifier-config.js` and magic will happen (or not if the config is wrong).

My config for example

```javascript
const { exec } = require('child_process');

module.exports = {
  onComplete(_compilation, status) { exec(`/home/edho/bin/notify-windows-webpack ${status === 'success' ? 'ok' : 'fail'}`) }
}
```


notbakaneko:
Wonder if it should be moved to `devDependencies` and only load if available 🤔

notbakaneko (approved):

PR #9802, opened 2023-01-27
Reformat `report-form.tsx` to match other forums

Blooshing:
First time contributing to osu-web, sorry if I am making any contribution issues, but I figure that this was a regression or was something unintentional and decided to check out the code for myself.

image (sorry peppy, it was just an example :P). 

It seems that it used to be that the `Other (type below)` option migrated from the very last selection to a different spot entirely. This doesn't tend to match up with any other online forums, but it definitely doesn't need to be in this part, just figured the team would want it to match. 

Not entirely sure if i need to move it elsewhere on other documents as well, but here we go.


venix12:
Would possibly be helpful to add a comment explaining that alphabetical order is not intended.

Blooshing:
> Would possibly be helpful to add a comment explaining that alphabetical order is not intended.

Curious as to if the order of the selection is intended due to priority? Or if putting it in alphabetical order would be better?

venix12:
I believe it makes most sense for "other" to be at the very bottom considering you'd logically read all other options before selecting one that fits whatever is supposely not specified.

nanaya (changes requested):
in addition to comment about it being display order, might as well fix the order for the whole thing

https://github.com/ppy/osu-web/blob/5f2d10a342e8136a6c867a640a3863bddd7db109/resources/assets/lib/components/report-form.tsx#L14-L22

Blooshing:
Should be good now? Added the note and reordered the form per the one that @nanaya provided.

Blooshing:
Should be good now? Added the note and reordered the form per the one that @nanaya provided.

nanaya (approved):

nanaya (approved):

PR #9803, opened 2023-01-28
clarify newline type

tybug:


nanaya (changes requested):
I think the splitter should be updated to handle all cases instead

PR #9808, opened 2023-01-30
Support timestamp type for supporter expiration time column

nanaya:
@peppy please merge this beforehand if we're going ahead with timestamp datatype update for the `osu_subscriptionexpiry` column. This should handle either type of column.


notbakaneko (commented):
`timestamp` or `datetime`? 👀

nanaya:
datetime because there are some people with >2038 supporter. from laravel standpoint it's the same thing in the end 😴

notbakaneko (approved):

PR #9810, opened 2023-01-30
Fairer duration for supporter tags

nanaya:
1 month is now 730 hours.

depends on:
- [x] #9808 (required before migration)
- [x] migrate the column in production and add entry `2023_01_30_080509_use_datetime_for_supporter_expiry`


notbakaneko (approved):

peppy:
Migration entry added, migration is running in background (~1 hour).

peppy (approved):

PR #9811, opened 2023-01-30
Convert Beatmap discussion url helpers to typescript

notbakaneko:
Convert remaining beatmap discussion helpers to typescript.

Was originally going to make the typing for `makeUrl` param stricter but the additional code to handle the "either but not both" cases properly turned out to be a mess ( ﾟ ヮﾟ)a


resources/js/utils/beatmapset-discussion-helper.ts
```diff
@@ -36,6 +49,26 @@ interface PropsFromHrefValue {
   target?: '_blank';
 }
 
+interface UrlOptions {
```
nanaya, line 52: I feel this and DiscussionPageUrlParams should have more similar name. Or maybe let the urlParse return this anyway.

nanaya (changes requested):

nanaya (approved):

PR #9813, opened 2023-01-31
Disallow taking over inactive top player usernames

### LINKED ISSUE #9804: Disallow all username changes to handles held by top players
As raised in an email to me recently, a specific, now-deceased top player's name is about to become available for name changes. This is an ongoing problem not just for deceased or inactive players, but also for any player who has been in the public eye by virtue of being at the top of the leaderboards.

An easy solution to fix this would be to add the usernames (including all previous ones) of any players from the top 100 via peak rank (in all game modes) to the disallowed usernames list a few times a year.

nanaya:
Resolves #9804.

### DIFF
--- a/app/Libraries/UsernameValidation.php
+++ b/app/Libraries/UsernameValidation.php
@@ -82,7 +82,7 @@ class UsernameValidation
     {
         $errors = new ValidationErrors('user');
 
-        $users = static::usersOfUsername($username);
+        $users = static::usersOfUsername($username)->load('rankHighests');
         foreach ($users as $user) {
             // has badges
             if ($user->badges()->exists()) {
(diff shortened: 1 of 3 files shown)


nanaya:
oh yeah I forgot about that 😐

app/Libraries/UsernameValidation.php
```diff
@@ -95,7 +96,10 @@ public static function validateUsersOfUsername(string $username): ValidationErro
             }
 
             // top 100
-            if ($user->rankHighests->contains(fn ($r) => $r->rank <= 100)) {
+            // Queried directly on model because User::rankHighests is disabled
+            // when experimental rank is set as default.
+            $highestRank = RankHighest::where('user_id', $user->getKey())->min('rank');
```
notbakaneko, line 101: Does this work outside the loop with `whereIn(..., $users` if it's using `min` since as long any previous user meets the requirement it should be locked?

nanaya: that seems applicable to the others as well but I'm changing those here 🤔

notbakaneko (approved):

PR #9818, opened 2023-02-01
Add coloured marker for Beatmap discussion type on mobile view

### LINKED ISSUE #7207: Can't distinguish issue types on modding pages on mobile

notbakaneko:
On the right side since the left side has the resolved marker and stacking the bars looked a bit odd.

closes #7207


resources/js/beatmap-discussions/discussion.tsx
```diff
@@ -119,6 +119,9 @@ export class Discussion extends React.Component<Props> {
         className={`${topClasses} js-beatmap-discussion-jump`}
         data-id={this.props.discussion.id}
         onClick={this.handleSetHighlight}
+        style={{
+          '--discussion-colour': `var(--beatmapset-discussion-colour--${kebabCase(this.props.discussion.message_type)})`,
```
nanaya, line 123: I think it's fine to use the value as is. It looks a bit weird but otherwise simplifies the handling of this.

...and also there are a few mixups in some places regarding casing of this thing

nanaya (changes requested):

nanaya:
oh an icon is probably be easier to tell 👀  (left to the votes button maybe?)

notbakaneko:
well the point is the bar is still visible even if the discussion thread is long but then again desktop view has the same problem 🤔

nanaya:
what's phpstan files doing here 🤨

notbakaneko:
oh wait 👀

resources/css/bem/beatmapset-discussions-chart.less
```diff
@@ -71,31 +71,31 @@
     }
 
     &--mapper-note {
-      --icon-colour: @beatmap-discussion--color-mapper-note;
+      --icon-colour: var(--beatmapset-discussion-colour--mapper-note);
```
nanaya, line 74: wrong case 👀

resources/css/bem/counter-box.less
```diff
@@ -5,7 +5,7 @@
   @top: counter-box;
 
   @color-deleted: black;
-  @color-mapper-notes: @beatmap-discussion--color-mapper-note;
+  @color-mapper-notes: var(--beatmapset-discussion-colour--mapper-note);
```
nanaya, line 8: also wrong case 👀

nanaya (changes requested):

resources/css/bem/beatmapset-discussions-chart.less
notbakaneko: somehow I corrected the names on the branch I was rebasing but not this one 🤨

resources/css/bem/beatmap-discussion-timestamp.less
```diff
@@ -18,13 +18,17 @@
 
   &__icon {
```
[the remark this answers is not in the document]
notbakaneko: `icons-container` seems to be there to offset both icons and timestamp text on timeline 🤔

resources/css/bem/counter-box.less
notbakaneko: wonder if the colour for praise should be change to use the variable as well, it's almost the same 👀

nanaya (approved):

PR #9822, opened 2023-02-01
Include pack tags in beatmapset search/show response

### LINKED ISSUE #8528: Add Beatmap Packs to ApiV2
Currently the beatmap object does not include the packs the beatmap is bundled in.
The only way to receive that info is using the ApiV1, which also seems to be using an old way of identifying the packs.  E.g. beatmap id `129891` would give back packs `R129,R83,R86,SA10,T66`, which if you were to put one of them into an url (https://osu.ppy.sh/beatmaps/packs/T66), it would redirect to a different pack id (https://osu.ppy.sh/beatmaps/packs/1119).

Might be related to packs not even showing on new site yet (https://github.com/ppy/osu-web/issues/2025).

nanaya:
I'll look into adding this by default if the need ever come (as in becoming part of beatmapset card) 🤷

Initially returning id but it looks like tag would be more useful? In that case it sure would be nice if the join table uses tag instead of id...

Resolves #8528.


notbakaneko (approved):
I though I reviewed this already 🤔

PR #9824, opened 2023-02-01
Recursively delete discussion post

### LINKED ISSUE #8700: Discussion replies are visible in Modding History if parent discussion is deleted
image
image

Really shouldn't be visible still. 

links:
https://osu.ppy.sh/users/15376632/modding/posts
https://osu.ppy.sh/beatmapsets/1487999/discussion/-/generalAll#/2928352

nanaya:
It's only doable by moderator when there are replies and in such cases the entire thread is intended to be gone.

This should resolve #8700 although it doesn't retroactively delete existing replies. <sub>I'll figure out a query to do it one of these days maybe</sub>

### DIFF
--- a/app/Models/BeatmapDiscussion.php
+++ b/app/Models/BeatmapDiscussion.php
@@ -659,6 +659,7 @@ class BeatmapDiscussion extends Model
                 BeatmapsetEvent::log(BeatmapsetEvent::DISCUSSION_RESTORE, $restoredBy, $this)->saveOrExplode();
             }
 
+            $this->beatmapDiscussionPosts()->where('deleted_at', $this->deleted_at)->update(['deleted_at' => null]);
             $this->update(['deleted_at' => null]);
             $this->refreshKudosu('restore');
         });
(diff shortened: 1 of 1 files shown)


app/Models/BeatmapDiscussion.php
```diff
@@ -659,6 +659,7 @@ public function restore($restoredBy)
                 BeatmapsetEvent::log(BeatmapsetEvent::DISCUSSION_RESTORE, $restoredBy, $this)->saveOrExplode();
             }
 
+            $this->beatmapDiscussionPosts()->where('deleted_at', $this->deleted_at)->update(['deleted_at' => null]);
```
notbakaneko, line 662: this means post previously deleted separately also get restored, right?

nanaya: the where should exclude those out 🤔

notbakaneko: I did not see that 👀

notbakaneko (approved):

PR #9828, opened 2023-02-02
Show link to wiki page in user restriction message

### LINKED ISSUE #4581: Make the "Account Restrictions" wiki page more visible to restricted users

nanaya:
Resolves #4581. Not sure about the wording. I removed the contact email link because the person should probably read the wiki page first.


resources/lang/en/users.php
```diff
@@ -157,7 +157,8 @@
     ],
     'restricted_banner' => [
         'title' => 'Your account has been restricted!',
-        'message' => 'While restricted, you will be unable to interact with other players and your scores will only be visible to you. This is usually the result of an automated process and will usually be lifted within 24 hours. If you wish to appeal your restriction, please <a href="mailto:accounts@ppy.sh">contact support</a>.',
+        'message' => 'While restricted, you will be unable to interact with other players and your scores will only be visible to you. This is usually the result of an automated process and will usually be lifted within 24 hours. :link.',
+        'message_link' => 'Check this page to learn more',
```
Walavouchey, line 161: small thing but if the link contains the whole sentence then it might as well contain the period too.

```suggestion
        'message' => 'While restricted, you will be unable to interact with other players and your scores will only be visible to you. This is usually the result of an automated process and will usually be lifted within 24 hours. :link',
        'message_link' => 'Check this page to learn more.',
```

notbakaneko (approved):

PR #9829, opened 2023-02-02
Add bbcode tag for inline code

### LINKED ISSUE #7126: Markup tag for inline monospace text blocks ("code") on forums
(has no relation to #1121)

in addition to `[code]`, this would help emphasize things and separate them from regular text description without having to use unnecessary quotes for enclosure marks:

> When playing a beatmap, hit `Alt` + `F4` to return to the beatmap listing.
> To submit a beatmap, go to `File` > `Upload beatmap...`
> Type `!help` for a list of commands.

nanaya:
Saw this issue and figured it's easy enough to add... if we do want it, that is. Using `[c]` tag. Style was mostly copied from markdown but with lighter background and colour that doesn't look like link 🤨

Resolves #7126.

### DIFF
--- a/app/Libraries/BBCodeForDB.php
+++ b/app/Libraries/BBCodeForDB.php
@@ -134,20 +134,15 @@ class BBCodeForDB
         return $text;
     }
 
-    /*
-    * Handles:
-    * - Bold (b)
-    * - Code (c)
-    * - Italic (i)
-    * - Strike (strike, s)
-    * - Underline (u)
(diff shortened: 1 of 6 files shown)


notbakaneko (changes requested):
but where's the style? 🤔

notbakaneko:
also it works for non-inline
```
[c]
- have
- a
- list
[/c]
```

notbakaneko (approved):

PR #9835, opened 2023-02-02
Fix body params documentation for mark read notification endpoint

### LINKED ISSUE #8886: v2 endpoints return error
/beatmapsets/{beatmapset}/favourites
```json
{"error":"Invalid action"}
```
image

/me/download-quota-check
```json
{"error":null}
```
image

/notifications/mark-read

returning empty string and notifications still unreaded
```json
""
```
image

nanaya:
I'm not quite sure how to document this ಠ_ಠ It definitely needs more detailed description but at least this one should be less wrong than before.

Resolves #8886.


notbakaneko (approved):
The inconsistency looks like a side effect of when the the params handling was unified; it should probably be standardized to accepting `identities` only (and other fields ignored if `id` is specified) or accept the old format of an array of ids, imo

PR #9836, opened 2023-02-02
Don't link user to specific ruleset for multiplayer score listing

nanaya:
The entries don't have ruleset attached to it.


notbakaneko (changes requested):
conflicts

notbakaneko (approved):

PR #9839, opened 2023-02-06
Update react markdown

notbakaneko:
step 1 of ? to supporting partial markdown syntax in beatmap discussions outside of reviews.

This uses an updated `react-markdown` which now uses `micromark` for parsing. As a result, the plugins previously used have all changed and the method of parsing and tokenization are now different.

- `timestamp` is now decorated at render time instead of tokenized
- links are parsed and tokenized with the gfm link plugins
- disabling tokenization (as most markdown features are actually disabled for reviews) is done by passing the option to `micromark`

The separate text/html parsing/mangling/construction in discussions will probably be replaced in the future 

~~(still WIP/testing the review editor)~~ seems to work 👀


resources/js/beatmap-discussions/editor-helpers.ts
```diff
@@ -77,38 +77,39 @@ function serializeEmbed(node: EmbedElement): DocumentIssueEmbed {
...
   }
 
-  return childOutput.join('');
+  return text.replace(trimmedText, formattedText);
+}
+
+function serializeParagraph(node: ParagraphElement) {
+  return node.children.map((child) => {
+    if (child.text !== '') {
+      const text = child.text.replace(/\*/g, '\\*');
```
nanaya, line 99: should probably at least escape `\` as well (although at this point this looks like a losing battle esp if we're to support more formatting)

nanaya (changes requested):

nanaya (approved):
There are still some weird edge cases but it's not like the current one is any better anyway

PR #9844, opened 2023-02-09
More reliable notification banner positioning

nanaya:
Use sync height so no need to match header modifier and notification banner existence. The header modifier approach can't even deal with multiple banners.


resources/css/bem/osu-page.less
```diff
@@ -237,9 +237,10 @@
 
   &--notification-banners {
     @media @desktop {
-      position: relative;
-      height: 0;
+      position: absolute;
```
notbakaneko, line 240: needs to be updated for mobile as well 
image

notbakaneko (changes requested):

notbakaneko (approved):

PR #9851, opened 2023-02-10
Fix very wide hype button

### LINKED ISSUE #9850: Hype button is too long with more than 50 hypes
It seems like now when your map has over 5 hypes, there's more bars below the hype count, for example like this :
osuwebissue20230210-6

But I discovered with one of my map (1606 hypes), the hype count went out of the screen (the button is 8025 px wide wtf!)

osuwebissue20230210-1600

_It's not the first time that [my map breaks the website](https://github.com/ppy/osu-web/issues/6285) lol_

This bug occurs either you're logged in or not. By looking across all maps, this bug starts to appears with 50 hypes or more. First, each bar is 0 or 1 px wide, but it seems that the padding between the bars remains the same, which should explain the bug. The more there are bars, the more the poor hype button is stretched, and in more extreme case, the hype count isn't even visible.
osuwebissue20230210-50

My screen is 1920x1080 btw

notbakaneko:
Fixes #9850

(when you expect clamp to be between 2 numbers, but it's not 👀 )

### DIFF
--- a/resources/js/components/discrete-bar.tsx
+++ b/resources/js/components/discrete-bar.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
 // See the LICENCE file in the repository root for full licence text.
 
-import { clamp, times } from 'lodash';
+import { times } from 'lodash';
 import * as React from 'react';
 import { classWithModifiers, Modifiers } from 'utils/css';
 
--- a/resources/js/components/discrete-bar.tsx
+++ b/resources/js/components/discrete-bar.tsx
@@ -14,7 +14,7 @@ interface Props {
 
 export default function DiscreteBar(props: Props) {
   const current = Math.max(0, props.current);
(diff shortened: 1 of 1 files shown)


resources/js/components/discrete-bar.tsx
```diff
@@ -14,7 +14,7 @@ interface Props {
 
 export default function DiscreteBar(props: Props) {
   const current = Math.max(0, props.current);
-  const total = Math.max(1, clamp(props.total, current, props.total));
+  const total = Math.max(1, clamp(props.total, Math.min(current, props.total), props.total));
```
nanaya, line 17: I have a feeling it was just me at one point thought the bars should be at least as many as current count. Otherwise might as well just
```suggestion
  const total = Math.max(1, props.total);
```

nanaya (changes requested):

resources/js/components/discrete-bar.tsx
notbakaneko: not sure how well hundreds of bars in a small space would render well 🤔

nanaya (approved):

PR #9860, opened 2023-02-14
Mark read on notification middle click

### LINKED ISSUE #7687: Notifications not dismissing when using Middle Click
It seems as if comments from watched comments sections do not dismiss once one takes a look at them

07f1xJpPwZ

nanaya:
Resolves #7687. Also fixes clicking on timestamp not marking read.


notbakaneko (commented):
Should use `onMouseUp` to match `onClick` but then you'd also have to track if it was released in the same area 🤔

notbakaneko (approved):

PR #9866, opened 2023-02-16
Update to php 8.1

nanaya:
Upgrading to laravel 9 turned out to be a bit more annoying than expected so this can go in first instead. As additional bonus we may be able to just jump directly to 10 instead with this.

I'm not updating php specification in composer because notification and thumbnailer servers' php need to be updated separately. It'll be done as soon as it's deployed so I don't bother keeping old version in test matrix.

I'll update the branch protection rule accordingly once this pr is approved 👀

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -27,7 +27,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           tools: composer:v2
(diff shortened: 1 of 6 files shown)


notbakaneko (approved):

nanaya:
Protection rules updated 👀

PR #9868, opened 2023-02-16
Update date casting and fix missing composite key settings

nanaya:
Apparently this has been silently deprecated and then removed in laravel 10 ಠ_ಠ

Updated some missing composite key things along the way.


notbakaneko (changes requested):
`Comment`, `Contest`, `Event`, `Message` still have `$dates` 👀

notbakaneko (approved):

PR #9877, opened 2023-02-18
Use native lazy loading to lazy load images

### LINKED ISSUE #9876: Some images in profile Me! section doesn't load
I recently remade my osu! profile me! section (https://osu.ppy.sh/users/15646039), which is quite long, with a lot of images, but the problem is the website don't load some images, as shown below:

bruh1

It's supposed to load the following image (which is only 35 kB):

osu_profile_section_header_2-min

By looking at the inspect element, with the network tool, it seems that when I loaded the page, my browser didn't downloaded this image. Even scrolling around in the me! section seems to not work.

But there's a fix, which leads me to think this is a bug: just scroll OUTSIDE of the me! section, and OwO? the browser send the request to get and display the image.

bruh2

Seems like the website doesn't know when downloading hidden images in me! section. (it ruins my whole me! section that I took hours to make please fix ;_;)

notbakaneko:
`loading="lazy"` works in safari now 🤔 

closes #9876


nanaya:
might as well replace with `loading="lazy"`

notbakaneko:
is that a thing now 🤔

nanaya (approved):

PR #9878, opened 2023-02-18
Fix page change event triggered before page load

### LINKED ISSUE #9869: Osu-web seems to have reload issues
Forum post: https://osu.ppy.sh/community/forums/topics/1722739

Seems that going moving through forums, to user profile, to forums, then user profile can cause the user profile to not load (not sure if indefinitely) and needing to refresh the browser to get it to load. 

Some have experienced this with Chrome and Opera, where I have experienced this on safari and chrome. When on safari I had no extensions added, but with chrome I have extensions. 

image

The way I reproduce it on safari: 

1. Home page 
2. User profile 
3. Forums 
4. Home page 
5. User profile 
6. Sometime in between there the user profile or page will not load and give this issue.

nanaya:
Mainly in core-legacy/forum where for some reason one of its function (topicMeta) uses newBody when the rest of it uses currently displayed document.

Resolves #9869. (going to user profile from a forum topic show)

### DIFF
--- a/resources/js/components/header-v4.tsx
+++ b/resources/js/components/header-v4.tsx
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 import HeaderLink from 'interfaces/header-link';
+import core from 'osu-core-singleton';
 import * as React from 'react';
 import { classWithModifiers, Modifiers, urlPresence } from 'utils/css';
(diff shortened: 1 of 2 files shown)


notbakaneko (approved):
this works but I get the impression that the forum class shouldn't be calling boot every navigation in the first place 👀

PR #9884, opened 2023-02-20
Use debian for docker image

nanaya:
This enables using debian chromium which is also available on arm instead of just amd64.

Seccomp has been removed and chromium is no longer sandboxed due to headaches surrounding it.


notbakaneko (commented):
js tests seems to run fine but dusk doesn't seem connect to chromedriver? 🤔 

```
1) Tests\Browser\AccountTest::testUpdatePassword
Facebook\WebDriver\Exception\WebDriverCurlException: Curl error thrown for http POST to /session with params: {"capabilities":{"firstMatch":[{"browserName":"chrome","goog:chromeOptions":{"args":["--disable-gpu","--headless"]}}]},"desiredCapabilities":{"browserName":"chrome","platform":"ANY","chromeOptions":{"args":["--disable-gpu","--headless"]}}}

Failed to connect to localhost port 9515: Connection refused
```

notbakaneko (approved):

PR #9887, opened 2023-02-20
add link to api wrappers in docs

tybug:
As discussed [on discord](https://discord.com/channels/188630481301012481/188630616286167041/1077064481189941258). Wording is a bit awkward since there's only one wrapper to link to. Feel free to make wording changes or move the section around.

Wrapper link: https://github.com/circleguard/ossapi


NiceAesth:
Sorry to barge in, but if you need more wrappers to add I have also written once which is used in production by multiple projects

https://github.com/NiceAesth/aiosu

Also for js, cyperdark's library is one of the more widely used ones: https://github.com/cyperdark/osu-api-extended

notbakaneko (commented):
It needs a disclaimer, especially if we're going to start adding more third party options and we're not going to be able to check all of them regularly.

resources/views/docs/info.md.blade.php
```diff
@@ -18,6 +18,12 @@
 
 Current rate limit is set at an insanely high 1200 requests per minute, with burst capability of up to 200 beyond that. If you require more, you probably fall into the above category of abuse. If you are doing more than 60 requests a minute, you should probably give [peppy](mailto:pe@ppy.sh) a yell.
 
+# Wrappers
+
+There are community maintained wrappers for the api:
```
peppy, line 23:
```suggestion
Below is a list of some language-specific wrappers maintained by the community. Your mileage may vary when using them – please report any issues to the wrapper first before reporting back to us.
```

Something like this?

peppy:
> Sorry to barge in, but if you need more wrappers to add I have also written once which is used in production by multiple projects
> 
> https://github.com/NiceAesth/aiosu
> 
> Also for js, cyperdark's library is one of the more widely used ones: https://github.com/cyperdark/osu-api-extended

Would be good to add these in this PR.

tybug:
Definitely agree with adding [osu-api-extended](https://github.com/cyperdark/osu-api-extended), since it's the most widely used and complete js library (not 100% complete as it doesn't implement forum posting endpoints, but those are rarely used in practice). But does it make sense to add more than one wrapper per language? Both [ossapi](https://github.com/circleguard/ossapi) and [aiosu](https://github.com/NiceAesth/aiosu) mentioned above are python wrappers, with ossapi much more commonly used ([ossapi dependents](https://github.com/circleguard/ossapi/network/dependents), [aiosu dependents](https://github.com/NiceAesth/aiosu/network/dependents)).

If we're enumerating all wrappers with 100% endpoint coverage, then there's also a third complete wrapper ([osu.py](https://github.com/Sheepposu/osu.py)) for python, and will probably be more as time goes on. I'm not sure that linking to all wrappers is better than only the most polished or most used one for each language.

NiceAesth:
`osu.py` is an APIv2 only wrapper, not sure about that one.
`ossapi` is a wrapper that seems to only be usable for certain usecases because of how it stores tokens in files for cache and the fact it cannot do async, which led me to make my own. I personally think that they accomplish completely different things.

tybug:
async is definitely a valid concern, I actually just released `v3.1.0` of ossapi earlier today which provides an `OssapiAsync`, you can check it out here: https://circleguard.github.io/ossapi/async.html.

As far as tokens go, you can pass a [token_dir](https://circleguard.github.io/ossapi/appendix.html#module-ossapi.ossapiv2) to change where the tokens get stored, or override `Ossapi._save_token` to prevent writing tokens to disk at all if you prefer.

NiceAesth:
my concern is that you should not be using files to store any sort of data in a library period. having to override an underline method is insanely jank. either way, this is becoming very offtopic and i am inclined to think that this discussion will not lead somewhere productive.

NiceAesth:
also, to address your concern of multiple libraries for the same language: https://discord.com/developers/docs/topics/community-resources#libraries you may check what discord does, which also does the same thing.

jacksonisiah:
Could I suggest linking to [this](https://github.com/CirclePeople/osu-list#api-libraries) instead of trying to list a bunch of libraries there?  Currently I'm working to get a community-maintained, complete list of libraries that can work with any language that has one, and may be a good resource for people that are looking for this information anyway.

tybug:
I don't think it's healthy to fracture the ecosystem by linking to many different api wrappers - discord does, yes, but they have an enormous developer ecosystem and all libraries linked have over 500 stars, most with thousands.

But, it appears I'm in the minority in that opinion, so I've updated with the following list:

- [ossapi](https://github.com/circleguard/ossapi) (python)
- [osu.py](https://github.com/Sheepposu/osu.py) (python)
- [aiosu](https://github.com/NiceAesth/aiosu) (python)
- [osu-api-extended](https://github.com/cyperdark/osu-api-extended) (javascript)
- [rosu-v2](https://github.com/MaxOhn/rosu-v2) (rust)

rosu-v2 has some endpoints missing (changelog, beatmapset discussions, forum posting), but since it's the only decently complete rust library I figured it deserves to be listed.

nanaya:
Wrappers which include/use lazer token (password grant) can't be listed as it's not supposed to be used by anything but lazer client.

tybug:
Good point. I've removed osu.py ([ref](https://github.com/Sheepposu/osu.py/blob/f5177a3c2ce851ce577672d1ccf65bcf89730f25/osu/auth.py#L233)) and osu-api-extended ([ref](https://github.com/cyperdark/osu-api-extended/blob/a56972b942224a4350acabeffb61b48039ab7a29/src/utility/auth.ts#L58)). As far as I can tell, the three remaining wrappers don't allow lazer authentication.

nanaya (approved):

PR #9896, opened 2023-02-21
Ignore deleted post when checking nearby beatmap discussions

### LINKED ISSUE #7591: Soft-deleted issues/suggestions should not warn you when posting new things.
If you're a gmt/nat and can see those, the modding panel warns you that there's issues nearby, even if they're deleted

ex

I think this is nonsense and shouldn't be a thing
number of users affected: very few

nanaya:
Resolves #7591. The conditional is a bit hard to read ಠ_ಠ


notbakaneko (commented):
Maybe we should invert it as it's more like `not not nearby discussion` right now ಠ_ಠ

notbakaneko (approved):

PR #9901, opened 2023-02-22
Replace anonymous function transformer with primitive

nanaya:
Reuse variable and transformers while at it.


notbakaneko (approved):
Looks correct

PR #9903, opened 2023-02-22
Allow sorting forum topics by creation date

### LINKED ISSUE #1705: [Forum] No way to check the date of thread creation in the thread listing

### LINKED ISSUE #9891: Allow forums to be sorted by date created
One of the things people request in the [skinning poll](https://osu.ppy.sh/home/news/2023-01-16-skinning-poll-2023).

Browsing skins is a major pain when you can't sort by newest post, only by newest reply. Thought there already was an issue for this but couldn't find it. Its generally a good thing to have, even outside of skinning. (I also think this used to be a thing on the old forums?)

nanaya:
As it turned out there's existing sort option already.

Resolves #9891, resolves #1705 (sort of). <sub>haha sort</sub>

### DIFF
--- a/app/Models/Forum/Topic.php
+++ b/app/Models/Forum/Topic.php
@@ -424,12 +424,15 @@ class Topic extends Model implements AfterCommit
         $tieBreakerOrder = 'desc';
 
         switch ($sort) {
+            case 'created':
+                $query->orderBy('topic_time', 'desc');
+                break;
             case 'feature-votes':
                 $query->orderBy('osu_starpriority', 'desc');
                 break;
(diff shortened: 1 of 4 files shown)


notbakaneko (changes requested):
should the index on production be created separately? 🤔

nanaya:
The table (`phpbb_topics`) isn't that big so I assume it should be fine. @peppy ?

peppy:
Should be fine.

peppy (approved):

notbakaneko (approved):

PR #9912, opened 2023-02-28
Manual render of blank user card

nanaya:
Reduces number of active react components.


resources/js/components/user-card-tooltip.tsx
```diff
@@ -60,12 +61,28 @@ function createTooltipOptions(card: HTMLElement) {
   };
 }
 
+let blankCardCache: HTMLElement|undefined;
```
notbakaneko, line 64:
```suggestion
let blankCardCache: HTMLElement | undefined;
```

notbakaneko (changes requested):

notbakaneko (approved):

PR #9913, opened 2023-02-28
Typescript register component script

nanaya:
This should be functionally the same apart of missing dataset throwing error instead of passing undefined. I've checked the relevant ones and they seem to always pass the correct value.

- [x] #9912


resources/js/register-components.tsx
```diff
@@ -0,0 +1,153 @@
...
+));
+
+core.reactTurbolinks.register('spotlight-select-options', () => (
+  <SpotlightSelectOptions {...parseJson('json-spotlight-select-options')} />
+));
+
+core.reactTurbolinks.register('comments', (container) => {
+  const props: CommentsManagerProps = {
+    component: Comments,
+    ...reqJson<Omit<CommentsManagerProps, 'component'>>(container.dataset.props),
```
notbakaneko, line 86: properties being assigned here should come after the spread, otherwise it's misleading as `reqJson<Omit..` doesn't actually exclude the omitted property from the parsing.

notbakaneko (changes requested):

resources/js/register-components.tsx
nanaya: but it's more of indication the json doesn't contain component 🤔

notbakaneko: that's why it should be the other way around because it'll still parse `component` if it's in the json overriding the already set `component` when it's expected `component` to be always `Comment`

notbakaneko (approved):

PR #9918, opened 2023-02-28
Add featured artist pack prefix

### LINKED ISSUE #4408: Featured Artist beatmap pack category
rather than splitting featured artist packs between the "theme" and "artist/album" categories, for the future it'd be a good idea to add an additional "featured artist" map pack category.

This is because as the mappers' guild and featured artist projects grow, there's gonna be lots more of these packs, and splitting them between the already existing categories is potentially confusing.
image
image
Like it already is a bit confusing cause just 2 packs are in a different category from the rest.

This would also be fantastic because then we can look into setting up packs that'd cover all ranked maps of licensed featured artist songs, even outside of those tied to medals. It'd serve as a good way organize, find, and dl featured artist maps using the current system until https://github.com/ppy/osu-web/issues/3777 sees implementation.

edit: also i do believe this requires old web implementation so i hope ppy reads this

nanaya:
Not sure if the label should say "Featured" or "Featured Artist" 🤔 

Also I feel it should be the first tab?

Resolves #4408.

### DIFF
--- a/app/Models/BeatmapPack.php
+++ b/app/Models/BeatmapPack.php
@@ -21,11 +21,14 @@ use Exception;
 class BeatmapPack extends Model
 {
     const DEFAULT_TYPE = 'standard';
-    private static $tagMappings = [
+
+    // also display order for listing page
+    const TAG_MAPPINGS = [
         'standard' => 'S',
+        'featured' => 'F',
+        'chart' => 'R',
         'theme' => 'T',
         'artist' => 'A',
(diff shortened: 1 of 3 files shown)


Walavouchey:
definitely should be "Featured Artist" i'd say. just "Featured" doesn't make sense as a category among the others, both because they're not displayed in any special way and because all packs essentially "feature" maps anyway

notbakaneko (approved):

PR #9926, opened 2023-03-01
Disable beatmap discussions on beatmaps which have been download limited

### LINKED ISSUE #9924: Disable ability to edit / post discussions on beatmaps with `disabled` flag(s)
Similar to beatmap comments, the discussions page should not allow interaction when this flag is set.

notbakaneko:
closes #9924

### DIFF
--- a/resources/lang/en/authorization.php
+++ b/resources/lang/en/authorization.php
@@ -12,6 +12,8 @@ return [
     'unauthorized' => 'Access denied.',
 
     'beatmap_discussion' => [
+        'discussion_locked' => 'This beatmap is locked for discussion.',
+
         'destroy' => [
             'is_hype' => 'Can not undo hyping.',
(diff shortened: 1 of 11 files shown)


resources/lang/en/authorization.php
```diff
@@ -12,6 +12,8 @@
     'unauthorized' => 'Access denied.',
 
     'beatmap_discussion' => [
+        'discussion_locked' => 'This beatmap is locked for discussion.',
```
nanaya, line 15: not sure if worth moving the key esp when it's still called 'beatmap discussion' (instead of beatmapset) - also if anything this key is probably more applicable to the original (only discussion specifically locked) instead of the current one (beatmapset also "locked")...

nanaya (changes requested):

nanaya (approved):
Existing translations will need to be readded

PR #9931, opened 2023-03-02
Add option to show contest entry usernames during voting

### LINKED ISSUE #9459: Add option to show contest entry usernames during voting
Contests that involve voting for given publicly available beatmaps (as seen in the end of year community choice polls and things like the Beatmap Spotlights players picks) often have to have scripts include the usernames in the masked or ordinary entry names.

An extra option available to make the contest entry's associated ``user_id``/username pairing visible during voting, similar to [how they're displayed when voting has finalized](https://puu.sh/JqH0t/3eccb3c48c.png) would be very welcome for these use cases.

nanaya:
Resolves #9459.

Option is in `extra_options` json, field `show_entry_user` with value `true`.


notbakaneko (changes requested):
What's the difference between this and `show_names`? 😕

notbakaneko (changes requested):
probably can do with some alignment with both enabled
image

notbakaneko (changes requested):
Position  now appears to be missing when `show_names: true` and `show_votes=1`

notbakaneko (approved):

PR #9940, opened 2023-03-03
Update mod definitions

bdach:
Most importantly, includes the new "Fade out hit circles earlier" setting for classic mod. The setting is out in a lazer release, so deployment should probably be expedited.

Addresses https://github.com/ppy/osu/discussions/22779.

### DIFF
--- a/database/mods.json
+++ b/database/mods.json
@@ -401,6 +401,10 @@
           {
             "Name": "always_play_tail_sample",
             "Type": "boolean"
+          },
+          {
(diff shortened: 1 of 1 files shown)


nanaya (approved):

nanaya:
ping @peppy for deployment

peppy:
Good catch, we probably need to automate this at some point.

Deploying now.

bdach:
Opened https://github.com/ppy/osu-web/issues/9942 to track the possible automation of these updates.

PR #9947, opened 2023-03-06
Rearrange spotlights ranking header

### LINKED ISSUE #6335: Move sort tab in spotlight ranking header
Because of changes in https://github.com/ppy/osu-web/pull/6306, the spotlight header now has more empty space.

image

How about move it up like this?

image

nanaya:
Resolves #6335. Involves moving a _bunch_ of things (due to needing to split the filters).


resources/css/bem/grid-items.less
```diff
@@ -2,26 +2,49 @@
...
+
+  &--spotlights {
+    width: 100%;
+    gap: 0 20px;
+  }
+
+  &__item {
+    &--spotlight-user-filter {
+      width: 100%;
+      padding: 15px 0;
```
notbakaneko, line 42: This ends up making the gap between title and value on desktop quite large
image

Can be removed and replaced with `align-self: center`

notbakaneko (changes requested):

notbakaneko (approved):

PR #9958, opened 2023-03-07
Prioritise accept-language header for api endpoint

### LINKED ISSUE #8113: Handle API requests with a language different from web preferred
Continuation to ppy/osu#14662.

Currently osu!lazer's preferred language setting is decoupled from web and therefore any error responses from web would appear in the language selected web-side rather than lazer-side.

While coupling the client and web's language setting is one way to resolve this, going for an alternative way might be an easier solution in the interim (and perhaps better).

One proposal was to use the `Accept-Language` HTTP header, but, looking at osu!web now, I assume from [this code](https://github.com/ppy/osu-web/blob/04e0f42aebde3e12f8e3eef9bbe855925d2b8b49/app/Http/Middleware/SetLocale.php#L17-L46) that the `Accept-Language` is actually already considered but only when the user isn't signed in to the website, as a fallback locale provided by the browser, and therefore it can't actually be used as a solution.

Another way would be to provide the localisation key (e.g. `beatmapsets.panel.empty`) next to the message in the response, for osu!lazer to parse and display in the locally selected language, but I'm not sure if it's feasible to get the key, however.

nanaya:
Different behaviour added for `*` so it can be used when the http client enforces the header.

Resolves #8113.

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -21,7 +21,7 @@ class Kernel extends HttpKernel
     protected $middlewareGroups = [
         'api' => [
             Middleware\AuthApi::class,
-            Middleware\SetLocale::class,
+            Middleware\SetLocaleApi::class,
             Middleware\CheckUserBanStatus::class,
         ],
         'web' => [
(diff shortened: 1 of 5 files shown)


notbakaneko (changes requested):
Could use some tests for this

notbakaneko (approved):

PR #9960, opened 2023-03-08
Remove nonexistent app service registration

nanaya:
It was from many years ago and never actually used.


notbakaneko (approved):
binding non-existent class doesn't explode ಠ_ಠ

PR #9964, opened 2023-03-08
Fix sometimes passing null to urldecode

nanaya:
More deprecated thing.


notbakaneko (approved):

notbakaneko (changes requested):
wait, this also means `$this->relativeUrlRoot === null` is always false

notbakaneko (approved):

PR #9969, opened 2023-03-09
Fix editing beatmap discussion post on modding profile

notbakaneko:
Only update the post when editing a beatmap discussion post on the modding profile page instead of adding more discussions from the beatmapset onto the page 👀 

Also disables the UI for adding new discussions in reviews when editing from the modding profile and discussion history page since it doesn't seem to make sense to be able to add more without the context of the beatmap itself...?

This is largely a workaround until we can properly refactor the top-level handling on the beatmap discussions component itself.

### DIFF
--- a/resources/js/beatmap-discussions-history/main.coffee
+++ b/resources/js/beatmap-discussions-history/main.coffee
@@ -21,6 +21,7 @@ export class Main extends React.PureComponent
 
     @eventId = "beatmapset-discussions-history-#{nextVal()}"
     @cache = {}
+    @reviewsConfig = Object.assign({}, props.reviewsConfig, disableInsert: true)
     @state = JSON.parse(props.container.dataset.discussionsState ? null)
     @restoredState = @state?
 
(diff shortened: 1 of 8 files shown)


resources/js/beatmap-discussions-history/main.coffee
```diff
@@ -46,40 +47,20 @@ export class Main extends React.PureComponent
...
-        users = _.reject users, id: newUser.id
-
-      users.push(newUser)
-
-    @cache.users = @cache.discussions = @cache.beatmaps = @cache.beatmapsets = @state.relatedDiscussions = null
-    @setState
-      discussions: _.reverse(_.sortBy(discussions, (d) -> Date.parse(d.starting_post.created_at)))
-      users: users
-      relatedDiscussions: relatedDiscussions
+    { post } = options
```
nanaya, line 50: update from delete/restore endpoints doesn't have this

resources/js/modding-profile/main.coffee
```diff
@@ -44,6 +46,7 @@ export class Main extends React.PureComponent
     @cache = {}
     @tabs = React.createRef()
     @pages = React.createRef()
+    @reviewsConfig = Object.assign({}, props.reviewsConfig, disableInsert: true)
```
nanaya, line 49: that's an attempt but pasting still works (and this also removes insert paragraph button)

nanaya (changes requested):

nanaya:
I guess this one isn't needed anymore?

Issue #8751, opened 2022-03-28
New syntax for image map


nanaya:
A way to support something that's like [html map](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/map) would be useful. It won't use that exact syntax (because it's not responsive) but just an image and a bunch of absolute-positioned links. It'll be limited to rectangular links since otherwise the syntax will be rather unwieldy.

Something like this for the syntax
```
[imagemap]
imageurl
x y width height url1 optionaltitle
x y width height url2 optionaltitle
[/imagemap]
```

The bbcode editor will have a button to open a map editor where image can be specified and then the links can be added. Or maybe a dedicated page where the specific syntax above can be copy pasted around.

This will remove the need for #2060 by removing the need to deal with sliced images.

solstice23:
HTML map has multiple types of areas(poly, circle and rect) rather than only rectangles, which can make collabs more flexible and customizable. Maybe we can draw up new syntaxes for these different types of areas?

nanaya:
As mentioned in the starting post, it's not going to use map element due to it not being responsive.

nanaya:
I guess the button and help link are good enough "frontend" for now 👀

PR #9972, opened 2023-03-09
Add imagemap support to bbcode

nanaya:
Syntax support for #8751. Frontend for it is missing.

I just hope enterprising user will create one if there's enough interest 👀

```
[imagemap]
imageurl
x y width height url1 optional title
x y width height url2 optional title
x y width height # hover title
[/imagemap]
```

All the units (x, y, width, height) are in percent without the percent sign. Passing `#` as url will make the area only show title on hover (or nothing if no title specified).

```
[imagemap]
https://assets.ppy.sh/osu-web-test-resources/placeholder-1280x720.jpg
0 0 50 50 https://google.com q1
0 50 50 50 https://yahoo.com q2
50 0 50 50 https://osu.ppy.sh q3
50 50 50 50 # q4
[/imagemap]
```


RockRoller01:
Do I see it correctly that the URL is **not** optional? From my own experience with current imagemaps, making the url optional seems benefitial. For example in skinning we often only want some parts of the imagemap to be a link

nanaya:
image url isn't optional otherwise there's nothing to map to...

RockRoller01:
ofc, my comment was solely focused on the areas you can define for an imagemap

nanaya:
Now I think again, I'm not sure if the positioning should use percent or pixel 🤔 if anyone have preference of one over the other please comment here.

RockRoller01:
> if the positioning should use percent or pixel 

Looking at the code, wouldn't pixel result in a non-responsive design?

nanaya:
It'll be converted to percent

app/Libraries/BBCodeFromDB.php
```diff
@@ -129,6 +129,54 @@ public function parseHeading($text)
...
+                                    "width: {$link[2]}%",
+                                    "height: {$link[3]}%",
+                                ]),
+                                'title' => $link[5] ?? '',
+                            ]),
+                            $links,
+                        ));
+
+                        $imageUrl = proxy_media($map['imageUrl']);
+                        $imageSize = fast_imagesize($imageUrl) ?? [0, 0];
```
notbakaneko, line 162: I think there needs to be something better than rendering a 0x0 image if it fails to get the size, or just skip rendering the image. It's also not apparent to the user if something goes wrong here, they just get an image that gets displayed at 0px

notbakaneko (changes requested):

notbakaneko (approved):

PR #9973, opened 2023-03-10
Cleanup header nav styling

notbakaneko:
Something I came across while borrowing the nav for the markdown editor.

Switch to using `gap` instead of margin/padding offsetting

### DIFF
--- a/resources/css/bem/header-nav-v4.less
+++ b/resources/css/bem/header-nav-v4.less
@@ -4,7 +4,7 @@
 .header-nav-v4 {
   @_top: header-nav-v4;
   @header-nav-gap: 10px;
-  --header-nav-item-padding-y: 4px;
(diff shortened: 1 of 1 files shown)


resources/css/bem/header-nav-v4.less
```diff
@@ -3,88 +3,74 @@
...
-    padding-top: 0;
-    padding-bottom: 0;
     display: flex;
   }
 
   &--breadcrumb {
     display: flex;
+
+    @media @mobile {
+      --header-nav-item-padding-y: 10px;
```
nanaya, line 27: this looks like should/can be in the main block instead. I don't know where the 4px version is even visible.

nanaya (changes requested):

nanaya (approved):

PR #9980, opened 2023-03-13
Exclude explicit beatmaps popular listing from dashboard

### LINKED ISSUE #9979: Hide explicit tagged beatmaps from appearing on the osu! front page.
A newly ranked map "Faggot" hit ranked not too long ago and thus appears on the "New Ranked Beatmaps" and "Popular Beatmaps" section in the bottom right corner of the osu! front page. We checked to see if this was hidden for people who do not have the explicit option enabled, but it still appears in the listing, just without the background image thumbnail present.

image

It would probably be better to hide explicitly tagged beatmaps entirely from these two sections of the front page though, so people opting out of seeing this sort of content don't have to encounter it.

notbakaneko:
...and I guess latest ranked since it'll be a similar issue..? 🤷 
(this is the quick fix; better would be to cache the values according to user setting 👀 )

closes #9979

### DIFF
--- a/app/Models/Beatmapset.php
+++ b/app/Models/Beatmapset.php
@@ -172,7 +172,8 @@ class Beatmapset extends Model implements AfterCommit, Commentable, Indexable, T
     public static function popularIds()
     {
         $recentIds = static::ranked()
-            ->where('approved_date', '>', now()->subDays(30))
+            ->where('approved_date', '>', now()->subDays(360))
+            ->where('nsfw', false)
             ->select('beatmapset_id');
 
         return FavouriteBeatmapset::select('beatmapset_id')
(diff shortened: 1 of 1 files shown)


app/Models/Beatmapset.php
```diff
@@ -172,7 +172,8 @@ public static function popular()
     public static function popularIds()
     {
         $recentIds = static::ranked()
-            ->where('approved_date', '>', now()->subDays(30))
+            ->where('approved_date', '>', now()->subDays(360))
```
nanaya, line 175: a bit too long

notbakaneko: oh whoops 👀

nanaya (approved):

Issue #2761, opened 2018-03-21
lack of ability to search beatmap packs


dorothy3242:
May not be _as_ sorely needed depending on https://github.com/ppy/osu-web/issues/2759

The old beatmap pack pages was one tab/page per type of pack, allowing a user to quickly go through and use their browser's ctrl+f function to find a specific pack.
As the pack listings are now paginated, this becomes more difficult. 
Now if a user, say, notices the achievment for "Internet! Pack vol.1", they would have to
- find the pack listing page
- go through potentially 40+ pages in order to find it on page 3 of the themes listing. (there's nothing indicating that it'd be under the themes listing specifically.)

By the time someone would complete that, most people would simply lose interest in it altogether.
A search function for the packs listing would prevent this issue.

PR #9988, opened 2023-03-14
Increase number of beatmap packs per page

nanaya:
Not really solving #2761 but should help reducing number of pages to browse through. Reduces padding a bit so it's quite too long.


notbakaneko:
I recall this partly being limited due to performance from the animation (hi Safari) 🤔

notbakaneko (approved):
Performance doesn't seem too bad...at least on this machine 👀

PR #9991, opened 2023-03-14
Support markdown input and preview in beatmap discussions

notbakaneko:
continuation of #9843

Tries to have less deadspace for the posts and input; text now wraps around the top action buttons.
Also separates various component styles instead of trying to share and contort them into the proper positions since they're really only tangentially related...

currently borrows `header-nav-v4`'s style for the tabs 👀 

WIP while I find more missing stuff and debug more weird behaviour...


notbakaneko:
closing this since it's _way_ out of date now

PR #9992, opened 2023-03-15
Convert NewReview component to use mobx

notbakaneko:
A bit annoying working with this not being mobx when the other `NewThing` components are mobx

### DIFF
--- a/resources/js/beatmap-discussions/new-review.tsx
+++ b/resources/js/beatmap-discussions/new-review.tsx
@@ -27,6 +27,17 @@ interface State {
 }
 
 export default class NewReview extends React.Component<Props, State> {
+  private get noPermissionText() {
+    if (downloadLimited(this.props.beatmapset)) {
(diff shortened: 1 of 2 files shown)


resources/js/beatmap-discussions/new-review.tsx
```diff
@@ -76,31 +76,27 @@ export default class NewReview extends React.Component<Props, State> {
                   {trans('beatmaps.discussions.review.new')}
                   <span className='page-title__button'>
                     <span
```
nanaya, line 78: unrelated but this looks like should be a button

nanaya (changes requested):

resources/js/beatmap-discussions/new-review.tsx
notbakaneko: probably will do that separately (along with the one in `NewDiscussion` 👀

PR #10014, opened 2023-03-22
Remove type in timestamp tooltip

### LINKED ISSUE #6193: Translation dependent on gender
[This translation](https://crowdin.com/translate/osu-web/16/en#10666) is dependent on the gender of ":type" in some languages. Using multiple demonstrative pronouns (like "dieser/dieses :type") looks bad. The simplest solution would be to just make this snippet translatable for each of the 3 cases for ":type".

nanaya:
Resolves #6193. Technically people translating this could've done this themselves ("If this post (:type)") but looking again the English version seems kind of weird anyway with capitalised type. And it already says "Posts" for All difficulties tooltip.


notbakaneko (changes requested):
Probably should do something about the ones with the existing placeholders..

notbakaneko (approved):

PR #10020, opened 2023-03-22
Support reporting chat message

nanaya:
For use in lazer.


notbakaneko (approved):

cdwcgt:
Is comment necessarily required?
stable chat report should not require comments
image

peppy:
Might be best to allow all report types (except `other`) be valid to submit without any attached comment.

PR #10025, opened 2023-03-23
Show profile page navigation on mobile

### LINKED ISSUE #3630: Add the ability to collapse section in User Profiles
I've seen an proposal opened with the same concept but limited to the "Me!" section (and closed with no reason), but making every section collapsible would solve the common issue of profiles being not efficient to navigate.

### LINKED ISSUE #9968: profile page tabs
the old osu! profile page was legendary because you could see everything **you wanted to see** at once. on the new site we gotta scroll and scroll... especially the medals section is pretty big itself.

how about tabs you have to manually open just like before? https://old.ppy.sh/u/5851253

9uJRCGZ
- following account standing just below it (without opening manually).
- make me! userpage fixed on the top and automatically opened
- arrows would be good, too

pretty sure you can make it look way better than what i created, but you get the idea.

nanaya:
Resolves #9968, resolves #3630.

### DIFF
--- a/resources/css/bem/page-mode.less
+++ b/resources/css/bem/page-mode.less
@@ -37,13 +37,13 @@
 
   &--profile-page-extra {
     .default-gutter-v2();
-    padding-top: 0;
-    padding-bottom: 0;
-    height: 40px;
+    padding-top: 10px;
+    padding-bottom: $padding-top;
     display: flex;
     align-items: center;
     gap: 20px;
     font-size: @font-size--title-small;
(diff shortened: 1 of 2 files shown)


notbakaneko (changes requested):
https://github.com/ppy/osu-web/blob/1d0e40a71e5f3e78d1fa1c1e829ead01ef301526/resources/js/profile-page/main.tsx#L99 needs updating to include the bar height for mobile

notbakaneko (approved):

PR #10039, opened 2023-03-27
Fix select box width

nanaya:
Noticed in forum search page as it contains forum name that can be more the page width.


notbakaneko (approved):
also seems to be an issue with beatmapset metadata editor which this change seems to have no effect on 🤔 
image

PR #10040, opened 2023-03-27
Put beatmapset panel component in correct place

nanaya:
It doesn't make sense to have `beatmapset-panel/beatmaps-popup` when it really belongs specifically to to `components/beatmapset-panel`. I'm not sure if the whole thing should be on the root js directory or still inside components though 🤔 

Also I just learned that calling the file `index` makes it importable by just specifying the directory. That seemed interesting so I put that in as well 👀


notbakaneko (commented):
implicit `index` has always been with case when using `node` module resolution and is pretty much the standard way of doing it 👀

notbakaneko (approved):
..that's not the right option

notbakaneko (changes requested):
now there're conflicts 😑

notbakaneko (approved):

PR #10043, opened 2023-03-27
Escape brackets in user link part of report for discord

### LINKED ISSUE #9195: Square brackets are not escaped in reports
image

Just saw this report and it seems like the square brackets arent escaped, causing markdown links to potentially break.

nanaya:
Resolves #9195. Or more like "resolves".

Attempted:
- ❌ escape with backslash
- ❌ use html entity
- ❌ percent encode


notbakaneko (approved):
We should probably consider sending to the discord native webhook instead of the slack-compatible one at some point 🤷

PR #10050, opened 2023-03-28
Fix spoiler link overflowing

nanaya:
( ﾟ◡ﾟ)

Also apparently `anywhere` is the value I've been looking for.


notbakaneko (approved):
looks like repackaged `word-break: break-word` and finally supported by safari 👀

PR #10054, opened 2023-03-28
Only require comment when the reason is Other

nanaya:
As mentioned in #10020.


RockRoller01:
This is a regression from #8719, reports should always require comments, too many people otherwise file reports without enough detail

PR #10062, opened 2023-03-29
Update Setup.MD

ntbosscher:
Added nginx configuration details (from Laravel recommendations).  Added link to nginx + php-fpm setup tutorial (link is to an article that I wrote)

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -7,7 +7,7 @@ There are a few different options to get started:
 ### Install prerequisites
 
 - MySQL 8.0+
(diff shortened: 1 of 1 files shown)


bdach:
I feel that it is relevant to point out that looking at your github contribution history, you have submitted several pulls to various repositories that contain various links to `statuslist.app`. Seems like a bit of a disingenuous attempt at self-promotion to me, if you ask me.

nanaya:
there are problems with current docs but setting up with php-fpm isn't one of them as octane is the web server being used and quite a few parts preload data which don't make sense with fpm

PR #10063, opened 2023-03-30
Update readme to mention opencollective instead of legacy bounty system

peppy:
The legacy bounty system has mostly been deprecated these days. I want to simplify the messaging here in a way that users can reach out to me directly if they have specific concerns, otherwise direct them to opencollective (which has been working amazingly for regular contributors!).

Once this has been approved here I'll apply to all other repos which have the same message without PRs to save on process.

cc/ @ppy/team-web @ppy/team-client


bdach (approved):
Not entirely sure who should be approving this, but since I was tagged in OP then this is an acknowledgement approval from me I guess.

frenzibyte (approved):
This reads pretty well to me. One key difference between the two systems is that Open Collective is a bit more advanced and requires the contributor to enter the amount by dollars rather than by approximate time spent, but I think that's what the inquiry part is for.

smoogipoo (approved):

PR #10074, opened 2023-04-03
Automatic dataset sync for artist track search component

nanaya:
Testing out an idea so the state sync happens automatically instead of having to manually sync whenever it's updated. Could also be used for profile page controller which has more points where it updates the state (and they all currently have explicit save state call).


notbakaneko (commented):
Is it even useful to update the container data outside of unmount? 🤔

nanaya:
manually entering url and then pressing back?

notbakaneko:
> manually entering url and then pressing back? Or maybe not? 🤔

This seems to be the only case 🤔 
(and `saveState` in unmount definitely doesn't seem to update anything 🤔 )

notbakaneko (approved):

PR #10081, opened 2023-04-05
Confirm when closing description editor

### LINKED ISSUE #10077: Clicking out of the edit box when editing beatmap description silently discards all changes
As the title says; this is pretty prone to misclick imo.
Since there is confirmation dialog when clicking the "cancel" button explicitly, I think the same thing should be done when clicking out of the edit box.
Or maybe clicking out of the edit box should not quit editing at all.

nanaya:
Resolves #10077.

### DIFF
--- a/resources/js/beatmapsets-show/info.tsx
+++ b/resources/js/beatmapsets-show/info.tsx
@@ -25,6 +25,7 @@ interface Props {
 
 @observer
 export default class Info extends React.Component<Props> {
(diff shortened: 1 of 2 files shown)


notbakaneko (changes requested):
Pressing `esc` with the editor in focus and escaping the confirmation causes the dialog to trigger again

notbakaneko (approved):

PR #10085, opened 2023-04-06
Backend part of legacy api management

nanaya:
That's rather large.

- [x] depends on #10111


app/Models/ApiKey.php
```diff
@@ -18,7 +25,72 @@
  */
 class ApiKey extends Model
```
notbakaneko, line 26: could do with including `* @property-read User|null $user`

app/Models/User.php
```diff
@@ -42,7 +42,7 @@
 
 /**
  * @property-read Collection<UserAccountHistory> $accountHistories
- * @property-read ApiKey|null $apiKey
+ * @property-read Collection<ApiKeys>|null $apiKeys
```
notbakaneko, line 45:
```suggestion
 * @property-read Collection<ApiKey> $apiKeys
```

notbakaneko (changes requested):

notbakaneko (approved):
still need a UI for this 👀

nanaya:
These prs are intended to be merged first btw 👀

PR #10086, opened 2023-04-07
Add class based factory for Multiplayer related models

nanaya:
Removed `->getKey()` on the related factory calls while at it.

- [ ] depends on #10084


database/factories/LegacyMatch/GameFactory.php
```diff
@@ -3,29 +3,37 @@
...
+use Database\Factories\Factory;
+
+class GameFactory extends Factory
+{
+    protected $model = Game::class;
+
+    public function definition(): array
+    {
+        return [
+            'beatmap_id' => Beatmap::factory(),
```
notbakaneko, line 22: Come to think of it, the keys should probably be sorted while we're updating these and move the ones with dependencies to the end (same for the others) 🤔

notbakaneko (changes requested):

PR #10087, opened 2023-04-07
Backend part of legacy IRC key management

nanaya:
It's also large but mostly copypasta from #10085. Also it most likely conflicts with the aforementioned PR on the route file.


app/Libraries/OsuAuthorize.php
```diff
@@ -1788,6 +1788,19 @@ public function checkIsSpecialScope(?User $user): string
         return 'unauthorized';
     }
 
+    public function checkLegacyIrcKeyStore(?User $user): string
+    {
+        $this->ensureLoggedIn($user);
+        $this->ensureCleanRecord($user);
+
+        // isBot checks user primary group
+        if (!$user->isGroup(app('groups')->byIdentifier('bot')) && $user->playCount() < 100) {
```
notbakaneko, line 1797: I'm not sure if we want to change this from being primary group only? @peppy

notbakaneko (changes requested):

app/Libraries/OsuAuthorize.php
nanaya: after looking at the requirement apparently you can't even be in bot group before running a bot first so this check probably doesn't matter whether or not it's primary group (the current one in old site doesn't require primary group).

Also having main account being in bot group isn't even a thing anymore anyway? (at least no new one)

notbakaneko: Skipping the playcount check on the old site seems to only apply to users where the primary group is the bot group

nanaya: the query is on `phpbb_user_group` 🤔

notbakaneko (approved):

notbakaneko (approved):

PR #10100, opened 2023-04-10
Add class based factory for Notification

nanaya:
The notifiable doesn't make sense but that's problem for another day 🤔


database/factories/NotificationFactory.php
```diff
@@ -3,14 +3,24 @@
...
-        'details' => [],
-    ];
-});
+class NotificationFactory extends Factory
+{
+    protected $model = Notification::class;
+
+    public function definition(): array
+    {
+        return [
```
notbakaneko, line 19: Could do with key order sort as well

notbakaneko (changes requested):

PR #10101, opened 2023-04-10
Use factory method for generic discussion markdown renderer

notbakaneko:
Also `children` actually resolves to `ReactNode[] & ReactNode` and since `ReactNode` also includes `null | undefined`, it means `children` is `nullable` but the typecheck doesn't pick it up:

```typescript
type Foo = number | null;

function bar(arg: Foo[] & Foo) {
  // doesn't register that arg can be null 
  return arg.map((x) => x);
}
```

The downside is this creates a new instance of the function every time but I guess we could cache it if needed? 🤷


nanaya:
Probably because `&` isn't `|` 🤔 (also it wouldn't make sense to map number)

nanaya (changes requested):

PR #10107, opened 2023-04-12
Convert beatmap discussions user filter component to typescript

notbakaneko:
Like #10088, converting the dependency of `beatmap-discussions/main` to typescript instead of defining a module


resources/js/beatmap-discussions/user-filter.tsx
```diff
@@ -0,0 +1,92 @@
...
+  };
+}
+
+export class UserFilter extends React.Component<Props> {
+  render() {
+    const options = [allUsers, ...this.props.users.map(mapUserProperties)];
+
+    const selected = this.props.selectedUser != null
+      ? mapUserProperties(this.props.selectedUser)
+      : noSelection;
```
nanaya, line 48: getter?

nanaya (changes requested):

nanaya (approved):

PR #10112, opened 2023-04-13
More customisable font size

### LINKED ISSUE #10099: Format text in indicated size on forums
It is very weird that the forum formats text in only 4 different sizes (50, 80, 100, and 150) while the bbcode syntax supports more sizes. For example, something like `[size=90]test[/size]` should reflect the text size indicated, but right now it defaults to size 100.

nanaya:
Resolves #10099.

### DIFF
--- a/app/Libraries/BBCodeFromDB.php
+++ b/app/Libraries/BBCodeFromDB.php
@@ -320,8 +320,12 @@ class BBCodeFromDB
 
     public function parseSize($text)
     {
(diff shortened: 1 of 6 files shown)


notbakaneko (changes requested):
Could also do with extra tests of the out of range sizes 👀

notbakaneko (approved):
I thought I approved this already 👀

PR #10116, opened 2023-04-17
Update shopify api library

nanaya:
There doesn't seem to be any change to the two api calls we're using? Although my dev setup doesn't even work with the older library so depending on how the live store is currently setup there may need some changes over there instead? 🤷‍♂️


notbakaneko (approved):

notbakaneko (commented):
(just needed shopify setting update, no other changes)

PR #10130, opened 2023-04-19
Add frontend for legacy api/irc key

nanaya:
The wordings may need some fixing.

Depends on:
- [x] #10123
- [x] #10117


resources/js/legacy-api-key/controller.ts
```diff
@@ -0,0 +1,83 @@
...
+  showing_form: boolean;
+}
+
+type DatasetState = Partial<State> & Required<Pick<State, 'legacy_api_key'>>;
+
+export default class Controller {
+  @observable state;
+  private stateSyncDisposer;
+  private xhrCreate?: JQuery.jqXHR<LegacyApiKeyJson>;
+  private xhrDelete?: JQuery.jqXHR<void>;
```
notbakaneko, line 20: needs `@observable`

resources/js/legacy-irc-key/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+    this.controller.destroy();
+  }
+
+  render() {
+    return this.controller.key == null ? this.renderEmpty() : this.renderDetails();
+  }
+
+  @action
+  private readonly onNewKeyClick = () => {
+    if (!confirm('legacy_irc_key.confirm_new')) return;
```
notbakaneko, line 38:
```suggestion
    if (!confirm(trans('legacy_irc_key.confirm_new'))) return;
```

notbakaneko (changes requested):
I've just realized this should be `app_url` not `$field` (and the also maybe the `foreach`?) https://github.com/ppy/osu-web/blob/master/app/Models/ApiKey.php#L67

resources/js/legacy-api-key/form.tsx
```diff
@@ -0,0 +1,118 @@
...
+  private errors = new FormErrors();
+
+  constructor(props: Props) {
+    super(props);
+
+    makeObservable(this);
+  }
+
+  @action
+  readonly onCancelClick = () => {
```
notbakaneko, line 31: `readonly`

resources/js/legacy-api-key/details.tsx
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  private readonly deleteClicked = () => {
+    if (!confirm(trans('common.confirmation'))) return;
+
+    this.props.controller.deleteKey();
+  };
+
+  @action
+  private onClickToggleKeyVisibility = () => {
```
notbakaneko, line 115: `readonly`

resources/js/legacy-irc-key/controller.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    makeObservable(this);
+
+    this.stateSyncDisposer = reaction(
+      () => JSON.stringify(this.state),
+      (stateString) => this.container.dataset.state = stateString,
+    );
+  }
+
+  createKey() {
```
notbakaneko, line 42: also needs `@action` to go with the `@observable` to properly show the spinner or disable button, same for the other create/delete keys

notbakaneko (changes requested):

resources/js/legacy-api-key/form.tsx
nanaya: but it's already readonly 🤔

notbakaneko (changes requested):
can also do with `attributes` key in model_validation to localise key name
image

notbakaneko (approved):

PR #10131, opened 2023-04-19
Declare missing return types

notbakaneko:
Match the interface they're implementing.
Also removes the unnecessary(?) `setCreatedAt`


nanaya:
Should the parameters also match?

notbakaneko:
It doesn't complain if `mixed` is missing from the parameters, strangely enough 🤔

nanaya (approved):

PR #10133, opened 2023-04-19
Declare consts for Validatable

notbakaneko:
Apparently, this is a thing since Traits just copy paste the code into the class. It'll satisfy phpstan since phpstan checks the trait from the context of the class it's used in.

(and then stuff was moved into the base class instead)


nanaya:
Maybe create an interface with default const instead? That would also remove the need for `method_exists` check in saveOrExplode function and could also add `isValid` check in the base `save` function so the manual override in each models can be removed.

notbakaneko:
Some classes have variations where `parent::save()` is called separately and having `isValid()` called in the parent doesn't seem to make sense for them; imo, we should also look at making `save()` throw on failing validation instead, considering the number of times `save()` failing as bitten us, especially in tests 🤨

tests/BroadcastNotificationTest.php
```diff
@@ -116,13 +117,14 @@ public function notificationJobClassesDataProvider()
     public function notificationNamesDataProvider()
     {
         // TODO: move notification names to different class instead of filtering
-        $constants = collect((new ReflectionClass(Notification::class))->getConstants())
-            ->except(['NAME_TO_CATEGORY', 'NOTIFIABLE_CLASSES', 'SUBTYPES', 'CREATED_AT', 'UPDATED_AT'])
+        $constants = collect((new ReflectionClass(Notification::class))->getReflectionConstants())
+            ->filter(fn (ReflectionClassConstant $constant) => (
+                $constant->getDeclaringClass()->name === Notification::class
+                    && !in_array($constant->name, ['NAME_TO_CATEGORY', 'NOTIFIABLE_CLASSES', 'SUBTYPES'], true)
+            ))
```
nanaya, line 124: I feel this should be a const array eventually...

(or enum or something)

nanaya (changes requested):

nanaya (approved):

PR #10138, opened 2023-04-20
Use markdown parser to create link in chat

nanaya:
The most basic syntax for now (markdown link). I'm adding the others later (flipped markdown, legacy client, wiki).

### DIFF
--- a/resources/js/remark-plugins/disable-constructs.ts
+++ b/resources/js/remark-plugins/disable-constructs.ts
@@ -11,35 +11,37 @@ interface Options {
   type?: DisabledType;
 }
 
-function makeDisabledListFromAllowlist(allowlist: Construct[]) {
-  return ([
-    // 'characterEscape', // escaping things is always useful
-    // 'content', // not sure what this is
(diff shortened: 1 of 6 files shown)


resources/js/remark-plugins/disable-constructs.ts
```diff
@@ -5,12 +5,40 @@ import type Constructs from 'micromark-core-commonmark';
...
+    'htmlFlow',
+    'htmlText',
+    'labelEnd',
+    'labelStartImage',
+    'labelStartLink',
+    'lineEnding',
+    'list',
+    'setextUnderline',
+    'thematicBreak',
+  ] as Construct[]).filter((item) => !allowlist.includes(item));
```
notbakaneko, line 39: should use a variable typed to `Construct[]` instead of `as Construct[]` as `as Construct[]` allows invalid types to go into the array

notbakaneko (changes requested):

notbakaneko (approved):

PR #10148, opened 2023-04-26
Support legacy link syntax for chat

nanaya:
(the `[https://wat wat]` one)

...seems working? 🤷 

<sub>probably should add test or something</sub>

- [x] depends on #10143 just because it's updating same file


notbakaneko (changes requested):
Could do with some tests particularly for the micromark part 🤔

notbakaneko (approved):
Seems to work...I'm sure someone will come across some kind of breaking edge case 👀

Issue #9942, opened 2023-03-04
Automating updates of mod definitions in `mods.json`


bdach:
Right now, updating [`mods.json`](https://github.com/ppy/osu-web/blob/master/database/mods.json), which is used for server-side validation of submiitted and multiplayer scores, is a semi-manual process using [`osu-tools`](https://github.com/ppy/osu-tools) and its [`mods` command](https://github.com/ppy/osu-tools/blob/master/PerformanceCalculator/Difficulty/ModsCommand.cs).

For a few releases now, the necessity of running this semi-manual process was forgotten about when new mods were added to client, which lead to having to expedite a few deploys as the lazer builds with the new mods had already been live at that point (https://github.com/ppy/osu-web/pull/9841, https://github.com/ppy/osu-web/pull/9940). Automating this could reduce the risk of forgetting about having to run the updates.

I'm imagining that this could work in the following way:

1. When a new release is created over at `osu`, a workflow is run at `osu`, that signals `osu-web` about the new release using the [`repository_dispatch`](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#repository_dispatch) event (requires issuance of a [personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token))
2. `osu-web` runs a workflow on `repository_dispatch` which:
   - clones the release tag of `osu`
   - clones the latest `master` of `osu-tools`
   - generates the `mods.json` file using the `mods` command
   - if the files are identical, the workflow early-aborts with no further action
   - if the files are not identical, a PR is opened, containing the necessary update to `mods.json` (and maybe a few assignees/required reviewers to bring attention to it?)

PR #10156, opened 2023-05-03
Add workflow for semi-automated mod definition updates

bdach:
This PR is supposed to be the starting point for https://github.com/ppy/osu-web/issues/9942. The goal here is to be able to automate the generation of `database/mods.json`, so that once new client releases drop, the process of updating the mod definitions is made much easier on the web side.

Right now, the workflow is written to be dispatched manually. What this means, is that to start it, you have to go to the "Actions" tab, choose the "Update mod definitions workflow", and fill the argument:

This part is probably not what we want, and is why I'm opening this as draft, to try and gauge what we *would* want. One idea in the issue was to coordinate the workflow such that `ppy/osu` signals `ppy/osu-web` on every new release by way of a `repository_dispatch` event, but this would require an extra workflow on the `ppy/osu` side and possibly issuance of a new personal access token. Notably, `ppy/osu` can't create the PR on the web side itself.

This was tested to work on my fork. See the following runs:

- [Run 1](https://github.com/bdach/osu-web/actions/runs/4870886979/jobs/8687174369): Wherein I specified the latest release tag when dispatching the workflow; this meant that the mod definitions were up to date, so no PR was opened (note that the "Create pull request with changes" step did not execute).
- [Run 2](https://github.com/bdach/osu-web/actions/runs/4870905479): Wherein I specified nothing, which ends up being latest `master`, and so, since on the game side we've merged a new mod since the last release, this resulted in [this PR](https://github.com/bdach/osu-web/pull/2) being opened.

@ThePooN cc'ing you here, since @peppy said in stream chat that you may have an opinion as to where and when this thing should run.

### DIFF
--- a/.github/workflows/update-mod-definitions.yml
+++ b/.github/workflows/update-mod-definitions.yml
@@ -1,0 +1,68 @@
+name: Update mod definitions
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        type: string
+        description: The name of the ref to check out. Will default to latest `master`.
+
+jobs:
+  update-mod-definitions:
+    name: Update mod definitions
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout ppy/osu-web
+      uses: actions/checkout@v3
+      with:
+        path: osu-web
+
+    - name: Install .NET 6.0.x
+      uses: actions/setup-dotnet@v3
+      with:
+        dotnet-version: "6.0.x"
+
+    - name: Checkout ppy/osu
+      uses: actions/checkout@v3
+      with:
+        repository: ppy/osu
+        ref: ${{ github.event.inputs.ref }}
+        path: osu
+
+    - name: Store ppy/osu commit hash
+      id: store-osu-hash
+      run: echo commit-hash=`git rev-parse HEAD` >> "$GITHUB_OUTPUT"
+      working-directory: ./osu
+
+    - name: Checkout ppy/osu-tools
+      uses: actions/checkout@v3
+      with:
+        repository: ppy/osu-tools
+        path: osu-tools
+
+    - name: Setup local game checkout for tools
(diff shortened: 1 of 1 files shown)


ThePooN:
I think this is fine, but as you pointed out, it should rather be ran on every commit to `master` on `osu-tools` or `osu`, otherwise this ain't solving much.

We already use the `repository_dispatch` flow heavily in this repo, but the other way around; osu-web dispatches events to the private `osu-kubernetes-config` repo for deployments, see [osu-web/.github/workflows/pack.yml](https://github.com/ppy/osu-web/blob/master/.github/workflows/pack.yml).  

Frankly, the fact that it relies on a workflow in two repos and especially that it requires PATs is very not ideal, I wish GitHub would provide a better solution. I know I'm going to migrate everything the day they do, but until then this is the way, and you can go ahead implementing it this way in the `osu` or `osu-tools` repo.

bdach:
> Actually, using a PAT you can also do everything from a single workflow in the osu/osu-tools repo and completely discard the idea of running a workflow in osu-web.

Appears that way. I initially thought the action I was using to open the PR had no power to do this [but it apparently does](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#creating-pull-requests-in-a-remote-repository). Kind of weird that's not in the main README.

I could move this to `osu-tools`, but I would still need `repository_dispatch`to make work, right? `osu-tools` does not really see much regular traffic, so running every commit there doesn't help much, unless I made it a cronjob or something. If I want to not muck around with any of that and we're going to be using PATs anyway, then the best place would be `osu` itself, then I could just set the trigger to `release` and that would fix it all, no acrobatics required.

bdach:
Makes sense. There isn't much remaining to do here then; I'll move this across to the game repo.

PR #10160, opened 2023-05-09
Fix new beatmap discussion pinning when switching tabs

notbakaneko:
- Fix dying when jumping to a review from non-review discussion while the new-discussion element is pinned.
- Fix the pinned position of element when switching discussion tabs.


nanaya (approved):
is it just me or they both have same wrapper container; structure, logic and all 🤨

notbakaneko:
I'm pretty sure `new-review` was originally a copy-paste of `new-discussion` 👀

PR #10161, opened 2023-05-09
Fix current user data being updated too late on navigation

### LINKED ISSUE #10149: Error occurs when logging in to friends page while logged out
When logging in from https://osu.ppy.sh/home/friends while logged out, an error occurs and the friend list is not displayed. However, refreshing the page resolves the issue.

If the site language setting when logged out and the site language setting when logging in (language setting associated with the account) are different, the site language will be automatically refreshed to change the language, so the error will be resolved without any action by the user.

PS: not using user script.

Please refer to the attached video for reproduction (the password field is intentionally hidden).

https://user-images.githubusercontent.com/65931718/234664974-b9fd2c7b-9e05-4d84-82c5-694cb2d83900.mp4

```
mobx.esm.js:1139  Uncaught Error: current user is null
    at get currentUserOrFail (osu-core.ts:71:13)
    at Ke (mobx.esm.js:1793:18)
    at t.computeValue_ (mobx.esm.js:1567:13)
    at t.trackAndCompute (mobx.esm.js:1538:25)
    at t.get (mobx.esm.js:1509:18)
    at t.getObservablePropValue_ (mobx.esm.js:4763:34)
    at R.get (mobx.esm.js:5308:26)
    at <anonymous>:1:718
    at react-turbolinks.ts:42:27
    at Ie (mobx.esm.js:1136:15)
```

nanaya:
On navigation after login, react components will be rendered before currentUser is updated which may cause error/invalid assumption. `newBody` is empty on very first page load so the json parser need to deal with that accordingly.

Resolves #10149.


resources/js/osu-core.ts
```diff
@@ -95,14 +119,20 @@ export default class OsuCore {
...
-    this.setCurrentUser(user);
+  readonly updateCurrentUser = () => {
+    // Remove from DOM so only new data is parsed on navigation.
+    window.currentUser = parseJsonNullable('json-current-user', true) ?? { id: undefined };
+    this.setCurrentUser(window.currentUser);
   };
 
-  private onPageLoad = () => {
-    this.setCurrentUser(window.currentUser);
+  private onCurrentUserUpdate = (event: unknown, user: CurrentUserJson) => {
```
notbakaneko, line 134: can `readonly` this while we're at it, same for `setCurrentUser`

notbakaneko (changes requested):

notbakaneko (approved):

PR #10164, opened 2023-05-10
Support old link syntax in chat

### LINKED ISSUE #8711: Closed bracket in link format get carried over in web chat
The close bracket get treated as part of URL if the user or BanchoBot use `(Text)[Link]` format to create URL. For example, if the link in the chat formatted like this `(Mamat)[https://osu.ppy.sh/u/3031177]` it will show this in web chat:

(Mamat)\[https://osu.ppy.sh/u/3031177]

BanchoBot Example

Thus, if user click it will include close bracket in URL bar.

nanaya:
Lifted mostly from the other link syntax 👀 

Resolves #8711.


notbakaneko (changes requested):
something like `()[https://psu.ppy.sh]` causes a 0 length chunk with items causing it to go into an infinite loop

notbakaneko (approved):
seems..to work 👀

PR #10165, opened 2023-05-10
Disallow javascript protocol in chat links

nanaya:
React is already blocking it but throws warning (on dev).


notbakaneko (changes requested):
this should already be disallowed by the default `uriTransformer` in `ReactMarkdown`? https://github.com/remarkjs/react-markdown#uritransformer

nanaya:
So it's the one generating that `void(0)` conversion 🤔 React still complains about it though (because it's still javascript link)

notbakaneko (approved):
would be useful if this `transformLinkUri` allowed the callback to return `null` or `undefined` to remove the link :|

PR #10147, opened 2023-04-26
Cleanup beatmap discussions Nominator component

notbakaneko:
noticed while cleaning up the conversion of `Nominations` component to typescript that `nominator`'s `props` have the wrong declaration.

Also added extra tooltips when the user can't nominate

PR #10166, opened 2023-05-10
Convert beatmap discussions nominations component to typescript 

notbakaneko:
- [ ] #10147


nanaya:
The title says nominations but the header is also converted 🤔

resources/js/beatmap-discussions/nominations.tsx
```diff
@@ -0,0 +1,668 @@
...
+  private readonly delete = () => {
+    const message = this.userIsOwner
+      ? trans('beatmaps.nominations.delete_own_confirm')
+      : trans('beatmaps.nominations.delete_other_confirm');
+
+    if (!confirm(message)) return;
+
+    showLoadingOverlay();
+
+    this.xhr.delete?.abort();
```
nanaya, line 149: all the xhr checks need to be switched to return if processing instead

notbakaneko:
https://github.com/ppy/osu-web/pull/10176 instead

PR #10172, opened 2023-05-11
Support multiple oauth callback url input

### LINKED ISSUE #10137: Allow multiple redirect URIs for an OAuth 2 application
Currently we're developing an application that is separated in a couple of different microservices. 

These all have authentication implemented using osu! in a different way due to structural differences, but we ran into a roadblock where only one Redirect URI is allowed in an application.

Maintaining multiple IDs and secrets in the environment is inconvenient but may also be confusing for users of the applications, considering they're part of a larger stack.  

Example OAuth 2 provider (discord) that does allows the same feature: 
CleanShot 2023-04-17 at 14 46 57@2x

Having this implemented would certainly improve the developer experience for us. 

- co-signed by @omkelderman

nanaya:
The model setter (both php and javascript) kind of sucks 😦 

Resolves #10137 (the actual interface instead of just being implicitly supported)

### DIFF
--- a/app/Models/OAuth/Client.php
+++ b/app/Models/OAuth/Client.php
@@ -9,6 +9,7 @@ use App\Exceptions\InvariantException;
 use App\Models\User;
 use App\Traits\Validatable;
 use DB;
+use Ds\Set;
 use Laravel\Passport\Client as PassportClient;
 use Laravel\Passport\RefreshToken;
 
@@ -56,6 +57,26 @@ class Client extends PassportClient
         );
     }
 
+    public function setRedirectAttribute($value)
+    {
+        if (is_string($value) && present($value)) {
+            $baseString = preg_replace("/[\r\n]+/", ',', $value);
+            $baseEntries = explode(',', $baseString);
+            $entries = new Set();
+            foreach ($baseEntries as $entry) {
+                $entry = presence(trim($entry));
+                if ($entry !== null) {
+                    $entries->add($entry);
+                }
+            }
+            if (!$entries->isEmpty()) {
+                $cleanValue = $entries->join(',');
+            }
+        }
+
+        $this->attributes['redirect'] = $cleanValue ?? null;
+    }
+
     public function isValid()
     {
         $this->validationErrors()->reset();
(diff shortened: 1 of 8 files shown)


app/Models/OAuth/Client.php
```diff
@@ -56,6 +57,26 @@ public function refreshTokens()
         );
     }
 
+    public function setRedirectAttribute($value)
+    {
+        if (is_string($value) && present($value)) {
+            $baseString = preg_replace("/[\r\n]+/", ',', $value);
+            $baseEntries = explode(',', $baseString);
```
notbakaneko, line 64: is this `preg_split` 🤔

resources/lang/en/model_validation.php
```diff
@@ -122,7 +122,7 @@
     'oauth' => [
         'client' => [
             'too_many' => 'Exceeded maximum number of allowed OAuth applications.',
-            'url' => 'Please enter a valid URL.',
+            'url' => 'Please enter valid URL.',
```
notbakaneko, line 125: wonder if this should be `URLs` now even though only one is required

resources/js/oauth/new-client.tsx
```diff
@@ -42,19 +44,40 @@ export class NewClient extends React.Component {
...
+              type='text'
+              value={this.params.name}
+            />
+          </label>
+
+          <label className='oauth-client-details__group'>
+            <div className='oauth-client-details__label'>
+              {trans('oauth.client.redirect')}
+            </div>
+            <TextareaAutosize
```
notbakaneko, line 65: Might be useful to initialize the textarea height to at least the number of callbacks within reason; that said, I feel like there should be a limit on the number of callbacks that can be set on a single application 🤔

notbakaneko (changes requested):

resources/js/oauth/new-client.tsx
nanaya: the min height is already set in the css though?

notbakaneko: I mean if there's 5 callbacks it should start at 5 rows visible, instead of the current one where it only gets larger after adding more callbacks

notbakaneko (approved):

PR #10180, opened 2023-05-12
Fix incorrect user state on navigation

### LINKED ISSUE #10179: Going back in browser history renders user as not signed in
started happening recently

using Chrome Version 113.0.5672.92 (Official Build) (64-bit)

https://github.com/ppy/osu-web/assets/62819481/65f28216-38fa-46e3-b1f9-15a1c41796ec

nanaya:
Only update current user state if the json exists otherwise going back will reset to guest as the json is removed on initial read.

Resolves #10179.


notbakaneko (approved):
looks exactly like the one I was going to push 🤔

PR #10201, opened 2023-05-16
Stop post edit counter after 64k

### LINKED ISSUE #10196: Getting internal server error when trying to edit a post
I have been running a forum game since 2018 called "Thread Necromancy" ([link here](https://osu.ppy.sh/community/forums/topics/802725)). The objective is to get points with respect to how much time passes since the last post. The first post displays the leaderboard as well as historical winners, and it is updated upon every new post in the thread. 

Fast forward to today - I have started getting "HTTP 500 Internal Server Error" upon trying to edit the post. That one post has accumulated over 65,535 edits and I suspect the data structure used to store that value is 16 bit, leading to the failure.

nanaya:
Resolves #10196. The second most edited post is at just a bit over 10k so there's not much point increasing the column size.

### DIFF
--- a/app/Models/Forum/Post.php
+++ b/app/Models/Forum/Post.php
@@ -377,10 +377,10 @@ class Post extends Model implements AfterCommit, Indexable, Traits\ReportableInt
 
         // record edit history
         if ($this->exists && $this->isDirty('post_text')) {
-            $this->fill([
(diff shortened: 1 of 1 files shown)


notbakaneko:
At that point there is little difference between 64k and 65k

notbakaneko (approved):

PR #10203, opened 2023-05-16
Add extra info on chat reports

nanaya:
<del>Currently it'll add context messages from either public channels or the specific channel (pm, etc) the message comes from. Channel name is in the original form though - e.g. `#pm_$userId1-$userId2` for pm.</del> Changed to always show non-pm messages.

Also fixed some weird params handling.


app/Models/Chat/Message.php
```diff
@@ -66,29 +66,24 @@ public function getAttribute($key)
...
 
-    public function reportableAdditionalInfo(?User $reporter): ?string
+    public function reportableAdditionalInfo(): ?string
     {
-        if (
-            !priv_check_user($reporter, 'ChatChannelRead', $this->channel)->can()
-            && !priv_check_user($reporter, 'ChatChannelJoin', $this->channel)->can()
-        ) {
-            return "reporter doesn't actually have access to the reported message";
-        }
```
notbakaneko: why was this removed?

notbakaneko (changes requested):

app/Models/Chat/Message.php
nanaya: the included backlog now only shows non-pm messages. And if we're to add this check it should probably be during report creation instead anyway and that's kind of unrelated to this PR.

notbakaneko (approved):

PR #10210, opened 2023-05-18
Simplify Comments Index component

nanaya:
Move header and pagination rendering to server side. Mainly to remove the stupid handling of pagination component.


nanaya:
Eh forget this I'm rolling this up to the ts conversion pr.

PR #10211, opened 2023-05-19
Dim forum topic entry if its older than 1 month

### LINKED ISSUE #4017: Make forum thread titles grey when they're getting old
my wish for new site:
saapcfg

atm:
l8vkzvi

would give some better overview i guess!

RockRoller01:
yay, first real contribution

Colors are placeholder, wasn't quite sure what the colours should be.

Also tried using opacity just like --deleted does, however this didn't feel quite right and wasn't clear enough when a user can see all 3 states:
image

closes #4017


RockRoller01:
Honestly not quite sure how necessary this is, I feel like with the new forums this doesn't do much, might be better to just close the original issue and this PR

Hiviexd:
honestly i feel like dimming inactive posts would cause confusion among moderators because deleted posts are also dimmed.

i'd suggest following this approach from the [forum figma design library](https://www.figma.com/file/Gz0MIrXMVVdwaKeecOcLlavu/Web%2FForum?type=design&node-id=0-1&t=Qxwa4fpSwOAIbDWe-0) instead:

image

RockRoller01:
Changed it to be in-line with the figma design, not sure if the level 4 base colours should also be added for the other colours as nothing uses them yet..

resources/css/bem/forum-topic-entry.less
```diff
@@ -4,27 +4,37 @@
...
   }
 
   @media @desktop {
     margin-left: -@forum-item-overflow-desktop;
     margin-right: -@forum-item-overflow-desktop;
     padding-left: @forum-item-overflow-desktop;
   }
+  
+  &--old {
+    opacity: 0.5;
```
nanaya, line 31: I feel this should be at least 0.7 otherwise it's kinda hard to read

nanaya (changes requested):

resources/css/bem/forum-topic-entry.less
```diff
@@ -4,27 +4,37 @@
...
     margin-right: -@forum-item-overflow-desktop;
     padding-left: @forum-item-overflow-desktop;
   }
+  
+  &--old {
+    opacity: 0.7;
+  }
 
   &--deleted {
+    --forum-item-background-color: hsl(var(--colour-red-hue), 40%, 30%);
```
nanaya, line 35: I only meant the `@osu-colour-x` lesscss variable being deprecated, not the whole `--hsl-x`...

resources/views/forum/forums/_topic.blade.php
```diff
@@ -7,8 +7,7 @@
 @endphp
 <li
     class="
-        forum-topic-entry
-        {{ $topic->trashed() ? 'forum-topic-entry--deleted' : '' }}
+        {{ class_with_modifiers('forum-topic-entry', $topic->trashed() ? 'deleted' : null, $topic->isOld() ? 'old' : null) }}
```
nanaya, line 10:
```php
['deleted' => $topic->trashed(), 'old' => $topic->isOld()]
```

app/Models/Forum/Topic.php
```diff
@@ -482,6 +482,15 @@ public function deletedPostsCount()
         });
     }
 
+    public function isOld()
+    {
+        // pinned and announce posts should never be considered old
+        if ($this->topic_type !== 0) {
```
nanaya, line 488:
```php
static::TYPES['normal']
```

nanaya (dismissed):

peppy:
@nanaya may be worth applying the outstanding review pieces and getting this in if there's nothing else blocking it.

I still think having inactive threads dimmed is beneficial.

notbakaneko (approved):

PR #10212, opened 2023-05-19
Add tooltips to discussion post buttons

### LINKED ISSUE #9588: Add a tooltip to discussion post buttons
as someone who rarely uses the discussion its very unintiuitve that these posts are "post" buttons, I always think they simply select the post type.

grafik

RockRoller01:
someone please double check that all the titles make sense, not too familiar with the discussion system

closes #9588


resources/lang/en/beatmaps.php
```diff
@@ -67,15 +67,42 @@
         ],
 
         'message_type' => [
-            'disqualify' => 'Disqualify',
```
nanaya, line 70: this should be done in a way that doesn't remove all existing translations. Add a suffix, use different group, etc.

nanaya (changes requested):

resources/lang/en/beatmaps.php
RockRoller01: resolved by using a different group

resources/js/beatmap-discussions/new-discussion.tsx
```diff
@@ -455,6 +455,7 @@ export class NewDiscussion extends React.Component<Props> {
         props={{
           'data-type': type,
           onClick: this.post,
+          title: trans(`beatmaps.discussions.message_type_title.${typeText}`),
```
nanaya, line 458: title tooltip behaviour is rather bugged in chrome when the button is disabled. Wrap it with a div or something and put the title there (it's still kind of buggy but at least it's less bad).

nanaya (changes requested):

peppy:
Another one which can probably be review-applied from our end just to clean up open PRs.

nanaya (approved):

PR #10213, opened 2023-05-19
Fix missing gap after comments on changelog page

### LINKED ISSUE #9593: Missing gap after comments section on changelog
Just a minor visual thing, but afaik there should be a gap after the comments container, however its quite clearly filled on changelog comments.

how it should be:
image

how it is on changelogs:
image

RockRoller01:
closes #9593 

tested for both mobile and desktop

### DIFF
--- a/resources/css/bem/comments.less
+++ b/resources/css/bem/comments.less
@@ -32,6 +32,10 @@
     }
   }
 
+  &--changelog{
+    margin-bottom: 0;
+  }
+
   &__count {
     font-size: 80%;
     padding: 5px 10px;
--- a/resources/css/bem/osu-page.less
+++ b/resources/css/bem/osu-page.less
@@ -101,7 +101,6 @@
     background-color: @osu-colour-b4;
     display: flex;
     flex-direction: column;
-    margin-bottom: 0;
   }
 
   &--comment {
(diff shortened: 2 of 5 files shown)


notbakaneko (commented):
https://github.com/ppy/osu-web/issues/9593#issuecomment-1559349453

RockRoller01:
@notbakaneko addressed all the ones in the comment and found some others

notbakaneko (changes requested):
After investigating the layout some more, I think it makes more sense to move the `margin-bottom` to the base `osu-page` itself and have a modifier with `margin-bottom: 0` to override as necessary - there's also a bunch of inner elements that have margin-bottom which aren'y necessary if it's on the container instead

RockRoller01:
I thought the very same thing, but I am not quite sure which of the classes are for components and which for actual layouts, so I didn't go with that route, as I wouldn't know which pages to check

notbakaneko:
I'll probably make the updates in a separate PR

RockRoller01:
@notbakaneko so is there anything blocking this PR considering that you wanted to the other fix in a seperate PR?

notbakaneko:
@RockRoller01 I mean I'm going to move the margin to the base style and remove them from the inner elements, instead

notbakaneko:
#10435

PR #10217, opened 2023-05-19
Remove redundant button disabling parameter

nanaya:
The `disabled` are all either redundant with `isBusy` or include larger test which covers the same thing. This removes the need to copy the check for the former.


notbakaneko (dismissed):
Could do with a jsdoc `/** */` on the prop itself saying `isBusy` implies `disabled` (`//` doesn't seem to work for vscode 🤔 )

notbakaneko:
also apparently I pressed the wrong button 👀

notbakaneko (approved):

PR #10219, opened 2023-05-19
Typescript Comments Index component

nanaya:
This includes moving a bunch of components back to blade as per previous PR.


resources/views/comments/index.blade.php
```diff
@@ -4,18 +4,43 @@
...
+        'theme' => 'comments',
+    ]])
+    <div class="osu-page osu-page--comments">
+        <div class="js-react--comments-index u-contents"></div>
 
-    {{-- temporary pagination to be used by react component above --}}
-    <div class="hidden">
-        <div class="js-comments-pagination">
-            @include('objects._pagination_v2', ['object' => $commentPagination])
-        </div>
```
notbakaneko, line 14: I think this was more not to have a pager with no body before the component finishes mounting 🤔

nanaya: it was because it's located inside the react component's `osu-page` (before this change)

PR #10221, opened 2023-05-19
Resolve OOM error when generating covers from big mapsets

ThePooN:
`RegenerateBeatmapsetCover` jobs on large sets occasionally fail in OOM. See: https://sentry.ppy.sh/organizations/ppy/issues/20008/?alert_rule_id=5&alert_timestamp=1684532190493&alert_type=email&environment=production-k8s&project=3

It appears the code is designed to use temporary files to optimize memory usage, but wastes all these efforts on download. Quickly fixed that using cURL.

Tested on my end using a bare PHP file with just the cURL instructions, further validation before merge is welcome.


app/Models/Beatmapset.php
```diff
@@ -469,17 +469,21 @@ public function fetchBeatmapsetArchive()
             return false;
         }
 
-        $contents = file_get_contents($url);
-        if ($contents === false) {
-            throw new BeatmapProcessorException('Error retrieving beatmap');
-        }
+        $curl = curl_init($url);
```
nanaya, line 472: does this need to handle redirects? (301/302)

nanaya (changes requested):

app/Models/Beatmapset.php
ThePooN: I don't believe so. But it's just a curl option that can't break anything, yeah?

nanaya (approved):

PR #10222, opened 2023-05-21
Add events listing API endpoint

cl8n:
unsure if any events need to be filtered out from this (is `private` used for anything..?)


peppy:
Would probably be good to get this in, as there have been cases of users scraping the page on the old site and maybe having a valid endpoint would improve that situation.

nanaya (approved):

PR #10225, opened 2023-05-22
Add datadog metric for notification digest mails sent

### LINKED ISSUE #10194: Add datadog metrics for sending mail notifications
Side-effect of #10183 is that there are no stats to monitor now

notbakaneko:
closes #10194


nanaya (approved):
A datadog prefix helper function seems like could be useful 🤔

PR #10228, opened 2023-05-22
Require being in a channel to report chat message

notbakaneko:
Seems weird to be able to report messages without being in a channel 👀 
(also apparently there's legit use-case for allowing reports without being in a channel)


nanaya:
should it be part of `checkUserReport` instead? There's already check for trashed item there.

notbakaneko:
Not sure where it should actually go, or if should just return a `404`

notbakaneko:
Figured it'd lake more sense to have the limitation only apply to pm-like channels 🤔

nanaya (approved):

PR #10247, opened 2023-05-25
Upgrade to Laravel 10

### LINKED ISSUE #10042: Upgrade Laravel
We are running two major versions behind with the release of Laravel 10. Outdated practices such as the use of Laravel Mix have been dropped and replaced with Vite, Artisan has gotten a makeover, amongst many other things. (I recommend reading the update logs)

While I wouldn't mind picking up this task by myself, I'd like to hear anything from maintainers to gather if they think it is worth the time spent.

nanaya:
- reset middleware order (to match existing behaviour; should probably readd it later)
- <del>reset connections when dealing with database outside normal test cycle</del> already done in a separate PR

Seems working but probably can use a bit more tests 🤔 

Resolves #10042.


notbakaneko (approved):
working so far 👀

PR #10252, opened 2023-05-26
Remove unused osu-page styles

RockRoller01:
I noticed while working on #10213 that a lot of osu-page styles appear to be not used. No matter how hard I searched I could not find them, nor find anything that breaks after removing them, so probably fine? A lot of the modifiers clearly indicate their past usage and those pages now all used one of the many generic versions of osu-page


resources/css/bem/osu-page.less
```diff
@@ -244,16 +209,6 @@
     }
   }
 
-  &--ranking-info {
```
notbakaneko, line 247: Used by seasons ranking

notbakaneko (changes requested):

notbakaneko (approved):

PR #10253, opened 2023-05-27
Add tiered level badge colouring

RockRoller01:
Couldn't find any details in figma, it only had black and a fancy gradient, so I followed what was implemented in lazer: https://github.com/ppy/osu/pull/22277

This probably is not the best way to implement this, but I couldn't find any other. Very open to any suggestions


resources/css/bem/user-level.less
```diff
@@ -6,7 +6,65 @@
   width: 50px;
   height: $width;
   font-size: @font-size--large;
-  background-image: url('~@images/backgrounds/level-hexagon.svg');
```
nanaya, line 9: put back the original svg except modify it

```svg
<svg xmlns="http://www.w3.org/2000/svg">
<defs>
  <clipPath id="clip" clipPathUnits="objectBoundingBox" transform="translate(3.35) scale(0.01)">
    <original path goes here/>
  </clipPath>
</defs>
</svg>
```

And the icon element

```css
&__icon {
  .full-size();
  clip-path: url('~@images/backgrounds/level-hexagon.svg#clip');
  background: linear-gradient(var(--bg));
}
```

Finally the base block modifiers (don't bother saving off typing `--tier`)

```scss
&--tier-<name> {
  --bg: @level-tier-<name>;
}
// putting the original variables in layout.less as css variables
// and using them inlined directly could also work
```

The icon element can be just an empty div.

nanaya (changes requested):

resources/css/bem/user-level.less
RockRoller01: These changes should be applied now

nanaya (changes requested):

peppy:
@RockRoller01 are you planning to apply the review changes (or @nanaya just go for it if those are the only blocking things i guess 🤷)

RockRoller01:
@peppy Yes I plan to apply them, just haven't had much time for this stuff since the skinning contest was wrapping up. Should be getting back to this and my other PRs soon

notbakaneko (approved):

PR #10255, opened 2023-05-29
Add more chat scopes

### LINKED ISSUE #10118: Open the WebSocket chat API to 3rd-party apps
Currently, the WebSocket API requires scopes that aren't available to 3rd-party apps (https://github.com/ppy/osu-notification-server/blob/693fab53e75d4e5d227426b15985228ade6f02f0/src/oauth-verifier.ts#L92); `*` can only be granted through the lazer-specific password grant, and `read` doesn't seem to exist.

This is necessary in order to actually switch from the legacy IRC API to osu-web APIs, as has been encouraged for years 😅

The API endpoint that returns the WebSocket gateway URL is also lazer-only.

Note that this cannot be restricted to whitelisted chat bots only; building a chat bot on a normal user account first is necessary per [current process](https://osu.ppy.sh/wiki/en/Bot_account) to be eligible.

Opened this issue here rather than osu-notification-server since OAuth scope stuff probably makes more sense to handle/track here, but feel free to move this issue across.

Received requests about this from @Tillerino and @RoanH.

nanaya:
Something like this maybe? 🤷 I don't remember what casing the scopes are supposed to be in. I'm adding another one for joining/leaving channels because otherwise it may unintentionally extend permission of existing tokens.

Also moved the `chat.write` scope from `Channels@store` (joining channel) to `Messages@store` (sending message to channel) as it looks like a bug.

depends on:
- [x] #10256
- [x] ppy/osu-notification-server#50
- [x] deploy updated notification server?

Resolves #10118.


notbakaneko (dismissed):
Scopes following the same casing as json with dot notation seems fine 🤔 

Waiting confirmation on https://github.com/ppy/osu-notification-server/pull/50#pullrequestreview-1449413679

nanaya:
The notification server has been updated and deployed.

notbakaneko (approved):

PR #10276, opened 2023-06-11
fix: Invalid endpoint for v1 in docs

minisbett:
The base url for the v1 api is https://osu.ppy.sh/api/, not https://osu.ppy.sh/api/v1.

I'm not sure what would be the best way to fix this in the docs, tried my best.

### DIFF
--- a/resources/views/docs/info.md.blade.php
+++ b/resources/views/docs/info.md.blade.php
@@ -106,8 +106,8 @@ This is combined with the base endpoint to determine where requests should be se
 
 URL                             | Status
 ------------------------------- | ---------------------------------------------------------------
-https://osu.ppy.sh/api/v2/      | current
-https://osu.ppy.sh/api/         | _legacy api provided by the old site, will be deprecated soon_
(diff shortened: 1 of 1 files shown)


peppy:
A bit confused, You're adding v1 specific documentation to a site which clearly says it's for v2. v1 documentation is found over at https://github.com/ppy/osu-api/wiki

minisbett:
> A bit confused, You're adding v1 specific documentation to a site which clearly says it's for v2. v1 documentation is found over at https://github.com/ppy/osu-api/wiki

That's a fair point but I guess the endpoint is still incorrect so why not fix it? Or if as you said the V2 docs aren't the place for that remove it completely.

peppy:
Probably best to remove it. Also maybe the whole version section should be replaced with a link to the changelog (https://osu.ppy.sh/docs/index.html#changelog)

notbakaneko (changes requested):
Probably should remove `v1` from the docs since we're no longer planning on implementing it on `osu-web`

PR #10283, opened 2023-06-16
Add timeout for curl

nanaya:
Apparently the default connect timeout is 5 minutes which is a bit too long.


notbakaneko (approved):
hope 10s total time doesn't end up being too short 👀

PR #10287, opened 2023-06-20
Set event parsing failure as tag for easier filtering

nanaya:
I can't find way to filter/group by additional data 🤔


notbakaneko (approved):
> I can't find way to filter/group by additional data 🤔

because there isn't 👀

PR #10288, opened 2023-06-20
Stick profile page show more buttons at the bottom

### LINKED ISSUE #9520: Bloated profile due to scrolling feature
Hello, 

I see you guys added the "Nominated Ranked Beatmaps" and it's a really cool feature, although the scrolling through that is really annoying considering you can have a high number... (example : Lasse with +700 nomination, good luck to find the first map he nominated)

Maybe the button show more makes the section with a scrolling bar? So it's easier to go to the first entry in the list. 

Anyways, keep up the good work!

Edit: Btw this applies for mostly all bloated categories, even ranked, graveyard maps is a pain to scroll through...

nanaya:
Kind of resolves #9520. Nested scrolling is horrible so this is the next best thing available (cue people smashing the button). The end result is hilarious enough so here's the PR (and the original issue has `priority:1` for some reason).

### DIFF
--- a/resources/css/bem/show-more-link.less
+++ b/resources/css/bem/show-more-link.less
@@ -70,6 +70,8 @@
   }
 
   &--profile-page {
+    position: sticky;
+    bottom: 5px;
     margin: 5px auto 0;
   }
 
--- a/resources/css/utilities.less
+++ b/resources/css/utilities.less
@@ -25,7 +25,7 @@
 }
 
 .u-contents {
-  display: contents;
+  display: contents !important;
 }
 
 .u-ellipsis-overflow {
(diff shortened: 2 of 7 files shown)


notbakaneko (commented):
This feels kind of confusing while scrolling down with the `show more`s appearing and disappearing when there's multiple sections :|

notbakaneko:
also this looks kind of weird 🤔 
image

nanaya:
I didn't say it looks good 👀 (although it can probably at least use a drop shadow or something)

notbakaneko (approved):
Maybe being able to jump back to the top of a section or collapsing sections is better?
Anyway, let's find out how well this works 🤷

nanaya:
I wonder if it should be sticky header with buttons (jump to top, to bottom, show more) 🤔 

Or add jump buttons to the sides of show more.

PR #10296, opened 2023-06-22
Add link to gift supporter tag from user card

### LINKED ISSUE #1587: Gift supporter tag from user card
Add link in user card to gift them supporter tag.

nanaya:
Resolves #1587.


peppy (approved):

notbakaneko (approved):
wonder if the user profile page should get a button too?

nanaya:
but it is added there as well 👀 I just forgot to mention it.

PR #10305, opened 2023-06-23
Include channel message length limit data

### LINKED ISSUE #9590: Expose the channel's max text length
A follow-up issue from https://github.com/ppy/osu/issues/21330, in order to fix this, we need to expose the channel's max text length to have different max lengths for different chat text boxes as said by @bdach https://github.com/ppy/osu/pull/21397#issuecomment-1325522484.

https://github.com/ppy/osu-web/blob/7d89312883198ebb8ba57d6afc5319db9fcc05b2/config/osu.php#L80

nanaya:
There's one problem with this approach that is the length for new channel isn't known until it's saved 🤔 

Resolves #9590.

### DIFF
--- a/app/Models/Chat/Channel.php
+++ b/app/Models/Chat/Channel.php
@@ -402,6 +402,13 @@ class Channel extends Model
         return $this->validationErrors()->isEmpty();
     }
 
+    public function messageLengthLimit(): int
+    {
+        return $this->isAnnouncement()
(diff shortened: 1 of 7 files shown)


notbakaneko:
> There's one problem with this approach that is the length for new channel isn't known until it's saved 🤔

I feel like we should have a config endpoint or something that returns the current values of configurable items like this (and review max blocks, etc)

notbakaneko (approved):

PR #10311, opened 2023-06-26
Convert modding profile component to typescript

notbakaneko:
now using a copy of the `pageScan/Jump` logic from the main profile page (also there's an existing problem of lazy loaded images shifting the scroll position)


resources/js/modding-profile/main.tsx
```diff
@@ -0,0 +1,330 @@
...
+                        <Badges badges={this.props.user.badges} />
+                        <Stats user={this.props.user} />
+                      </div>
+                    </>
+                  )}
+                  <DetailBar user={this.props.user} />
+                </div>
+                <div
+                  ref={this.pagesOffsetRef}
+                  className='hidden-xs page-extra-tabs page-extra-tabs--profile-page'
```
nanaya, line 194: the hidden-xs shouldn't be here anymore

resources/js/modding-profile/main.tsx
```diff
@@ -0,0 +1,330 @@
...
+        ? pageBottom > 0
+        : pageBottom > this.stickyHeaderOffset && pageDims.top < window.innerHeight;
+
+      if (match) {
+        matching.add(pageId);
+      }
+    }
+
+    let preferred: Page | undefined;
+    const pageIds = [...matching.values()];
```
nanaya, line 294: `.values()` not needed?

resources/js/beatmap-discussions/review-editor-config-context.ts
```diff
@@ -4,4 +4,4 @@
 import ReviewEditorConfigJson from 'interfaces/review-editor-config-json';
 import * as React from 'react';
 
-export const ReviewEditorConfigContext = React.createContext({} as ReviewEditorConfigJson);
+export const ReviewEditorConfigContext = React.createContext<ReviewEditorConfigJson>({ max_blocks: 0 });
```
nanaya, line 7: what's this change doing here 🤔 although I guess it's simple enough so this can stay

resources/js/modding-profile/main.tsx
```diff
@@ -0,0 +1,330 @@
...
+                </div>
+              </div>
+            </UserProfileContainer>
+          </BeatmapsContext.Provider>
+        </BeatmapsetsContext.Provider>
+      </DiscussionsContext.Provider>
+    );
+  }
+
+  private extraPage = (name: ModdingExtraPage) => {
```
nanaya, line 235: I think some of the components should be observer? ...or not? 🤔

nanaya (changes requested):

resources/js/beatmap-discussions/review-editor-config-context.ts
notbakaneko: `max_blocks` isn't supposed to be `undefined` and this just makes the default 0 so pages that don't need editing can skip adding a provider without causing `max_blocks` to be `undefined`

resources/js/modding-profile/main.tsx
notbakaneko: I wonder about that 🤔  (apparently everything that receives an `observable` prop should also be an `observer`, otherwise it's supposed to be `toJS`d even if the props don't change? 🤔 )

nanaya: yeah because otherwise when passing object and any of its properties get updated the non-observer component will be outdated. The toJS tells that the current component do access the properties so it'll update its children accordingly. Although in this case there's no update involved so it should be fine...? I'd probably still make them observer anyway just in case.

nanaya (approved):

PR #10314, opened 2023-06-27
Improve `SETUP.md` in a few ways

peppy:
### 1c4bf8947c Reword seeding section to allow searching using `populate` keyword

Some users may be searching for this and not using the uncommon "seeding" terminology.

### e91a9d927a Remove `$ ` prefixes in documentation

This causes issues when using the "copy to clipboard" functionality on the github website.

### ca54d53a0b Change default port of `beatmap-difficulty-lookup-cache` to avoid macOS conflict

macOS uses port 5000 for upnp stuff.


nanaya (approved):
the port 5000 on macos is for airplay btw 👀

PR #10319, opened 2023-06-29
Mark `UserCompact.default_group` as nullable in api docs

tybug:
If a user has a default usergroup of e.g. `34`, `default_group` will be null for them in api responses. I'm guessing this is because `identifier` is null at a database level for certain groups, including the above.

I was going to suggest that `identifier` should be set in the db for all groups. But that could lead to the identifier being leaked for hidden groups if someone manages to find a user in that group and make the appropriate api call, so I'm not sure that's such a good idea anymore.


nanaya (changes requested):
the typescript typing needs fixing as well (UserJson in interfaces, User in models)

tybug:
I think I've updated things correctly, let me know. Not 100% on the difference between the typescript `User` model and the php `User` model.

nanaya (approved):

PR #10321, opened 2023-06-29
Remove store address editing

### LINKED ISSUE #9490: Remove address management portions of store
osu-web doesn't handle this part anymore

notbakaneko:
No longer supported, handling of physical item shipping has gone via shopify partners for a while now.
Extra validation that forces an error if a non-shopify shippable item tries to checkout.
Placeholders hardcoded for old invoices.

closes #9490


nanaya:
Should the invoice printing related stuff be removed as well?

notbakaneko:
printing and shipping-only versions probably don't needed either without shipping being handled on our end 🤔

nanaya (approved):

nanaya (changes requested):
I clearly forgot about the comment I made a bit ago

nanaya (approved):

PR #10325, opened 2023-07-03
Include type of model being indexed when running EsDocument

notbakaneko:
everything being `EsDocument` in metrics isn't very useful...
Doesn't change the Laravel default output which is less than useful and less than trivial to override.


app/Jobs/EsDocument.php
```diff
@@ -66,5 +68,18 @@ public function handle(): void
...
+    }
+
+    private static function incrementStat(string $action, string $class)
+    {
+        Datadog::increment(
+            config('datadog-helper.prefix_web').'.es_document',
+            1,
+            [
+                'action' => $action,
+                'class' => $class,
```
nanaya, line 81: doesn't datadog tag have specific format requirement? (snake case etc)

nanaya: it will and the class names will look weird I think? Something like `app_models_artisttrack`

app/Jobs/EsDocument.php
```diff
@@ -66,5 +68,18 @@ public function handle(): void
         $keyName = $model->getKeyName();
         $model->setAttribute($keyName, $id);
         $model->esDeleteDocument();
+        static::incrementStat('delete', get_class($model));
```
nanaya, line 71: why one up there uses `$this->modelMeta(...)` and this one uses `get_class(...)` and I think they are all really just `$class`? or maybe don't bother passing the class name at all if the incrementStat function is an instance method?

app/Jobs/EsDocument.php
```diff
@@ -66,5 +68,18 @@ public function handle(): void
...
+    }
+
+    private static function incrementStat(string $action, string $class)
```
nanaya, line 74: static functions should be above constructor shouldn't they? And set a return type

nanaya (changes requested):

nanaya (approved):

PR #10328, opened 2023-07-03
Add missing fields to several models in api docs

tybug:
Hopefully relatively uncontroversial. The only unusual commit here is a22d29620fb10c71bb9ef81d9105d0aca862b5fe, which renames an incorrectly named field rather than adding a new one.


resources/views/docs/_structures/beatmapset.md
```diff
@@ -2,28 +2,29 @@
...
-storyboard                     | boolean                  | |
-submitted_date                 | [Timestamp](#timestamp)? | |
-tags                           | string                   | |
+Field                          | Type                         | Description
+------------------------------ | ---------------------------- | -----------------------------------------------------------------------
+availability.download_disabled | boolean                      | |
+availability.more_information  | string?                      | |
+bpm                            | float                        | |
+can_be_hyped                   | boolean                      | |
+current_nominations            | [Nomination](#nomination)[]? | |
```
nanaya, line 11: this is part of optional BeatmapsetCompact

resources/views/docs/_structures/build.md
```diff
@@ -33,6 +33,7 @@ Field             | Type                                | Notes
 ------------------|-------------------------------------|------
 changelog_entries | [ChangelogEntry](#changelogentry)[] | If the build has no changelog entries, a placeholder is generated.
 versions          | [Versions](#build-versions)         | |
+youtube_id        | string                              | |
```
nanaya, line 36: The field isn't optional but the value may be null

resources/views/docs/_structures/chat_channel.md
```diff
@@ -39,6 +40,7 @@ Optional attributes:
 Field                   | Type                                             | Description
 ----------------------- | ------------------------------------------------ | ------------------
 current_user_attributes | [CurrentUserAttributes](#currentuserattributes)? | only present on some responses
+first_message_id        | number?                                          |
```
nanaya, line 43: I don't think this exists anymore

nanaya (changes requested):

resources/views/docs/_structures/build.md
tybug: good catch, thanks. I have it marked null locally as well.

resources/views/docs/_structures/chat_channel.md
tybug: Doesn't seem returned on testing (though I don't have announce usergroup access to test that, in case it differs). It's been 22 months since I added support for the endpoint which uses this, so could definitely have been removed since then. 

I've removed it.

resources/views/docs/_structures/beatmapset_discussion.md
```diff
@@ -28,11 +28,13 @@ user_id                 | number
...
-praise      | |
-problem     | |
-review      | |
-suggestion  | |
+Name             | Description
+---------------- | -----------
+disqualify       | |
+hype             | |
+mapper_note      | |
+nomination_reset | |
```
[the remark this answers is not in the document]
tybug: hmm, I think they might have been mistaken from `BeatmapsetEvent` somehow. Doesn't seem to belong here on a double check. I've removed them.

resources/views/docs/_structures/build.md
nanaya: should still be moved to the main table instead of optional

resources/views/docs/_structures/beatmapset_compact.md
```diff
@@ -11,34 +11,38 @@ creator         | string                              | |
...
-pack_tags               | string[]              | |
-ratings                 |                       | |
-recent_favourites       |                       | |
-related_users           |                       | |
-user                    |                       | |
+Field                   | Type                         | Description
+----------------------- | ---------------------------- | -----------
+beatmaps                | [Beatmap](#beatmap)[]        | |
+converts                |                              | |
+current_nominations     | [Nomination](#nomination)[]? | |
```
nanaya, line 31: this field isn't nullable

nanaya (changes requested):

resources/views/docs/_structures/beatmapset_compact.md
tybug: ah, I guess you guys make a distinction between "nullable" and "not included at all". Fair enough.

nanaya: yeah. I think there are some optional fields that's nullable.

nanaya (approved):

PR #10334, opened 2023-07-04
Add notification count to page title

### LINKED ISSUE #9208: Show unread notification count in the browser tab title
Sometimes I keep the https://osu.ppy.sh/community/chat tab open to wait for a friend to come online. It'd be great if the browser tab title said "Friend・chat (1) | osu!" if I have one unread notification so that I don't have to check manually every few minutes or so, or even forget to check completely.

nanaya:
Something like this? 🤔 

The class name is a bit funky. The `With...` part is probably not needed as it should be handling all browser title updates?

Resolves #9208.


resources/js/core/browser-title-with-notification-count.ts
```diff
@@ -0,0 +1,43 @@
...
+  set title(newTitle: string) {
+    runInAction(() => {
+      this.origTitle = `${newTitle} | osu!`;
+    });
+  }
+
+  constructor(private core: OsuCore) {
+    document.addEventListener('turbolinks:load', this.setTitle);
+    document.addEventListener('turbolinks:before-cache', this.resetTitle);
+    makeObservable(this);
```
notbakaneko, line 21: `makeObservable` should be before the listeners get added otherwise the `@action` has no effect (which also does nothing by coincidence due to the `autorun` observing `origTitle` is always disposed when `setTitle` runs)

notbakaneko (changes requested):
Sometimes there's a flash of no notification count when navigating but not sure if worth fixing here atm 🤔

notbakaneko (approved):

PR #10344, opened 2023-07-05
Add kudosu ranking page

### LINKED ISSUE #6228: Kudosu ranking page on the new website
https://osu.ppy.sh/p/kudosu

This page exists and has yet to be ported to the new website in any form, it's quite cool to look at and can give modders a small goal to be proud of as an overall sign of their experience/progress. It would be a loss to see it go away entirely when the old site is decommissioned, and it could do well to have a new website page with some additions. The issue it currently has is that it is not very dynamic.

This page would be most useful with monthly and weekly leaderboards, in addition to an all-time one like the current page. Weekly could possibly be the default. Besides being a cool thing for modders to check out, this would be very useful for mappers as well - it could allow them to find out what modders are currently active in the community, giving them some idea of potential people to reach out to for feedback! Having weekly/monthly options would be much more dynamic than just one all-time ranking as well.

Since it's a rankings page, the design could be done similarly to the performance ranking page, just with the appropriate columns.

nanaya:
Yay? ┐(･_･)┌

There's a min-width of 650px on the table itself which isn't helpful (especially mobile) so I removed it.

- [x] ~Someone write description for the page please (if needed)~ not needed apparently

Resolves #6228.


peppy:
Probably fine to go without a description for now...

notbakaneko (approved):
seems fine, I guess

PR #10347, opened 2023-07-06
Replace generic page change event with specific event (if needed)

nanaya:
There were layzr and timeago. And now the former is gone and the latter doesn't depend on anything to have it updated correctly. All that's left is sync-height so those which actually need it can just do it directly.

As for sticky header/footer, I don't remember why it's needed in the first place 🤷 And scroll event should trigger it right away anyway.


notbakaneko (changes requested):
conflicts

notbakaneko (approved):

PR #10361, opened 2023-07-11
Support `after` parameter for comment pagination

nanaya:
Simpler alternative to sending cursor in some specific format which kind of depend on the sort type.


notbakaneko (approved):
extra query but simpler, I guess 🤷

PR #10362, opened 2023-07-11
Allow limited set of console usage

nanaya:
Let's get this one over already. It's either this or remove console usage entirely.


notbakaneko (commented):
I feel like `debug` shouldn't be allowed and the existing `debug`s replaced with `error` or something since they're basically just error messages (and the one that's just a trace message can be removed 🤔 )

notbakaneko (approved):

PR #10363, opened 2023-07-12
Paginate kudosu ranking page

nanaya:
Up to 1000 instead of `MAX_RESULTS` (10k) because the query gets a bit slow. Also it seems rather silly with the end of the table only having single digit kudosu.


notbakaneko (approved):

nanaya:
oh wait, the paginator didn't actually limit the page number

app/Http/Controllers/RankingController.php
```diff
@@ -223,10 +223,15 @@ public function index($mode, $type)
...
+        static $maxResults = 1000;
+
+        $maxPage = $maxResults / static::PAGE_SIZE;
+        $page = min(get_int(request('page')) ?? 1, $maxPage);
+
         $scores = User::default()
             ->with('country')
             ->orderBy('osu_kudostotal', 'desc')
-            ->paginate(static::PAGE_SIZE, ['*'], 'page', null, 1000);
+            ->paginate(static::PAGE_SIZE, ['*'], 'page', $page, $maxResults);
```
notbakaneko, line 234: actually, doesn't `paginate` only have 4 params? 🤔

notbakaneko (changes requested):

app/Http/Controllers/RankingController.php
nanaya: that's what you think 👀

notbakaneko: oh it's a magic param :|

notbakaneko (approved):

PR #10378, opened 2023-07-17
Add option to enable Centili payments for checkouts

notbakaneko:
Option to disable was requested; env is `CENTILI_ENABLED`, default is `false` (disabled).

I left the callbacks working, maybe they should be disabled too? 🤔


app/Libraries/OrderCheckout.php
```diff
@@ -227,7 +227,8 @@ private function allowCentiliPayment()
         // Geolocation header from Cloudflare
         $isJapan = strcasecmp(request_country(), 'JP') === 0;
 
-        return $isJapan
+        return config('payments.centili.enabled')
+            && $isJapan
```
nanaya, line 231: the variable can be inlined

nanaya (changes requested):

nanaya (approved):

PR #10387, opened 2023-07-19
Workaround invalid legacy match game ruleset

nanaya:
Apparently it's a thing.

- [ ] maybe not needed as it'll be fixed at the backend


nanaya:
Not needed anymore.

PR #10396, opened 2023-07-21
Change default rulesets for NAT group

### LINKED ISSUE #10390: Allow NAT members with no game mode to nominate maps from all game modes
discussed internally

nanaya:
Resolves #10390.


notbakaneko (changes requested):
Could do with some tests for some/no modes specified

notbakaneko (approved):

PR #10398, opened 2023-07-21
Allow changing country from account settings

### LINKED ISSUE #10382: Add support for user initiated flag changes
We've now got the required data dating back 6 months, so we can add proposed support for flag changes.

## Data structure

```sql
root@db-master:osu> show create table user_country_history\G;
Table        | user_country_history
Create Table | CREATE TABLE `user_country_history` (
  `user_id` int unsigned NOT NULL,
  `year_month` char(4) NOT NULL,
  `country_acronym` char(2) NOT NULL,
  `count` int unsigned NOT NULL DEFAULT '1',
  `last_updated` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  PRIMARY KEY (`user_id`,`year_month`,`country_acronym`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci COMMENT='tracking user access from different countries for the purpose of country change requests'

root@db-master:osu> select * from user_country_history where user_id = 2;
| user_id | year_month | country_acronym | count | last_updated        |
|---------|------------|-----------------|-------|---------------------|
| 2       | 2302       | JP              | 23    | 2023-02-28 10:13:00 |
| 2       | 2303       | JP              | 19    | 2023-03-30 07:09:52 |
| 2       | 2304       | JP              | 2     | 2023-04-04 07:30:10 |
| 2       | 2305       | JP              | 24    | 2023-05-31 02:24:18 |
| 2       | 2306       | JP              | 21    | 2023-06-26 15:34:06 |
```

## Initial rules

A flag change can be initiated by a user when:

- There are 6 months of data
- At least 4 of the 6 months have logins from only one `country_acronym`, and all are the same `country_acronym`
- The user is not registered for any ongoing tournaments

We may relax or change this in the future.

## Proposed UX

Country change available:

2023-07-18 22 41 48@2x

Country change unavailable:

2023-07-18 22 39 23@2x

nanaya:
Resolves #10382.

Depends on
- [x] #10395
- [x] #10394


resources/views/accounts/_edit_country.blade.php
```diff
@@ -0,0 +1,43 @@
...
+    use App\Libraries\User\CountryChangeTarget;
+    use App\Models\Country;
+
+    $countryChangeTarget = CountryChangeTarget::get($user);
+@endphp
+<div class="account-edit-entry account-edit-entry--read-only">
+    <div class="account-edit-entry__label account-edit-entry__label--top-pinned">
+        {{ osu_trans('accounts.edit.profile.country') }}
+    </div>
+    <div>
```
notbakaneko, line 15: could do with some better wrapping and flag centering 
image

app/Http/Controllers/AccountController.php
```diff
@@ -156,6 +159,21 @@ public function update()
         return json_item($user, new CurrentUserTransformer());
     }
 
+    public function updateCountry()
+    {
+        $newCountry = get_string(Request('country_acronym'));
```
notbakaneko, line 164: `request`

tests/Controllers/AccountControllerTest.php
```diff
@@ -64,6 +67,26 @@ public function testInvalidIdsInProfileOrder()
             ->assertJsonFragment(['profile_order' => $newOrder]);
     }
 
+    /**
+     * @dataProvider dataProviderForUpdateCountry
+     *
+     * More complete tests are done through CountryChange and CountryChangeTarget.
+     */
+    public function testUpdateCountry(int $months, bool $success): void
```
notbakaneko, line 75: I think this test case should explicitly be testing against the countries matching (instead of assigning number of months) and leave the number of months testing to the other tests

notbakaneko (changes requested):

app/Libraries/User/CountryChangeTarget.php
```diff
@@ -0,0 +1,106 @@
...
+        // First group countries by year_month
+        $byMonth = [];
+        foreach ($history as $entry) {
+            $byMonth[$entry->year_month] ??= [];
+            $byMonth[$entry->year_month][] = $entry->country_acronym;
+        }
+
+        // For each year_month, summarise each countries
+        $byCountry = [];
+        foreach ($byMonth as $month => $countries) {
```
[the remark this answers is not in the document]
nanaya: not sure what you mean by regular array ?_?

although I noticed the month variable isn't actually used

notbakaneko (approved):

PR #10402, opened 2023-07-24
Add `guest` and `nominated` types to get user beatmaps docs

tybug:
`guest` has been available since https://github.com/ppy/osu-web/pull/8734.


notbakaneko (changes requested):
might as well add `nominated` while we're at it.

tybug:
didn't even realize that was a thing...

notbakaneko (approved):

PR #10404, opened 2023-07-25
Remove redundant isNew attribute

nanaya:
Laravel already have the exactly same attribute just with different name.


notbakaneko (changes requested):
conflicts 💥

notbakaneko (approved):

PR #10406, opened 2023-07-25
Move playlist item high score functions to its model

nanaya:
Also put score pass and total score together instead of two separate locations.

Additionally add default values for PlaylistItemUserHighScore because otherwise the new score will be compared with null.

Also I just checked UserScoreAggregate and remembered why it has such weird class name related to its table name ಠ_ಠ


notbakaneko (changes requested):
conflicts 💥

notbakaneko (approved):

PR #10414, opened 2023-07-26
Add support for non-AWS S3 compatible storage

nanaya:
I _think_ a null endpoint should be fine for using aws s3 since that's the current default.

- [x] Depends on #10412


notbakaneko (approved):
seems to working for non-aws 🤔 

Something to note before we try switching, is any operation using the `s3` filesystem will fail silently if there's an error unless `'throw' => true` is set in the config

notbakaneko (approved):
🤔

PR #10422, opened 2023-07-30
Add link to osu.js API wrapper

L-Mario564:
I maintain a JS/TS API wrapper called osu.js that supports API v1 and v2 and would like to see it added to the current list of API wrappers.

[Repository](https://github.com/L-Mario564/osu.js)


notbakaneko:
Are there any public project using this?

L-Mario564:
To my knowledge, there's only two public repos that use this package.

[https://github.com/squink13/sega-backend](https://github.com/squink13/sega-backend)
[https://github.com/squink13/sega-frontend](https://github.com/squink13/sega-frontend)

peppy (approved):

PR #10423, opened 2023-07-31
Typescript comments manager component

nanaya:
This combines the functionalities in comments manager and various data stores into one controller. Events are called directly instead of relying on pubsub.

other changes include
- bundled comments are not sorted client side as it's assumed the order returned from server is correct
- only exception is for newly posted comment by the user which it's added at the top
- only updated comment/commentablemeta/user model/json are updated as otherwise unrelated things will be rerendered when new bundle is loaded


notbakaneko (changes requested):
conflicts

resources/js/components/comments-controller.ts
```diff
@@ -0,0 +1,538 @@
...
+    this.state.isFollowing = bundle.user_follow;
+    this.state.hasMore[bundle.has_more_id] = bundle.has_more;
+    if (bundle.top_level_count != null && bundle.total != null) {
+      this.state.topLevelCount = bundle.top_level_count;
+      this.state.total = bundle.total;
+    }
+
+    for (const comment of [...bundle.comments, ...bundle.included_comments, ...bundle.pinned_comments]) {
+      const id = comment.id;
+      if (this.state.comments[id]?.updatedAt !== comment.updated_at) {
```
notbakaneko, line 473: The bundle returned when replying uses the original timestamp instead of the updated one, so the reply count doesn't get updated correctly (also means first reply to a comment doesn't show).

notbakaneko (changes requested):

resources/js/components/comments-controller.ts
```diff
@@ -0,0 +1,538 @@
...
+  isFollowing: boolean; // only for the first commentableMetaItem
+  pinnedCommentIds: number[];
+  sort: string;
+  topLevelCount: number;
+  total: number;
+  users: Partial<Record<number, UserJson>>;
+  votedCommentIds: Set<number>;
+}
+
+interface StateJson {
```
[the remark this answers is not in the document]
nanaya: Not sure about this. Some model or non-json js object may accidentally end up here if deriving from state.

resources/js/components/comments-controller.ts
nanaya: more like the updated_at column wasn't touched at all when incrementing...

notbakaneko (changes requested):
And now I've just noticed: reply -> go to another page -> back -> reply is gone 👀 (or just any posted comment, in general)

PR #10426, opened 2023-08-01
Fix saving new user playlist high score

nanaya:
"Updating" nonexistent record is noop in laravel because apparently it previously updates the whole table instead (laravel/framework#12083) (ha ha ha) and then updated to be noop because filling the model and saving it as requested makes too much sense.


notbakaneko (approved):
ಠ_ಠ

PR #10430, opened 2023-08-02
Add group history page

### LINKED ISSUE #6216: Record and show group changes

cl8n:
resolves #6216

- [ ] depends on #9488

I've tested this to work & look how I expected, but it's been a while since I've written some of this code, so I'm going to keep it marked as draft until I do another full pass over it myself

<details><summary>design w/ screenshots</summary>

the search form is basically same as featured artists track search and the events list is basically same as user profile recent activity area. the icon BGs light up with the group's color, if it has one.

</details>


resources/js/group-history/group-store.ts
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import GroupJson from 'interfaces/group-json';
+import { sortBy } from 'lodash';
+import { action, computed, makeObservable, observable } from 'mobx';
+
+class GroupStore {
+  @observable byId = observable.map<number, GroupJson>();
```
nanaya, line 9: observable.map can just be normal object

resources/js/group-history/event.tsx
```diff
@@ -0,0 +1,87 @@
...
+    const mappings: Record<string, React.ReactNode> = {
+      group: (
+        <a href={route('groups.show', { group: event.group_id })}>
+          {event.group_name}
+        </a>
+      ),
+    };
+
+    if ('playmodes' in event && event.playmodes != null) {
+      mappings.playmodes = transArray(
```
nanaya, line 31: should probably be called ruleset

resources/css/bem/group-history-event.less
```diff
@@ -0,0 +1,65 @@
...
+      user-remove: user-minus;
+      user-remove-playmodes: user-tag;
+      user-set-default: user-cog;
+    };
+    each(@icons, {
+      .@{_top}--@{key} & {
+        @icon-var: 'fa-var-@{value}';
+        --icon: @@icon-var;
+      }
+    });
```
nanaya, line 28: set at block level? Also for the message bold styling

nanaya (changes requested):

resources/js/group-history/search-form.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+    return (
+      <form className={bn} data-loading-overlay='0' onSubmit={this.onSubmit}>
+        <div className={`${bn}__content ${bn}__content--inputs`}>
+          <InputContainer for={`${bn}-group`} labelKey='group_history.form.group'>
+            <div className='form-select form-select--group-history'>
+              <select
+                className='form-select__input'
+                id={`${bn}-group`}
+                name='group'
```
nanaya, line 46: `for`, `name`, and `id` aren't needed?

nanaya (changes requested):

resources/css/bem/group-history-event.less
```diff
@@ -0,0 +1,65 @@
...
+  display: flex;
+  font-size: @font-size--title-small;
+  gap: 15px;
+
+  @bold-events: group-add, group-remove, group-rename;
+  each(@bold-events, {
+    &--@{value} {
+      --message-weight: bold;
+    }
+  });
```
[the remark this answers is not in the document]
cl8n: not sure how you wanted me to do that. do u mean something like 
```
@icons: {
    group_add: @fa-var-users, bold;
    // ....
}
```
and use both values in the `each()` below?

resources/js/group-history/search-form.tsx
cl8n: `name` is for `onChange()` 

and hm I thought `for`/`id` would give the same kind of click target behavior as on the other inputs but i guess not. I've seen this used on label+select before, does it help with accessibility or something? :thinking:

nanaya: it's not needed when the input is inside the label.

resources/css/bem/group-history-event.less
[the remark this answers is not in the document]
nanaya: yes

Hiviexd:
Maybe we can get this in as-is (with only necessary changes) as an MVP for people to use? I got a few inquiries recently asking about why is the current group history page just a json response.

notbakaneko:
@nanaya do you want to just make the updates directly?

nanaya:
yeah maybe

resources/js/group-history/group-store.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  @computed
+  get byIdentifier() {
+    const byIdentifier = new Map<string, GroupJson>();
+
+    for (const group of this.byId.values()) {
+      byIdentifier.set(group.identifier, group);
+    }
+
+    return byIdentifier;
```
notbakaneko: can use `mapBy` (and add `Iterable` to param type)

notbakaneko (changes requested):
Navigating via history doesn't restore the correct page, everything gets offset

resources/js/group-history/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+import { navigate } from 'utils/turbolinks';
+import { updateQueryString, wikiUrl } from 'utils/url';
+import Events from './events';
+import groupStore from './group-store';
+import GroupHistoryJson from './json';
+import SearchForm from './search-form';
+
+type MoreParams = GroupHistoryJson['params'] & { cursor_string: string };
+
+export const formParamKeys = ['group_id', 'max_date', 'min_date', 'user'] as const;
```
notbakaneko: should probably be exported from `search-form` or `json.ts`  instead

resources/js/group-history/search-form.tsx
```diff
@@ -0,0 +1,132 @@
...
+        </div>
+      </form>
+    );
+  }
+
+  @action
+  private readonly onChange = (event: React.ChangeEvent<HTMLInputElement | HTMLSelectElement>) => {
+    event.preventDefault();
+
+    const name = event.currentTarget.name as Exclude<typeof formParamKeys[number], 'sort'>;
```
notbakaneko: the `Exclude` doesn't do anything

resources/js/group-history/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+        data: params,
+        dataType: 'json',
+        method: 'GET',
+      },
+    )
+      .done(action((response: GroupHistoryJson) => {
+        groupStore.update(response.groups);
+        this.setMoreParamsFromJson(response);
+
+        if (this.loading === 'new') {
```
notbakaneko: the entire method only seems to be used for load more

resources/js/group-history/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+          this.events = response.events;
+        } else {
+          this.events.push(...response.events);
+        }
+      }))
+      .fail(onErrorWithCallback(() => this.loadEvents(params)))
```
notbakaneko: There doesn't seem to be anything to trigger the callback? There's no login required.

resources/js/group-history/index.tsx
```diff
@@ -56,8 +56,8 @@ export default class GroupHistory extends React.Component<Props> {
     // If the "group" param doesn't match any group we can show as a select
     // option, set it to null in the new params. This prevents the new params
     // from initially being out of sync with the displayed form controls
-    if (json.params.group != null && !groupStore.byIdentifier.has(json.params.group)) {
-      this.newParams.group = null;
+    if (json.params.group_id != null && !groupStore.byId.has(json.params.group_id)) {
```
notbakaneko, line 54: `byIdentifier` isn't need any more making the other change unnecessary :D

notbakaneko (changes requested):

PR #10433, opened 2023-08-03
Add `ProfileBanner` `image@2x` field to api docs

Joehuu:
Wasn't added in https://github.com/ppy/osu-web/pull/9861.

### DIFF
--- a/resources/views/docs/_structures/user_compact.md
+++ b/resources/views/docs/_structures/user_compact.md
@@ -83,8 +83,8 @@ Field         | Type        | Description
 --------------|-------------|------------
 id            | number      | |
 tournament_id | number      | |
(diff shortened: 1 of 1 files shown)


resources/views/docs/_structures/user_compact.md
```diff
@@ -84,6 +84,7 @@ Field         | Type        | Description
 id            | number      | |
 tournament_id | number      | |
 image         | string      | |
+image@2x      | string      | |
```
notbakaneko, line 87: should be updated to `string?` with `image`

notbakaneko (changes requested):

notbakaneko (approved):

PR #10435, opened 2023-08-03
Cleanup layout footer gap

notbakaneko:
Moves the `margin-bottom` gap to the section container.
Apparently it's supposed to be everywhere; except it looks really odd in chat, especially on mobile, so I removed it there.

### DIFF
--- a/resources/css/bem/osu-layout.less
+++ b/resources/css/bem/osu-layout.less
@@ -118,12 +118,10 @@
     &--page {
       ._page();
       padding: 20px;
-      margin-bottom: 10px;
     }
 
     &--page-compact {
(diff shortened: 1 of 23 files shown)


resources/css/bem/osu-layout.less
```diff
@@ -136,13 +130,15 @@
   &__section {
     display: flex;
     flex-direction: column;
+    margin-bottom: 10px;
```
nanaya, line 133: account settings page now has 15px margin (from previously also incorrect 5px)

nanaya (changes requested):

nanaya (approved):

PR #10444, opened 2023-08-07
Clean up group page variant of `UserList` component

cl8n:
pass in the group rather than splitting into a bunch of related props


nanaya (approved):
not sure if all that useful but sure I guess 🤷 (if anything this makes the component less flexible with title and description only available for group page but nothing else is using it atm either so)

PR #10446, opened 2023-08-07
Add history button to group pages

cl8n:
- [ ] depends on #10430
- [ ] depends on #10444


peppy:
As with others, would be helpful to fill the opening post with more details like screenshots of how the system looks.

Hiviexd:
superseded by https://github.com/ppy/osu-web/pull/12450

PR #10449, opened 2023-08-08
Fix nested links rendering in chat messages

notbakaneko:
Also, make `linkRenderer` itself skip rendering a link if it's already rendering a link.
This also makes chat messages render shortened beatmap discussion urls if the markdown doesn't have any other text content (or maybe it should alway override like discussions? or discussions not always override? 🤔 )


resources/js/beatmap-discussions/renderers.tsx
```diff
@@ -14,16 +14,27 @@ export function createRenderer(ElementType: React.ElementType) {
...
 
-export function linkRenderer(astProps: JSX.IntrinsicElements['a']) {
+export function linkRenderer(astProps: JSX.IntrinsicElements['a'], alwaysShortenUrls = true) {
   const props = propsFromHref(astProps.href);
   const href = safeReactMarkdownUrl(props.href);
 
+  const override = alwaysShortenUrls
+    || astProps.children instanceof Array
+      && astProps.children.length > 0
+      && astProps.children[0] === astProps.href;
```
nanaya, line 24: Is this supposed to make `[thing](https://..discussion.link..)` link says `thing`? It doesn't seem to work

resources/js/chat/message-item.tsx
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
 // See the LICENCE file in the repository root for full licence text.
 
+import { linkRenderer } from 'beatmap-discussions/renderers';
```
nanaya, line 4: why is chat importing thing from beatmap discussions

nanaya (changes requested):

resources/js/beatmap-discussions/renderers.tsx
notbakaneko: image

works for me ?_?

notbakaneko: discussion page always uses the discussion id text; that was the question of maybe it shouldn't or the text should always be the discussion/post id when it's a discussion.

> (or maybe it should alway override like discussions? or discussions not always override? 🤔 )

nanaya: oh, I thought that was a bug. What was the reason it currently always overrides the text?

nanaya: also the original question doesn't really make sense to me. Why would chat have shortened link just for beatmapset discussion? How about never shorten it for chat? Or alternatively label it with `modding:` or something and then maybe we can do `user:` for user link etc. (although that seems to be another rabbit hole)

notbakaneko: I fell like the answer is that it was just simpler to overwrite the content  when it was implemented, instead of checking if it needed to be replaced, which is moot now since it needs to check 👀

nanaya (approved):

Issue #10453, opened 2023-08-10
Migrations which upgrade `mediumint` columns to `int` do not have any effect


bdach:
Originally reported [on discord](https://discord.com/channels/188630481301012481/188630616286167041/1139257326801530910).

It appears that since https://github.com/ppy/osu-web/pull/10247 (I've confirmed experimentally that it's since the laravel 10 upgrade specifically), migrations that upgrade/widen `mediumint` columns to `int` have stopped functioning correctly. One such instance is this:

https://github.com/ppy/osu-web/blob/36d197ddf7ffcbdea041c7a088ebd104042a0851/database/migrations/2015_01_01_133337_base_tables.php#L175

to this:

https://github.com/ppy/osu-web/blob/36d197ddf7ffcbdea041c7a088ebd104042a0851/database/migrations/2020_05_15_083037_sync_structure.php#L223

After running all migrations, the table should have type `int unsigned`, while in fact it appears to have type `mediumint unsigned`.

The direct implications of this are that:

- [test databases](https://github.com/ppy/osu-web/blob/master/SETUP.md#initializing-the-test-database) will not be consistent schema-wise with production
- in local setups, databases may fail to seed as the seeding data will not be insertable to tables due to columns being too narrow

[Here is a gist](https://gist.github.com/bdach/8d756f9bcaf8b42a4d2b587bb6695f27) with 2 excerpts of mysql statements executed during  captured using `general_log` while the aforementioned migration was running.

- On 2023.706.0 (last commit before merge of aforementioned PR) the column width changes are successfully generated
- On current master the column width changes are skipped entirely.

There are several threads of possibility where this might have broken (changes in dbal, laravel itself...) but given I know very little about the php ecosystem I'm not willing to hazard guesses or link any links.

It may be wise to run a full cross-check of test/local db schemas against production to make sure no other cases of such silent breakage have occurred somewhere else.

nanaya:
Fixed in #10454

PR #10454, opened 2023-08-10
Remove `doctrine/dbal` and fix migration improper column types

NiceAesth:
Addresses #10453

## Summary

Ever since https://github.com/laravel/framework/pull/45487, Laravel has added support for column modifying which dropped `doctrine/dbal` as a dependency. 

However, if the package is still installed it will continue to use it for type mappings. This caused issues with certain types which the maintainers chose to not map to mysql/mariadb specific types. (e.g. mediumint and int both being mapped to INTEGER, causing migrations between them to effectively do nothing)

## Changes done

- Removed `doctrine/dbal` as a dependency
- Updated migrations to remove previous workarounds caused by dbal
- Changed migrations between `increment()` and `bigIncrement()` since Laravel always sets it as a primary key causing errors such as ```SQLSTATE[42000]: Syntax error or access violation: 1068 Multiple primary key defined (Connection: mysql, SQL: alter table `osu_scores_high` modify `score_id` bigint unsigned not null auto_increment primary key)```

Note: JSON type and other MySQL types which previously did not have mappings due to dbal maintainers not wishing to support different versions of SQL now have proper mappings in Laravel 10. Additionally, nullable now also specifies default as null with MySQL.

You may see a schema dump obtained from running the new migrations on a fresh database [here](https://gist.github.com/NiceAesth/9f533f2ef7a13a80261860d8b0d8ed37)


nanaya (changes requested):
It looks like nullable setting isn't preserved on schema change with native migration.

NiceAesth:
Of note is that all of the `nullable(false)` calls are technically not needed anymore. Did not touch those.

nanaya (changes requested):
Still missing nullable:
- `changelog_entries.github_user_id` in `database/migrations/2020_05_15_083037_sync_structure.php`
- `contest_entries.user_id` in `database/migrations/2020_05_15_083037_sync_structure.php`

nanaya (approved):

PR #10458, opened 2023-08-11
Lock usernames for users with Loved maps or guest difficulties

cl8n:
- [x] depends on #10457

the Loved part was brought up in a project loved chat, ppl were surprised that Loved worked any different from Ranked here. and then I also added guest diffs cuz that seemed right too :thinking:


app/Libraries/UsernameValidation.php
```diff
@@ -82,33 +85,42 @@ public static function validateUsername($username)
...
+                ->withStates(['approved', 'loved', 'qualified', 'ranked'])
+                ->where(function (Builder $query) use ($userIds) {
+                    $query
+                        ->whereIn('user_id', $userIds)
+                        ->orWhereHas(
+                            'beatmaps',
+                            fn (Builder $query) => $query->whereIn('user_id', $userIds),
+                        );
+                })
+                ->exists()
```
nanaya, line 115: Checking beatmap state itself should be sufficient? Should usernames of pending guest diff that's part of loved map made available?

```
Beatmap::scoreable()->whereIn('user_id', $userIds)->exists()
```

nanaya (changes requested):

app/Libraries/UsernameValidation.php
cl8n: good point. fixed, although I have it checking beatmapset still because there are cases where the host doesn't mark themselves as owner of any map (common on collabs from what i can tell)

nanaya (approved):

Issue #10207, opened 2023-05-18
Remaining features exclusively on old.ppy.sh


cl8n:
this is a tracking issue for anything that should be done prior to shutting down the old website. the topic comes up now and then so I think it'd be good to at least address all of the remaining use cases the old website has. please edit if I missed anything

- [x] [Ban history](https://old.ppy.sh/p/bans) (#6484)
- [ ] [Event feed](https://old.ppy.sh/p/events) (#3678)
- [x] [Kudosu ranking](https://old.ppy.sh/p/kudosu) (#6228)
- [x] Uploading beatmap packs (#7427)
- [x] Editing beatmap display title (#6023)
- [x] Mapping contest judging support (#903)
- [ ] [Viewing legacy mapping contest entries and results](https://old.ppy.sh/p/contestresults)
- [ ] Reading legacy forum PMs (functionality has been replaced, but there is currently no way to read old messages)
- [ ] List of people currently browsing a forum ([screenshot](https://user-images.githubusercontent.com/4358724/239418330-1b9c583d-b4b2-46da-ba72-d8a05dd54be7.png))

peppy:
Is there a reason why two of these are in the list but already completed?

Also, for the time being the old web will remain but login will be disabled (so completely read-only). This makes a lot of the remaining pieces a non-issue. Only real once is forum PMs, which I guess we'll offer an export feature for or something. like that

WitherFlower:
Has the stance changed on the overall unweighted accuracy stat ? It would be cool to see it return on the score ranking (it was only available on the score ranking on the old website)

Related issue : 
- https://github.com/ppy/osu-web/issues/2665

Hiviexd:
uploading beatmap packs is already revamped with a new workflow that involves discord, so that can be marked as done

cl8n:
@peppy I listed everything I could find that was missing -- those issues are marked resolved, but to my knowledge there is no replacement so I still included them. can mark them off if the final verdict is to just not have them anymore

and yeah as Hivie pointed out I wasn't up to date on the packs process, I thought that page was still needed. Stefan told me he still uses the old site for this but apparently that will change soon :eyes:

abraker95:
For completeness, the new site is missing people currently browsing the forum

![image](https://github.com/ppy/osu-web/assets/4358724/1b9c583d-b4b2-46da-ba72-d8a05dd54be7)

[...]

PR #10460, opened 2023-08-13
Add contest judging system

### LINKED ISSUE #903: Contest judging system
The contest system should have the following features for proper judge-based voting:

* Support for up to 16 judges
* Support for double-blind voting by masking entrant usernames with a masked name, including altering .osu file metadata to reflect this masked name
* Support for up to 12 custom judging score categories on a 1-10 rating scale. Score categories are things like "flow and structure", or "judge's impression", though ideally provision would exist to create these categories per contest instead of being limited to a certain set (as per [the old system](https://puu.sh/t8yJE/e96dce870f.png))
* Per entry, a text field for judge comments
* Results page per user entry including judging comments and per judge scoring in the various aforementioned score categories
* Contest ranking page that displays winning entries and all other entries sorted by score
* Panel which displays overall judging completion percent (visible only to administrators)

Basically if /p/contestresults and /p/contest have it, it should be in the new contest system as well.

venix12:
resolves #903 (except the altering metadata part, i don't think it's in scope of that PR)
part of #10207

<details>
  <summary>Preview screenshots</summary>
  
  Contest judging panel, visible only to assigned judges on a per-contest basis.
  contest judge

  Button in contest results that takes you to the judging results.
  contest results
  
  Contest judging results. 
  contest judge results
  
</details>

it could probably be more pretty :eye:


peppy:
@venix12 if possible could you add a video or screenshots showing what this is adding?

venix12:
I've put the screenshots in the description.

notbakaneko (changes requested):
conflicts

peppy:
@notbakaneko are you able to resolve conflicts in cases like this on their behalf? It's a bit weird to not review something from our end for a  month then fire back that there's conflicts 😅 . Seems like an easy resolution.

app/Models/ContestJudgeCategoryVote.php
```diff
@@ -0,0 +1,33 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+declare(strict_types=1);
+
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+/**
+ * @property ContestJudgeCategory $category
```
notbakaneko, line 13: For reference, stuff that goes through relations are readonly on the actual property itself and should be `@property-read`

venix12:
resolved the conflicts :eyes:

resources/js/entrypoints/contest-judge.tsx
```diff
@@ -0,0 +1,18 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+import Main from 'contest-judge/main';
+import core from 'osu-core-singleton';
+import * as React from 'react';
+import ContestEntryStore from 'stores/contest-entry-store';
+import { parseJson } from 'utils/json';
+
+const store = new ContestEntryStore();
+store.updateWithJson(parseJson('json-entries'));
```
notbakaneko, line 11: This should go in the register callback, top level module code run when the module is first loaded, not when the page is loaded.

notbakaneko (changes requested):
Some quick notes first:
- is judging limited to specific types of contests?
- link to judging page from somewhere would be useful
- more context on the judging page on what's being judged would be useful, now it looks like a bunch of titles referring to what?

venix12:
> Some quick notes first:
> 
> * is judging limited to specific types of contests?

It was traditionally used for beatmap contests only, don't really think there are currently any other use cases (?). If anything things can be adjusted when there's a need I'd say (actually would boil down to just adding judging results button to other listing types probably).
> * link to judging page from somewhere would be useful

Right, was an oversight from my side, added.
> * more context on the judging page on what's being judged would be useful, now it looks like a bunch of titles referring to what?

I added a question-mark-icon-tooltip kind of thing along with description field on categories, it should suffice i think?

notbakaneko:
> I added a question-mark-icon-tooltip kind of thing along with description field on categories

I mean on the judging page itself you'd see a bunch of titles with sliders and text boxes with no context of the thing being judged

venix12:
Ah well, it's using entry masked name for title now. Assumption was the entries would be pulled and anonymized at database level using some external script (basically what's done at the moment I think) and then sent to judges so the judges should know which title goes for which entry (the masked name being set for author in beatmap file).

notbakaneko:
@cl8n do you have any feedback for this?

cl8n:
yeah I've tested this one a bit and will post a full review here. I think I can get to it this weekend but if not I'll give an update regardless

resources/js/components/basic-select-options.tsx
```diff
@@ -32,6 +32,8 @@ export default class RankingSelectOptions extends React.PureComponent<Props> {
 
   private href(id: number | null) {
```
notbakaneko, line 33: At this point the component should probably just accept a callback to get the `href`.
Also, `modifiers` passed to `SelectOptions` should no longer be `spotlight`.

resources/views/contests/voting.blade.php
```diff
@@ -32,6 +38,12 @@
         <div class='contest__voting-notice'>
             {{ $noVoteReason }}
         </div>
+
+        @if ($contestMeta->isJudged() && priv_check('ContestJudge', $contestMeta)->can())
```
notbakaneko, line 42: Do admins automatically become judges? 🤔

notbakaneko (changes requested):
Quick one for the things that stood out while I go through the rest 👀

database/migrations/2023_07_20_153416_create_contest_judge_category_votes_table.php
```diff
@@ -0,0 +1,39 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('contest_judge_category_votes', function (Blueprint $table) {
```
notbakaneko, line 19: I feel like these are scores or grades, not votes 🤔

notbakaneko (changes requested):
The pivot tables with unique keys don't really need `id`s and should probably use a composite key instead.

`_votes` should probably be called `_scores` instead, especially if the grading criteria is always numeric and the table with the `comments` should probably be `contest_judge_comments`.

resources/views/contests/voting.blade.php
venix12: guess this certain check could be added to the ``alwaysCheck`` stuff :eyes:

venix12:
Personally feels like ``votes`` make more sense (for table with comments) given its usage, as it's got ``scores`` belonging to it (``contest_judge_vote_id`` with hasMany on ``votes``), pretty much always used that way. ``comments`` as a base for ``scores`` feels kinda off? 🤨

...unless you want me to disattach the hasMany relationship there and rather have both comments and scores relate to ``entry`` and ``user`` on its own?

resources/js/components/basic-select-options.tsx
venix12: well, multiplayer one is still being casted from blade via registerComponents thing so not sure how to go with callback thing considering that :thinking:

notbakaneko:
I'll get back to this after the finishing going through the session and oauth verification stuff 👀

app/Http/Controllers/ContestEntriesController.php
```diff
@@ -5,20 +5,142 @@
...
+            foreach ($entry->contest->scoringCategories as $category) {
+                $score = $scores
+                    ->where('contest_scoring_category_id', $category->getKey())
+                    ->first();
+
+                if ($score === null) {
+                    throw new InvariantException(osu_trans('contest.judge.validation.missing_score'));
+                }
+
+                $currentScore = ContestJudgeScore::where('contest_judge_vote_id', $vote->getKey())
```
notbakaneko, line 103:
```php
$vote->scores()->firstOrNew([
    'contest_judge_vote_id' => $vote->getKey(),
    'contest_scoring_category_id' => $category->getKey(),
])->fill([...])->save();
```

app/Http/Controllers/ContestEntriesController.php
```diff
@@ -5,20 +5,142 @@
...
+            } else {
+                $vote = ContestJudgeVote::create([
+                    'comment' => $comment,
+                    'contest_entry_id' => $entry->getKey(),
+                    'user_id' => auth()->user()->getKey(),
+                ]);
+            }
+
+            foreach ($entry->contest->scoringCategories as $category) {
+                $score = $scores
```
notbakaneko, line 95: Can use `Ds\Set` and get the intersection/diff for missing ids.
Or, just default missing values to 0.

app/Http/Controllers/ContestEntriesController.php
```diff
@@ -5,20 +5,142 @@
...
+            }
+        });
+
+        $updatedEntry = ContestEntry::with('judgeVotes')
+            ->with('judgeVotes.scores')
+            ->findOrFail($id);
+
+        $updatedEntryJson = json_item($updatedEntry, 'ContestEntry', ['current_user_judge_vote.scores']);
+
+        return $updatedEntryJson;
```
notbakaneko, line 131: can inline the return

app/Http/Controllers/ContestEntriesController.php
```diff
@@ -5,20 +5,142 @@
...
+                    ContestJudgeScore::create([
+                        'contest_scoring_category_id' => $category->getKey(),
+                        'contest_judge_vote_id' => $vote->getKey(),
+                        'value' => $value,
+                    ]);
+                }
+            }
+        });
+
+        $updatedEntry = ContestEntry::with('judgeVotes')
```
notbakaneko, line 125: `$entry->refresh()->load('judgeVotes.scores')`

notbakaneko (changes requested):
still checking....

app/Http/Controllers/ContestEntriesController.php
```diff
@@ -5,20 +5,142 @@
...
+        $params = get_params(request()->all(), null, [
+            'scores:array',
+            'comment',
+        ]);
+
+        $scores = collect($params['scores']);
+        $comment = $params['comment'];
+
+        DB::transaction(function () use ($scores, $comment, $entry) {
+            $vote = $entry->judgeVotes->where('user_id', auth()->user()->getKey())->first();
```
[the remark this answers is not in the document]
venix12: thought the ``auth()->user()`` was the preference looking all around the code everywhere :thinking:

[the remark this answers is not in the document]
notbakaneko: apparently using the facade is better for performance, so we've been moving back to that 👀

resources/js/contest-judge-results/header.tsx
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    return ret;
+  }
+
+  render() {
+    const { contest, entry } = this.props;
+
+    const score = entry.results?.votes;
+    const maxScore = (contest.max_judging_score ?? 0) * (entry.judge_votes?.length ?? 0);
```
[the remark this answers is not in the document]
venix12: well max_judging_score is sum of scoring_category values so technically yes 🤨

all judges should be required to vote on all entries for things to make sense though so it shouldn't differ per-entry of one contest, ended up moving this part to backend anyway since it probably makes way more sense.

resources/js/contest-judge/entry.tsx
```diff
@@ -0,0 +1,191 @@
...
+  @observable private posting = false;
+  @observable private readonly scores: ContestJudgeScoreJson[];
+  @observable private xhr?: JQuery.jqXHR;
+
+  constructor(props: Props) {
+    super(props);
+
+    this.comment = props.entry.current_user_judge_vote?.comment ?? '';
+    this.initialVote = toJS(props.entry.current_user_judge_vote);
+    this.scores = props.entry.current_user_judge_vote?.scores ?? [];
```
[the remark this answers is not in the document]
notbakaneko: Looking at this again, I think the whole `current_user_judge_vote` can be abstracted out to its own class where the scores are in a `Record` or `Map` and the comment is non-nullable?

notbakaneko (changes requested):

resources/js/components/basic-select-options.tsx
```diff
@@ -32,6 +32,8 @@ export default class RankingSelectOptions extends React.PureComponent<Props> {
...
     switch (this.props.type) {
+      case 'judge_results':
+        return route('contest-entries.judge-results', { contest_entry: id });
```
notbakaneko, line 36: oh yeah, `contest_entry` needs to be a non-nullable value

venix12: how would you go about resolving that? ...honestly not really ceratin about what to do with it🤔

notbakaneko: just set it to 0 or return an empty href for now

resources/js/contest-judge/entry.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+      return false;
+    });
+
+    if (
+      !scoresHaveChanged &&
+      this.currentVote.comment === this.initialVote.comment
+    ) return true;
+
+    return false;
```
notbakaneko, line 58: Also should check if all the scores are set (I'm assuming the comment is not required here?)

```suggestion
    return !(
      this.currentVote.scores.size === this.props.store.scoringCategories.length
        && (scoresHaveChanged || this.currentVote.comment !== this.initialVote.comment)
    );
```

notbakaneko (changes requested):

resources/js/contest-judge/entry.tsx
venix12: the comment is not required yes

notbakaneko (approved):

notbakaneko (commented):
Will do for now, I guess 👀 

(I also re-dated the migrations)

PR #10461, opened 2023-08-14
Fix migrations with dependencies failing if run by themselves

cl8n:
random problem i found while testing #10454. see `2016_10_28_000000_update_oauth_tables` for example

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -7,6 +7,7 @@ use App\Libraries\LocaleMeta;
 use App\Models\LoginAttempt;
 use Egulias\EmailValidator\EmailValidator;
 use Egulias\EmailValidator\Validation\NoRFCWarningsValidation;
+use Illuminate\Database\Migrations\Migration;
 use Illuminate\Support\Arr;
 use Illuminate\Support\HtmlString;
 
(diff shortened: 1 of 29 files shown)


nanaya (changes requested):
this won't be useful going forward. I'd rather update the related migrations to be in the new format and have them imported as needed (`(include $migration)->down()` etc)

app/helpers.php
```diff
@@ -1847,3 +1848,21 @@ function unmix(string $resource): HtmlString
...
+    );
+
+    if (!$instance instanceof Migration) {
+        throw new Exception(
+            "\"{$name}\" does not return an instance of ".Migration::class.'.',
+        );
+    }
+
+    return $instance;
+}
```
nanaya, line 1868: that's a lot of fluff just for this

```php
function migration(string $name): Migration
{
    return require database_path("migrations/{$name}.php");
}
```

- getRequire is just a fancy wrapper over plain require. The only benefit from it I can see is it throws full path?
- php already has way to check return type
- as for directory separator thing, just use `/`. It'll be fine

nanaya (changes requested):

app/helpers.php
cl8n: return type i couldnt do `Migration` because up() and down() aren't actually methods on that. but okay ill change the rest. i was just trying to follow as close as possible to what the migrator was doing normally

nanaya: but the check here isn't checking for the methods either...? what's the difference between this check with return your check

cl8n: oh my bad it was just a linter getting mad at me for `migration('...')->up()`. i guess i never actually tested it with Migration return type...

will change to what you wrote then :ok_hand:

nanaya (approved):

PR #10462, opened 2023-08-14
Enforce string query and mode search params

nanaya:
Also make sure to not trim nulls.

(this is mainly to not throw error when `query` parameter is an array)

### DIFF
--- a/resources/views/home/_search_results.blade.php
+++ b/resources/views/home/_search_results.blade.php
@@ -25,7 +25,7 @@
 
             <a
                 class="search-result__more-button {{ $showMore ? '' : 'search-result__more-button--hidden' }}"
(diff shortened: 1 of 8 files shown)


resources/views/home/_search_results.blade.php
```diff
@@ -25,7 +25,7 @@
 
             <a
                 class="search-result__more-button {{ $showMore ? '' : 'search-result__more-button--hidden' }}"
-                href="{{ route('search', ['mode' => $mode, 'query' => request('query')]) }}"
```
notbakaneko, line 28: I remember this being done intentionally to preserve the query as-is instead of changing `foo   ` to `foo` but I suppose it should be fine as long as the input box's text itself isn't changing 🤔

notbakaneko (approved):

PR #10468, opened 2023-08-15
Move WIP/pending beatmaps to graveyard when approving their beatmapset

cl8n:
rationale in the comment. always feels a little weird to me when i see loved mapsets with wip/pending diffs...


peppy (approved):
Probably makes sense. Have not tested.

app/Models/Beatmapset.php
```diff
@@ -558,11 +558,23 @@ public function setApproved($state, $user, ?array $beatmapIds = null)
         $approvedState = static::STATES[$state];
         $beatmaps = $this->beatmaps();
 
-        if (isset($beatmapIds)) {
-            if ($beatmaps->whereKey($beatmapIds)->count() === count($beatmapIds)) {
-                $beatmaps = $beatmaps->whereIn('beatmap_id', $beatmapIds);
-            } else {
-                throw new InvariantException('beatmap_ids contains invalid id');
+        if ($beatmapIds !== null) {
```
nanaya, line 561: can use a test

nanaya (changes requested):

nanaya (approved):

Issue #10251, opened 2023-05-26
Implement badges carousel to profile page


Hiviexd:
seems like this got forgotten when implementing the profile redesign around a year ago

currently, we're stuck with a big and ugly wall of pngs (also with an incorrect placement, at the top instead of the bottom)

![image](https://github.com/ppy/osu-web/assets/62819481/0aec2dbd-9e0a-48b2-a116-51fc01473118)

new design (from the [osu! figma library](https://www.figma.com/file/YIVQm9d6pvOemHsdjq6j9w/Web%2FProfile-Page-Proposal?type=design&node-id=0-1&t=3jA2sOH5kMVOd2eT-0))

![image](https://github.com/ppy/osu-web/assets/62819481/6e52efaa-ca2b-458c-8bd9-60d681d251df)

RockRoller01:
To note: it seems like a decent chunk of users may actively dislike this design, as a wall of badge has a certain prestige and epicness to it, so probably worth to consider if this is really the way to go forward with

cl8n:
yea def agree with rockroller, on top of the prestige factor it just doesn't feel right to me to push these awards off-screen when they usually require a lot of skill or dedication to get

not really sure why they ended up on top and without the bg color though :thinking:

Hiviexd:
any chance we could implement the new design without hiding badges in a carousel? i.e. moving badges to the bottom (optional?), adding the counter, and adding a proper bg color

RockRoller01:
moving them and adding the bg color is not a big task really, the badge counter however might look weird when it essentially fills the first badge spot in a wall of badges, its kinda meant for a single row design afterall

cl8n:
@arflyte considering the above conversation, can you give your input on whether this should change any further?

this is what the badges area looks like right now, for someone with a lot of badges

![](https://github.com/ppy/osu-web/assets/9954349/9480f1f5-7aeb-4ee2-92c1-869b47f8cd80)

nanaya:
I think the conclusion when we moved it back to the top is to keep the current one as is.

PR #10469, opened 2023-08-16
Update profile badge layout

cl8n:
part of #10251. although personally I think it is all that needs to be done (see comments on that issue)


peppy:
@cl8n could you please update the PR with screenshots for those of us that want to provide feedback without deploying ourselves?

nanaya (approved):

PR #10470, opened 2023-08-16
Update to php 8.2

nanaya:
Seems working.


notbakaneko (approved):
should be fine since I've been running php8.2 🤔

notbakaneko:
I guess this is going to need to have the required checks updated too 👀

PR #10473, opened 2023-08-17
Allow non-consecutive login month for country change condition

nanaya:


peppy:
@notbakaneko can you take a look at this one? should probably get it in as it's blocking some users from valid country changes currently.

notbakaneko (approved):

PR #10475, opened 2023-08-18
Add API route for kudosu ranking

### LINKED ISSUE #10374: make kudosu leaderboard available via api
title, would be handy for community projects

cl8n:
resolves #10374

### DIFF
--- a/app/Http/Controllers/RankingController.php
+++ b/app/Http/Controllers/RankingController.php
@@ -12,6 +12,7 @@ use App\Models\Spotlight;
 use App\Models\User;
 use App\Models\UserStatistics;
 use App\Transformers\SelectOptionTransformer;
+use App\Transformers\UserCompactTransformer;
 use DB;
 use Illuminate\Pagination\LengthAwarePaginator;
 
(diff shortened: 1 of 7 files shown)


resources/views/docs/_structures/user_compact.md
```diff
@@ -118,3 +119,12 @@ awarded_at  | Timestamp | |
 description | string    | |
 image_url   | string    | |
 url         | string    | |
+
+<div id="usercompact-kudosu" data-unique="usercompact-kudosu"></div>
```
nanaya, line 123: sort alphabetically

app/Http/Controllers/RankingController.php
```diff
@@ -229,10 +243,19 @@ public function kudosu()
         $page = min(get_int(request('page')) ?? 1, $maxPage);
 
         $scores = User::default()
-            ->with('country')
             ->orderBy('osu_kudostotal', 'desc')
             ->paginate(static::PAGE_SIZE, ['*'], 'page', $page, $maxResults);
 
+        if (is_json_request()) {
+            return json_collection(
```
nanaya, line 250: this should be inside `['ranking' => json_collection(...)]` just like other responses here (also new responses shouldn't return array in the first place)

nanaya (changes requested):

nanaya (approved):

PR #10483, opened 2023-08-21
Sync database structure

nanaya:
There are a few additional differences in prod which seems better suited to be updated directly instead:

- [x] `osu.contest_entries`: `entry_url` should have default value `null` (not `'NULL'` string)
- [x] `osu.osu_user_ircauth`: `user_id` should not have `auto_increment`
- [x] `osu_chat.channels`: `last_message_id` should be unsigned

And then there's a rather questionable one:
- `osu.migrations`: newer version has `id` primary key (unsigned int auto increment) 🤷

Migration entry:
- [x] `2023_08_17_133006_sync_structure`


peppy:
> osu.migrations: newer version has id primary key (unsigned int auto increment) 🤷

is this a laravel change?

peppy (approved):

nanaya:
> is this a laravel change?

yeah, the current schema looks like this:

```
CREATE TABLE `migrations` (
  `id` int unsigned NOT NULL AUTO_INCREMENT,
  `migration` varchar(255) NOT NULL,
  `batch` int NOT NULL,
  PRIMARY KEY (`id`)
)
```

peppy:
we can probably just apply that to prod.

notbakaneko (approved):
seems fine 🤔

peppy:
I've applied all the migrations to production, including the primary key change to `migrations`.

nanaya:
there's also a new migration entry you haven't checked 👀

peppy (approved):

PR #10484, opened 2023-08-22
Update country in ruleset variant statistics table when changing user country

nanaya:
Forgot about this 👀

Todo (after merging):
- [ ] fix country in variant tables


notbakaneko (approved):
apparently I forgot to add the comment about the variants in the original PR 🙈

nanaya:
A reminder for @peppy to fix the tables

peppy:
I've resynced both tables now. I've confirmed the original issue has been fixed.

PR #10488, opened 2023-08-23
Remove the Laravel Mix docs link in SETUP.md

agatemosu:
It confused me because this version is not used now


notbakaneko (changes requested):
We don't use mix at all, the line should be removed instead.

notbakaneko (approved):

PR #10493, opened 2023-08-25
Remove filtering multiplayer rooms by route path parameter

### LINKED ISSUE #10491: Filtering multiplayer Rooms by path mode parameter does not work
e.g. `rooms?mode=owned` works fine but `rooms/owned` doesn't do anything (since #6278 it seems)

Probably better to remove it instead or re-adding it back in, since it hasn't worked for a long time and no one has made an issue about it; also doesn't make as much sense anymore as a route as `rooms/latest` exists and returns a single room

notbakaneko:
closes #10491

Added some basic docs for `@queryParam` so I don't have to keep digging for it; `category` is omitted due to being deprecated.

### DIFF
--- a/app/Http/Controllers/Multiplayer/RoomsController.php
+++ b/app/Http/Controllers/Multiplayer/RoomsController.php
@@ -26,7 +26,7 @@ class RoomsController extends BaseController
      * ---
      *
      * @queryParam limit Maximum number of results. No-example
-     * @queryParam mode Filter mode; `all`, `ended`, `participated`, `owned`. Defaults to active rooms only if not sepcified. No-example
+     * @queryParam mode Filter mode; `active` (default), `all`, `ended`, `participated`, `owned`. No-example
      * @queryParam season_id Season ID to return Rooms from. No-example
      * @queryParam sort Sort order; `ended`, `created`. No-example
      * @queryParam type_group `playlists` (default) or `realtime`. No-example
--- a/app/Http/Controllers/Multiplayer/RoomsController.php
+++ b/app/Http/Controllers/Multiplayer/RoomsController.php
@@ -18,6 +18,19 @@ class RoomsController extends BaseController
         $this->middleware('require-scopes:public', ['only' => ['index', 'leaderboard', 'show']]);
     }
 
+    /**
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Multiplayer/RoomsController.php
```diff
@@ -18,6 +18,19 @@ public function __construct()
         $this->middleware('require-scopes:public', ['only' => ['index', 'leaderboard', 'show']]);
     }
 
+    /**
+     * Get Multiplayer Rooms
```
nanaya, line 22: need a `@category` specifier otherwise this goes to "Undocumented" 🤔

app/Http/Controllers/Multiplayer/RoomsController.php
```diff
@@ -18,6 +18,19 @@ public function __construct()
...
+     *
+     * Returns a list of multiplayer rooms.
+     *
+     * ---
+     *
+     * @queryParam limit Maximum number of results. No-example
+     * @queryParam mode Filter mode; `all`, `ended`, `participated`, `owned`. Defaults to active rooms only if not sepcified. No-example
```
nanaya, line 29: can just pretend `active` param for the default (and maybe add that as part of the switch)

nanaya (changes requested):

app/Http/Controllers/Multiplayer/RoomsController.php
notbakaneko: wait, that's why it was `default` in the first place 🙈

app/Http/Controllers/Multiplayer/RoomsController.php
```diff
@@ -18,6 +18,21 @@ public function __construct()
         $this->middleware('require-scopes:public', ['only' => ['index', 'leaderboard', 'show']]);
     }
 
+    /**
+     * Get Multiplayer Rooms
+     *
+     * Returns a list of multiplayer rooms.
+     *
+     * ---
```
nanaya, line 26: should only be here if there's response half of the docs otherwise this just adds a random line

nanaya (approved):

PR #10495, opened 2023-08-25
Add has_replay attribute in score json to match with database column

nanaya:
Forgot to PR this 👀

Less work for other components when implementing same json interface.


app/Transformers/ScoreTransformer.php
```diff
@@ -78,21 +78,25 @@ public function transformSolo(ScoreModel|SoloScore $score)
...
+
+        return [
+            ...$score->data->jsonSerialize(),
             'best_id' => $bestId ?? null,
+            'has_replay' => $hasReplay,
             'id' => $score->getKey(),
             'legacy_perfect' => $legacyPerfect ?? null,
             'pp' => $pp ?? null,
-            'replay' => $replay ?? false,
+            'replay' => $hasReplay,
```
peppy, line 97: Might be good to add a comment mentioning it can be removed in 6 months or so? That's our standard deprecation policy for lazer versions.

notbakaneko (approved):

PR #10500, opened 2023-08-28
Highlight new top rank (1-50) in event

### LINKED ISSUE #8877: Top 50 scores should be highlighted in recent feed as it was on the old site
I can't be the only one that liked when my top 50 score was standing out on my profile.
Of course it is not necessary and probably most of you guys don't care but I think it's cool 
before
opera_WnZp8kY3xT

nanaya:
Resolves #8877?

It's just a `<strong>` and no font resize. But the current legacy css doesn't have specific styling either and I can't find reference to the original `epicN` classes.

Also the bold is not quite visible 🤷

### DIFF
--- a/resources/js/profile-page/parse-event.tsx
+++ b/resources/js/profile-page/parse-event.tsx
@@ -1,11 +1,13 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
 // See the LICENCE file in the repository root for full licence text.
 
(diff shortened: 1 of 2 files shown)


peppy:
Would be good to have a screenshot in the OP for changes like this so I can give a quick assessment of how it looks visually.

peppy:
The bold really isn't that visible is it.. windows font rendering? Or just Torus?

notbakaneko:
Bold in torus is just kind of subtle probably because the normal weight is quite thick already 👀 
image
image

peppy (approved):

PR #10501, opened 2023-08-28
Fix beatmap discussion jump to anchor position

### LINKED ISSUE #10496: Beatmap Discussion modpost links dont properly link to modposts because of Markdown
On beatmaps discussions with a lot of images in the discussion the modpost links link to the right post, but then the images push down the discussion so that the user no longer is on the linked mod post

Example: https://osu.ppy.sh/beatmapsets/2007015/discussion/4174629/timeline#/3908168

This post is meant to link to this post 
image

but when loaded the images move it all the way to this post
image

This makes it really hard for users to link modposts on big discussions since its really hard to find the post that the link intended to go to

notbakaneko:
Something like this, maybe? 🤔 
Uses native `scrollIntoView` function ~~and `scroll-margin`~~; `instant` scroll avoids triggering lazy loaded images except when the image is close to the current view position. Chrome and Firefox should still handle it fine since they support scroll anchoring, Safari will still jump, not sure about Edge. The default of `auto` won't work because Safari just takes it to mean `smooth` even on load....

Updating `scroll-padding` instead of adding `scroll-margin` to the elements as it doesn't seem to use the correct position to offset of layout shift when images get loaded in.
Moved some of the element `padding` to `margin` otherwise Chrome still anchors to the wrong position 🤨 (also noticed some of the element structure could be updated to better handle it, but that's a problem for another time? 👀 )

fixes #10496 (unless you're Safari and doesn't scroll anchor)

### DIFF
--- a/resources/css/bem/beatmap-discussion-post.less
+++ b/resources/css/bem/beatmap-discussion-post.less
@@ -11,8 +11,8 @@
 
   @_content-gutter-base: 15px;
 
-  scroll-margin-top: var(--scroll-margin-top, 36px);
   transition: background-color .1s ease-in-out;
+  margin: 10px 0;
 
   &--deleted {
     opacity: 0.5;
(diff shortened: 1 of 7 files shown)


nanaya (changes requested):

nanaya (approved):
Without scroll animation the post highlighting may need some update but we'll worry about that later...

PR #10504, opened 2023-08-29
Move badge to the top

nanaya:
Readjust the user info bar colour so it doesn't turn into one big section with the badges.


notbakaneko (approved):
I hope this doesn't conflict with my other branch 👀

cl8n:
I don't have opinion on the actual design, but the reason I updated it how I did was to match closer to what flyte has on figma, and now this is deviating further away again. this change should be ported back to the figma doc if it's the preferred design.

also this makes another kind of "one big section" effect on bot profiles (https://osu.ppy.sh/users/3)

PR #10508, opened 2023-08-31
Update to webpack 5

notbakaneko:
Should be safe to move from webpack 4 since we're no longer blocked by the node version requirement.
Some `loaders` are being deprecated for `Asset Modules` but that will be a separate update.

Originally did this to support updating to TS5 which was to update DOM typings to fix scroll anchoring and jumping on beatmap discussions but that turns out to be a bit more complicated and expected, so might as well PR this first. 👀 

- [x] #10511

Because it now builds with the `development` export from `micromark` debugging can be enabled in browser by setting `localStorage.debug = 'micromark'` and reloading; `localStorage.debug = null` to disable. Performance will take a huge dip with debug enabled; the flag has no effect on `production` builds.


notbakaneko:
Apparently, webpack5 also correctly uses the dev versions of micromark for development builds where 4 didn't, and the links parsers should have been failing on assertions all this time ( ﾟ ヮﾟ)

nanaya (approved):

PR #10512, opened 2023-09-01
Add 2x image url to user badge

### LINKED ISSUE #10434: `UserBadge` from api doesn't return `@2x` image link string
https://osu.ppy.sh/docs/index.html#usercompact-userbadge

Similar to https://github.com/ppy/osu-web/issues/9816.

kj415j45:
Resolve #10434

~~I'm not sure if naming the new field as `image@2x` and duplicating `image_url` as `image` would be better or not.~~

### DIFF
--- a/app/Transformers/UserBadgeTransformer.php
+++ b/app/Transformers/UserBadgeTransformer.php
@@ -17,8 +17,8 @@ class UserBadgeTransformer extends TransformerAbstract
         return [
             'awarded_at' => json_time($badge->awarded),
             'description' => $badge->description,
-            'image_url' => $imageUrl,
             'image@2x_url' => $image2xUrl,
+            'image_url' => $imageUrl,
             'url' => $badge->url,
         ];
     }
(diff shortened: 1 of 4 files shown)


app/Transformers/UserBadgeTransformer.php
```diff
@@ -11,10 +11,14 @@ class UserBadgeTransformer extends TransformerAbstract
 {
     public function transform(UserBadge $badge)
     {
+        $image_url = $badge->imageUrl();
+        $image2x_url = retinaify($image_url);
```
nanaya, line 15: camel case variable name

resources/js/interfaces/user-badge-json.ts
```diff
@@ -5,5 +5,6 @@ export default interface UserBadgeJson {
   awarded_at: string;
   description: string;
   image_url: string;
+  'image@2x_url': string;
```
nanaya, line 8: as per lint check, this isn't sorted correctly

nanaya (changes requested):

nanaya (changes requested):
the code doesn't work

kj415j45:
🤦 I should use rename symbol next time.

nanaya (changes requested):
just noticed the docs in `user_compact.md` needs an update as well

nanaya (approved):

PR #10519, opened 2023-09-04
Add more information about test groups in documentation

### LINKED ISSUE #10466: Country change tests fail without active solo scores schema
https://github.com/ppy/osu-web/blob/38cb333fd0a988abc40c68a970e282fee79e7960/app/Libraries/User/CountryChange.php#L40-L44

I didn't want to add the tests to EsSoloScores group since they're testing unrelated things. all the tests pass if you comment out that command, which is what I've been doing for now

notbakaneko:
also rename the groups to `Requires{Thing}`, that seems to make more sense to me 🤷 

closes #10466

### DIFF
--- a/tests/Commands/EsIndexScoresQueueTest.php
+++ b/tests/Commands/EsIndexScoresQueueTest.php
@@ -48,6 +48,7 @@ class EsIndexScoresQueueTest extends TestCase
 
     /**
      * @dataProvider dataProviderForTestQueueScores
+     * @group EsSoloScores
      */
     public function testQueueScores(callable $setUp, array|callable $params, int $change): void
     {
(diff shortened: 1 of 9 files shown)


tests/Commands/EsIndexScoresQueueTest.php
```diff
@@ -48,6 +48,7 @@ public function testParameterValidity(array $params, bool $isValid)
 
     /**
      * @dataProvider dataProviderForTestQueueScores
+     * @group RequiresScoreIndexer
```
nanaya, line 51: this doesn't need score indexer as all it does is pushing to a different queue (`static::SCHEMA`)

nanaya (changes requested):

nanaya (approved):

PR #10523, opened 2023-09-06
Make ruleset id column on playlist items not nullable

nanaya:
It's already required on create.


peppy (approved):

peppy:
Probably fine to let this run as a blocking migration.

PR #10527, opened 2023-09-06
Remove unnecessary beatmap discussion wrapper

notbakaneko:
Remove creating an necessary layer (immediate `beatmap-discussion` child element already is own layer); the overflow also doesn't do anything anymore.
There's a different overflow/overlap problem which `overflow-x` here doesn't fix, anyway.

### DIFF
--- a/resources/js/beatmap-discussions/discussions.tsx
+++ b/resources/js/beatmap-discussions/discussions.tsx
@@ -172,18 +172,17 @@ export class Discussions extends React.Component<Props> {
     const parentDiscussion = discussion.parent_id != null ? this.props.currentDiscussions.byFilter.total.reviews[discussion.parent_id] : null;
 
     return (
-      <React.Fragment key={discussion.id}>
-        <Discussion
-          beatmapset={this.props.beatmapset}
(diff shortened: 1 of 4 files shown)


resources/js/beatmap-discussions/discussions.tsx
```diff
@@ -172,10 +172,7 @@ export class Discussions extends React.Component<Props> {
     const parentDiscussion = discussion.parent_id != null ? this.props.currentDiscussions.byFilter.total.reviews[discussion.parent_id] : null;
 
     return (
-      <div
-        key={discussion.id}
-        className={`${bn}__discussion`}
-      >
+      <React.Fragment key={discussion.id}>
```
nanaya, line 175: what's this for? (instead of just returning discussion component as is)

nanaya (changes requested):

resources/js/beatmap-discussions/discussions.tsx
notbakaneko: wait, I thought there were 2 elements there when I changed it despite saying it wasn't needed 🙈

PR #10530, opened 2023-09-07
Make entire quick search user card a link

### LINKED ISSUE #10317: Performing a middle/right click on player card on quick search does not work as expected
When you perform a middle/right click on player card on quick search, only while hovering over player name (or even profile picture and user title) does work as expected (for example, opening a page in new tab). The whole card outside of it doesn't seem to work in a similar way. Only left clicking works as expected. This is different from other cards like beatmaps.

Notice on video whether it shows a URL link on bottom left corner or not depending on where you hover.

https://github.com/ppy/osu-web/assets/75444413/c5b4943d-45c8-438e-a6c9-cef855e79b08

notbakaneko:
fixes #10317

### DIFF
--- a/resources/js/quick-search/user.tsx
+++ b/resources/js/quick-search/user.tsx
@@ -14,7 +14,7 @@ export default function User({ user, modifiers = [] }: { modifiers?: string[]; u
   const url = route('users.show', { user: user.id });
 
   return (
-    <div className={`${classWithModifiers('user-search-card', modifiers)}`}>
+    <div className={classWithModifiers('user-search-card', modifiers)}>
       <a className='user-search-card__background-container' href={url} />
       <div className='user-search-card__container'>
         <a className='user-search-card__avatar-container' href={url}>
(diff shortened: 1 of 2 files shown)


resources/js/quick-search/user.tsx
```diff
@@ -14,31 +14,34 @@ export default function User({ user, modifiers = [] }: { modifiers?: string[]; u
...
-      </a>
-
-      <div className='user-search-card__details'>
-        <div className='user-search-card__col  user-search-card__col--flag'>
-          <FlagCountry country={user.country} />
-        </div>
-
-        <a className='user-search-card__col user-search-card__col--username clickable-row-link' href={url}>
-          {user.username}
+    <div className={`${classWithModifiers('user-search-card', modifiers)}`}>
```
nanaya, line 17: string interpolation not needed anymore

nanaya (changes requested):

nanaya (approved):

PR #10542, opened 2023-09-11
Match object docs with TypeScript interface

nanaya:
One of these days we'll finally rename the transformers...


notbakaneko (changes requested):
info.md blade needs to be updated as well

nanaya:
I don't know about updating old entries so I removed the links instead 🤔

notbakaneko:
I think updating the link is fine since it's just a rename, not to mention the name itself is already changed so it's not like it's referring to the old one anymore?

notbakaneko (approved):

PR #10543, opened 2023-09-11
Remove react-autosize-textarea

notbakaneko:
`react-autosize-textarea` hasn't been updated in a while and doesn't have current versions of React in its dependencies.
It's just wrapping `autosize`, anyway, so we might as well use `autosize` directly - the component is just a convenience wrapper to attach and detach `autosize` and doesn't use `forwardRef`, making the implementation more straight-forward.

### DIFF
--- a/resources/js/chat/input-box.tsx
+++ b/resources/js/chat/input-box.tsx
@@ -4,6 +4,7 @@
 import { ChatMessageSendAction } from 'actions/chat-message-send-action';
 import { dispatch } from 'app-dispatcher';
 import BigButton from 'components/big-button';
+import TextareaAutosize from 'components/textarea-autosize';
 import { trim } from 'lodash';
 import { action, autorun, computed, makeObservable, reaction } from 'mobx';
 import { disposeOnUnmount, observer } from 'mobx-react';
@@ -11,7 +12,6 @@ import { maxMessageLength } from 'models/chat/channel';
 import Message from 'models/chat/message';
 import core from 'osu-core-singleton';
 import * as React from 'react';
-import TextareaAutosize from 'react-autosize-textarea';
 import { classWithModifiers } from 'utils/css';
 import { trans } from 'utils/lang';
 import { isModalShowing } from 'utils/modal-helper';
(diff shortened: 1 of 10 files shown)


resources/js/components/textarea-autosize.tsx
```diff
@@ -0,0 +1,69 @@
...
+}
+
+export default class TextareaAutosize extends React.Component<Props> {
+  static readonly defaultProps = {
+    async: false,
+    rows: 1,
+  };
+
+  private lineHeight?: number;
+  private ref = this.props.innerRef ?? React.createRef<HTMLTextAreaElement>();
```
nanaya, line 20: should this be a getter instead so the value doesn't potentially get outdated when the props change? (although probably practically impossible? also there are a bunch of other ref-related things that's being set once in componentDidMount and assumed being the same in componentDidUpdate)

resources/js/chat/input-box.tsx
```diff
@@ -109,10 +109,10 @@ export default class InputBox extends React.Component<Props> {
     return (
       <div className='chat-input'>
         <TextareaAutosize
-          ref={this.inputBoxRef}
           autoComplete='off'
           className={classWithModifiers('chat-input__box', { disabled: this.inputDisabled })}
           disabled={this.inputDisabled}
+          innerRef={this.inputBoxRef}
           maxLength={channel?.messageLengthLimit ?? maxMessageLength}
           maxRows={channel?.type === 'ANNOUNCE' ? 10 : 3}
```
nanaya, line 117: looks like the maximum rows option doesn't get applied correctly

nanaya (changes requested):

resources/js/components/textarea-autosize.tsx
notbakaneko: Refs aren't managed by React other than committing a value to it after render, so messing with a ref wouldn't normally be triggering renders anyway

resources/js/components/textarea-autosize.tsx
```diff
@@ -0,0 +1,79 @@
...
+    autosize.destroy(this.ref.current);
+  }
+
+  render() {
+    const { async, innerRef, maxRows, style, ...otherProps } = this.props;
+
+    const maxHeight = this.maxHeight;
+
+    return (
+      <textarea
```
nanaya, line 72: not sure exactly where but the styling seems to be slightly wrong

old


this pr


note the wrong background colour for the padding

<details>
<summary>it's slightly more <em>interesting</em> in chrome</summary>



</details>

nanaya (changes requested):

resources/js/components/textarea-autosize.tsx
notbakaneko: The old round uses `Math.round` before multiplying the rows, this one applies `Math.ciel` after multiplying; current `line-height` is `17.5px`

notbakaneko: white box appears to be a change in `autosize` between 4 and the current version where previously it applied `overflow: hidden scroll` but not it doesn't 🤔

nanaya (approved):

PR #10545, opened 2023-09-11
Add user table column to lock changing email

### LINKED ISSUE #10544: Add the ability to lock changing email of accounts
We need this for support usage. An internal flag can be added to `phpbb_user`, something like `lock_email_changes` which disallows the user from changing their email, instead showing a text message "Please contact [accounts](accounts@ppy.sh) if you need your email updated."

nanaya:
Resolves #10544.


resources/lang/en/accounts.php
```diff
@@ -18,6 +18,10 @@
             'new' => 'new email',
             'new_confirmation' => 'email confirmation',
             'title' => 'Email',
+            'locked' => [
+                '_' => 'Please contact :accounts if you need your email updated.',
+                'accounts' => 'osu! support',
```
Walavouchey, line 23: if it's `accounts@ppy.sh` as suggested in the linked issue then it should be this (see https://osu.ppy.sh/wiki/en/People/Account_support_team)

```suggestion
                'accounts' => 'the account support team',
```

Walavouchey, line 23: if it's `accounts@ppy.sh` as suggested in the linked issue then it should be this (see https://osu.ppy.sh/wiki/en/People/Account_support_team)

```suggestion
                '_' => 'Please contact the :accounts if you need your email updated.',
                'accounts' => 'account support team',
```

notbakaneko (approved):

PR #10548, opened 2023-09-12
Workaround beatmap discussions reply box having wrong background colour

### LINKED ISSUE #10536: Broken highlighting in beatmap discussion reply area
seems like it broke with the latest deploy

image

image

notbakaneko:
closes #10536

Will do for now since the reply area doesn't have a jump so it doesn't need to use `margin`

### DIFF
--- a/resources/css/bem/beatmap-discussion-post.less
+++ b/resources/css/bem/beatmap-discussion-post.less
@@ -30,6 +30,10 @@
   }
 
   &--new-reply {
+    // use padding because this section is a different background colour to the parent
+    // and fortunately doesn't need an anchor.
+    margin: 0;
+    padding: 10px 0;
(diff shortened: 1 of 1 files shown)


resources/css/bem/beatmap-discussion-post.less
```diff
@@ -30,6 +30,10 @@
   }
 
   &--new-reply {
+    // use padding because this section is a different background colour to the parent
+    // and fortunately doesn't need an anchor.
+    margin: 0;
+    padding: 10px 0;
```
nanaya, line 36: The bottom padding isn't supposed to be that big. Also the close button alignment seems wrong?



this is how it previously looked like

nanaya (changes requested):

nanaya (approved):

PR #10549, opened 2023-09-12
Fix sending multi-line message plain text chat error

notbakaneko:
Sending multi-line messages in channels only supporting plain text will error when markdown parsing assertions are enabled; they will still render fine on refresh


resources/js/remark-plugins/disable-constructs.ts
```diff
@@ -76,6 +76,7 @@ const disabled: Record<DisabledType, Construct[]> = {
     'autolink',
     'labelEnd',
     'labelStartLink',
+    'lineEnding',
```
nanaya, line 79: remove from "all disabled list" instead? If things are broken with this disabled

nanaya (changes requested):

nanaya (approved):

PR #10552, opened 2023-09-12
Add beatmapset search param for beatmapset events

### LINKED ISSUE #10071: Allow get beamapset events by beatmapset_id
For now, we can query beatmapset events by user id. I don't know why there's no support to use `beatmapset_id` as a query parameter

tsunyoku:
resolves #10071


app/Models/BeatmapsetEvent.php
```diff
@@ -92,6 +92,7 @@ public static function search($rawParams = [])
 
         $query = static::limit($params['limit'])->offset($pagination['offset']);
         $searchByUser = present($rawParams['user'] ?? null);
+        $searchByBeatmapset = present($rawParams['beatmapset'] ?? null);
```
cl8n, line 95: can just be in the `if` condition

app/Models/BeatmapsetEvent.php
```diff
@@ -106,6 +107,18 @@ public static function search($rawParams = [])
...
         }
 
+        if ($searchByBeatmapset) {
+            $params['beatmapset'] = $rawParams['beatmapset'];
+            $beatmapset = Beatmapset::find($params['beatmapset']);
+
+            if ($beatmapset === null) {
+                $query->none();
+            }
+            else {
```
cl8n, line 117:
```suggestion
            } else {
```
`phpcs` can help you fix minor code style issues like this

app/Models/BeatmapsetEvent.php
cl8n, line 95: also just something to note: when searching by user, certain event types related to moderation are filtered out (see later usage of `$searchByUser`). I'm not sure if the same should apply when searching by beatmapset. the behavior is kind of confusing to me considering everything is public when not directly searching by user...

cl8n (changes requested):

app/Models/BeatmapsetEvent.php
tsunyoku: sure, was trying to be consistent but i see they're that `$searchByUser` is used in more places than one

tsunyoku: i assume this comment is resolved by https://github.com/ppy/osu-web/pull/10552/commits/d5dac9355196499b094966daf4baf65691d26ec4? not sure what the relevance is otherwise, i'm not doing any filtering out..

app/Models/BeatmapsetEvent.php
```diff
@@ -106,6 +106,17 @@ public static function search($rawParams = [])
             }
         }
 
+        if (isset($rawParams['beatmapset'])) {
```
cl8n, line 109: it should probably still check for empty string as in `present($rawParams['beatmapset'] ?? null)` to match behavior of other param

app/Models/BeatmapsetEvent.php
cl8n: yeah your new code isn't filtering anything out, I'm saying that's what `$searchByUser` currently does, and so maybe searching by beatmapset also *should*. someone else needs to give their input on it, I don't fully understand the purpose

cl8n (approved):
need someone to look at https://github.com/ppy/osu-web/pull/10552#discussion_r1324937006 but otherwise :ok_hand:

app/Models/BeatmapsetEvent.php
nanaya: it's for some moderation events where the event itself is visible but user_id isn't public. Showing them when filtered by user id means leaking the user id itself.

app/Models/BeatmapsetEvent.php
```diff
@@ -106,6 +106,17 @@ public static function search($rawParams = [])
             }
         }
 
+        if (present($rawParams['beatmapset'] ?? null)) {
+            $params['beatmapset'] = $rawParams['beatmapset'];
```
nanaya, line 110: I'm pretty sure the parameter is called `beatmapset_id` everywhere else.

nanaya (changes requested):

app/Models/BeatmapsetEvent.php
```diff
@@ -106,6 +106,13 @@ public static function search($rawParams = [])
             }
         }
 
+        if (present($rawParams['beatmapset_id'] ?? null)) {
+            $params['beatmapset_id'] = $rawParams['beatmapset_id'];
+            $beatmapset = Beatmapset::find($params['beatmapset_id']);
+
+            $query->where('beatmapset_id', '=', $beatmapset->getKey());
```
nanaya, line 113:
```suggestion
            $query->where('beatmapset_id', '=', $params['beatmapset_id']);
```
otherwise it'll error when the beatmapset id doesn't exist. Also this way skips additional query.

nanaya (changes requested):

nanaya (approved):

PR #10555, opened 2023-09-13
Update to node 18

nanaya:
Fun stuff with deprecated md4 🤷


notbakaneko (changes requested):
`"css/app.css": "/assets/css/app.[sha512:contenthash:hex:8].css",` 🤔

notbakaneko (commented):
#10557 and we don't have to change the webpack output config (or can use `xxhash64` instead)

notbakaneko (approved):

PR #10557, opened 2023-09-14
Use webpack Asset Modules

notbakaneko:
Replaces the deprecated `file-loader` (and other loaders we don't use).
Also updates the other loaders. (Apparently older versions of loaders and copy plugin required `crypto` directly and only used providers from openssl directly instead of supported by webpack)

`[ext]` also now includes the `.` as part of the extension


nanaya (approved):
the file extension thing totally isn't confusing ಠ_ಠ

PR #10559, opened 2023-09-14
Use existing manifest plugin instead of a custom one

nanaya:
Seems to work fine now? At least it outputs more or less the same as the custom one.

Also ran yarn-deduplicate while at it.


notbakaneko (approved):
Looks like recent versions removed the thing which made me write the custom plugin in the first place 👀

PR #10561, opened 2023-09-15
Add type guard to string.present

notbakaneko:
Implicitly makes the return type of `presence` `string | null` which it should be instead of `string | null | undefined`

### DIFF
--- a/resources/js/utils/string.ts
+++ b/resources/js/utils/string.ts
@@ -5,6 +5,6 @@ export function presence(value?: string | null) {
   return present(value) ? value : null;
 }
 
-export function present(value?: string | null) {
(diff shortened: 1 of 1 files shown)


nanaya:
wouldn't this make the false condition incorrectly infer the variable as `null | undefined` when it may still be a string?

similar issue with microsoft/TypeScript#15048

notbakaneko:
oh right empty string is still there 🤔

notbakaneko:
Going to close this as the case where we need the type narrowing is rare and can be worked around when needed; adding a workaround here with the options currently available is a bit problematic since `''` itself is a subtype of `string` and typescript will complain about that instead...

PR #10564, opened 2023-09-16
Show badge `awarded_at` date in tooltip

### LINKED ISSUE #10541: Show badge `awarded_at` value in tooltip
a badge's award date is stored within every badge object, so i'm wondering why not actually show said date instead of only hiding it behind the api?

here's a mock-up design on how it could look

image

Swekka:
Closes #10541

### DIFF
--- a/resources/css/bem/profile-badges.less
+++ b/resources/css/bem/profile-badges.less
@@ -28,6 +28,7 @@
   }
 
   &__date {
-    color: @osu-colour-l1;
+    color: hsl(var(--hsl-l1));
+    margin-top: 5px
   }
 }
--- a/resources/css/bem/profile-badges.less
+++ b/resources/css/bem/profile-badges.less
@@ -26,4 +26,8 @@
     height: 40px;
     object-fit: contain;
   }
+
+  &__date {
+    color: @osu-colour-l1;
+  }
 }
--- a/resources/css/bem/profile-badges.less
+++ b/resources/css/bem/profile-badges.less
@@ -29,6 +29,6 @@
 
   &__date {
     color: hsl(var(--hsl-l1));
-    margin-top: 5px
+    margin-top: 5px;
   }
 }
--- a/resources/js/profile-page/badges.tsx
+++ b/resources/js/profile-page/badges.tsx
@@ -3,6 +3,7 @@
 
 import Img2x from 'components/img2x';
 import UserBadgeJson from 'interfaces/user-badge-json';
+import * as moment from 'moment';
 import * as React from 'react';
 import { Modifiers, classWithModifiers } from 'utils/css';
 import { present } from 'utils/string';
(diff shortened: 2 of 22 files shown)


resources/js/profile-page/badges.tsx
```diff
@@ -24,6 +24,7 @@ export default class Badges extends React.PureComponent<Props> {
           const img = (
             <Img2x
               className='profile-badges__badge'
+              data-date={badge.awarded_at}
```
notbakaneko, line 27: Format the description with the date here and pass as `data-html-title` instead of adding new functions to `TooltipDefault`

notbakaneko (changes requested):

resources/js/profile-page/badges.tsx
```diff
@@ -21,11 +22,16 @@ export default class Badges extends React.PureComponent<Props> {
     return (
       <div className={classWithModifiers('profile-badges', this.props.modifiers)}>
         {this.props.badges.map((badge) => {
+          const hasDate = present(badge.awarded_at);
+          const htmlTitle = hasDate ?`<div>${badge.description}</div>
+            <div class='tooltip-default__date'>${moment(badge.awarded_at).format('LL')}</div>` : null;
```
nanaya, line 27: element class (`tooltip-default__date`) can't be used when the block (`tooltip-default`) isn't in the same file/component.

Swekka: I'm not entirely sure what you mean, as far as I'm aware I just did what `TimeWithTooltip` did.
TimeWithTooltip:
image
This PR:
image

nanaya: the other one is used together in the same file where the tooltip-default is set.

Swekka: Oh okay, am I supposed to make a new class then?

notbakaneko: We use [BEM](https://getbem.com/naming/) conventions for css naming; the profile badge component uses `profile-badge` as its corresponding block name.

resources/js/profile-page/badges.tsx
```diff
@@ -21,11 +22,16 @@ export default class Badges extends React.PureComponent<Props> {
...
+          const htmlTitle = hasDate ? `<div>${badge.description}</div>
+            <div class='profile-badges__date'>${moment(badge.awarded_at).format('LL')}</div>` : null;
+
           const img = (
             <Img2x
               className='profile-badges__badge'
+              data-html-title={htmlTitle}
               src={badge.image_url}
-              title={badge.description}
+              title={hasDate ? '' : badge.description}
```
notbakaneko, line 34: just always show the description as title

resources/css/bem/profile-badges.less
```diff
@@ -26,4 +26,8 @@
     height: 40px;
     object-fit: contain;
   }
+
+  &__date {
```
notbakaneko, line 30: Could use a gap for the date like `margin-top: 5px`, otherwise it's too close
Screenshot 2023-09-21 at 20 11 22
image

resources/css/bem/profile-badges.less
```diff
@@ -26,4 +26,8 @@
...
+    color: @osu-colour-l1;
```
notbakaneko, line 31: use the css var version instead `hsl(var(--hsl-l1))`

The less variable `@osu-colour-*`versions are legacy variables

notbakaneko (changes requested):
I wonder if the display should be something like `Awarded on ${date}` like the achievements have but that can be a separate issue 🤔

notbakaneko (approved):

PR #10568, opened 2023-09-19
Simplify Webpack config

nanaya:
Somehow ended with this instead when attempting to enable es modules ಠ_ಠ thanks karma (I can shape it to make it somehow "work" but it looks amazingly ugly).

Seems to work here 🤷

- simplify config reading for karma
- remove explicitly passing config filename for webpack
- fix poll option
- combine `config` and `unmix`
- remove touching file after updating localisation
- remove extra logic for external file watching
  - the directory watch problem seems to be fixed?
- update postcss (leftover from trying esm)

protip: the original unmix part of the config is copied almost as-is with exception for postcss config update and the exporting part.


notbakaneko (approved):
should be ...safe 🤔

Issue #22, opened 2015-09-02
Files Not Found when running gulp


ITzNybble:
When running ./node_modules/.bin/gulp I am getting multiple file not found log entries:

[02:29:21] Merging: resources/assets/js/messages.js
[02:29:21] File not found: bower_components/jquery-ujs/src/rails.js
[02:29:21] File not found: bower_components/jquery-timeago/jquery.timeago.js
[02:29:21] File not found: bower_components/jquery-zoom/jquery.zoom.js
[02:29:21] File not found: bower_components/history.js/scripts/bundled-uncompres
sed/html5/jquery.history.js
[02:29:21] File not found: bower_components/ResponsiveSlides.js/responsiveslides
.js
[02:29:21] gulp-notify: [Error running notifier] Could not send message: not fou
nd: notify-send
[02:29:21] Finished 'scripts' after 250 ms
[02:29:21] Starting 'coffee'...
[02:29:21] Running CoffeeScript: resources/assets/coffee/osu!live.coffee, resour
ces/assets/coffee/osu_common.coffee, resources/assets/coffee/turbolinks-mod.coff
ee, resources/assets/coffee/bbcode.coffee, resources/assets/coffee/main.coffee,
resources/assets/coffee/store.coffee, resources/assets/coffee/forum.coffee, reso
urces/assets/coffee/forum/post-box.coffee, resources/assets/coffee/forum/topic-a
jax.coffee, resources/assets/coffee/ujs-common.coffee, resources/assets/coffee/b
ootstrap-modal.coffee, resources/assets/coffee/login-modal.coffee, resources/ass
ets/coffee/logout.coffee, resources/assets/coffee/shared.coffee
[02:29:22] gulp-notify: [Laravel Elixir] CoffeeScript Compiled!
[02:29:22] Finished 'coffee' after 671 ms
[02:29:22] Starting 'browserify'...
[02:29:22] Running Browserify: ./resources/assets/js/index.js
[02:29:22] Running Browserify: ./resources/assets/js/jsx/modding_react.jsx
[02:29:22] Running Browserify: ./resources/assets/js/jsx/profile_page.jsx
[02:29:22] gulp-notify: [Error in notifier] Error in plugin 'gulp-notify'
not found: notify-send
[02:29:25] Finished 'browserify' after 3.8 s
[02:29:25] Starting 'version'...
[02:29:25] Versioning: public/css/app.css, public/js/app.js, public/js/bundle.js
[...]

PR #10570, opened 2023-09-20
Add opengraph header to user profile

notbakaneko:
replaces #7518 since I can't push to that branch.

Not sure if `toMetaDescription` should accept an array of options or just the mode string directly until I look at the other pages 🤷 

Shows up in  `og:description` (and `description`, I guess) in `head` as something like `Rank (osu): Global #2 | Country #1` or `Rank (mania): Global #2 (4k #22) (7k -) | Country #1 (4k #1) (7k -)` for rulesets with variants. Appearance entirely dependent on whatever is displaying the card.


nanaya:
I wonder if the function should be put somewhere else. The User class is a bit too big.

Also using `ruleset` for new variable name instead of `mode` is probably good idea.

app/Models/User.php
```diff
@@ -1814,6 +1814,28 @@ public function titleUrl(): ?string
...
+    public function toMetaDescription(array $options = []): ?string
+    {
+        $ruleset = $options['ruleset'] ?? $this->playmode;
+        $stats = $this->statistics($ruleset);
+        if ($stats === null) {
+            return null;
+        }
+
+        $countryRank = $stats->countryRank();
+        $globalRank = $stats->globalRank();
```
nanaya, line 1826: I wonder if there should be something about variants

nanaya (changes requested):

app/Models/User.php
notbakaneko: I omitted the variants to keep the description simple.

Also, figuring out how to remove the `where` query from the transformer so it doesn't always do a separate query 🤔

nanaya: can probably just be updated to use the now-updated `statistics(...)` and preload as usual

notbakaneko: `UserStatisticsTransformer` doesn't use `User`, though and preloading a relation on `User` for the transformer to go through seems kind of obtuse? 🤔

nanaya: I mean preloaded through wherever it's being used. or in case of single user use the non query version of it so it'll be cached on next usage `$user->statistics('mania', variant: $variant)`

app/Http/Controllers/UsersController.php
```diff
@@ -629,6 +629,10 @@ public function show($id, $mode = null)
             abort(404);
         }
 
+        // preload and set relation for toMetaDescription and transformer sharing data
+        $stats = $user->statistics($currentMode);
+        $stats?->setRelation('user', $user);
+
```
nanaya, line 635: also can use similar thing for `me` action.

nanaya:
I wonder if for unranked case it should just say something like `Unranked` instead of `Global - | Country -` (or even worse `Global - (4k -, 7k -) | Country - (4k -, 7k -)`). And in case of unranked variant, just don't display it at all maybe? Something like `Global #10 (7k #3) | ...`

notbakaneko:
Yeah, I dunno, the `-` is just consistent with how it displays in other places, although in the case of variants it doesn't show 🤔

app/Models/User.php
```diff
@@ -1814,6 +1814,39 @@ public function titleUrl(): ?string
...
+            ]);
+
+            $variants = Beatmap::VARIANTS[$ruleset] ?? [];
+
+            $variantsTexts = null;
+            foreach ($variants as $variant) {
+                $variantStats = $this->statistics($ruleset, false, $variant);
+                if ($variantStats !== null) {
+                    $variantsTexts[] = $variant.' '.format_rank($variantStats->$method());
+                }
```
nanaya, line 1839: I'll need to check again but there may be case where the variant entry exists but not ranked.

nanaya (approved):

PR #10571, opened 2023-09-20
Combine lookup username array generation logic

nanaya:
Should've done this in the other pr but forgot to actually do it 👀

Also found another place it can be used.

### DIFF
--- a/app/Libraries/User/UsernamesForDbLookup.php
+++ b/app/Libraries/User/UsernamesForDbLookup.php
@@ -1,0 +1,30 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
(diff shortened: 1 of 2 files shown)


app/Libraries/User/UsernamesForDbLookup.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    public static function make($username, bool $trimPrefix = true): array
+    {
+        $searchUsername = presence(get_string($username));
+
+        if ($searchUsername === null) {
+            return [];
+        }
+
+        if ($trimPrefix && $searchUsername[0] === '@') {
```
cl8n, line 20: looking at usages of `findByUsernameForInactive()`, I don't think `$trimPrefix` option is necessary

cl8n (changes requested):

notbakaneko (approved):

PR #10575, opened 2023-09-21
Fix User mania 7k statistics relation

notbakaneko:
🤦


nanaya (approved):
oh whoops 👀

nanaya:
@peppy mania7k table need some fixing on its country column 👀

PR #10582, opened 2023-09-22
Add missing semicolons and spaces in CSS

Swekka:


cl8n:
docs.less should probably not be changed here because it's mostly copied from Scribe -- keeping the original formatting makes comparing changes easier

this also reminds me I have an old branch with a basic less linter setup :eyes:

cl8n (approved):

Swekka:
Replaced by #10592

PR #10584, opened 2023-09-22
Remove support for legacy sessions

cl8n:
because you can't log into old site anymore

I don't know if this migration will break something on old site so maybe ignore it.


database/migrations/2023_09_22_194226_drop_phpbb_sessions.php
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        migration('2016_08_16_101048_create_sessions_table')->up();
+
+        // Relevant part of 2020_05_15_083037_sync_structure
+        Schema::table('phpbb_sessions', function (Blueprint $table) {
```
peppy, line 30: Yeah I don't think this is required.

nanaya: I'd rather keep the table for now and test renaming the table separately to see what breaks.

nanaya (changes requested):

nanaya (approved):

PR #10585, opened 2023-09-23
Misc. UserGroup clean-up

cl8n:
applying some of what I was thinking while working around usergroups / group permissions / group history / etc  for a while. I thought this would be good to get out of the way before (eventually) looking into expanding group permissions over the API for some project loved things.

imo the main problem with a lot of group checking code is that it factors in the *current request's* oauth token, so then these methods are only really valid when called on the current user, and as the developer you have to be mindful of which methods concern authorization/permissions/etc and which just get the data they suggest. while not solved in this PR, I consolidated the various kinds of group checks so that in most cases it will be clear when things are intentionally bypassing the contained oauth check

some other time I will look into getting that oauth token check out of the User model. but this was already a rabbit hole so I'm just PRing what I did so far


tests/Models/OAuth/GroupPermissionTest.php
```diff
@@ -3,56 +3,46 @@
...
 
-    public function groupsDataProvider()
+    public function booleanDataProvider(): array
     {
-        return [
-            ['admin', 'isAdmin', false],
-            ['alumni', 'isAlumni', false],
-            ['announce', 'isChatAnnouncer', true],
-            ['bng', 'isBNG', false],
-            ['bot', 'isBot', true],
```
nanaya, line 51: the test for this is missing

nanaya (changes requested):

app/Models/User.php
```diff
@@ -1118,49 +1076,62 @@ public function defaultGroup(): Group
...
+                ->all();
         });
     }
 
-
-    public function findUserGroup($group, bool $activeOnly): ?UserGroup
+    /**
+     * Get the user's usergroup corresponding to a specified group.
+     */
+    public function findUserGroup(Group|string $group, bool $includePending = false): ?UserGroup
```
notbakaneko, line 1098: I'm not keen on this type of param type overloading and having conditionals in the body to convert it when the specific type is known before hand; this isn't a parsing function.

It also means `findUserGroup` implicitly picks up the auto group creating behaviour of `Groups::byIdentifier` as an unexpected side-effect.

PR #10590, opened 2023-09-24
Update mod definitions

peppy:
This PR has been auto-generated to update the mod definitions to match ppy/osu@2023.924.1.


nanaya:
when mod settings are removed like this here, what should be displayed for existing score with the settings?

peppy:
I think they can safely be not shown. Any plays with the setting will have it forced going forward.

nanaya (approved):

PR #10592, opened 2023-09-24
Add linter for Less

cl8n:
mostly so things can be fixed automatically and you don't have to care about formatting so much. the enabled rules that *can't* be fixed automatically seem like good suggestions and already had very few errors in this project

- will conflict with  #10568 because postcss dependency
- I haven't tested the formatting for GitHub action
- some of the rules should probably be warnings instead of errors


nanaya:
I don't think stylelint-stylistic is supposed to be used for new projects.

cl8n:
I looked into stylelint's recommended migration path for their stylistic rules (tl;dr try Prettier) -- it doesn't really align with the goal I had here of being non-intrusive. I wanted the result to be similar to osu-web's ESLint config in that you can still make some formatting decisions yourself where it makes sense, it just lints the tedious things that we agree should *always* be done a certain way

so I went with their next recommendation which is to depend on someone else maintaining similar rules, and this is the most popular option.

nanaya:
prettier can be configured to only handle css files

I took a quick look of its output and it seems fine (apart of mod names need quoting)

as for customization, I think we would care more about customizing actual lint than formatting

nanaya:
The formatter part is now done using prettier (#11355). The linting part would need some more considerations. Closing this one for now and feel free to make another pr if you want to add stylelint.

PR #10597, opened 2023-09-26
Show rank changes on performance ranking

### LINKED ISSUE #7391: Show changes in rank on leaderboards
Hi there. The Issue I'll be proposing is one I've been thinking of lately. Across the gamemodes, leaderboards are constantly changing. In some, they have a certain country dominating in the rankings like japan in taiko for example or standard where the top 100 is having new players every week. As a solution not only for leaderboards but for profiles I suggest the following:
Implement an arrow system. Something like FIFA's performance arrows. This arrows would be used both in profiles and leaderboards to reflect a player's growth in the past 7 days globally. When entering to their profile it would also show their growth countrywise. This would help not only tournament players to get more info faster but also players who just like looking at new players in the leaderboards making it a lot easier to grasp information. I made 2 examples out of taiko with wikipedia's arrows. These can be changed. I had 3 type of arrows:
Neutral
Positive
Negative
Neutral can have a +0 or just nothing that's up to whoever decides to implement this feature. 
Last but not least, It would probably look better if the arrows were placed next to the rank in the same column to make it more aesthetic. However, a small place should be added to fit it.
GITHUB TEMPLATE PROFILE
GITHUB TEMPLATE

cl8n:
resolves #7391

---

rank-change-demo


cl8n:
I need some help with caching this per request because currently it gets called 50 times on a ranking page. I don't know the best way to do it
https://github.com/ppy/osu-web/blob/69bc11a2180ac88064087372e7b219b5aab59a25/app/Models/RankHistory.php#L123

nanaya:
is it normal to have "least recent" ranking to compare against when the others are 30 days? Isn't it usually just `-` to indicate new ranker?

cl8n:
> is it normal to have "least recent" ranking to compare against when the others are 30 days? Isn't it usually just `-` to indicate new ranker?

yeah I guess this could be misleading in some way. I've changed it so that only the history at exactly 30 days is checked, and everything has an icon of some form now. also further tooltip to explain the "not enough history" case

resources/css/bem/ranking-page-table.less
```diff
@@ -51,6 +51,28 @@
       opacity: 1;
     }
 
+    each({
+      down: @fa-var-arrow-down, hsl(var(--hsl-red-1));
+      none: @fa-var-equals, hsl(0, 0%, var(--c-lightness-1));
+      pending: @fa-var-ellipsis-h, hsl(0, 0%, var(--c-lightness-1));
```
nanaya, line 57: `--hsl-f1`? That above is barely any different to just using `#bbb` directly

nanaya (changes requested):

nanaya (approved):

PR #10598, opened 2023-09-26
 Reduce user last visit recording frequency to 5 minutes 

nanaya:
Also skipped converting the times to object and instead work directly on the ints as it's currently stored as unix timestamp.


notbakaneko (approved):
I was thinking more along the lines of only slowing down the indexing rate to 5 minutes but that'd be more complicated (and a bit weird with the times not matching, I guess 🤔 )

notbakaneko:
backlink to the reason `user_lastvisit` gets updated in ES #7185

PR #10600, opened 2023-09-27
Stringify beatmapsets json for caching

nanaya:
Wasn't originally done as it may break if the json structure is changed but nowadays we do care about backward compat so it should be fine? 🤷


notbakaneko (commented):
what's the performance difference here? I can't really benchmark it atm as there's something strange going on with my setup where it returns the same time for everything 👀

nanaya:
Just the specific bit (single core/thread):
- string: 4854 ops/s
- array: 1773 ops/s

I couldn't get consistent from wrk and I'm noticing a memory leak with memory usage getting higher and the performance getting lower over time if I set the max requests threshold to something high ಠ_ಠ

30k max requests, 6 connections, 6 threads, 20s x3:
- string: 2931, 1304, 1059 req/s
- array: 2838, 820, 818 req/s

peppy (approved):

PR #10608, opened 2023-09-29
Add option to hide review posts

### LINKED ISSUE #10606: Beatmap review posts duplicated in recent discussions page
As can be seen in the image below. Since individual posts are accessible through links in the review post, I don't see the point of duplicating them under the review post. This makes recent discussions page annoying to scroll sometimes if the review post is long.
Clipboard_20230928_170309

nanaya:
Active by default.

Resolves #10606.


resources/lang/en/beatmap_discussions.php
```diff
@@ -26,6 +26,7 @@
             'deleted' => 'Include deleted discussions',
             'mode' => 'Beatmap mode',
             'only_unresolved' => 'Show only unresolved discussions',
+            'show_review_embeds' => 'Show review posts',
```
notbakaneko, line 29: Is it posts or embeds? `Review posts` also implies the actual review post, not the posts in it 🤔

notbakaneko (changes requested):

resources/lang/en/beatmap_discussions.php
nanaya: yeah but I don't know if embed means anything to people

notbakaneko (approved):

PR #10614, opened 2023-10-02
Fix markdown wiki link locale and case

### LINKED ISSUE #10613: Wiki links in chat always use English locale
- what I posted was: `[[ru/Rules]]`
- the link displayed by the website was: `https://osu.ppy.sh/wiki/en/ru/rules` (case normalization may be ignored, I get that it probably saves a redirect, and that is good)

feels like this is a recent regression, and `en` in the URL is not tied to my wiki locale (tried changing it, to no effect)

image

nanaya:
Resolves #10613.


notbakaneko (approved):
I feel like the default behaviour should be to avoid the redirect, but oh well 🤷

PR #10621, opened 2023-10-06
Fix page scrolling when typing into new discussions text box 

### LINKED ISSUE #10578: Typing or selecting text causes beatmap discussion page to scroll up
I'm having a difficult time describing the issue, so here's a video:

https://github.com/ppy/osu-web/assets/117247146/04c41047-fb8a-4239-9315-52ddd1052428

Basically, if you go to a random discussion point from a link, it anchors you to that point as addressed in https://github.com/ppy/osu-web/issues/10496. If you try to make a new suggestion, typing will gradually bring you up the page which I'm sure is unintended (albeit a little funny as we've had a few laughs in the mania BN channel). Regardless, it does make referencing posts kind of difficult with the unintended scrolling, so just wanted to point it out in the hopes something can be done about it. I'm guessing this arose after https://github.com/ppy/osu-web/pull/10501 which had been intended to fix the anchoring issue due to images loading with markdown.

notbakaneko:
`position: sticky` and `scroll-padding` don't quite behave well together.
Typing into the text area or selecting text constantly triggers scrolling into view and the combination of `position: sticky` and `scroll-padding` makes the entire page scroll; this doesn't happen if the total `scroll-padding-top` is less vertical position where it's focusing into view (which it normally isn't). 

closes #10578

### DIFF
--- a/resources/js/beatmap-discussions/main.coffee
+++ b/resources/js/beatmap-discussions/main.coffee
@@ -359,6 +359,7 @@ export class Main extends React.PureComponent
 
 
   handleNewDiscussionFocus: =>
+    # Bug with position: sticky and scroll-padding: https://bugs.chromium.org/p/chromium/issues/detail?id=1466472
     document.documentElement.style.removeProperty '--scroll-padding-top-extra'
 
 
--- a/resources/js/beatmap-discussions/main.coffee
+++ b/resources/js/beatmap-discussions/main.coffee
@@ -164,6 +164,7 @@ export class Main extends React.PureComponent
                   currentDiscussions: @currentDiscussions()
                   innerRef: @newDiscussionRef
                   mode: @state.currentMode
+                  onFocus: @handleNewDiscussionFocus
                   pinned: @state.pinnedNewDiscussion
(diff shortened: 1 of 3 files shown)


resources/js/beatmap-discussions/main.coffee
```diff
@@ -357,6 +358,10 @@ export class Main extends React.PureComponent
     @cache.groupedBeatmaps ?= BeatmapHelper.group _.values(@beatmaps())
 
 
+  handleNewDiscussionFocus: =>
+    document.documentElement.style.removeProperty '--scroll-padding-top-extra'
```
nanaya, line 362: can use a note it's for [this chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1466472)

nanaya (changes requested):

resources/js/beatmap-discussions/main.coffee
notbakaneko: actually forgot to put a comment after figuring out the cutoffs where the scrolling starts 👀

nanaya (approved):

PR #10627, opened 2023-10-10
Refactor password reset backend

nanaya:
Move out password reset from user session so requests for same user use same password reset data.

Laravel is updated to get access to the new mail queue count assertion.


notbakaneko (dismissed):

app/Http/Controllers/PasswordResetController.php
```diff
@@ -24,138 +28,112 @@ public function __construct()
         $this->middleware('throttle:20,1440,password-reset:');
```
peppy, line 28: kinda not related, but i noticed this seems quite high. even in the case of a proxy it could probably be lowered considerably..

notbakaneko (approved):

app/Http/Controllers/PasswordResetController.php
peppy: I guess it's okay. Just wondering if it would make sense to make it something like 2 per 2 hours instead, to reduce the ability to spam a heap at once.

PR #10628, opened 2023-10-10
Limit password verification code issue attempts for user

notbakaneko:
A simple one for now, 5 per hour, doesn't reset after entering the correct verification code nor does it handle how to contact support or anything if you've hit the limit. Also there appears to be duplicate flash popups on the verification code page...?

not sure if key should be `password_reset` or `password-reset` (like the throttle key) 🤔


notbakaneko:
use #10627 instead

PR #10630, opened 2023-10-11
Replace multiplayer ruleset const with enum

nanaya:
Just replacing the const with enum one to one so there's not much benefit in term of actual enum usage yet.

Also I don't know what to do with catch/fruits name discrepancy 🤷 (probably just rename internally but keep the api output)


notbakaneko:
> catch/taiko name discrepancy

did you mean fruits....❓

notbakaneko (approved):

PR #10631, opened 2023-10-11
Improve UI performance of long chat channel lists

notbakaneko:
Improve the performance of long channel lists especially on Safari; slight improvement on Chrome but nowhere near the same extent as Safari.

The channel list is broken down into channel chunks so that each chunk can be it's own render layer; 100 (arbitrary) seems to be a reasonable size - too big and Safari's performance takes a dive, too small creates too many layers...and Safari's performance also takes a steep dive. (edit: too many layers causes too many layout recalculations even it their content doesn't change, and layout is relatively expensive compared to re-rendering)

React-keyed elements ending up in different chunks when the list changes shouldn't be an issue since the list usually gets rearranged at that point.


resources/js/chat/conversation-list.tsx
```diff
@@ -21,14 +22,20 @@ function renderChannels(type: SupportedChannelType) {
...
+
     return (
       <React.Fragment key={type}>
         <div className='chat-conversation-list__group'>
           <div className='chat-conversation-list__header'>
             <span className='chat-conversation-list__header-text'>{title}</span>
             <span className='chat-conversation-list__header-icon' title={title}><i className={icons[type]} /></span>
           </div>
-          {channels.map((channel) => <ConversationListItem key={channel.channelId} channel={channel} />)}
+          {chunks.map((c, index) => (
```
peppy, line 34: Might be worth adding a comment somewhere in here mentioning that this is an optimisation.

nanaya (changes requested):
this kinda breaks mobile

nanaya (approved):

PR #10634, opened 2023-10-12
Move some of score link function to score token

nanaya:
Renames (solo prefix removal, score link rename to something else, etc) will be done later.

The json returned by score token creation is changed (from ScoreTransformer to ScoreTokenTransformer) but doesn't seem to affect anything client side.

- [x] #10635
- [x] update existing `multiplayer_scores_high.score_id` and `multiplayer_rooms_high.last_score_id`
- [x] add `playlist_item_id` (and relevant index) to `solo_score_tokens`
  - column: `` `playlist_item_id` bigint unsigned DEFAULT NULL AFTER ruleset_id ``
  - index: `` KEY `solo_score_tokens_user_id_playlist_item_id_index` (`user_id`,`playlist_item_id`) ``
  - migration entry: `2023_10_12_071549_add_playlist_item_id_to_score_tokens`

The playlist item attempts thing will be done later since the change is quite involved and the current one isn't too bad anyway.


peppy:
I'm running the migrations for this one. May take a day or to due to `solo_score_tokens` never being cleaned up to date. I'll probably look to add partition rotation as part of this.

peppy:
Migrations run and marked run.

app/Models/Multiplayer/Room.php
```diff
@@ -688,7 +692,8 @@ private function assertValidStartPlay(User $user, PlaylistItem $playlistItem)
         }
 
         if ($playlistItem->max_attempts !== null) {
-            $playlistAttempts = $playlistItem->scoreLinks()->where('user_id', $user->getKey())->count();
+            $playlistAttempts = $playlistItem->scoreTokens()->where('user_id', $user->getKey())->whereNull('score_id')->count()
```
notbakaneko, line 695: What happens when score tokens finally get purged? Or will there be something that keeps score tokens for playlist items that are still playable? @peppy

peppy: I'm not yet sure. Arguably we could keep score tokens around long enough that this didn't matter, but that could be months worth..

Shall we open a separate issue to track this? Might need a new table storage if we want this functionality.

nanaya: I'll add the count to scores_high table later similar to current rooms_high (and make sure it doesn't get reset etc)

notbakaneko (approved):
seems fine 👀

PR #10646, opened 2023-10-17
Fix permission for local storage upload

nanaya:
Theoretically this should also apply to s3 but it seems fine over there? 🤷


notbakaneko:
is there something that needs to happen first, I never noticed any problems? 🤔

nanaya:
the files and directories are created with 600/700 permission which nothing but the current user can read. It's only noticeable if the files are served by something ran by different user.

notbakaneko (approved):

PR #10650, opened 2023-10-18
Support annotations to handle cursor_string parameter for apidocs

notbakaneko:
Standardize the `cursor_string` description instead of repeating the description text everywhere; some other params could probably be standardized as well.
The down side is they are all ordered at the bottom at the moment 🤔


app/Docs/Strategies/UsesCursor.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
nanaya, line 1: missing license

nanaya (approved):

PR #10652, opened 2023-10-18
Use user_id for profile customisation model primary key

nanaya:
I don't remember why it has its own primary key ಠ_ಠ

I'll drop the column in a different PR since this needs to be deployed first.

On unrelated note, considering how frequent this model is needed just for the cover, I'm thinking of moving the cover column over to user model. It doesn't need to be a full json column, a normal varchar is sufficient.


notbakaneko (approved):
I'm guessing "auto primary key all the models" was the in-thing to do at the time 👀

PR #10639, opened 2023-10-16
Add more Opengraph headers to more pages

notbakaneko:
Add more opengraph headers for news, featured artists, beatmapsets, forums, contests.
Moves the previous `pageDescription` for `show` pages into opengraph info since they're always the same.
More detailed headers can be added separately.

The interface isn't really necessary at the moment other than to make linter/style checker/validators complain if not implemented 🤷 

closes #1295
closes #7652

PR #10655, opened 2023-10-19
Extract opengraph methods

notbakaneko:
- [x] directly continues #10639
- moves the opengraph methods to their own helper classes instead of filling up the existing model files
- no need to pass the variable into the view factory since only the metadata blade cares about it
- wiki page parser now has a thing that pulls out first paragraph block as the excerpt (maybe news can use it too?)

Also I did think of removing the `Opengraph` suffixes from the class names, except, they're usually used in places where there'd already be an existing model class import with the same name, so they'd have to be aliased to `as xxOpengraph` anyway 👀 

Not sure what to with the user multiplayer tabs descriptions as the room/playlist counts seems...not useful 🤔


notbakaneko:
Adding per beatmap opengraph on the beatmapset url isn't currently possible as the beatmap portion of the url is part of the fragment and client-side

app/helpers.php
```diff
@@ -1690,6 +1691,15 @@ function seeded_shuffle(array &$items, int $seed = 0)
     mt_srand();
 }
 
+function set_opengraph(HasOpengraph $model, ...$options)
+{
+    $className = 'App\\Libraries\\Opengraph\\'.get_class_basename(get_class($model)).'Opengraph';
+
+    View::share([
+        'opengraph' => (new $className($model, ...$options))->get(),
+    ]);
```
nanaya, line 1700: this needs to be reset after each request because the value persists to the next requests until overwritten

nanaya (changes requested):

app/Libraries/Opengraph/CommentOpengraph.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    public function __construct(private Comment $comment)
+    {
+    }
+
+    public function get(): array
+    {
+        $user = $this->comment->user;
+
+        return priv_check('CommentShow', $this->comment)->can()
```
nanaya, line 22: priv_check_user with null user?

app/Libraries/Opengraph/CommentOpengraph.php
```diff
@@ -0,0 +1,29 @@
...
+            ? [
+                'description' => html_excerpt($this->comment->message_html, 100),
```
nanaya, line 24: `blade_safe`? return of `html_excerpt` is already escaped

although I noticed discord double unescape description by default 🤔 and twitter is even worse

app/Libraries/Opengraph/Wiki/PageOpengraph.php
```diff
@@ -0,0 +1,31 @@
...
+    public function get(): array
+    {
+        if (!$this->page->isVisible()) {
+            return [];
+        }
+
+        $pageData = $this->page->get();
+
+        return [
+            'description' => $pageData['excerpt'] ?? html_excerpt($pageData['output']), // html_excerpt fallback
```
nanaya, line 27: another one needing blade_safe

app/Libraries/Markdown/Osu/DocumentProcessor.php
```diff
@@ -263,6 +269,15 @@ private function proxyImage()
         }
     }
 
+    private function recordExcerpt(): void
+    {
+        if ($this->excerpt !== null || !$this->node instanceof Paragraph || !$this->event->isEntering()) {
```
nanaya, line 274: with first two paragraphs may not even contain actual content I wonder if it's better to just show first 100 characters instead.

nanaya (changes requested):

app/Libraries/Opengraph/CommentOpengraph.php
notbakaneko: originally, I had everything assume opengrapgh would be viewed only by `null` user, but if you go to the page with a deleted comment as a moderator, you see a different thing, so I'm not so sure now 👀

resources/views/layout/metadata.blade.php
```diff
@@ -16,23 +21,33 @@
...
+        @php
+            if ($key === 'title') {
+                $isHtmlString = $value instanceof HtmlString;
+                $value .= ' · '.page_title();
+                if ($isHtmlString) {
+                    $value = new HtmlString($value);
+                }
+            }
+        @endphp
+        <meta property="og:{{ $key }}" content="{{ $value }}">
```
nanaya, line 48: a bit more repetitive but a lot shorter. (also the `page_title()` should be escaped)

```suggestion
        @if ($key === 'title')
            <meta property="og:{{ $key }}" content="{{ $value }} · {{ page_title() }}">
        @else
            <meta property="og:{{ $key }}" content="{{ $value }}">
        @endif
```

app/Libraries/Opengraph/HasOpengraph.php
```diff
@@ -0,0 +1,12 @@
...
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+declare(strict_types=1);
+
+namespace App\Libraries\Opengraph;
+
+interface HasOpengraph
+{
```
nanaya, line 11: I wonder if this should require something like `$opengraphClass` and use it for `set_opengraph`. Not sure if helps anything at all though. And this currently isn't enforcing anything. One can have HasOpengraph model without corresponding class and it'll only explode on runtime. That said, specifying the class name won't help either since it's not checked and can't specify inheritance either.

eh 🤷

nanaya (changes requested):

app/Libraries/Opengraph/BeatmapsetOpengraph.php
```diff
@@ -0,0 +1,29 @@
...
+class BeatmapsetOpengraph
+{
+    public function __construct(private Beatmapset $beatmapset)
+    {
+    }
+
+    public function get(): array
+    {
+        $section = osu_trans('layout.menu.beatmaps._');
+        $title = "{$this->beatmapset->artist} - {$this->beatmapset->title}"; // opengrah header always intended for guest.
```
nanaya, line 21: oh right, opengraph missing p

app/Libraries/Opengraph/HasOpengraph.php
notbakaneko: I got rid of it since the type check is more of "belongs to `App\Models`" than anything 🤷

nanaya (approved):

PR #10659, opened 2023-10-20
Store config in global variable

nanaya:
Noticed when trying to clean up storage a bit more, the main reason for caching the storage class (or creating singleton) is to cache config value. And thus I went on trying to just not use the laravel's config thing.

Apparently this shoves off about 3ms on scores endpoint `http://127.0.0.1:8000/beatmaps/397534/scores?mode=osu&type=global` (from ~70ms down to ~67ms), and about 6ms on user profile page (75 &rarr; 69).

That said, the number varies a lot because octane so I can't really tell how much it actually improves. But at least I don't need to worry about caching config value, so, yay? 🤷

Raw number just for retrieving config:

```
> $st = hrtime(true); $n = 1_000_000; for ($i=0;$i<$n;$i++){config('filesystems.default');}echo ((hrtime(true)-$st)/1000000000).PHP_EOL;
9.041199092
> $st = hrtime(true); $n = 1_000_000; for ($i=0;$i<$n;$i++){$GLOBALS['cfg']['filesystems']['default'];}echo ((hrtime(true)-$st)/1000000000).PHP_EOL;
0.047803389
```

replacer script (only for the most basic format `config('...')`, the rest are done by hand):
```php
<?php

$files = explode("\n", trim(file_get_contents('./files.txt')));

foreach ($files as $file) {
    $content = file_get_contents($file);
    $replacedContent = preg_replace_callback("/config\('([^)']+)'\)/", function ($matches) {
        $str = strtr($matches[1], ['.' => "']['"]);

        return '$GLOBALS[\'cfg\'][\''.$str.'\']';
    }, $content);

    file_put_contents($file, $replacedContent);
}
```

Extra changes:
- `osu_url` and related config has been flattened
- download url uses `osu_url` function


notbakaneko (changes requested):
conflicts

notbakaneko:
```
> $st = hrtime(true); $n = 1_000_000; for ($i=0;$i<$n;$i++){config('filesystems.default');}echo ((hrtime(true)-$st)/1000000000).PHP_EOL;
0.936425944

> $st = hrtime(true); $n = 1_000_000; for ($i=0;$i<$n;$i++){Config::get('filesystems.default');}echo ((hrtime(true)-$st)/1000000000).PHP_EOL;
0.513163456

> $st = hrtime(true); $n = 1_000_000; for ($i=0;$i<$n;$i++){$GLOBALS['cfg']['filesystems']['default'];}echo ((hrtime(true)-$st)/1000000000).PHP_EOL;
0.030464694
```

🤔

nanaya:
oh huh apparently I have xdebug loaded and set to develop mode (default). Hmm 🤔

nanaya:
I guess the impact is way less than I initially thought. The difference is still there but generating 300k urls/s is still good enough maybe? (the globals one is 1.7M urls/s)

```
> $st = hrtime(true); $n = 1_000_000; for ($i=0;$i<$n;$i++){App\Libraries\StorageUrl::makeGl(null, 'hello/world.jpg');}echo ((hrtime(true)-$st)/1000000000).PHP_EOL;
0.133792562
> $st = hrtime(true); $n = 1_000_000; for ($i=0;$i<$n;$i++){App\Libraries\StorageUrl::makeFn(null, 'hello/world.jpg');}echo ((hrtime(true)-$st)/1000000000).PHP_EOL;
3.573589783
```

nanaya:
config isn't supposed to be immutable... (except inside test)

notbakaneko (dismissed):
forgot about this 👀 
Should deploy separately so we can measure its effect independent of other PRs

notbakaneko (approved):

Issue #6840, opened 2020-10-31
Error on website, only one of the languages is not uppercase on first, while all the rest yes. Please correct this. :)


ErrorFinder-osu:
Only happens on Spanish/Español.
![Screenshot_2020-10-31 dashboard osu ](https://user-images.githubusercontent.com/73762816/97791863-d4e8f700-1bb5-11eb-86cc-201b3bccb099.png)

qauland:
According to #6241,

> Language names in Spanish are common nouns (_nombres comunes_) and the first letter shouldn't be capitalised.

bdach:
While this issue is a huge nitpick, if each language's capitalisation standards are to be used, they should probably be enforced in all cases. As far as I can tell there are some other language names on that list that do not abide by their language's capitalisation rules - from the ones I know without having to google, I'm pretty sure both "polski" and "français" should be lower-cased to match.

ErrorFinder-osu:
> 
> 
> While this issue is a huge nitpick, if each language's capitalisation standards are to be used, they should probably be enforced in all cases. As far as I can tell there are some other language names on that list that do not abide by their language's capitalisation rules - from the ones I know without having to google, I'm pretty sure both "polski" and "français" should be lower-cased to match.

That's it. All uppercase/not uppercase or all in their respective language rules.

peppy:
When adding these I would copy from the wikipedia entry as that seemed to be correct. I'll fix the above ones accordingly (to match the language's rules) but please list any others which look incorrect.

cl8n:
"common noun" doesn't mean "lowercase" though... if it's sentence or title cased for style, you'd have the appropriate caps to go along with it in all of these languages. I don't think I've seen a language selector *not* capitalize all of them, including the one on Wikipedia...

peppy:
Feel free to provide an alternative PR, probably faster.

notbakaneko:
Wikipedia itself isn't consistent; the list of languages [here](https://meta.wikimedia.org/wiki/List_of_Wikipedias) follows the rules of each language but at the same time, the [Spanish version](https://es.wikipedia.org/wiki/Anexo:Wikipedias) of the list in the table doesn't, even for Spanish itself, but the [French version](https://fr.wikipedia.org/wiki/Liste_des_%C3%A9ditions_de_Wikip%C3%A9dia) does.

Helen0903:
All 3 pages seem to follow 2 fairly standard capitalisation rules:
[...]

PR #10665, opened 2023-10-21
Rename language names

agatemosu:


agatemosu:
wrong commit sorry

Poyo-SSB:
Related issue/pull: #6840/#6843

Looks correct according to native-language Wikipedias. Maybe consider changing `português do Brasil` to `português brasileiro`? Not sure if that's idiomatic.

agatemosu:
I asked a Brazilian, and he told me that he sees "português do Brasil" as more suitable. but since it is like another language I think it is better to put "português brasileiro"

nanaya (approved):

Cavitedev:
This tiny change broke osn! picking the website language. It's now fixed in version 3.4.7

PR #10668, opened 2023-10-22
Add country flag tooltips in wiki

### LINKED ISSUE #10661: Country flags in osu! wiki don't have tooltips
They used to have tooltips before the custom syntax. Maybe also link them to the country rankings to be consistent with all the other flags on the site.

Note that lazer already has tooltips but not the linking.

Swekka:
Closes #10661


Swekka:
The tests aren't working as I'd expect, if you look at the [first test failure](https://github.com/ppy/osu-web/actions/runs/6603821744/job/17937338932#step:11:141) it says:

`1) Tests\Libraries\Markdown\ProcessorTest::testHtml with data set #1 ('attribute_flag', '/home/runner/work/osu-web/osu...amples')
ErrorException: Attempt to read property "acronym" on null`

Meaning `Country::find("AU")` was somehow `null`

notbakaneko (changes requested):
I'm pretty sure we don't want to be doing a db query for every single flag

Swekka:
Makes sense, what would be the best way to do this without a db query?

Swekka:
I just realized that if `AU` happens to be in the test db the test will always fail, but I have no idea what to do about that since the countries in the test db are always random

notbakaneko:
I don't think we want to make a network request to redis for every flag either; use a local cache like https://github.com/ppy/osu-web/blob/3b91b06/app/Libraries/Groups.php or https://github.com/ppy/osu-web/blob/3b91b06/app/Libraries/Medals.php does it.

notbakaneko:
I mean making a service object to cache the country names like `Groups` does and adding it to `AppServiceProvider` https://github.com/ppy/osu-web/blob/master/app/Providers/AppServiceProvider.php#L39-L43

notbakaneko (changes requested):
Fetch the countries (preferably just the name and country code) once, store it in an a `Map` or keyed array and do the lookup against the `Map/array` as needed.

notbakaneko (changes requested):
```php
    private function allByCode(): Collection
    {
        return $this->memoize(__FUNCTION__, fn () => Country::get(['acronym', 'name'])->keyBy('acronym'));
    }

    public function byCode(string $code): ?Country
    {
        return $this->allByCode()->get($code);
    }
```

notbakaneko (approved):

PR #10671, opened 2023-10-23
Increase cost of API cost of Beatmap scores country ranking request

notbakaneko:
These queries are kind of expensive; it should be reflected in the API rate limit cost.

~~whoops~~ 👀


notbakaneko:
replay download throttle doesn't apply on non-api route, so 🤷

nanaya:
what does reply download have to do with anything here

PR #10675, opened 2023-10-24
Prevent qualified maps with unresolved issues from getting ranked

### LINKED ISSUE #10654: Prevent qualified maps with open issues/suggestions from getting ranked
While most issues are currently resolved before maps get ranked sometimes a few maps slip through and either never get to apply the suggestion/issue or has to be unranked.

The solution to prevent this from happening is to prevent maps with suggestions/issues from being ranked until the posts have been resolved either by the original poster or an NAT (the NAT will make sure that these posts get resolved)

notbakaneko:
plus an attempt at cleaning up some test methods
closes #10654

- [x] exclude from ranking queue as well?


nanaya:
Yeah I think it should be excluded from the queue otherwise potentially fewer maps will be ranked at each runs.

notbakaneko:
I tried adding tests for hybrid beatmapsets but I'm not sure the results make sense 🤔 
I wonder if these are related  https://github.com/ppy/osu-web/pull/4911/files#diff-4cf8d05043400424e732b329df53d237b69b3d5c10ad285c894aba37cd4835a4R54 
https://github.com/ppy/osu-web/pull/4489/files#diff-03627ef6eb157031bb0c701c8990ea1c6e7713fc129ef82683b5949bb67ebb39R291-R293

nanaya:
I think the order doesn't matter. By ruleset id:
- 0: anything with 0
- 1: anything with 1 except 0
- 2: anything with 2 except 0 and 1
- 3: anything with 3 and nothing else

notbakaneko:
nevermind, I'm counting them wrong 👀

nanaya (changes requested):

app/Console/Commands/ModdingRankCommand.php
```diff
@@ -25,60 +25,81 @@ class ModdingRankCommand extends Command
...
-        foreach ($modeInts as $modeInt) {
+        foreach ($rulesets as $ruleset) {
             $this->waitRandom();
-            $this->rankAll($modeInt);
+
+            if ($this->countOnly) {
+                $stats = static::getStats($ruleset);
+                $this->info($ruleset->name);
+                foreach ($stats as $key => $value) {
+                    $this->line("{$key}: {$value}");
```
notbakaneko, line 73: would have liked to use `twoColumnDetail` but didn't find any option to make the columns narrower 👀

database/factories/BeatmapsetFactory.php
```diff
@@ -70,9 +71,9 @@ public function pending()
         ]);
     }
 
-    public function qualified()
+    public function qualified(?Carbon $approvedAt = null)
```
nanaya, line 74: `\DateTimeInterface`?

nanaya (changes requested):

nanaya (approved):

PR #10677, opened 2023-10-25
Support showing multiple tournament banners

nanaya:
They just stack and no limit for now 🙃


notbakaneko (approved):
what could go wrong 👀

PR #10684, opened 2023-10-27
Create main groups before test

nanaya:
Only create the expected groups. Those are originally just for testing anyway.


nanaya:
Should the identifiers be enum while at it 🤔

notbakaneko:
maybe in a different PR if a lot of files change 👀

notbakaneko (approved):

PR #10685, opened 2023-10-27
Fix stuttering while scrolling on mobile Safari

notbakaneko:
Revert (temporarily, hopefully) to an older version of autosize without the issue

steps to reproduce:
- go to any page with `TextareaAutosize` or autosizable textarea on mobile Safari
- scroll
- scrolling gets interrupted when address bar changes size

~~Uses a version of `autosize` with a workaround for mobile Safari's address bar hiding/showing.
Fixes an issue with autosizing textareas causing inertial scrolling to stop when the address bar resizes; the address bar resizing triggers `resize` events, causing `autosize` to set `scrollTop` which interrupts scrolling.~~

~~https://github.com/notbakaneko/autosize/compare/master...notbakaneko:5bb2b796ade5b5bfd2dffff35d43e7a8ca883551~~

~~I can't seem to trigger the same issue on the autosize demo site, so there might be something more involved in triggering it.~~


peppy:
Is this something that could potentially be PR'd upstream? I guess hard if it can't be repro'd there?

notbakaneko:
Actually, looking into it further, this appears to be a regression in the latest version which the demo doesn't use.

nanaya:
it sure would be nice if we can put note on the package.json entry...

nanaya (changes requested):
Writing down the repro steps here seems like a good idea when someone comes back and check if it's fixed in future version

peppy:
I'd probably report upstream if we can? Better than pinning an old version with no resolution in sight.

nanaya (approved):

PR #10686, opened 2023-10-27
Add column to record playlist item attempts

nanaya:
The table ends up a bit weird with `score_id` changed to nullable but the existing data already partially pointing to old `multiplayer_scores.id` (which means now pointing to wrong `solo_scores.id`). And I've added a bunch of `whereHas` check when migrating to score link before 🤷

- [ ] reminder to update existing attempts
  ```
  update multiplayer_scores_high h set attempts = (select (select count(*) from solo_score_tokens t where h.user_id = t.user_id and t.playlist_item_id = h.playlist_item_id) + (select count(*) from multiplayer_score_links l where h.user_id = l.user_id and l.playlist_item_id = h.playlist_item_id));
  ```
- [ ] migration: `2023_10_26_000000_add_attempts_to_multiplayer_scores_high`
  ```
  alter table `multiplayer_scores_high` modify `score_id` bigint unsigned null;
  alter table `multiplayer_scores_high` add `attempts` tinyint unsigned not null default '0' after `pp`;
  ```


notbakaneko (changes requested):
Do existing attempts need to be accounted for in the migration?

nanaya:
it probably should be done after deployment

notbakaneko (changes requested):

notbakaneko (approved):

PR #10693, opened 2023-10-30
Split storage and url generator classes

nanaya:
This removes the need to initialise class entirely just to get file url. The places which use previous underlying `Storage` aren't performance critical and can just use the facade directly.

There may be some performance hit depending on how it's used/tested but that's what #10659 is for.


notbakaneko (approved):

nanaya:
hol up (more facades)

notbakaneko (approved):

PR #10696, opened 2023-10-31
Fix multiplayer score mods not being validated properly when playlist item has no allowed mods

bdach:
The `allowed_mods` list being empty is not supposed to mean that any mod is allowed. Any and all allowed / "free" mods must (and will) be explicitly listed in `allowed_mods`; if the list is empty, then it means that the score's mods must be exactly set-equal to `required_mods`.

I discovered this when working on https://github.com/ppy/osu/issues/5378 - I was very surprised when web set me submit a play with the "Touch Device" mod active even when it was (by design) not included in either the required or allowed mods. As it turns out, it would only let me do this if no allowed / "free" mods were set; if at least one allowed mod was specified, the client would correctly report an error when trying to submit the score.

(And yeah that means I'm likely going to be adding a carve-out for the Touch Device mod so that it can be submitted regardless of required or allowed mods. In discussing this with the client team [we came to a conclusion that it'll probably be a hardcoded check for that specific mod for now](https://discord.com/channels/188630481301012481/188630652340404224/1168936077344657452) as we don't foresee this being needed very often in the future.)

Dunno if the included tests are up to scratch. They do seem to be working like I want them to, but they were mostly written using the process of trial and error (mostly error).

### DIFF
--- a/tests/Models/Multiplayer/ScoreLinkTest.php
+++ b/tests/Models/Multiplayer/ScoreLinkTest.php
@@ -8,7 +8,6 @@ declare(strict_types=1);
 namespace Tests\Models\Multiplayer;
 
 use App\Exceptions\InvariantException;
-use App\Models\Beatmap;
 use App\Models\Multiplayer\PlaylistItem;
 use App\Models\Multiplayer\ScoreLink;
 use App\Models\User;
@@ -20,7 +19,6 @@ class ScoreLinkTest extends TestCase
     public function testRequiredModsMissing()
     {
         $user = User::factory()->create();
-        $beatmap = Beatmap::factory()->create();
         $playlistItem = PlaylistItem::factory()->create([
             'required_mods' => [[
                 'acronym' => 'HD',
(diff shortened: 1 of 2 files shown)


tests/Models/Multiplayer/ScoreLinkTest.php
```diff
@@ -0,0 +1,165 @@
...
+use App\Models\User;
+use Carbon\Carbon;
+use Tests\TestCase;
+
+class ScoreLinkTest extends TestCase
+{
+    public function testRequiredModsMissing()
+    {
+        $user = User::factory()->create();
+        $beatmap = Beatmap::factory()->create();
```
nanaya, line 23: beatmap is already created by the playlist item factory

(and the `complete` function call below can use `$playlistItem->beatmap_id` and `$playlistItem->ruleset_id`)

tests/Models/Multiplayer/ScoreLinkTest.php
```diff
@@ -0,0 +1,165 @@
...
+        $playlistItem = PlaylistItem::factory()->create([
+            'required_mods' => [[
+                'acronym' => 'HD',
+            ]],
+        ]);
+        $scoreLink = ScoreLink::factory()->create([
```
nanaya, line 29: this should either be `->make` or `new ScoreLink([...])` because otherwise it'll create an already completed score which doesn't make sense in this context.

(looking again I should've changed the `complete` function as static method to prevent this from being allowed at all)

nanaya (changes requested):

nanaya (approved):

PR #10697, opened 2023-11-01
Move score link complete function to static method

nanaya:
It makes more sense to have model creation as part of it and prevents "completing" existing one.

### DIFF
--- a/app/Models/Multiplayer/ScoreLink.php
+++ b/app/Models/Multiplayer/ScoreLink.php
@@ -9,6 +9,7 @@ namespace App\Models\Multiplayer;
 
 use App\Exceptions\InvariantException;
 use App\Models\Model;
+use App\Models\ScoreToken;
 use App\Models\Solo\Score;
 use App\Models\User;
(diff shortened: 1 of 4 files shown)


app/Models/Multiplayer/ScoreLink.php
```diff
@@ -24,6 +25,39 @@
...
+            if (!empty(array_diff($mods, $requiredMods, $allowedMods))) {
+                throw new InvariantException('This play includes mods that are not allowed.');
+            }
+
+            $scoreId = $score->getKey();
+            $token->fill(['score_id' => $scoreId])->saveOrExplode();
+
+            $ret = new static([
+                'playlist_item_id' => $playlistItem->getKey(),
+                'score_id' => $scoreId,
```
notbakaneko, line 52: also `associate($score)`? It's used right after `ScoreLink::complete`

notbakaneko (changes requested):

notbakaneko (approved):

PR #10703, opened 2023-11-02
Support multiple file upload in single model

nanaya:
(technically already supported but creating new class for every single of them is rather silly)

Basically move the trait to its own class and standardise relevant attributes to single filename string.

I think the array was initially to allow additional attributes but then the later tables have it as table column instead so that's kind of pointless. There's also no need for it so far either 🤷

The image processing option is kind of questionable but should work for now 🕺 

- [x] #10693


resources/views/admin/forum/forum_covers/index.blade.php
```diff
@@ -18,19 +18,19 @@
                 </h2>
 
                 @foreach ([
-                    'main' => ['cover' => $forum->cover, 'key' => 'main_cover'],
+                    'main' => ['cover' => $forum->cover->file(), 'key' => 'main_cover'],
```
notbakaneko, line 21: `cover?->file()`

app/Libraries/Uploader.php
```diff
@@ -0,0 +1,108 @@
...
+        private array $processors = [],
+    ) {
+        $this->filename = $this->model->{$this->attr};
+    }
+
+    private static function process(string $name, ?array $options, string $srcPath): ?string
+    {
+        switch ($name) {
+            case 'image':
+                $processor = new ImageProcessor($srcPath, $options['maxDimensions'], $options['maxFileSize'] ?? 10_000_000);
```
notbakaneko, line 32: can move to const

app/Models/UserProfileCustomization.php
```diff
@@ -268,6 +267,21 @@ public function setProfileCoverExpandedAttribute($value)
         $this->setOption('profile_cover_expanded', get_bool($value));
     }
 
+    public function customCover()
+    {
+        return $this->customCover ??= new Uploader(
+            'user-profile-covers',
+            $this,
+            'custom_cover_filename',
+            ['image' => ['maxDimensions' => [2400, 640]]],
```
notbakaneko, line 276: dimensions can go to const

notbakaneko (changes requested):

notbakaneko (approved):

PR #10704, opened 2023-11-02
Allow plays with touch device mod to be submitted in multiplayer regardless of required/allowed mods

bdach:
Companion piece to https://github.com/ppy/osu/pull/25348

Opening as draft as I'm not sure the direction will stick.

The general problem is thus: we want to distinguish scores with the touch device mod. For now, regardless of whether we want to specify that mod as required or allowed in a playlist room, the mod _must_ be accepted at submission time _even if_ the Touch Device mod is not required or allowed. So it gets it own special flag `AlwaysValidForSubmission` to avoid hardcoding the acronym, and that gets checked and excluded at score completion time to ensure that submission can go through.


database/mods.json
bdach, line 1: These changes were one-off manually generated from https://github.com/ppy/osu/pull/25348 and an un-PRed osu-tools branch (https://github.com/ppy/osu-tools/compare/master...bdach:osu-tools:mods-command-new-flag). I did it this way because there wasn't clear consensus on discord whether this should be [a hardcode](https://discord.com/channels/188630481301012481/188630652340404224/1168936176237940767) or [a flag](https://discord.com/channels/188630481301012481/188630652340404224/1169161156041121843), so I basically picked one I liked better to force the issue. I'll PR the osu-tools change if this gets accepted.

app/Libraries/Mods.php
```diff
@@ -112,6 +112,19 @@ public function assertValidForMultiplayer(int $rulesetId, array $ids, bool $isRe
...
+    {
+        $this->validateSelection($rulesetId, $ids);
+
+        $rulesetMods = $this->mods[$rulesetId];
+
+        return collect($ids)
+            ->filter(function ($id) use ($rulesetMods) {
+                return !$rulesetMods[$id]['AlwaysValidForSubmission'];
+            })
+            ->all();
```
nanaya, line 125: No need to use `collect` here. A plain `array_values(array_filter(` works.

Also `fn ($id) => $...` is simpler and doesn't need `use (...)`

database/mods.json
nanaya: what's the plan here? get this PR merged first or wait for the tools etc to be updated?

nanaya (changes requested):

database/mods.json
bdach: I suppose the web change should be merged and deployed before game-side for 100% correctness. I'll PR the tools branch now but it is not urgent.

app/Libraries/Mods.php
```diff
@@ -112,6 +112,19 @@ public function assertValidForMultiplayer(int $rulesetId, array $ids, bool $isRe
         }
     }
 
+    public function excludeModsAlwaysValidForSubmission(int $rulesetId, array $ids): array
+    {
+        $this->validateSelection($rulesetId, $ids);
```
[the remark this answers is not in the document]
bdach: "monkey see, monkey do, monkey no understand" moment. Will remove.

app/Libraries/Mods.php
bdach: "idiot that doesn't write php tries to write php" moment. Will apply in a sec.

database/mods.json
nanaya: just need to remember to not have the changes here gone by unrelated release, I guess 🤔

nanaya (approved):

Issue #10377, opened 2023-07-17
[Discussion] Add support for matching full terms at song select using suffixed !


Skubers:
https://github.com/ppy/osu/pull/24058

Can we have this on osu-web too? Since searching on web returns a MUCH bigger amount of maps than the locally installed ones, I'd think it's a no-brainer to add that

peppy:
Would definitely be cool to add. Would also be cool to have some sort of hinting text so users could be made aware of the text filter options they have available.

PR #10713, opened 2023-11-07
Surround query with quotes when clicking on title, artist or source on beatmapset page

### LINKED ISSUE #8841: Increase the search accuracy when clicking source's super link on map's description
In my recent practice on deciding whether i should put the full event name"THE BMS OF FIGHTERS XVI -NEO DYSTOPIA-" or abbreviation"BOFXVI" on my map's source 
Turns out i just discovered how misleading putting full BOFXVI event name into source:
https://media.discordapp.net/attachments/951524313763287071/967182643651297300/unknown.png?width=474&height=671
when i click the source on my map's desc and it lead you to map searching page and none of them are legit from the event:<
but with Abbreviation"BOFXVI" it's a lot relevant:
https://cdn.discordapp.com/attachments/849692819799277578/967192792910659614/unknown.png

As a very clueless mapper myself i asked other experienced mappers how can i solve this and they said you can customize in beatmap searching:
Mapper: The reason why you see so many bad result is cuz of THE
which is very common word
only search keyword with fuzzy search (without quotes)
if u know exactly what u want, always use ""

that did answered my problem but another question came to my mind
why they don't put the "" automatically? when leading to beatmap searching from map's description?
Cause what you put in source should be seen as a whole instead of separated letters.
Manually adding that like the current one is fine but seems like a easier way tis to fix it for everyone that not knowing this searching tech beforehand!

An alternative:
Having a button as default 
Like with " or not
Search by a whole phrase
Search by each in search engine is also a solution

Hoping this issue will be reviewed so i can put my epic event name on my map before qualified(X
I mean,increase the total search accuracy on the website:>

Swekka:
- [x] Depends on #10718

Closes #8841


nanaya (changes requested):
It's not quite working yet (to be improved by #10716) but it should be double (double) quoted (`title=""the title""`).

Swekka:
Isn't it planned to use `title="the title"!` for full matching? (#10377)

nanaya:
I don't know when that will happen. Adding support for that is a bit more involved. The double double quoting is for searching the words in specified order.

Swekka:
Oh, I see, I thought single quotes already did that.

nanaya (approved):

PR #10719, opened 2023-11-09
Prevent enter key on ime completion from triggering blur

nanaya:
Being able to type things without having to refocus is kinda important.

On firefox it's "Process" on keydown but somehow "Enter" on keyup 😐

On chrome it's both "Process"

And then there's safari

|                                    | firefox                                           | chrome                                                         | safari                          |
|------------------------------------|---------------------------------------------------|----------------------------------------------------------------|---------------------------------|
| keydown isComposing check          | supported                                         | supported                                                      | no isComposing flag at any case |
| keyup                              | no difference between composing and non-composing | Process event.key (windows only?), otherwise indistinguishable | same as above                   |
| keypress                           | deprecated                                        | deprecated                                                     | deprecated                      |
| beforeinput on non-composing enter | supported                                         | supported                                                      | missing                         |


notbakaneko (approved):

notbakaneko (approved):
seem to work 👀

Issue #7641, opened 2021-05-24
There is no on-site way to learn about what each ranked category does


cl8n:
... besides the wiki, which should never be assumed as information that everyone will read, especially if not directly linked to from somewhere more visible.

## Problem

Ranked categories are the backbone of how online maps work, yet there's very little information presented to players about what they are. The only in-your-face messaging the game gives in this regard is when you load a Qualified or Loved map and get the banner saying your score is temporary and won't award pp. Everything about the purpose of segregating maps into categories to what processes maps go through to get there is left to word of mouth and community-provided resources that aren't straightforward to find or be made aware of.

I suspect that the lack of presented information is in large part responsible for common misunderstandings about Ranked and Loved, but it's hard to prove relation there without interviewing lots of players (which I haven't done before opening this issue). Even if this isn't the case, I think it would be beneficial to explain more parts of the ranking system & other categories on-site where they are relevant, instead of deferring all not-super-obvious info to the wiki without reference.

## Feature idea

This is just one random idea for tackling part of the problem but I think it would be very impactful as a start.

On the beatmap listing, below the filter options and above the maps list, there could be a brief document explaining about the currently selected ranked category: the practical function (e.g. "these maps have leaderboards, award PP, and cannot be updated"), the purpose, the process, and whatever else is relevant. Links to wiki pages would be provided for more in-depth reading.

To me this feels like the perfect place to explain what the categories are. The beatmap listing is the central hub for finding those maps.

It may be annoying for experienced players that already know what all the categories do, so it could be toggled closed and saved as preference.

### Including current events

This info on the listing could also be adapted to include current events, for example a banner about a Project Loved voting cycle for Loved, or ongoing BN/NAT projects for Ranked, like the extra incentives to help people get first-ranked-maps recently.

To make management easy, it could be implemented as markdown documents in osu-wiki, similar to how news articles are sourced from there.

---

I got this idea from a project [DigitalHypno](https://osu.ppy.sh/users/4384207) spent some time on recently, where he interviewed a spread of random players about UX of things related to Project Loved. A common thread among most respondents was that it's not clear where to find current info about Loved, or just annoying to get there (the latest news post). After talking about this we agreed that there should be some "hub" area for Loved that's not out of the way, and the beatmap listing seems like the current best candidate to host that. IMO the issue is not specific to Loved so I've made this proposal more general.

Other issues that could help address the similar problems:
- #2737
- #3976

peppy:
[...]

PR #10722, opened 2023-11-10
Make beatmapset status label on beatmapset page link to wiki

Swekka:
Related: #7641
- [x] Depends on ppy/osu-wiki/pull/10426


nanaya:
This needs update on osu-wiki to have uniform id across languages.

nanaya (approved):

PR #10732, opened 2023-11-13
Fix missing `{{` `}}` brackets in blade

Swekka:


nanaya (approved):

notbakaneko:
probably better to just remove this whole section as we don't physically handle shipping anymore, and all the existing tracking codes would have already expired...

PR #10737, opened 2023-11-14
Fix beatmap and spotlight selector blackout when navigating via history

notbakaneko:
Blackout doesn't get cleared/set properly on the discussion beatmap list and spotlights selector when they're open and navigating back/forwards to the page.

### DIFF
--- a/resources/js/components/select-options.tsx
+++ b/resources/js/components/select-options.tsx
@@ -48,8 +48,7 @@ export default class SelectOptions<T extends Option> extends React.Component<Pro
     super(props);
     makeObservable(this);
     disposeOnUnmount(this, autorun(() => {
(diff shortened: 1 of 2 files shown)


resources/js/components/select-options.tsx
```diff
@@ -52,6 +52,7 @@ export default class SelectOptions<T extends Option> extends React.Component<Pro
       (value) => {
         blackoutToggle(value ?? false, 0.5);
       },
+      { fireImmediately: true },
```
nanaya, line 55: might as well use autorun instead

nanaya (changes requested):

nanaya (approved):
This whole blackout handling can use some refactor 😐 there's one in Modal component but it needs to also deal with all blackouts

PR #10738, opened 2023-11-14
Country model cleanup

nanaya:
Combine relations and remove unused transformer includes.

### DIFF
--- a/app/Models/Country.php
+++ b/app/Models/Country.php
@@ -5,6 +5,9 @@
 
 namespace App\Models;
 
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Relations\HasOne;
+
 /**
  * @property string $acronym
  * @property int $display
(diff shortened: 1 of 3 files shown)


app/Models/Country.php
```diff
@@ -35,29 +38,24 @@ public function scopeForStore($query)
...
     {
-        return $this->hasOne(CountryStatistics::class, 'country_code')->where('mode', Beatmap::MODES['osu']);
+        return $this->hasMany(ProfileBanner::class, 'country_acronym');
     }
 
-    public function statisticsTaiko()
+    public function statistics(): HasOne
     {
-        return $this->hasOne(CountryStatistics::class, 'country_code')->where('mode', Beatmap::MODES['taiko']);
+        return $this->hasOne(CountryStatistics::class, 'country_code');
```
notbakaneko, line 58: shouldn't this be `hasMany`?

notbakaneko (changes requested):

notbakaneko (approved):

PR #10741, opened 2023-11-17
Tidy up socket worker

notbakaneko:
There's only 1 `xhr` and can probably use that to track loading state (and remove the nullable boolean)


resources/js/socket-worker.ts
```diff
@@ -169,23 +168,22 @@ export default class SocketWorker {
...
-
-    this.xhr.startWebSocket = $.get(route('notifications.endpoint'))
+    this.xhr = $.get(route('notifications.endpoint'));
+    this.xhr
       .always(action(() => {
-        this.xhrLoadingState.startWebSocket = false;
+        this.xhr = null;
       }))
-      .done(action((data: NotificationFeedMetaJson) => {
+      .done((data) => runInAction(() => {
```
nanaya, line 182: nothing inside seem to need to be run in action 🤔 also for the other callbacks

nanaya (changes requested):

nanaya (approved):

PR #10742, opened 2023-11-17
Remove ESLint strict-boolean-expressions exclusions

notbakaneko:
Disabling the exclusions to `@typescript-eslint/strict-boolean-expressions`; PR'ing these 3 rules separate from `allowNullableBoolean` as that one is a lot more changes by itself 👀


nanaya (changes requested):

.github/workflows/lint.yml
```diff
@@ -49,7 +49,7 @@ jobs:
       - name: Install js dependencies
         run: yarn --frozen-lockfile
 
-      - run: 'yarn lint --max-warnings 96 > /dev/null'
+      - run: 'yarn lint --max-warnings 95 > /dev/null'
```
nanaya, line 52: now 94

nanaya (approved):

PR #10765, opened 2023-11-27
Add video display to artist pages

venix12:
instead of #10761


notbakaneko:
I'm guessing the video isn't necessarily going to be on youtube only?

venix12:
Assumed so considering the newspost ones are hosted on assets.ppy.sh and we are talking the same videos here, @peppy may know better though.

peppy:
It should always be self hosted on assets.

resources/views/artists/show.blade.php
```diff
@@ -46,6 +46,14 @@
                     <h1>{{ $artist->name }}</h1>
 
                     {!! markdown($artist->description) !!}
+
+                    @if ($artist->video_url !== null)
+                        <div class="artist__video">
+                            <div class="u-embed-wide">
+                                <iframe src="{{ $artist->video_url }}" allowfullscreen></iframe>
+                            </div>
```
notbakaneko, line 54: If they're always self-hosted then just use `video`; putting video url into iframe also clips the controls on Safari

notbakaneko (changes requested):

notbakaneko (approved):

PR #10771, opened 2023-11-29
Change data provider to static method

nanaya:
Non-static is deprecated in phpunit 10.

Also:
- updated phpunit dusk schema
- skip creating app in provider if possible
  - move base64url to its own class (helper isn't available)
  - use Arr::random instead of array_rand_val
  - use config in a callback
  - read config file directly instead of relying on config function
  - change some test class helpers to static method

I'm not rearranging them because there are just a lot of them ಠ_ಠ


tests/CreatesApplication.php
```diff
@@ -9,16 +9,21 @@
 
 trait CreatesApplication
 {
+    public static function createApp()
```
notbakaneko, line 12: replace the other `(new static())->createApplication();` calls with this?

notbakaneko (changes requested):

PR #10772, opened 2023-11-29
Add test for verification link

nanaya:
Change default session and cache driver to redis because otherwise persistent session would be _rather tricky_. It would also allow testing user session management eventually - still missing a few things for that though which includes having correct id for logged in session.

Initially thinking of clearing the database after the test finished but then it also deletes score indexer active schema info.


notbakaneko (commented):
Wonder if it's worth setting the `*_DB` redis envs to a different value in `.env.testing.example` 🤔

nanaya:
there are too many places to set the variable including the indexer (docker compose, github, dusk env, testing env).

PR #10776, opened 2023-11-30
Simplify session store

nanaya:
- make non-instance specific session functions static
- combine various instance delete functions
- let laravel handles osu-next prefix
- make the whole `sessions:$user:id` the id instead of considering parts of them prefix of the actual id
- always return current session when looking up by session id
- obtain session user id from attribute instead of prefix
- add more session related tests


config/session.php
```diff
@@ -71,7 +71,7 @@
     |
     */
 
-    'connection' => null,
+    'connection' => 'cache',
```
notbakaneko, line 74: should be using the main redis db, not the cache db

notbakaneko (changes requested):
still checking 👀

app/Libraries/Session/Store.php
```diff
@@ -3,229 +3,153 @@
...
+        UserSessionEvent::newLogout($userId, $ids)->broadcast();
     }
 
-    /**
-     * Destroys a session owned by the current user that is identified by the given id.
-     *
-     * @return bool
-     */
-    public function destroyUserSession($sessionId)
+    public static function findOrCreate(?string $id = null): static
```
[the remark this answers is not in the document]
nanaya: also, an id is always assigned so in a way it does look like it's created...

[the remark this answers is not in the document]
nanaya: actually, when doing something else I noticed it's not even currently possible to tell whether or not the session actually exists

app/Libraries/Session/Store.php
```diff
@@ -235,63 +159,58 @@ public function getIdWithoutKeyPrefix()
...
-    protected function generateSessionId(int $userId = null)
+    public function userId(): ?int
     {
-        // Overriden to allow namespacing the session id (used as the redis key)
-
-        return static::keyPrefix($userId).':'.Str::random(static::SESSION_ID_LENGTH);
+        // From `Auth::getName()`.
+        // Hardcoded because Auth depends on this class instance which then
+        // calls this functions and would otherwise cause circular dependency.
+        return $this->attributes['login_web_59ba36addc2b2f9401580f014c7f58ea4e30989d'] ?? null;
```
notbakaneko, line 203: should also mention its using the sha1 of the class in `Illuminate\Auth\SessionGuard::getName()` which is always the same 
...I'm pretty sure there'll be other trollings if the hash ever gets changed, anyway 🤔

app/Libraries/Session/Store.php
```diff
@@ -235,63 +159,58 @@ public function getIdWithoutKeyPrefix()
      * @param  string  $sessionId
      * @return bool
      */
-    public function migrate($destroy = false, int $userId = null)
+    public function migrate($destroy = false)
```
notbakaneko, line 162: Can delete outdated phpdoc

notbakaneko (changes requested):

app/Libraries/Session/Store.php
nanaya: The detail shouldn't matter and I've added relevant test.

app/Http/Controllers/Account/SessionsController.php
```diff
@@ -21,15 +20,16 @@ public function __construct()
...
             // current session
             logout();
         } else {
-            Request::session()->destroyUserSession($id);
+            $session = SessionStore::findOrNew($id);
+            if ($session->userId() === \Auth::user()->getKey()) {
+                $session->delete();
+            } else {
+                abort(404);
+            }
```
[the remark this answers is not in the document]
nanaya: it's a special case due to cookies and current Auth handling

app/Libraries/Session/Store.php
```diff
@@ -3,295 +3,227 @@
...
+        if ($userId !== null) {
+            $listIds = [
+                ...$ids,
+                // Also delete ids that were previously stored with prefix.
+                ...array_map(
+                    fn ($id) => "osu-next:{$id}",
+                    $ids,
+                ),
+            ];
+            $redis->srem(self::listKey($userId), ...$ids);
```
notbakaneko, line 43:
```suggestion
            $redis->srem(self::listKey($userId), ...$listIds);
```
👀

notbakaneko (changes requested):

app/Libraries/UserVerificationState.php
```diff
@@ -119,7 +108,7 @@ public function markVerified()
         $this->session->put('verified', true);
         $this->session->save();
 
-        UserSessionEvent::newVerified($this->user->getKey(), $this->session->getKey())->broadcast();
+        UserSessionEvent::newVerified($this->user->getKey(), $this->session->getId())->broadcast();
```
notbakaneko, line 111: this needs to the cache-prefixed key because that's what osu-notification-server still uses and checks against; the same goes for logout; the `verified` event currently isn't being relayed to the client.

notbakaneko (changes requested):

app/Libraries/Session/Store.php
```diff
@@ -3,295 +3,238 @@
...
-    {
-        if (!static::isUsingRedis()) {
-            return [];
+        $redis = self::redis();
+        $redis->del($ids);
+        if ($userId !== null) {
+            $idsForEvent = self::keysForEvent($ids);
+            // Also delete ids that were previously stored with prefix which is
+            // the full redis key just like for event.
+            $redis->srem(self::listKey($userId), ...[...$ids, ...$idsForEvent]);
```
notbakaneko, line 38: just should work `$redis->srem(self::listKey($userId), ...$ids, ...$idsForEvent);`

app/Libraries/Session/Store.php
```diff
@@ -3,295 +3,238 @@
...
-        if (!static::isUsingRedis()) {
-            return [];
+        $redis = self::redis();
+        $redis->del($ids);
+        if ($userId !== null) {
+            $idsForEvent = self::keysForEvent($ids);
+            // Also delete ids that were previously stored with prefix which is
+            // the full redis key just like for event.
+            $redis->srem(self::listKey($userId), ...$ids, ...$idsForEvent);
+            UserSessionEvent::newLogout($userId, $ids)->broadcast();
```
notbakaneko, line 39: `$idsForEvent`

notbakaneko (changes requested):

notbakaneko (approved):
seems fine™️

PR #10779, opened 2023-12-02
Fix incorrect check for completed playlist item

### LINKED ISSUE #10778: Lazer Non-osu! standard scores are not recorded in Multiplayer lobbies
Looks like scores from other modes are not recorded or represented in the frontend when viewing a multiplayer lobby when done on Lazer.

image

nanaya:
The change to always create playlist item high score entry broke the exist check.

Note: untested.

Resolves #10778 (probably)

### DIFF
--- a/app/Models/Multiplayer/UserScoreAggregate.php
+++ b/app/Models/Multiplayer/UserScoreAggregate.php
@@ -202,7 +202,7 @@ class UserScoreAggregate extends Model
 
     private function updateUserTotal(ScoreLink $currentScoreLink, PlaylistItemUserHighScore $prev)
     {
-        if ($prev->exists) {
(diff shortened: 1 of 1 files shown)


peppy (approved):

app/Models/Multiplayer/UserScoreAggregate.php
```diff
@@ -202,7 +202,7 @@ public function userRank()
 
     private function updateUserTotal(ScoreLink $currentScoreLink, PlaylistItemUserHighScore $prev)
     {
-        if ($prev->exists) {
+        if ($prev->last_score_id !== null) {
```
notbakaneko, line 205: it's `score_id`....

PR #10790, opened 2023-12-07
Readd prefix for new score related tables

nanaya:
- [ ] figure out the actual plan 🙃


nanaya:
Stopping the live version from returning current lazer scores instead.

PR #10800, opened 2023-12-13
Add verification for password oauth token

nanaya:
```
# reissue verification code
POST api/v2/session/verify/reissue

# verify token
POST api/v2/session/verify
parameter: verification_key (string)
```

On `/me` endpoint response, there's a new attribute `session_verified`.

Additionally, notification server will also send verification event when the verification link is opened (`{ "event": "verified" }`)

- [x] depends on #10796
- [ ] implemented and tested and released in client (ppy/osu#25480)


notbakaneko (changes requested):
A verified session sending `verification_key` probably should be issuing a new verification code? 🤔

app/Http/Middleware/VerifyUser.php
```diff
@@ -20,6 +20,7 @@ class VerifyUser
         'notifications_controller@endpoint' => true,
         'sessions_controller@destroy' => true,
         'sessions_controller@store' => true,
+        'users_controller@me' => true,
```
notbakaneko, line 23: Also, does this do anything when API requests don't even use this middleware?

tests/Controllers/OAuth/TokensControllerTest.php
```diff
@@ -36,4 +49,76 @@ public function testDestroyCurrentClientGrant()
...
+            'scope' => '*',
+            'username' => $user->username,
+            'password' => UserFactory::DEFAULT_PASSWORD,
+        ])->assertSuccessful();
+        $tokenJson = json_decode($tokenResp->getContent(), true);
+
+        $meResp = $this->json('GET', route('api.me'), [], [
+            'Authorization' => "Bearer {$tokenJson['access_token']}",
+        ])->assertSuccessful();
+        $meJson = json_decode($meResp->getContent(), true);
```
notbakaneko, line 77: There's this `TestResponse::decodeResponseJson()` and `assertPath`/`assertJsonPath` thing

notbakaneko (changes requested):

notbakaneko (dismissed):
Reminder https://github.com/ppy/osu-notification-server/pull/55 also should be deployed after this is deployed.

peppy:
> Additionally, notification server will also send verification event when the verification link is opened

Given that we don't yet have any setup for notification server in the game yet, is it fine to ignore this for now? Do we need to poll or something to handle the scenario?

nanaya:
ignoring it would confuse users. I guess a token status endpoint could be added...

peppy:
Cancel that, we do have support for notification websocket handling.

But work on this will likely happen early next year from the client side, so let's leave this PR in limbo for now.

nanaya:
> I would like to get started on this client-side, but I'm not sure I get what fits where exactly, so I'm gonna have a few questions here.
> 
> * From what I can gather, the basic flow is as follows:
>
>
>   Do I have that right?
> 

seems mostly right. The newVerified event is always published regardless of the source. I don't think it matters much in the context of client though.

> * In [this](https://github.com/ppy/osu/pull/25480) comment, the following was stated:
>   > Basically there will be a case where web will let us know via **a response to any API request** that we need verification, which will cause the client to bring up the prompt.
>       
>       
>    Is this really the case here? Do I need to be potentially handling (re-)verification on any response? From source it kinda appears to me that it's only scoped to `GET /api/v2/users/me` but I may well be wrong.
> 

mainly only POST requests will return verification required response (exceptions etc applies). It's the same as web at the moment.

> * Do I need to be handling token _revocation_? As in, if a `UserSessionEvent::newLogout()` event is broadcast, do I need to be doing anything? Is that even something that can happen for a client token right now?

a logout event means the session/token is no longer valid. One case would be when the user changes password (on web) it'll log out all other sessions and tokens.

technically even without handling it any further requests will be denied anyway but at least in osu-web a forced logout will be immediately applied.

peppy (approved):

PR #10804, opened 2023-12-15
Store profile cover data in user table

nanaya:
This will transfer the cover data from profile customization table on access. The whole data should be transferred once deployed (manually by looping through users and accessing their cover) and then later remove the profile customization subquery/include when it's only used for cover.

- [x] depends on #10703


notbakaneko (approved):
Means any access to a page with a user's cover will trigger the cover migration, but I guess we'll see how that goes 🤷

PR #10805, opened 2023-12-17
Don't render spoiler button in button

notbakaneko:
Nested `buttons` are not valid and will result in stupid rendering.
The best part is dev inspector will helpfully 'fix' the html tree so you don't actually see what's actually being rendered ( ﾟ ヮﾟ)

### DIFF
--- a/app/Libraries/BBCodeFromDB.php
+++ b/app/Libraries/BBCodeFromDB.php
@@ -62,7 +62,16 @@ class BBCodeFromDB
 
     public function parseBox($text)
     {
-        $text = preg_replace("#\[box=((\\\[\[\]]|[^][]|\[(\\\[\[\]]|[^][]|(?R))*\])*?):{$this->uid}\]\n*#s", $this->parseBoxHelperPrefix('\\1'), $text);
+        $text = preg_replace_callback("#\[box=((\\\[\[\]]|[^][]|\[(\\\[\[\]]|[^][]|(?R))*\])*?):{$this->uid}\]\n*#s", function ($m) {
+            // don't render button in button.
+            $escapedMatch = strtr($m[1], [
+                "[spoilerbox:{$this->uid}]" => '',
+                "[/spoilerbox:{$this->uid}]" => '',
+            ]);
+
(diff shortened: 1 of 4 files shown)


nanaya:
maybe just disable block in button in cleanhtml's purifier option?

nanaya:
actually, replacing button with anchor is probably safer as its definition is already included

app/Libraries/BBCodeFromDB.php
```diff
@@ -75,7 +75,7 @@ public function parseBoxHelperPrefix($linkText = null)
     {
         $linkText = presence($linkText) ?? 'SPOILER';
 
-        return "<div class='js-spoilerbox bbcode-spoilerbox'><button class='js-spoilerbox__link bbcode-spoilerbox__link' type='button'><span class='bbcode-spoilerbox__link-icon'></span>{$linkText}</button><div class='bbcode-spoilerbox__body'>";
+        return "<div class='js-spoilerbox bbcode-spoilerbox'><a class='js-spoilerbox__link bbcode-spoilerbox__link'><span class='bbcode-spoilerbox__link-icon'></span>{$linkText}</a><div class='bbcode-spoilerbox__body'>";
```
nanaya, line 78: needs a href so it responds to both tab and enter key

app/Libraries/BBCodeFromDB.php
```diff
@@ -75,7 +75,7 @@ public function parseBoxHelperPrefix($linkText = null)
     {
         $linkText = presence($linkText) ?? 'SPOILER';
 
-        return "<div class='js-spoilerbox bbcode-spoilerbox'><button class='js-spoilerbox__link bbcode-spoilerbox__link' type='button'><span class='bbcode-spoilerbox__link-icon'></span>{$linkText}</button><div class='bbcode-spoilerbox__body'>";
```
nanaya, line 78: the incorrect button element definition in cleanhtml can be removed

nanaya (changes requested):

nanaya (approved):

PR #10822, opened 2023-12-22
Add toggle for legacy score only view

nanaya:
This doesn't do anything by itself but should theoretically allow me to update the other new score PRs without going back to interleaving the commits with each other.

I hope.

Added helper because it's needed in quite a few places.

### DIFF
--- a/app/Libraries/Search/ScoreSearchParams.php
+++ b/app/Libraries/Search/ScoreSearchParams.php
@@ -11,11 +11,26 @@ use App\Libraries\Elasticsearch\SearchParams;
 use App\Libraries\Elasticsearch\Sort;
 use App\Models\Solo\Score;
 use App\Models\User;
+use App\Models\UserProfileCustomization;
 
 class ScoreSearchParams extends SearchParams
 {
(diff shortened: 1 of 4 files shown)


app/Libraries/Search/ScoreSearchParams.php
```diff
@@ -39,19 +54,10 @@ public static function fromArray(array $rawParams): static
...
-    public ?int $rulesetId = null;
-    public $size = 50;
-    public ?User $user = null;
-    public ?int $userId = null;
-
-    private ?string $countryCode = null;
-    private string $type = self::DEFAULT_TYPE;
+    public static function showLegacyForUser(?User $user): bool
+    {
+        return $user?->userProfileCustomization?->legacy_score_only ?? UserProfileCustomization::DEFAULT_LEGACY_ONLY_ATTRIBUTE;
```
notbakaneko, line 59: using a private const

notbakaneko (changes requested):

notbakaneko (approved):
does this need merging yet?

nanaya:
yeah, I'll update the other PRs to use this.

Issue #10560, opened 2023-09-14
Ranking page tabs can use some rearranges


nanaya:
[Something like this](https://discord.com/channels/188630481301012481/188630616286167041/1151425194326507580)
```
performance
score
country
multiplayer
seasons
spotlights (old)
kudosu
```

nanaya:
Fixed in the linked pr.

PR #10831, opened 2023-12-26
Rearrange ranking page tabs

mayacopeland:
Changed the order for tabs as per #10560, changed the english text to spotlights (old).


resources/lang/en/rankings.php
```diff
@@ -24,7 +24,7 @@
     ],
 
     'type' => [
-        'charts' => 'spotlights',
```
nanaya, line 27: the top menu also needs update (defined somewhere in `app/helpers.php`)

nanaya (changes requested):

mayacopeland:
ughhhhh, why did it remove those files???

PR #10832, opened 2023-12-27
👷 Also build images from private-staging-v2

ThePooN:
Going with branch builds, unless we go with tags instead? Would have to for automated deploys.


peppy:
I think tag will do for now. This should hopefully be quite quick and painless..

PR #10833, opened 2023-12-27
Add menu to switch score display mode

nanaya:
( ﾟ ヮﾟ)a


resources/views/layout/_score_mode_toggle.blade.php
```diff
@@ -0,0 +1,24 @@
+{{--
+    Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+    See the LICENCE file in the repository root for full licence text.
+--}}
+@php
+    $legacyScoreOnlyValue = App\Libraries\Search\ScoreSearchParams::showLegacyForUser(Auth::user());
+    $icon = $legacyScoreOnlyValue
```
notbakaneko, line 7: Is this backwards? lazer mode = legacy score only...?

PR #10834, opened 2023-12-27
Support joining public chat channels from web interface

### LINKED ISSUE #4661: Add the ability to join channels in Chat
Recently noticed you can chat in osu! channels like #osu #taiko etc. through Chat. Is it possible to add the ability to add channels into the chat list like how the osu! client allows you to do?

Something like this:
image

notbakaneko:
closes #4661

(also needs a better name for `ChannelId` because it's not actually a channel 👀 )


resources/js/entrypoints/chat.tsx
```diff
@@ -77,8 +81,10 @@ core.reactTurbolinks.register('chat', action(() => {
     core.dataStore.channelStore.lastReceivedMessageId = initial.last_message_id ?? 0;
   }
 
-  if (currentUrl().hash === '#create') {
-    core.dataStore.chatState.selectChannel('create', 'replaceHistory');
+  const hash = getChannelHash(currentUrl().hash);
+
+  if (hash === 'create' || hash === 'join') {
```
nanaya, line 86: is it any better than just comparing with `#create` and `#join`? Also `getChannelHash` is really just get hash without the hash sign, nothing to do with channel?

resources/js/chat/chat-state-store.ts
```diff
@@ -121,50 +144,59 @@ export default class ChatStateStore implements DispatchListener {
...
-    const json = this.createAnnouncement.toJson();
-    this.waitJoinChannelUuid = json.uuid;
-    // TODO: when adding more channel types to join, remember to add separate busy spinner states for them.
-
-    createAnnouncement(json)
-      .fail(action((xhr: JQueryXHR) => {
-        onError(xhr);
-        this.waitJoinChannelUuid = null;
-      }));
+  async joinChannel(channelId?: number) {
```
nanaya, line 151: addChannel? Seems weird calling this joinChannel but it sometimes call createAnnouncement instead of (a different) joinChannel

also kinda match the channel id type

resources/js/chat/join-channels.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  let statusElement: React.ReactNode | undefined;
+  if (status === 'joined') {
+    statusElement = <i className='fas fa-check' />;
+  } else if (status === 'joining') {
+    statusElement = <Spinner />;
+  }
+
+  return (
+    <button key={channel.channel_id} className={classWithModifiers('chat-join-channel__channel', { joined: status === 'joined' })} onClick={handleClick}>
```
nanaya, line 38: key not valid/needed in this context

resources/js/chat/join-channels.tsx
```diff
@@ -0,0 +1,152 @@
...
+    this.error = false;
+
+    try {
+      this.xhr = getPublicChannels();
+      const channels = await this.xhr;
+      runInAction(() => {
+        this.channels = channels;
+      });
+    } catch (error) {
+      if (!isJqXHR(error)) throw error;
```
nanaya, line 106: should at least handle abort (navigating away somewhere else)

```
Uncaught (in promise) 
Object { readyState: 0, getResponseHeader: getResponseHeader(key), getAllResponseHeaders: getAllResponseHeaders(), setRequestHeader: setRequestHeader(name, value), overrideMimeType: overrideMimeType(type), statusCode: statusCode(map), abort: abort(statusText), state: state(), always: always(), catch: catch(fn), … }
chat:1036:95
```

nanaya (changes requested):

resources/js/entrypoints/chat.tsx
notbakaneko: it narrows the type down for `selectChannel`

nanaya: I see... ಠ_ಠ

The function name is still weird though and I don't think it's any better than straight out `.hash.slice(1)` since it must have the `#` by definition and clear enough in context (and slicing empty string doesn't do any harm).

notbakaneko: I just got rid of the function since it didn't end up being used for anything else.

resources/js/components/textarea-autosize.tsx
```diff
@@ -24,7 +24,7 @@ export default class TextareaAutosize extends React.PureComponent<Props, State>
   private shouldUpdate = true;
 
   private get maxHeight() {
-    return this.props.maxRows != null && this.state.lineHeight != null
+    return this.props.maxRows != null && this.state.lineHeight != null && Number.isFinite(this.state.lineHeight)
```
nanaya, line 27: not really important but if lineHeight is typed non nullable and initialized with NaN, only isFinite check is needed

resources/js/chat/public-channels.ts
```diff
@@ -0,0 +1,54 @@
...
+
+      runInAction(() => {
+        this.lastRefreshTime = new Date();
+        this.channels = channels;
+      });
+    } catch (error) {
+      if (!isJqXHR(error)) throw error;
+
+      runInAction(() => {
+        this.error = true;
```
nanaya, line 46: also reset lastRefreshTime?

resources/js/chat/chat-state-store.ts
```diff
@@ -121,50 +187,27 @@ export default class ChatStateStore implements DispatchListener {
...
       }
 
       return;
     }
 
-    if (channelId == null) return;
-
     const channel = this.channelStore.get(channelId);
 
     if (channel == null) {
```
nanaya, line 213: feels like this should reset `this.selected` as well or maybe even have `this.selected` as `Channel|AddChannelType|null` if the channel models don't change.

resources/js/chat/chat-state-store.ts
```diff
@@ -213,12 +256,17 @@ export default class ChatStateStore implements DispatchListener {
...
-
-    if (this.waitJoinChannelUuid != null && this.waitJoinChannelUuid === event.json.uuid) {
-      this.selectChannel(event.json.channel_id);
-      this.waitJoinChannelUuid = null;
-      if (event.json.type === 'ANNOUNCE') {
+    const json = event.json;
+    this.channelStore.update(json);
+
+    if (this.waitAddChannelType === 'announce' && this.waitAddChannelId === json.uuid
+      || this.waitAddChannelType === 'existing' && this.waitAddChannelId === json.channel_id
```
nanaya, line 263: isn't it enough to just check type of waitAddChannelId and match the uuid/channel_id accordingly? That way the waitAddChannelType wouldn't be needed

(which also a bit confusing why waitAddChannelId being a string means it's announce type)

resources/js/chat/chat-state-store.ts
```diff
@@ -47,24 +50,47 @@ export default class ChatStateStore implements DispatchListener {
 
   // Should not join/create another channel if still waiting for a pending request.
   @computed
-  get isJoiningChannel() {
-    return this.waitJoinChannelUuid != null;
+  get isAddingChannel() {
```
nanaya, line 53: is it actually useful having `@computed` for simple comparison like this and some others here

nanaya (changes requested):

resources/js/chat/chat-state-store.ts
notbakaneko: This shouldn't happen (since the selection comes from the list of channels), unless some funny business is being done, so it's better not to mask it, and not have `this.selected` changing to a value that wasn't requested.

nanaya: so, throw exception instead of ignoring it?

notbakaneko: It doesn't throw so chat doesn't become unusable if something triggers a race condition for the user.

nanaya: 🤔 so it's expected to receive an invalid channel id? and then assign it to then re-check it whenever it's used?

resources/js/chat/chat-state-store.ts
```diff
@@ -121,50 +175,27 @@ export default class ChatStateStore implements DispatchListener {
...
     // Marking as read is done here to avoid constantly sending mark-as-read requests
     // while receiving messages when autoScroll is enabled on the channel.
     this.selectedChannel?.throttledSendMarkAsRead();
-
     this.selected = channelId;
 
-    if (channelId === 'create') {
+    if (channelId == null) return;
+
+    if (typeof channelId === 'string') {
```
nanaya, line 188: I just noticed the page title is kinda wrong if coming from a different channel...

resources/css/bem/chat-join-channel.less
```diff
@@ -0,0 +1,42 @@
...
+    }
+  }
+
+  &__channels {
+    display: grid;
+    gap: 10px;
+    padding: 10px;
+
+    // checkmark is 14px wide
+    grid-template-columns: minmax(14px, max-content) 200px auto;
```
nanaya, line 31: `... auto 1fr;` is better? (at least on mobile)

resources/js/chat/join-channels.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  let statusElement: React.ReactNode | undefined;
+  if (status === 'joined') {
+    statusElement = <i className='fas fa-check' />;
+  } else if (status === 'joining') {
+    statusElement = <Spinner />;
+  }
+
+  return (
+    <button className={classWithModifiers('chat-join-channel__channel', { joined: status === 'joined' })} onClick={handleClick}>
```
nanaya, line 36: should this switch to the channel if already joined? (instead of doing nothing)

nanaya (changes requested):

resources/css/bem/chat-join-channel.less
nanaya: and as subgrid is a thing now, might as well use it so the button can cover the entire area (instead of having random gaps) and have hover effect

notbakaneko: guess subgrid should be fine if safari supports it now 🤔

resources/js/chat/chat-state-store.ts
```diff
@@ -96,15 +109,57 @@ export default class ChatStateStore implements DispatchListener {
...
+      if (channelId != null) {
+        if (this.channelStore.channels.has(channelId)) {
+          this.selectChannel(channelId);
+          return;
+        }
+
+        showImmediateLoadingOverlay();
+
+        this.waitAddChannelId = channelId;
+        await joinChannel(channelId, core.currentUserOrFail.id);
```
nanaya, line 143: fun behavior 🤔 

0. add sleep to `ChannelsController#join` to emulate slow connection
1. go to a page
2. navigate to chat
3. navigate to join channel
4. join something
5. go back (browser) (should go to whatever previous channel was)
6. wait
7. after a while it switches channel (maybe expected but seems a bit rude?)

Or instead of wait at 6:

6. go back again (to whatever initial page was)
7. wait backend sleep to finish
8. go forward (to chat page)
9. select join channel
10. try joining any channel
11. nothing happens (not sure what happens here)

nanaya (changes requested):

resources/js/chat/main-view.tsx
```diff
@@ -28,6 +29,17 @@ export default class MainView extends React.Component<Record<string, never>> {
...
       }),
     );
+
+    disposeOnUnmount(
+      this,
+      autorun(() => {
+        if (core.dataStore.chatState.isAddingChannel) {
+          showImmediateLoadingOverlay();
+        } else {
+          hideLoadingOverlay();
```
notbakaneko, line 39: I know there is an issue when going back/forward and then returning, but I'm not planning to fixing it in this PR.

resources/css/bem/chat-join-channel.less
```diff
@@ -0,0 +1,47 @@
...
+
+  &__channel {
+    .reset-input();
+
+    display: grid;
+    grid-template-columns: subgrid;
+    grid-column: span 3;
+    color: hsl(var(--hsl-l3));
+    text-align: left;
+    padding: 5px 10px;
```
nanaya, line 19: looks like this causes overflow in firefox when the element type is button 🤨 seems fine with anchor or div at least

nanaya (changes requested):

resources/js/chat/conversation-list.tsx
```diff
@@ -19,7 +19,7 @@ const icons: Record<SupportedChannelType, string> = {
 
 function renderChannels(type: SupportedChannelType) {
   const channels = core.dataStore.channelStore.groupedChannels[type];
-  if (channels.length > 0 || type === 'ANNOUNCE' && core.dataStore.chatState.canChatAnnounce) {
+  if (channels.length > 0 || type ==='PUBLIC' ||type === 'ANNOUNCE' && core.dataStore.chatState.canChatAnnounce) {
```
nanaya, line 22: might as well fix the missing space here

resources/js/chat/join-channels.tsx
```diff
@@ -0,0 +1,136 @@
...
+  let statusElement: React.ReactNode | undefined;
+  if (status === 'joined') {
+    statusElement = <i className='fas fa-check' />;
+  } else if (status === 'joining') {
+    statusElement = <Spinner />;
+  }
+
+  return (
+    // anchor instead of button due to Firefox having an issue with button padding in subgrid.
+    <a className={classWithModifiers('chat-join-channel__channel', { joined: status === 'joined' })} onClick={handleClick}>
```
nanaya, line 37: anchor without href isn't all that different from just using span... (most notably no tabstop, also wrong cursor)

nanaya (changes requested):
I thought I posted it as normal comment ?_?

nanaya (approved):

PR #10836, opened 2023-12-28
React typing fixes

notbakaneko:
Some updates of the react types to prepare for updating to React 18.

`ReactChild` and `ReactFragment` are deprecated, `ReactFragment` not even actually a `Fragment`

### DIFF
--- a/resources/js/components/beatmapset-event.tsx
+++ b/resources/js/components/beatmapset-event.tsx
@@ -114,11 +114,11 @@ export default class BeatmapsetEvent extends React.PureComponent<Props> {
   }
 
   private content() {
-    let discussionLink: React.ReactElement | string = '';
-    let discussionUserLink: React.ReactElement | string = '[unknown user]';
-    let text: React.ReactElement | string = '';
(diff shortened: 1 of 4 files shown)


resources/js/components/beatmapset-event.tsx
```diff
@@ -114,11 +114,11 @@ export default class BeatmapsetEvent extends React.PureComponent<Props> {
...
   private content() {
-    let discussionLink: React.ReactChild = '';
-    let discussionUserLink: React.ReactChild = '[unknown user]';
-    let text: React.ReactChild = '';
+    let discussionLink: React.ReactElement | string = '';
+    let discussionUserLink: React.ReactElement | string = '[unknown user]';
+    let text: React.ReactElement | string = '';
     let url = '';
-    let user: React.ReactChild | undefined;
+    let user: React.ReactElement | string | undefined;
```
nanaya, line 121: might as well just use `React.ReactNode`? That's what's these variables end up being used for later in this function anyway

nanaya (changes requested):

nanaya (approved):

PR #10842, opened 2024-01-03
Queue scores jobs on non-default queue

notbakaneko:
So they don't block the default queue while other stuff is being done and they can be run later by running that queue, instead.

I didn't bother updating any of the tests to run the queue.


notbakaneko:
not needed now

PR #10844, opened 2024-01-04
Temporarily disable forum topics page count

notbakaneko:


bdach:
Not sure it's helpful but just in case, can confirm that this appears to eliminate the problematic query on a local checkout.

PR #10849, opened 2024-01-09
Update `BanchoStats` to include `users_lazer` for online user count(s)

### LINKED ISSUE #10816: Home page user count should include lazer users
Currently, it's sourcing from an `osu_counts` row that bancho is responsible for.

Instead, it should add up all users across all active builds (`allow_bancho = 1`) and probably cache once a minute or so. Probably quite a simple one:

```
root@db-master:osu> select sum(users) from osu_builds where allow_bancho = 1;
+------------+
| sum(users) |
+------------+
| 8323       |
+------------+
```

### LINKED ISSUE #10991: User count on home page doesn't include lazer players
Supersedes and closes https://github.com/ppy/osu-web/pull/10849.

osu-web should add a migration:

```sql
alter table osu_banchostats add users_lazer mediumint unsigned default 0 not null after users_osu, algorithm=inplace;
```

This column will be updated by bancho, to keep things simple. It is using the same logic as `osu-server-spectator` to aggregate user totals. It will be deployed on monday.

A proposed (untested) osu-web patch would be something like:

```diff
diff --git a/app/Libraries/CurrentStats.php b/app/Libraries/CurrentStats.php
index c3b163601c..2eecacd4d8 100644
--- a/app/Libraries/CurrentStats.php
+++ b/app/Libraries/CurrentStats.php
@@ -25,7 +25,7 @@ class CurrentStats
             $latest = array_last($stats);
 
             return [
-                'currentOnline' => $latest['users_osu'] ?? 0,
+                'currentOnline' => $latest['users_osu'] + $latest['users_lazer'] ?? 0,
                 'currentGames' => $latest['multiplayer_games'] ?? 0,
                 'graphData' => array_to_graph_json($stats, 'users_osu'),
                 'totalUsers' => Count::totalUsers()->count,
diff --git a/app/Models/BanchoStats.php b/app/Models/BanchoStats.php
index c9a3f01f6d..a8460baa1b 100644
--- a/app/Models/BanchoStats.php
+++ b/app/Models/BanchoStats.php
@@ -23,7 +23,7 @@ class BanchoStats extends Model
     {
         return array_reverse(
             static::whereRaw('banchostats_id mod 10 = 0')
-                ->select(['users_irc', 'users_osu', 'multiplayer_games', 'date'])
+                ->select(['users_irc', 'users_osu + users_lazer', 'multiplayer_games', 'date'])
                 ->orderBy('banchostats_id', 'DESC')
                 ->limit(24 * 60 / 10)
                 ->get()

```

tsunyoku:
Resolves #10991.

Something to note is that this doesn't change the user counts used for the user graph, since this works by Bancho inserting user counts on an interval and that doesn't seem possible with the `osu_builds` table.

The issue mentions caching the user count for around a minute, however the homepage is already doing a 5 minute cache on all of the stats so I don't think it's necessary.

### DIFF
--- a/app/Models/Build.php
+++ b/app/Models/Build.php
@@ -252,7 +252,7 @@ class Build extends Model implements Commentable
         // In order to not double-count users, we must use the "main" releases which will never have a stream_id.
         return static
             ::where('allow_bancho', 1)
-            ->whereNull('stream_id')
+            ->whereNotNull('stream_id')
             ->sum('users');
     }
 }
--- a/app/Models/Build.php
+++ b/app/Models/Build.php
@@ -248,6 +248,11 @@ class Build extends Model implements Commentable
 
     public static function onlineUsers()
     {
-        return static::where('allow_bancho', 1)->sum('users');
+        // Lazer builds uses "fake" release rows to track platform-specific builds
+        // In order to not double-count users, we must use the "main" releases which will never have a hash.
+        return static
+            ::where('allow_bancho', 1)
+            ->whereNull('hash')
+            ->sum('users');
     }
 }
--- a/app/Models/Build.php
+++ b/app/Models/Build.php
@@ -245,14 +245,4 @@ class Build extends Model implements Commentable
     {
         return preg_replace('#[^0-9.]#', '', $this->version);
     }
-
-    public static function onlineUsers()
-    {
-        // Lazer builds uses "fake" release rows to track platform-specific builds
-        // In order to not double-count users, we must use the "main" releases which will always have a stream_id.
-        return static
-            ::where('allow_bancho', 1)
-            ->whereNotNull('stream_id')
-            ->sum('users');
-    }
 }
--- a/app/Models/Build.php
+++ b/app/Models/Build.php
@@ -249,10 +249,10 @@ class Build extends Model implements Commentable
     public static function onlineUsers()
     {
         // Lazer builds uses "fake" release rows to track platform-specific builds
-        // In order to not double-count users, we must use the "main" releases which will never have a hash.
+        // In order to not double-count users, we must use the "main" releases which will never have a stream_id.
         return static
             ::where('allow_bancho', 1)
-            ->whereNull('hash')
+            ->whereNull('stream_id')
             ->sum('users');
     }
 }
--- a/app/Models/Build.php
+++ b/app/Models/Build.php
@@ -249,7 +249,6 @@ class Build extends Model implements Commentable
 
     public static function onlineUsers()
     {
-        return static::where('allow_bancho', 1)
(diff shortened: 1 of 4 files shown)


app/Models/Build.php
```diff
@@ -235,4 +235,9 @@ public function displayVersion()
     {
         return preg_replace('#[^0-9.]#', '', $this->version);
     }
+
+    public static function onlineUsers()
+    {
+        return static::where('allow_bancho', 1)->sum('users');
```
bdach, line 241: I'm not sure this is correct. I think it'll double-count lazer users.

The reason for that is that lazer build tracking is a little wonky. See [this discord message](https://discord.com/channels/188630481301012481/188630652340404224/1184782370105200753) for an example, but in short, for every _actual_ lazer release:

- there will be a "main" lazer release row with the actual version, with `stream_id` set correctly to 7, but with no hash,
- and there will be "fake" lazer release rows, one per platform, with `null` `stream_id` but hashes set to match the appropriate platform.

And notably all of them will have `allow_bancho` set to 1.

I'd recommend looking through https://github.com/ppy/osu-server-spectator/pull/204 to see how `osu-server-spectator` handles that right now. Maybe we want to clean this up at some point (maybe even now), but I'd rather point this out now before this PR goes too far.

tsunyoku: Ah, thanks for that. I've updated to add an extra `where` that ensures `hash` is `null`. I thought about adding an extra `where` to ensure that `stream_id` is `7` however I'm unsure if this is correct? In the PR linked, the tests are using any build ids while `GetAllMainLazerBuildsAsync` is doing `WHERE stream_id = 7` so would appreciate some clarification there.

bdach: I'd say that basing it on stream not being null is more likely correct than on null hash. And I'm not sure how the build counts look when it comes to stable releases. Probably needs @peppy double-check.

tsunyoku: That makes more sense. I've applied that, but will keep this open awaiting a double-check from peppy regarding stable.

peppy: I don't think the main lazer build will ever get a hash, but either sounds fine for now.

tsunyoku: Sounds good, any concerns regarding stable?

nanaya (dismissed):

nanaya:
On hold since the graph would still be outdated. I think @peppy (or someone other) is supposed to take a look at updating whatever handling the osu_banchostats table instead.

nanaya:
@tsunyoku the updated system is outlined here: #10991.

tsunyoku:
Adjusted PR to meet new requirements.

nanaya (approved):

PR #10851, opened 2024-01-10
Update to php 8.3

nanaya:
Seems working.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -94,7 +94,7 @@ jobs:
   tests:
     strategy:
       matrix:
-        php: ['8.2']
+        php: ['8.3']
     name: Tests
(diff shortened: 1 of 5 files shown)


peppy (approved):

notbakaneko (commented):
needs `php8.3-ds` `1.5.0` which isn't up on `packages.sury.org` yet (and test runner passes because ubuntu)

```
PHP Warning:  PHP Startup: Unable to load dynamic library 'ds.so' (tried: /usr/lib/php/20230831/ds.so (/usr/lib/php/20230831/ds.so: undefined symbol: fast_add_function), /usr/lib/php/20230831/ds.so.so (/usr/lib/php/20230831/ds.so.so: cannot open shared object file: No such file or directory)) in Unknown on line 0
```

PR #10852, opened 2024-01-10
Update ziggy

notbakaneko:
Originally looked at updating this to get the route typings and then discovering ziggy doesn't export the parameter typings so there's still the workaround typings for the wrapper (because we're using `Ziggy` as an export, not global). 
~~We can't use the `Parameters<T>` utility type to infer `route` argument types either because it's declared as an overload and it only uses the last one...so I'm not sure what to do with it that doesn't involve just copying their whole `d.ts` and adding `export` ಠ_ಠ~~
Turns out the generic type was important 👀

### DIFF
--- a/resources/js/ziggy.d.ts
+++ b/resources/js/ziggy.d.ts
@@ -2,14 +2,12 @@
 // See the LICENCE file in the repository root for full licence text.
 
 declare module 'ziggy' {
-  interface ZiggyClass {
+  // Reduced version of unexported ziggy Config for typing when setting the port and url.
+  // Using Parameters<typeof route>[3] doesn't quite work since route has overloaded parameters.
+  interface ZiggyGlobal {
     port: number | null;
     url: string;
   }
 
-  export const Ziggy: ZiggyClass;
-}
-
-declare module 'ziggy-route' {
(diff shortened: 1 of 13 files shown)


resources/js/ziggy.d.ts
```diff
@@ -2,14 +2,23 @@
 // See the LICENCE file in the repository root for full licence text.
 
 declare module 'ziggy' {
-  interface ZiggyClass {
+  // Reduced version of unexported ziggy Config for typing when setting the port and url.
+  // Using Parameters<typeof route>[3] doesn't quite work since route has overloaded parameters.
+  interface ZiggyGlobal {
```
nanaya, line 7: this?

```typescript
  import route, { RouteList } from 'ziggy-js';

  export const Ziggy: NonNullable<Parameters<typeof route<keyof RouteList>>[3]>;
```

nanaya (changes requested):

resources/js/ziggy.d.ts
notbakaneko: Turns out I never tried `<keyof RouteList>` here since I only started using it after trying to strip the parameter for the wrapper 🤔

PR #10853, opened 2024-01-10
Add client hash check

nanaya:
Also applies to user registration.


app/Http/Controllers/Multiplayer/Rooms/Playlist/ScoresController.php
```diff
@@ -205,6 +207,7 @@ public function update($roomId, $playlistItemId, $tokenId)
         $transformer = ScoreTransformer::newSolo();
         if ($score->wasRecentlyCreated) {
             $scoreJson = json_item($score, $transformer);
+            ClientCheck::queueToken($clientTokenData, $score->getKey());
```
notbakaneko, line 210: Test should also check that token is queued, also for the other one

notbakaneko (changes requested):

notbakaneko (approved):

PR #10854, opened 2024-01-10
More useful return type for user score legacy preference

nanaya:
As it turned out the original value is a bit cumbersome to use.


notbakaneko (commented):
where does it help if it's true/null instead of true/false?

notbakaneko (approved):

PR #10856, opened 2024-01-11
Revert back to php 8.2

nanaya:
More waiting for the [php repository to be updated](https://github.com/oerdnj/deb.sury.org/issues/2063).

Either that or replace the thing with alpine instead.

(partial revert - the updated packages are kept)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -94,7 +94,7 @@ jobs:
   tests:
     strategy:
       matrix:
-        php: ['8.3']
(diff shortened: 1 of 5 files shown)


notbakaneko (commented):
I'd rather switch to ubuntu and match the github runner versions

notbakaneko (approved):

nanaya:
ubuntu wouldn't help since it'd be the same php repository. not only that, they used to break jhead and their chromium is some snap package

PR #10865, opened 2024-01-13
Support displaying F rank and make it default for fail scores

### LINKED ISSUE #10524: Failed solo score handling needs reconsidering with respect to stored score rank
[Back in the olden days](https://github.com/ppy/osu/pull/13926), when client would submit a failed play, it would set D rank, because we didn't have F rank client-side yet.

Since https://github.com/ppy/osu/pull/18785, we now have an F rank client-side. But this logic still lives:

https://github.com/ppy/osu-web/blob/f7ef5553dc31e2a2dd7ea43644fd092bfe93091a/app/Models/Solo/Score.php#L57-L62

and overwrites the F rank sent by client to D again.

You would say that, well, no biggie, just do this:

```diff
diff --git a/app/Models/Solo/Score.php b/app/Models/Solo/Score.php
index 875ade1318..284aff7c04 100644
--- a/app/Models/Solo/Score.php
+++ b/app/Models/Solo/Score.php
@@ -60,7 +60,7 @@ class Score extends Model implements Traits\ReportableInterface, Traits\SoloScor
         // older lazer builds potentially submit incorrect details here (and we still want to
         // accept their scores.
         if (!$score->data->passed) {
-            $score->data->rank = 'D';
+            $score->data->rank = 'F';
         }
 
         $score->saveOrExplode();

```

which, sure, fixes that immediate part of it, but then you attempt to go to `/scores/{score_id}` web-side and see this:

Note missing rank letter, note rank badges on the left all lit up. So that is at least _one_ place that is not ready web-side to handle an F rank. Maybe there are more.

I guess the things to discuss here are:

- Should failed scores be even displayable via `/scores/{score_id}`?
- Maybe something to do with the `preserve` flag makes this issue go away?
- If not, does this need design? Which other places may need to handle an F rank?

nanaya:
~~No small icon but I don't think it's visible anywhere...?~~ Added the thing. Not in the score page icons "tower" as it's weird "achieving" failed rank.

Resolves #10524

### DIFF
--- a/app/Models/Solo/Score.php
+++ b/app/Models/Solo/Score.php
@@ -64,7 +64,7 @@ class Score extends Model implements Traits\ReportableInterface
         // older lazer builds potentially submit incorrect details here (and we still want to
         // accept their scores.
         if (!$score->data->passed) {
-            $score->data->rank = 'D';
+            $score->data->rank = 'F';
         }
 
         $score->saveOrExplode();
--- a/public/images/badges/score-ranks-v2019/GradeSmall-F.svg
+++ b/public/images/badges/score-ranks-v2019/GradeSmall-F.svg
@@ -1,0 +1,22 @@
+<svg width="32" height="16" viewBox="0 0 32 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<g clip-path="url(#clip0_8410_3113)">
+<rect width="32" height="16" rx="8" fill="#3F3F3F"/>
+<g style="mix-blend-mode:darken" opacity="0.75">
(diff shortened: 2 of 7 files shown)


notbakaneko (changes requested):
What about the icons on the user profile page?

nanaya:
I don't think it's visible there? The only way to make it visible is by specifically requesting failed scores but the web doesn't have that option iirc.

notbakaneko:
but is it possible to go to the score url and pin it?

nanaya:
hmm right. well 🤷‍♀️

notbakaneko (approved):

PR #10866, opened 2024-01-14
Fix the chat not checking the month and the year before adding a header with the date on it

Corne2Plum3:
I sent a private message to someone on May 14th 2023, then sent another today (January 14th) on the next year, but the message, separated from 8 months, are showing up like it was the same day.

image

I somehow discovered that when adding the date headers, only the date number is checked, and not the month and the year. So with this fix, the month and year is also checked.

Is this what they meant by "making this less dodgy"?

Used the documentation of Moment.js to find the correct functions btw


resources/js/chat/conversation-view.tsx
```diff
@@ -66,13 +68,15 @@ export default class ConversationView extends React.Component<Props> {
       }
 
       // check whether the day-change header needs to be shown
-      if (isEmpty(conversationStack) || moment(message.timestamp).date() !== currentDay /* TODO: make check less dodgy */) {
+      if (isEmpty(conversationStack) || (moment(message.timestamp).date() !== currentDay && moment(message.timestamp).month() !== currentMonth && moment(message.timestamp).year() !== currentYear)) {
```
bdach, line 71: Surely this cannot be the best way to do this?

Seems like something along the lines of `moment().isSame(message.timestamp, 'day')` may work better ([see docs](https://momentjs.com/docs/#/query/is-same/))?

Corne2Plum3: As `message.timestamp` is generated, using `moment().toISOString()`, it works.

Thanks.

I'm going to do that.

resources/js/chat/conversation-view.tsx
```diff
@@ -66,13 +65,12 @@ export default class ConversationView extends React.Component<Props> {
       }
 
       // check whether the day-change header needs to be shown
-      if (isEmpty(conversationStack) || moment(message.timestamp).date() !== currentDay /* TODO: make check less dodgy */) {
+      if (isEmpty(conversationStack) || moment().isSame(message.timestamp, 'day')) {
```
notbakaneko, line 68: This is supposed to compare with the timestamp of the previous message (store `moment(message.timestamp)` instead of `.date()` in `currentDay` and use `currentDay.isSame`), not the current point in time

or just remove `moment`:
```
new Date(message.timestamp).toLocaleDateString() !== currentDay
.
.
currentDay = new Date(message.timestamp).toLocaleDateString();
```

notbakaneko (changes requested):

resources/js/chat/conversation-view.tsx
Corne2Plum3: I also forgot a logical not lol

Corne2Plum3: Should be fixed, now I'm comparing the good stuff and I'm also added a `!` when using `isSame()`, as we want `true` when we got 2 different days.

Corne2Plum3: I forgot to check if currentDay isn't defined bruh, and in this situation the output is incorrect...

Corne2Plum3: Finally I did what you suggested, the thing I was doing ignored the timezone, which shouldn't be ignored in this context.

notbakaneko (changes requested):
remove the now unused `moment` import

Corne2Plum3:
Done.

notbakaneko (approved):

Issue #10801, opened 2023-12-15
Qualified maps that are blocked from getting ranked need a visible warning


Hiviexd:
currently, these kind of maps lack visibility which is causing nats (who are supposed to swiftly address them) a hard time keeping track of them, and it's causing complaints among players/mappers because of unexpected rank delays because of not understanding the feature / not being aware of it.

the following text could be an acceptable approach, though it'd be preferred to have some kind of a more visible warning in the modding page.

![image](https://github.com/ppy/osu-web/assets/62819481/d3d1de0f-f850-4900-9636-5f497bc8f590)

i'd really appreciate if this could be prioritized a bit because of the UX concerns mentioned above.

tsunyoku:
could you expand on what actually "blocks" a map from being qualified? i'm unsure if it should display if there's any open suggestions or if it's only for open issues

Hiviexd:
any unresolved suggestion or problem post blocks the map from leaving qualified.

PR #10872, opened 2024-01-16
Display a warning on discussions page if a qualified map has unresolved discussions that are blocking it from ranked

tsunyoku:
Resolves https://github.com/ppy/osu-web/issues/10801.

image

Clicking `these problems` will filter to only display pending discussions and will jump you to the first one. Intentionally larger so that I was asked to make sure it stands out.


resources/js/beatmap-discussions/nominations.tsx
```diff
@@ -634,12 +634,36 @@ export class Nominations extends React.PureComponent<Props> {
             : formatDate(rankingEta)
           : trans('beatmaps.nominations.rank_estimate.soon');
 
+        const unresolvedProblems = this.props.beatmapset.discussions.filter((d) => !d.resolved && ['problem', 'suggestion'].includes(d.message_type));
```
notbakaneko, line 637: Use `this.props.discussionsState.discussionsByFilter.pending`

resources/js/beatmap-discussions/nominations.tsx
```diff
@@ -634,12 +634,36 @@ export class Nominations extends React.PureComponent<Props> {
...
+        const rankEstimatePrefixText = unresolvedProblems.length > 0
+          ? trans('beatmaps.nominations.rank_estimate.unresolved_problems')
+          : trans('beatmaps.nominations.rank_estimate.no_problems');
+
+        const problems = unresolvedProblems.map((problem) => (
```
notbakaneko, line 642: Should limit the maximum number of issues displayed to avoid a runoff of issues.

resources/js/beatmap-discussions/nominations.tsx
```diff
@@ -634,12 +634,36 @@ export class Nominations extends React.PureComponent<Props> {
...
+          <a
+            key={problem.id}
+            href={makeUrl({ discussion: problem })}
+            rel='noreferrer'
+            target='_blank'
+          >
+            #{problem.id}
+          </a>
```
notbakaneko, line 650: Use the same format as https://github.com/ppy/osu-web/blob/a7f53f734fc847a924611ebf6b964863122a2362/resources/js/beatmap-discussions/nominations.tsx#L261
It's strange if this link behaves differently

resources/js/beatmap-discussions/nominations.tsx
tsunyoku: hm, how would you suggest using this...? `this.props.currentDiscussions.byFilter.pending` is ` Record<"reviews" | "generalAll" | "general" | "timeline", Partial<Record<number, BeatmapsetDiscussionDefault>>>` so it doesn't work as nicely

notbakaneko: `discussionsState.discussionsByFilter.pending` is new, pull the latest changes from `master`

resources/js/beatmap-discussions/nominations.tsx
notbakaneko: hmm, then again, considering the message is for the mapper, maybe it's better to just list them all out 🤔 
I'd consider looking at a way to enable the `pending` filter to make it easier to find the discussions that need resolving, though.

tsunyoku: after a conversation with hivie, it seems that the warning should be separated and apply a pending filter. i'll make some changes and then resolve this

tsunyoku: updated. check PR description for updated functionality

cl8n:
for the design I think a color change (orange/yellow warning color or somethn) would feel less out of place here, and probably stand out even more than changing font size

notbakaneko (changes requested):
I agree a different colour instead of a font size change is probably a better idea, maybe `hsl(var(--hsl-orange-2))`; it looks like any other info text in the box at the moment

notbakaneko (approved):

PR #10874, opened 2024-01-18
Fix browser test on docker dev

nanaya:
This seems to work. Some more clean up is still needed to make it not randomly fail due to leftover data but that's for another day (also for normal phpunit tests).


notbakaneko (approved):
Turns out my setup has been going though nginx this whole time 👀

PR #10876, opened 2024-01-19
Add message on forced password reset

nanaya:
Also fixed bug where ujs redirect conflicts with user login handler.


notbakaneko (commented):
unrelated to this, but I wonder if we should also start taking into account `user_country_history` as well? 🤔

notbakaneko (approved):

PR #10879, opened 2024-01-23
Fix colors of links in changelogs

### LINKED ISSUE #9656: links in changelog text don't look like links
I didn't immediately recognize that "gifting a supporter tag", the contest, and the nekodex song were links, and thought they were just bold text instead:

https://osu.ppy.sh/home/changelog/stable40/20221220.1

image

They also stand out a bit too much in my opinion, since all the other changelog detail text is so muted. Changing to the regular purple link color (or even a more muted version of it) would probably help with both of these issues.

HonKLam:
*(Still very new to open source PRs. Any feedback to improve them is appreciated!)*

Resolves #9656

Since changelogs now use the `changelog-entry` CSS-class that already fixed this issue a long time ago, I wanted to fix this problem for previous posts that used the `changelog-md` CSS-class and match their colors for better readability.

I was not able to fetch the original changelogs to my dev-environment (I think you aren't supposed to either way), so I created my own elements and applied the class to test it:

image


resources/css/bem/changelog-md.less
```diff
@@ -26,6 +26,14 @@
     }
   }
 
+  &__link {
```
nanaya, line 29: just removing the `a` styling is enough (or more like it's overriding this styling and this change doesn't do anything)

also
> Since changelogs now use the `changelog-entry` CSS-class that already fixed this issue a long time ago, I wanted to fix this problem for previous posts that used the `changelog-md` CSS-class and match their colors for better readability.

they're for different part of the changelog and still used.

nanaya (changes requested):

resources/css/bem/changelog-md.less
HonKLam: Removed the a-styling and the part of `&__link` that would override the a-styling. Hover-part is staying to correctly match colors with `changelog-entry`. Hope that's fine now!

nanaya (changes requested):
after some discussions, these changes should be applied instead:

- links should have the default colours, including the `changelog-entry__link` (no white colour on hover)
- `changelog-md` should have default colour (just remove the definition)
- `changelog-entry__row--message` should have top margin of 8px instead of 2px

resources/css/bem/changelog-entry.less
```diff
@@ -16,19 +16,11 @@
...
   }
 
-  &__link {
-    color: @osu-colour-l1;
-
-    &:hover {
-      color: white;
-    }
-  }
-
```
HonKLam, line 26: Tested it - you can remove everything here to get default colored links.

resources/css/bem/changelog-entry.less
```diff
@@ -16,19 +16,11 @@
...
   &__row {
     word-wrap: break-word;
 
     &--message {
-      margin-top: 2px;
+      margin-top: 8px;
```
HonKLam, line 23: Message looks like this now:

image

nanaya: the custom text colour for changelog-md still needs to be removed (which the original reason the link is coloured white)

HonKLam: I hope I understood it correctly, but it should be fixed now.

From this:

image

To this:

image

(only using `changelog-md`-class)

nanaya (approved):

PR #10881, opened 2024-01-23
Deprecate active_tournament_banner property in User API response

notbakaneko:
Should use `active_tournament_banners` for the list of banners instead

### DIFF
--- a/resources/views/docs/_structures/user.md
+++ b/resources/views/docs/_structures/user.md
@@ -24,7 +24,7 @@ default_group   | string?                   | Identifier of the default [Group](
 id              | integer                   | unique identifier for user
 is_active       | boolean                   | has this account been active in the last x months?
 is_bot          | boolean                   | is this a bot account?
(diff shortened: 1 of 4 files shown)


resources/views/docs/_structures/user.md
```diff
@@ -41,7 +41,8 @@ Following are attributes which may be additionally included in the response. Rel
 Field                      | Type
 ---------------------------|-----
 account_history            | [User.UserAccountHistory](#user-useraccounthistory)[]
-active_tournament_banner   | [User.ProfileBanner](#user-profilebanner)?
+active_tournament_banner   | Deprecated, use `active_tournament_banners` instead. [User.ProfileBanner](#user-profilebanner)?
```
nanaya, line 44: why is the note in type column 🤨

nanaya (changes requested):

nanaya (approved):

PR #10884, opened 2024-01-23
Add option to disable score submission

nanaya:
Instead of current workaround of removing tokens table and/or having outdated structure.


notbakaneko (commented):
Do we need this in production or is it for the pre-production score stuff?

nanaya:
apparently yes

"might be handy"

notbakaneko (approved):

peppy (approved):

PR #10887, opened 2024-01-24
Update score table structure

nanaya:
Reopening #10869 because I pushed wrong branch to master 🕺 

This uses raw sql for migration due to laravel not supporting float until laravel 11. Also the syntax just not being very helpful in the first place (if anything the migration result has been occasionally changed between laravel versions).

Indexer image points to my own published image because the changes need ppy/osu-elastic-indexer#154 to be properly tested.

I considered making everything flat - as in handling maximum_statistics, statistics, and mods directly from Score model - but it'd involve even more changes dealing with partial ScoreData assignment so maybe later if at all.

- [x] depends on #10867 (although not sure if that much meaningful being a rather small change by itself)
- [x] migration entry `2024_01_12_115738_update_scores_table_final`


peppy (approved):

nanaya:
(already approved in original pr and no change since)

PR #10888, opened 2024-01-24
Add client hash check

nanaya:
Reopening #10853 🕺

~~Also applies to user registration.~~

- [x] #10887


notbakaneko (dismissed):

nanaya:
Undid the check for registration as it'll be on a separate PR since it involves a bit more than just adding the check.

peppy (approved):

PR #10889, opened 2024-01-24
Add menu to switch score display mode

nanaya:
Reopening #10833


resources/views/layout/_score_mode_toggle.blade.php
```diff
@@ -0,0 +1,24 @@
+{{--
+    Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+    See the LICENCE file in the repository root for full licence text.
+--}}
+@php
+    $legacyScoreOnlyValue = App\Libraries\Search\ScoreSearchParams::showLegacyForUser(Auth::user());
+    $icon = $legacyScoreOnlyValue
+        ? 'far fa-square'
```
tsunyoku, line 8: shouldn't this be:

```suggestion
        ? 'fas fa-square'
```

nanaya: no because it'll look like it's enabled

notbakaneko (dismissed):

peppy (approved):

PR #10903, opened 2024-01-29
Add ranked field in score params

nanaya:
As requested to help score processing.


notbakaneko (approved):

notbakaneko:
I'm assuming it won't explode or anything if merged 👀

PR #10913, opened 2024-01-30
Limit recent scores to 48h

### LINKED ISSUE #10912: "Recent plays" section on website shows scores older than 24h

nanaya:
This will roll through all the user's scores but apparently it's already doing something like it anyway due to the user index in prod has a lot more columns in it (instead of just user_id and ruleset_id). So maybe it'll be fine? 🤷‍♀️

One of the worst cases is going through something like 200k rows.

Resolves #10912.


notbakaneko (changes requested):
I think the count returned needs to be updated too? 🤔

notbakaneko (approved):

PR #10918, opened 2024-01-30
Change display for scores not awared pp

### LINKED ISSUE #10911: Scores set with mods which don't support pp show as "This score is still being calculated"
Example (ignore that this actually has no mods):

Safari 2024-01-30 at 07 06 44

They should say something more like "pp was not awarded for this score due to invalid beatmap or mods".

notbakaneko:
closes #10911


nanaya (changes requested):

nanaya:
Maybe the pp value render logic should just be split between legacy and solo score.

nanaya (approved):

PR #10930, opened 2024-01-31
Validate score token creation time

### LINKED ISSUE #10924: Beatmap state change after token provision should cause score submission to fail

nanaya:
Resolves #10924.


peppy:
best to get this in as soon as we can 😄

notbakaneko (approved):

PR #10944, opened 2024-02-01
Use logo without shadow to fix blurriness on Safari

peppy:


frenzibyte:
Is the "triangles" part supposed to have the logo label on it? The previous version doesn't have one.

peppy:
That might explain the thickness.

PR #10949, opened 2024-02-01
Increment replays watched count when downloading a user's score

### LINKED ISSUE #10948: Replay download requests should increment "Replays Watched by Others" count on profile
Also known as `` `osu_user_stats_*`.`replay_popularity` ``.

Probably the only feasible way this can work for lazer scores. In stable this is handled by osu-web-10.

tsunyoku:
Resolves #10948.

Modified `ScoresController` tests to check it was incremented.


tsunyoku:
drafting while I fix tests (local workflow is fucked)

app/Http/Controllers/ScoresController.php
```diff
@@ -54,6 +58,8 @@ public function download($rulesetOrSoloId, $id = null)
             'Content-Type' => 'application/x-osu-replay',
         ];
 
+        $score->user->statistics($ruleset, true)->increment('replay_popularity');
```
cl8n, line 61: this should be skipped if the user is downloading their own replay

cl8n, line 61: there are some other places tracking replay counts that should be updated at the same time:
- monthly watch statistics - insert or increment `$score->user->replaysWatchedCounts` for the current `year_month`
- the replay's individual replay count - for legacy scores increment `$score->replayViewCount`, for lazer scores idk if it even exists...

cl8n (changes requested):

app/Http/Controllers/ScoresController.php
bdach: For what it's worth - as far as I can tell - the two above should be about it, looking at `osu-web-10` source. The point about `osu_replays_*` not really applying for lazer scores is valid though. Will probably need to be ran by @peppy (do we do a separate table? attempt to add a column to current? maybe put it in the `data` json blob?)

tsunyoku: I'll apply the first one, and the second one if it's a legacy score but will wait on a response for lazer scores.

app/Http/Controllers/ScoresController.php
```diff
@@ -54,6 +61,18 @@ public function download($rulesetOrSoloId, $id = null)
...
+
+        if (Auth::user()->user_id !== $score->user->user_id) {
+            $score->user->statistics($ruleset, true)->increment('replay_popularity');
+
+            $currentMonth = UserCountryHistory::formatDate(CountryChangeTarget::currentMonth());
+            $score->user->replaysWatchedCounts->where('year_month', $currentMonth)->first()->increment('count');
+
+            if ($score instanceof ScoreBest) {
+                $score->replayViewCount()->increment('play_count');
+            }
```
nanaya, line 73: the relevant rows need to be created first if not exist yet?

app/Http/Controllers/ScoresController.php
```diff
@@ -54,6 +61,18 @@ public function download($rulesetOrSoloId, $id = null)
             'Content-Type' => 'application/x-osu-replay',
         ];
 
+
+        if (Auth::user()->user_id !== $score->user->user_id) {
+            $score->user->statistics($ruleset, true)->increment('replay_popularity');
+
+            $currentMonth = UserCountryHistory::formatDate(CountryChangeTarget::currentMonth());
```
nanaya, line 68: the `currentMonth` function is rather... unique for the class and not what one may expect

app/Http/Controllers/ScoresController.php
```diff
@@ -54,6 +61,18 @@ public function download($rulesetOrSoloId, $id = null)
...
+            $score->user->replaysWatchedCounts->where('year_month', $currentMonth)->first()->increment('count');
```
nanaya, line 69: this would fetch all rows first before finding the correct one which is rather wasteful (and then explode if the row doesn't exist)

nanaya (changes requested):

app/Http/Controllers/ScoresController.php
tsunyoku: forgive me for not knowing much about user flow, but shouldn't they always exist? if not, which ones might not so I can account for it?

app/Http/Controllers/ScoresController.php
tsunyoku: isn't fetching all rows unavoidable when accessing it with `$score->user->replaysWatchedCounts`? I guess if I query `UserReplaysWatchedCount` directly then I can avoid it

app/Http/Controllers/ScoresController.php
nanaya: I think only `statistics` is guaranteed to exist

app/Http/Controllers/ScoresController.php
nanaya: use `replaysWatchedCounts()` like everywhere else?

tsunyoku: ah, whoops

nanaya (changes requested):

app/Http/Controllers/ScoresController.php
```diff
@@ -54,6 +61,26 @@ public function download($rulesetOrSoloId, $id = null)
...
+                ->where('year_month', $currentMonth)
+                ->firstOrCreate(['year_month' => $currentMonth], ['count' => 0])
+                ->increment('count');
+
+            if ($score instanceof ScoreBest) {
+                if (!$score->replayViewCount()->exists()) {
+                    $score->replayViewCount()->create(['play_count' => 0]);
+                }
+
+                $score->replayViewCount()->increment('play_count');
```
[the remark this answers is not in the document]
tsunyoku: i tried this initially and when testing it gave errors about the expected count being 0 and not 1 🤔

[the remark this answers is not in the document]
nanaya: that's because the replayviewcount model is missing `public $incrementing = false;` setting

nanaya (approved):

PR #10952, opened 2024-02-02
Fix converted score type

nanaya:
Fun 🫥


notbakaneko (commented):
if it passed and doesn't have a`legacy_score_id` it's a high score? How does that work 🤔

notbakaneko (approved):

PR #10955, opened 2024-02-02
Change recent plays time limit text to 48h

Swekka:


Joehuu (approved):

cl8n (approved):
seems right but I'm curious why it was changed to 48 :thinking:

nanaya:
it was originally some performance concern but the count display made it moot. If anything it should probably be reverted to 24h instead.

PR #10958, opened 2024-02-02
Fix error on invalid ruleset

nanaya:
Not sure if should throw error or use default 🤔 (the same controller is already doing both on different places)


notbakaneko (commented):
I feel like returning an error makes more sense, it seems odd to request one thing and have something else returned, especially from an API perspective (not to mention misleadingly thinking something is working when the wrong param is passed in) 🤷

notbakaneko (approved):

PR #10972, opened 2024-02-05
Fix supporter being required for `played=any` in beatmap listing

Swekka:


resources/js/beatmapset-search-filters.ts
```diff
@@ -85,7 +85,7 @@ export class BeatmapsetSearchFilters {
 
   @computed
   get supporterRequired() {
-    return keyNames.filter((key) => this[key] != null && filtersRequireSupporter.includes(key));
+    return keyNames.filter((key) => filtersRequireSupporter.includes(key) && this[key] != null && this[key] !== this.getDefault(key));
```
nanaya, line 88: I have a feeling the update function should reset the value if default instead

@notbakaneko

notbakaneko: It should reset the filter instead, invalid options should also be reset to the default

PR #10976, opened 2024-02-06
Add `speed_note_count` to docs

### LINKED ISSUE #10345: Speed note count not documented
Hi, i'm currently writing a PP calculator using the API, and I noticed the `speed_note_count` attribute in the osu `BeatmapDifficultyAttributes` object ([here](https://osu.ppy.sh/docs/index.html#beatmapdifficultyattributes)) is undocumented. I just wanted to let you know so it can be included in the documentation :)

HonKLam:
# Resolves #10345

API call for `BeatmapDifficultyAttributes` looks like this (in osu):

image

So I added the missing `speed_note_count` to docs and it looks like this now:

image

### I also checked the other gamemodes and found some things off:

# Taiko

API

image

Docs

image

(peak_difficulty vs approach_rate)

# Mania

API 

image

Docs

image

(score_multiplier just missing in general?)

---

Is this intended? Otherwise I can offer to fix it in the docs.

### DIFF
--- a/resources/views/docs/_structures/beatmap_difficulty_attributes.md
+++ b/resources/views/docs/_structures/beatmap_difficulty_attributes.md
@@ -17,6 +17,7 @@ flashlight_difficulty | float
 overall_difficulty    | float
 slider_factor         | float
 speed_difficulty      | float
+speed_note_count      | float
 
 ### taiko
 


nanaya:
> If `BeatmapDifficultyAttributes` should not be documented in osu docs, but can be requested via API, where should it be documented then?

the linked project which the endpoint proxies to

> Could also put in a note in docs that states that `BeatmapDifficultyAttributes` is subject to change, but that applies to every endpoint or output, doesn't it?

the code and documentation for other endpoints are in the same repository whereas this one the code is in another castle

HonKLam:
> the code and documentation for other endpoints are in the same repository whereas this one the code is in another castle

I will leave it to the other castle then to update the docs if it's more fitting there. If it happens, maybe put a note in osu-web docs that refers to it?

Feel free to close PR or keep it open as a reminder

peppy:
Will just close this for now.

Issue #10679, opened 2023-10-25
Inconsistent "perfect combo" display in osu!taiko and osu!mania


Walavouchey:
Assuming that a green combo count on beatmap leaderboards is supposed to indicate the maximum possible combo, there are 3 cases where this is inconsistent. The causes are not obvious without checking the code or extensive testing in the game (nor are they documented).

There are similar inconsistencies with when the stable client displays `Perfect` on the results screen graph, but not sure if they are worth reporting separately.

### Scores on maps with swells/dendens in osu!taiko are never marked "perfect"

Swells/dendens don't affect combo in stable ScoreV1 (they do in ScoreV2) and lazer. (This is the same for drumrolls but they don't have this issue regarding perfect combo.)

[Example map](https://osu.ppy.sh/beatmapsets/422261#taiko/913007) (combo is not marked green):

![image](https://github.com/ppy/osu-web/assets/36758269/3a9ef42d-30de-4221-b50b-39bc66f2f4e1)

### Combo values greater than what Auto gets is possible in osu!mania 

This seems to stem from hold notes giving a variable amount of combo in stable (even when not breaking combo) depending on hit timing. It is not a problem in lazer from what I can tell.

[Example map](https://osu.ppy.sh/beatmapsets/2070252#mania/4331793) (different combo counts are marked green):

![image](https://github.com/ppy/osu-web/assets/36758269/cc70a786-22aa-4ab2-9b13-06aaf03b9d6e)

### "Perfect combo" in osu!mania requires zero 100s and zero 50s

[Example map](https://osu.ppy.sh/beatmapsets/847996#mania/1773372) (scores with 100s or 50s are not marked green, regardless of combo):

![image](https://github.com/ppy/osu-web/assets/36758269/afa1d99d-ea3c-46cf-9b55-4bfdaf0595e3)

This is due to the `perfectCombo` check here:

https://github.com/ppy/osu-web/blob/546c34b640163e171ae9f5ce1175d3901a39a92f/resources/js/utils/score-helper.ts#L40-L58

(There is a [similar and more complicated check](https://github.com/peppy/osu-stable-reference/blob/3ea48705eb67172c430371dcfc8a16a002ed0d3d/osu!/GameModes/Play/Rulesets/Mania/RulesetMania.cs#L201-L206) in stable.)

Not sure what `legacy_perfect` in that osu-web function corresponds to.

nanaya:
[...]

PR #10978, opened 2024-02-07
An attempt to get legacy perfect for mania closer to what it should be

nanaya:
Calling this a mess is a bit of understatement. Also it's still kind of wrong.

```csharp
            if ((CurrentScore.CountMiss + CurrentScore.Count50 + CurrentScore.Count100) == 0 && ((float)CurrentScore.CountKatu / CurrentScore.TotalHits) < 0.1)
                CurrentScore.Perfect = true;
            else
                CurrentScore.Perfect = false;
```

Related: #10679

### DIFF
--- a/app/Models/Solo/Score.php
+++ b/app/Models/Solo/Score.php
@@ -274,24 +274,31 @@ class Score extends Model implements Traits\ReportableInterface
      */
     public function isPerfectLegacyCombo(): ?bool
     {
+        // This is best effort as there's no way to re-generate the correct
+        // value short of checking the source legacy score.
         if ($this->ruleset_id === Ruleset::mania->value) {
             if (!$this->passed) {
                 return false;
             }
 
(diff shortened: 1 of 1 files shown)


app/Models/Solo/Score.php
```diff
@@ -274,26 +274,38 @@ public function isPerfectCombo(): bool
...
                     return false;
                 }
             }
 
+            $hits = $statistics->good + $statistics->great + $statistics->perfect;
+            if ($hits !== $this->data->maximumStatistics->perfect) {
+                return false;
+            }
+
+            $totalHits = $hits + $statistics->meh + $statistics->ok + $statistics->miss;
```
notbakaneko, line 304: but if any of these are non-zero, if should be false already, so how does it work :psy:

nanaya: good point 🤔

notbakaneko (approved):

PR #10984, opened 2024-02-08
Update number of nominations required to qualify Beatmap

notbakaneko:
- Updated required nominations
- Moved the nomination logic out to its own class
- Removed the `try-catch` dance on `nominate()`
- went through the nomination and qualification tests to make sure they were passing/failing on the things they were supposed to be
 - the `expectExceptionCallable` wrapper is so that we can run additional assertions after the test throws to make sure it didn't do the thing it wasn't supposed to do.
 - `withHypes()` test factory isn't used by any test, it's for quickly seeding a `Beatmapset` with necessary number of hypes for nomination.
- throws an error if nominating puts the mapset in an invalid state that can't be qualified (e.g. main ruleset changing and making the number of nominations invalid)
- how long has this been using the wrong translation key 👀 
- [x] we're just going to deal with it when it happens/reset the nominations.  ~~might be a problem with existing nominations since they might fail the nomination count check~~ 🤔 
  - ~~there's appears to be only 1 beatmapset with a limited nomination that might fail the check so maybe it's better to handle it manually when it arises instead of coding in an exception for it.~~
- [x] Main ruleset probably needs to be stored somewhere, constantly re-querying everything to get the summary counts for `BeatmapPanel` is probably not a good idea...
- remove ability to do legacy nominations

I'm still not quite satisfied with the number of repeated things the tests do each time, but we've had questions in the past about notifications, etc not being sent due to a user setting, when they have been dispatched.

- [x] updating logic if 1 and 2 not met
Probationary BNs cannot be the only nomination on their ruleset (i.e. can only nominate the main ruleset)

---

The new rules for nomination on hybrid sets are:

2 nominations required for the main ruleset and 1 on every other ruleset.
There must be one full nominator on every ruleset.

1. The main ruleset is the one with the most difficulties
2. If the number of difficulties are the same, then the one with the most diffs by the host if the main ruleset.
If 1 and 2 do not result in a clear main ruleset, then any ruleset that satisfies the previous criteria is used.


notbakaneko:
(draft while I fix some missing stuff + rebase)

database/migrations/2024_04_04_000001_add_eligible_main_rulesets_to_beatmapsets.php
```diff
@@ -0,0 +1,27 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('osu_beatmapsets', function (Blueprint $table) {
+            $table->json('eligible_main_rulesets')->nullable();
```
notbakaneko, line 17: I feel like this should be mode instead of ruleset and all the ruleset stuff in beatmap discussions/nominations reverted since all the existing ones are already stored using the legacy name, and the switch to `Ruleset` makes all the logic more confusing.

nanaya: more like should store ruleset id instead, I think

notbakaneko: switched to storing the ids instead, not sure if better but at least they're in a sorted now 👀

app/Libraries/NominateBeatmapset.php
```diff
@@ -0,0 +1,190 @@
...
+class NominateBeatmapset
+{
+    /** @var Set<int> */
+    private Set $beatmapRulesetIds;
+    /** @var Set<int|null> */
+    private Set $nominatedRulesetIds;
+
+    public function __construct(private Beatmapset $beatmapset, private User $user, array $playmodes)
+    {
+        $this->beatmapRulesetIds = new Set($beatmapset->playmodes()->toArray());
```
nanaya, line 28: this returns array of string for cached value from `difficulty_names`

app/Libraries/NominateBeatmapset.php
```diff
@@ -0,0 +1,190 @@
...
+            if ($count > 1) {
+                if ($maybeHasMainRuleset || !$eligibleRulesetIds->contains(Beatmap::modeInt($name))) {
+                    throw new InvariantException(osu_trans('beatmapsets.nominate.too_many_non_main_ruleset'));
+                }
+
+                $maybeHasMainRuleset = true;
+            }
+        }
+
+        foreach ($rulesetIds as $rulesetId) {
```
nanaya, line 134: if there's no existing nominations, a limited bng can end up nominating multiple eligible main rulesets

nanaya (changes requested):
part 1/?

app/Libraries/BeatmapsetMainRuleset.php
```diff
@@ -0,0 +1,129 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+declare(strict_types=1);
+
+namespace App\Libraries;
```
nanaya, line 8: oh yeah, this and the other beatmapset related library classes should probably go in `App\Libraries\Beatmapset`?

nanaya (changes requested):
part 2/?

app/Libraries/NominateBeatmapset.php
```diff
@@ -0,0 +1,190 @@
...
+{
+    /** @var Set<int> */
+    private Set $beatmapRulesetIds;
+    /** @var Set<int|null> */
+    private Set $nominatedRulesetIds;
+
+    public function __construct(private Beatmapset $beatmapset, private User $user, array $playmodes)
+    {
+        $this->beatmapRulesetIds = new Set($beatmapset->playmodes()->toArray());
+        $this->nominatedRulesetIds = new Set(array_map(fn (string $name) => Beatmap::MODES[$name] ?? null, $playmodes));
```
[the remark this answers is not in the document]
notbakaneko: I was planning on everything being ids eventually but probably better to do it all together instead of part of it now 😬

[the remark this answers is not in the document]
nanaya: well maybe at least convert the `nominationCount` one to id as well instead of staying as name...

app/Models/Beatmapset.php
```diff
@@ -1314,6 +1221,45 @@ public function hasFullBNNomination($mode = null)
...
+            // legacy nomination, only check group
+            if ($nomination->modes === null) {
+                if ($nomination->user->isLimitedBN()) {
+                    $result['limited'][] = null;
+                } else if ($nomination->user->isBNG() || $nomination->user->isNAT()) {
+                    $result['full'][] = null;
+                }
+            } else {
+                foreach ($nomination->modes as $mode) {
+                    $nominationType = $userNominationModes[$mode] ?? null;
```
nanaya, line 1252: this would potentially invalidate the nominations displayed on web, wouldn't it?

app/Models/Beatmapset.php
```diff
@@ -1265,6 +1140,38 @@ static function ($ele) {
         );
     }
 
+    /**
+     * Returns all the Rulesets that are eligible to be the main ruleset.
+     * This will _not_ query the current beatmapset nominations if there is an existing value in `eligible_main_ruleset_ids`
+     *
+     * @return int[]
+     */
+    public function eligibleMainRulesetIds(): array
```
nanaya, line 1149: should also be reset on disqualification/nomination reset

app/Libraries/Beatmapset/NominateBeatmapset.php
```diff
@@ -0,0 +1,213 @@
...
+
+        // check if there are any outstanding issues still
+        if ($this->beatmapset->beatmapDiscussions()->openIssues()->count() > 0) {
+            throw new InvariantException(osu_trans('beatmaps.nominations.unresolved_issues'));
+        }
+    }
+
+    private function assertRulesetNomination()
+    {
+        // NOTE: This assumption is only valid for the current nomination rules.
```
nanaya, line 119: similar note should probably also in the `requiredNominationsConfig`?

Because that's where it looks like the number is configurable but it's actually hardcoded/assumed in quite a few places (at least here and nominator's `userCanNominateMode` from a quick glance).

resources/js/beatmap-discussions/nominator.tsx
```diff
@@ -301,14 +323,27 @@ export class Nominator extends React.Component<Props> {
...
   };
 
   private userCanNominateMode(mode: GameMode) {
-    if (!this.userHasNominatePermission || this.nominationCountMet(mode)) {
+    // Always enable the selected one so it can be unselected.
+    if (this.selectedModes.includes(mode)) return true;
+
+    if (!this.userHasNominatePermission
+      || this.nominationCountMet(mode)
+      || (core.currentUserOrFail.is_limited_bn && this.selectedModes.length > 0)) {
```
nanaya, line 346: I think this is missing check for whether or not it's main ruleset

nanaya (changes requested):

app/Libraries/Beatmapset/NominateBeatmapset.php
notbakaneko: the assumption should be gone now 🤔

tests/Models/BeatmapsetTest.php
```diff
@@ -76,110 +75,216 @@ public function testLoveBeatmapApprovedStates(): void
...
+        }
 
-        Bus::assertDispatched(CheckBeatmapsetCovers::class);
+        // Assertions that nomination doesn't qualify
+        $this->assertTrue($beatmapset->isPending());
+        Bus::assertNotDispatched(CheckBeatmapsetCovers::class);
     }
 
-    public function testLimitedBNGQualifyingNominationNATNominated()
+    public function testNominateMainRulsetInvariant()
```
nanaya, line 246: rul...set :thonk:

app/Libraries/Beatmapset/NominateBeatmapset.php
notbakaneko: and the assumption is back 😬

tests/Models/BeatmapsetTest.php
```diff
@@ -255,243 +389,232 @@ public function testGlobalScopeSoftDelete()
...
-        $this->fillNominationsExceptLastForMode($beatmapset, 'bng_limited', 'taiko');
+        $beatmapset->refresh();
 
-        $result = $beatmapset->fresh()->nominate($user, ['osu', 'taiko']);
+        $this->expectCountChange(fn () => $beatmapset->bssProcessQueues()->count(), 1);
 
-        $this->assertFalse($result['result']);
-        $this->assertSame($result['message'], osu_trans('beatmapsets.nominate.full_bn_required'));
-        $this->assertTrue($beatmapset->fresh()->isPending());
+        $beatmapset->nominate($nominator, ['taiko']);
```
nanaya, line 581: this test fails if it's `$beatmapset->fresh()->nominate(...)` instead. Replacing the refresh with `$beatmapset = $beatmapset->fresh()` doesn't help so the `bssProcessQueues()` maybe somehow do something to the nominate function? 🤷 

Also the nomination throws this error (not sure if it's tested somewhere in this file):



- map with taiko and fruits being eligible
- existing nominations (in order):
  1. fruits (bn)
  2. fruits (limited bn)
- trying to nominate (qualify) taiko (bn)

(I copied this test and noticed it doesn't fail unless I call fresh beforehand just like in this test)

nanaya (changes requested):

tests/Models/BeatmapsetTest.php
nanaya: oh it was just the model used for assertion was different 🤔 still no idea why I'm getting that nomination error above though

notbakaneko: removing the `bssProcessQueues()` check doesn't have any affect 🤔

app/Libraries/Beatmapset/BeatmapsetMainRuleset.php
```diff
@@ -0,0 +1,136 @@
...
+
+    private function mainRuleset(): ?string
+    {
+        if ($this->eligibleRulesets->count() === 1) {
+            return $this->eligibleRulesets->first();
+        }
+
+        // First mode with more nominations that others becomes main mode.
+        // Implicity implies that limited BN nominations becomes main mode.
+        $nominations = $this->beatmapset->beatmapsetNominations()->current()->get();
```
nanaya, line 68: looks like this needs an explicit order. In my dev db the new row somehow is returned first and messed things up

nanaya (changes requested):

app/Libraries/Beatmapset/BeatmapsetMainRuleset.php
nanaya: either that or the logic needs some rework to not immediately finish on the first ruleset with single nomination...

nanaya (approved):
seems working <sub>(inb4 things fall over right after deploy)</sub>

nanaya (approved):

PR #10985, opened 2024-02-09
Fix disqualify button not focusing on new discussion text box

notbakaneko:
`window.setTimeout` is for if the page has to switch from the History tab, otherwise the focus won't trigger.


resources/js/beatmap-discussions/nominations.tsx
```diff
@@ -243,8 +243,11 @@ export class Nominations extends React.Component<Props> {
     // and thus new discussion box isn't visible.
     if (this.props.discussionsState.currentPage === 'events') {
       this.props.discussionsState.changeDiscussionPage('generalAll');
-      this.focusNewDiscussion();
     }
+
```
nanaya, line 247: the button still doesn't do anything if on review page

nanaya (changes requested):

nanaya (approved):

PR #10988, opened 2024-02-10
Hide global rank display if a score is non-ranked or non-preserved

### LINKED ISSUE #10966: "Global rank" should not be shown on score pages for non-preserved and non-ranked scores
Example 1: https://osu.ppy.sh/scores/2282646880

image

Score I set on a graved map. Shows "number 1 global rank", yet score is not ranked.

Example 2: https://osu.ppy.sh/scores/2272890904

image

Annotated as non-preserved yet still shows global rank.

tsunyoku:
Resolves #10966

### DIFF
--- a/resources/js/scores-show/player.tsx
+++ b/resources/js/scores-show/player.tsx
@@ -17,13 +17,10 @@ export default function Player(props: Props) {
   let title: string;
   let content: React.ReactNode;
 
-  if (props.score.ranked === false || props.score.preserve === false)
-  {
+  if (props.score.ranked === false || props.score.preserve === false) {
     title = trans('scores.show.player.no_rank');
     content = '-';
-  }
-  else
-  {
+  } else {
     title = '';
     content = <>#{formatNumber(props.score.rank_global)}</>;
   }
--- a/resources/js/scores-show/player.tsx
+++ b/resources/js/scores-show/player.tsx
@@ -17,7 +17,7 @@ export default function Player(props: Props) {
   let title: string;
   let content: React.ReactNode;
 
-  if (props.score.ranked === false || props.score.preserve === false) {
+  if (props.score.rank_global == null || props.score.ranked === false || props.score.preserve === false) {
     title = trans('scores.show.player.no_rank');
     content = '-';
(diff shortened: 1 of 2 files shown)


cl8n (approved):
change looks good

...but reviewing design of this page again, i dont think it's great to only show `-` for a case where something isn't included, it kind of prompts the question of "why?" if the user isn't already familiar with how this stuff works

peppy:
We already have tooltips on the "recent plays" to describe the "why" of this, so it could likely also be done here.

resources/js/scores-show/player.tsx
```diff
@@ -50,7 +50,7 @@ export default function Player(props: Props) {
           {trans('scores.show.player.rank.global')}
         </div>
         <div className='score-player__rank score-player__rank--value'>
-          {props.score.rank_global == null ? '-' : `#${formatNumber(props.score.rank_global)}`}
+          {props.score.rank_global == null || props.score.ranked === false || props.score.preserve === false ? '-' : `#${formatNumber(props.score.rank_global)}`}
```
notbakaneko, line 53: Can probably add a tooltip for non-ranked/non-preserved scores like that in `pp-value.tsx` (I think a single generic message should be fine?)
Also, split the line.

notbakaneko (changes requested):

resources/js/scores-show/player.tsx
```diff
@@ -14,6 +14,17 @@ interface Props {
 }
 
 export default function Player(props: Props) {
+  let title: string;
+  let content: React.ReactNode;
+
+  if (props.score.rank_global == null || props.score.ranked === false || props.score.preserve === false) {
+    title = trans('scores.show.player.no_rank');
```
notbakaneko, line 21: `scores.status.no_rank`

notbakaneko (changes requested):

resources/js/scores-show/player.tsx
tsunyoku: oops

notbakaneko (approved):

PR #10996, opened 2024-02-14
Update to php 8.3

nanaya:
Seems working now.

- [x] wait for php-ds 1.5.0 arm64


notbakaneko (changes requested):
Their repo is missing the `arm64` versions of `php-ds-1.5.0` ಠ_ಠ

nanaya:
the package has been added

notbakaneko (approved):

PR #10999, opened 2024-02-15
Use docker builtin init

nanaya:
This speeds up shutdown time since artisan queue doesn't quite handle signal correctly. Similarly with schedule loop.


notbakaneko (changes requested):
I think we should remove the `version:` line while we're at it since it's ignored anyway. (`init` is a `3.7` feature)

notbakaneko (approved):

PR #11002, opened 2024-02-15
Update docs relating to score

nanaya:
Something like this? 🤷 

(I'm not touching the score object docs just yet)

### DIFF
--- a/app/Http/Controllers/BeatmapPacksController.php
+++ b/app/Http/Controllers/BeatmapPacksController.php
@@ -87,7 +87,7 @@ class BeatmapPacksController extends Controller
      *
      * @urlParam pack string required The tag of the beatmap pack to be returned.
      *
-     * @queryParam legacy_only integer Whether or not to consider lazer scores for user completion data. Defaults to 1. Example: 0
+     * @queryParam legacy_only integer Whether or not to consider lazer scores for user completion data. Defaults to 0. Example: 0
      */
     public function show($idOrTag)
     {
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/BeatmapPacksController.php
```diff
@@ -86,6 +86,8 @@ public function index()
      * user_completion_data |
      *
      * @urlParam pack string required The tag of the beatmap pack to be returned.
+     *
+     * @queryParam legacy_only integer Whether or not to consider lazer scores for user completion data. Defaults to 1. Example: 0
```
respektive, line 90: Doesn't it default to 0? I'm at least getting lazer scores without setting this param in the request and have to manually set it to 1 to exclude them.

notbakaneko: I think this should say it's also based on the user preference setting?

notbakaneko (changes requested):

app/Http/Controllers/BeatmapPacksController.php
nanaya: user preference is not used for api request

resources/views/docs/info.md.blade.php
```diff
@@ -124,6 +124,16 @@
 v2      | current
 v1      | _legacy api provided by the old site, will be deprecated soon_
 
+### API Response Version
+
+Sometimes API response needs updating in non-backward compatible ways. In such cases, `x-api-version` header is used to determine which version of the response will be returned.
```
notbakaneko, line 129: forgot about this line 👀 
```suggestion
Sometimes, an API response need to be updated in non-backward compatible ways. In such cases, the `x-api-version` header is used to determine which version of the response will be returned.
```

notbakaneko (changes requested):

notbakaneko (approved):

Issue #10922, opened 2024-01-30
Wrong beatmap leaderboard ranking after recent update


kafibu:
This [player](https://osu.ppy.sh/users/15434024/osu) lost his no1 ranking on this [beatmap](https://osu.ppy.sh/beatmapsets/2060600#osu/4324907) after latest update (It the same whether lazer mode is on or not)

![2024-01-30 (1)](https://github.com/ppy/osu-web/assets/111474179/d9e0c5ec-e4e2-47c7-9d4f-bd8936766cdc)

Note his no1 ranking still appear in his profile

![2024-01-30 (2)](https://github.com/ppy/osu-web/assets/111474179/0ac43267-6fcd-4a5f-8e6c-56cb41295e93)

nanaya:
The id order probably doesn't match when the score is imported. Not sure how to get that fixed...

One would be to include legacy_score_id in the index and sort by that but that wouldn't work for lazer mode. Another alternative would be to sort by one of the timestamp columns (started_at etc) before by id.

@peppy

peppy:
Order was matched on import, I made special care to ensure that. Will investigate.

nanaya:
It's the same on non-lazer.

| score id   | legacy score id |
|------------|-----------------|
| 1808465719 | 4561877924      |
| 1808465739 | 4561808159      |

peppy:
Other cases brought up in linked threads:

ppy/osu#27731:

| score id   | legacy score id |
|------------|-----------------|
| [1808465719](https://osu.ppy.sh/scores/1961650557) | [214575709](https://osu.ppy.sh/scores/fruits/214575709) |
| [1808465739](https://osu.ppy.sh/scores/1961650466) | [214593084](https://osu.ppy.sh/scores/fruits/214593084) |


#11048

| score id   | legacy score id |
|------------|-----------------|
| [1961647015](https://osu.ppy.sh/scores/1961647015) | [214615400](https://osu.ppy.sh/scores/fruits/214615400) |
| [1961647016](https://osu.ppy.sh/scores/1961647016) | [214584096](https://osu.ppy.sh/scores/fruits/214584096) |

peppy:
Repairs have been run for this across all beatmaps / rulesets!

Issue #10725, opened 2023-11-10
Multiplayer rooms are missing their scores


TTTaevas:
(note I'm only talking about **the type of rooms where players play against each other in real time**, not the type of room where players have 7 days to play specific maps)

Up until somewhat recently, on the page of a room, the beatmaps (the items) that were played in that room would be displayed
Now they simply do not display anymore, and the "Map Count" remains on 0
Example: https://osu.ppy.sh/multiplayer/rooms/413097

Similarly, on the side of the API, [the endpoint to get scores from a room's item](https://osu.ppy.sh/docs/index.html#get-scores) doesn't seem to work anymore, but it's important to note [the undocumented endpoint that is used to list a room's items](https://osu.ppy.sh/docs/index.html#get-apiv2roomsroom) **does work**, its `playlist` array has all the items you'd expect it to have
What I mean by "the endpoint to get scores from a room doesn't seem to work anymore" is that it'll always give this as a response: `{
  params: { limit: 50, sort: 'score_desc' },
  scores: [],
  total: 0,
  user_score: null,
  cursor: null,
  cursor_string: null
}`

So there must be some issue going on with playlist items

nanaya:
It was some migration failure before (somehow score links aren't created properly) and then the remaining scores got wiped/archived sometime in jan.

peppy:
I'll track progress in here.

So far I have removed all links to old `solo_scores`.

I'm currently working on re-importing these scores with `ranked=0` so they can be restored in the leaderboards.

peppy:
All recoverable scores have now been reimported!

TTTaevas:
Thank you very much!!

PR #11003, opened 2024-02-15
Record new score id when pinning

nanaya:
First part of moving pinned score to new table. The column name is a bit unfortunate 🫥

Index to be added later as it should be unique but the current system allows pinning both legacy and imported score.

- [x] legacy score re-import or whatever solution for the incorrect import order ppy/osu-web#10725
- [x] ~~depends on #10922 as the score id should not change after this is merged and the linked issue may end with changing score id~~

~~Draft while looking into doing the whole thing in one go instead.~~ Nope


notbakaneko (approved):

notbakaneko:
is this still waiting to the reimport task above? @peppy @nanaya

peppy:
Yeah sorry, I'm making slow progress (see https://github.com/ppy/osu-web/issues/10725 for the latest).

nanaya:
Oh yeah, the score id reordering thing apparently will be handled separately by @peppy so this can just go ahead.

notbakaneko:
Would help if I pressed the update button too 👀

PR #11005, opened 2024-02-15
Display changelog entries count on user profile link section

### LINKED ISSUE #10962: Show changelog contribution count & GitHub profile on user page
I got a few requests for this immediately after the github linking thing went live. there is a short discussion about it in the dev server: <https://discord.com/channels/188630481301012481/188630616286167041/1165766213830447185> (tl;dr i couldn't figure out a way to display it that I was happy with, but I was probably overthinking it)

tsunyoku:
Resolves #10962.

## GitHub user linked
image

## GitHub user not linked
image

### DIFF
--- a/app/Transformers/UserTransformer.php
+++ b/app/Transformers/UserTransformer.php
@@ -48,6 +48,8 @@ class UserTransformer extends UserCompactTransformer
             'title_url' => $user->titleUrl(),
             'twitter' => $user->user_twitter,
             'website' => $user->user_website,
+            'changelog_entries_count' => $user->githubUser?->changelogEntries->count(),
+            'github_url' => $user->githubUser?->githubUrl(),
         ]);
     }
 }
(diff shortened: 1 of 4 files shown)


nanaya:
I really would rather not. It looks like it may motivate people in the wrong way.

tsunyoku:
I understand where you're coming from, but I don't think it's going to motivate people in the wrong way any more than somebody may already be (changelogs display the user for each changelog entry, GitHub has contributors tab with similar statistics)

notbakaneko:
I don't see how this is meaningful? Might as well count the number of lines I write
and I put
everything on a new line
since that increases 
it
faster

cl8n:
I can understand the issue as well and I usually oppose similar things using nanaya's reasoning, but I don't view this one as much different from the neighboring forum post or comment counts. it's just an attempt to attribute somehting to the user, even though the value itself is measuring a very arbitrary thing. it would be cool if this could instead link to some page showing *what* the user contributed, like a filtered changelog or somethn, but I imagine that is not straightforward to implement like this currently is.

but if the number is really that big of a deal then I guess just link github alongside the other `mediaKeys` ..?

peppy:
i dunnnnnno

tsunyoku:
okay, understandable. how about clayton's suggestion and we just display the user's github as another social with no stats?

frenzibyte:
That just feels weird. GitHub doesn't really have any place in osu!.

PR #11006, opened 2024-02-15
:pencil2: Fix minor mistakes on Norwegian localization

HilmTien:
"sjekk tilbake senere" is wrong, it should be "kom tilbake senere" (come back later) or "sjekk igjen senere" (check again later)

"registret" is wrong, it should be "registrert"

"resultatene" is a more precise wording than "resultater" since the results are definite from the implication that the tournament has concluded

### DIFF
--- a/resources/lang/no/tournament.php
+++ b/resources/lang/no/tournament.php
@@ -5,7 +5,7 @@
 
 return [
     'index' => [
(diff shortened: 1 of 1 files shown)


sheeshDotExe (approved):
Looks good to me 👍

PR #11009, opened 2024-02-16
Remove unused locales from `LocaleMeta`

cl8n:
some parts of the app relying on LocaleMeta let you interact with these locales despite there being no normal way to select them and no actual translations. example: <https://osu.ppy.sh/wiki/zh-hk/Sitemap>


app/Libraries/LocaleMeta.php
```diff
@@ -9,6 +9,7 @@
 
 class LocaleMeta
 {
+    // Keep keys in sync with app.available_locales config
```
nanaya, line 12: no, this array is supposed to allow more languages for wiki. It does need to include all available locales but as superset, not in sync.

nanaya (changes requested):

app/Libraries/LocaleMeta.php
cl8n: oh so it is intentional :sweat_smile:  

on osu-wiki's end, only `availables_locales` are allowed right now, and both of the locales I removed here have no articles. I did some digging and found what happened to each of them...
- `tl` caused some confusion because the translators expected it to appear when you select Filipino on the website, but that's actually `fil`, so they moved the few `tl` articles over to `fil`
- `zh-hk` was deleted, seems like people just assumed it was unsupported because you can't select this language on website. it was only a few articles and all extremely outdated so I don't think that's worth reverting now

given the confusion in each case I'm not sure if more locales here *should* be supported :thinking:

nanaya: it's less of a problem now with most of the popular languages being supported but I don't think it needs to be strictly in sync.

Removing the unused locales is fine but the comment should be adjusted

nanaya (approved):

PR #11010, opened 2024-02-19
Limit score pinning to new score table

### LINKED ISSUE #10813: Stable scores don't show classic mod on lazer website in the pinned setion
image
image

nanaya:
Previous addition of solo_score support wasn't quite complete and messed up a few things...

- [x] Depends on #11003.
- [x] Update existing pinned scores
    ```sql
    update score_pins p
    set new_score_id = (select id from scores where ruleset_id = p.ruleset_id and legacy_score_id = p.score_id)
    where score_type like 'score_best_%' and new_score_id is null;
    ```

Resolves #10813


peppy:
> Update existing pinned scores

I guess this should be run post-deploy of this PR, yeah?

nanaya:
No, before. The previous PR does it for new pins and the query above fills up the older ones.

peppy:
Pinned scores have been migrated.

tests/Controllers/ScorePinsControllerTest.php
```diff
@@ -79,7 +79,7 @@ public function testReorderMoveBottom()
         $user = User::factory()->create();
         $pins = collect([0, 1])->map(fn ($order) => ScorePin
             ::factory(['display_order' => $order])
-            ->withScore(static::createScore($user, Ruleset::osu))
+            ->withScore(static::createScore($user, Beatmap::MODES['osu']))
```
notbakaneko, line 82: I'm not sure if this is any better (or worse 🤷 ) than using `Ruleset` in this case

notbakaneko (approved):

PR #11028, opened 2024-02-24
Fix legacy score rank calculation

### LINKED ISSUE #11025: ``/beatmaps/{beatmap}/scores/users/{user}`` returns wrong data
As shown here on a score submitted on lazer, ``scores`` field returns '0' and ``rank`` field is incorrect, as though it was calculated in classic mod.

Also, on screenshot below, ``rank`` returns wrong value. It should be an ``SH`` (just like it does on the beatmap page) because the score has exactly 1% 50s, as explained on the [wiki](https://osu.ppy.sh/wiki/en/Gameplay/Grade#osu!).

nanaya:
Missed an equal sign.

Resolves #11025.

### DIFF
--- a/app/Models/Traits/Scoreable.php
+++ b/app/Models/Traits/Scoreable.php
@@ -93,7 +93,7 @@ trait Scoreable
                 $this->rank = match (true) {
                     $ratio300 === 1 =>
                         $this->shouldHaveHiddenRank() ? 'XH' : 'X',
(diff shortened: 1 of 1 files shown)


peppy:
Confirming this now matches stable:

2024-02-24 17 12 55@2x

peppy (approved):

PR #11035, opened 2024-02-26
Add documentation for new score json

nanaya:
Something like this maybe 🤔 The version thing seems a bit weird.


YoruNoKen:
wasn't `rank` renamed to `grade`? I might be remembering wrong

notbakaneko (approved):

PR #11036, opened 2024-02-26
Update to nodejs 20

nanaya:
Original plan of updating to debian 12 went to bin due to swoole's curl breakage 🚮


notbakaneko (approved):
shouldn't break anything...I think 🤔

PR #11042, opened 2024-02-27
Add missing default on bancho stats table

nanaya:


database/migrations/2024_02_27_134104_add_default_to_banchostats_users_lazer.php
```diff
@@ -0,0 +1,27 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('osu_banchostats', function (Blueprint $table) {
```
peppy, line 16: The other columns don't seem to have defaults either, should we be adding them?

Also as mentioned on discord, may want to update the seeder:

```diff
diff --git a/database/factories/BanchoStatsFactory.php b/database/factories/BanchoStatsFactory.php
index fe90b74e41..c75a595e09 100644
--- a/database/factories/BanchoStatsFactory.php
+++ b/database/factories/BanchoStatsFactory.php
@@ -19,6 +19,7 @@ class BanchoStatsFactory extends Factory
         return [
             'users_irc' => fn() => 100 + $this->faker->randomNumber(2),
             'users_osu' => fn() => 10000 + $this->faker->randomNumber(4),
+            'users_lazer' => fn() => 1000 + $this->faker->randomNumber(6),
             'multiplayer_games' => fn() => 200 + $this->faker->randomNumber(3),
             'date' => fn() => Carbon::now(),
         ];

```

tsunyoku: I'm not convinced that we need defaults on any of the columns... Bancho should always be inserting numbers to these (even if 0, which'll never happen) right? Adding a default to `users_lazer` (or any of the columns) is only a reactive fix to the issue happening during migrations - modifying the factory seems like the correct solution

nanaya: added all defaults and updated factory.

having sane default is occasionally useful.

notbakaneko (approved):

PR #11045, opened 2024-02-28
Remove trailing space of reference link tooltip

nanaya:
There are nbsp's between the backlinks.

### DIFF
--- a/resources/js/core/wiki/reference-link-tooltip.ts
+++ b/resources/js/core/wiki/reference-link-tooltip.ts
@@ -44,12 +44,9 @@ export default class ReferenceLinkTooltip {
 
     if (targetId == null) return;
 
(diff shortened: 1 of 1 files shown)


notbakaneko (commented):
There's still an extra `&nbsp` at end of the text node...

the space prepended to backrefs also seems to be hardcoded into `FootnoteExtension`... 🤔 
`return '&nbsp;' . new HtmlElement('a', $attrs->export(), \htmlspecialchars($symbol), true);`

nanaya:
The extra nbsp is because the text is part of the rest of the tooltip and I'm not going to slice and dice text node especially as they're really hard to work on (if possible at all - I couldn't even figure out how to get the object to return `&nbsp;`). One extra space doesn't really matter.

notbakaneko (approved):

PR #11047, opened 2024-02-28
Specify defaults for medal unlock notification options

nanaya:
Explicitly set so it doesn't depend on the base class providing the expected defaults.


notbakaneko:
makes sense I guess 🤔

PR #11049, opened 2024-02-29
Update ranked flag to remove scores from index

nanaya:
Prevents scores from being actually deleted, for things like multiplayer scores which scores should stay.


notbakaneko (approved):
does this need to wait for anything to be deployed? @peppy

peppy (approved):

peppy:
> does this need to wait for anything to be deployed? @peppy

already deployed the other components 😄

PR #11051, opened 2024-02-29
Don't escape supporter gift message mail

### LINKED ISSUE #11031: Some characters do not show up when you write a message for supporter
I recently gifted someone supporter and they showed me the message I sent them however it showed up with a bunch of weird characters.

image

Apparently, some characters aren't supported at all when you write messages to others. For this example, I wrote ' in the message, but it came out saying &039; instead. If this is happening for this character in specific, it might be happening to others as well.

nanaya:
Resolves #11031 (I think). I don't think there's any potential harm here...? 🤔


notbakaneko (approved):
well, the mail _is_ supposed to be plain text... 🤔

PR #11060, opened 2024-03-04
Include fade in for hidden rank check

### LINKED ISSUE #11059: Mania FI S/SS scores are not shown as silver unless lazer mode is enabled
without lazer mode:
image

with lazer mode: 
image

without lazer mode:
image

with lazer mode:
image

example score: https://osu.ppy.sh/scores/2402484733

nanaya:
Resolves #11059.


notbakaneko (approved):
I guess the list isn't too long... yet 🤔

PR #11062, opened 2024-03-04
Fix inconsistency in displayed miss count between stable & lazer catch scores

bdach:
Related: https://github.com/ppy/osu/issues/27473

In stable, both fruit misses and large droplet misses were counted as full misses and shown in the miss column. In lazer they are two disparate statistics (misses and large tick misses).

Because we cannot retroactively get this data out for all scores on stable, this PR changes the display instead so that lazer scores have misses and large tick misses summed up in the "miss" column, which restores parity (stable scores will be full misses which actually include large tick misses, plus zero "actual" large tick misses, and lazer scores will just sum actual full misses and actual large tick misses).

In the process this fixes silent CSS breakage due to implicit dependency of the CSS rule changed on the textual value of `attribute` key of each label.

Naming suggestions appreciated, I kinda didn't know what to do there.

### DIFF
--- a/resources/js/utils/score-helper.ts
+++ b/resources/js/utils/score-helper.ts
@@ -48,40 +48,67 @@ export function isPerfectCombo(score: SoloScoreJson) {
     : score.is_perfect_combo;
 }
 
-interface AttributeData {
-  attribute: SoloScoreStatisticsAttribute;
+interface AttributeDisplayMapping {
+  key: string,
+  attributes: SoloScoreStatisticsAttribute[];
   label: string;
+}
+
+interface AttributeDisplay {
+  key: string,
+  label: string;
+}
+
+interface AttributeDisplayTotal extends AttributeDisplay {
+  total: number;
 }
 
 const labelMiss = trans('beatmapsets.show.scoreboard.headers.miss');
 
-export const modeAttributesMap: Record<GameMode, AttributeData[]> = {
+const modeAttributesMap: Record<GameMode, AttributeDisplayMapping[]> = {
   fruits: [
-    { attribute: 'great', label: 'fruits' },
-    { attribute: 'large_tick_hit', label: 'ticks' },
-    { attribute: 'small_tick_miss', label: 'drp miss' },
-    { attribute: 'miss', label: labelMiss },
+    { key: 'great', attributes: ['great'], label: 'fruits' },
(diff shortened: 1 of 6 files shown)


resources/js/utils/score-helper.ts
```diff
@@ -48,41 +48,68 @@ export function isPerfectCombo(score: SoloScoreJson) {
...
-    { attribute: 'great', label: 'great' },
-    { attribute: 'ok', label: 'good' },
-    { attribute: 'miss', label: labelMiss },
+    { attributes: ['great'], key: 'great', label: 'great' },
+    { attributes: ['ok'], key: 'ok', label: 'good' },
+    { attributes: ['miss'], key: 'miss', label: labelMiss },
   ],
 };
 
+export function attributeDisplayLabels(ruleset: GameMode): AttributeDisplay[] {
```
nanaya, line 97: this one is a bit pointless, isn't it? All it does is literally just removing the attributes field from the items' object. Might as well just use the original array directly.

resources/js/utils/score-helper.ts
```diff
@@ -48,41 +48,68 @@ export function isPerfectCombo(score: SoloScoreJson) {
...
+  return modeAttributesMap[ruleset].map((mapping) => ({
+    key: mapping.key,
+    label: mapping.label,
+  }));
+}
+
+export function attributeDisplayTotals(ruleset: GameMode, score: SoloScoreJson): AttributeDisplayTotal[] {
+  const modeMapping = modeAttributesMap[ruleset];
+  return modeMapping.map((mapping) => ({
```
nanaya, line 106: can be inlined like the other function
```suggestion
  return modeAttributesMap[ruleset].map((mapping) => ({
```

resources/js/utils/score-helper.ts
```diff
@@ -48,41 +48,68 @@ export function isPerfectCombo(score: SoloScoreJson) {
...
+    key: mapping.key,
+    label: mapping.label,
+    total: mapping.attributes.map((attribute) => score.statistics[attribute] ?? 0).reduce((sum, elem) => sum + elem, 0),
```
nanaya, line 109: skip map?
```suggestion
    total: mapping.attributes.reduce((sum, attribute) => sum + (score.statistics[attribute] ?? 0), 0),
```

nanaya (changes requested):

nanaya (approved):

PR #11065, opened 2024-03-05
Use correct deny flow on oauth authorisation

nanaya:
Use library oauth to deal with auth deny.

### DIFF
--- a/app/Http/Controllers/Passport/AuthorizationController.php
+++ b/app/Http/Controllers/Passport/AuthorizationController.php
@@ -8,7 +8,6 @@ namespace App\Http\Controllers\Passport;
 use Illuminate\Http\Request;
 use Laravel\Passport\ClientRepository;
 use Laravel\Passport\Exceptions\AuthenticationException;
-use Laravel\Passport\Exceptions\OAuthServerException;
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Passport/AuthorizationController.php
```diff
@@ -35,21 +37,20 @@ public function authorize(
...
-        $redirectUri = presence(trim(get_string($request['redirect_uri']) ?? ''));
-
-        abort_if($redirectUri === null, 400, osu_trans('model_validation.required', ['attribute' => 'redirect_uri']));
-
-        if (!auth()->check()) {
-            // Breaks when url contains hash ("#").
-            $separator = strpos($redirectUri, '?') === false ? '?' : '&';
-            $cancelUrl = "{$redirectUri}{$separator}error=access_denied";
+        try {
+            \Log::debug($request->get('prompt'));
```
notbakaneko, line 41: is this supposed to be there

nanaya:
Wait, our exception handler is supposed to handle oauth exception. Let me check why didn't do the correct render for me.

notbakaneko (approved):

PR #11073, opened 2024-03-07
Combine ranking url generation and order constant

nanaya:
- [x] depends on #10831


notbakaneko (changes requested):
some conflicts now 🤔

notbakaneko (approved):

PR #11075, opened 2024-03-07
Include more detail when GitHub returns unexpected response

nanaya:
Somehow github returned an object without content (twice) but it's not supposed to happen so having an idea what it actually returns may help.

### DIFF
--- a/app/Libraries/OsuWiki.php
+++ b/app/Libraries/OsuWiki.php
@@ -16,8 +16,12 @@ class OsuWiki
 {
     const CACHE_DURATION = 60;
 
-    public $path;
-    public $data;
+    public array $data;
+
+    public function __construct(public string $path)
+    {
+        $this->data = static::fetch($this->path);
+    }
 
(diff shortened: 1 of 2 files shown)


app/Libraries/OsuWiki.php
```diff
@@ -40,12 +45,19 @@ public static function getTree(?string $sha = null, bool $recursive = true): arr
         return GitHub::gitData()->trees()->show(static::user(), static::repository(), $sha ?? static::branch(), $recursive);
     }
 
-    public static function fetch($path)
+    public static function fetch($path): array
```
nanaya, line 48: I notice the fetch doesn't seem to be used outside so maybe it's fine as private? 🤔 (and type `$path` as string?)

notbakaneko: Might as well make it `private`, only case I see for it public public is being able to call it directly in console, but we can just the results from `->data` anyway

notbakaneko (approved):

PR #11081, opened 2024-03-11
Add user cover preset model

nanaya:
Just the models. Not linked to user just yet (and would otherwise depend on #10804). The management page ended up weirder than expected 👀 probably can use some more javascripts/react but this basic one seems to somehow work.

No idea what group should have access to this yet so admin only for now.

Tests to be added later. Maybe.

Oh and there's bug with simple form's input which I'm not fixing here.


routes/web.php
```diff
@@ -290,6 +290,9 @@
     Route::post('session', 'SessionsController@store')->name('login');
     Route::delete('session', 'SessionsController@destroy')->name('logout');
 
+    Route::post('user-cover-presets/batch-activate', 'UserCoverPresetsController@batchActivate')->name('user-cover-presets.batch-activate');
+    Route::resource('user-cover-presets', 'UserCoverPresetsController', ['only' => ['index', 'store', 'update']]);
+
```
notbakaneko, line 295: shouldn't these go under `/admin`?

notbakaneko (changes requested):

routes/web.php
nanaya: not if we're going to allow this to be accessible by non-admin

notbakaneko:
> Oh and there's bug with simple form's input which I'm not fixing here.
is it the input is resizable 👀

notbakaneko (approved):

routes/web.php
Hiviexd: would be very nice to have it under `/admin` now in the console page, highly doubt any other usergroup will get access to this.

(another side effect is me having to look back at this PR each time to remember the route lol)

Issue #10961, opened 2024-02-03
Forum topic view counts aren't consistent with phpbb view count


cl8n:
on phpbb, the view count of a topic counted visits to the page (I haven't checked the exact details but that should be close to correct). the osu website now tracks it as the number of unique users who have visited the page.

- "views" is a misleading descriptor for the modern behavior. it's more like "users visited"
- topics that were live on phpbb have much greater view counts due to how it was tracked

PR #11082, opened 2024-03-11
Don't increment replay counts if replay download is coming from an API request

### LINKED ISSUE #11019: Only increment "Replays watched" when downloading through lazer
I'm running a discord bot that downloads replays through the API to show stats like the UR and also render parts of the replay for analyzing and I've been getting a few reports of people being confused about their "Replays watched by Others" stat being increased by using the bot and thinking it might be a bug.

I'm thinking it might be better to only increment it when someone downloads the replay from lazer (e.g. actually watching it in-game). This would also help with potential abuse, as I've already seen people talking about "farming" this stat now.

tsunyoku:
Resolves #11019.

Took the direction peppy suggested as it makes sense to me, but I can change direction if something else is wanted.

### DIFF
--- a/app/Http/Controllers/ScoresController.php
+++ b/app/Http/Controllers/ScoresController.php
@@ -52,7 +52,7 @@ class ScoresController extends Controller
             abort(404);
         }
 
-        if (\Auth::user()->getKey() !== $score->user_id) {
+        if (\Auth::user()->getKey() !== $score->user_id && !is_api_request()) {
             $score->user->statistics($score->getMode(), true)->increment('replay_popularity');
 
             $month = CarbonImmutable::now();


nanaya:
that function also returns true from lazer (also this won't help with abuse)

tsunyoku:
> also returns true from lazer
right.. i forgot about that. do we have a pattern for detecting lazer request so we can allow it, or is this direction simply not going to work?

nanaya:
I think the only way at the moment is checking the token client's `password_client` attribute

respektive:
Correct me if I'm wrong, but wouldn't explicitly checking for the lazer oauth client_id work? Also I think it would still make sense to additionally only count the first download per user as nanaya suggested, to also avoid any abuse using the lazer api client.

cl8n:
I don't agree with counting only one download per user; that seems like it would create an issue similar to #10961 but with a much more important metric. I think these counts should increment whenever web10 would do the same... so just a download counter, unless I'm missing something

nanaya:
While not as strict as I'm proposing, web10 is actually doing something pretty close to it.

cl8n:
superseded by #11126

PR #11085, opened 2024-03-12
Limit image type when regenerating beatmap covers

nanaya:
scthumber doesn't support bmp but not sure if should just convert to jpeg instead 🤨


notbakaneko (approved):
eta someone complains about webp 👀

PR #11086, opened 2024-03-12
Exclude scores without pp when displaying pack item pass indicator

bdach:
This is a port of https://github.com/ppy/osu-queue-score-statistics/pull/234 to web. See that pull for reasoning, or tl;dr: we want to exclude scores with mods that don't give pp ("unranked mods") from passing challenge packs, so we check pp even though there are more reasons why a score might have no pp (i.e. beatmap status).

I'm not super sure this is entirely valid. For the purposes of pack medals that was an okay workaround to apply. As long as challenge packs are only used for medals and/or as long as no challenge packs contain unranked/loved maps, this should hold up in web too.

If it turns out that any of the above assumptions is incorrect in a way that makes this partially incorrect this can probably be dumpstered since things are going to look slightly wrong either with or without this diff.


app/Libraries/Search/ScoreSearchParams.php
```diff
@@ -24,6 +24,7 @@ class ScoreSearchParams extends SearchParams
     public bool $excludeConverts = false;
     public ?array $excludeMods = null;
     public ?bool $isLegacy = null;
+    public bool $excludeScoresWithoutPP = false;
```
notbakaneko, line 27: I would think the scores part would be redundant in this context...

nanaya: I think so but this line should also be renamed 🤔

And I think osu-web uses `Pp` everywhere... (when not lowercased)

nanaya (approved):

PR #11092, opened 2024-03-14
Remove usage of profile customization model for user cover

nanaya:
🧹 

<del>drafting since I forgot to add migration to remove the column</del> Will do that after this one is deployed instead.


notbakaneko (approved):

nanaya:
I forgot to add "finish the migration" task 👀 (currently at id 19M out of ~35M max)

PR #11095, opened 2024-03-14
Remove cover json from user profile customization model

nanaya:
- [ ] depends on #11092
- [ ] and the above being deployed

<sub>lol the datatype is still `text`</sub>


nanaya:
Why is this still open 🤔

notbakaneko (dismissed):
I keep forgetting to merge this one ಠ_ಠ

PR #11105, opened 2024-03-19
Ignore broken beatmap archive for cover regeneration

nanaya:
For some reason there are beatmaps with invalid zip file ಠ_ಠ


notbakaneko (changes requested):
I think there should be a change to the metrics being recorded in `RegenerateBeatmapsetCover` if it fails on invalid zip? 🤔

nanaya:
I think it's the same as any other ignored cases unless we want to track them all which then would be a different pr.

notbakaneko:
What's the reason does this get recorded as `ok` instead of `error`?

nanaya:
Because it'll be fixed once the file is fixed. The same as missing file and invalid format.

notbakaneko (approved):

PR #11107, opened 2024-03-21
Remove unnecessary function and variable

nanaya:
`get_stream_filename` doesn't need specific position cursor (and nor has it been reset for things processed through image processor).

Also removed bytesWritten as it's not used anywhere.


notbakaneko:
Isn't it more that `file_get_contents` in `storeCover` always starts at `0` without passing offset?

notbakaneko (approved):

nanaya:
Well, only the file name (not handler) is being passed around, so...

PR #11108, opened 2024-03-21
Remove unused variable in Users controller

nanaya:
🤨


notbakaneko (approved):
oh yeah, I keep seeing this and forgetting to remove it 🤔

Issue #10589, opened 2023-09-23
Add note to content created or edited via third-party OAuth app


cl8n:
just got this idea from github:
![](https://github.com/ppy/osu-web/assets/9954349/cc725eb1-a6fd-4bc0-bfef-f21bab338e07)

currently, forum posts and chat messages can be posted by third-party client on behalf of a user, but there is no indication that it came from a third party. I think it'd be good to add a small message or icon that says "posted with \<client name\>" similar to the above image.

notbakaneko:
This is adds a lot of noise imo, especially for chat.

cl8n:
yeah for chat it would need to be just a small/dim icon with tooltip or something. I haven't looked into potential design for this yet.

PR #11113, opened 2024-03-24
Add API endpoints for creating/editing beatmap discussion posts

### LINKED ISSUE #10199: Add create/edit/delete beatmap discussion post API endpoints
some cool uses for this would be:

- NAT bot automatically posting veto conclusions
- a bot that automatically posts detected unrankables in a given beatmap

minetoblend:
Closes #10199 

Adds the ability to create new beatmap discussions, reply to existing discussions, and updating discussion posts, via the API. The whole thing is based on the forum API for how OAuth scopes are set up (`forum.write` scope).

## Use cases
- Bots that automatically detect unrankables and create posts about it
- Ability to write mods in third party mapping tools & editors
- Would help pave the way for modding in the lazer editor

## List of all changes:
- Added API endpoints for `BeatmapDiscussionPostsController.store` and `BeatmapDiscussionPostsController.update`
- Added a `beatmap_discussion.write` OAuth scope
- Added API Documentation for `BeatmapDiscussionPostsController.store` and `BeatmapDiscussionPostsController`
- Made adjustments to expected endpoints in `RouteScopesTest`
- Added English & German translations for `beatmap_discussion.write` OAuth scope

### DIFF
--- a/app/Http/Controllers/BeatmapDiscussionPostsController.php
+++ b/app/Http/Controllers/BeatmapDiscussionPostsController.php
@@ -27,8 +27,7 @@ class BeatmapDiscussionPostsController extends Controller
         $this->middleware('require-scopes:public', ['only' => ['index']]);
         $this->middleware('require-scopes:beatmapset_discussion.write', ['only' => [
             'store',
-            'update',
-            'destroy'
+            'update'
         ]]);
 
         parent::__construct();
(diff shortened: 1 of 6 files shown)


peppy:
> Ability to write mods in third party mapping tools & editors

I'm against supporting this, so it should be limited for now to not allow that.

Maybe client-auth only for lazer to start with.

minetoblend:
In that case, would adding it to the the scopes that require client delegation (like `chat.write`) work?
That way it can still be used by bot accounts & lazer, while not being able to post as regular users.

cl8n:
> This is in line with other restrictions we currently have like third party apps should not be able to post on forums as a user.

there is no such restriction :see_no_evil: maybe you're thinking of chat messages, which do have some additional requirements

what problems do you see with adding more write scopes like this? if it's going to be a blocker for third party, we should at least be tracking what needs to be done somewhere. maybe #10589 is related

PR #11117, opened 2024-03-25
Convert TwitchPlayer to typescript

notbakaneko:
Just a direct(ish) conversion


resources/js/core/twitch-player.tsx
```diff
@@ -0,0 +1,77 @@
...
+    addEventListener(action: string, callback: () => void): void;
+  }
+
+  interface Window {
+    Twitch?: {
+      Embed: unknown; // don't care.
+      Player: TwitchPlayerConstructor;
+    };
+  }
+}
```
nanaya, line 23: use class declaration instead?

```typescript
declare class TwitchEmbedPlayer {
  static PLAY: string;

  constructor (id: string, options: Record<string, unknown>);
  addEventListener(action: string, callback: () => void): void;
}

declare global {
  interface Window {
    Twitch?: {
      Embed: unknown; // don't care.
      Player: typeof TwitchEmbedPlayer;
    };
  }
}
```

nanaya (changes requested):

resources/js/core/twitch-player.tsx
```diff
@@ -0,0 +1,77 @@
...
+    PLAY: string;
+  }
+
+  interface TwitchPlayerClass {
+    addEventListener(action: string, callback: () => void): void;
+  }
+
+  interface Window {
+    Twitch?: {
+      Embed: unknown; // don't care.
```
nanaya, line 19: it's more like "not used" rather than "don't care" isn't it?

resources/js/core/twitch-player.tsx
notbakaneko: declaring it as class sets off eslint `max-classes-per-file` unless ignored and moving the declaration to a separate `.d.ts` adds it to the global namespace instead of limiting it to the file, so there's this roundabout workaround :|

nanaya: Do it like `shopify-buy.d.ts` then...?

nanaya (approved):

PR #11118, opened 2024-03-25
Always preserve multiplayer scores regardless of pass/fail

bdach:
Related: https://github.com/ppy/osu-queue-score-statistics/issues/141, https://github.com/ppy/osu-web/pull/10946

We want to preserve failed scores if they come from multiplayer so that the results of the multiplayer room can be accessed in the indeterminate future. Currently this would not happen if a multiplayer score was a fail.

### DIFF
--- a/app/Models/Multiplayer/ScoreLink.php
+++ b/app/Models/Multiplayer/ScoreLink.php
@@ -29,6 +29,10 @@ class ScoreLink extends Model
     {
         return \DB::transaction(function () use ($params, $token) {
             $score = Score::createFromJsonOrExplode($params);
+            
+            // multiplayer scores are always preserved.
+            $score->preserve = true;
+            $score->saveOrExplode();
 
             $playlistItem = $token->playlistItem;
             $requiredMods = array_column($playlistItem->required_mods, 'acronym');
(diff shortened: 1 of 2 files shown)


app/Models/Multiplayer/ScoreLink.php
```diff
@@ -29,6 +29,10 @@ public static function complete(ScoreToken $token, array $params): static
     {
         return \DB::transaction(function () use ($params, $token) {
             $score = Score::createFromJsonOrExplode($params);
+            
+            // multiplayer scores are always preserved.
+            $score->preserve = true;
+            $score->saveOrExplode();
```
bdach, line 35: This probably looks funny/awkward but there was an intention behind doing things that way. In most cases, scores are created by passing raw json params to `Score::extractParams()` and then passing that to `Score::createFromJsonOrExplode()`. But I didn't want to add a parameter to control `preserve` to `extractParams()` (which I would have had to do to avoid this silliness), because I don't want to allow an arbitrary api consumer to set `preserve` directly in the first place?

If I'm overthinking this do let me know and I'll shuffle across, but yeah that was the intention.

nanaya: seems fine but I think it can also be done with `createFromJsonOrExplode([...$params, 'preserve' => true])`

bdach: Yeah that works, good call. Applied, thanks :+1:

nanaya (approved):

PR #11120, opened 2024-03-26
Include original exception on image processor service errors

nanaya:
Would be useful 😐


nanaya:
Nevermind, the original exception isn't useful either.

PR #11123, opened 2024-03-26
Show model not found message

nanaya:
The model naming and the way it's done don't seem quite right but 🤷

Also I hope crowdin doesn't choke on the key name...

Related: ppy/osu#27720


app/Exceptions/Handler.php
```diff
@@ -90,6 +90,29 @@ private static function isOAuthSessionException(Throwable $e): bool
...
+            'model' => trans_exists($modelTransKey, $GLOBALS['cfg']['app']['fallback_locale'])
+                ? osu_trans($modelTransKey)
+                : trim(strtr($model, ['App\Models\\' => '']), '\\'),
+        ];
+
+        if (empty($ids)) {
+            $key = 'models.not_found_no_ids';
+        } else {
+            $key = 'models.not_found';
+            $params['ids'] = implode(', ', $ids);
```
nanaya, line 110: I wonder if the id shouldn't be visible. I took it from the original laravel exception message

notbakaneko: I'm thinking it shouldn't be shown in general, but in the context of the other issue, it'd make sense to know which one is missing? 🤔

notbakaneko (changes requested):
Yeah, I'm not sure if this should become the default behaviour...

image

app/Exceptions/Handler.php
```diff
@@ -90,6 +90,22 @@ private static function isOAuthSessionException(Throwable $e): bool
             && $e->getMessage() === 'Authorization request was not present in the session.';
     }
 
+    private static function modelNotFoundMessage(ModelNotFoundException $e): string
+    {
+        $ids = $e->getIds();
```
notbakaneko, line 95: now it's unused 👀

notbakaneko (changes requested):

notbakaneko (approved):

PR #11126, opened 2024-03-27
Limit replay file downloads count

nanaya:
Like #11082 but only for client downloads and counted once per day.

~~The test is a bit of a mess atm.~~ A bit better now.


notbakaneko (approved):

respektive:
Can probably close #11019 and #11082 now

PR #11129, opened 2024-03-29
Convert StoreSupporterTag to typescript

notbakaneko:
And react + mobx it


resources/js/store/store-supporter-tag.tsx
```diff
@@ -0,0 +1,316 @@
...
+      this.user = null;
+      this.debouncedGetUser(this.username);
+    } else {
+      this.user = core.currentUserOrFail;
+    }
+
+    disposeOnUnmount(
+      this,
+      autorun(() => {
+        toggleCart(this.isValidUser);
```
nanaya, line 97: this is triggered too early, causing wrong button state if `target` is specified and the username check takes too long or the user doesn't exist

(irrelevant if not doing any workarounds)

resources/js/store/store-supporter-tag.tsx
```diff
@@ -0,0 +1,316 @@
...
+
+  componentWillUnmount() {
+    document.removeEventListener('turbolinks:before-cache', this.handleBeforeCache);
+    this.xhr?.abort();
+  }
+
+  render() {
+    return (
+      <>
+        <div className='js-store store-supporter-tag'>
```
nanaya, line 179: js-store not needed anymore

resources/js/store/store-supporter-tag.tsx
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('turbolinks:before-cache', this.handleBeforeCache);
+    this.xhr?.abort();
+  }
+
+  render() {
+    return (
+      <>
```
nanaya, line 178: not needed

resources/js/register-components.tsx
```diff
@@ -106,17 +107,24 @@ core.reactTurbolinks.register('ranking-variant-filter', () => (
   <RankingVariantFilter {...parseJson('json-variant-filter')} />
 ));
 
+core.reactTurbolinks.register('store-supporter-tag', (container) => {
+  const productId = getInt(container.dataset.productId);
+  if (productId == null) {
+    throw new Error('missing productId');
+  }
+
+  const maxMessageLength = getInt(container.dataset.maxMessageLength) ?? 0;
```
nanaya, line 116: weird missing attribute handling with one throwing and one having nonsense default

(and similar to quantity, having product id inside the component just adds complexity compared to just having it directly on plain html)

resources/js/store/store-supporter-tag.tsx
```diff
@@ -0,0 +1,316 @@
...
+}
+
+function sliderValueFrom(value: number) {
+  return value * resolution;
+}
+
+@observer
+export default class StoreSupporterTag extends React.Component<Props> {
+  private readonly debouncedGetUser;
+  private readonly giftMessage: string = '';
```
nanaya, line 64: the attribute name is rather misleading as it's not the actual state...

resources/js/store/store-supporter-tag.tsx
```diff
@@ -0,0 +1,316 @@
...
+    $(slider).find('.ui-slider-range').remove();
+
+    return $(slider).slider({
+      animate: true,
+      change: this.handleSliderValueChanged,
+      max: sliderValueFrom(maxValue),
+      min: sliderValueFrom(minValue),
+      range: 'min',
+      slide: this.handleSliderValueChanged,
+      step: 1,
```
nanaya, line 312: apparently `sliderValueFrom` isn't needed if `step` is set to something like 0.1? (or 0.125 to copy the current "resolution")

nanaya (changes requested):

resources/js/store/store-supporter-tag.tsx
```diff
@@ -0,0 +1,316 @@
...
+
+  @computed
+  get cost() {
+    return Math.floor(this.sliderValue / resolution);
+  }
+
+  get discount() {
+    if (this.duration >= 12) {
+      return 46;
+    }
```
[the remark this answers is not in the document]
notbakaneko: You'll have to ask @peppy about that 🤷

nanaya (approved):

nanaya:
Unrelated but I don't think the slider/button colour is right ಠ_ಠ

notbakaneko:
The slider is supposed to be pink like the supporter icon 👀

PR #11131, opened 2024-03-30
Fix incorrect query syntax

nanaya:
How ಠ_ಠ


notbakaneko (approved):

notbakaneko:
I didn't actually run the previous one 😅

Issue #10238, opened 2023-05-23
Wiki search text box shouldn't have yellow border color when unfocused


Joehuu:
I keep thinking that it's focused and type something but nothing is happening:
![XW4R3DQmUC](https://github.com/ppy/osu-web/assets/35318437/445dfbdf-c9f5-46d4-b045-93511165c6d2)

I would expect the same behavior as the text box in beatmap listing:
![x4nNzFgOG0](https://github.com/ppy/osu-web/assets/35318437/8efea9f7-4c25-46f1-a69c-820e7410c312)

notbakaneko:
I think it makes sense to remove the outline or at least make it look less "highlighted" when not focused @arflyte

peppy:
Agree with removing.

frenzibyte:
Resolved in #11132

PR #11132, opened 2024-03-30
Remove yellow border from wiki search text box when unfocused and added styling when focused.

danielmyren:
This pull request addresses issue #10238

Changes Made:
- Removed the yellow border color from the wiki search text box when unfocused to align with user expectations.
- Added focus styling to enhance user interaction, including a yellow/orange border and a subtle box shadow.  The styling matches the behavior of the search box in the beatmap listing.

osu-css-border-style

### DIFF
--- a/resources/css/bem/wiki-search.less
+++ b/resources/css/bem/wiki-search.less
@@ -20,7 +20,12 @@
     font-size: @font-size--wiki-search;
     background-color: @osu-colour-b6;
 
-    border: @_border-size solid @osu-colour-orange-1;
+    border: @_border-size solid @osu-colour-b6;
+    
+    &:focus-within {
+      box-shadow: 0 0 10px @osu-colour-orange-1;
+      border: @_border-size solid @osu-colour-orange-1
+    }
   }
 
   &__input {


resources/css/bem/wiki-search.less
```diff
@@ -20,7 +20,12 @@
     font-size: @font-size--wiki-search;
     background-color: @osu-colour-b6;
 
-    border: @_border-size solid @osu-colour-orange-1;
+    border: @_border-size solid @osu-colour-b6;
+    
+    &:focus-within {
+      box-shadow: 0 0 10px @osu-colour-orange-1;
+      border: @_border-size solid @osu-colour-orange-1
```
nanaya, line 27: the style doesn't quite match with beatmap search box (wrong border size, rounding, and colour). Also the colour should use page hue css variable, not actual orange.

nanaya (changes requested):

resources/css/bem/wiki-search.less
danielmyren: I have adjusted the border size and the rounding of the border. Do we prefer the thinner border? I've also changed the color to match the other elements on the page instead of the color of wiki-notice. In order to keep the code consistent I'm still using the pre-defined colors from color.less. 


osu-css-border-style2

resources/css/bem/wiki-search.less
```diff
@@ -12,15 +12,19 @@
...
-    .default-border-radius();
     display: flex;
     flex-direction: row;
 
     padding: 13px @_gutter;
     font-size: @font-size--wiki-search;
     background-color: @osu-colour-b6;
 
-    border: @_border-size solid @osu-colour-orange-1;
+    border: 1px solid @osu-colour-b6;
```
nanaya, line 22: update `@_border-size` instead otherwise the suggestion box would have wrong alignment

also don't use `@osu-colour-` anymore. New codes should always use the css variable. Adding more lines that need update isn't helping.

resources/css/bem/wiki-search.less
```diff
@@ -12,15 +12,19 @@
...
+    
+    &:focus-within {
+      box-shadow: 0 0 10px @osu-colour-h2;
```
nanaya, line 25: matching the colour with beatmap search would be preferred but I'll probably refactor the search box later anyway so it's probably fine for now (if not a bit weird when it changes again later)

nanaya (changes requested):

nanaya:
@danielmyren still interested in finishing this PR?

nanaya (approved):

PR #11135, opened 2024-04-01
Add option to always reset password of inactive users

nanaya:
An option to skip country check to trigger reset.


.env.example
```diff
@@ -336,3 +336,6 @@ CLIENT_CHECK_VERSION=false
 
 # USER_COUNTRY_CHANGE_MAX_MIXED_MONTHS=2
 # USER_COUNTRY_CHANGE_MIN_MONTHS=6
+
+#USER_INACTIVE_DAYS_VERIFICATION=180
+#USER_INACTIVE_FORCE_PASSWORD_RESET=false
```
notbakaneko, line 341: can do with a space to be consistent with the reset

notbakaneko (approved):

PR #11141, opened 2024-04-03
Mark score not ranked if not passed

nanaya:
Related to ppy/osu#26369

### DIFF
--- a/app/Models/Solo/Score.php
+++ b/app/Models/Solo/Score.php
@@ -118,11 +118,12 @@ class Score extends Model implements Traits\ReportableInterface
         $params['started_at'] = $scoreToken->created_at;
         $params['user_id'] = $scoreToken->user_id;
 
+        $params['passed'] ??= false;
+        $params['preserve'] = $params['passed'];
+
         $beatmap = $scoreToken->beatmap;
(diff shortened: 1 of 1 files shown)


peppy (approved):

peppy:
I'll run a fixup on all existing scores after rolling this out.

nanaya:
I thought the initial value here is just temporary and there's another process which updates it using a more detailed check?

bdach:
> I thought the initial value here is just temporary and there's another process which updates it using a more detailed check?

If you're talking about `osu-queue-score-statistics` then it doesn't really touch this flag right now. The only intended use case where it'd change value from what this change gives it would be beatmaps getting ranked, or scores getting nuked due to a restrict or something. Which... I'm not sure either of those is actually implemented right now?

`osu-queue-score-statistics` _will_ touch `preserve` once we're done with https://github.com/ppy/osu-queue-score-statistics/issues/141.

I attempted to document this better in https://github.com/ppy/osu-infrastructure/blob/master/score-submission.md#-decide-on-storage-method-and-structure, but given the ongoing perpetual confusion I'm not sure that did much...

PR #11145, opened 2024-04-04
Sort recent scores based on end time instead of update

nanaya:
The update filter stays for partition reason.


peppy (approved):
seemsok

notbakaneko (dismissed):

peppy:
Hold off on merging this momentarily while we discuss the forward direction. I've fixed the specific case which required this manually for now.

peppy (approved):

PR #11148, opened 2024-04-04
Record visible replies count

nanaya:
To be used to determine if a deleted comment should be shown. Draft while I test if this is actually useful.


nanaya:
This ended up useless (the full logic is a bit more complicated than this).

PR #11151, opened 2024-04-09
Fix scores set with unranked mods showing as processing indefinitely (again)

bdach:
- Closes https://github.com/ppy/osu-web/issues/10911 once more.
- Regressed with https://github.com/ppy/osu-queue-score-statistics/pull/227.

The actual fix is the inclusion of and check for existence of the `score_process_history` row as it is the most reliable indicator of processing having concluded. (cc @peppy as you expressed [being hesitant about using it](https://github.com/ppy/osu-queue-score-statistics/pull/227#issuecomment-1970238447))

This also includes a drive-by inclusion of a `preserve === false` check that won't really do anything right now, but will if / once https://github.com/ppy/osu-queue-score-statistics/issues/141 is resolved.

The goal here is to only show pp values on scores that contribute to the user's total. The query determining can be found [here](https://github.com/ppy/osu-queue-score-statistics/blob/f65dec335a914ee171d69981719d6ccb1d3f0ebe/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs#L66-L76).


notbakaneko:
At this point I feel like we're better off checking with mods award pp or just storing when a score isn't supposed to have any pp.
Is every score guaranteed to have a corresponding `score_process_history` row? Does `score_process_history` get truncated?

bdach:
> At this point I feel like we're better off checking with mods award pp

That feels like a bad ending IMO. The logic isn't simple and is (hopefully?) going to be subject to frequent changes as we bring pp to more mods over time.

> or just storing when a score isn't supposed to have any pp.

I guess I could stomach this although that's yet another flag to the collection in the table. Maybe fine (although `ranked` has been confusing everyone multiple times over now, but I can't really bring myself to harp on about this any longer - I think I did what I could in that respect in [documenting what it does](https://github.com/ppy/osu-infrastructure/pull/28)).

> Is every score guaranteed to have a corresponding `score_process_history` row? Does `score_process_history` get truncated?

As far as I'm aware, yes. See: https://github.com/ppy/osu-queue-score-statistics/blob/f65dec335a914ee171d69981719d6ccb1d3f0ebe/osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs#L270-L280.

notbakaneko:
> I dunno what "truncation" would mean in this context.

I mean if old rows every get dropped or deleted from the table.

bdach:
> I mean if old rows ever get dropped or deleted from the table.

Unless the score itself is deleted, currently I don't think so. Even if the score is pushed for reprocessing, the row remains, it'll just get updated after reprocessing is done.

peppy:
As a heads up, I was considering removing the process history table as it's not used for much at all. So there's no guarantee it will exist forever (I'd rather avoid new uses of it >.>).

bdach:
Well that's fine and all, but that leaves me without a clear alternative path. Some were already proposed above, but will require either further implementation or a db schema change and back-migration somehow.

peppy:
Yep understood. I'm gonna have to take a step back and decide how complicated we want/need to make this whole system, because it is feeling increasingly unmaintainable (and I am sure I'm not alone in this train of thought).

app/Transformers/ScoreTransformer.php
```diff
@@ -102,6 +102,7 @@ public function transformSolo(MultiplayerScoreLink|ScoreModel|SoloScore $score)
         if ($score instanceof SoloScore) {
             $extraAttributes['ranked'] = $score->ranked;
             $extraAttributes['preserve'] = $score->preserve;
+            $extraAttributes['processed'] = $score->processHistory()->first() !== null;
```
peppy, line 105: I'm a bit concerned about this being in the transformer here – is there potential that it adds an overhead (+1 query?) in places we don't want it to?

peppy:
@bdach the table is actually very small for process history so i'm fine with using it here for now. the main reason for potentially doing away with it was database overheads, but that is a non-issue for now.

app/Transformers/ScoreTransformer.php
bdach: Hmm well it probably does but I'm not super sure how to do better... It does kinda need to be in the transformer because we'll want to use this in client to mirror this logic too.

peppy:
Let's go ahead with this solution for now. Will defer for review.

app/Transformers/ScoreTransformer.php
notbakaneko: `processHistory` definitely needs to be eager loaded in all the collections it's used in at the very least.

bdach: Well this is where I probably expose myself for a fraud but... do you have any good ways of determining _where_ that should be added? It's a bit impossible to do a "find all usages" as I would with a more statically-typed language (unless I am using inadequate tools, which I might be, since my tools for php are just vscode right now).

I see `MULTIPLAYER_BASE_PRELOAD` in `ScoreTransformer`, and `USER_PROFILE_INCLUDES_PRELOAD` next to it, and that they get fed into `with()` in queries, but those appear to only be specific usage contexts too... I also see `$defaultIncludes` but I don't think that is going to help me here any is it?

notbakaneko, line 105: Should be `$score->processHistory !== null` instead.

notbakaneko (changes requested):
the following places need 'processHistory' added into the eager loads:
https://github.com/ppy/osu-web/blob/97f41e3cb1f1fc985e984454313a90e908da014a/app/Transformers/ScoreTransformer.php#L35

https://github.com/ppy/osu-web/blob/97f41e3cb1f1fc985e984454313a90e908da014a/app/Http/Controllers/BeatmapsController.php#L78

https://github.com/ppy/osu-web/blob/97f41e3cb1f1fc985e984454313a90e908da014a/app/Http/Controllers/BeatmapsController.php#L497
`->records()->loadMissing('processHistory');`

app/Transformers/ScoreTransformer.php
notbakaneko: `$defaultIncludes` is just the transformer defaults

bdach: I changed to `->exists()` in https://github.com/ppy/osu-web/pull/11151/commits/e25f0333a1bb8cfce9691425fce4264f2c3c072c since it seems better but yeah I dunno. Can't seem to get it working without some method call at the end.

bdach:
I've added the requested preloads I think but not entirely sure how to test. As far as I can tell the related views load fine...?

app/Transformers/ScoreTransformer.php
notbakaneko: oh, whoops, forgot to mention you have to add `processHistory` here for the accessor to work:
https://github.com/ppy/osu-web/blob/97f41e3cb1f1fc985e984454313a90e908da014a/app/Models/Solo/Score.php#L242

`processHistory()` returns the relation so any eager loads with be ignored, `processHistory` will use the existing value if already loaded.

bdach: Yeah okay, that helped, thanks :+1:

app/Transformers/ScoreTransformer.php
```diff
@@ -102,6 +104,7 @@ public function transformSolo(MultiplayerScoreLink|ScoreModel|SoloScore $score)
         if ($score instanceof SoloScore) {
             $extraAttributes['ranked'] = $score->ranked;
             $extraAttributes['preserve'] = $score->preserve;
+            $extraAttributes['processed'] = $score->legacy_score_id !== null || $score->processHistory !== null;
```
bdach, line 107: Uhhh... So earlier I said:

> > Is every score guaranteed to have a corresponding `score_process_history` row? Does `score_process_history` get truncated?
> 
> As far as I'm aware, yes.

So as it turns out, I lied here (or if you wanna be charitable, "forgot"). It is not always the case that a `score_process_history` row will exist for a score that is processed.

The exception is scores that have been initially manually batch-imported from `osu_scores_high_*` tables. All legacy scores generally are fully processed, because first `osu-web-10` processes them synchronously, and then the score watchers import them without doing any further processing relevant to the score - so at the point the `scores` row exists for a legacy score, it is already fully processed.

I can't exactly double-check production to confirm this but I realised this when testing on my local database wherein I had a local score import from seeded `osu_scores_high_*` rows into `scores`.

What this says about this particular PR, and also the complexity in and around the new score table, I'm not entirely sure, but I do acknowledge that it's pretty bad...

app/Transformers/ScoreTransformer.php
```diff
@@ -37,6 +38,7 @@ class ScoreTransformer extends TransformerAbstract
         'beatmap.beatmapset',
         // it's for user profile so the user is already available
         // 'user',
+        'processHistory',
```
notbakaneko, line 41: Should be above the comment.

notbakaneko (approved):

PR #11153, opened 2024-04-10
Exclude scores without pp from user best performance

notbakaneko:
Needed since #11086


bdach:
I suppose https://github.com/ppy/osu-queue-score-statistics/pull/227 is the _actual_ point of regression here (again), but at a very quick glance I guess this makes sense?

Could sorta feed into https://github.com/ppy/osu-web/pull/11151. If that flag mentioned over there happens we should probably be using that all over.

peppy (approved):

PR #11162, opened 2024-04-17
Prevent score removal jobs from overlapping

notbakaneko:
Use `WithoutOverlapping` middleware to lock the job. There will be a bogus try added (and also gets counted as successful run) if the job takes longer than `retry_after` to run and the job gets moved to a `:delayed` queue. `InteractsWithQueue` is needed for the actual `$job->release()` behaviour of the queue. There's still the problem if something kills the worker before it can run any fail handlers and the job won't be retried until the timeout/lock expires, but that's not any different than if `retry_after` is changed.

---

The issue is workers will automatically retry reserved jobs after `retry_after` has expired even it they're still running, so the same job starts being run by multiple workers simultaneously and have their attempt counts incremented, eventually causing `MaxAttemptsExceededException` to be thrown (while the job is still running on the first worker). If the job _does_ fail now, it won't be retried anymore.


nanaya:
where did you get that? That's not what the handle function says at least

https://github.com/laravel/framework/blob/ad758500b47964d022addf119600a1b1b0230733/src/Illuminate/Queue/Middleware/WithoutOverlapping.php#L70-L85

`$job->release()` is "Release the job back into the queue after (n) seconds."

notbakaneko:
oh, it also needs the `InteractsWithQueue` trait added to the job to get the default release behaviour :|

It's still not great:
```
worker 1 dequeues job, increments attempts, acquires lock
retries_after expires
worker 2 dequeues job, increments attempts, fails to lock, places on delayed queue, returns job as done
releaseAfter expires
worker 2 dequeues job, increments attempts...
```
It might be passable but it's also still a roundabout way of having another setting that acts like `retries_after` and doesn't apply the first time; you'd think it should always use the `releaseAfter` value 🤔

nanaya:
it's probably still better than adding different queue just to have different timeout

nanaya (changes requested):

nanaya (approved):

PR #11165, opened 2024-04-17
Use command name instead of display name in datadog metric

notbakaneko:
`resolveName()` uses `displayName` which isn't the best for datadog 👀 
(`displayName` defaults to the same value as `commandName` if not specified)


app/Providers/AppServiceProvider.php
```diff
@@ -64,7 +64,7 @@ public function boot()
                 $GLOBALS['cfg']['datadog-helper']['prefix_web'].'.queue.run',
                 1,
                 [
-                    'job' => $event->job->resolveName(),
+                    'job' => $event->job->payload()['data']['commandName'],
```
nanaya, line 67: can use a test to avoid getting trolled by laravel

nanaya (changes requested):

PR #11167, opened 2024-04-18
Convert CurrentUserObserver to typescript

notbakaneko:
Convert `CurrentUserObserver` and `AccountEditBlocklist` as well as remove some unneeded usages of `user:update`


resources/js/core/account-edit-blocklist.ts
```diff
@@ -0,0 +1,35 @@
...
+import { trans } from 'utils/lang';
+
+const contentClass = '.js-account-edit-blocklist-content';
+const countClass = '.js-account-edit-blocklist-count';
+const labelClass = '.js-account-edit-blocklist';
+
+export default class AccountEditBlocklist {
+  constructor() {
+    $(document).on('click', labelClass, this.toggle);
+    $(() => reaction(() => core.currentUser?.blocks, this.updateBlockCount));
```
nanaya, line 15: pass `core` in constructor instead?

resources/js/osu-core.ts
```diff
@@ -140,13 +147,13 @@ export default class OsuCore {
...
     if (currentUser != null) {
-      window.currentUser = currentUser;
-      this.setCurrentUser(window.currentUser);
+      this.setCurrentUser(currentUser);
     }
   };
 
   private readonly onCurrentUserUpdate = (event: unknown, user: CurrentUserJson) => {
     this.setCurrentUser(user);
+    this.currentUserObserver.setAvatars();
```
nanaya, line 156: add reaction in the observer instead?

resources/js/osu-core.ts
```diff
@@ -89,6 +93,8 @@ export default class OsuCore {
       document.addEventListener('DOMContentLoaded', this.updateCurrentUser);
     }
     $.subscribe('user:update', this.onCurrentUserUpdate);
+    // ensure currentUser is updated early enough.
```
nanaya, line 96: I kinda doubt it's still relevant especially with `main.coffee` (where it was before) being loaded after this file

resources/js/osu-core.ts
```diff
@@ -89,6 +93,8 @@ export default class OsuCore {
       document.addEventListener('DOMContentLoaded', this.updateCurrentUser);
     }
     $.subscribe('user:update', this.onCurrentUserUpdate);
```
nanaya, line 95: maybe should look into replacing this with direct function call sometime

nanaya (changes requested):

nanaya (approved):

Issue #10925, opened 2024-01-31
Website should make users aware of lazer mode on own profile when the user has played lazer recently (or only)


peppy:

nanaya:
...what should be shown there?

nphuracm:
Maybe add a banner at somewhere like between the pfp, name and location row and the ranking graph row saying something like
_"Your scores set on osu!(lazer) are being displayed because of lazer mode being active. The status of these scores are subject to change, and might even be wiped altogether in the future."_
It doesn't have to be a really bright and scary, just a nice notice with a dark background (i.e. blends well with the overall color scheme) but the text in yellow or something to ensure that there's still some emphasis there.

nanaya:
with recent scoring updates, is this still needed?

peppy:
I think the minimum requirement is to set a user to their lazer profile if they've only ever played on lazer.

Right now, a new user can create an account, never play on stable, and have their profile stuck in non-lazer-mode until they find the toggle (or ask why they can't see any scores on their profile).

nanaya:
With the new default, is it still needed?

(lazer mode for guests and users with no plays, last score type otherwise)

bdach:
Yeah as far as I'm aware latest consensus was that https://github.com/ppy/osu-web/pull/11168 was enough. Let's close for now until it is deemed not enough.

PR #11168, opened 2024-04-19
Change score display mode to default based on most recent score

notbakaneko:
see https://github.com/ppy/osu-web/issues/10925#issuecomment-2021968877

Changes the lazer mode default for scores displayed setting to be based on the user's most recently set score and saves the setting.

Added some tests for the expected setting based on the conditions.


app/Models/UserProfileCustomization.php
```diff
@@ -195,7 +195,18 @@ public function setForumPostsShowDeletedAttribute($value)
 
     public function getLegacyScoreOnlyAttribute(): bool
     {
-        return $this->options['legacy_score_only'] ?? static::DEFAULTS['legacy_score_only'];
+        $option = $this->options['legacy_score_only'] ?? null;
+        if ($option === null) {
+            $lastScore = $this->user->soloScores()->last();
```
nanaya, line 200: can search by user id directly?

app/Models/UserProfileCustomization.php
```diff
@@ -195,7 +195,18 @@ public function setForumPostsShowDeletedAttribute($value)
...
+            $legacyOnly = $lastScore?->isLegacy() ?? static::DEFAULTS['legacy_score_only'];
+
+            $this->setOption('legacy_score_only', $legacyOnly);
+            $this->save();
+
+            return $legacyOnly;
```
nanaya, line 206: this wouldn't be needed if assigned to `$option` instead

nanaya (changes requested):

nanaya (approved):

Issue #11169, opened 2024-04-21
Hide scores with id 0


tybug:
e.g. https://osu.ppy.sh/scores/osu/0, same with other modes.

> peppy — Yesterday at 10:52 PM
> probably a side-effect of the mapping
> i think we could make an osu-web code rule not to show 0 id
> the case you're looking at is a score which doesn't have an "old" high score id, but does have a new score id. this is because it's not a permanent score
> https://discord.com/channels/188630481301012481/188630652340404224/1231075114028896337

PR #11173, opened 2024-04-22
Prevent bogus score lookups when querying `/scores/{ruleset}/0`

bdach:
Closes https://github.com/ppy/osu-web/issues/11169.

The `scores` table can contain records with `legacy_score_id = 0`. These records correspond to rows from `osu_scores_*` (non-high) tables which are always ephemeral (they get recycled every 24 hours).

To avoid bogus lookups when querying the show controller endpoint, just explicitly disallow 0 (and negative numbers) from being unmapped via `legacy_score_id`.

Decided to just fix this locally because it was the only usage of `legacy_score_id` that I could see being susceptible to this sort of bug - all remaining usages at least try to reach into the source table to get the row and invariably fail since there is no actual row in any of the old score tables with id 0.

Dunno if this needs tests. I didn't seem to find any for this endpoint.


peppy (approved):
Superficial review, looks like what we want logic wise.

nanaya (approved):

PR #11178, opened 2024-04-24
Force cache reset after one minute

nanaya:
Maybe? Although this will incur check/reset before request is handled instead because otherwise there may be one outdated cache being used before it's reset.

<sub>~And probably should make the timeout configurable... 👀~ added </sub>


peppy:
Is this just config or is there more to it?

notbakaneko (approved):

notbakaneko:
Should be just config

PR #11190, opened 2024-05-02
Fix another float/int mismatch

nanaya:
ಠ_ಠ

Could also be updated to compare with int but the variable name kinda implies a float.


peppy (approved):
php scares me

PR #11195, opened 2024-05-07
Skip beatmapset_id for converting to legacy score

nanaya:
It's not actually used in the json.

It's also causing additional queries during conversion for searching played beatmapset because of missing preload.


notbakaneko (changes requested):
Function can do with a note that it doesn't match the model.
...then again, `beatmapset_id` is only on non-best score tables 🤔

nanaya:
I'd rather disable the beatmapset_id attribute if anything

notbakaneko (approved):

PR #11196, opened 2024-05-07
Filter by best legacy scores when searching beatmap in legacy mode

nanaya:
Seems to make more sense 🤷‍♀️


app/Libraries/Search/BeatmapsetSearch.php
```diff
@@ -435,7 +435,12 @@ private function getPlayedBeatmapIds(?array $rank = null)
 
         $topScores = [];
         $showLegacyOnly = ScoreSearchParams::showLegacyForUser($this->params->user) ?? false;
-        $scoreField = $showLegacyOnly ? 'legacy_total_score' : 'total_score';
+        if ($showLegacyOnly) {
+            $scoreField = 'legacy_total_score';
+            $query->where('legacy_score_id', '>', 0);
```
notbakaneko, line 440: Shouldn't this apply to the base query then?

notbakaneko (changes requested):

notbakaneko (approved):

PR #11199, opened 2024-05-08
Avoid negative notification count in document title

MinatoHikari:
emmm


resources/js/models/notification-type.ts
```diff
@@ -90,7 +90,12 @@ export default class NotificationType implements NotificationReadable, Notificat
   @action
   removeStack(stack: NotificationStack) {
     const exists = this.stacks.delete(stack.id);
-    if (exists) this.total -= stack.total;
+    if (exists) {
```
nanaya, line 93: move the comment and negative check logic in https://github.com/ppy/osu-web/blob/e505ae19d805724082ee779c6a1c76ce50f9cbae/resources/js/components/notification-icon.tsx#L15-L18 to a `total` getter (and also create the relevant setter).

nanaya (changes requested):

resources/js/models/notification-type.ts
MinatoHikari: rename `total` to `_total` and create a `total` with getter and setter,
and fix the incorrect judgement of negative total in the other file.

MinatoHikari: wait, in this way, once i opened 2 browser tabs, if  the one's unread is -10, another is 0, i can never get two tabs' synced to something like (2)/(2) and (4)/(4), it will always be (0)/(2) (0)/(4)

MinatoHikari: or should i revert the change about `if (this.isEmpty)` to `if (this.total < 0)`, the origin logic is correct?

nanaya: how did you get into that state in the first place?

MinatoHikari: without the patch, open 2 tabs in browser, than if one tab's unread messages didn't be updated correctly, i will get (-8) shown in one tab and (2) in another

and after i clear the unread notifictions in (2) tab, it becomes

nanaya: how did you get the "than if one tab's unread messages didn't be updated correctly"?

MinatoHikari: > how did you get the "than if one tab's unread messages didn't be updated correctly"?

i don't how, just, opened 2 tabs and leave one active and another inactive, bn post mods, clicked the unread message in active tab, the number in document in 2 tabs became (-4)/(0) (not displayed), and i repeated this behavior for 2-3 times finally it became  (-10)/(0)

MinatoHikari: than bn post new mods, it became (-8)/(2).

nanaya: just remove all the `= 0` assignment changes for now. Those don't fix the underlying problem and only add confusion.

MinatoHikari: fine

resources/js/models/notification-type.ts
```diff
@@ -25,12 +25,12 @@ export function getValidName(value: unknown) {
 }
 
 export default class NotificationType implements NotificationReadable, NotificationDeletable {
+  @observable _total = 0;
```
nanaya, line 28: private

resources/js/components/notification-icon.tsx
```diff
@@ -13,8 +13,7 @@ interface Props {
 }
 
 function format(count: number) {
```
nanaya, line 15: remove this function and call formatNumber directly

nanaya (changes requested):

resources/js/components/notification-icon.tsx
MinatoHikari: sure

resources/js/models/notification-type.ts
MinatoHikari: added

nanaya (approved):

PR #11201, opened 2024-05-09
Only request captcha challenge if visible

nanaya:
That's a rather questionable method of visibility check 👀


notbakaneko (changes requested):
I think it also needs to clean up on navigation 
`[Cloudflare Turnstile] Cannot find Widget cf-chl-widget-jp108, consider using turnstile.remove() to clean up a widget..`

nanaya:
I'll do that in a separate pr as that needs different typing and stuff

notbakaneko (approved):

PR #11204, opened 2024-05-09
Implement creating rooms for queued beatmaps for "daily challenge" system

bdach:
RFC. Opening this for early feedback because I want to continue building on this going forward (and because this is likely to go through a few review rounds because I suck at php).

---

This implements the queueing mechanism described in point (1) of the execution plan of the [design document for the new "daily challenge" system](https://github.com/ppy/osu/issues/28136). The queueing is based around a new database table:

```
MySQL [osu]> describe daily_challenge_queue;
+---------------------+--------------------+------+-----+---------+----------------+
| Field               | Type               | Null | Key | Default | Extra          |
+---------------------+--------------------+------+-----+---------+----------------+
| id                  | bigint unsigned    | NO   | PRI | NULL    | auto_increment |
| beatmap_id          | mediumint unsigned | NO   |     | NULL    |                |
| ruleset_id          | smallint unsigned  | NO   |     | NULL    |                |
| allowed_mods        | json               | YES  |     | NULL    |                |
| required_mods       | json               | YES  |     | NULL    |                |
| order               | smallint unsigned  | YES  |     | NULL    |                |
| multiplayer_room_id | bigint unsigned    | YES  |     | NULL    |                |
+---------------------+--------------------+------+-----+---------+----------------+
```

The table is designed to serve the MVP first and foremost (`beatmap_id` and `ruleset_id`), but also has some future-proofing built in (`allowed` and `required_mods`, also `order` for being able to e.g. swap the items if need be).

Items will be added to this table externally (most likely via ASS, cc @peppy for confirm). There is no selection automation in place.

This PR introduces a new artisan command - `daily-challenge:queue-next` - which takes the item with the lowest outstanding `order` (or `id`, if tied), and creates a new playlists room based on the parameters stored in the queue (beatmap id / ruleset id / mods). The playlist is assumed to run until midnight of the current day.

This is not hooked up to cron yet, because the full flow isn't yet ready. I'll PR a change adding it to cron when the dust has settled on all parts and we have a flow working end-to-end, between web, spectator server, and client.

## Possible contention points

- Naming - willing to listen to suggestions.
- Creating a room requires a user because both the room itself and the playlist items are presumed to be owned by a user. In this case I used BanchoBot, or user w/ ID `$GLOBALS['cfg']['osu']['legacy']['bancho_bot_user_id']`, as a replacement for a user.

  This is especially a bit dodgy due to restrictions placed on users with respect to the number of allowable rooms open at a time:

  https://github.com/ppy/osu-web/blob/8a374412312ec3cc5f4bea7e3c025f807dd23a49/app/Models/Multiplayer/Room.php#L644-L646

  That is why I wrote the command to die if another "daily challenge" room was already open when it was ran, so that check above won't be hit in practice, but yeah. I briefly considered adding a carve-out for the check for BanchoBot specifically but that felt really dodgy so I didn't.

  I imagine this may be considered suboptimal, especially given the `legacy` name in that key lookup there. We could probably try to decouple the multiplayer room model from always needing to have a user although that may be a operation that requires care to a degree that I'm not necessarily sure I'm able to perform it confidently.
- Structurally this reuses the whole `$room->startGame()` logic even though many validation checks are technically superfluous in there. It feels alright to me but it could be argued that this is a server-side flow and thus one might want to refactor that `startPlay()` method into smaller chunks and reuse those. Dunno.
- The created room uses a new category - `daily_challenge`. I'm pretty sure this will be needed, at least for client-side filtering (because we won't want to show these in open room listings, just directly from main menu), but if not, then we can probably talk about alternatives.
 
  One thing that I already know is that this new category as added here will break old lazer clients due to serialisation woes (old versions won't know the new enum value). It is unfortunately too late to do anything about this, we'll just need to tell users to upgrade if this sticks.
- af761d5b41ed02b92d354d9f7d24ae06394be356 is a notable outlier in being a functional change to existing code rather than something completely new. It happened because I was having an issue when writing tests that rooms were getting out by the `active()` scope up to a second after having been opened. I'm not sure why that is - guessing it may be time comparison precision being set at 1 second or something - but it appears to me that a weak inequality doesn't really make sense there anyhow so I just made the change inline.

### DIFF
--- a/app/Console/Commands/DailyChallengeQueueNext.php
+++ b/app/Console/Commands/DailyChallengeQueueNext.php
@@ -5,33 +5,33 @@
 
 namespace App\Console\Commands;
 
-use App\Models\Multiplayer\BeatmapOfTheDayQueueItem;
+use App\Models\Multiplayer\DailyChallengeQueueItem;
 use App\Models\Multiplayer\Room;
 use App\Models\User;
 use Illuminate\Console\Command;
 use Illuminate\Support\Facades\DB;
 
-class BeatmapOfTheDayQueueNext extends Command
+class DailyChallengeQueueNext extends Command
 {
-    protected $signature = 'beatmap-of-the-day:queue-next';
+    protected $signature = 'daily-challenge:queue-next';
 
-    protected $description = 'Creates a new "beatmap of the day" multiplayer room for the first item in queue.';
+    protected $description = 'Creates a new "daily challenge" multiplayer room for the first item in queue.';
 
     public function handle()
     {
-        $nextQueueItem = BeatmapOfTheDayQueueItem::whereNull('multiplayer_room_id')
+        $nextQueueItem = DailyChallengeQueueItem::whereNull('multiplayer_room_id')
             ->orderBy('order', 'asc')
             ->orderBy('id', 'asc')
             ->first();
 
         if ($nextQueueItem === null) {
-            $this->error('Beatmap of the day queue is empty.');
+            $this->error('"Daily challenge" queue is empty.');
             return 1;
         }
 
-        $existing = Room::where('category', 'beatmap_of_the_day')->active()->first();
+        $existing = Room::where('category', 'daily_challenge')->active()->first();
         if ($existing !== null) {
-            $this->error("Another 'beatmap of the day' room is open (id {$existing->id}).");
+            $this->error("Another \"daily challenge\" room is open (id {$existing->id}).");
             return 1;
         }
 
@@ -43,7 +43,7 @@ class BeatmapOfTheDayQueueNext extends Command
                 User::lookup($ownerId),
                 [
                     'ends_at' => today()->addDay(),
-                    'name' => "Beatmap of the Day: {$startTime->toFormattedDateString()}",
+                    'name' => "Daily Challenge: {$startTime->toFormattedDateString()}",
                     'type' => 'playlists',
                     'playlist' => [[
                         'beatmap_id' => $nextQueueItem->beatmap_id,
@@ -53,7 +53,7 @@ class BeatmapOfTheDayQueueNext extends Command
                     ]],
                 ]
             );
-            $room->update(['category' => 'beatmap_of_the_day']);
+            $room->update(['category' => 'daily_challenge']);
 
             $nextQueueItem->multiplayer_room_id = $room->id;
             $nextQueueItem->save();
--- a/app/Console/Commands/DailyChallengeQueueNext.php
+++ b/app/Console/Commands/DailyChallengeQueueNext.php
@@ -40,7 +40,7 @@ class DailyChallengeQueueNext extends Command
             $ownerId = $GLOBALS['cfg']['osu']['legacy']['bancho_bot_user_id'];
 
             $room = (new Room())->startGame(
-                User::lookup($ownerId),
+                User::findOrFail($ownerId),
                 [
                     'ends_at' => today()->addDay(),
                     'name' => "Daily Challenge: {$startTime->toFormattedDateString()}",
--- a/app/Console/Commands/DailyChallengeQueueNext.php
+++ b/app/Console/Commands/DailyChallengeQueueNext.php
@@ -13,9 +13,9 @@ use Illuminate\Support\Facades\DB;
 
 class DailyChallengeQueueNext extends Command
 {
-    protected $signature = 'daily-challenge:queue-next';
-
     protected $description = 'Creates a new "daily challenge" multiplayer room for the first item in queue.';
+
+    protected $signature = 'daily-challenge:queue-next';
 
     public function handle()
     {
--- a/tests/Commands/DailyChallengeQueueNextTest.php
+++ b/tests/Commands/DailyChallengeQueueNextTest.php
@@ -12,8 +12,6 @@ use Tests\TestCase;
 
 class DailyChallengeQueueNextTest extends TestCase
 {
-    private int $originalUtilityUserId;
-
     public function testBasicItem()
     {
         $queueItem = DailyChallengeQueueItem::factory()->create();
@@ -111,14 +109,6 @@ class DailyChallengeQueueNextTest extends TestCase
         parent::setUp();
 
         $utilityUser = User::factory()->create();
-        $this->originalUtilityUserId = $GLOBALS['cfg']['osu']['legacy']['bancho_bot_user_id'];
         config_set('osu.legacy.bancho_bot_user_id', $utilityUser->getKey());
-    }
-
-    protected function tearDown(): void
-    {
-        parent::tearDown();
-
-        config_set('osu.legacy.bancho_bot_user_id', $this->originalUtilityUserId);
     }
 }
--- a/tests/Commands/DailyChallengeQueueNextTest.php
+++ b/tests/Commands/DailyChallengeQueueNextTest.php
@@ -5,20 +5,20 @@
 
 namespace Tests\Commands;
 
-use App\Models\Multiplayer\BeatmapOfTheDayQueueItem;
+use App\Models\Multiplayer\DailyChallengeQueueItem;
 use App\Models\Multiplayer\Room;
 use App\Models\User;
 use Tests\TestCase;
 
-class BeatmapOfTheDayQueueNextTest extends TestCase
+class DailyChallengeQueueNextTest extends TestCase
 {
     private int $originalUtilityUserId;
 
     public function testBasicItem()
     {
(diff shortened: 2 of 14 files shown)


cl8n:
`BeatmapOfTheDayQueueItem` having most of the data of  `PlaylistItem` makes me feel like it would be more straightforward to manage as orphaned playlist items (with room ID 0 or something), without a new table. the whole purpose of this queue is to eventually translate into playlist items anyway, right?

or even creating the entire room in advance to act as the "queue" (with `starts_at` set to the future)

bdach:
> `BeatmapOfTheDayQueueItem` having most of the data of `PlaylistItem` makes me feel like it would be more straightforward to manage as orphaned playlist items (with room ID 0 or something), without a new table

I really don't like this idea in general if I'm honest. It feels sloppy - I've really been historically bitten *way* too much by "special" in-band values like zero IDs having some special meaning like this, which then leads to pitfalls when somebody tries to follow them like a normal foreign key. It also closes off potential avenues of extension. For instance, imagine if this takes and a few months down the line we want to add something like collective goals attached to the beatmap of the day, like getting however many A ranks across the playerbase. At that point I dunno where you put this information if not into this table - some new table?

Although you could arguably say that if that is the eventual direction then the `queue` suffix needs to be dropped from the table. Which I guess is easy enough if we decide on that.

nanaya:
Also mentioned above, how about creating the room with starts_at set to the future? Whatever the custom settings it'll have, I imagine it'll be linked to the room (or the respective playlist items) anyway.

bdach:
Room with a future date sounds much cleaner from a design angle (no more weird detached playlist items whose existence must be remembered), but also it implies that whatever will be responsible for queueing the beatmaps of the day will have to reimplement the logic responsible for creating the actual room, e.g. the entirety of `$room->startGame()`. And there are ruleset validity checks there, there is chat channel creation for the playlist, etc.

I dunno. I'm most likely not going to be the one implementing the enqueueing of items so from my perspective I can't pretend I can push back too hard on that but it does appear to me that if ASS is to create a full playlists room by itself it may become redundant (with respect to code duplication) and/or inconvenient.

nanaya:
Or can just implement an interop endpoint which accepts some set of parameters and then create the room...

bdach:
Yep that would also work. Really need a shot-call from @peppy at this point to commit to either direction because I can't see into ASS source and don't have much of an insight as to what would be easiest/best here.

peppy:
Hmm... if I was implementing this I too would have gone for a separate table as done here, just to implement cleanly and in an isolated manner. We've learnt the hard way from trying to make one model work for too many different usages, and that feels like it could be the case here.

As for whether we use a future `starts_at`, I guess it depends on whether we want the complexity at the queueing mechanism or the storage. I'm weakly leaning towards how it's done here.

I would see ASS being a very simple "queue more beatmap / mod criteria to queue table" and osu-web handling the actual room creation. If this was to change, it would be hard to orchestrate checks such as "queue is running low!" being broadcast to discord or otherwise, as ASS doesn't have any kind of cron concept.

I guess osu-web could still do the creation ahead of time and use a future `starts_at` if that makes timings better though? It would be a good tool to have in our arsenal to be able to schedule future rooms.

nanaya:
well the table here is just an interim table before eventually transformed into rooms and playlists. Having an interop which creates the room immediately skips that additional step.

If the actual system doesn't use room/playlist at all then that'd make more sense I guess.

peppy:
> With a separate table scheduling future rooms is finicky 

Yeah I meant more for manual stuff in the future.

> Having an interop which creates the room immediately skips that additional step.

It'll get a bit annoying if we need to make changes, especially to ordering. Like rather than having a cron task that runs once a day and pulls a new room in, we have all these future rooms already lined up, which potentially need date changes, deletion etc.

bdach:
Blocking temporarily pending confirmation of rename of the feature (see https://discord.com/channels/188630481301012481/188630652340404224/1239804074355068959)

bdach:
Rename of feature to "daily challenge" has been applied.

If anyone other than me has ran this on their local envs, first wipe all `beatmap_of_the_day` category rooms (or move them elsewhere), and then run `artisan migrate:rollback --step=1` **before pulling**, then pull, then re-apply the migration from this PR with the naming changes again. Not sure if there's a better way to do this, but also not sure anybody except me cares.

nanaya:
from what I gather, there will be a few other things to be run when the new room is open. How will the schedule be coordinated? Is it checking the rooms table and see if the new room has been created? What if the creation failed due to some reason (like db error)?

bdach:
Currently the intended flow for this to work is as such:

1. "Daily challenge" beatmaps are queued manually by a selected team, via internal discord / ASS commands. This translates to insertion of `daily_challenge_queue` rows.
2. The command added here will be hooked up to a cron and will pick up the queue items on a daily basis and make rooms out of them, say every day on 00:05 UTC. (The five minute gap would be to avoid various funniness around the previous room not quite completing yet or something.)
3. `osu-server-spectator` will poll the db for open `daily_challenge` rooms on an interval. Something like once every five minutes. Once it detects a change in the active daily challenge room, it broadcasts this info to all connected clients.

> from what I gather, there will be a few other things to be run when the new room is open

Not super sure what this is referring to.

> How will the schedule be coordinated?

I'm not sure what the specific question is here but for now it is best-effort, hard-coded, daily. Rooms are created daily as long as there exist any queued beatmaps for the daily challenge thing.

> What if the creation failed due to some reason (like db error)?

The presumption is that step (2) will fail and `osu-server-spectator` will advertise nothing until we fix the job and manually re-run it.

> an alternative would be the table to be completely managed by ass and it just hit an interop endpoint to create the room when it starts maybe?

I'm not sure what else to say to that other than what has been already said above.

nanaya:
> Not super sure what this is referring to.

the number 3 there.

> I'm not sure what else to say to that other than what has been already said above.

it'll allow all the process to be run in sync by a single process, with its own failure handling. And possibly removes the need to poll the table

bdach:
> And possibly removes the need to poll the table

Unless ASS was to RPC spectator server too to tell it that there is a new room, it doesn't. Unless you're thinking that the interop endpoint could issue like a notifications server message, which, again, I'm neutral on this being the case, were if not for @peppy expressing clear opposition to using the notifications server for delivery to clients on this feature (which is understandable, since expansion of this feature will be much easier to do in spectator server *if* the expansion happens).

nanaya:
well, I guess this is fine as long there's someone to handle rerunning osu-web job in case of failure...

tests/Commands/DailyChallengeQueueNextTest.php
```diff
@@ -0,0 +1,124 @@
...
+        $this->assertNotNull($firstItem->fresh()->multiplayerRoom()->first());
+        $this->assertNull($secondItem->fresh()->multiplayerRoom()->first());
+    }
+
+    public function testNothingDoneOnEmptyQueue()
+    {
+        $this->artisan('daily-challenge:queue-next')
+            ->expectsOutputToContain('"Daily challenge" queue is empty')
+            ->assertFailed();
+        $this->assertSame(0, Room::all()->count());
```
nanaya, line 86: use `expectCountChange` instead

app/Console/Commands/DailyChallengeQueueNext.php
```diff
@@ -0,0 +1,64 @@
...
+
+        $existing = Room::where('category', 'daily_challenge')->active()->first();
+        if ($existing !== null) {
+            $this->error("Another \"daily challenge\" room is open (id {$existing->id}).");
+            return 1;
+        }
+
+        DB::transaction(function () use ($nextQueueItem) {
+            $startTime = today();
+            $ownerId = $GLOBALS['cfg']['osu']['legacy']['bancho_bot_user_id'];
```
nanaya, line 40: I think it's more like `host` than `owner`

nanaya (changes requested):

app/Console/Commands/DailyChallengeQueueNext.php
bdach: Yeah there's "room host" but also "playlist item owner". Mixed these up a bit :sweat_smile:

nanaya (approved):
seems working although should be remembered that any failure (including transient errors like db temporarily blipping) on the cron job means it'll need to be manually rerun by someone who has access to osu-web console as I don't think it's got any retry mechanism

(either that or schedule the cron multiple times or something)

nanaya:
oh yeah, the title should probably be adjusted because it's not queuing anything but more like creating queued room.

peppy (approved):

Issue #10673, opened 2023-10-23
Beatmap contest entry titles should be clickable


Walavouchey:
[example listing](https://osu.ppy.sh/community/contests/180)

it's not obvious that the thumbnails are clickable, while i'd expect the text to be

![image](https://github.com/ppy/osu-web/assets/36758269/b2c170d3-2e77-42b3-a99b-8dc0dd39e9f5)

![image](https://github.com/ppy/osu-web/assets/36758269/aad07a5d-8914-4e5c-9f84-8e62e2710278)

Joehuu:
I'd also want the thumbnails to play the song preview to be consistent with beatmap and featured artist listing behavior.

itsayopapi:
@Walavouchey I would like to work on this..please direct me to the codebase where the issue is coming from

PR #11206, opened 2024-05-11
Add links to contest entry title

Exotica0122:
Resolves: #10673

Added links to beapmap in contest entry list


resources/css/bem/contest-voting-list.less
```diff
@@ -123,6 +123,11 @@
       width: 100%;
       padding: 2px 10px;
     }
+
+    &--link {
+      color: white;
+      .link-hover({ color: white });
```
cl8n, line 129: helper for this: `.link-white();`

resources/js/contest-voting/entry.coffee
```diff
@@ -61,11 +61,19 @@ export class Entry extends React.Component
       if @props.contest.show_votes
         div className: 'contest-voting-list__title contest-voting-list__title--show-votes',
           div className: 'contest-voting-list__votes-bar', style: { width: "#{relativeVotePercentage}%" }
-          div className: 'u-relative u-ellipsis-overflow', entry_title
+          div className: 'u-relative u-ellipsis-overflow',
+            a
```
cl8n, line 65: the extra layer of div isn't necessary, as long as you make the link `display: block`. to combine these "utility" styles into the title link, add lines like `.u-ellipsis-overflow();` into the styles

so u may end up with something like...
```less
&--link {
  .u-ellipsis-overflow();
  .u-relative();

  display: block;
  
  // ...
}
```

also, it seems `u-relative` may not be needed either? I'm not sure why it's here

resources/js/contest-voting/entry.coffee
```diff
@@ -61,11 +61,19 @@ export class Entry extends React.Component
...
+              className: 'contest-voting-list__title--link',
```
cl8n, line 66: in [BEM](https://getbem.com/naming/) the "modifiers" (`--link`) are for differetn versions of an element, but here it's something below `title` and not a variation of it. so i think this can be `contest-voting-list__title-link` instead, and put `&__title-link { ... }` at the same level as `&__title` in the css

resources/js/contest-voting/entry.coffee
```diff
@@ -61,11 +61,19 @@ export class Entry extends React.Component
...
+              href: route('beatmapsets.show', beatmapset: @props.entry.preview)
```
cl8n, line 67: you pointed this out on discord already but just reminding, yeah this is a special case for `submitted_beatmaps` contests and `@props.entry.preview` would otherwise be a url on its own. deciding the URL should share logic with the existing icon links on line 50 & 57

cl8n (changes requested):

nanaya:
it doesn't even compile

Exotica0122:
> it doesn't even compile

My bad had a typo in css

resources/js/contest-voting/entry.coffee
```diff
@@ -90,6 +89,19 @@ export class Entry extends React.Component
             target: '_blank'
             i className: 'fas fa-fw fa-lg fa-external-link-alt'
 
+  renderTitle: (entry_title) ->
```
nanaya, line 92: entry title should just be set inside this function and make sure to not put `a` inside `a`.

(also new variable names should be camelCase)

resources/js/contest-voting/entry.coffee
```diff
@@ -61,12 +61,11 @@ export class Entry extends React.Component
...
         div className: 'contest-voting-list__title contest-voting-list__title--show-votes',
           div className: 'contest-voting-list__votes-bar', style: { width: "#{relativeVotePercentage}%" }
-          div className: 'u-relative u-ellipsis-overflow', entry_title
-          @renderUserLink()
+          @renderTitle(entry_title)
       else
         div className: 'contest-voting-list__title',
-          div className: 'u-ellipsis-overflow', entry_title
-          @renderUserLink()
+        @renderTitle(entry_title)
```
nanaya, line 67: can be inlined

resources/js/contest-voting/entry.coffee
```diff
@@ -90,6 +89,19 @@ export class Entry extends React.Component
             target: '_blank'
             i className: 'fas fa-fw fa-lg fa-external-link-alt'
 
+  renderTitle: (entry_title) ->
+    if @props.options.showLink && @props.entry.preview && @props.contest.submitted_beatmaps
+      el React.Fragment, null,
```
nanaya, line 94: put the fragment and user link render outside instead

```
el Fragment, ...
  if ...
    ...
  else
    ...
  @renderUserLink()
```

resources/css/bem/contest-voting-list.less
```diff
@@ -125,6 +125,13 @@
     }
   }
 
+  &__title-link {
+    .link-white();
+    .u-relative();
```
nanaya, line 130: don't include utility class like this; use normal css instead

nanaya (changes requested):

resources/css/bem/contest-voting-list.less
Exotica0122: Is it a code style thing? Just curious because I"m seeing other files use utility functions just like this

resources/js/contest-voting/entry.coffee
Exotica0122: Removed `entry_title` since the `a` tag is used only in the title anyways

resources/css/bem/contest-voting-list.less
nanaya: yes those are mistakes

Exotica0122: Is it ok to use it in an element className?

nanaya: as in the js? well yeah

nanaya: utility classes are the ones with `u-` prefix...

Exotica0122: oh so the link-white is fine to use?

nanaya (changes requested):

resources/css/bem/contest-voting-list.less
nanaya: well yes, it's a function, not a class

resources/js/contest-voting/entry.coffee
```diff
@@ -90,6 +77,23 @@ export class Entry extends React.Component
             target: '_blank'
             i className: 'fas fa-fw fa-lg fa-external-link-alt'
 
+  renderTitle: ->
+    el React.Fragment, null,
+      if @props.contest.type == 'external'
+        a
+          className: 'contest-voting-list__title-link'
```
nanaya, line 84: this should be styled the same as the beatmap link below

nanaya (changes requested):

nanaya (approved):

PR #11209, opened 2024-05-13
Convert AccountEdit and AccountEditAvatar to typescript

notbakaneko:
Moved the extensions on the html element handling the state into a separate class. 

Also removed the `value == prevValue` check since the request can be aborted before or after the server has already processed it, which means the last value may or may not be wrong already. 

~~A proper check would be to check against the known saved value or initial value which doesn't happen since the way it's currently set up, the first known value is after the change.~~
~~only fixed initial value check for where the `defaultValue` is usable (so not checkboxes)~~
Removed the no update on default check because things may not update as expected when navigating away and restoring the page state, regardless of cancelling the xhr or not due to the actually state being nebulous unless all the settings are fully reloaded.


resources/js/core/account-edit-state.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    this.saving();
+    this.debouncedUpdate();
+  }
+
+  saved() {
+    window.clearTimeout(this.timeout);
+    this.dataset.accountEditState = 'saved';
+    this.timeout = window.setTimeout(() => {
+      this.dataset.accountEditState = '';
```
nanaya, line 63: `setTimeout(this.clear, ...)`?

resources/js/core/account-edit-state.ts
```diff
@@ -0,0 +1,130 @@
...
+    }, 3000);
+  }
+
+  saving() {
+    window.clearTimeout(this.timeout);
+    this.dataset.accountEditState = 'saving';
+  }
+
+  private getMultiValue() {
```
nanaya, line 72: this and `getValue` can probably be merged into `data`?

(and if anything this is more like `getMultiData` since it includes the field name)

resources/js/core/account-edit-state.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    return data;
+  }
+
+  private getValue() {
+    let value: string | string[] | undefined;
+
+    if (this.dataset.accountEditType === 'array') {
+      value = [];
```
nanaya, line 86: the initial empty string is required to allow unselecting everything

nanaya (changes requested):

nanaya (changes requested):

nanaya (approved):

PR #11211, opened 2024-05-13
Read & store score total without mods on submission

bdach:
This is being sent from the client after https://github.com/ppy/osu/pull/27912. There's no actual release with that pull out yet, so if you wanna test full-stack, test with latest game `master`.

The intended place of storage is the `data` json. @peppy initially wanted to do a new column for this, but unfortunately due to the fact that columns cannot be added online to `ROW_FORMAT=COMPRESSED` tables, this is not very easy to do on a table with several billion records inside. Thus in the json it goes. It's not guaranteed json will be much better either due to the unpredictable behaviour of json compression, but it seems to be the best way forward.

The param is nullable / allowed-missing for now. It will probably cease to be after submission is disabled for clients that don't send it, and all rows that have it missing are back-populated.

The end goal in sight here is to be able to do perpetual mod multiplier rebalances without precision loss, e.g. https://github.com/ppy/osu/issues/27685. In the future this could also prove useful for something like custom mod multiplier support for tournaments.


bdach:
> seems working apart of the interface in `solo-score-json.ts` can use an update.

Correct me if I'm wrong but that looks like a frontend-facing interface? I wasn't really intending on exposing this to the frontend at all, it's basically only intended for internal / maintenance usage for when we do need to rebalance mod multipliers.

nanaya:
yeah but the it kind of documents what attributes the json has

bdach:
Does https://github.com/ppy/osu-web/pull/11211/commits/1ded710c9554c22a102430807b6e825c60ea9955 work?

resources/js/interfaces/solo-score-json.ts
```diff
@@ -48,6 +48,7 @@ type SoloScoreJsonDefaultAttributes = {
   started_at: string | null;
   statistics: Partial<Record<SoloScoreStatisticsAttribute, number>>;
   total_score: number;
+  total_score_without_mods?: boolean;
```
nanaya, line 51: yes but why is it a boolean 🤔

nanaya (changes requested):

resources/js/interfaces/solo-score-json.ts
bdach: whoooooooooooooooooops

nanaya (approved):
kinda wonder about the length of the key though (`total_score_without_mods`)

peppy (approved):

PR #11216, opened 2024-05-16
Use mobx for follow user button

notbakaneko:
Also fix the `being read outside a reactive context` warning on user card tooltips


nanaya (changes requested):

resources/js/components/follow-toggle.tsx
```diff
@@ -3,58 +3,50 @@
...
-  };
-
-  state: State;
-
-  private readonly eventId = `follow-toggle-${nextVal()}`;
-  private toggleXhr: null | JQueryXHR = null;
+@observer
+export default class FollowToggle extends React.PureComponent<Props> {
+  @observable private _following = true;
+  private readonly jsonId = `json-follow-toggle-${this.props.follow.subtype}-${this.props.follow.notifiable_type}-${this.props.follow.notifiable_id}`;
```
nanaya, line 26: this should be a getter? doesn't matter now but it'll probably break if the component is ever shuffled around and get passed a different set of props

nanaya (changes requested):

nanaya (approved):

PR #11222, opened 2024-05-20
Support multiple length scales on beatmap filtering

### LINKED ISSUE #10938: Support beatmap length filtering with multiple scales
In the lazer song select menu, you can filter using a query like `length<=2m30s`. Relevant code: https://github.com/ppy/osu/blob/7cdedfb6ca408b8ace311944bc3bf8b8454c515d/osu.Game/Screens/Select/FilterQueryParser.cs#L342

From what I can tell, this doesn't work in the web search. `length<=2m` works and `length<=30s` works but not with the minute and second scales combined (`length<=2m30s`). I don't think there should be a discrepancy when using the same query but in the web search.

tsunyoku:
Resolves #10938.

Intending to match the length filtering available in the client so added some test cases which are taken from there.


app/helpers.php
```diff
@@ -1439,36 +1439,57 @@ function get_length($string): ?array
...
-    $scaleKey = substr($string, -2);
+    $time = null;
+    $minScale = 3600000;
 
-    if (!isset($scales[$scaleKey])) {
-        $scaleKey = substr($scaleKey, -1);
-    }
+    foreach ($patterns as $pattern) {
+        $match = preg_match($pattern, $string, $matches);
+        if ($match === 0 || $match === false) {
```
nanaya, line 1459: check for `!== 1` instead?

app/helpers.php
```diff
@@ -1439,36 +1439,57 @@ function get_length($string): ?array
         'h' => 3600,
     ];
 
+    static $patterns = [
+        '/^((?P<hours>\d+):)?(?P<minutes>\d+):(?P<seconds>\d+)$/',
+        '/^((?P<hours>\d+(\.\d+)?)h)?((?P<minutes>\d+(\.\d+)?)m)?((?P<seconds>\d+(\.\d+)?)s)?((?P<milliseconds>\d+(\.\d+)?)ms)?$/',
+        '/^(?P<seconds>\d+(\.\d+)?)$/',
```
nanaya, line 1445: the `P`s are useless and just adding noise

app/helpers.php
```diff
@@ -1439,36 +1439,57 @@ function get_length($string): ?array
...
+        if (isset($matches['hours'])) {
+            $scale = $scales['h'];
+            $minScale = min($minScale, $scale);
+            $time += get_float($matches['hours']) * $scale;
+        }
+
+        break;
+    }
+
+    return ['value' => $time, 'min_scale' => $minScale];
```
nanaya, line 1492: `ComparatorParam` needs an update to handle this change

nanaya (changes requested):

peppy:
@tsunyoku ping on this one (requires review application)

PR #11223, opened 2024-05-21
Record user country from client connection

nanaya:
Reference: ppy/osu-server-spectator#228

- [ ] confirm if it's fine to ignore the count column

### DIFF
--- a/app/Http/Middleware/UpdateUserInfo.php
+++ b/app/Http/Middleware/UpdateUserInfo.php
@@ -11,7 +11,7 @@ use Carbon\Carbon;
 use Closure;
 use Illuminate\Contracts\Auth\Guard;
 
-class UpdateUserLastvisit
+class UpdateUserInfo
 {
     protected $auth;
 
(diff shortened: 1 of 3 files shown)


app/Http/Middleware/UpdateUserInfo.php
```diff
@@ -42,6 +42,13 @@ public function handle($request, Closure $next)
                         $user->update([
                             'user_lastvisit' => $currentLastVisit,
                         ], ['skipValidations' => true]);
+
+                        if ($token !== null) {
```
peppy, line 46: What are the cases where token is non-null here? Just thinking about the `count` column – we'd probably want to increment it, but not if it's going to increment too often.

nanaya: when it's a password client token

peppy: on every request?

peppy: Hmm, might be okay to update count if it's debounced to 5 minutes.

peppy (approved):

peppy (approved):

PR #11224, opened 2024-05-21
Check against specific sql unique constraint exception

nanaya:
Apparently this is a thing now.

Not actually sure if all the unique exceptions are now wrapped in that class but there's one way to find out 👀


notbakaneko (approved):
seeing as the laravel check for throwing it is just a wrapper around a regex for `Integrity constraint violation: 1062` 🤷

PR #11227, opened 2024-05-23
Fix forum topic cover handling

### LINKED ISSUE #11226: Topic covers not viewable or editable
I briefly looked into this and it seems like somewhere around #10703 or #10655 is where this all stopped working. not 100% sure though

if it's as bad as it sounds (entirely broken for 6 months) and nobody reported it, I feel like that says something about how low the demand is for this feature... so maybe relevant as a future design consideration or something.

nanaya:
Resolves #11226.

- incorrect cover file check
- missing topic id on relevant model

### DIFF
--- a/app/Http/Controllers/Forum/TopicsController.php
+++ b/app/Http/Controllers/Forum/TopicsController.php
@@ -431,7 +431,7 @@ class TopicsController extends Controller
         $topic->incrementViewCount($currentUser, \Request::ip());
         $posts->last()->markRead($currentUser);
 
-        // initiate new cover model separately to prevent it from getting used by opengraph
+        // Instantiate new cover model separately to prevent it from getting used by opengraph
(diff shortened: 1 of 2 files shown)


cl8n:
can still reproduce part of the issue by:
- make a new post
- upload a topic cover (using the edit cover button)
- refresh page

app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -431,7 +431,8 @@ public function show($id)
         $topic->incrementViewCount($currentUser, \Request::ip());
         $posts->last()->markRead($currentUser);
 
-        $coverModel = $topic->cover ?? new TopicCover();
+        // initiate new cover model separately to prevent it from getting used by opengraph
```
cl8n, line 434:
```suggestion
        // Instantiate new cover model separately to prevent it from getting used by opengraph
```
:thinking:

cl8n (approved):

notbakaneko (approved):

PR #11236, opened 2024-05-27
Add datadog increment wrapper

notbakaneko:
Not sure if useful but needing to prefix most of the time `$GLOBALS['cfg']['datadog-helper']['prefix_web']` seems repetitive 🤔  (and samplerate always being 1)

Renamed the `thumbdonger.*`counters to match the class name and removed the `processed` counter since its just the sum of the other 2.


nanaya:
I wonder if we should just move everything (thumbs and chats) to `osu.web.` prefix

nanaya:
it looks like osu-web is the only source now

notbakaneko:
I propose renaming the `thumbdonger.*` stats to something else since the thumbnailer itself already sends its own stats...

nanaya:
I guess we can look into renaming the metrics later and get this in first. Also maybe use named parameter and pass everything as is to the actual function?

nanaya (approved):

PR #11240, opened 2024-05-29
Do not return daily challenge rooms in `GET /rooms` endpoint if API version too low

bdach:
RFC.

Currently old lazer clients die on these due to deserialisation woes (it can't figure out what enum value to deserialise `daily_challenge` to as it doesn't exist yet), which makes deployment of the feature a bit tricky logistically.

Since we can't fix old clients (because they're already released, and we would have to push an update either way), hide daily challenge rooms behind an `x-api-version` header check. Note that challenge rooms can still be accessed directly by ID regardless of API version, but that's fine for this case, since the client won't be doing it.

### DIFF
--- a/app/Http/Controllers/Multiplayer/RoomsController.php
+++ b/app/Http/Controllers/Multiplayer/RoomsController.php
@@ -46,7 +46,15 @@ class RoomsController extends BaseController
         }
 
         $search = Room::search($params);
-        $rooms = $search['query']
+        $query = $search['query'];
+
+        // temporary workaround for lazer client failing to deserialise `daily_challenge` room category
+        // can be removed 20241129
+        if ($apiVersion < 20240529) {
+            $query->whereNot('category', 'daily_challenge');
+        }
+
+        $rooms = $query
             ->with($includes)
             ->withRecentParticipantIds()
             ->get();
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Multiplayer/RoomsController.php
```diff
@@ -46,7 +46,15 @@ public function index()
         }
 
         $search = Room::search($params);
-        $rooms = $search['query']
+        $query = $search['query'];
+
+        // temporary workaround for lazer client failing to deserialise `daily_challenge` room category
+        // can be removed 20241129
+        if ($apiVersion < 20240529) {
```
bdach, line 53: Not sure if you do this sort of thing around these parts, but I don't foresee this living too long, so I did the standard that we do in client of only keeping this around for ~6 months at which point it can be nuked.

peppy (dismissed):
conceptually correct

app/Http/Controllers/Multiplayer/RoomsController.php
nanaya: maybe note that it's temporary in the docs?

nanaya (approved):

PR #11254, opened 2024-06-05
Rename GameMode to Ruleset

nanaya:
Just get this done already. Variables can be updated later as needed.

Also took a quick look into using typescript enum but not quite sure how useful it'll be 🤷 

There doesn't seem to be a way to obtain array of enum elements so that'll be at least one duplicated thing.


notbakaneko (approved):
should be safe 🤔

PR #11255, opened 2024-06-06
Remove support for Centili payments

notbakaneko:
Getting this part out of the way first without all the other store-related cleanups so it can be reverted by itself if needed.
We haven't enabled centili for a while now so we can just remove it completely.


resources/css/bem/store-payment-button.less
```diff
@@ -24,10 +24,6 @@
     box-shadow: 0 0 0 3px @osu-colour-l4;
   }
 
-  &--centili {
-    background-image: url('~@images/store/ycoins.png');
```
nanaya, line 28: the image can also be deleted

nanaya (changes requested):

nanaya (approved):

PR #11257, opened 2024-06-09
Add recent score index

peppy:
Used primarily for new medals.

@nanaya may want to switch the recent section of profile query to use this index now.

### DIFF
--- a/database/migrations/2024_06_09_091512_add_recent_score_index.php
+++ b/database/migrations/2024_06_09_091512_add_recent_score_index.php
@@ -1,0 +1,33 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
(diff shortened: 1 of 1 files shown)


database/migrations/2024_06_09_091512_add_recent_score_index.php
```diff
@@ -26,7 +26,7 @@ public function up(): void
      */
     public function down(): void
     {
-        Schema::table('daily_challenge_queue', function (Blueprint $table) {
+        Schema::table('scores', function (Blueprint $table) {
```
peppy, line 29: oops

bdach:
> It may get pretty slow when also filtering by ruleset id

For whatever it's worth this index was originally for medal purposes (to look up latest N scores of a user) and it seems to perform pretty well even when filtering by ruleset ID which some medals do. I guess that said the biggest N scores that medals will be fetching is like 4, so that may not exactly match the use case.

nanaya (approved):

PR #11261, opened 2024-06-11
Allow reporting same object after some period

nanaya:
- [x] decide the cooldown duration (or maybe make it configurable?)

### DIFF
--- a/resources/lang/en/errors.php
+++ b/resources/lang/en/errors.php
@@ -33,4 +33,7 @@ return [
         'invalid_cursor_exception' => 'Invalid cursor parameter specified.',
         'operation_timeout_exception' => 'Search is currently busier than usual, try again later.',
     ],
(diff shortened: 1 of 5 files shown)


resources/lang/en/errors.php
```diff
@@ -33,4 +33,7 @@
         'invalid_cursor_exception' => 'Invalid cursor parameter specified.',
         'operation_timeout_exception' => 'Search is currently busier than usual, try again later.',
     ],
+    'user_report' => [
+        'recently_reported' => 'Already reported recently.',
```
notbakaneko, line 37: `You've already reported this recently`.
Type might be nice to have but maybe later 🤷

notbakaneko (changes requested):

notbakaneko (approved):

PR #11264, opened 2024-06-12
Log payment validation errors to Sentry

notbakaneko:
These used to be reported to a now non-existent channel, so this is about getting visibility on payment processing validation errors back (because standard `ValidationException`s don't get reported to sentry).

### DIFF
--- a/app/Exceptions/InvalidSignatureException.php
+++ b/app/Exceptions/InvalidSignatureException.php
@@ -9,9 +9,9 @@ use Exception;
 
 class InvalidSignatureException extends Exception implements HasExtraExceptionData
 {
-    public function __construct(private array $extras = [])
+    public function __construct(string $message = '', private array $extras = [])
     {
-        parent::__construct();
+        parent::__construct($message);
     }
 
     // doesn't really contain anything
--- a/app/Exceptions/InvalidSignatureException.php
+++ b/app/Exceptions/InvalidSignatureException.php
@@ -14,7 +14,6 @@ class InvalidSignatureException extends Exception implements HasExtraExceptionDa
         parent::__construct($message);
     }
 
-    // doesn't really contain anything
     public function getContexts(): array
     {
         return [];
--- a/app/Exceptions/InvalidSignatureException.php
+++ b/app/Exceptions/InvalidSignatureException.php
@@ -7,7 +7,21 @@ namespace App\Exceptions;
 
 use Exception;
 
-class InvalidSignatureException extends Exception
+class InvalidSignatureException extends Exception implements HasExtraExceptionData
 {
+    public function __construct(private array $extras = [])
+    {
+        parent::__construct();
+    }
+
     // doesn't really contain anything
+    public function getContexts(): array
+    {
+        return [];
+    }
+
+    public function getExtras(): array
+    {
+        return $this->extras;
+    }
 }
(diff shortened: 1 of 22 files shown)


notbakaneko:
I've been made aware of a whoops 👀

app/Libraries/Payments/PaymentProcessor.php
```diff
@@ -341,11 +341,8 @@ protected function getOrder()
 
     /**
      * Convenience method that calls dispatchValidationFailed() and then throws the supplied exception.
-     *
-     * @param Exception $exception
-     * @return void
      */
-    protected function throwValidationFailed(Exception $exception)
+    protected function throwValidationFailed(Exception $exception): void
```
nanaya, line 345: this wrapper is kinda useless now with no actual channel? (along with all the event listeners/handlers)

app/Libraries/Payments/PaymentProcessor.php
```diff
@@ -216,7 +216,7 @@ public function cancel()
         $this->sandboxAssertion();
 
         if (!$this->validateTransaction()) {
-            $this->throwValidationFailed(new PaymentProcessorException($this->validationErrors()));
+            $this->throwValidationFailed(new PaymentProcessorException($this->getOrder(), $this->validationErrors()));
```
nanaya, line 219: not related to this pr but it looks like there should be an `assertValidTransaction` which validates transaction and throw if not valid

nanaya (changes requested):

app/Libraries/Payments/PaymentProcessor.php
notbakaneko: Not doing it in this PR, there's going to be a separate cleanup for all this; this PR is only for getting the errors reported again.

app/Exceptions/InvalidSignatureException.php
```diff
@@ -7,7 +7,21 @@
...
 
-class InvalidSignatureException extends Exception
+class InvalidSignatureException extends Exception implements HasExtraExceptionData
 {
+    public function __construct(string $message = '', private array $extras = [])
+    {
+        parent::__construct($message);
+    }
+
     // doesn't really contain anything
```
nanaya, line 17: not sure if needed

app/Exceptions/Store/FulfillmentException.php
```diff
@@ -3,10 +3,10 @@
...
 // See the LICENCE file in the repository root for full licence text.
 
-namespace App\Libraries\Fulfillments;
+declare(strict_types=1);
 
-use App\Exceptions\ValidationException;
+namespace App\Exceptions\Store;
 
-class FulfillmentException extends ValidationException
+class FulfillmentException extends OrderException
```
nanaya, line 10: was this caught in the controller before? 🤔 looks like the listeners are dispatched in sync and thus previously potentially caught by the controller?

nanaya (changes requested):

app/Exceptions/Store/FulfillmentException.php
notbakaneko: That's a good point; originally it was planned to have the fulfillers eventually moved to workers so it wasn't intended to have their exceptions caught by the controller, but we never went through with that 🤔

nanaya: I guess it should be handled correctly then? Either queue the event handlers or make such errors return the expected code (200?) so the payment processor receives the correct status.

nanaya (approved):
seems correct code wise but I can't quite manually test all the code paths, so 😬 🤞

PR #11268, opened 2024-06-13
Fix saving status alignment on account settings page

notbakaneko:
The text alignment of the saving status on the account edit isn't quite right when using a CJK font.
Was wondering wtf was going on until noticing there's actually 2 checkbox status classes...

(Ignore the part where all the indicators are forced to display :eyes:)

image


resources/css/bem/account-edit-status.less
```diff
@@ -9,18 +9,24 @@
...
         opacity: 1;
       }
     }
 
     &--saving {
-      width: 0;
+      position: absolute;
+
+      @media @desktop {
+        .account-edit-entry__checkbox-status--left & {
```
nanaya, line 26: apply modifier to the account-edit-status block instead?

nanaya (changes requested):

nanaya (approved):

PR #11269, opened 2024-06-13
Use weighted performance for country leaderboards

### LINKED ISSUE #10525: Change to Country section on Rankings page
Right now, the country page lists each country and rankes them according to the cumulative sum of pp achieved by members with the country flag. This offers no information apart from which countries have the most users.

image

My suggestion is to change the way this page works. Countries should be ranked by the weighted sum of their players' pp values, in the same way as players are ranked by the weighted sum of their scores' pp values. To elaborate, the rank 1 player in a country should contribute 100% of their pp to the country's total, the rank 2 contributing 95% etc etc down to a suitable value, possibly even just top 100.

image
(ignore central column - the value being sorted is the right-most one)

This in my opinion would bring a greater sense of community to the country ranking boards, and can lead to community efforts such as attempting to gain ranks on this board instead of only for solo goals.

Other metrics can still stay visible, and I can see that this _could_ be confusing for newer players, however it would be an immensely cool metric to have access to on the main site and gives a greater purpose to this country ranking page.

tsunyoku:
Resolves #10525.

Added the user count as a config variable just in case it's something that wants to be messed with but seems general consensus from [peppy's poll](https://x.com/ppy/status/1707702245712974039) that top 1000 is the best idea.

Unsure if the column on the page should be reworded to say something like "Weighted Performance" so that the difference is clear for those who've looked at the page in the past but it might seem weird since pp leaderboards for example just use "Performance".

Formula for weighted performance is the same as user total performance.

I don't know how often this command is ran in production but if this gets in then it might be worth running the recalculation if it's not usually ran often.


nanaya:
the formula question is still open. I think the 0.95 factor is too steep for country-sized population. It's effectively 0 after around 300th.

tsunyoku:
if we want to encapsulate the entire top 1000 then we should use something around 0.99 for the factor, for the 999th user it's ~0.004

nanaya:
maybe make it configurable for now with 0.99 as the default

app/Models/CountryStatistics.php
```diff
@@ -34,18 +34,34 @@ public static function recalculate($countryAcronym, $modeInt)
             ->where('rank_score', '>', 0)
             ->whereHas('user', function ($userQuery) {
                 return $userQuery->default();
-            })->select(DB::raw('sum(ranked_score) AS ranked_score, sum(playcount) AS playcount, count(*) AS usercount, sum(rank_score) AS rank_score'))
+            })->select(DB::raw('sum(ranked_score) AS ranked_score, sum(playcount) AS playcount, count(*) AS usercount'))
             ->first();
 
         if ($stats->ranked_score > 0) {
+            $userPerformances = UserStatistics\Model::getClass(Beatmap::modeStr($modeInt))
```
nanaya, line 41: missing default user scope

nanaya (changes requested):

config/osu.php
```diff
@@ -312,4 +312,8 @@
         'github_token' => env('CHANGELOG_GITHUB_TOKEN'),
         'update_streams' => array_map('intval', explode(' ', env('UPDATE_STREAMS', '5 1'))),
     ],
+    'rankings' => [
+        'country_performance_user_count' => intval(env('COUNTRY_PERFORMANCE_USER_COUNT', 1000)),
```
nanaya, line 316: forgot to mention before but the new envs should be added to the `.env.example` file

nanaya:
@peppy please confirm whether or not we want to go on with this

app/Models/CountryStatistics.php
```diff
@@ -34,18 +34,37 @@ public static function recalculate($countryAcronym, $modeInt)
...
+                ::select('rank_score')
+                ->where('country_acronym', $countryAcronym)
+                ->where('rank_score', '>', 0)
+                ->whereHas('user', function ($userQuery) {
+                    return $userQuery->default();
+                })
```
nanaya, line 47: might as well make a query variable first and use it for both queries. Also the anonymous function can be shortened with `fn ($userQuery) => $userQuery->default()`

nanaya (changes requested):

peppy:
@nanaya i'm on board with this change, yes.

nanaya:
merge conflicts 👀

tsunyoku:
presumedly from https://github.com/ppy/osu-web/pull/11282, i'll fix

nanaya (approved):

PR #11273, opened 2024-06-14
Remove unused score pins table columns

nanaya:
- [x] deployed #11270


notbakaneko (approved):

cl8n:
not a problem but just curious why are these written in raw sql instead of query builder?

PR #11277, opened 2024-06-17
Show legacy forum icons on topic listing, posts, and post search

### LINKED ISSUE #2814: Add legacy support for forum icons
This needs to be added so that future users can understand the context of what was going on in an older v1 thread. 
[Example](https://osu.ppy.sh/community/forums/topics/582342?start=6554643) and [old forum thread](https://osu.ppy.sh/forum/p/6554643)
If in the future someone was to look at this thread, they would have no idea who bubbled or hearted the set. It would be especially confusing to try and understand controversial threads (which are looked at more often) as it would be much harder to track the state of whether a map was bubbled, popped, or hearted.

cl8n:
resolves #2814

it's pretty ugly with most icons but I hope that can be forgiven since it's just a historical thing.

---

### DIFF
--- a/app/Models/Forum/Post.php
+++ b/app/Models/Forum/Post.php
@@ -397,6 +397,13 @@ class Post extends Model implements AfterCommit, Indexable, Traits\ReportableInt
 
     public function legacyIcon(): ?LegacyIcon
     {
+        // TODO: $this->icon_id shouldn't be null, but there are some cases
+        // where we make a new Post model and don't refresh it with the default
+        // attributes from database
+        if ($this->icon_id === null) {
+            return null;
+        }
+
         return app('forum-legacy-icons')->byId($this->icon_id);
     }
 


database/migrations/2024_06_17_203000_create_phpbb_icons.php
```diff
@@ -0,0 +1,42 @@
...
+
+        Schema::create('phpbb_icons', function (Blueprint $table) {
+            $table->mediumIncrements('icons_id');
+            $table->string('icons_url')->default('');
+            $table->tinyInteger('icons_width')->default(0);
+            $table->tinyInteger('icons_height')->default(0);
+            $table->unsignedMediumInteger('icons_order')->default(0);
+            $table->boolean('display_on_posting')->default(true);
+
+            $table->index(['display_on_posting'], 'display_on_posting');
```
cl8n, line 31: could use a double-check for accuracy bc I just copied thsi from phpbb3.0

resources/css/bem/forum-legacy-icon.less
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+.forum-legacy-icon {
+  margin: 0 5px;
```
nanaya, line 5: I think this can use gap on __icons instead

resources/views/objects/search/_forum_post.blade.php
```diff
@@ -14,6 +14,11 @@
     >
         <img class="search-forum-post__avatar-image" src="{{ $user->user_avatar }}">
     </a>
+
+    @if ($post->legacyIcon() !== null)
```
nanaya, line 18: I don't think this is a good place to put the icon. It just looks wrong

nanaya (changes requested):
also the icons are very ugly especially with dark background. I wonder if the non-essentials should just be removed

resources/views/objects/search/_forum_post.blade.php
notbakaneko: Icons should go before or after the title, or in the information lines at the bottom, imo. It should definitely not be shifting the layout of entire blocks.

Issue #10963, opened 2024-02-03
Beatmap playmode tabs show different unlabeled stats on "info" vs. "discussions" pages


cl8n:
on the info page:
![](https://github.com/ppy/osu-web/assets/9954349/729cabcb-4661-4524-8ebe-b5ad150e69ec)
on the same map's discussion page:
![](https://github.com/ppy/osu-web/assets/9954349/a08e3ad3-daf8-4ef7-b98f-a5a949612008)

if you aren't already familiar with how each page works, then the fact that these numbers mean different things is hard to pick up. it appears more like a bug on first sight (which is how this was initially reported in discord just now)

also, while not the main issue, I find it weird that the discussion page shows zeroes instead of nothing for modes that aren't present.

PR #11286, opened 2024-06-20
Add tooltip on Beatmap discussions page for per-ruleset unresolved issue count 

notbakaneko:
Not sure if tooltip should be separate or combined with the icon's own tooltip

Maybe helps with #10963 🤷


resources/js/beatmap-discussions/header.tsx
```diff
@@ -73,6 +73,7 @@ export class Header extends React.Component<Props> {
                 count: this.discussionsState.unresolvedDiscussionCounts.byMode[mode],
                 disabled: (this.discussionsState.groupedBeatmaps.get(mode)?.length ?? 0) === 0,
                 mode,
+                tooltip: transChoice('beatmaps.discussions.unresolved_count', this.discussionsState.unresolvedDiscussionCounts.byMode[mode] ?? 0),
```
nanaya, line 76: countTooltip? 🤔

nanaya (changes requested):

nanaya (approved):

PR #9872, opened 2023-02-17
Render chat markdown with react-markdown

notbakaneko:
closes #9870

`content_html` can be removed after giving clients some time to refresh 🤔

PR #11287, opened 2024-06-21
Remove content_html from Message

notbakaneko:
Deprecated since #9872 and ended up being forgotten 👀 
@ppy/team-client  If anyone can confirm it's not being used in the client (I can't find any reference to `content_html`)


bdach:
> If anyone can confirm it's not being used in the client (I can't find any reference to `content_html`)

As far as lazer goes it's not. The requisite counterpart is [this](https://github.com/ppy/osu/blob/8bbbedaec3a1af9a255a32e3f186cfebd25d6783/osu.Game/Online/Chat/Message.cs), and we only have plaintext `content` in there. We don't have any means to render html in client anyways (actual markdown at best).

peppy (approved):

PR #11294, opened 2024-06-24
Set default discussion filter according to ranked state of Beatmap

### LINKED ISSUE #1392: Default filter modes should be set better
- For non-ranked maps (qualified or below) "pending" should be the default filter.
- For ranked maps "praise" should be the default filter.

notbakaneko:
closes #1392

Also changes it so the url updates when jumping.


nanaya (approved):

Hiviexd:
this change has resulted in backlash in many mapping spaces, specifically defaulting to pending posts in unranked beatmaps. it'd be greatly appreciated to partially revert it so the filter is only set on ranked maps.

Issue #10974, opened 2024-02-06
``api/v2/matches/{match}`` missing params in documentation


ArtemOsuskyi:
According to #9589, for ``api/v2/matches/{match}`` there should be `before` and `after` params which seem to be missing

![image](https://github.com/ppy/osu-web/assets/96246908/70780122-549f-4e9c-af42-7b3cc1db6de9)

cl8n:
well the whole endpoint is undocumented (look at what section it's in). I'll keep this issue open because I've seen a lot of questions about this endpoint in particular, probably because it's supposed to be a direct alternative to something api v1 has.

PR #11298, opened 2024-06-25
Add documentation to matches API endpoints

ssz7-ch2:
Resolves https://github.com/ppy/osu-web/issues/10974

Not sure if matches should go in its own group or under Multiplayer, but I've placed it in its own group for now.


resources/views/docs/_structures/match_game.md
```diff
@@ -0,0 +1,15 @@
+## MatchGame
+
+Field        | Type                     | Description
+------------ | ------------------------ | -----------
+id           | integer                  | |
+beatmap      | [Beatmap](#beatmap)      | Includes `beatmapset`.
+beatmap_id   | integer                  | |
+start_time   | [Timestamp](#timestamp)  | Uses `Z` instead of `+00:00` for timezone.
```
notbakaneko, line 8: I don't think this needs to be mentioned, ISO 8601 defines both and parsing ISO 8601 means you have to be able to accept either form.

notbakaneko (changes requested):

notbakaneko (approved):

PR #11300, opened 2024-06-26
Allow ctrl+enter to resolve beatmap discussion when replying

### LINKED ISSUE #2875: Pressing enter when replying to mods doesn't resolve
When replying to a mod, pressing enter to post the reply will not resolve the post in addition to putting the reply yet. Issue is more often you'd be closing an issue after replying to it than leaving it open, so it would be more practical to have pressing enter also resolve the issue. This would help make replying to mods quicker and more efficient.

notbakaneko:
closes #2875

### DIFF
--- a/resources/lang/en/beatmaps.php
+++ b/resources/lang/en/beatmaps.php
@@ -25,7 +25,7 @@ return [
         'message_placeholder_silenced' => "Can't post discussion while silenced.",
         'message_type_select' => 'Select Comment Type',
         'reply_notice' => 'Press enter to reply.',
-        'reply_resolve_notice' => 'Press enter to reply. Press crtl + enter to reply and resolve.',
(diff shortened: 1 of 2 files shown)


resources/lang/en/beatmaps.php
```diff
@@ -25,6 +25,7 @@
         'message_placeholder_silenced' => "Can't post discussion while silenced.",
         'message_type_select' => 'Select Comment Type',
         'reply_notice' => 'Press enter to reply.',
+        'reply_resolve_notice' => 'Press enter to reply. Press crtl + enter to reply and resolve.',
```
nanaya, line 28: ctrl. Also there shouldn't be spaces around the `+` sign

nanaya (changes requested):

nanaya (approved):

PR #11301, opened 2024-06-26
Remove average performance column on country rankings page

nanaya:
Broken after the weighted performance update. It's either this or add a new column 🤔


peppy (approved):
let's remove for now

PR #11304, opened 2024-06-27
Add backend support for customising profile page hue

nanaya:
The frontend needs a bit more time to cook 🍳

```javascript
$.ajax('/home/account', { method: 'PUT', data: { user: { user_style: 80 } } })
```

The variables are moved to body to allow turbolinks-compatible preview as otherwise it'll need to modify `<html>` styling which persists on navigation.


notbakaneko (approved):
This part seems fine

PR #11305, opened 2024-06-27
Remove user rank cache lookup

nanaya:
Use elasticsearch instead. I don't think it's used at all after the solo score migration though. I can't find it in datadog either (the osu-web one, not legacy web)

BeatmapModeStats table to be removed later™


notbakaneko (approved):
seeeems fine 🤔

PR #11312, opened 2024-07-03
Ignore more Paypal IPN transaction types

notbakaneko:
Or maybe should ignore everything without `invoice` field set 🤔 (which I recall wasn't done to make sure that callbacks for outgoing transactions that had them missing would explode)

### DIFF
--- a/app/Libraries/Payments/PaypalPaymentProcessor.php
+++ b/app/Libraries/Payments/PaypalPaymentProcessor.php
@@ -70,7 +70,9 @@ class PaypalPaymentProcessor extends PaymentProcessor
         static $rejectedStatuses = ['Declined', 'Denied', 'Expired', 'Failed', 'Voided'];
 
         $status = $this->getNotificationTypeRaw();
-        if (in_array($status, $paymentStatuses, true)) {
+        if ($this->shouldIgnore($status)) {
+            return NotificationType::IGNORED;
+        } elseif (in_array($status, $paymentStatuses, true)) {
             return NotificationType::PAYMENT;
         } elseif (in_array($status, $refundStatuses, true)) {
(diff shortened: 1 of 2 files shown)


app/Libraries/Payments/PaypalPaymentProcessor.php
```diff
@@ -70,16 +70,16 @@ public function getNotificationType(): string
         static $rejectedStatuses = ['Declined', 'Denied', 'Expired', 'Failed', 'Voided'];
 
         $status = $this->getNotificationTypeRaw();
-        if (in_array($status, $paymentStatuses, true)) {
+        if ($this->shouldIgnore($status)) {
```
nanaya, line 73: probably can use a note on why this needs to be first (or maybe just inline the function)

app/Libraries/Payments/PaypalPaymentProcessor.php
```diff
@@ -70,16 +70,16 @@ public function getNotificationType(): string
...
+            return NotificationType::IGNORED;
+        } elseif (in_array($status, $paymentStatuses, true)) {
```
nanaya, line 75: semi unrelated but this looks like a job for switch/match.

nanaya (changes requested):

app/Libraries/Payments/PaypalPaymentProcessor.php
notbakaneko: I don't think it's worth switching it right now

nanaya (approved):
the comment doesn't really make sense without knowing how `getNotificationTypeRaw` works but oh well ಠ_ಠ

(the abstraction itself doesn't seem to be useful anymore either)

PR #11314, opened 2024-07-03
Frontend for profile page hue customisation

nanaya:
~Drafting due to potential field name change.~


nanaya:
Seems okay 🤷 The range doesn't fill the entire width because the handle needs some extra space and I couldn't figure out how to fake the ends correctly.

notbakaneko (changes requested):
conflicts 💥

PR #11317, opened 2024-07-04
Convert HSL to HEX in `theme-color` meta tag

Hiviexd:
this makes it compatible with more stuff in general, most notably Discord embed colors.


app/helpers.php
```diff
@@ -414,6 +414,73 @@ function get_valid_locale($requestedLocale)
...
+        $sv;
+        $sextant;
+        $fract;
+        $vsf;
+        $mid1;
+        $mid2;
+
+        $m = $l + $l - $v;
+        $sv = ($v - $m) / $v;
+        $h *= 6.0;
```
nanaya, line 436: the input hue is in different unit

app/helpers.php
```diff
@@ -414,6 +414,73 @@ function get_valid_locale($requestedLocale)
     }
 }
 
+function hsl_to_hex($h, $s, $l)
+{
+    $r = $l;
+    $g = $l;
+    $b = $l;
+
+    $v = $l <= 0.5 ? $l * (1.0 + $s) : $l + $s - $l * $s;
```
nanaya, line 423: the input saturation and lightness are in different unit. the one used here makes more sense though so just need to fix the value being passed

app/helpers.php
```diff
@@ -414,6 +414,73 @@ function get_valid_locale($requestedLocale)
...
+        $m = $l + $l - $v;
+        $sv = ($v - $m) / $v;
+        $h *= 6.0;
+        $sextant = floor($h);
+        $fract = $h - $sextant;
+        $vsf = $v * $sv * $fract;
+        $mid1 = $m + $vsf;
+        $mid2 = $v - $vsf;
+
+        switch ($sextant) {
```
nanaya, line 443: can be made shorter with match and array destructuring

app/helpers.php
```diff
@@ -414,6 +414,73 @@ function get_valid_locale($requestedLocale)
     }
 }
 
+function hsl_to_hex($h, $s, $l)
+{
+    $r = $l;
+    $g = $l;
+    $b = $l;
```
nanaya, line 421: these can be delayed until the end (or just use null coalescing)

nanaya (changes requested):

app/helpers.php
```diff
@@ -414,6 +414,73 @@ function get_valid_locale($requestedLocale)
...
+    $v = $l <= 0.5 ? $l * (1.0 + $s) : $l + $s - $l * $s;
+
+    if ($v > 0) {
+        $m;
+        $sv;
+        $sextant;
+        $fract;
+        $vsf;
+        $mid1;
+        $mid2;
```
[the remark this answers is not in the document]
Hiviexd: stackoverflow trolled me...

Hiviexd:
this should properly work now, tested all the hues we use for `$currentHue` and it outputs correct hex for all of them.

resources/views/layout/metadata.blade.php
```diff
@@ -16,7 +16,7 @@
 <link rel="manifest" href="{{ $appUrl }}/site.webmanifest">
 <link rel="mask-icon" href="{{ $appUrl }}/images/favicon/safari-pinned-tab.svg" color="#e2609a">
 <meta name="msapplication-TileColor" content="#603cba">
-<meta name="theme-color" content="hsl({{ $currentHue }}, 10%, 40%)"> {{-- @osu-colour-b1 --}}
+<meta name="theme-color" content="{{ hsl_to_hex($currentHue, 40, 50) }}"> {{-- @osu-colour-b1 --}}
```
nanaya, line 19: that's not `b1`?

app/helpers.php
```diff
@@ -414,6 +414,31 @@ function get_valid_locale($requestedLocale)
     }
 }
 
+function hsl_to_hex($h, $s, $l)
+{
+    $s /= 100;
+    $l /= 100;
```
nanaya, line 420: as mentioned before, passing percentage value as is without the unit (`100%` as `100` instead of `1`) is weird. It's not how it works in the css either

nanaya (changes requested):

resources/views/layout/metadata.blade.php
nanaya: I'm not sure about that. It's a bit light and chrome actually set the text/icon colour to black for wiki section.

Hiviexd: hmm, i'll just revert to `b1` for now

nanaya (approved):

PR #11319, opened 2024-07-08
Move more colour variables to body

nanaya:
Whoops 👀


notbakaneko (approved):

LunexSan:
Is it possible that this change broke https://osu.ppy.sh/community/forums/topics/802445?n=1 ?

cl8n:
@LunexSan  the download for that seems broken so I can't check for sure but yes it's likely that this PR or some of the preceding color-related PRs broke it. you'll have to see if there are any alternatives to that extension or update it yourself

PR #11328, opened 2024-07-11
Add missing strict mode to daily challenge command

nanaya:
Not sure why but most of the commands are missing strict mode ಠ_ಠ


bdach:
Might have been a copy & paste moment. Or a "phpstorm generated this for me" moment.

notbakaneko (approved):

PR #11339, opened 2024-07-17
Remove total score without mods field if 0

nanaya:
The client explodes when the value is null, and then there's no point returning 0 total score 🤷


bdach (approved):
have not tested but looks logical

notbakaneko (approved):

PR #11342, opened 2024-07-18
Add touch support for jquery ui

### LINKED ISSUE #11334: Profile color hue slider isn't draggable on mobile
Just the title

Doesn't work either on Arc (WebKit) on my iPhone or Safari on my iPad. I don't have an android phone on me so I can't test whether it doesn't work on other browser engines or if this is just a webkit skill issue

https://github.com/user-attachments/assets/23c20dad-5caf-40fa-b1e0-cf47839f1800

nanaya:
Unfortunately this has side effect of breaking profile page's section tabs on own profile page as the sortable gesture conflicts with vertical scrolling... so I just disable the specific gesture on mobile.

Resolves #11334.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "jets": "^0.15.0",
     "jquery": "^3.5.0",
     "jquery-ui": "^1.13.2",
+    "jquery-ui-touch-punch": "^0.2.3",
     "jquery-ujs": "^1.2.2",
     "jquery.scrollto": "^2.1.2",
     "js-cookie": "^2.1.2",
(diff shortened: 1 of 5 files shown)


notbakaneko (commented):
Not sure if removing sortable tabs is the solution when it's still possible to have mobile layout on a desktop just from the window size.

And ios seems to prioritize it's link drag and drop thing so the sortable always doesn't work on ios, anyway 🤔

nanaya:
with the handle and overall page layout being different I think it's fine to also disable the thing on that layout

notbakaneko (approved):
🤷 

(Also I could have sworn it used to work on ios before.. a long time ago 🤔  )

PR #11344, opened 2024-07-19
Fix passing `passed` flag for Score factory

nanaya:
Spent a few minutes figuring out why things didn't work as expected 🤨


notbakaneko (approved):
😅

PR #11354, opened 2024-07-24
Include user score position when indexing playlist item scores

bdach:
Needed for daily challenge (without this client can't show position numbers for the user score on the leaderboard).

A second review for performance implications is probably best-advised, this monkey just fix no number showing with hammer. [I tried to ask once](https://discord.com/channels/188630481301012481/188630616286167041/1258386604754604083) but got no reply.

### DIFF
--- a/app/Http/Controllers/Multiplayer/Rooms/Playlist/ScoresController.php
+++ b/app/Http/Controllers/Multiplayer/Rooms/Playlist/ScoresController.php
@@ -82,7 +82,14 @@ class ScoresController extends BaseController
             )->first();
 
             if ($userHighScoreLink !== null) {
-                $userScoreJson = json_item($userHighScoreLink, $transformer, ScoreTransformer::MULTIPLAYER_BASE_INCLUDES);
+                $userScoreJson = json_item(
+                    $userHighScoreLink,
+                    $transformer,
+                    [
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Multiplayer/Rooms/Playlist/ScoresController.php
```diff
@@ -66,7 +66,10 @@ public function index($roomId, $playlistId)
         $scoresJson = json_collection(
             $highScores->pluck('scoreLink'),
             $transformer,
-            ScoreTransformer::MULTIPLAYER_BASE_INCLUDES
+            [
+                ...ScoreTransformer::MULTIPLAYER_BASE_INCLUDES,
+                'position',
```
bdach, line 71: I can _maybe_ skip this one and hardcode a 1..50 position client-side if it helps.

nanaya: yeah this will cause as many additional queries as there are scores

bdach: Have force-pushed this out.

nanaya (approved):

PR #11355, opened 2024-07-24
Add prettier formatter for css

nanaya:
Seems good enough 🤷 

- [x] depends on #11353


notbakaneko (changes requested):

notbakaneko (changes requested):
there's 2 spaces here 🤔  https://github.com/ppy/osu-web/blob/4bcba5445f7752bdda7e043a812c513fbf7ce446/resources/css/utilities.less#L70

`yarn pretty`: pass
`yarn prettier --check resources/css/**/*.less`: fail
?_?

notbakaneko (approved):

PR #11358, opened 2024-07-25
Change daily challenge week start to thursday for now

peppy:


ReinIsNOTaDev:
what would be the transition to a proper week start (if it happens) and how would it happen?

cl8n:
this will feel weird to be out-of-sync with the plan of ramping up difficulty from Monday to Sunday, unless that's changing too?

nanaya:
uh yeah, the news post is showing "next week" is starting from monday...

nanaya (approved):

PR #11360, opened 2024-07-25
Fix off by one percentile score

nanaya:
I think...? 🤔


notbakaneko (changes requested):
If there are equal scores like `[20, 20, 20, 20, 20, 70, 70, 70, 70, 70]` which percentile would they fall into?

nanaya:
all the 70s would go to 10p and 20s would be none.

notbakaneko:
Maybe include a test for number of expected entries in each percentile?

nanaya:
the match cases already imply the count though

notbakaneko (approved):

PR #11363, opened 2024-07-26
Update daily challenge streak count on play instead of at the end of the day

nanaya:
More useful in term of visibility but the percentile stats being updated later may look a bit weird...?


cl8n:
maybe a note like "Placement counts will update when the challenge ends in x hours" in the popover when you participated today would help w the feeling weird problem

notbakaneko (changes requested):
I noticed if you `caculate($dateFromWeekBefore)` on already existing stats, it adds to them again and `last_weekly_streak` gets set to the week before 🤔

nanaya:
Calculating backward isn't really supported and things will break, not just the weekly streak.

notbakaneko (approved):

PR #11377, opened 2024-07-30
Support multiple mappers on beatmap difficulties

### LINKED ISSUE #7602: Allow to choose several mappers as difficulty owners
This is a common practice for a difficulty to be mapped by two or more mappers as a collaboration, however the new system lets users to choose one guest mapper per difficulty only. Perfectly, it would be the best solution if we could mark each part of the map to show who mapped it through web interface, but considering the complexity of this thing I guess it would be a better idea to implement adding of multi users per difficulty.

notbakaneko:
closes #7602

Also refactors out the username input/search thing a separate component

- [x] fix some display consistency things.
- [x] missing users should still be in response.
- [x] preserve existing guest mapper even if missing or restricted when updating.
- [x] user profile page 🤔 
- [x] #11451
- [x] search and indexing


notbakaneko:
I'm thinking all the `*Owner` should be renamed `*Mapper` to be consistent, but then there's the existing beatmap events which uses `*owner`...

nanaya:
can try renaming the events...

nanaya (changes requested):
- decide on either mapper or owner name; not both
- split the user lookup change to its own pr (and combine with `checkUsernameExists` while at it)

nanaya:
can use a merge conflict fix

notbakaneko:
Not sure renaming to mappers is a great idea now that I've gotten to the event itself, but using `mapper` for the display text at least seems to make more sense than `owner`...

notbakaneko:
Should be no new introduction of `mapper` instead of `owner` now, I think 🤔

nanaya (changes requested):
part 1?

resources/css/bem/beatmap-owner-editor.less
```diff
@@ -37,32 +31,23 @@
...
 
-  &__input {
-    .default-border-radius();
-    .reset-input();
-    background-color: hsl(var(--hsl-b6));
-    color: inherit;
-    padding: 1px 3px;
-    width: auto;
-    border: 2px solid transparent;
+  &__owners {
```
nanaya, line 40: can use some wrapping and more gap between avatar and previous user

nanaya:
still has conflict btw 👀 

(also the elasticsearch thing)

notbakaneko:
I'll get to that 👀

app/Models/Beatmap.php
```diff
@@ -376,6 +374,30 @@ private function getDiffSize()
         return $value;
     }
 
+    private function getOwners(): Collection
```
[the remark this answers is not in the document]
notbakaneko: Elasticsearch already stores multiple values in a field

nanaya (changes requested):

resources/css/bem/input-container.less
```diff
@@ -50,6 +50,11 @@
     }
   }
 
+  &--beatmap-owner-editor {
```
nanaya, line 53: block interleaving aside, this seems wrong...? barely any way to differentiate editing and non-editing state. and clicking on the highlighted parts focuses somewhere unrelated if there's an edit state somewhere. or if there's nothing being edited, it just does nothing which is also weird despite having a hover outline.

I think the "beatmap owner" block can still use a background here... the &times; button can still look like belong to either to its left or right.

and looking again, maybe just remove the non-editing state and have them all editable right away.

peppy:
prod on this one, let's try and get it live >_<

app/Http/Controllers/BeatmapsController.php
```diff
@@ -25,7 +24,7 @@
 class BeatmapsController extends Controller
 {
     const DEFAULT_API_INCLUDES = ['beatmapset.ratings', 'failtimes', 'max_combo'];
-    const DEFAULT_SCORE_INCLUDES = ['user', 'user.country', 'user.cover'];
+    const DEFAULT_SCORE_INCLUDES = ['user', 'user.country', 'user.cover', 'beatmap.owners'];
```
nanaya, line 27: I don't think including same beatmap multiple times when fetching leaderboard would be useful?

the commit log says for userScore but then userScore also included this manually...?

nanaya (changes requested):

app/Http/Controllers/BeatmapsetsController.php
```diff
@@ -402,8 +403,10 @@ private function showJson($beatmapset)
             'beatmaps',
             'beatmaps.failtimes',
             'beatmaps.max_combo',
+            'beatmaps.owners',
             'converts',
             'converts.failtimes',
+            'converts.owners',
```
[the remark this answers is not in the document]
nanaya: as a bonus(?), if just need the user ids, it can potentially be selected together with the beatmap itself with something like in https://github.com/ppy/osu-web/blob/1fbc73baa8e8be6759b1cc4f4ab509d8ae53a165/app/Models/Multiplayer/Room.php#L296-L321

resources/css/bem/beatmap-owner-editor-owners.less
```diff
@@ -0,0 +1,20 @@
...
+  padding: var(--input-padding);
+
+  border-radius: @border-radius-large;
+
+  &--editing {
+    background: var(--input-bg);
+  }
+
+  .input-container--error & {
+    box-shadow: inset 0 0 0 2px var(--input-border-error-colour);
```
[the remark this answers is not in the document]
notbakaneko: I think the error highlight shouldn't be overridden by the focus highlight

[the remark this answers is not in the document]
nanaya: then it should be changed globally instead? 🤔

app/Http/Controllers/BeatmapsetsController.php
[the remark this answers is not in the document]
notbakaneko: `owners` is included on `converts` so the UI side doesn't have to go digging the for the non-convert to check for mappers.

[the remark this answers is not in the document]
notbakaneko: `owners` is also contains the minimum to display on pages that don't have `related_users` like the score page instead of a separate include for just the ids, otherwise, the score page should sideload the users instead.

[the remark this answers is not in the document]
nanaya: > `owners` is included on `converts` so the UI side doesn't have to go digging the for the non-convert to check for mappers.

just include the ids then?

> `owners` is also contains the minimum to display on pages that don't have `related_users` like the score page instead of a separate include for just the ids, otherwise, the score page should sideload the users instead.

include it only on that specific page then? or add the related users.

seems like a lot of data being returned for everything else just to deal with single page with its single beatmap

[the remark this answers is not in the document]
notbakaneko: `owners` is just the id and username...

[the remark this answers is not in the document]
notbakaneko: I'm going to remove the duplicated data from `converts` but it may or not be part of this PR since it needs some extra type and client-side beatmap lookup changes

app/Models/Beatmapset.php
```diff
@@ -1278,18 +1280,30 @@ public function status()
...
+        ]);
+
+        foreach ($this->allBeatmaps as $beatmap) {
+            $beatmap->setRelation('beatmapset', $this);
+        }
+
+        foreach ($this->beatmapDiscussions as $discussion) {
+            // set relations for priv checks.
+            $discussion->setRelation('beatmapset', $this);
+            $beatmap = $this->allBeatmaps->find($discussion->beatmap_id);
```
nanaya, line 1298: keying the beatmaps seems like a good idea

resources/js/beatmap-discussions/beatmap-owner-editor.tsx
```diff
@@ -102,67 +93,55 @@ export default class BeatmapOwnerEditor extends React.Component<Props> {
     );
   }
 
+  @action
+  private readonly handleBeforeVisit = (event: Event) => {
+    if (this.editing && !confirm(trans('common.confirmation_unsaved'))) {
```
nanaya, line 98: can use this when closing the dialogue too

resources/css/bem/beatmap-owner-editor-owners.less
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+.beatmap-owner-editor-owners {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  gap: 2px;
+  padding: var(--input-padding);
+  box-shadow: inset 0 0 0 2px var(--input-border-colour);
```
nanaya, line 10: should be only for `--editing`?

resources/js/beatmap-discussions/beatmap-owner-editor.tsx
```diff
@@ -206,70 +185,61 @@ export default class BeatmapOwnerEditor extends React.Component<Props> {
...
+      <InputContainer
+        for='beatmap-owner-editor-username-input'
+        hasError={!this.canSave}
+        modifiers='beatmap-owner-editor'
+        showError={this.showError}
+      >
+        <div className={classWithModifiers('beatmap-owner-editor-owners', { editing: this.editing })}>
+          {this.editing ? (
+            <UsernameInput
+              id='beatmap-owner-editor-username-input'
```
nanaya, line 203: should have unique id, otherwise clicking on any entry label will always focus the first box

app/Models/Beatmap.php
```diff
@@ -274,6 +284,34 @@ public function getAttribute($key)
...
+
+        // TODO: remove when everything writes to beatmap_owners.
+        if (!$owners->contains(fn ($beatmapOwner) => $beatmapOwner->user_id === $this->user_id) && $this->user !== null) {
+            $owners->prepend($this->user);
+        }
+
+        return $owners;
+    }
+
+    public function isOwner(User $user): bool
```
nanaya, line 304: can use similar update for js side (being able to post note and showing "guest" badge on posts)

nanaya (changes requested):

resources/js/beatmap-discussions/beatmap-owner-editor.tsx
nanaya: apparently if you render the input box first and move it to the end using flex's `order: 1`, the id isn't needed

resources/js/utils/beatmap-helper.ts
```diff
@@ -99,6 +101,14 @@ export function group<T extends BeatmapJson>(beatmaps?: T[] | null, includeEmpty
   return ret;
 }
 
+export function hasGuestOwners(beatmap: WithBeatmapOwners<BeatmapJson>, beatmapset: BeatmapsetJson) {
+  return beatmap.owners.some((owner) => owner.id !== beatmapset.user_id);
+}
+
+export function isOwner(userId: number, beatmap: WithBeatmapOwners<BeatmapJson>) {
```
nanaya, line 108: there's at least another one missing usage of this in `nominator.tsx` (nominate button permission)

nanaya (changes requested):
there'

app/Models/Beatmap.php
```diff
@@ -279,6 +289,34 @@ public function getAttribute($key)
...
+     * @return Collection<User>
+     */
+    public function getOwners(): Collection
+    {
+        $owners = $this->beatmapOwners->loadMissing('user')->map(
+            fn ($beatmapOwner) => $beatmapOwner->user ?? new DeletedUser(['user_id' => $beatmapOwner->user_id])
+        );
+
+        // TODO: remove when everything writes to beatmap_owners.
+        if (!$owners->contains(fn ($beatmapOwner) => $beatmapOwner->user_id === $this->user_id) && $this->user !== null) {
```
nanaya, line 302: related to the comment above, can't this be simplified to just `if ($owners->empty() && $this->user !== null) {`?

(there's also a potential weird bug/case already existing here)

notbakaneko: Is it a simplification if more states have to be considered instead of user is in or not? 🤨

nanaya: what more state? it's just whether there's anything in it or not

notbakaneko: It's also making another unnecessary assumption instead of just `user in or not`

nanaya: it skips looping through the array and so the check is way simpler

nanaya: also I think it's a fine assumption - there's either anything in the owners table or there's none

notbakaneko: It's also very pointless bike shedding

nanaya: more like optimization instead of bike shedding to me?

resources/js/beatmap-discussions/beatmap-owner-editor.tsx
```diff
@@ -1,58 +1,55 @@
...
   @observable private updatingOwner = false;
-  private userLookupTimeout?: number;
-  private readonly xhr: Partial<XhrCollection> = {};
+  @observable private validUsers = new Map<number, UserJson>();
 
-  @computed
-  private get inputUser() {
-    return this.props.userByName.get(normaliseUsername(this.inputUsername));
+  private get canSave() {
+    return this.validUsers.size > 0 && normaliseUsername(this.inputUsername).length === 0;
```
nanaya, line 43: normaliseUsername is now just a rather verbose way to do a `.trim`

app/Models/Beatmap.php
```diff
@@ -279,6 +289,34 @@ public function getAttribute($key)
...
+    /**
+     * @return Collection<User>
+     */
+    public function getOwners(): Collection
+    {
+        $owners = $this->beatmapOwners->loadMissing('user')->map(
+            fn ($beatmapOwner) => $beatmapOwner->user ?? new DeletedUser(['user_id' => $beatmapOwner->user_id])
+        );
+
+        // TODO: remove when everything writes to beatmap_owners.
```
nanaya, line 301: more like should eventually decide whether an empty collection is acceptable or not (in case of broken owners or just simplifying simple cases)

resources/js/beatmap-discussions/beatmap-owner.tsx
```diff
@@ -0,0 +1,46 @@
...
+import * as React from 'react';
+import { classWithModifiers } from 'utils/css';
+
+interface Props {
+  editing: boolean;
+  onRemoveUser?: (user: UserJson) => void;
+  user: UserJson;
+}
+
+function createRemoveOwnerHandler(user: UserJson, onRemoveClick?: NonNullable<Props['onRemoveUser']>) {
```
nanaya, line 16: there's no need for this lint workaround anymore now that the component is split out - this can just be part of the component

resources/js/beatmap-discussions/beatmap-owner-editor.tsx
```diff
@@ -206,70 +175,59 @@ export default class BeatmapOwnerEditor extends React.Component<Props> {
...
+              initialUsers={this.owners}
+              // initialValue not set for owner editor as value is reset when cancelled.
+              modifiers='beatmap-owner-editor'
+              onEnterPressed={this.handleSaveClick}
+              onValidUsersChanged={this.handleValidUsersChanged}
+              onValueChanged={this.handleUsernameInputValueChanged}
+              renderUser={this.renderOwner}
+            />
+          ) : (
+            this.owners.map((owner) => <BeatmapOwner key={owner.id} editing={this.editing} user={owner} />)
```
nanaya, line 201: this can just call `renderUser`

database/factories/BeatmapFactory.php
```diff
@@ -64,6 +66,15 @@ public function inactive(): static
         ]);
     }
 
+    public function owner(User $user): static
```
nanaya, line 69: this should (will?) probably be `owners` instead

app/Libraries/Beatmapset/ChangeBeatmapOwners.php
```diff
@@ -0,0 +1,80 @@
...
+use App\Models\User;
+use Ds\Set;
+
+class ChangeBeatmapOwners
+{
+    private Set $userIds;
+
+    public function __construct(private Beatmap $beatmap, array $newUserIds, private User $source)
+    {
+        priv_check_user($source, 'BeatmapUpdateOwner', $beatmap->beatmapset)->ensureCan();
```
nanaya, line 24: seems like this should be part of controller (as it previously was) as indicated by the need to create additional user in test

nanaya (approved):
should mostly work... additional fixes can be done later as necessary

PR #11405, opened 2024-08-06
Use github existing chromedriver

nanaya:
I think I still some fails due to chromedriver somehow exiting midway...


notbakaneko (approved):
seems to run on my m1 mac (where chromedriver always breaks)

PR #11406, opened 2024-08-06
Add better default value to all user stats timestamp columns

nanaya:
This removes the need to pretend default separately when updating things.

(it'll also be used in verification function)


notbakaneko (approved):

peppy:
Test failure looks relevant..

nanaya:
it looks like a different bug

PR #11407, opened 2024-08-06
Split percentile calculation to its own function

nanaya:
This shortens `calculate` function and will be reused by fixup function.

### DIFF
--- a/app/Models/DailyChallengeUserStats.php
+++ b/app/Models/DailyChallengeUserStats.php
@@ -56,7 +56,6 @@ class DailyChallengeUserStats extends Model
         $highScoresByUserId = $playlist
             ->highScores()
             ->where('total_score', '>', 0)
-            ->orderBy('total_score', 'DESC')
             ->get()
             ->keyBy('user_id');
         $statsByUserId = static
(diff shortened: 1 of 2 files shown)


app/Models/DailyChallengeUserStats.php
```diff
@@ -47,27 +48,20 @@ public static function calculate(CarbonImmutable $date): void
...
+            ->get()
+            ->keyBy('user_id');
         $statsByUserId = static
             ::where('last_weekly_streak', '>=', $previousWeek->subDays(1))
-            ->orWhereIn('user_id', array_keys($highScoresByUserId))
+            ->orWhereIn('user_id', $highScoresByUserId->keys())
             ->get()
             ->keyBy('user_id');
-        $userIds = new Set([...$statsByUserId->keys(), ...array_keys($highScoresByUserId)]);
+        $percentile = $playlist->scorePercentile();
```
notbakaneko, line 62: this means it's doing the `$playlist->highScores()->where('total_score', '>', 0)->orderBy` query twice, right?

notbakaneko (changes requested):

app/Models/DailyChallengeUserStats.php
nanaya: yes. it's a once a day thing and indexed columns and the other only selects one column, so 🤷

notbakaneko (approved):

PR #11412, opened 2024-08-07
Set daily challenge room start time to be at the start of the day

nanaya:
This fixes test when it runs exactly within 30 minutes before midnight (23:30 - 00:00).

I don't know if the behaviour is acceptable for live so it'll need maybe @bdach's confirmation (or anyone from client).

Alternatively I can just mock the time used during testing.

### DIFF
--- a/app/Console/Commands/DailyChallengeCreateNext.php
+++ b/app/Console/Commands/DailyChallengeCreateNext.php
@@ -53,7 +53,10 @@ class DailyChallengeCreateNext extends Command
                         'allowed_mods' => $nextQueueItem->allowed_mods,
                         'required_mods' => $nextQueueItem->required_mods,
                     ]],
-                ]
+                ],
+                [
+                    'starts_at' => $startTime,
+                ],
             );
             $room->update(['category' => 'daily_challenge']);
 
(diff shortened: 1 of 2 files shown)


bdach:
Spectator server queries `daily_challenge` rooms every minute to broadcast the active room to clients. The query is [here](https://github.com/ppy/osu-server-spectator/blob/30d269ffc2e4cf5d86869a129cb8263f1716097b/osu.Server.Spectator/Database/DatabaseAccess.cs#L397-L402). The thing to avoid there is having multiple rooms matching this query at the same time (it won't kill spectator server, but the server _will_ start yelling about multiple rooms active simultaneously to sentry).

In principle 00:00:00+00:00 should be a fine start date with that respect, but I do wonder if we don't want to make that start date 00:05:00+00:00 to match the cron time:

https://github.com/ppy/osu-web/blob/80d90de322358013db1328f3bcb5c4fa364034ba/app/Console/Kernel.php#L144-L146

nanaya:
Sounds good. I've adjusted it.

bdach (approved):
No objections here

peppy (approved):

PR #11414, opened 2024-08-07
More sophisticated wait for github action chromedriver

nanaya:
but why ಠ_ಠ


notbakaneko (approved):
just wait for the wind to blow and break it again ಠ_ಠ

PR #11415, opened 2024-08-07
Add command to fix daily challenge stats

nanaya:
```
artisan daily-challenge:user-stats-recalculate --all

# or individual user(s)
artisan daily-challenge:user-stats-recalculate 0 2 4 6 9
```

The query isn't exactly optimal but I think it should be fine for an adhoc operation...? 🤔

### DIFF
--- a/app/Console/Commands/DailyChallengeUserStatsRecalculate.php
+++ b/app/Console/Commands/DailyChallengeUserStatsRecalculate.php
@@ -28,7 +28,7 @@ class DailyChallengeUserStatsRecalculate extends Command
         }
 
         if (count($userIds) === 0 && !$isAll) {
-            $this->error("either user ids or all option is required");
+            $this->error('either user ids or all option is required');
 
             return 1;
         }
--- a/app/Console/Commands/DailyChallengeUserStatsRecalculate.php
+++ b/app/Console/Commands/DailyChallengeUserStatsRecalculate.php
@@ -27,6 +27,12 @@ class DailyChallengeUserStatsRecalculate extends Command
             return 1;
         }
 
+        if (count($userIds) === 0 && !$isAll) {
+            $this->error("either user ids or all option is required");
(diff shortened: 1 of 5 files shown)


app/Console/Commands/DailyChallengeUserStatsRecalculate.php
```diff
@@ -0,0 +1,71 @@
...
+    protected $signature = 'daily-challenge:user-stats-recalculate {--all} {id?*}';
+
+    protected $description = 'Recalculate user daily challenge stats';
+
+    public function handle(): int
+    {
+        $isAll = $this->option('all');
+        $userIds = $this->argument('id');
+
+        if (count($userIds) > 0 && $isAll) {
```
notbakaneko, line 24: Also something to differentiate `--all` with no changes and no options passed? They look the same right now.

app/Models/DailyChallengeUserStats.php
```diff
@@ -95,6 +95,30 @@ public function user(): BelongsTo
...
     }
 
+    public function fix(): void
+    {
+        $highScores = PlaylistItemUserHighScore
+            ::where('user_id', $this->user_id)
+            ->whereRelation('playlistItem.room', 'category', 'daily_challenge')
+            ->with('playlistItem.room')
+            ->orderBy('created_at')
+            ->get();
```
notbakaneko, line 105: This seems to imply only only days with plays are considered, so it's possible for the last play to be several days previously and it'll still have a current streak when it shouldn't?

notbakaneko (changes requested):

app/Models/DailyChallengeUserStats.php
nanaya: it'll eventually be fixed by the daily job

notbakaneko (approved):

PR #11419, opened 2024-08-09
Fix supporter being required when selecting default options in beatmap listing

notbakaneko:
Check and normalize for default value when updating search filters instead of just the final url.

closes #10972


tests/karma/beatmapset-search-filters.spec.ts
```diff
@@ -2,11 +2,21 @@
...
   let subject: BeatmapsetSearchFilters;
 
   describe('.query', () => {
+    beforeAll(() => {
+      core.setCurrentUser(testCurrentUserJson);
+    });
+
+    afterAll(() => {
+      core.setCurrentUser({ id: undefined });
+    });
```
nanaya, line 18: what's this for here?

nanaya (changes requested):

tests/karma/beatmapset-search-filters.spec.ts
notbakaneko: actual problem was if tests that simulated having a user ran first and didn't have `user_preferences` set left it undefined when logging out

nanaya (approved):

PR #11420, opened 2024-08-09
Add separate container for javascript test

nanaya:
Prevents normal assets builder and karma test from conflicting with each other ಠ_ಠ

There doesn't seem to be any way to prevent docker compose service from starting by default.


notbakaneko (commented):
Setting `profiles` like `profiles: [test]` prevents it from running by default iirc

notbakaneko (approved):

PR #11428, opened 2024-08-11
Add platform issue tags

venix12:
This adds buttons for "[osu!lazer]", "[osu!stable]", and "[osu!web]" thread title tagging, as of request of help forum moderators.

Uses separate const from regular tags as it doesn't have icon thing, and is supposed to be visible in thread names in the listing.

~~Also, genuinely had no idea for icons so they could be better possibly, though a mod-only feature so probably shouldn't matter as much 🤷.~~ It is text now.

In order: osu!lazer, osu!stable, osu!web
image


peppy:
Do these buttons actually fit? Can you provide a screenshot showing how this looks?

Also I don't think we need localisation for any of this.

venix12:
Updated the OP with a screenshot. 

As for translations, it's built into issue tags workflow so is easier to just go that way rather than omitting it, it's only tooltips and success messages that get translated anyway, the tag itself doesn't.

peppy:
is it possible to replace the icons with text? might make it more legible than almost-understandable icons. laz / stb / web maybe.

app/helpers.php
```diff
@@ -1100,6 +1100,15 @@ function concat_path($paths)
     return implode('/', array_filter($paths, 'present'));
 }
 
+function platform_issue_text($issue)
+{
+    return match ($issue) {
+        'lazer' => 'laz',
```
Hiviexd, line 1106: `lzr` feels like a better acronym than `laz`, thoughts?
```suggestion
        'lazer' => 'lzr',
```

resources/views/forum/topics/_nav.blade.php
```diff
@@ -10,7 +10,7 @@
         @include('forum.topics._moderate_move', compact('topic'))
 
         @if ($topic->isIssue())
-            @foreach ($topic::ISSUE_TAGS as $type)
+            @foreach (array_merge($topic::ISSUE_TAGS, $topic::PLATFORM_ISSUE_TAGS) as $type)
                 @include("forum.topics._issue_tag_{$type}")
```
cl8n, line 14: or
```blade
@include('forum.topics._issue_tag', ['issueTag' => $type])
```
and no extra blade templates

app/Models/Forum/Topic.php
```diff
@@ -811,6 +817,10 @@ public function setIssueTag($tag)
     {
         $this->topic_type = static::typeInt($tag === 'confirmed' ? 'sticky' : 'normal');
 
+        if (in_array($tag, static::PLATFORM_ISSUE_TAGS, true)) {
+            $tag = "osu!{$tag}";
+        }
```
cl8n, line 822: probably move this logic to its own method cuz it comes up 3 times (`tagForTopicTitle()` or something)

alternatively include the prefix in each PLATFORM_ISSUE_TAGS

cl8n (changes requested):

nanaya:
feels like the tags should just be moved to a menu with this amount

peppy:
I wouldn't be against that.

Blooshing:
Making a new panel for the buttons would be nice, I wouldn't mind that. Though this design works fine as well since only a few people are using the buttons it really wouldn't matter either way. 

Thanks for the quick work Venix, I appreciate it.

app/Models/Forum/Topic.php
```diff
@@ -103,6 +103,12 @@ class Topic extends Model implements AfterCommit
         'topic_title' => 100,
     ];
 
+    const PLATFORM_ISSUE_TAGS = [
+        'lazer',
```
nanaya, line 107: `str_slug` the type (class name, translation, filename) and use the actual tag string everywhere else instead? The icon helper can just generate the full content (`stb` etc string for the new tags and `<i ...>` for everything else)

resources/views/forum/topics/_issue_tag.blade.php
```diff
@@ -25,6 +25,10 @@ class="
     data-method="post"
 >
     <span class="btn-circle__content">
-        <i class="{{ issue_icon($issueTag) }}"></i>
+        @if (in_array($issueTag, $topic::ISSUE_TAGS, true))
```
nanaya, line 28: view `forum.forums._topic` also needs adjustment

nanaya (changes requested):

tatitatai:
I'm still adding the help thread tags manually, just asking because I don't really know github workflow, are we still going to add this at some point? I also would like to ask, is it possible for each icon to have their own color? So that the overview of help forum looks nice too. 
And, why no icons? Can we look for free ones?

By the way, we're currently using these tags:
- question 
- osu!web
- osu!lazer
- osu!stable
- technical support

notbakaneko (approved):

PR #11433, opened 2024-08-14
Better ApiDoc definitions

notbakaneko:
Fixing apidocs with wrong auto-generated examples, etc.
Moving repetitive definitions to attributes so their values can be set be params instead of repeating the whole thing every time. Using attributes because customizing the parsing for phpdoc blocks is not so great...

Draft while I get better ideas on handling params


nanaya:
I wonder if all the default limit should be unified to the const

notbakaneko:
redraft again while I check if there was something else to be updated after the other PR was merged 🤔

PR #11435, opened 2024-08-15
Add download mirror fallback

nanaya:
This prevents error when all of the mirrors are disabled. It itself is usually caused by temporary routing error from monitoring process so returning main server is usually fine.

Also cleaned up stuff along the way.


peppy (approved):
Logic looks about right

notbakaneko (approved):

PR #11436, opened 2024-08-16
Support prefixing session name

nanaya:
CSRF being in the cookie ended up pretty annoying so I removed it instead. notification server also need an update to handle the new name.


notbakaneko (approved):

notbakaneko:
does ppy/osu-notification-server#59 need to be deployed first? 🤔

nanaya:
shouldn't be. Also we're only prefixing dev so probably doesn't matter.

PR #11437, opened 2024-08-17
Increase Beatmap Discussion auto refresh interval to 30 seconds

notbakaneko:
Maybe 10 seconds is too frequent if changes keep happening 🤔


nanaya (approved):
sure 🤷

peppy:
I do wonder if at the point it's 30 seconds, people are just going to be refreshing the page anyway. Like unless there's something visually showing the user that there's polling / auto-refreshing going on, a lot of users aren't going to realise and just reload the page out of habit.

peppy:
Maybe just remove it altogether >_>

notbakaneko:
We decided to try a different thing for this instead

PR #11451, opened 2024-08-26
Refactor username input lookup component

notbakaneko:
To be used by #11377

Combined the `checkUsernameExists` into the new lookup as well.


notbakaneko:
I wonder if having a `ref` to the component and just getting the value out directly would be better than passing the updated value through a callback 🤔

resources/js/store/store-supporter-tag.tsx
```diff
@@ -223,16 +229,15 @@ export default class StoreSupporterTag extends React.Component<Props> {
 
   @action
   private readonly getUser = (username: string) => {
-    this.xhr = $.ajax({
-      data: { username },
-      dataType: 'json',
-      type: 'POST',
-      url: route('users.check-username-exists'),
-    });
+    this.xhr = apiLookup([username]);
```
nanaya, line 232: the gift link generated in user menu should be `@` prefixed as it was previously strictly username lookup

resources/js/store/store-supporter-tag.tsx
```diff
@@ -223,16 +229,15 @@ export default class StoreSupporterTag extends React.Component<Props> {
...
 
     this.xhr
-      .done((data) => runInAction(() => {
-        this.user = data;
+      .done((response) => runInAction(() => {
+        if (response.users.length > 0) {
+          this.user = response.users[0];
+        } else {
+          this.user = userNotFoundJson;
```
nanaya, line 239: the response isn't an observable so might as well just `response.users[0] ?? userNotFoundJson`?

nanaya (changes requested):

resources/js/store/store-supporter-tag.tsx
nanaya: or maybe just prefix it before sending the request

resources/js/components/username-input.tsx
```diff
@@ -0,0 +1,224 @@
...
+    // spinner should trigger even before request is sent.
+    this.busy = true;
+    this.debouncedLookupUsers();
+
+    if (immediate) {
+      this.debouncedLookupUsers.flush();
+    }
+  }
+
+  private validUsersContains(userIdOrUsername?: string | null) {
```
nanaya, line 213: ...shouldn't this be `Contain` instead? 🤔

nanaya (changes requested):

nanaya (approved):

PR #11453, opened 2024-08-26
Adjust sort button colour

nanaya:
Links are link coloured and hover background is more visible (on ranking page).


notbakaneko (approved):
makes more sense to me 🤷

PR #11461, opened 2024-08-30
Change parameter for country flag view

nanaya:
Also pass single modifier as string.

Simplifies passing the parameter for exactly one scenario (country model with title enabled) so not quite sure if worth it 🤔


resources/views/objects/_flag_country.blade.php
```diff
@@ -3,8 +3,8 @@
     See the LICENCE file in the repository root for full licence text.
 --}}
 <span class="{{ class_with_modifiers('flag-country', $modifiers ?? []) }}"
-    @if (isset($countryName))
-        title="{{ $countryName }}"
+    @if (isset($country['name']))
+        title="{{ $country['name'] }}"
     @endif
-    style="background-image: url('{{ flag_url($countryCode) }}');"
+    style="background-image: url('{{ flag_url($country['acronym']) }}');"
```
notbakaneko, line 9: Maybe check if `string` or object and handle accordingly? I'm not sure if wrapping everything in `['acronym' => ]` is an improvement 🤔

notbakaneko (approved):

PR #11465, opened 2024-09-02
Always show own score on multiplayer room leaderboard

nanaya:
Another update which mainly is for daily challenge but also useful in general.


notbakaneko (approved):
such row 🤔

PR #11471, opened 2024-09-04
Beatmap ranking timer changes

### LINKED ISSUE #11327: Reset the Ranking queue timer of a beatmap in certain cases
Following a discussion with the NAT a while ago, we want the ranking queue timer to be adjusted in the following ways:

- Reset if a new difficulty is added
- Reset if the map is renominated by a BN who wasn't one of the original nominators responsible for the latest qualification
- Have the timer go back up 1 day for each passing week that the map remains disqualified, until it reaches 7 days

main reason for this change is the increasing number of cases where new diffs and maps that went through significant changes under a prolonged time get ranked almost immediately with barely any time for them to have enough exposure in the qualified section.

notbakaneko:
closes #11327

- Resets if a new difficulty is added after disqualification
 - existing disqualified maps before this change will have the timer reset (because there's no `beatmap_ids` saved in the old events)
- Doesn't reset queue timer if renominated by a nominator that was part of the most recent nomination that qualified the map before disqualification (does not include the nominators that only nominated before a nomination reset, because they were not the nominations that got the map qualified); otherwise it resets.
- Applies a penalty of 1 day for each week the map remains disqualified, up to `BEATMAPSET_MAXIMUM_DISQUALIFIED_RANK_PENALTY_DAYS` days (default 7)
- [x] Also test timing issues when it crosses over to the next second during a test...which happens more often that expected
- [x] Some kind of warning if map will requalify with timer reset if qualified with different nominators

Going to look at refactoring some of the `qualify`/`setApproved` separately because that thing is getting long ಠ_ಠ


app/Models/Beatmapset.php
```diff
@@ -621,9 +628,30 @@ public function setApproved($state, $user, ?array $beatmapIds = null)
...
             $this->queued_at = null;
         } elseif ($this->isPending() && $state === 'qualified') {
-            $maxAdjustment = ($GLOBALS['cfg']['osu']['beatmapset']['minimum_days_for_rank'] - 1) * 24 * 3600;
-            $adjustment = min($this->previous_queue_duration, $maxAdjustment);
-            $this->queued_at = $currentTime->copy()->subSeconds($adjustment);
+            // Check if any beatmaps where added after most recent invalidated nomination.
+            $lastResetNominationTime = $this->beatmapsetNominations()->current(false)->max('reset_at');
+            if (
+                $lastResetNominationTime !== null
+                && $this->beatmaps()->where('last_update', '>', $lastResetNominationTime)->exists()
```
nanaya, line 635: doesn't updating beatmap also update this column?

notbakaneko: Indeed it does 🤔

nanaya: should deleting a map reset the time?

notbakaneko: The reason for resetting on new difficulties is so they don't skip the queue when they could be introducing new problems and end up needing last minute popping/unrank or whatever; deleting won't be introducing new issues in the map set.

tests/Models/BeatmapsetRequalifyTest.php
```diff
@@ -0,0 +1,261 @@
...
+
+        // sanity
+        $this->assertSame(0, $beatmapset->previous_queue_duration);
+        $this->assertEquals($qualifiedDate, $beatmapset->approved_date);
+
+        $this->travelTo($disqualifiedDate);
+
+        $discussion = $this->disqualifyOrResetNominations($beatmapset);
+        $beatmapset = $beatmapset->fresh();
+        // TODO: timing issues 86401 != 86400
```
nanaya, line 48: just strip the last digit maybe? or check if less than few seconds difference

app/Models/Beatmapset.php
```diff
@@ -714,13 +745,29 @@ public function disqualifyOrResetNominations(User $user, BeatmapDiscussion $disc
...
-        DB::transaction(function () use ($user) {
+        $this->getConnection()->transaction(function () use ($user) {
+            // If beatmapset was previously disqualified, reset the queue timer
+            // if the current nominators are current nominators are different from the previous qualified nominations.
+            $disqualifyEvents = $this->events()->where('type', BeatmapsetEvent::DISQUALIFY)->orderBy('id', 'desc')->limit(2)->get();
+            if ($disqualifyEvents !== null) {
+                $previousNominations = $this->beatmapsetNominations()->current(false);
+                $previousNominators = new Set($disqualifyEvents->count() === 2
+                    ? $previousNominations->where('event_id', '>', $disqualifyEvents[1]->getKey())->pluck('user_id')
+                    : $previousNominations->pluck('user_id'));
```
nanaya, line 762: what if there's nomination reset before disqualification?

app/Models/Beatmapset.php
```diff
@@ -714,13 +745,29 @@ public function disqualifyOrResetNominations(User $user, BeatmapDiscussion $disc
...
     {
         if (!$this->isPending()) {
-            return;
+            throw new InvariantException('cannot qualify a beatmapset not in a pending state.');
         }
 
-        DB::transaction(function () use ($user) {
+        $this->getConnection()->transaction(function () use ($user) {
+            // If beatmapset was previously disqualified, reset the queue timer
+            // if the current nominators are current nominators are different from the previous qualified nominations.
```
nanaya, line 756: this comment doesn't read quite right. It doesn't help the second line is longer than the first one. Maybe something like "Reset ... if ... and ..." instead?

nanaya (changes requested):

nanaya (changes requested):

resources/js/beatmap-discussions/nominator.tsx
```diff
@@ -293,12 +301,26 @@ export class Nominator extends React.Component<Props> {
...
       </>
     );
   }
 
   private renderModalContentNormal() {
-    return trans('beatmapsets.nominate.dialog.confirmation');
+    return (
+      <>
+        {trans('beatmapsets.nominate.dialog.confirmation')}
+        {this.renderNominatorWarning()}
```
nanaya, line 313: as in both version this is located at the end of the content and not wrapped in anything, why not just put this directly in the main render function?

nanaya (changes requested):

resources/js/beatmap-discussions/nominator.tsx
```diff
@@ -300,7 +313,6 @@ export class Nominator extends React.Component<Props> {
   private renderModalContentNormal() {
     return trans('beatmapsets.nominate.dialog.confirmation');
   }
-
```
nanaya, line 303: this can stay, I think

notbakaneko: how did that happen ಠ_ಠ

PR #11477, opened 2024-09-06
Sync user preferences when updated from a different tab

### LINKED ISSUE #11295: Scores on leaderboards sometimes display in wrong order
Context: I general view the osu! site while not on lazer mode. However, in the midst of a debate with a friend, I switched to lazer mode while looking at the beatmap page for [Padoru Padoru](https://osu.ppy.sh/beatmapsets/1073074#osu/2245774) to grab a reference for lazer score multipiers. It was here where I noticed that the scores were being displayed in the wrong order.
image
I'm not entirely sure what caused this, as when looking at the site on my computer, the scores display in the correct order. The issue is only present on the specific difficulty I was initially looking at, on only global leaderboards while not filtering by mods. The issue did not persist after reloading the page, and I don't know how to reproduce it.

Phone is iPhone SE (2nd generation), browser is Google Chrome version 216.0.6478.54

nanaya:
Resolves #11295... kind of.

1. (tab A) open a beatmap page
2. (tab B) open some page
3. (tab B) toggle lazer mode
4. (tab A) switch to different difficulty

The result is the scores are fetched using the new option but sorted using the old one.

This PR doesn't change the header correctly but there's currently no framework in place to do it so maybe if someone complains about it.

### DIFF
--- a/resources/js/core/user/user-preferences.ts
+++ b/resources/js/core/user/user-preferences.ts
@@ -7,6 +7,8 @@ import { route } from 'laroute';
 import { action, makeObservable, observable } from 'mobx';
 import { onErrorWithCallback } from 'utils/ajax';
 
+const localStorageKey = 'userPreferences';
+
 export default class UserPreferences {
   @observable private current: UserPreferencesJson;
   private updatingOptions = false;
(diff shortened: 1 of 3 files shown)


resources/js/core/user/user-preferences.ts
```diff
@@ -13,9 +13,13 @@ export default class UserPreferences {
   private user?: CurrentUserJson;
 
   constructor() {
-    this.current = Object.assign({}, defaultUserPreferencesJson, this.fromStorage());
+    this.current = this.fromStorageWithDefaults();
 
     makeObservable(this);
+
+    window.addEventListener('storage', action(() => {
```
notbakaneko, line 20: should be limited to the `userPreferences` key and maybe `null` for `clear()`

notbakaneko (changes requested):

notbakaneko (approved):

PR #11482, opened 2024-09-11
Fix spotlight dropdown height

notbakaneko:
Spotlight dropdown should scroll, not make the page scroll...


resources/views/multiplayer/rooms/show.blade.php
```diff
@@ -18,7 +18,7 @@
 @section('ranking-header')
     <div class="osu-page osu-page--ranking-info">
         <div class="js-react--basic-select-options">
-            <div class="select-options select-options--spotlight">
```
nanaya, line 21: remove the style as well? and somehow I put this modifier for daily challenge blade view 🤨

nanaya (changes requested):

resources/views/multiplayer/rooms/show.blade.php
notbakaneko: what the.... I thought I removed this already 🤨

nanaya: oh wait the style has been gone since forever, I was looking at different thing.

nanaya (approved):

PR #11487, opened 2024-09-13
Don't show restricted user in daily challenge ranking

notbakaneko:


notbakaneko:
was looking at the wrong thing

PR #11488, opened 2024-09-13
Filter out restricted user from Playlist scores

notbakaneko:


app/Http/Controllers/Multiplayer/Rooms/Playlist/ScoresController.php
```diff
@@ -53,7 +53,12 @@ public function index($roomId, $playlistId)
         $limit = clamp(get_int($params['limit'] ?? null) ?? 50, 1, 50);
         $cursorHelper = PlaylistItemUserHighScore::makeDbCursorHelper($params['sort'] ?? null);
 
-        $highScoresQuery = $playlist->highScores()->whereHas('scoreLink');
+        $highScoresQuery = $playlist
+            ->highScores()
+            ->whereHas('user', function ($userQuery) {
+                $userQuery->default();
+            })
```
nanaya, line 60:
```suggestion
            ->whereHas('user', fn ($userQuery) => $userQuery->default())
```

nanaya (changes requested):

app/Models/Multiplayer/PlaylistItemUserHighScore.php
```diff
@@ -130,4 +131,9 @@ public function updateWithScoreLink(ScoreLink $scoreLink): bool
             'total_score' => $score->total_score,
         ])->save();
     }
+
+    public function user(): BelongsTo
```
nanaya, line 135: oh right, this should be in relation function area?

nanaya (changes requested):

app/Models/Multiplayer/PlaylistItemUserHighScore.php
notbakaneko: are we still doing that 🤔

nanaya: I think so for scopes, relations and laravel model setters/getters? 🤔

nanaya (approved):

PR #11491, opened 2024-09-13
Fix BeatmapsetRequalifyTest failing due to timing variance.

notbakaneko:
At least until we explicitly pass the timestamp into the qualify function


tests/Models/BeatmapsetRequalifyTest.php
```diff
@@ -262,6 +262,11 @@ protected function setUp(): void
         Bus::fake([CheckBeatmapsetCovers::class]);
     }
 
+    private function assertDiffWithinRange(int $expected, int $actual, int $range = 1)
```
nanaya, line 265: shouldn't return anything? also the function call reads kinda weird with the default

nanaya (changes requested):

nanaya (approved):

PR #11492, opened 2024-09-13
Use fa-var variables

nanaya:
Reference: `node_modules/@fortawesome/fontawesome-free/less/_variables.less`


notbakaneko (changes requested):
might as well replace the `\f105` as well?

notbakaneko (approved):

PR #11493, opened 2024-09-13
Use subgrid for store cart

nanaya:
Randomly remembered about this when looking for some other stuff (form select styling). Looks like it can use some minimum sizing though? 🤔


notbakaneko (changes requested):
Can do with some minimum sizing (also a bit weird when changing the drop down option causes the column to resize due to different width)

notbakaneko (approved):

PR #11499, opened 2024-09-18
Cleanup discussion related search query and params helper

notbakaneko:
`pagination()` isn't used by anything else anymore, so might as well streamline it for the only places it's used...

...also I'm not sure the original reason the page size of these is 20, other than 50 gets kind of long when displaying discussions or discussion posts in one page, especially if there're reviews in them 👀


nanaya (changes requested):
I think it's fine to bump the default to 50 🤔

nanaya (approved):

PR #11503, opened 2024-09-23
Add support for rejecting payments

notbakaneko:
Also adds a command for getting an `Order`'s corresponding details from paypal.

### DIFF
--- a/app/Console/Commands/StoreGetPaypalOrder.php
+++ b/app/Console/Commands/StoreGetPaypalOrder.php
@@ -3,6 +3,8 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
 // See the LICENCE file in the repository root for full licence text.
 
+declare(strict_types=1);
+
 namespace App\Console\Commands;
 
 use App\Libraries\Payments\PaypalApiContext;
--- a/app/Console/Commands/StoreGetPaypalOrder.php
+++ b/app/Console/Commands/StoreGetPaypalOrder.php
@@ -23,13 +23,13 @@ class StoreGetPaypalOrder extends Command
         $order = Order::findOrFail(get_int($this->argument('orderId')));
         if ($order->provider !== 'paypal') {
             $this->error('Not a Paypal order');
-            return;
+            return static::INVALID;
         }
 
         $paypalOrderId = $order->reference;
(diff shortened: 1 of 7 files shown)


app/Models/Store/PaypalBanned.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace App\Models\Store;
+
+/**
+ * @property int $id
+ * @property string|null $account_id
+ * @property string|null $email
+ */
+class PaypalBanned extends Model
+{
```
nanaya, line 14: `$timestamps = false` somewhere?

app/Console/Commands/StoreGetPaypalOrder.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
```
nanaya, line 2: can use the strict thing

app/Console/Commands/StoreGetPaypalOrder.php
```diff
@@ -0,0 +1,39 @@
...
+        $order = Order::findOrFail(get_int($this->argument('orderId')));
+        if ($order->provider !== 'paypal') {
+            $this->error('Not a Paypal order');
+            return;
+        }
+
+        $paypalOrderId = $order->reference;
+        if ($paypalOrderId === null) {
+            $this->error('Missing Paypal order id');
+            return;
```
nanaya, line 30: these should probably return 1 (or `Symfony\Component\Console\Command\Command::INVALID` or something)

app/Http/Controllers/Payments/PaypalController.php
```diff
@@ -61,6 +62,8 @@ public function approved()
             (new PaypalExecutePayment($order))->run();
         } catch (HttpException $e) {
             return $this->setAndRedirectCheckoutError($order, $this->userErrorMessage($e));
+        } catch (PaymentRejectedException $e) {
```
nanaya, line 65: apparently php 8 allows skipping the variable

app/Models/Store/PaypalBanned.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
```
nanaya, line 2: can use strict as well

nanaya (changes requested):

nanaya (approved):

PR #11505, opened 2024-09-25
Remove redundant user card endpoint

nanaya:
Now I look closer the endpoint is a bit different compared to the others but it should only affect admins...


notbakaneko (approved):
I think admins being able to see the card was one of the points but the change does make the visual behaviour more consistent (as opposed to "the card works for me but not others"), so we'll see how it goes 🤷

PR #11506, opened 2024-09-25
Open up user lookup api

nanaya:
I think the limit is about 50 times higher than the index one (for 50 users lookup). The endpoint itself is relatively lightweight 🤷 

Renamed things so they're less redundant.

Related: ppy/osu#29982


peppy (approved):

notbakaneko:
I just remembered a difference between the user lookup and removed user card endpoint in #11505 is the user card one allows restricted users to lookup themselves and see their own card.

notbakaneko (commented):
I wonder if the lookup should also exclude `DeletedUser_*` or have the transformer return an object stripped of info if this endpoint is just going to be open

notbakaneko (approved):

PR #11510, opened 2024-09-26
Don't send user ids for announcement channels

nanaya:
In the meantime until a separate function for fetching users is added.


peppy (approved):
Probably fine

notbakaneko (approved):

PR #11511, opened 2024-09-26
Add helper command for getting Shopify-related checkout details

notbakaneko:
Mainly to get the Shopify order `orderNumber` or `id`.
It calls the public Storefront API so it can use the same token instead of needed a separate oauth client/secret for the admin API.
API version is matched to the current `shopify-buy` sdk's version (and different version has different schema and I don't feel like updating it now)


peppy:
Do you think it would be feasible to store these at some point in the checkout process to the database? This would allow for easier lookups from external tooling.

notbakaneko:
That's part of the plan.
The current public token is also rate-limited by IP which normally isn't a problem during checkout since they're initiated from the user's end;

nanaya (approved):
Seems like a lot of things for what should be just storing `order_number` during the webhook callback or something 🤔 I guess it's useful for checking existing orders

PR #11518, opened 2024-09-30
Put back announcement channel user listing

nanaya:
Now with limit.

~~The user loading flash for those with no listing permission gets a bit worse here but I think fixing it should be a different PR 🤔~~ added additional priv attribute to check if it should load user list

(or maybe someone can come up with better design)


resources/js/chat/conversation-view.tsx
```diff
@@ -259,16 +259,33 @@ export default class ConversationView extends React.Component<Props> {
...
-        ) : (
-          this.currentChannel.announcementUsers.map((user) => (
-            <UserCardBrick key={user.id} user={user.toJson()} />
-          ))
+      <div className='chat-conversation__users-container'>
+        <div className={classWithModifiers('chat-conversation__users', { loading: users == null })}>
+          {users == null ? (
+            <>
+              <Spinner modifiers='self-center' /><span>{trans('chat.loading_users')}</span>
+            </>
```
notbakaneko, line 274: I think we can just skip trying to load users if they don't have permission to view it anyway. 
Or maybe return `users: []` with the channel if they have no permission to load users and `null` otherwise 🤔  (or both)

notbakaneko (changes requested):

resources/js/chat/conversation-view.tsx
nanaya: I'll look into that... eventually...

probably not changing existing json fields though.

notbakaneko (approved):

PR #11529, opened 2024-10-03
Form styling cleanup

nanaya:
The smaller cleanup before going with a larger one. Mainly to unify input container styles.

Notable changes include same background colour and padding.

### DIFF
--- a/resources/css/bem/input-container.less
+++ b/resources/css/bem/input-container.less
@@ -8,7 +8,7 @@
   display: flex;
   flex-direction: column;
   background: var(--input-bg);
-  border-radius: 5px;
+  border-radius: @border-radius-large;
   text-transform: initial;
   font-weight: initial;
   font-size: inherit;
--- a/resources/css/bem/input-container.less
+++ b/resources/css/bem/input-container.less
@@ -46,7 +46,7 @@
 
   &--genre {
     --label-colour: hsl(var(--hsl-c1));
-    background: none;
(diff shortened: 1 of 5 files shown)


resources/css/bem/input-container.less
```diff
@@ -3,20 +3,37 @@
...
   --label-colour: hsl(var(--hsl-c2));
-  --input-bg: hsl(var(--hsl-b5));
+  --label-padding-base: 5px;
+  --input-bg: hsl(var(--hsl-b6));
+  --input-border-colour: var(--input-border-base-colour);
+  --input-border-base-colour: transparent;
+  --input-border-hover-colour: hsl(var(--hsl-l3));
+  --input-border-focus-colour: hsl(var(--hsl-l1));
+  --input-border-error-colour: hsl(var(--hsl-red-2));
+  --input-padding: var(--label-padding-base) var(--input-padding-base)
```
notbakaneko, line 13: What's the reason for splitting up the container's inner padding into label and input outer padding (aka margin)?

notbakaneko (changes requested):
padding is missing
image

resources/css/bem/input-container.less
nanaya: the end result would be to absolute position the label and have the input take the whole area. It's mainly for select input so clicking anywhere will show the dropdown

notbakaneko (approved):

PR #11530, opened 2024-10-03
Add warning before sending message to announcement channel

nanaya:
Can use a better message but it's a start, I suppose :D

The function name is a bit questionable as well.


notbakaneko (approved):
ಠ_ಠ

PR #11534, opened 2024-10-07
Update dropdown on store page

nanaya:
Some hack involving floating the label is involved 🔧

### DIFF
--- a/resources/css/bem/input-container.less
+++ b/resources/css/bem/input-container.less
@@ -2,26 +2,25 @@
 // See the LICENCE file in the repository root for full licence text.
 
 .input-container {
+  --label-height: 25px;
   --label-colour: hsl(var(--hsl-c2));
-  --label-padding-base: 5px;
+  --label-pointer-events: auto;
   --input-bg: hsl(var(--hsl-b6));
   --input-border-colour: var(--input-border-base-colour);
   --input-border-base-colour: transparent;
   --input-border-hover-colour: hsl(var(--hsl-l3));
   --input-border-focus-colour: hsl(var(--hsl-l1));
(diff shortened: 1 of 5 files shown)


resources/css/bem/input-container.less
```diff
@@ -74,12 +71,33 @@
...
+
   &__label {
     display: flex;
     justify-content: space-between;
     margin: 0;
     color: var(--label-colour);
     font-size: @font-size--normal;
-    padding: var(--label-padding-base) var(--input-padding-base) 0;
+    padding: 5px var(--input-padding-base) 0;
+    position: absolute;
```
notbakaneko, line 99: can do with a comment on why it's `position: absolute`, also a note for the sizing of `--label-height` (or calculate it instead?)

notbakaneko (changes requested):

resources/css/bem/input-container.less
```diff
@@ -3,7 +3,9 @@
 
 .input-container {
   --label-font-size: @font-size--normal;
-  --label-height: calc(var(--label-font-size) + var(--input-padding-base) + 5px);
+  --label-height: calc(
```
notbakaneko, line 6: I think this should be `font-size * 1.25 (the line-height) + padding base` tbh 🤔

nanaya: eh close enough 🤷

notbakaneko (approved):

nanaya:
wait no, that was with wrong padding - it should be 5px both sides

notbakaneko:
wut

notbakaneko (approved):

PR #11536, opened 2024-10-08
Fix incorrect cutoffs on grade dial when displaying lazer scores and stable scores in lazer mode

bdach:
As reported on discord: https://discord.com/channels/188630481301012481/1097318920991559880/1293089915394985985

| web, master | client | web, this PR |
| :-: | :-: | :-: |
| before | client | after |

Also contains amendments to the colour values to match client 1:1.


resources/js/scores-show/info.tsx
```diff
@@ -27,7 +27,7 @@ export default function Info({ score }: Props) {
       </div>
 
       <div className='score-info__item score-info__item--dial'>
-        <Dial accuracy={accuracy(score)} mode={rulesetName(score.ruleset_id)} rank={rank(score)} />
+        <Dial accuracy={accuracy(score)} mode={rulesetName(score.ruleset_id)} rank={rank(score)} rank_cutoffs={rankCutoffs(score)} />
```
nanaya, line 30: mode can be removed, also it should be `rankCutoffs` as props that's not json are camelCased

resources/js/utils/score-helper.ts
```diff
@@ -105,6 +105,67 @@ export function rank(score: SoloScoreJson) {
     : score.rank;
 }
 
+export function rankCutoffs(score: SoloScoreJson): number[] {
+  // <rank>: minimum acc => (higher rank acc - current acc)
+  // for SS, use minimum accuracy of 0.99 (any less and it's too small)
+  // actual array is reversed as it's rendered from D to SS clockwise
+
+  if (shouldReturnLegacyValue(score)) {
+    return {
```
nanaya, line 114: just a matter of style but I wonder if a switch is easier to read. And maybe a different function to calculate the result array from the actual accuracy requirements - transforming the 0.99, 0.9801, etc to the expected values - while at it.

nanaya (changes requested):

resources/js/utils/score-helper.ts
bdach: I made an attempt in https://github.com/ppy/osu-web/pull/11536/commits/a86ce4138c86fa0a834860f180321a0587703228, I guess see if you like it better?

nanaya: yeah that's definitely easier to understand 😄 

(the switch still kinda sucks but that's just javascript)

nanaya (approved):

PR #11537, opened 2024-10-08
Adjust dashboard banner position

nanaya:
The glow is kinda intentional 👀

### DIFF
--- a/resources/css/bem/menu-image.less
+++ b/resources/css/bem/menu-image.less
@@ -10,7 +10,8 @@
 
   &__image {
     object-fit: contain;
-    width: 100%;
     height: 100%;
+    margin: auto;
+    display: block;
   }
 }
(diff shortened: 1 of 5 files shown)


peppy (approved):
Went through design changes IRL 👍  Better than what we have live so let's get this out and then work forward from there.

notbakaneko:
Is the glow supposed to bleed everywhere?

nanaya:
yes

notbakaneko (changes requested):
umm
image
image
image

(also it's kind of hard to tap anything other than the banner link on mobile)

nanaya:
yeah I'm not sure why the indicator button isn't clickable on the content padding area ಠ_ಠ

notbakaneko (approved):

PR #11538, opened 2024-10-09
Remove some react container from page layout

nanaya:
Just a random cleanup 🧹


notbakaneko (approved):
Should be fine 🤔 (they're just repeating the `osu-layout__section--full` layout anyway)

PR #11544, opened 2024-10-10
Remove usage of old score es index

nanaya:
Everything is using the solo scores now.


notbakaneko (approved):
Does anything else need to be updated (like stop calling the lio)?

peppy:
It should be fine. That endpoint is specifically used for cleaning up ES scores after a change in user state. The return value is never used so I'll just remove all calls to it.

PR #11549, opened 2024-10-12
add beatmap favourite count search parameter

### LINKED ISSUE #11462: Add filtering by favorite count
Would be nice if you could set a minimum favorite count while searching on the website (favorites>=5) like you could on osusearch. I feel like its the only thing missing from the search being perfect.

LBlend:
Closes #11462

One thing I'm uncertain about is whether or not there should be support for the American English spelling of "favorite". As I see no references to the American spelling in the code I decided not to.


LBlend:
I noticed you ran the tests and that it failed. Should be fixed now.

nanaya (approved):

PR #11556, opened 2024-10-15
Use origin check for guests

nanaya:
It seems that with "recent" browsers, origin headers can be used in place of csrf token.


app/helpers.php
```diff
@@ -816,12 +817,15 @@ function ext_view($view, $data = null, $type = null, $status = null)
...
-    // Add trailing slash so people can't just use https://osu.web.domain.com
-    // to bypass https://osu.web referrer check.
+    $headers = ($request ?? Request::instance())->headers;
+    $appUrl = $GLOBALS['cfg']['app']['url'];
+    // Add trailing slash for referer check to avoid matching
+    // https://osu.web.domain.com.
     // This assumes app.url doesn't contain trailing slash.
-    return starts_with(request()->headers->get('referer'), $GLOBALS['cfg']['app']['url'].'/');
+    return $headers->get('origin') === $appUrl
+        || str_starts_with($headers->get('referer'), "{$appUrl}/");
```
notbakaneko, line 828: As far as I know, it's possible to do something stupid like run chrome with `--disable-web-security` which allows changing the `Referer` to a different domain with `replaceState`

nanaya: play stupid game, win stupid prize

notbakaneko (approved):
should be fine 🤷

PR #11557, opened 2024-10-15
Fix error on null reportable

nanaya:
Not sure if a report with null reportable makes sense 🤔


notbakaneko:
what is it reporting...

nanaya:
my guess is it's deleted right after the report

notbakaneko (approved):

PR #11563, opened 2024-10-17
Add basic team backend

nanaya:
Just the models and a very basic show page 🤔

Coming next: settings page


resources/views/teams/show.blade.php
```diff
@@ -0,0 +1,100 @@
...
+                                {{ osu_trans('teams.show.sections.members') }}
+                            </h2>
+                            <div class="team-summary__members">
+                                <div class="team-members team-members--owner">
+                                    <div class="team-members__meta">
+                                        {{ osu_trans('teams.show.members.owner') }}
+                                    </div>
+                                    <div
+                                        class="js-react--user-card u-contents"
+                                        data-user="{{ json_encode($teamMembers['leader'][0]) }}"
```
notbakaneko, line 67: does this need a check if the leader exists? (or if the team has members at all? 🤔 )

notbakaneko (changes requested):

resources/views/teams/show.blade.php
nanaya: leader_id is required so I'd hope it'll always exist. But otherwise the handling the case when the leader is gone is currently missing and I'm not even quite sure what should happen yet. Maybe in a later pr...

notbakaneko:
Is this going to be used without without the UI for setting up teams or is it going to sit there until then?

nanaya:
it's just going to sit there until then

resources/css/bem/profile-info.less
```diff
@@ -2,17 +2,26 @@
 // See the LICENCE file in the repository root for full licence text.
 
 .profile-info {
+  --avatar: "";
```
agatemosu, line 5: Shouldn't this and the one below be "none"?

notbakaneko (changes requested):
description area could use some forced wrapping (also mobile)
image

app/Models/Team.php
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+
+    public function descriptionHtml(): string
+    {
+        $description = presence($this->description);
+
+        return $description === null
+            ? ''
+            : bbcode((new BBCodeForDB($description))->generate());
```
notbakaneko, line 60: are we still going for bbcode instead of markdown for new stuff? 🤔

nanaya: that's a good question but markdown barely covers what bbcode does and this is supposed to be similar to user page's 🤷

notbakaneko:
The images could do with some max width as well 👀

nanaya:
I've split the bbcode fixups to a different pr.

notbakaneko (approved):

PR #11566, opened 2024-10-18
Adjust timeout to match proxy

nanaya:
- [ ] check if there's anything that's been running for over a minute without the return code being confirmed (mainly for interop endpoint?)

### DIFF
--- a/config/octane.php
+++ b/config/octane.php
@@ -218,7 +218,7 @@ return [
     |
     */
 
-    'max_execution_time' => 180,
+    'max_execution_time' => 55,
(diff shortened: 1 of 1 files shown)


notbakaneko:
I remember contest admin downloads taking too long or something 🤔

nanaya:
I think I've fixed that one by streaming the response

nanaya:
hmm, I wonder if it's still needed once the transfer is started

nanaya:
need to fix this through a different method.

PR #11569, opened 2024-10-18
Fix wording in Daily Challenge Ranking page

### LINKED ISSUE #11533: Incorrect use of the term "10th percentile" in daily challenge
10th percentile supposed to describe lower part of the scores, basically a score that is better than only 10% of all scores.
firefox_e125CVaOJs
However, what it actually shows on the website is **90th percentile score**, or a score to get in top 10%.

I would also propose to change it to "top 10% score" and "top 50% score" instead, as not every person is familiar about percentiles and it might be confusing. That will be also consistent with the way it shows on user profiles.
firefox_Vk6IHvl6jr

chromb:
Replaces 10th Percentile (which was wrong) with Top 10% Score, which is likely easier to understand for most players.

Also fixes #11533

(I have also fixed some grammar in the Ukrainian translation at request of a friend).


bdach:
Non-English copy cannot be modified via a direct PR. Such changes must go through crowdin.

notbakaneko:
If we're making this change, the key should be renamed as well since keeping it will make even less sense.

nanaya (approved):
let's just put this in for now

PR #11581, opened 2024-10-22
Add api endpoint for adding/removing friends

nanaya:
Related to ppy/osu#30386.

- [ ] maybe decide whether or not to change the response for store/destroy to just the relevant user relation?


cdwcgt:
can `GET api/v2/friends` also give the information like: `"relation_type":"friend","mutual":true` in this pr?
As far as I know, the current API will not provide these information

nanaya:
that's on the other pr (#11582)

notbakaneko (commented):
https://github.com/ppy/osu-web/pull/11584#pullrequestreview-2401366270

notbakaneko (approved):

Issue #10789, opened 2023-12-07
Add information about blocks / mapping subscriptions for lazer consumption


bdach:
Currently, lazer only has the list of friends of the logged-in user available, via the response of the `GET /me` API endpoint. However - as far as I can tell - there is no way to tell via any API request whether the logged-in user is blocked by another user, or is a mapping subscriber of another user.

This is important as it makes it impossible to replicate the operation of the friend / mapping subscriber button from web on the in-game user profile overlay.

nanaya:
I'm not sure the relation with friend operation though...

bdach:
https://github.com/ppy/osu-web/blob/f4f6d25fb6c02d6caf64cf0aaf1ebe851b355ea8/resources/js/components/friend-button.tsx#L52-L61

Note block check specifically. Unless you're suggesting we should just show the button but presumably let the friend operation fail due to the block?

nanaya:
I don't remember why it's that way on frontend but the friend operation will actually still work

PR #11584, opened 2024-10-22
Add user block management api

nanaya:
That the destroy/store endpoint is returning both friends and blocks feels a bit whack 🤨

Also it means three additional requests just to get full relation lists (friend, block, mapper (once added))...

Related to #10789.

- [x] depends on #11581 for handling the new response

### DIFF
--- a/app/Http/Controllers/BlocksController.php
+++ b/app/Http/Controllers/BlocksController.php
@@ -49,11 +49,9 @@ class BlocksController extends Controller
             abort(404);
         }
 
-        $existingRelation = $currentUser
-            ->relations()
-            ->where('zebra_id', $targetId)
-            ->first();
+        $relationQuery = $currentUser->relations()->where('zebra_id', $targetId);
 
+        $existingRelation = $relationQuery->first();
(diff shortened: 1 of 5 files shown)


notbakaneko (changes requested):
I think maybe a separate relations listing and add/remove block/friend returning only the relevant record? 🤔

nanaya:
but then again for friends list you probably want to fetch the user data as well whereas for blocklist all that's needed is just the id so I'm not sure if combining them is useful

notbakaneko (changes requested):
Yeah I think it's kind of weird returning both friends and blocks, if anything at all...

I can see the case of knowing that the user is removed from the friends list if blocked, but that can be assumed to be implicit from blocking, so it isn't actually needed. 🤷 
The most I can see is if you're blocking/adding friends from multiple clients open but it would see strange to me if blocking one user in the UI causes the whole list to change. (and I feel like the reason for returning the whole list was 'shove everything into React to render and not think about it' 🤔 )

notbakaneko (approved):

PR #11587, opened 2024-10-24
Use native uuid generator function

nanaya:
It's been a thing since 2022 (thx safari).


notbakaneko (approved):

notbakaneko:
A note that safari need needs the site to be https for it to work

nanaya:
wow this secure context thing is kinda garbage 😐 at least all of our deployments are https and localhost is assumed secure...

notbakaneko:
huh, I guess blank tab is considered secure context in Chrome but not Safari 🤔

PR #11593, opened 2024-10-26
Move user registration step in `SETUP.md` to be after database seeding

frenzibyte:
As I was going through setting up a local web server, I've noticed user registration does not work, and gives an error:
```
> (new App\Libraries\UserRegistration(["username" => "frenzi", "user_email" => "frenzibyte@gmail.com", "user_password" => "1234"]))->save();

   Error  Call to a member function getKey() on null.
```

I was at a complete loss as to what's causing this, then I opened discord and searched in osu and came across [this](https://discord.com/channels/188630481301012481/188630616286167041/1259342156624302164). Turns out I have to perform the seeding step first so that the `default` user group becomes present for user registration to work (or so is what I can deduce from the error message, it might be something entirely different).

I expect to get things working by just reading the setup documentation from top to bottom, so I just moved the user registration step in this PR to be done after the seeding step.

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -208,15 +208,6 @@ p artisan tinker
 
 # Development
 
-## Generating assets
-
-```bash
-# build assets (should be done automatically if using docker)
-yarn run development
-```
-
-Note that if you use the bundled docker compose setup, yarn/webpack will be already run in watch mode.
-
 ## Reset the database + seeding sample data
 
 ```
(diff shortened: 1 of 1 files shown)


peppy:
Feels like we might want to make this more fool-proof though. cc @ppy/team-web

peppy (approved):

notbakaneko (changes requested):
I think we can reorder the first few sections under `Development` to
1. Reset the database + seeding sample data
2. Creating your initial user
3. Generating assets
4. Continuous asset generation while developing

notbakaneko (approved):

PR #11600, opened 2024-10-29
Replace Turbolinks with Turbo

nanaya:
Seems to work to some extent.

jquery-ujs stays for now as turbo doesn't emit as many form events but I think it can still be converted cleanly.

Also coming next is removing jquery sync script loading which will also fix console error log spewing nonsense file/line number due to how jquery does it.


notbakaneko (changes requested):
don't the `data-turbolinks-` attributes also have to be updated?

notbakaneko (changes requested):
there's still `data-turbolinks='false'` and equivalent

resources/js/utils/turbolinks.ts
```diff
@@ -25,20 +25,47 @@ export function currentUrlRelative() {
...
 
 export function reloadPage(keepScroll = true) {
   $(document).off('.ujsHideLoadingOverlay');
-  Turbolinks.clearCache();
+  Turbo.cache.clear();
 
   navigate(currentUrl().href, keepScroll, { action: 'replace' });
 }
+
+export function updateHistory(url: string, action: 'advance' | 'replace') {
```
notbakaneko, line 46: just make it `'push' | 'replace'` to match the methods? (and maybe add a type for it)

notbakaneko (changes requested):
seems fine other than another change 👀

notbakaneko (approved):

PR #11606, opened 2024-10-31
Don't create user notification entry if mail and push are both disabled

nanaya:
~~(untested and probably will break test)~~ oh wow nothing tests this specifically


peppy (approved):

peppy:
Won't deploy this until I get confirmation it's okay to go ahead with, merge may have been too eager.

notbakaneko:
Should be fine, this might have been set up with the expectation the option itself was deleted originally and it fell through, or the bit just flipped for testing 😬

PR #11610, opened 2024-10-31
Sort user notifications with existing notifications table index

nanaya:
This removes extra filesort as mentioned [here](https://discord.com/channels/90072389919997952/1083320217259749376/1301366162046582887) but there doesn't seem to be any difference, at least on small queries (including the linked one).

Original query:
``
SELECT * FROM `user_notifications` WHERE `user_notifications`.`user_id` = 2387883 and `user_notifications`.`user_id` IS not NULL and `delivery` & 1 and exists (SELECT * FROM `notifications` WHERE `user_notifications`.`notification_id` = `notifications`.`id` and `notifiable_type` = 'build' and `notifiable_id` = 7629 and `name` in ('comment_new')) and `is_read` = 0 ORDER BY `id` DESC LIMIT 50
``

This pr:
``
SELECT `user_notifications`.* FROM `user_notifications` inner join `notifications` on `notifications`.`id` = `user_notifications`.`notification_id` WHERE `user_notifications`.`user_id` = 2387883 and `user_notifications`.`user_id` IS not NULL and `delivery` & 1 and `notifications`.`notifiable_type` = 'channel' and `notifications`.`notifiable_id` = 56380464 and `notifications`.`name` in ('channel_message') and `is_read` = 0 ORDER BY `notifications`.`created_at` DESC, `notifications`.`id` DESC LIMIT 50
``

Also imported `reedware/laravel-relation-joins` to "help" doing the join.

### DIFF
--- a/app/Libraries/NotificationsBundle.php
+++ b/app/Libraries/NotificationsBundle.php
@@ -87,6 +87,7 @@ class NotificationsBundle
                     ->whereIn($q->qualifyColumn('name'), Notification::namesInCategory($category))
                     ->orderByDesc($q->qualifyColumn('created_at'))
                     ->orderByDesc($q->qualifyColumn('id'));
+
                 if ($this->cursorId !== null) {
                     $q->where($q->qualifyColumn('id'), '<', $this->cursorId);
                 }
--- a/app/Libraries/NotificationsBundle.php
+++ b/app/Libraries/NotificationsBundle.php
@@ -76,33 +76,27 @@ class NotificationsBundle
             return;
         }
 
-        $notificationModel = new Notification();
-        $userNotificationModel = new UserNotification();
         $query = $this
             ->user
             ->userNotifications()
-            ->select($userNotificationModel->qualifyColumn('*'))
             ->hasPushDelivery()
-            ->join(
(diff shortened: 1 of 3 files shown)


app/Libraries/NotificationsBundle.php
```diff
@@ -76,23 +76,27 @@ private function fillStacks(string $objectType, int $objectId, string $category)
...
+            ->userNotifications()
+            ->hasPushDelivery()
+            ->joinRelation('notification', function ($q) use ($category, $objectId, $objectType) {
+                $q
+                    ->where($q->qualifyColumn('notifiable_type'), $objectType)
+                    ->where($q->qualifyColumn('notifiable_id'), $objectId)
+                    ->whereIn($q->qualifyColumn('name'), Notification::namesInCategory($category))
+                    ->orderByDesc($q->qualifyColumn('created_at'))
+                    ->orderByDesc($q->qualifyColumn('id'));
+                if ($this->cursorId !== null) {
```
notbakaneko, line 90: can do with a newline after the big block above

notbakaneko (changes requested):

notbakaneko (approved):

Issue #9439, opened 2022-10-31
Highest rank pop-up misaligned on mobile


0x84f:
![Screenshot_20221031_170130_org mozilla firefox](https://user-images.githubusercontent.com/24534022/199054104-4e5a8320-4571-481f-bf9e-e30d573cc1ba.jpg)

Tested on latest Firefox and Chrome mobile versions.

PR #11613, opened 2024-11-01
Fix highest rank tooltip being misaligned on mobile

Joehuu:
- Fixes https://github.com/ppy/osu-web/issues/9439

| Before | After |
| --- | --- |
| Screenshot 2024-10-31 at 8 19 26 PM | Screenshot 2024-10-31 at 8 19 18 PM |

Unsure if this is the correct fix. Wanted to keep the hover/click area width at least the label width, so one digits are still easy to hover/click.

### DIFF
--- a/resources/js/components/value-display.tsx
+++ b/resources/js/components/value-display.tsx
@@ -16,9 +16,11 @@ interface Props {
 export default function ValueDisplay({ description, label, modifiers, value }: Props) {
   return (
     <div className={classWithModifiers(bn, modifiers)}>
-      <div className={`${bn}__label`}>{label}</div>
-      <div className={`${bn}__value`}>{value}</div>
-      {description != null && <div className={`${bn}__description`}>{description}</div>}
+      <div style={{ width: `max-content` }}>
+        <div className={`${bn}__label`}>{label}</div>
+        <div className={`${bn}__value`}>{value}</div>
+        {description != null && <div className={`${bn}__description`}>{description}</div>}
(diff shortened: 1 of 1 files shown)


resources/js/components/value-display.tsx
```diff
@@ -16,9 +16,11 @@ interface Props {
 export default function ValueDisplay({ description, label, modifiers, value }: Props) {
   return (
     <div className={classWithModifiers(bn, modifiers)}>
-      <div className={`${bn}__label`}>{label}</div>
-      <div className={`${bn}__value`}>{value}</div>
-      {description != null && <div className={`${bn}__description`}>{description}</div>}
+      <div style={{ width: 'max-content' }}>
```
nanaya, line 19: this breaks description area (like in kudosu section). just add `align-items: flex-start` to the `value-display` class instead.

nanaya (changes requested):

resources/js/components/value-display.tsx
Joehuu: That makes the hover/click area confined to the rank (bad for `-` and one digits). I've moved the description out of the new div.

nanaya: I see...? Although might as well position it at the bottom left instead (with `data-tooltip-position`).

Also, inlining css is generally avoided.

Joehuu:
Superseded by https://github.com/ppy/osu-web/pull/11836.

PR #11615, opened 2024-11-01
Allow Lazer replay download via client credential grant

### LINKED ISSUE #11579: Allow replay download with client credentials token
im trying to do an app for getting the top 10 monthly plays(including replays) of my country, but i do not seem to be able to get lazer replays download with client credential flow (public scope), can it be added?

minisbett:
Resolves #11579 

This adds the ability for client-credential authorization to download Lazer replays.

I haven't worked on osu-web ever before and I'm not sure if this is how it's done; I [asked on the Discord](https://discord.com/channels/188630481301012481/188630616286167041/1299056552501579859) and talked to nanaya and from what I understood this should be everything needed.

Let me know if there's anything wrong.


app/Http/Controllers/ScoresController.php
```diff
@@ -22,6 +22,7 @@ public function __construct()
 
         $this->middleware('auth', ['except' => [
             'show',
+            'download',
```
nanaya, line 25: this will cause the files to be downloadable without logging in on web which I don't think is an intended side effect.

also the current function assumes current user exists and thus it will throw error when the user is null.

the function needs to be adjusted to handle those cases.

nanaya (changes requested):

app/Http/Controllers/ScoresController.php
minisbett: That is basically what I was asking on the Discord, since I don't know how I would 1. allow client credential auth and 2. check for such auth to not run the code getting the resource owner and increasing the view count. I couldn't quite figure out how to based on what you said.

nanaya: if you look at the actual function, there's a variable which stores the current user which will be null if accessed through web without logging in or client credentials.

As for api vs web check, there's a function being called in the function which should be relatively obvious on what it does.

minisbett: If accessed through client credential auth, shouldn't the user still be null? Since client credential does not have a resource owner. Because of that I thought there's nothing to do there.

minisbett: In that case, I don't seem to understand what code even disallows client-credential auth. `is_api_request()` is true so it won't redirect. Having to handle nullability on the user object makes sense

nanaya: it's this auth middleware. but as I mentioned above, the download function needs to be adjusted to handle the change this caused

nanaya (changes requested):
still need the check for guest user on non-api access

minisbett:
> still need the check for guest user on non-api access

yea I figured, me and someone else we trying to setup a development environment but ultimately failed due to a lack of documentation or troubleshooting or whatever. I will look into it soon

minisbett:
> still need the check for guest user on non-api access

We tried it locally and.. everything seems to work?

1. You can download with code flow & client credential auth
2. You can download on the website when logged in
3. You cannot download via api *nor* via osu.ppy.sh/scores/id/download (the website download button) if not logged in

I'm lost on what's missing here.

nanaya:
try going to the score page first and click the download button. what you saw probably was the referer check

minisbett:
> try going to the score page first and click the download button. what you saw probably was the referer check

`$shouldRedirect = !is_api_request() && (!from_app_url() || \Auth::user() === null);` does this make sense? If from the app url (which I suppose is download via the website) and guest user?

minisbett:
Oh, it does do that in the current state of the PR. If not logged in, clicking on the download button brings up the login pop-up.

nanaya:
it's because of additional javascript handling on the button itself. It won't work if the link is anywhere else like in forum etc.

minisbett:
I hope this should be right

app/Http/Controllers/ScoresController.php
```diff
@@ -42,6 +44,10 @@ private static function parseIdOrFail(string $id): int
 
     public function download($rulesetOrSoloId, $id = null)
     {
+        if(from_app_url() && \Auth::user() === null) {
```
nanaya, line 47: move and use the currentUser variable, and the api/web check function is the other one (it's right in the name)

app/Http/Controllers/ScoresController.php
```diff
@@ -42,6 +44,10 @@ private static function parseIdOrFail(string $id): int
...
+            throw new AuthorizationException('User is not logged in.');
```
nanaya, line 48: authentication exception, not authorization

nanaya (changes requested):

app/Http/Controllers/ScoresController.php
minisbett: But isn't from_app_url() checking if called via the website? That's what we wanna check for, no? Like, the osu.ppy.sh/scores/id/download url. The API endpoint /api/v2/scores/id/download works as expected and fails if not authorized

minisbett: I hope I finally understood how things work, thanks for being patient with me and giving me the chance to learn this 😅

app/Http/Controllers/ScoresController.php
```diff
@@ -42,6 +44,11 @@ private static function parseIdOrFail(string $id): int
 
     public function download($rulesetOrSoloId, $id = null)
     {
+        $currentUser = \Auth::user();
+        if(is_api_request() && $currentUser === null) {
```
nanaya, line 48: "if api request and guest user"?

minisbett: Didn't you just tell me to use the other method? (is_api_request instead of from_app_url)

nanaya: yes but is it "if api request" or "if not api (aka web) request"?

minisbett: Or do you mean I should negate the is_api_request check

nanaya (changes requested):
please fix the lint and test errors ([see this to update the route test file](<https://github.com/ppy/osu-web/pull/11629#issuecomment-2461483618>))

nanaya (approved):

wyvrtn:
Yay! I should also probably start doing some development on this soon to add more features...
BTW, this will be documented soon right?

minisbett:
The API docs are automatically generated from the laravel API controllers

Also regarding this PR, maybe I shouldn't touch web stuff and stick to client stuff

wyvrtn:
> The API docs are automatically generated from the laravel API controllers
> 
> Also regarding this PR, maybe I shouldn't touch web stuff and stick to client stuff

Yeah I'm looking at this php and I still don't understand how any of this makes sense
Feels like it's running on pure magic (Is there a framework involved?)
Anyways, the docs have way too many optional parameters and parameters with no types
I have no idea half the time what the endpoint wants

PR #11616, opened 2024-11-01
Async script loading

nanaya:
Properly registered with filename and map and stuff instead of being unmapped inlined stuff.

- [x] depends on #11600

### DIFF
--- a/resources/js/core/turbolinks-reload.ts
+++ b/resources/js/core/turbolinks-reload.ts
@@ -3,51 +3,50 @@
 
 export default class TurbolinksReload {
   private readonly loaded = new Set<string>();
-  private readonly loading = new Map<string, JQuery.jqXHR<void>>();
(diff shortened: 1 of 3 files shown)


resources/js/core/turbolinks-reload.ts
```diff
@@ -22,7 +22,7 @@ export default class TurbolinksReload {
 
   load(src: string) {
     if (this.loaded.has(src) || this.loading.has(src)) {
-      return this.loading.get(src)?.[0];
+      return this.loading.get(src)?.[0] ?? Promise.resolve();
```
notbakaneko, line 25: wouldn't it be `Promise.reject()`?

nanaya: this indicates the file has been loaded successfully

Issue #8300, opened 2021-11-05
Add the ability for users to add custom tags to beatmaps


apollo-dw:
Keeping this super short and sweet, there were talks about [adding a guideline to ranking criteria to tag the "style" of a beatmap](https://osu.ppy.sh/community/forums/topics/1208305), however there were concerns with standardizing this sort of thing. 

I propose delegating this task to players. Imagining a system similar to what Steam has (screenshot below), there will be no need to define terms - players will know that stuff best. My proposal is strictly for terms that would describe a map, since I'm not sure whether user tags would be useful for describing other things...

![image](https://user-images.githubusercontent.com/83023433/140540938-a24e2596-3474-48f3-b82a-dbaa9a1a2d59.png)

cl8n:
for the RC thread, the perceived issue was that direction/organisation would be required to use the tags meaningfully, not that it needed standardization from the beginning. I think the same issue applies to this

also
- comparison to steam assumes that people have common sets of words to describe osu gameplay, I don't think this is true beyond the simplest of things
- I don't see how u could implement the restriction to gameplay tags
- moderating this seems impossible

seems cool but like the RC thing I don't think it's practical to implement (yet!)

Hiviexd:
i think having something like this could greatly benefit searchability for players and mappers alike. to solve the moderation/consistency issue, we can have players submit tags for "approval" (a 5-second task that can be done by NAT/GMT), and have those approved tags to be usable on any map by players.

if a map has misleading user tags, that can be handled through the usual beatmap report where any moderator can fix it accordingly.

Walavouchey:
https://omdb.nyahh.net/descriptors/ list of tags used by a community beatmap database site

apollo-dw:
> peppy mentioned making the list predefined instead of user-defined in today's catch-up meeting. in that case the nat or bn would likely be the ones to manage the tag list and fix wrong tags whenever needed

just to give more detail here, the "user-defined" tags are added by virtue of proposals that are critiqued and voted on by users -- they only pass if they have a majority

i would love to elect myself to look after the tags/descriptor list on osu-web if this goes through, as someone whos bn and already oversees omdb... and has cared about this feature since 2021 according to the issue open date 😅

Hiviexd:
the general consensus from the NAT meeting about how this should work was:

- start with incorporating the tags from https://omdb.nyahh.net/descriptors/
- implement the system in a way that makes it easy to add/edit tags via `osu-web` PRs
[...]

PR #11629, opened 2024-11-06
Add basic beatmap tags backend

tsunyoku:
Implements the initial proposal found in #8300.

I've pre-emptively added a `priv_check` for beatmap tags since beatmap tags could be maliciously added to beatmaps and we may want to restrict users from using the feature. I've also added logic to ensure the user has a score on the beatmap before they can add a tag to try and improve quality of beatmap tags (mentioned on the proposal also).

Initially have forced auth on the store & destroy endpoints so they're effectively lazer only API endpoints.


nanaya:
there's an artisan command to generate that file btw 👀 
```bash
artisan route:convert --to-json=tests/api_routes.json
```

tsunyoku:
I had a feeling there was a command for it 🤦🏻‍♂️

peppy:
Would be good to keep momentum on this one, users will appreciate.

database/factories/BeatmapTagFactory.php
```diff
@@ -0,0 +1,27 @@
...
+class BeatmapTagFactory extends Factory
+{
+    protected $model = BeatmapTag::class;
+
+    public function definition(): array
+    {
+        return [
+            'beatmap_id' => fn () => Beatmap::factory(),
+            'tag_id' => fn () => Tag::factory(),
+            'user_id' => fn () => User::factory(),
```
notbakaneko, line 24: using `factory()` doesn't need to be in a closure.

app/Models/BeatmapTag.php
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * @property Beatmap $beatmap
+ * @property int $beatmap_id
+ * @property int $tag_id
+ * @property User $user
+ * @property int $user_id
+ */
+class BeatmapTag extends Model
+{
```
notbakaneko, line 16: needs to be set for composite keys
```php
protected $primaryKeys = ['beatmap_id', 'tag_id', 'user_id'];
```

notbakaneko (changes requested):
getting the ones I already noted down out first

nanaya:
getting the tags of a beatmap seems like a potential nightmare if it got tagged by thousands of people...?

peppy:
> getting the tags of a beatmap seems like a potential nightmare if it got tagged by thousands of people...?

If we are actually allowing custom/freeform tags in the first place, we'd only show top n tags which are tagged by x users, ordered by count of user votes for tag descending.

nanaya:
getting those count - which also involves excluding restricted users - and then sorting them doesn't seem exactly fast though

peppy:
Probably just a table with the raw count number cached to it?

notbakaneko:
I think we'd want to cache the tags being shown on the beatmap or have something update them on a schedule; it doesn't seem like something that needs to be realtime

tsunyoku:
I've added caching for both tags and beatmap tags - hopefully this should resolve any concerns mentioned above.

> Also, I think the initial idea is for the available tags to be fixed, initially?
Yes, there's intentionally no `approved` or similar flag on the `tags` table as they're supposed to only be fixed rather than user-customisable.

app/Http/Controllers/BeatmapTagsController.php
```diff
@@ -0,0 +1,101 @@
...
+
+        $beatmap = Beatmap::find($beatmapId);
+        $tag = Tag::find($tagId);
+
+        abort_if($beatmap === null, 422, "specified beatmap couldn't be found");
+        abort_if($tag === null, 422, "specified tag couldn't be found");
+
+        $user = \Auth::user();
+    
+        $userHasScore = Score::where('user_id', $user->getKey())->where('beatmap_id', $beatmapId)->exists();
```
notbakaneko, line 79: `$user->soloScores()->where('beatmap_id', $beatmapId)->exists()`

app/Http/Controllers/BeatmapTagsController.php
```diff
@@ -0,0 +1,101 @@
...
+        $hasExistingBeatmapTag = $user->beatmapTags()->where('beatmap_id', '=', $beatmapId)->exists();
+
+        if (!$hasExistingBeatmapTag) {
+            priv_check('BeatmapTag', $beatmap)->ensureCan();
+
+            try {
+                (new BeatmapTag(['tag_id' => $tagId]))
+                    ->user()->associate($user)
+                    ->beatmap()->associate($beatmap)
+                    ->saveOrExplode();
```
notbakaneko, line 91:
```php
BeatmapTag::create([
    'beatmap_id' => $beatmapId,
    'tag_id' => $tagId,
    'user_id' => $user->getKey(),
])
```

or add a relation on `Tag` and use `$tag->beatmapTags()->create(...)`

app/Http/Controllers/TagsController.php
```diff
@@ -0,0 +1,32 @@
...
+    public function __construct()
+    {
+        parent::__construct();
+
+        $this->middleware('require-scopes:public');
+    }
+
+    public function index()
+    {
+        $tags = \Cache::remember(
```
notbakaneko, line 22: Use `cache_remember_mutexed`.
Also it's probably fine to store in redis for now but depending on how large the list is and how frequently it's accessed, it might be better stored as a local cache singleton.

app/Http/Controllers/BeatmapTagsController.php
```diff
@@ -0,0 +1,101 @@
...
+                ->join('tags', 'beatmap_tags.tag_id', 'tags.id')
+                ->where('beatmap_id', '=', $beatmapId)
+                ->whereHas('user', function ($userQuery) {
+                    $userQuery->default();
+                })
+                ->groupBy('tag_id')
+                ->orderBy('tag_count', 'desc')
+                ->limit(50)
+                ->get()
+                ->all(),
```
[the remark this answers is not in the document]
notbakaneko: Put it in a function in one of the models; also `joinRelation` can be used like in https://github.com/ppy/osu-web/blob/019fdc3b40bac7538f401a22ae00e7446e5acbf2/app/Libraries/NotificationsBundle.php#L83
and use `qualifyColumn` instead of hardcoding the table name in the join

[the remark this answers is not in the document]
tsunyoku: what exactly is `joinRelation` going to help with here? i'm only joining `tags` to access the name of the tag, it feels like added complexity although with only that example to go by maybe i'm missing something

app/Http/Controllers/BeatmapTagsController.php
```diff
@@ -0,0 +1,96 @@
...
+    {
+        $tagId = get_int(request('tag_id'));
+
+        $beatmap = Beatmap::findOrFail($beatmapId);
+        $tag = Tag::findOrFail($tagId);
+
+        $user = \Auth::user();
+
+        $userHasScore = $user->soloScores()->where('beatmap_id', $beatmapId)->exists();
+        abort_if(!$userHasScore, 400, 'you must set a score on a beatmap to add a tag');
```
notbakaneko, line 76: Looks like it should be part of the `priv_check` (and add a localization key for the message)

config/osu.php
```diff
@@ -312,4 +312,8 @@
         'country_performance_user_count' => intval(env('COUNTRY_PERFORMANCE_USER_COUNT', 1000)),
         'country_performance_weighting_factor' => floatval(env('COUNTRY_PERFORMANCE_WEIGHTING_FACTOR', 0.99)),
     ],
+    'tags' => [
+        'tags_cache_interval' => 60 * intval(env('TAGS_CACHE_INTERVAL', 60)), // in minutes, converted to seconds
+        'beatmap_tags_cache_interval' => 60 * intval(env('BEATMAP_TAGS_CACHE_INTERVAL', 60)), // in minutes, converted to seconds
```
notbakaneko, line 317: We're standardizing on `_duration` suffix for cache durations and the keys should be alphabetically ordered (I just noticed the existing keys at the bottom are in the wrong place but that's a different problem).

Also, use `get_int(env(ENVKEY)) ?? 60` instead of `intval`.

app/Models/Tag.php
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * @property int $id
+ * @property string $name
+ * @property string $description
+ * @property-read Collection<BeatmapTag> $beatmapTags
+ */
+class Tag extends Model
+{
+    protected $table = 'tags';
```
notbakaneko, line 21: `$table` doesn't need to be specified.

app/Models/BeatmapTag.php
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * @property-read Beatmap $beatmap
+ * @property int $beatmap_id
+ * @property int $tag_id
+ * @property-read User $user
+ * @property int $user_id
+ */
+class BeatmapTag extends Model
+{
+    protected $table = 'beatmap_tags';
```
notbakaneko, line 19: `$table` doesn't need to be specified if the table and model already follow laravel's magic naming convention.

notbakaneko (changes requested):
I haven't gotten to the tests yet 🫠

tests/Controllers/BeatmapTagsControllerTest.php
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public function testStore(): void
+    {
+        $user = User::factory()->create();
+        Score::factory()->create(['beatmap_id' => $this->beatmap->getKey(), 'user_id' => $user->getKey()]);
+
+        $this->expectCountChange(fn () => BeatmapTag::count(), 1);
+
+        $this->actAsScopedUser($user, ['*']);
```
notbakaneko, line 46: `['*']` is already the default.

tests/Controllers/BeatmapTagsControllerTest.php
```diff
@@ -0,0 +1,83 @@
...
+                    ->where('beatmap_tags.0.tag_id', $this->tag->getKey())
+                    ->where('beatmap_tags.0.name', $this->tag->name)
+                    ->where('beatmap_tags.0.tag_count', 1)
+                    ->etc());
+    }
+
+    public function testStore(): void
+    {
+        $user = User::factory()->create();
+        Score::factory()->create(['beatmap_id' => $this->beatmap->getKey(), 'user_id' => $user->getKey()]);
```
notbakaneko, line 42: Not a big deal but this can also be
`$user = User::factory()->has(Score::factory()->state(['beatmap_id' => $this->beatmap]), 'soloScores')->create();`

tests/Controllers/BeatmapTagsControllerTest.php
```diff
@@ -0,0 +1,83 @@
...
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->tag = Tag::factory()->create();
+        $this->beatmap = Beatmap::factory()->create();
+        $this->beatmapTag = BeatmapTag::factory()->create([
+            'tag_id' => $this->tag->getKey(),
+            'beatmap_id' => $this->beatmap->getKey(),
```
notbakaneko, line 80: The model or factory can be used directly when creating via the factory: `'tag_id' => $this->tag`

tests/Controllers/TagsControllerTest.php
```diff
@@ -0,0 +1,38 @@
...
+                    ->where('tags.0.id', $this->tag->getKey())
+                    ->etc());
+    }
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->tag = Tag::factory()->create();
+    }
```
notbakaneko, line 37: Kind of unnecessary for single test 🤔

notbakaneko (changes requested):

tests/Controllers/TagsControllerTest.php
tsunyoku: Initially added this thinking there'd be other routes that might need a tag in these tests, but yeah not necessary currently.

notbakaneko (approved):

PR #11635, opened 2024-11-08
Include user colour icon as part of the link

nanaya:
Seems weird it's not part of the link.

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -835,7 +835,7 @@ function forum_user_link(int $id, string $username, string|null $colour, int|nul
         'style' => user_color_style($colour, 'background-color'),
     ]);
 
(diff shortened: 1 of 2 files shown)


app/helpers.php
```diff
@@ -835,12 +835,12 @@ function forum_user_link(int $id, string $username, string|null $colour, int|nul
         'style' => user_color_style($colour, 'background-color'),
     ]);
 
-    $link = link_to_user($id, $username, null, []);
+    $link = link_to_user($id, blade_safe($icon.e($username)), null, []);
```
notbakaneko, line 838: ngl, this part reads weird to me 🫠 
`blade_safe($icon.e($username))`

notbakaneko (approved):

PR #11644, opened 2024-11-13
Fix legacy match score rank value

### LINKED ISSUE #11591: osu api always show "F" rank from get_match
Code
Code_mo1cjrEchp
Result
WindowsTerminal_x77y3RhNZK
Match link: https://osu.ppy.sh/community/matches/115977090
Score: https://osu.ppy.sh/scores/3744935251

nanaya:
Resolves #11591.

This recalculates the value every time but I think it's already done for legacy score rank anyway, so 🤷

### DIFF
--- a/app/Models/Traits/Scoreable.php
+++ b/app/Models/Traits/Scoreable.php
@@ -74,7 +74,7 @@ trait Scoreable
 
     public function recalculateRank(): void
     {
-        if (!$this->passed) {
+        if (!$this->pass) {
             $this->rank = 'F';
 
             return;
--- a/app/Models/LegacyMatch/Score.php
+++ b/app/Models/LegacyMatch/Score.php
@@ -70,9 +70,18 @@ class Score extends Model
         return $this->_getEnabledMods($value | ($this->game->getAttributes()['mods'] ?? 0));
     }
 
+    public function getPassedAttribute(): bool
+    {
+        return $this->pass;
+    }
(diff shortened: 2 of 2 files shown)


app/Models/Traits/Scoreable.php
```diff
@@ -74,7 +74,7 @@ public function accuracy(): float
 
     public function recalculateRank(): void
     {
-        if (!$this->passed) {
+        if (!$this->pass) {
```
notbakaneko, line 77: `Solo\Score` doesn't have a `pass`; actually, the function should probably be moved out of the trait since it's only used by legacy scores.

notbakaneko (changes requested):

app/Models/Traits/Scoreable.php
nanaya: but Solo\Score doesn't use this trait? 🤔

notbakaneko: does it not? 🤔

notbakaneko: Oh wait, I mean because the naming in `Solo\Score::makeLegacyEntry` with the variable being `$score` and the method being named `recalculateRank` makes it confusing and the method is only used in legacy scores now.

I guess that can be a separate issue 🤷

notbakaneko (approved):

PR #11645, opened 2024-11-13
Update laravel framework

nanaya:
It's been a while so might as well do this.

There's still a lot of other outdated libraries 😬


notbakaneko (approved):
Nothing seems broken so far 👀

PR #11647, opened 2024-11-13
Do a quick build on dev start

nanaya:
This adds about 4s on my system with fully updated images 🤔 

```
$ time docker compose build
...
real    0m3.557s
user    0m0.146s
sys     0m0.117s
```


notbakaneko (approved):
I'd just be running `docker compose up` directly when nothing needs re-setting up 🤷

PR #11651, opened 2024-11-14
Try fixing user country on login

### LINKED ISSUE #11592: Attempt to assign country more frequently when initial assignment failed
I guess this is happen until they have atleast 1 pp - https://osu.ppy.sh/users/35107890/osu

nanaya:
Resolves #11592?


peppy (approved):
Seems logical

notbakaneko (approved):

PR #11655, opened 2024-11-15
Allow silenced user to play in playlist

nanaya:
I think that's all it needed? 🤔 

Related: ppy/osu#30627


peppy (approved):

notbakaneko (approved):
seems fine 🤔

PR #11656, opened 2024-11-15
Clean up authorize assertions

nanaya:
- reject null user in some cases
- don't return anything

### DIFF
--- a/app/Singletons/OsuAuthorize.php
+++ b/app/Singletons/OsuAuthorize.php
@@ -976,6 +976,7 @@ class OsuAuthorize
     {
         $prefix = 'chat.';
 
+        $this->ensureLoggedIn($user);
         $this->ensureSessionVerified($user);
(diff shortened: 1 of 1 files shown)


app/Singletons/OsuAuthorize.php
```diff
@@ -2040,30 +2041,24 @@ public function ensureLoggedIn(?User $user, string $prefix = ''): void
...
 
         if ($user->isSilenced()) {
             throw new AuthorizationCheckException($prefix.'silenced');
         }
-
-        return 'ok';
     }
 
     /**
      * @param User|null $user
```
notbakaneko, line 2056: also needs updating, or just removed since it's not adding any new info, same with the other one

notbakaneko (changes requested):
Maybe now is a good time to make the methods private as well

notbakaneko (approved):

PR #11661, opened 2024-11-15
Fix incorrect form parameter

nanaya:
That must be a leftover before the flow is changed 🤨 

Drafting as there might be others with similar bug.


notbakaneko:
The button I got stuck on reloading before seems to work fine now, so it was probably something else 🤔

notbakaneko (approved):

PR #11662, opened 2024-11-16
Fix parsing page parameter of ranking page

nanaya:
So I did miss at least one place when updating to laravel builtin function <_<

This is a rare edge case and must be manually typed... I think.


notbakaneko (approved):
Isn't this a case of the `?? 1` being in the wrong place ( ﾟ ヮﾟ)

PR #11667, opened 2024-11-19
Add endpoint for closing playlists within grace period of creation

bdach:
Intended for use in https://github.com/ppy/osu/issues/30401


tests/Controllers/Multiplayer/RoomsControllerTest.php
```diff
@@ -429,6 +429,100 @@ public function testJoinWithPassword()
         $this->assertSame($initialUserChannelCount + 1, UserChannel::count());
     }
 
+    public function testDestroy()
+    {
+        $token = Token::factory()->create(['scopes' => ['*']]);
+        $start = now();
+        $end = $start->addMinutes(60);
```
nanaya, line 436: the object created by `now()` mutates itself so the start and end are both the same. Either `clone()` it first or use `Carbon\CarbonImmutable`

tests/Controllers/Multiplayer/RoomsControllerTest.php
```diff
@@ -429,6 +429,100 @@ public function testJoinWithPassword()
         $this->assertSame($initialUserChannelCount + 1, UserChannel::count());
     }
 
+    public function testDestroy()
+    {
+        $token = Token::factory()->create(['scopes' => ['*']]);
```
nanaya, line 434: there's `$this->actAsScopedUser($room->user)`

tests/Controllers/Multiplayer/RoomsControllerTest.php
```diff
@@ -429,6 +429,100 @@ public function testJoinWithPassword()
...
+        $start = now();
+        $end = $start->addMinutes(60);
+        $room = Room::factory()->create([
+            'user_id' => $token->user,
+            'starts_at' => $start,
+            'ends_at' => $end,
+            'type' => Room::PLAYLIST_TYPE,
+        ]);
+        $end = $room->ends_at; // creation truncates fractional second part, so refetch here
```
nanaya, line 443: technically the truncation happens during assignment

app/Http/Controllers/Multiplayer/RoomsController.php
```diff
@@ -228,4 +228,17 @@ public function store()
             return error_popup($e->getMessage(), $e->getStatusCode());
         }
     }
+
+    public function destroy($id)
```
nanaya, line 232: the functions are sorted alphabetically

nanaya (changes requested):

nanaya:
oh and here's a protip for the routing json thing https://github.com/ppy/osu-web/pull/11629#issuecomment-2461483618

tsunyoku:
> oh and here's a protip for the routing json thing [#11629 (comment)](https://github.com/ppy/osu-web/pull/11629#issuecomment-2461483618)

at this point I feel like this should be placed somewhere as it seems to be referenced every time someone makes a PR which adds/changes a route 😆

could this not be automated via CI or something?

nanaya:
> at this point I feel like this should be placed somewhere as it seems to be referenced every time someone makes a PR which adds/changes a route 😆
> 
> could this not be automated via CI or something?

I think the point is to make us aware about api changes but yeah it can be better documented at least.

tests/Controllers/Multiplayer/RoomsControllerTest.php
bdach: I'm not sure how to use that. I assumed it'd be something like

```diff
diff --git a/tests/Controllers/Multiplayer/RoomsControllerTest.php b/tests/Controllers/Multiplayer/RoomsControllerTest.php
index 7cd7096c82..ae457dcb9d 100644
--- a/tests/Controllers/Multiplayer/RoomsControllerTest.php
+++ b/tests/Controllers/Multiplayer/RoomsControllerTest.php
@@ -431,11 +431,9 @@ class RoomsControllerTest extends TestCase
 
     public function testDestroy()
     {
-        $token = Token::factory()->create(['scopes' => ['*']]);
         $start = now();
         $end = $start->clone()->addMinutes(60);
         $room = Room::factory()->create([
-            'user_id' => $token->user,
             'starts_at' => $start,
             'ends_at' => $end,
             'type' => Room::PLAYLIST_TYPE,
@@ -443,8 +441,8 @@ class RoomsControllerTest extends TestCase
         $end = $room->ends_at; // creation truncates fractional second part, so refetch here
         $url = route('api.rooms.destroy', ['room' => $room]);
 
+        $this->actAsScopedUser($room->user, ['lazer']);
         $this
-            ->actingWithToken($token)
             ->delete($url)
             ->assertSuccessful();
 

```

but trying that, it barfs at me with `App\Exceptions\InvalidScopeException: * is not allowed with Client Credentials`.

app/Http/Controllers/Multiplayer/RoomsController.php
```diff
@@ -20,6 +20,13 @@ public function __construct()
         $this->middleware('require-scopes:public', ['only' => ['index', 'leaderboard', 'show']]);
     }
 
+    public function destroy($id)
+    {
+        $room = Room::findOrFail($id);
+        $room->endGame(\Auth()->user());
```
nanaya, line 26: it's a different thing 😔  `::user()`

tests/Controllers/Multiplayer/RoomsControllerTest.php
nanaya: oh wow the model doesn't have user relation defined (and thus it passes `null` for the user) o_O

adding it is probably a good idea...

app/Http/Controllers/Multiplayer/RoomsController.php
bdach: i don't even wanna know

bdach: oh apparently i had the braces in there which i should not have? i give up

nanaya: it's a different way to access the same thing but slightly faster because the other method (`auth()`) does additional processing for some unknown reason

tests/Controllers/Multiplayer/RoomsControllerTest.php
bdach: I think it might be `->host` actually. Seems to work locally at least.

https://github.com/ppy/osu-web/blob/1fbc73baa8e8be6759b1cc4f4ab509d8ae53a165/app/Models/Multiplayer/Room.php#L557

app/Http/Controllers/Multiplayer/RoomsController.php
```diff
@@ -20,6 +20,13 @@ public function __construct()
         $this->middleware('require-scopes:public', ['only' => ['index', 'leaderboard', 'show']]);
     }
 
+    public function destroy($id)
+    {
+        $room = Room::findOrFail($id);
+        $room->endGame(\Auth::user());
```
nanaya, line 26:
```suggestion
        Room::findOrFail($id)->endGame(\Auth::user());

```

nanaya (approved):

Issue #7302, opened 2021-03-11
Store version history of beatmaps


VINXIS:
Having version history for beatmaps  (the folders themselves and/or the .osu files) would be greatly beneficial for a ton of things, just to name a few:
- Being able to access previous versions lost when looking for things that were changed in previous versions of your beatmap that may have been lost
- Being able to more easily see what was changed between the different versions of the beatmap
- Possibly streamline the process of changing things in a beatmap while it's in Qualified because of the 2nd point; allow the beatmap to update by the mapper and create a new temporary state for the beatmap -> allow the member of the BNG to check the changes of the beatmap -> member of the BNG confirms the change and is merged into the "official" version of the beatmap and leaderboard resets
- Allow players to be able to play an older version of a beatmap if they wish to
- Help mitigate issues presented like https://osu.ppy.sh/community/forums/topics/1268711 or https://osu.ppy.sh/community/forums/topics/1068618 for example

Would be beneficial if we could see the changes in a .osu file + song folder, and maybe also having a short description for "commits" would be great for osu! mapping, similar to the git process of changes -> commit -> push and with changes in Qualified being more treated like a "branch/fork" that needs a member of the BNG to push into the "main/master" branch, though it would be a lot of work to implement this

VINXIS:
Was also talked about here https://github.com/ppy/osu-web/issues/1953 in 2018

peppy:
This is scheduled as part of a large BSS update with no current timeline. It is already mostly implemented in the lazer client, so you will likely see this functionality available in-client before online.

VINXIS:
I'd like to suggest if possible placing this at a higher priority (after seeing the response [here](https://osu.ppy.sh/community/forums/posts/7978812)) especially for online-side version history handling (moreso than in-client) since online-side version history will alleviate a ton of issues stated in this issue op, some of which are considered as "large" issues such as the third and fifth points in the op

peppy:
As a stop gap measure, would making the last n versions available (just as osz downloads, not properly listing what changes were made) be of benefit?

If you were to get a minimum viable product what would that be? I'd rather try and hack something in rather than attempt to do thing the right way, as the right way is going to take considerable time still.

dorothy3242:
Tools that can check changes between versions of maps exist with mapset verifier, so working with naxess to work into making that a native feature would be beneficial. Because of that having previous oszs would be useful already.

VINXIS:
[...]

PR #11668, opened 2024-11-19
Add tables for beatmap versioning support

bdach:
Related: https://github.com/ppy/osu-web/issues/7302

Schema as proposed in https://github.com/ppy/osu-server-beatmap-submission/pull/1. Can be negotiated further here as required.


nanaya:
I thought we're doing away with foreign keys? are we not anymore?

nanaya:
also let's not prefix the tables with `osu_`...

bdach:
You tell me, I have no idea. As I talked in that other PR everyone does db differently and I'm only learning the ropes here. Can delete the FKs if you think that's necessary, but I'd like to leave at least indices on these columns.

peppy:
I mentioned the foreign keys on the other PR. We should avoid them because it limits some kind of replication stuff in some weird way. Also it's just a PITA to deal with when FK fails, I'd rather have to clean up orphaned things myself and not have to worry about super breakage when things go wrong.

Generally we just convert these to a local INDEX / UNIQUE INDEX

database/migrations/2024_11_19_130731_create_beatmap_versioning_tables.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+use Illuminate\Database\Migrations\Migration;
```
nanaya, line 5: a blank line between license and code would be nice

database/migrations/2024_11_19_130731_create_beatmap_versioning_tables.php
```diff
@@ -0,0 +1,56 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
```
nanaya, line 8: I suppose this comes from an ide. Not harmful but also not necessary as those are already aliased to just `DB` and `Schema` respectively

nanaya (changes requested):

database/migrations/2024_11_19_130731_create_beatmap_versioning_tables.php
```diff
@@ -0,0 +1,56 @@
...
+            `sha2_hash` BINARY(32) NOT NULL,
+            `file_size` INT UNSIGNED NOT NULL,
+
+            UNIQUE (`sha2_hash`)
+        )');
+
+        Schema::create('beatmapset_versions', function (Blueprint $table) {
+            $table->bigIncrements('version_id');
+            $table->mediumInteger('beatmapset_id')->unsigned();
+            $table->dateTime('created_at')->default(DB::raw('NOW()'));
```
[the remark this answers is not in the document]
bdach: Yeah see date types is another one of these that everything and everyone everywhere does differently. Two-minute research seems to indicate timestamp is slightly different (and probably better) mostly because it seems to be storing UTC from what I can gather from mysql docs, so will change.

database/migrations/2024_11_19_130731_create_beatmap_versioning_tables.php
bdach: I'm not even sure whether this was phpstorm or artisan at this point, but will trim

database/migrations/2024_11_19_130731_create_beatmap_versioning_tables.php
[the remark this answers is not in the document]
nanaya: yeah the benefit is it's stored in utc and the main drawback is it'll explode in 14 years (search mysql 2038)

nanaya (approved):

PR #11670, opened 2024-11-21
Don't override InputContainer error colour when focused

notbakaneko:
The error highlight should still remain active when the field is focused. Left the hover highlight alone since it seems logical it should follow what's being hovered but not yet focused? 🤔

### DIFF
--- a/resources/css/bem/input-container.less
+++ b/resources/css/bem/input-container.less
@@ -2,6 +2,8 @@
 // See the LICENCE file in the repository root for full licence text.
 
 .input-container {
+  @top: input-container;
+
   --label-font-size: @font-size--normal;
   --label-line-height: 1.25;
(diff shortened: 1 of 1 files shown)


resources/css/bem/input-container.less
```diff
@@ -32,6 +34,10 @@
 
   &:focus-within {
     --input-border-colour: var(--input-border-focus-colour);
+
+    .@{top}--error& {
+      --input-border-colour: var(--input-border-error-colour);
```
nanaya, line 39: override `--input-border-focus-colour` in `--error` modifier instead? Also maybe have a lighter/darker colour than the normal error one?

nanaya (changes requested):

nanaya (approved):

PR #11673, opened 2024-11-21
Make test user session/token helpers chainable

nanaya:
I don't think there's any specific reason they shouldn't be chainable. Most of the use is immediately followed by the related request 🤔


bdach:
Completely unqualified opinion but upon recent use of these I did find it a little weird that these weren't chainable.

notbakaneko (approved):

Issue #11636, opened 2024-11-08
Allow pulling only open multiplayer lobbies from ``/matches`` endpoint


venix12:
Currently just goes by id, no matter if the room is still open or not, would be helpful to add an option to pull active rooms only.

peppy:
Seems like an easy/quick win. Also an easy one for someone to grab?

stilettocode:
> Seems like an easy/quick win. Also an easy one for someone to grab?

will give it a shot

PR #11681, opened 2024-11-22
Add `status` to `multiplayer_rooms`

smoogipoo:
Enum to support possible future expansion.

This doesn't contain the full array the client would normally use `( open, open_private, playing, ended )` because the "ended" state is difficult to handle at osu!web's end for playlist rooms.

So `idle` now represents `( open, open_private, ended )` and the client will figure out it from there.

### DIFF
--- a/app/Models/Multiplayer/Room.php
+++ b/app/Models/Multiplayer/Room.php
@@ -27,6 +27,7 @@ use Illuminate\Database\Eloquent\SoftDeletes;
 
 /**
  * @property string $category
+ * @property string $status
  * @property Channel $channel
  * @property int|null $channel_id
  * @property \Carbon\Carbon|null $created_at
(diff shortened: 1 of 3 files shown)


nanaya (approved):

peppy:
As mentioned elsewhere, `open` and `open_private` shouldn't be grouped in the same enum as `playing`/`ended` because you're losing the `private` state for the latter two.

But since the opening post, these have been removed so I guess it's fine.

peppy (approved):

baharsah:
it resolve #11636 ?

smoogipoo:
> This adds the column at the end

I only put it there because I noticed `category` is already there. I imagine you can do a migration to move it around later on?

> But since the opening post, these have been removed so I guess it's fine

The OP is post- that idea. I.e. it's a catch all state for all of those that have to be decided by some other method later on depending on on end date / password.

PR #11684, opened 2024-11-25
Update http string check

nanaya:
One place I missed thanks to its different way of checking.


nanaya:
Combined with other fixes.

PR #11688, opened 2024-11-25
Fix turbo redirect

nanaya:
The previous fix doesn't quite work as turbo just ends up doing a browser native visit and breaks on form action due to 200 status.

### DIFF
--- a/resources/js/setup-turbo.ts
+++ b/resources/js/setup-turbo.ts
@@ -25,7 +25,7 @@ document.addEventListener('turbo:submit-end', (e) => {
 });
 
 document.addEventListener('turbo:before-fetch-response', (e) => {
-  if (!e.detail.fetchResponse.contentType.match(/^text\/osu-turbo-redirect[ ;]*/)) {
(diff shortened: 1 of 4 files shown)


resources/js/setup-turbo.ts
```diff
@@ -24,6 +24,24 @@ document.addEventListener('turbo:submit-end', (e) => {
   }
 });
 
+document.addEventListener('turbo:before-fetch-response', (e) => {
+  if (!e.detail.fetchResponse.contentType.match(/^text\/osu-turbo-redirect[ ;]*/)) {
```
notbakaneko, line 28: `contentType` is null when replacing an existing user-cover-preset or toggling enabled/disabled (and maybe similar buttons?)

notbakaneko (changes requested):

resources/js/setup-turbo.ts
nanaya: well those only need to reload the page, not redirect

PR #11706, opened 2024-12-02
Automatically create beatmap set description topic on edit if missing

bdach:
> [!NOTE]
> Please consider this PR as a request for comments / starting point of discussion. It is by no means intended to be 100% complete, and as such I'm opening it as draft until that initial discussion has ran its course and any potential objections have been addressed.

This is a solution to the problem of the new beatmap submission server not creating a description topic for the set like the old one did. Notably, this solution implies several shortcuts:

- If the main metadata of the set (artist, title) don't change, the forum thread will not be renamed
- The forum thread will be posted in a hardcoded forum and never moved, while old BSS would move the thread between forums 6 and 10 depending on whether the beatmap is pending or WIP. Not doing this saves a whole bunch of busy work.
- There is no actual body in the forum thread OP, contrary to what old submission did ([example](https://osu.ppy.sh/community/forums/topics/2010392?n=1)). The 15 dashes are the minimum content that is required to have the existing post parsing logic work.

The aforementioned simplifications were made with the assumption that subforums 6 and 10 on production were going to be hidden away from users due to not containing any actual content that anyone would ever want to see. (Same thing applies to forum 19, which is graveyard - although old threads pre-modding V2 actually do have useful content there. Not sure what to do about that.)

The *one and only goal* here is that descriptions should be editable for maps submitted using new BSS. Thus I am taking pretty much the same shortcuts that places like the profile page already seem to be.


nanaya:
well this is originally my idea, so 🤷 and yay? 👍 

and I think it doesn't have to reuse the existing forum...?

peppy:
I have no issue with the direction, FWIW.

bdach:
Given there are no objections to general direction I'll undraft I suppose. That said the main question I'd have at this point is which forum ID to put the relevant forum topics in - is there going to be a new forum for them, or are we reusing existing?

peppy:
Following existing expectations should be fine (ie. using https://osu.ppy.sh/community/forums/6 / https://osu.ppy.sh/community/forums/10), unless I'm missing a reason for it to be separated?

bdach:
> unless I'm missing a reason for it to be separated?

Um well nowhere in this diff am I moving the relevant thread from the pending forum to the WIP forum or the other way around as the user changes the beatmap's status from pending to WIP or the other way around. web-10 does this on submission. Doing this would require special handling, likely triggered by `osu-server-beatmap-submission`, likely requiring another LIO endpoint or something. Unless your point is that it's fine to ignore it and just use one of the two...?

peppy:
Moving between the forums is largely a historic thing, and I don't think it needs to happen. We could just choose one (likely the pending forum) and go with it for now.

bdach:
> Right. To clarify on that, moving between the forums is largely a historic thing, and I don't think it needs to happen (the important movements to ranked or to graveyard are handled elsewhere anyways). We could just choose one (likely the pending forum `id:6`) and go with it for now

Switched to this one in https://github.com/ppy/osu-web/pull/11706/commits/6f8dadeacc4c2b4f2f5e64095cb4310e82d10283.

If we're using that one then I guess the other question would be whether I should bother to give the post any actual content. As in, old BSS topics had actual content before the description (example: https://osu.ppy.sh/community/forums/topics/2021687?n=1), while the ones created here don't really bother with that and just do the description part.

nanaya (approved):
Forgot about this 👀 Seems working

PR #11709, opened 2024-12-03
Replace compound primary key on `beatmapset_version_files` with synthetic one

bdach:
An oops from https://github.com/ppy/osu-web/pull/11668.

The primary key was on `(file_id, version_id)`, which looks fine in theory, but in practice I hit a case where I got primary key uniqueness violations, and realised that this setup fails if a beatmap set has two or more files in the package with identical contents but different filenames.

This is the case, since `beatmapset_files` has a column for the files' SHA256, with a uniqueness constraint, so two files with identical content must refer back to the same `beatmapset_files` row.

I considered adding `filename` to the PK but decided against it since it's probably slow and at this point attempting to conjure a natural key is clearly a lost cause.


peppy:
@bdach just for my ease of review, can you output a `SHOW TABLE` for this table post this PR?

bdach:
```
MySQL [osu]> SHOW CREATE TABLE `beatmapset_version_files`\G
*************************** 1. row ***************************
       Table: beatmapset_version_files
Create Table: CREATE TABLE `beatmapset_version_files` (
  `id` bigint unsigned NOT NULL AUTO_INCREMENT,
  `file_id` bigint unsigned NOT NULL,
  `version_id` bigint unsigned NOT NULL,
  `filename` varchar(500) NOT NULL,
  PRIMARY KEY (`id`),
  KEY `beatmapset_version_files_file_id_index` (`file_id`),
  KEY `beatmapset_version_files_version_id_index` (`version_id`)
) ENGINE=InnoDB AUTO_INCREMENT=88 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci
1 row in set (0.001 sec)
```

peppy (approved):
Not having `beatmap_id` in this table may be a bit annoying but we can add it later if required.

nanaya (approved):

PR #11710, opened 2024-12-03
Fix beatmapset preload on room

nanaya:
Doing another `load` overrides existing loaded relations.

### DIFF
--- a/app/Http/Controllers/Multiplayer/RoomsController.php
+++ b/app/Http/Controllers/Multiplayer/RoomsController.php
@@ -212,10 +212,11 @@ class RoomsController extends Controller
     private function createJoinedRoomResponse($room)
     {
         return json_item(
-            $room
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Multiplayer/RoomsController.php
```diff
@@ -212,10 +212,11 @@ public function store()
     private function createJoinedRoomResponse($room)
     {
         return json_item(
-            $room
-                ->load('host.country')
-                ->load('playlist.beatmap.beatmapset')
-                ->load('playlist.beatmap.baseMaxCombo'),
+            $room->loadMissing([
+                'host',
```
nanaya, line 216: apparently I yeeted the country along the way but it's not needed anyway 👀

notbakaneko (approved):

PR #11715, opened 2024-12-05
Workaround guest beatmaps not showing on user profile

notbakaneko:
For the unmigrated entries or anything not updating the new table yet.

~~I was going to update another fallback but it seems to have introduced a display bug for some reason so I'll make that separate~~ 👀 
nvm it's just wrong method

### DIFF
--- a/app/Models/Beatmap.php
+++ b/app/Models/Beatmap.php
@@ -298,9 +298,10 @@ class Beatmap extends Model implements AfterCommit
             fn ($beatmapOwner) => $beatmapOwner->user ?? new DeletedUser(['user_id' => $beatmapOwner->user_id])
         );
 
+        // Fallback if unpopulated
         // TODO: remove when everything writes to beatmap_owners.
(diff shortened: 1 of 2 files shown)


nanaya:
I'm running the migration so the fallback query is probably not needed.

notbakaneko:
I was thinking new submissions but that can't have guests set during submissions....or can they? 🤔

nanaya:
I don't think that's possible, no.

app/Models/Beatmap.php
```diff
@@ -298,9 +298,10 @@ public function getOwners(): Collection
             fn ($beatmapOwner) => $beatmapOwner->user ?? new DeletedUser(['user_id' => $beatmapOwner->user_id])
         );
 
+        // Fallback if unpopulated
         // TODO: remove when everything writes to beatmap_owners.
```
nanaya, line 302: not sure the pr title is relevant anymore. also this todo should be adjusted/removed depending on how empty owners should be handled

notbakaneko:
I will think about it later 💤

PR #11716, opened 2024-12-05
Upgrade ESLint

agatemosu:


nanaya (commented):
not a yes or no but I don't mind just dropping all the eslint custom styling and go with whatever prettier spits out (yes, including the double quotes)

agatemosu:
> not a yes or no but I don't mind just dropping all the eslint custom styling and go with whatever prettier spits out (yes, including the double quotes)

Prettier also has the `singleQuote` and `jsxSingleQuote` settings, could that be used?

agatemosu:
I did another branch replacing stylistic with prettier: https://github.com/agatemosu/osu-web/commits/prettier

notbakaneko (approved):
We'll do the upgrade to 9 later.

I (or rather, `yarn` did) fixed and pushed the lockfile conflicts

PR #11720, opened 2024-12-06
Support displaying longer guest mapper lists in Beatmap Discussions

notbakaneko:
- Fixed maximum width for beatmap drop down (simpler than calculating proportional view width)
- Expand area for showing guest mappers
- rescaled the chart height to keep the previous line height and appearance.

image
image


resources/css/bem/beatmap-list.less
```diff
@@ -86,7 +86,7 @@
     padding: 10px 0;
     min-width: 100%;
     width: max-content;
-    max-width: ~"max(90vw, 100%)";
+    max-width: ~"min(max(90vw, 100%), @{container-sm})";
```
nanaya, line 89: this escaping nonsense is no longer needed

nanaya (changes requested):

syzyxy:
what is now the max users you can add in a single diff as owner?

PR #11721, opened 2024-12-06
Show team flag on user display

nanaya:
I think this covers everything? 🤔 

- [x] depends on #11714 and I want to rebase it (and clean up a bit more) so drafting first


notbakaneko (changes requested):
image
`null` flags might need a placeholder or something...

notbakaneko (approved):
I feel like the default colour of null flags isn't very visible in the listing where it'll show up, and could do with a `?` marker or something like unknown country flags but we'll see how it goes 🤷

PR #11726, opened 2024-12-08
Remove unnecessary process declaration

agatemosu:
They are already defined in `@types/node` with the same `string | undefined` type.


notbakaneko:
We should look at excluding `@types/node` and only having the exported values instead, since the project is being built for browsers, not node.

PR #11728, opened 2024-12-09
Exclude restricted user scores from score listing

nanaya:
Maybe 🤔


nanaya:
do you mean the es index? it also excludes non-preserved scores.

notbakaneko (approved):

PR #11730, opened 2024-12-09
Fix flag size on team page

nanaya:
I wonder if the image itself should be resized to 240x120 instead of 256x128 (or the avatar area should be of that size instead, which also integer scaled with current user avatar) ಠ_ಠ


notbakaneko (approved):
Think of weird calling this flag everywhere while using profile avatar 🤨

PR #11731, opened 2024-12-09
Account for retina size team flag

nanaya:
Either this or just make it a quarter of its current size on team page instead (display size of 128x64). Especially considering they're pretty tiny everywhere else.


notbakaneko:
I think it can just use retina size everywhere, the downscaling quality isn't going to make much difference on non-retina displays imo. The file size for larger pngs is more of an issue if they're being used everywhere for the small flags, etc...

nanaya:
using 512x256 image for 40x20 display just sounds silly...

notbakaneko:
That's what I said in a different way

nanaya:
Apparently we're going with this for now so we have the highest possible images and can add smaller version later (either an additional file column or resizing service through nginx or whatever)

peppy:
To give a bit more context, I think uploading and storing the larger 256x512 is best for now specifically to match the height of avatars and to give the most flexibility to team creators and to future usages of the graphics, ie. news posts or tournament overlays.

peppy (approved):
i think

notbakaneko (approved):

PR #11733, opened 2024-12-09
Fix switching locale doesn't set js locale correctly

### LINKED ISSUE #11707: Mixed (cached?) translations in page after switching the website language
Dashboard / any other tab (in Italian):
image

When I go to the friends section:
**Note the tab names in English, and I could swear that the sort options were localized before.**

image

nanaya:
Resolves #11707. Or maybe works around? There's currently no way to change js locale without full reload.


notbakaneko (approved):
Oh yeah, I forgot about this while checking something else 👀

PR #11737, opened 2024-12-10
Notify all owners on beatmap owner change

nanaya:
Something like this? 🤔 

I suppose it should only notify the delta but there's no such data at the moment and this is probably better than nothing?


notbakaneko (commented):
could store and pass the results of the `xor` in`ChangeBeatmapOwners` 🤔

notbakaneko:
Should users getting removed be notified 🤔

nanaya:
then the scope keep getting larger when thinking about this so I just went with the simplest one for now 😬

(also handling change in job class structure etc)

notbakaneko (approved):

PR #11740, opened 2024-12-12
Replace ts-loader and terser with swc

agatemosu:


peppy:
Would be nice to have a bit of context for this in the opening post.

notbakaneko:
I don't think we're going to be simply changing our tooling "just because"

nanaya:
apparently it cut off build time.

| System | Build | Before | After | Speedup |
|--------|-------|--------|-------|---------|
| dev    | dev   | 10     | 7     | 3       |
| dev    | prod  | 22     | 14    | 8       |
| gha    | prod  | 42     | 24    | 18      |

I haven't checked the output or anything else.

agatemosu:
Sorry for not including an opening post, the reason for the change is to make builds faster.

The output of swc is smaller than ts.

|      | excluding .map | including .map |
| :--- | -------------: | -------------: |
| ts   |        9128 KB |       18288 KB |
| swc  |        8512 KB |       18448 KB |
| diff |        -616 KB |        +160 KB |

PR #11742, opened 2024-12-13
Add new API endpoints for forums and topics

### LINKED ISSUE #7440: Add API route to get topic listing on a forum
continuation of #6072

should have some way to request the pinned topics as well. bonus points if poll data and first post can be requested as includes, would save a lot of requests for my use case :^)

Sheppsu:
Closes #7440 

It would be nice to have these available, and in my case, I want to automate some things on my [otdb](https://otdb.sheppsu.me) project by using the api to comb through the tournament forum. I would love to at least have the topic listing endpoint available if not the others.

Adds three endpoints:
- Get a list of top-level forums and their subforums (max 2 deep)
- Get a single forum with pinned topics and recent topics
- Get a sorted list of topics optionally under a certain forum


nanaya (changes requested):

app/Http/Controllers/Forum/ForumsController.php
```diff
@@ -88,8 +154,20 @@ public function show($id)
...
-            ->paginate();
+            ->recent(compact('sort', 'withReplies'));
+
+        if (is_api_request()) {
+            $topics->limit(Topic::PER_PAGE);
+
+            return [
+                'forum' => json_item($forum, 'Forum/Forum', ['subforums.subforums']),
+                'topics' => json_collection($topics, 'Forum/Topic'),
+                'last_topics' => json_collection($lastTopics, 'Forum/Topic'),
```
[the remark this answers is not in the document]
Sheppsu: yeah that's true

app/Http/Controllers/Forum/ForumsController.php
```diff
@@ -88,8 +151,19 @@ public function show($id)
...
-            ->recent(compact('sort', 'withReplies'))
-            ->paginate();
+            ->recent(compact('sort', 'withReplies'));
+
+        if (is_api_request()) {
+            $topics->limit(Topic::PER_PAGE);
+
+            return [
+                'forum' => json_item($forum, new ForumTransformer(), ['subforums.subforums']),
+                'topics' => json_collection($topics, new TopicTransformer()),
```
nanaya, line 161: need a `->get()` here (or above, or combined here)

app/Models/Forum/Topic.php
```diff
@@ -78,8 +79,20 @@ class Topic extends Model implements AfterCommit
...
         restore as private origRestore;
     }
+    use WithDbCursorHelper;
 
     const DEFAULT_SORT = 'new';
+    const SORTS = [
+        'new' => [
+            // type 'timestamp' because the values are stored as integer in the database
+            ['column' => 'topic_last_post_time', 'order' => 'DESC', 'type' => 'timestamp'],
+            ['column' => 'topic_last_post_id', 'order' => 'DESC'],
```
nanaya, line 89: this destroys performance due to no usable index

nanaya (changes requested):

app/Models/Forum/Topic.php
Sheppsu: should I just change it to sort by topic time & id?

nanaya: just remove this and only rely on the post time for now

Sheppsu: ah I understand now. I misread as meaning both columns didn't have an index

PR #11743, opened 2024-12-13
Require multiplayer api to join multiplayer room

nanaya:
Mainly to disallow people from rejoining room after they're kicked and the room password changed. It also makes more sense because it should have same check as joining multiplayer room instead of generic chat room.

The downside is there's currently no way to join multiplayer room through other than client api but apparently we don't really want to support that at the moment? 🤷 @peppy


notbakaneko (approved):
I'm only approving this because web chat doesn't handle rooms with passwords (or maybe prevent rejoining if password set? 🤷 ) @peppy

peppy (approved):

PR #11744, opened 2024-12-13
Add member management page

nanaya:
Currently can only remove.

💯 design.


app/Http/Controllers/Teams/MembersController.php
```diff
@@ -0,0 +1,51 @@
...
+class MembersController extends Controller
+{
+    public function __construct()
+    {
+        parent::__construct();
+
+        $this->middleware('auth');
+    }
+
+    public function destroy(string $teamId, string $userId): Response
```
nanaya, line 24: from what I gather the type will always be string (or nullable string if optional)

notbakaneko (changes requested):

notbakaneko (approved):
Appearance can be a different problem™️

PR #11747, opened 2024-12-16
Fix profile page reorder not working without full reload

nanaya:
Not quite sure what's happening here ಠ_ಠ


notbakaneko (approved):
Not sure how long this hasn't been working for, but it's because the elements are still on `newBody` instead of `window.document` so the extension isn't going to find them

nanaya:
I think all the selectors and stuff are based on the initial dom being passed, and all the elements did get the relevant `ui-` classes so it looks like to be something else...

PR #11748, opened 2024-12-17
Use casting class for forum timestamps

nanaya:
Relatively straightforward...? 🤔 

Should get around actually fixing them one day...


notbakaneko (approved):
should be fine 🤔

PR #11750, opened 2024-12-17
Show custom user tags on beatmapset

notbakaneko:
Something like this?

This isn't the final display, just something to get going (There was some discussion about show counts and colour coding the tag types as well, differentiating user/mapper tags, etc), those will be separate; I don't think there's anything set up to write tags yet either? 👀 

I just combined the mapper and user tags into the same tag it they're the same for now.
The Beatmap top tag cache now only caches the id and count, and the tags are moved into a local singleton cache.
The `api.beatmap.tags.index` endpoint is also removed since that doesn't seem useful when you're probably wanting the tags together with the Beatmap/Beatmapset with the mapper tags as well?

I'm also splitting up the search and indexing part into a separate PR after shuffling some things around.


app/Models/Tag.php
```diff
@@ -23,19 +23,19 @@ public function beatmapTags(): HasMany
         return $this->hasMany(BeatmapTag::class);
     }
 
-    public static function topTags($beatmapId)
+    public static function topTagIds(int $beatmapId, int $limit = 50)
```
nanaya, line 26: there isn't anything in this that require `Tag` anymore, is there?

resources/js/beatmapsets-show/controller.ts
```diff
@@ -70,6 +82,51 @@ export default class Controller {
...
+    for (const tag of this.beatmapset.related_tags) {
+      map.set(tag.id, tag);
+    }
+
+    return map;
+  }
+
+  @computed
+  get tags() {
+    const mapperTagSet = new Set(this.beatmapset.tags.split(' ').filter(present));
```
nanaya, line 97: I don't know about deduplicating the old tag words. It may "break" titles/names like repeated `no` and other stuff which is why it was not done in the first place.

nanaya (changes requested):

resources/js/beatmapsets-show/controller.ts
```diff
@@ -70,6 +82,51 @@ export default class Controller {
...
+    const tagMap = new Map<number, TagJsonWithCount>();
+
+    for (const tag of this.beatmapset.related_tags) {
+      tagMap.set(tag.id, asTagJsonWithCount(tag));
+    }
+
+    for (const beatmap of this.beatmapset.beatmaps) {
```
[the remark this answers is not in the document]
notbakaneko: 🤔 
@peppy do we want the tag list to change when selecting a different beatmap difficulty?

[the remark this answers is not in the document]
peppy: they are intended to also be able to describe gameplay mechanics, so i think per-difficulty makes sense. we'd probably have two buttons in the in-game interface – one to suggest a tag for all difficulties and another for only the current difficulty.

resources/js/beatmapsets-show/controller.ts
```diff
@@ -70,6 +82,41 @@ export default class Controller {
...
+    const userTags: TagJsonWithCount[] = [];
+
+    if (this.currentBeatmap.top_tag_ids != null) {
+      for (const tagId of this.currentBeatmap.top_tag_ids) {
+        const maybeTag = this.relatedTags.get(tagId.tag_id);
+        if (maybeTag == null) continue;
+
+        const tag = asTagJsonWithCount(maybeTag);
+        tag.count = tagId.count;
+        userTags.push(asTagJsonWithCount(tag));
```
nanaya, line 107: can just push `{ ...maybeTag, count: tagId.count }`

(also why is it converting the thing twice)

resources/js/beatmapsets-show/info.tsx
```diff
@@ -191,14 +196,14 @@ export default class Info extends React.Component<Props> {
...
+          {this.tags.length > 0 &&
             <div className='beatmapset-info__row beatmapset-info__row--value-overflow'>
               <h3 className='beatmapset-info__header'>
                 {trans('beatmapsets.show.info.tags')}
               </h3>
               <div className='beatmapset-info__value-overflow'>
-                {tags.map((tag, i) => (
-                  <React.Fragment key={`${tag}-${i}`}>
+                {this.tags.map((tag) => (
+                  <React.Fragment key={`${tag}`}>
```
nanaya, line 206: the `-${i}` is required to avoid duplicate keys (mainly for mapper tags but also now in conjunction with user tags)

app/Models/BeatmapTag.php
```diff
@@ -10,14 +10,29 @@
...
  * @property-read User $user
  * @property int $user_id
  */
 class BeatmapTag extends Model
 {
     protected $primaryKey = ':composite';
     protected $primaryKeys = ['beatmap_id', 'tag_id', 'user_id'];
+    public $incrementing = false;
+
+    public static function topTagIdsQuery(int $beatmapId, int $limit = 50)
```
nanaya, line 25: why is it not a scope

app/Models/BeatmapTag.php
```diff
@@ -10,14 +10,29 @@
...
  * @property int $tag_id
+ * @property \Carbon\Carbon $updated_at
  * @property-read User $user
  * @property int $user_id
  */
 class BeatmapTag extends Model
 {
     protected $primaryKey = ':composite';
     protected $primaryKeys = ['beatmap_id', 'tag_id', 'user_id'];
+    public $incrementing = false;
```
nanaya, line 23: position public before protected?

nanaya (changes requested):

tsunyoku:
> The `api.beatmap.tags.index` endpoint is also removed since that doesn't seem useful when you're probably wanting the tags together with the Beatmap/Beatmapset with the mapper tags as well?

The intent with this one was to have a full list of tags you can use when adding a tag to a beatmap. If we don't have a way to retrieve this, how are users supposed to add new tags on web/client?

nanaya:
> The intent with this one was to have a full list of tags you can use when adding a tag to a beatmap. If we don't have a way to retrieve this, how are users supposed to add new tags on web/client?

that'd be `api.tags.index`. The beatmap one is specific tags for the beatmap.

tsunyoku:
Oh, right. If the tags are attached to the Beatmap itself then yeah that endpoint is useless.

nanaya (approved):
the way tag count is loaded is questionable but it can be dealt with later if needed

PR #11757, opened 2024-12-19
Add leave team button

nanaya:
Apart of the message that needs fixing, I wonder if the route should have team id in it... (there's also question of if it's worth adding an OsuAuthorize check for this)

### DIFF
--- a/app/Http/Controllers/TeamsController.php
+++ b/app/Http/Controllers/TeamsController.php
@@ -8,17 +8,40 @@ declare(strict_types=1);
 namespace App\Http\Controllers;
 
 use App\Models\Team;
+use App\Models\TeamMember;
 use App\Transformers\UserCompactTransformer;
 use Symfony\Component\HttpFoundation\Response;
 
 class TeamsController extends Controller
 {
+    public function __construct()
+    {
+        parent::__construct();
+        $this->middleware('auth', ['only' => ['part']]);
+    }
+
     public function edit(string $id): Response
     {
         $team = Team::findOrFail($id);
         priv_check('TeamUpdate', $team)->ensureCan();
 
         return ext_view('teams.edit', compact('team'));
+    }
+
+    public function part(): Response
+    {
+        $member = TeamMember::findOrFail(\Auth::user()->getKey());
+
+        if ($member === null) {
+            \Session::flash('popup', osu_trans('teams.part.no_team'));
+
+            return ujs_redirect(route('home'));
+        }
+
+        $member->delete();
+        \Session::flash('popup', osu_trans('teams.part.ok'));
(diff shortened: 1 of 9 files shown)


resources/views/teams/show.blade.php
```diff
@@ -13,6 +13,12 @@
     $teamMembers['member'] ??= [];
     $teamMembers['leader'] ??= $toJson([$team->members()->make(['user_id' => $team->leader_id])->userOrDeleted()]);
     $headerUrl = $team->header()->url();
+
+    $currentUser = Auth::user();
+    $buttons = new Ds\Set();
+    if ($currentUser !== null && $currentUser->team?->getKey() === $team->getKey() && $currentUser->getKey() !== $team->leader_id) {
+        $buttons->add('part');
+    }
```
notbakaneko, line 21: Are there more buttons coming 👀

app/Http/Controllers/TeamsController.php
```diff
@@ -21,6 +28,22 @@ public function edit(string $id): Response
...
 
+    public function part(): Response
+    {
+        $member = TeamMember::findOrFail(\Auth::user()->getKey());
+
+        if ($member === null) {
+            \Session::flash('popup', osu_trans('teams.part.no_team'));
+
+            return ujs_redirect(route('home'));
+        }
```
notbakaneko, line 39: always going to be false with `findOrFail` above

notbakaneko (changes requested):

resources/views/teams/show.blade.php
nanaya: why yes 👀

app/Http/Controllers/TeamsController.php
notbakaneko: localization key also unused

resources/views/teams/show.blade.php
```diff
@@ -69,6 +75,21 @@ class="btn-circle btn-circle--page-toggle"
...
+        @if (!$buttons->isEmpty())
+            <div class="profile-detail-bar profile-detail-bar--team">
+                @if ($buttons->contains('part'))
+                    <form
+                        action="{{ route('teams.part') }}"
+                        data-turbo-confirm="{{ osu_trans('common.confirmation') }}"
+                        method="POST"
+                    >
+                        <button class="team-action-button team-action-button--part">
+                                {{ osu_trans('teams.show.bar.part') }}
```
notbakaneko, line 87: too many idents btw

nanaya:
sec, I think it's more convenient to just add an OsuAuthorize check

notbakaneko (approved):

PR #11758, opened 2024-12-20
Add disband team button

nanaya:
This PR should contain all the things involved in deleting a team 🤔 (well, in the sense of whatever implemented at the moment which doesn't include pending team application because the model doesn't exist yet and the team chat channel)

Which currently just involves deleting the team itself and related files and member entries ┐(ﾟヮﾟ)┌

Extra things I can think of:
1. requiring the team to be empty first (but why)
2. notifying the members

### DIFF
--- a/app/Models/Team.php
+++ b/app/Models/Team.php
@@ -76,6 +76,19 @@ class Team extends Model
             : bbcode((new BBCodeForDB($description))->generate());
     }
 
+    public function delete()
+    {
+        $ret = parent::delete();
+
+        if ($ret) {
+            $this->header()->delete();
+            $this->logo()->delete();
+            $this->members()->delete();
+        }
+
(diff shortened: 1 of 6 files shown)


app/Models/Team.php
```diff
@@ -76,6 +76,19 @@ public function descriptionHtml(): string
...
     }
 
+    public function delete()
+    {
+        $ret = parent::delete();
+
+        if ($ret) {
+            $this->header()->delete();
+            $this->logo()->delete();
+            $this->members()->delete();
```
notbakaneko, line 86: I feel like the db stuff should be deleted together in a transaction, and the assets should be removed first so there's no dangling stuff if something somehow fails (can just try delete the team again)

notbakaneko (approved):

PR #11763, opened 2024-12-25
Include user tags in Beatmapset search

notbakaneko:
Needs a new beatmap search index to be built since the tags are being added to the nested beatmap object instead of the top level tags so  that they can be targetted separately.

- [x] (not really but it's simpler not to reimplement the same function) requires #11750


app/Libraries/Search/BeatmapsetSearch.php
```diff
@@ -60,6 +60,12 @@ public function getQuery()
                     ->should(['term' => ['_id' => ['value' => $this->params->queryString, 'boost' => 100]]])
                     ->should(QueryHelper::queryString($this->params->queryString, $partialMatchFields, 'or', 1 / count($terms)))
                     ->should(QueryHelper::queryString($this->params->queryString, [], 'and'))
+                    ->should([
+                        'nested' => [
```
nanaya, line 64: shouldn't this be also shared with the beatmap specific `$nested` query down below?

app/Jobs/EsDocumentUnique.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace App\Jobs;
+
+use Illuminate\Contracts\Queue\ShouldBeUnique;
+
+class EsDocumentUnique extends EsDocument implements ShouldBeUnique
+{
+    public int $uniqueFor = 600;
+
+    public function __construct($model)
```
nanaya, line 16: what's this overriding

app/Libraries/Search/BeatmapsetQueryParser.php
```diff
@@ -82,6 +82,9 @@ public static function parse(?string $query): array
                 case 'source':
                     $option = static::makeTextOption($op, $m['value']);
                     break;
+                case 'tag':
+                    $option = static::makeTextOption($op, $m['value']);
```
nanaya, line 86: being able to search for multiple tags would be useful?

app/Models/Traits/Es/BeatmapsetSearch.php
```diff
@@ -92,23 +97,30 @@ private function esBeatmapsValues()
...
-                    $convert->convert = true;
-                    $convertValues = [];
-                    foreach ($mappings as $field => $mapping) {
-                        $convertValues[$field] = $convert->$field;
-                    }
-
-                    $convertValues['user_id'] = $beatmapValues['user_id']; // just add a copy for converts too.
-                    $values[] = $convertValues;
+                    $values[] = [
+                        ...$beatmapValues,
```
nanaya, line 101: some fields have different value when in convert mode

nanaya (changes requested):

app/Libraries/Search/BeatmapsetSearch.php
notbakaneko: The scoring and matching of the query string is not the same in the other nested filter

nanaya: would this match tag that isn't part of beatmap specific field match? searching by this and difficulty for example

notbakaneko: Searching for `potato foo difficulty=mayday` returns beatmapsets with `mayday` difficulty with relevance score according to how the fields in the query match `potato foo`

app/Libraries/Search/BeatmapsetSearch.php
```diff
@@ -398,6 +405,17 @@ private function addTextFilter(BoolQuery $query, string $paramField, array $fiel
...
 
+    private function addTagsFilter(BoolQuery $query): void
+    {
+        if (!present($this->params->tags)) {
+            return;
+        }
+
+        foreach ($this->params->tags as $tag) {
+            $query->filter(QueryHelper::queryString($tag, ['beatmaps.top_tags'], 'and'));
+        }
```
notbakaneko, line 416: This filter means that specifying multiple tags requires the beatmapset to have a beatmap that has all the tags; having the tags across different beatmaps in the set will not match.

app/Models/Traits/Es/BeatmapsetSearch.php
```diff
@@ -109,6 +119,18 @@ private function esBeatmapsValues()
         return $values;
     }
 
+    private function esBeatmapTags(Beatmap $beatmap): array
```
nanaya, line 122: can be a static function

app/Libraries/Search/BeatmapsetSearch.php
```diff
@@ -398,6 +405,17 @@ private function addTextFilter(BoolQuery $query, string $paramField, array $fiel
...
     }
 
+    private function addTagsFilter(BoolQuery $query): void
+    {
+        if (!present($this->params->tags)) {
+            return;
+        }
+
+        foreach ($this->params->tags as $tag) {
+            $query->filter(QueryHelper::queryString($tag, ['beatmaps.top_tags'], 'and'));
```
nanaya, line 415: apparently searching for `tag="a b"` will also include beatmaps which has tag `a` and `b` in addition to `a b`...? not sure if intentional.

also `tag="a"` will also include `a b` and I can't find way to filter just `a` (although there isn't such tag at the moment)

app/Libraries/Search/BeatmapsetSearch.php
nanaya: that'd be weird anyway. just like specifying multiple different difficulty parameters should match beatmap which has all those parameters, not when they're spread across multiple beatmaps within same beatmapset.

nanaya (changes requested):

app/Libraries/Search/BeatmapsetSearch.php
notbakaneko: Well that's partly a parsing issue unless we want to start going `tag=""a""` for exact matches...
`creator="a"` already includes `a b` and `a c` so it's not really doing anything different at the moment.

nanaya: `creator=a` does only include user `a` though 🤔 (it uses user id lookup if the user exists)

nanaya (approved):
eh, can be adjusted later I guess 🤷‍♀️

app/Libraries/Search/BeatmapsetSearch.php
```diff
@@ -398,6 +406,25 @@ private function addTextFilter(BoolQuery $query, string $paramField, array $fiel
...
     }
 
+    private function addTagsFilter(BoolQuery $query): void
+    {
+        if ($this->params->tags === null) {
+            return;
+        }
+
+        $tagSet = new Set(array_map('mb_strtolower', $this->params->tags));
+        $tags = Tag::whereIn('name', $this->params->tags)->limit(10)->pluck('name');
```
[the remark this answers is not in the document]
notbakaneko: so you can't put hundreds of tags in; the limit should be in the parser but I didn't decide whether it should be throwing or how the UI is supposed to handle it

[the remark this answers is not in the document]
nanaya: well, the remaining hundreds of tags will still be looked up anyway... presumably with slower text query instead of keyword

Issue #6736, opened 2020-10-06
getting anonymous id 1 user by profile bbcode tag


lmnyx:
![oop](https://user-images.githubusercontent.com/13962537/95254676-ab001700-0839-11eb-958c-3d083c5229b6.png)
`[profile]1[/profile]`

PR #11764, opened 2024-12-26
Exclude bot and no profile special groups from being set as guest mappers

### LINKED ISSUE #11752: Members of the `BOT` usergroup shouldn't be able to be assigned guest difficulties
Self-explanatory I guess.

Related to the recent incident on [this map](http://osu.ppy.sh/s/2287978), where the mapset host tried to set the ownership of his map to bot accounts such as "[BanchoBot](http://osu.ppy.sh/u/3)" and "[Nomination Assessment Team](http://osu.ppy.sh/u/6616586)" :

image

notbakaneko:
closes #11752

Also changes Chat Announcements and supporter tag gift lookups from the UI to exclude those groups, but I didn't change the actual check where those are applied yet.

### DIFF
--- a/app/Http/Controllers/Users/LookupController.php
+++ b/app/Http/Controllers/Users/LookupController.php
@@ -22,7 +22,12 @@ class LookupController extends Controller
     public function index()
     {
         // TODO: referer check?
-        $ids = array_slice(array_reject_null(get_arr(request('ids'), presence(...)) ?? []), 0, 50);
+        $params = get_params(request()->all(), null, [
+            'exclude_bots:bool',
+            'ids:string[]',
+        ]);
+
+        $ids = array_slice(array_reject_null(get_arr($params['ids'] ?? [], presence(...))), 0, 50);
 
         $numericIds = [];
         $stringIds = [];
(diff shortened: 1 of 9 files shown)


app/Http/Controllers/Users/LookupController.php
```diff
@@ -35,8 +35,7 @@ public function index()
         }
 
         $users = User::where(fn ($q) => $q->whereIn('user_id', $numericIds)->orWhereIn('username', $stringIds))
-            ->where('group_id', '<>', app('groups')->byIdentifier('no_profile')->getKey())
-            ->default()
+            ->defaultForLookup()
```
nanaya, line 38: this also affects user card which we probably want to keep showing for bots?

app/Http/Controllers/Users/LookupController.php
```diff
@@ -35,13 +40,15 @@ public function index()
         }
 
         $users = User::where(fn ($q) => $q->whereIn('user_id', $numericIds)->orWhereIn('username', $stringIds))
-            ->where('group_id', '<>', app('groups')->byIdentifier('no_profile')->getKey())
```
nanaya, line 38: I think this should stay regardless of exclude bots option? See #6736

nanaya (changes requested):

nanaya (approved):

PR #11768, opened 2025-01-02
Add user season score calculation workflow

venix12:

Serving as a base for implementation of seasonal rankings in nearing feature.

This mimics the Beatmap Spotlights seasonal rankings calculation process that used to be done via an external script.

Also adds an artisan command for score recalculating.

Tried to optimize the calculation process as much as possible, any further potential improvements on that matter would be vastly appreciated. :thinking:

### Rooms

1. The seasons are designed as a series of few (currently: three) playlists that are usually labeled by consecutive letters of alphabet (A, B, C).

2. Each of the playlists is ran twice over a course of the season. For the sake of this implementation, the rooms of the first run of the playlists are referred to as "parent rooms", while the rooms of the second run are called "children rooms".

This PR adds ``group_indicator`` to the ``season_rooms`` table, to group parent and children rooms together.

### Score calculation

*Total (user) season score* is sum of best *playlist total scores* achieved throughout the season, with added factor-based weighting.

1. Total scores of a each playlist are compared between the parent and child rooms, with higher being taken for *total season score* calculation.

2. Said scores are sorted from highest to lowest, and season-specified factors are being applied in descending order to each of them (e.g. 1.0, 0.75, 0.5). This results in final *total season score*.

This PR adds the following:
- ``user_season_score_aggregates`` table, ``total_score`` of which is calculated using related model ``calculate()`` function, that is called whenever user places a score on related season's room.
- ``score_factors`` field on ``seasons`` table that stores factors used for *total season score* calculation.

#### Example user total season score calculation

Let's say a user achieves following scores during the six playlists of the season:

| playlist | score (parent) | score (child) |
|----------|----------------|---------------|
| A        | 10             | **12**        |
| B        | **20**         | 15            |
| C        | 5              | **40**        |

The season's score factors are set to following: 1, 0.75, 0.5.

Playlist final scores are being sorted in descending order, resulting in following values: 40, 20, 12. The final season score calculation goes as below:

```
40 * 1 + 20 * 0.75 + 12 * 0.5 = 40 + 15 + 6 = 61
```

The *total season score* is **61**. That value would be used for sake of placing user in seasonal leaderboard.


app/Models/Multiplayer/Room.php
```diff
@@ -37,6 +37,7 @@
  * @property int $id
  * @property int|null $max_attempts
  * @property string $name
+ * @property int|null $parent_id
```
cl8n, line 40: should be in `SeasonRoom` table instead. I also think it may be easier to manage/implement by storing the playlist letter itself, instead of linking rooms together

app/Models/UserSeasonScore.php
```diff
@@ -0,0 +1,79 @@
...
+
+        foreach ($parentRooms as $room) {
+            $totalScore = $userScores->where('room_id', $room->getKey())
+                ->first()
+                ?->total_score;
+
+            $childRoomId = $this->season->rooms
+                ->where('parent_id', $room->getKey())
+                ->first()
+                ?->getKey();
```
cl8n, line 50: should allow more than one child so you can re-run playlists more than once. well idk if that would ever actually happen but there's no reason to hardcode this in either (same with unique constraint on parent ID)

cl8n (changes requested):
mostly commenting on the db structure, I didn't review all the code too closely

app/Models/SeasonScoreFactor.php
```diff
@@ -0,0 +1,22 @@
...
+
+declare(strict_types=1);
+
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+
+/**
+ * @property int $id
+ * @property float $factor
```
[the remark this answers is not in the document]
venix12: factors should really be consistent across the seasons these days, though used to be changed in the past (e.g. i believe 4-playlist format we had before used a different set than what we use now), so this is mostly for backwards compatibility i think

app/Models/UserSeasonScore.php
```diff
@@ -0,0 +1,79 @@
...
+    public function calculate(): void
+    {
+        $userScores = UserScoreAggregate::whereIn('room_id', $this->season->rooms->pluck('id'))
+            ->where('user_id', $this->user_id)
+            ->get();
+
+        $factors = $this->season->scoreFactorsOrderedForCalculation();
+        $parentRooms = $this->season->rooms->where('parent_id', null);
+
+        if ($parentRooms->count() > count($factors)) {
```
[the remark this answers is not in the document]
venix12: having too many factors shouldn't be an issue when they are ordered in descending order for calculation, it is handy to just set them all per-season without having to update with every new playlist launch :thinking:

database/factories/SeasonRoomFactory.php
```diff
@@ -0,0 +1,21 @@
...
+
+use App\Models\SeasonRoom;
+
+class SeasonRoomFactory extends Factory
+{
+    protected $model = SeasonRoom::class;
+
+    public function definition(): array
+    {
+        // pivot table...
```
nanaya, line 18: it can still have the related models defined here

app/Models/Multiplayer/Room.php
```diff
@@ -446,6 +446,19 @@ public function completePlay(ScoreToken $scoreToken, array $params): ScoreLink
...
 
+            // spotlight playlists should always be linked to one season exactly
+            if ($this->category === 'spotlight' && $agg->total_score > 0 && $this->seasons()->count() === 1) {
+                $seasonId = $this->seasons()->first()->getKey();
+
+                $seasonScore = $user->seasonScores()
+                    ->where('season_id', $seasonId)
+                    ->firstOrNew();
+
+                $seasonScore->season_id = $seasonId;
```
nanaya, line 457: `->season()->associate($season)` (this also loads the relation so it's not queried again in calculate)

nanaya (changes requested):

tests/Models/UserSeasonScoreAggregateTest.php
```diff
@@ -0,0 +1,193 @@
...
+    private function createRoomWithPlay(float $totalScore, string $groupIndicator): Room
+    {
+        $room = $this->createRoom($groupIndicator);
+
+        $playlistItem = PlaylistItem::factory()->create([
+            'owner_id' => $room->host,
+            'room_id' => $room,
+        ]);
+
+        $this->roomAddPlay($this->user, $playlistItem, [
```
nanaya, line 186: more like `static::roomAddPlay` but either works I guess 🤷

tests/Models/UserSeasonScoreAggregateTest.php
```diff
@@ -0,0 +1,193 @@
...
+        SeasonRoom::factory()->create([
+            'group_indicator' => $groupIndicator,
+            'room_id' => $room,
+            'season_id' => $this->season,
+        ]);
+
+        return $room;
+    }
+
+    private function createRoomWithPlay(float $totalScore, string $groupIndicator): Room
```
nanaya, line 177: I wonder if `$groupIndicator` should be first argument instead. It's like `createRoom` (with indicator argument) extended with `WithPlay` (with additional score argument)

nanaya (changes requested):

tests/Models/UserSeasonScoreAggregateTest.php
venix12: UserScoreAggregateTest used ``$this->roomAddPlay()`` so just mimicked it here, but static probably makes more sense indeed 🤔

app/Models/UserSeasonScoreAggregate.php
```diff
@@ -0,0 +1,69 @@
...
+    protected $primaryKeys = ['user_id', 'season_id'];
+
+    public function calculate(bool $muteExceptions = true): void
+    {
+        $seasonRooms = SeasonRoom::where('season_id', $this->season->getKey())->get();
+        $userScores = UserScoreAggregate::whereIn('room_id', $seasonRooms->pluck('room_id'))
+            ->where('user_id', $this->user_id)
+            ->get();
+
+        $factors = $this->season->score_factors;
```
nanaya, line 35: need a `?? []` otherwise `count(null)` throws

database/migrations/2025_01_22_190453_create_user_season_score_aggregates_table.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('user_season_score_aggregates', function (Blueprint $table) {
+            $table->bigInteger('user_id')->unsigned();
+            $table->integer('season_id')->unsigned();
+            $table->float('total_score');
```
nanaya, line 22: this explodes on values as small as 7M

```
SQLSTATE[22003]: Numeric value out of range: 1264 Out of range value for column 'total_score' at row 1 (Connection: mysql, SQL: insert into `user_season_score_aggregates` (`user_id`, `season_id`, `total_score`) values (189617, 40, 7643867))
```

It needs to be either actual float or double (`float` here is `double(8,2)`). Or maybe rounded to int instead. Maybe @peppy can weigh in.

nanaya (changes requested):

database/migrations/2025_01_22_190453_create_user_season_score_aggregates_table.php
peppy: is that a laravel thing which maps `float` to `double(8,2)`?

i'm fine with any suitable data type. i don't think this table should get big enough for the choice to matter.

nanaya: yeah afaict up to laravel 10 it's not possible to use this syntax to create actual float column

venix12: went ahead and just changed it to regular double

nanaya (changes requested):

nanaya (approved):

PR #11770, opened 2025-01-06
Add `osu_logins` migration

peppy:
For use in https://github.com/ppy/osu-server-spectator/pull/253.

### DIFF
--- a/database/migrations/2025_01_06_000000_create_osu_logins.php
+++ b/database/migrations/2025_01_06_000000_create_osu_logins.php
@@ -23,14 +23,14 @@ return new class extends Migration
             $table->string('ip', 100)->default('');
             $table->timestamp('date')->useCurrent();
 
-            $table->index('user_id');
-            $table->index('date');
-            $table->index('ip');
+            $table->index('user_id', 'user_id');
+            $table->index('date', 'date');
+            $table->index('ip', 'ip');
         });
     }
 
     public function down(): void
     {
-        Schema::dropIfExists('teams');
+        Schema::dropIfExists('osu_logins');
     }
 };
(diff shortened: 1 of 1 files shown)


database/migrations/2025_01_06_000000_create_osu_logins.php
```diff
@@ -0,0 +1,36 @@
...
+
+            $table->index('user_id');
+            $table->index('date');
+            $table->index('ip');
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::dropIfExists('teams');
```
nanaya, line 34:
```suggestion
        Schema::dropIfExists('osu_logins');
```

database/migrations/2025_01_06_000000_create_osu_logins.php
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        Schema::create('osu_logins', function (Blueprint $table) {
+            $table->unsignedInteger('user_id')->default(0);
+            $table->string('ip', 100)->default('');
+            $table->timestamp('date')->useCurrent();
+
+            $table->index('user_id');
+            $table->index('date');
+            $table->index('ip');
```
nanaya, line 28: can use a matching index name with production
```suggestion
            $table->index('user_id', 'user_id');
            $table->index('date', 'date');
            $table->index('ip', 'ip');
```

nanaya (changes requested):

database/migrations/2025_01_06_000000_create_osu_logins.php
peppy: 🙈

nanaya (approved):

PR #11772, opened 2025-01-07
Show user card on discussion page user link

nanaya:
Seems useful? Noticed this missing when checking something else.


notbakaneko (approved):
I think this was more the user card didn't really show much more than what was already on the discussion page...

PR #11775, opened 2025-01-08
Rework score token validation

nanaya:
Combining hash and ruleset validation in the token model itself. build id stays as is because it's quite a bit more involved.

The way `beatmap_hash` is passed and validated is rather questionable 👀


peppy:
Is this intended to ease the implementation of freestyle, or related to something else?

nanaya:
it's related to the freestyle thing, yes

notbakaneko (approved):
I feel like there should be assertions some of the values don't change if the token already exists (because it doesn't make sense for them to change) but that's a different problem

Issue #11777, opened 2025-01-09
Running `testjs` in Docker results in root owning `.docker/js-build`


AJGranowski:
Running the Javascript tests with Docker results in root owning `.docker/js-build`.

**Reproduction steps**
1. `git clone git@github.com:ppy/osu-web.git`
2. `cd osu-web/`
3. `bin/docker_dev.sh --wait; sleep 10 && docker compose down`
4. [`docker compose run --rm testjs`](https://github.com/ppy/osu-web/blob/88dc4d289bb24d2cc748da02dbbb817cff1d22eb/SETUP.md?plain=1#L355-L359)
5. `find -user 0`
   * Observe that `.docker/js-build` is owned by root.

**Expected behavior**
Files created by running Docker commands are owned by the current host user, not root.

AJGranowski:
`find -user 0` now returns zero results, meaning no items are owned by root after running `testjs` with Docker.

Thank you!

PR #11778, opened 2025-01-09
Add Docker User Mirror

AJGranowski:
## What are these changes?
This change adds [**Docker User Mirror**](https://github.com/AJGranowski/docker-user-mirror).

## Why are these changes being made?

This change resolves:
* #11777

The [current approach](https://github.com/ppy/osu-web/blob/88dc4d289bb24d2cc748da02dbbb817cff1d22eb/docker/development/entrypoint.sh) mirrors the host user well, but relies on hard coded paths to fix ownership issues. Rather than hard coding these paths, we should automate the Docker environment setup using information already contained in the compose file. I made a project called Docker User Mirror that does just that, and I think it would be a great fit for this project!

This change replaces `entrypoint.sh` with two scripts:
1. `user-mirror`: Creates items on the host before the Docker daemon does (so they're owned by the current host user rather than root), and passes environment variables into the container to mirror the host user.
2. `entrypoint`: Mirrors the host user, sets the ownership of a list of items to that mirrored user, and steps down from root to that user using [`setpriv`](https://www.man7.org/linux/man-pages/man1/setpriv.1.html) to execute a command.

(there's also a bit of glue added to `docker-compose.yml` and `Dockerfile.development`)

Fairly simple user experience: rather than running `docker ...` you instead run `./user-mirror docker ...`

## Notes

* I didn't touch any documentation. Let me know what you would like to see there.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -103,13 +103,14 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Services
-        run: './user-mirror docker compose up --wait
+        run: './user-mirror docker compose up --quiet-pull --wait
           beatmap-difficulty-lookup-cache
           db
           elasticsearch
           notification-server
           redis
-          score-indexer'
+          score-indexer
+          '
 
       - name: Setup node.js
         uses: actions/setup-node@v4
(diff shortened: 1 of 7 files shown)


nanaya:
no

PR #11779, opened 2025-01-09
Fix first-time run of `docker_dev.sh`

AJGranowski:
## What are these changes?
This change makes octane commands wait a bit for `manifest.json` to exist if it's missing.

## Why are these changes being made?

The first run of `bin/docker_dev.sh` causes both `nginx` and `php-dusk` fail.

php:
```
Exception
The manifest does not exist.
at app/Singletons/AssetsManifest.php:22
   18▕     {
   19▕         $manifestPath = public_path('assets/manifest.json');
   20▕
   21▕         if (!file_exists($manifestPath)) {
➜ 22▕             throw new Exception('The manifest does not exist.');
   23▕         }
   24▕
   25▕         $this->manifest = json_decode(file_get_contents($manifestPath), true);
   26▕     }
1   app/Providers/AppServiceProvider.php:96
    App\Singletons\AssetsManifest::__construct()
```

php-dusk:
```
php-dusk exited with code 157
```

nginx:
```
[emerg] host not found in upstream "php-dusk" in /etc/nginx/conf.d/default.conf:53
nginx exited with code 1
```

Successive runs pass and serve the website, suggesting a race condition in `docker compose up`. Running `docker compose run --rm assets` beforehand seems to fix this.

### DIFF
--- a/docker/development/run.sh
+++ b/docker/development/run.sh
@@ -20,7 +20,19 @@ _migrate() {
 }
 
 _octane() {
-  exec ./artisan octane:start --host=0.0.0.0 "$@"
+    echo_counter=0
+    manifest_path='./public/assets/manifest.json'
+    while [ ! -f "$manifest_path" ]; do
+        if [ "$echo_counter" -le 0 ]; then
+            echo_counter=5
+            echo "waiting to start octane: ${manifest_path##*/} not found" >&2
+        fi
+
+        : $(( echo_counter -= 1 ))
+        sleep 1
+    done
+
+    exec ./artisan octane:start --host=0.0.0.0 "$@"
 }
 
 _schedule() {


nanaya:
what should be done is adding health check on assets service, checking for existence of the manifest file.

AJGranowski:
I'll take a look into that.

AJGranowski:
Hmm, looks like things break if I just use health checks... I think `assets` depends on `composer install`?

```
PHP Warning:  require(/app/vendor/autoload.php): Failed to open stream: No such file or directory in /app/artisan on line 18
PHP Fatal error:  Uncaught Error: Failed opening required '/app/vendor/autoload.php' (include_path='.:/usr/share/php') in /app/artisan:18

Extracting localizations...
[webpack-cli] Error: ENOENT: no such file or directory, open '/app/resources/builds/messages.json'
    at Object.openSync (node:fs:573:18)
    at Object.readFileSync (node:fs:452:35)
    at getAllMesssages (/app/resources/js/cli/generate-localizations.js:36:22)
    at extractLanguages (/app/resources/js/cli/generate-localizations.js:20:20)
    at Object.generateLocalizations [as callback] (/app/resources/js/cli/generate-localizations.js:66:21)
    at /app/webpack.config.js:292:40
    at Array.forEach (<anonymous>)
    at configFunction (/app/webpack.config.js:292:11)
    at loadConfigByPath (/app/node_modules/webpack-cli/lib/webpack-cli.js:1445:37)
    at async WebpackCLI.loadConfig (/app/node_modules/webpack-cli/lib/webpack-cli.js:1515:38) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/app/resources/builds/messages.json'
}
error Command failed with exit code 2.
```

nanaya:
maybe add a compose install before the yarn install in the watch command

AJGranowski:
Yep, that seems to work. Let me know if you were thinking of a different approach.

nanaya (changes requested):
looking again at how the service is actually used, maybe just monitor for the file (with a while loop) inside the octane specific entrypoint function instead

(also, moving the composer install before setting up github token for it is bad)

AJGranowski:
Were you thinking of something like this latest change?

app/Singletons/AssetsManifest.php
```diff
@@ -12,14 +12,23 @@
...
+    private const MANIFEST_TIMEOUT_SECONDS = 60; // Time limit to wait for a missing manifest file before throwing.
+
     private $manifest;
 
     public function __construct()
     {
         $manifestPath = public_path('assets/manifest.json');
+        $startTimeSeconds = microtime(true);
+
+        while (!file_exists($manifestPath)) {
```
nanaya, line 25: I wouldn't do it here but in `run.sh` instead as it shouldn't be needed for anything but docker dev (esp if something is broken on prod it'll probably never finish and just needlessly waiting).

At least this would need a logging somewhere because otherwise it will just hang with no indication of what's going on while waiting.

nanaya (changes requested):

app/Singletons/AssetsManifest.php
nanaya: alternatively, loading the manifest on first request is probably fine

AJGranowski:
Migrated the manifest check to `run.sh`. On my machine, the script waits between 15 and 70 seconds for the manifest file to generate.

Here's what the stderr output looks like:
stderr_log

docker/development/run.sh
```diff
@@ -20,7 +39,19 @@ _migrate() {
 }
 
 _octane() {
-  exec ./artisan octane:start --host=0.0.0.0 "$@"
+    start_time_seconds="$(uptime_seconds)";
+    while [ ! -f "$MANIFEST_PATH" ]; do
+        elapsed_time_seconds="$(( $(uptime_seconds) - $start_time_seconds ))";
+        if [ "$elapsed_time_seconds" -ge "$MANIFEST_TIMEOUT_SECONDS" ]; then
+            echo_error "$0 [pid $$] [${elapsed_time_seconds}s]: The manifest does not exist.";
+            exit 1;
```
nanaya, line 47: I don't think this is needed. If there's problem on the assets and fixed later (and thus the manifest is created), it should just then work right away

docker/development/run.sh
```diff
@@ -20,7 +39,19 @@ _migrate() {
...
+        fi
+
+        echo_warning "$0 [pid $$] [${elapsed_time_seconds}s]: ${MANIFEST_PATH} not found.";
+        sleep $MANIFEST_TIMEOUT_INTERVAL_SECONDS;
```
nanaya, line 51: wait 1s instead. Use a different logic to reduce logging frequency with something like `n=0; while ...; do if [ $n = 4 ]; then n=0; echo ...; else : $((n++)); fi; ...; done` (unformatted)

docker/development/run.sh
```diff
@@ -20,7 +39,19 @@ _migrate() {
...
+    done
+
+    exec ./artisan octane:start --host=0.0.0.0 "$@";
```
nanaya, line 54: I've never seen shell script with trailing semicolon everywhere before...

docker/development/run.sh
```diff
@@ -20,7 +39,19 @@ _migrate() {
...
-  exec ./artisan octane:start --host=0.0.0.0 "$@"
+    start_time_seconds="$(uptime_seconds)";
+    while [ ! -f "$MANIFEST_PATH" ]; do
+        elapsed_time_seconds="$(( $(uptime_seconds) - $start_time_seconds ))";
+        if [ "$elapsed_time_seconds" -ge "$MANIFEST_TIMEOUT_SECONDS" ]; then
+            echo_error "$0 [pid $$] [${elapsed_time_seconds}s]: The manifest does not exist.";
+            exit 1;
+        fi
+
+        echo_warning "$0 [pid $$] [${elapsed_time_seconds}s]: ${MANIFEST_PATH} not found.";
```
nanaya, line 50: the prefix is too long and I don't think it's necessary nor useful

docker/development/run.sh
```diff
@@ -3,6 +3,25 @@
...
+MANIFEST_TIMEOUT_INTERVAL_SECONDS=5;
+MANIFEST_TIMEOUT_SECONDS=90;
+
+echo_error() {
+    echo "${COLOR_RED}$@${COLOR_NONE}" >&2;
+}
+
+echo_warning() {
+    echo "${COLOR_YELLOW}$@${COLOR_NONE}" >&2;
+}
```
nanaya, line 19: seems overkill and the color kind of pollutes the log if it's output to a file

nanaya (changes requested):

docker/development/run.sh
AJGranowski: Should we just wait forever here?

docker/development/run.sh
AJGranowski: Sure thing.

docker/development/run.sh
AJGranowski: Built the habit after working with inline dockerfile scripts. I'll remove.

docker/development/run.sh
AJGranowski: Any suggestions? Keep in mind that this is only logged if the manifest is missing, which usually shouldn't happen.

I think knowing what's waiting, why, and for how long helps with understanding the state of the service during a non-normal startup.

docker/development/run.sh
AJGranowski: Sure, I can remove the coloring.

docker/development/run.sh
nanaya: yes. the octane itself runs forever anyway

docker/development/run.sh
nanaya: the service name is right there in the docker log output and `$0` is just the default entrypoint file so it's not telling anything either

just say "the manifest file not found" or maybe prefix with "waiting to start octane: "

AJGranowski: Any preference from these?

1. `./public/assets/manifest.json not found (123s)`
2. `manifest.json not found (123s)`
3. `manifest.json not found`
4. `waiting to start octane: ./public/assets/manifest.json not found (123s)`
5. `waiting to start octane: manifest.json not found (123s)`
6. `waiting to start octane: manifest.json not found`

nanaya: number 6 is probably fine. The duration can be checked through assets log anyway

docker/development/run.sh
```diff
@@ -20,7 +28,20 @@ _migrate() {
...
+    start_time_seconds="$(uptime_seconds)"
+    echo_counter=0
+    while [ ! -f "$MANIFEST_PATH" ]; do
+        if [ "$echo_counter" -le 0 ]; then
+            echo_counter="$MANIFEST_MISSING_LOG_STEPS"
+            elapsed_time_seconds="$(( $(uptime_seconds) - ${start_time_seconds} ))"
+            echo "${MANIFEST_PATH} not found (${elapsed_time_seconds}s)" >&2
+        fi
+
+        echo_counter="$(( ${echo_counter} - 1 ))"
```
nanaya, line 40: `: $(( echo_counter += 1 ))` should work on posix shell

AJGranowski: Oh neat!

docker/development/run.sh
```diff
@@ -20,7 +24,18 @@ _migrate() {
 }
 
 _octane() {
-  exec ./artisan octane:start --host=0.0.0.0 "$@"
+    echo_counter=0
+    while [ ! -f "$MANIFEST_PATH" ]; do
+        if [ "$echo_counter" -le 0 ]; then
+            echo_counter="$MANIFEST_MISSING_LOG_STEPS"
+            echo "waiting to start octane: $(basename "$MANIFEST_PATH") not found" >&2
```
nanaya, line 31: minor thing but there's no need to call external program just get basename
```suggestion
            echo "waiting to start octane: ${MANIFEST_PATH##*/} not found" >&2
```

docker/development/run.sh
```diff
@@ -3,6 +3,10 @@
 export CHROME_BIN=/usr/bin/chromium
 export DUSK_WEBDRIVER_BIN=/usr/bin/chromedriver
 
+MANIFEST_PATH='./public/assets/manifest.json'
+MANIFEST_MISSING_CHECK_INTERVAL=1
+MANIFEST_MISSING_LOG_STEPS=5 # Print an error log every (MANIFEST_MISSING_LOG_STEPS * MANIFEST_MISSING_CHECK_INTERVAL) seconds.
```
nanaya, line 8: plain shell variables should be lowercased. also I don't know if it's useful being here instead of inside the relevant function itself (and if the interval and step need to be variable in the first place)

nanaya (changes requested):
also please squash the commits

notbakaneko:
What's wrong with putting an empty `manifest.json` and letting the workers reload once the new one finishes building, instead of jumping through all the hoops checking for the file?

docker/development/run.sh
AJGranowski: I'm usually against magic numbers, but if you're fine with that we can inline them.

nanaya:
that's a good point 🤔

AJGranowski:
That seems to work! I just get served some error pages until it's finished.

screenshot

AJGranowski:
Here's an alternative PR:
* #11792

Either work, it just depends on the UX you're looking for when starting (either getting served nothing, or an error page during startup).

nanaya (approved):
let's just go with this one for now (there's different problem with the other one in its current state and I would probably take a different approach)

AJGranowski:
Sounds good! Thank you for being patent with me.

Issue #11597, opened 2024-10-29
Played variant for daily challenge counter


cl8n:
see [the design file](https://www.figma.com/design/vpLHaQ7CcjUuPH2hZiZ3WE/Quick-Fix%2FDaily-Challenge-Streak). this is a small update adding a variant of the daily challenge streak for when the user has completed today's challenge, already approved by flyte and myself

![](https://github.com/user-attachments/assets/cd5527a8-e796-423c-8885-df866a54c926)

also, the corner radius of that darker block is supposed to be 3px, it's 6px on osu-web rn for some reason.

I wasn't sure whether to just post it here or open a discussion at https://github.com/ppy/osu-design, the latter felt kind of overkill for such a tiny thing but maybe I'm overthinking it, I'm not sure. we're still figuring this out lol

this is a response to <https://discord.com/channels/188630481301012481/1097318920991559880/1300323179457151069>

![](https://github.com/user-attachments/assets/0a5a6c53-9c5b-4160-9140-57dc5de33d0c)

bdach:
I'll do the client change for this after the web pull has gone through review paces, since I'm not 100% on the web pull to begin with.

PR #11780, opened 2025-01-09
Add completion marker to daily challenge profile counter

bdach:
Implements https://github.com/ppy/osu-web/issues/11597 for the website.

### DIFF
--- a/resources/css/bem/daily-challenge.less
+++ b/resources/css/bem/daily-challenge.less
@@ -8,10 +8,11 @@
   position: relative;
   display: flex;
   align-items: center;
-  padding: 3px;
+  padding: 1px;
+  border: 2px solid transparent;
 
   &--played-today {
-    border: 2px solid @osu-colour-lime-1;
+    border-color: @osu-colour-lime-1;
 
     &::before {
       .fas();
@@ -19,10 +20,11 @@
       border-radius: 50%;
       color: @osu-colour-lime-1;
       content: @fa-var-check-circle;
-      font-size: 16px;
+      font-size: 16px; // icon size
       position: absolute;
-      right: -8px;
-      top: -8px;
+      right: 0;
+      top: 0;
+      transform: translate(50%, -50%);
     }
   }
 
--- a/resources/css/bem/daily-challenge.less
+++ b/resources/css/bem/daily-challenge.less
@@ -15,8 +15,8 @@
     border-color: @osu-colour-lime-1;
 
     &::before {
-      .center-content();
       .fas();
+      .center-content();
       background-color: @osu-colour-lime-1;
       border-radius: 50%;
       color: @osu-colour-b6;
--- a/resources/css/bem/daily-challenge.less
+++ b/resources/css/bem/daily-challenge.less
@@ -5,9 +5,29 @@
   background: hsl(var(--hsl-b4));
   border-radius: @border-radius-large;
   min-width: 0;
-  display: flex;
-  align-items: center;
-  padding: 3px;
+  position: relative;
+
+  &--played-today {
+    border: 2px solid @osu-colour-lime-1;
+
+    &::before {
+      .fas();
+      background-color: @osu-colour-b6;
+      border-radius: 50%;
+      color: @osu-colour-lime-1;
+      content: @fa-var-check-circle;
+      font-size: 16px;
+      position: absolute;
+      right: -8px;
+      top: -8px;
+    }
+  }
+
+  &__content {
+    display: flex;
+    align-items: center;
+    padding: 3px;
+  }
 
   &__name {
     font-size: @font-size--normal;
--- a/resources/css/bem/daily-challenge.less
+++ b/resources/css/bem/daily-challenge.less
@@ -15,16 +15,19 @@
     border-color: @osu-colour-lime-1;
 
     &::before {
+      .center-content();
       .fas();
-      background-color: @osu-colour-b6;
+      background-color: @osu-colour-lime-1;
       border-radius: 50%;
-      color: @osu-colour-lime-1;
-      content: @fa-var-check-circle;
-      font-size: 16px; // icon size
+      color: @osu-colour-b6;
+      content: @fa-var-check;
+      font-size: 8px; // icon size
+      height: 16px;
       position: absolute;
       right: 0;
       top: 0;
       transform: translate(50%, -50%);
+      width: 16px;
     }
   }
 
--- a/resources/css/bem/daily-challenge.less
+++ b/resources/css/bem/daily-challenge.less
@@ -6,6 +6,9 @@
   border-radius: @border-radius-large;
   min-width: 0;
   position: relative;
+  display: flex;
+  align-items: center;
+  padding: 3px;
 
   &--played-today {
     border: 2px solid @osu-colour-lime-1;
@@ -21,12 +24,6 @@
       right: -8px;
       top: -8px;
     }
-  }
-
-  &__content {
-    display: flex;
-    align-items: center;
-    padding: 3px;
   }
 
   &__name {
--- a/resources/css/bem/daily-challenge.less
+++ b/resources/css/bem/daily-challenge.less
@@ -40,7 +40,7 @@
   }
 
   &__value-box {
-    border-radius: @border-radius-large;
+    border-radius: @border-radius-small;
     background: hsl(var(--hsl-b6));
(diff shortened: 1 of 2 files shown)


resources/css/bem/daily-challenge.less
```diff
@@ -5,9 +5,29 @@
   background: hsl(var(--hsl-b4));
   border-radius: @border-radius-large;
   min-width: 0;
-  display: flex;
-  align-items: center;
-  padding: 3px;
+  position: relative;
+
+  &--played-today {
+    border: 2px solid @osu-colour-lime-1;
```
bdach, line 11: This is 1px on the [design file](https://www.figma.com/design/vpLHaQ7CcjUuPH2hZiZ3WE/Quick-Fix%2FDaily-Challenge-Streak) but I find that 1px looks terrible on the rounded corners:

| 1px | 2px |
| :-: | :-: |
|  |  |

Arguably even 2px looks not great. Maybe 3px even could be considered. cc @ppy/team-design

resources/css/bem/daily-challenge.less
```diff
@@ -20,7 +40,7 @@
   }
 
   &__value-box {
-    border-radius: @border-radius-large;
+    border-radius: @border-radius-small;
```
bdach, line 43: Adjusted as per mention in https://github.com/ppy/osu-web/issues/11597:

> also, the corner radius of that darker block is supposed to be 3px, it's 6px on osu-web rn for some reason.

resources/js/profile-page/daily-challenge.tsx
```diff
@@ -122,26 +123,30 @@ export default class DailyChallenge extends React.Component<Props> {
       return null;
     }
 
+    const playedToday = this.props.stats.last_update != null && moment.utc(this.props.stats.last_update).isSame(Date.now(), 'day');
```
bdach, line 126: I *think* this is gonna do timezones correctly:

https://momentjs.com/docs/#/query/is-same/

> If the two moments have different timezones, the timezone of the first moment will be used for the comparison.

and it looked correct when I messed around with my timezone...

resources/js/profile-page/daily-challenge.tsx
```diff
@@ -122,26 +123,30 @@ export default class DailyChallenge extends React.Component<Props> {
...
-        <div className='daily-challenge__value-box'>
-          <div
-            className='daily-challenge__value'
-            style={tierStylePlaycount(this.props.stats.playcount)}
-          >
-            {trans(
-              'users.show.daily_challenge.unit.day',
-              { value: formatNumber(this.props.stats.playcount) },
-            )}
+        <div className='daily-challenge__content'>
```
bdach, line 134: The reason for the existence of this `div` is that I need the `::before` pseudo-element that is the check in the corner to be excluded from the flexbox, it takes space otherwise. If you have better ideas on how to implement this I'm open to changes.

nanaya: absolute-positioned element, which the element is, doesn't take space...?

resources/css/bem/daily-challenge.less
nanaya: the extra spacing between the day count box and the green border seems a bit weird?

also dunno if matters but the popup box covers the check mark a little bit

resources/js/profile-page/daily-challenge.tsx
bdach: Huh. I guess I must have not tried some particular combination of css incantations because it does look to be not required now??? I love css.

resources/css/bem/daily-challenge.less
bdach: > the extra spacing between the day count box and the green border seems a bit weird?

I mean yeah maybe. I'm just following design direction here :shrug: 

> also dunno if matters but the popup box covers the check mark a little bit

Yeah it does and was also unsure if it mattered. I guess maybe the design peoples can say?

Just for visual reference here's a screenshot:

resources/js/profile-page/daily-challenge.tsx
nanaya: yeah that seems correct (until someone complains it's not)

alternatively `new Date(...).toISOString()` spits out iso8601 string which has utc timezone and can then be sliced _appropriately_ (not sure if better)

nanaya (changes requested):

resources/css/bem/daily-challenge.less
```diff
@@ -5,10 +5,27 @@
   background: hsl(var(--hsl-b4));
   border-radius: @border-radius-large;
   min-width: 0;
+  position: relative;
   display: flex;
   align-items: center;
   padding: 3px;
 
+  &--played-today {
+    border: 2px solid @osu-colour-lime-1;
```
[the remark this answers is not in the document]
bdach: In the process of applying this I noticed that the border even if transparent was also stacking onto the existing 3px of padding (see screenshots below, with the border green colour changed to transparent to show the difference):

| before | after |
| :-: | :-: |
|  |  |

which I didn't want to do and I don't think the design was going for, so I reduced the padding to 1px to compensate.

resources/js/profile-page/daily-challenge.tsx
bdach: > and can then be sliced _appropriately_ (not sure if better)

I uh... am not confident in my ability to do so, so I'd rather not if that's fine with you :sweat_smile:

resources/css/bem/daily-challenge.less
```diff
@@ -5,9 +5,28 @@
...
+  padding: 1px;
+  border: 2px solid transparent;
+
+  &--played-today {
+    border-color: @osu-colour-lime-1;
+
+    &::before {
+      .fas();
+      background-color: @osu-colour-b6;
+      border-radius: 50%;
```
nanaya, line 20: while at it, this doesn't actually do anything, does this? The fa icon is already rounded and all

bdach: The fontawesome icon is rounded but it has a hollow interior where the tick is. So without some hackage the border will peek out of the hollow interior, thus this + the `background-color` + `border-radius` specs are supposed to cover up anything that's in the background (particularly the border).

nanaya: oh. So that also explained the weirdly subtle black outline... 🤔 

I suppose an alternative would be to use normal check mark coupled with circle background, centered using `.center-content()`

bdach: Yeah that's a fair shout. Adjusted further in https://github.com/ppy/osu-web/pull/11780/commits/bf0333ab6ad91e4aa4b16b9c9d66c6a92c65697b.

I also took the opportunity to resize the tick itself because for whatever reason when using the pre-baked fontawesome `check-circle` the check was much larger than it actually was in the figma. It should be there-or-thereabouts matching now.

bdach:
> wait, is it only for own profile page or for any users? the design file says it's only for own page but this pr is currently for everyone

Yeah that was a reading fail... Attempted to address this in https://github.com/ppy/osu-web/pull/11780/commits/0600f1b781a6dd6c84d9ce6720f4bb03c92a8b7e. I [shamelessly stole the code to do this from somewhere else](https://github.com/ppy/osu-web/blob/f1d2b727cfd381bdf1dad2cc3b45218ce7729dc2/resources/js/profile-page/links.tsx#L172), let me know if that is fine.

resources/css/bem/daily-challenge.less
```diff
@@ -5,9 +5,31 @@
...
-  padding: 3px;
+  padding: 1px;
+  border: 2px solid transparent;
+
+  &--played-today {
+    border-color: @osu-colour-lime-1;
+
+    &::before {
+      .center-content();
+      .fas();
```
nanaya, line 19: the order is kinda important as both define `display:` 😬

resources/css/bem/daily-challenge.less
```diff
@@ -5,9 +5,31 @@
...
+      border-radius: 50%;
+      color: @osu-colour-b6;
+      content: @fa-var-check;
+      font-size: 8px; // icon size
+      height: 16px;
+      position: absolute;
+      right: 0;
+      top: 0;
+      transform: translate(50%, -50%);
+      width: 16px;
```
nanaya, line 30: can use `$height` so only need to define it once

nanaya (changes requested):

PR #11781, opened 2025-01-09
Add ability to search by circle or slider count

p2d0:
to search for stream maps like length<90 circles>500
ideally there would be something like cpm (circles per minute) but that would require bigger changes.


peppy:
Any reason for this being closed before it was reviewed?

p2d0:
> Any reason for this being closed before it was reviewed?

I changed branch name it autoclosed, i created same one

PR #11786, opened 2025-01-10
Add interop endpoint to create and join multiplayer room

nanaya:
To be used by some new multiplayer system in lazer.


smoogipoo:
Fwiw, I requested a review of myself because I was going to apply or suggest small changes as I get server-spectator side of things working using this. I'll just PR those changes separately now.

notbakaneko:
Oh, I saw that and forgot about it when pressing the merge button 😅

PR #11791, opened 2025-01-14
Don't preload singletons

nanaya:
Alternative fix for #11779. I think I put it there for benchmarking purpose but otherwise it's probably fine to init the singletons on first access.

### DIFF
--- a/config/octane.php
+++ b/config/octane.php
@@ -1,7 +1,6 @@
 <?php
 
 use App\Listeners\OctaneResetLocalCache;
-use App\Providers\AppServiceProvider;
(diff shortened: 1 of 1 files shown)


notbakaneko:
If they're not in `warm` they get constructed every request which defeats the purpose of the ones like `Mods`

nanaya:
but then how did the `LOCAL_CACHE_SINGLETONS` survive 🤔

nanaya:
holy fuck that was some braindead logic there. nvm then

https://github.com/laravel/octane/issues/265#issuecomment-841507887

PR #11801, opened 2025-01-20
Always wrap Beatmap Discussion guest mapper list

### LINKED ISSUE #11798: Overflowing containers in Discussion page depending on language chosen
As seen [in the following Discussion page](https://osu.ppy.sh/beatmapsets/769661/discussion/-/generalAll/total) if `lang="ca"`, if the contributors list is too long, the objects inside the `<span>` in `beatmap-discussions-header-top__owners` will overflow with its grid sibling `beatmap-discussions-header-top__basic-stats`.

Other languages do not have this issue as a white space is properly added after every comma, leading me to believe the localized string for the language is either not adding it, or overwriting the white space.

Adding the white spaces solved the issue:

notbakaneko:
fixes #11798

But the localization should still be updated to include a space.

### DIFF
--- a/resources/js/beatmap-discussions/header.tsx
+++ b/resources/js/beatmap-discussions/header.tsx
@@ -177,14 +177,12 @@ export class Header extends React.Component<Props> {
           <div className={`${bn}__beatmap-stats`}>
             <div className={`${bn}__owners`}>
               {hasGuestOwners(this.currentBeatmap, this.beatmapset) && (
-                <>
-                  <StringWithComponent
-                    mappings={{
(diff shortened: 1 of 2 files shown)


resources/js/beatmap-discussions/header.tsx
```diff
@@ -177,14 +177,14 @@ export class Header extends React.Component<Props> {
           <div className={`${bn}__beatmap-stats`}>
             <div className={`${bn}__owners`}>
               {hasGuestOwners(this.currentBeatmap, this.beatmapset) && (
-                <span>
+                <>
```
nanaya, line 180: not needed?

nanaya (approved):

PR #11804, opened 2025-01-20
Add seasonal leaderboards

venix12:
Depends on #11768.

This adds per-season rankings to [current seasons page](https://osu.ppy.sh/seasons/latest), including seasonal total score (as of #11768) and assigned division display (calculation of which is part of this PR).

image

---

``image_url`` column of ``divisions`` table isn't currently being utilized for purpose of the ranking display, but will be used for user profile division display in the future.


peppy:
@venix12 able to add some screenshots and  showing what we're adding here? 😄

venix12:
> @venix12 able to add some screenshots showing what we're adding here? 😄

updated the opening message accordingly :eye:

app/Models/Season.php
```diff
@@ -45,6 +48,36 @@ public static function latestOrId($id)
...
+    public function divisions(): HasMany
+    {
+        return $this->hasMany(Division::class);
+    }
+
+    public function divisionsOrdered(): Collection
+    {
+        return cache_remember_mutexed(
+            "divisions:{$this->id}",
+            $GLOBALS['cfg']['osu']['seasons']['divisions_cache_duration'],
```
venix12, line 60: saw duration being pulled from config/env in other places (tags) recently so just mimicked that, but wondering if this is actually needed here to begin with :thinking:

nanaya: no it's not. the other place involves some aggregation and stuff

nanaya:
what's this division/threshold/absolute threshold thing here? What's the difference between absolute and not?

venix12:
Divisions are tiers that are assigned to each of season's players based on their rank.

The divsisions (set per-season) are assigned for top x% of season players, therefore:
 - ``threshold`` is a (0-1) fraction meant to represent top amount of season's playerbase that should be assigned the division (e.g. 0.05, so 5%), it's stored in divisions table
 - absolute threshold is amount actual amount of players that will be assigned said division, aka. ``threshold * season players count``, dynamically calculated based on current player count

threshold column should probably use decimal(4,3) or so instead of float now that i think about this 🤔

resources/views/seasons/_rankings_table.blade.php
```diff
@@ -0,0 +1,79 @@
...
+                        if ($rank <= $division['absolute_threshold']) {
+                            $currentDivision = $division;
+                            break;
+                        }
+
+                        // this shouldn't ever happen if divisions are set up properly :)
+                        $currentDivision = null;
+                    }
+                }
+            @endphp
```
nanaya, line 40: less looping the divisions
```suggestion
        @foreach ($scores as $index => $score)
            @php
                $rank = $scores->firstItem() + $index;

                if (!isset($currentDivision) || $rank > $currentDivision['absolute_threshold']) {
                    while (($currentDivision = array_shift($divisions)) !== null) { {
                        if ($rank <= $currentDivision['absolute_threshold']) {
                            break;
                        }
                    }
                }
            @endphp
```

app/Models/Season.php
```diff
@@ -45,6 +48,28 @@ public static function latestOrId($id)
...
+    {
+        $divisions = $this->divisions()->orderBy('threshold')->get();
+        $userCount = $this->userScores()->count();
+
+        $divisionsWithThresholds = [];
+
+        foreach ($divisions as $division) {
+            $divisionsWithThresholds[] = [
+                'division' => $division,
+                'absolute_threshold' => $division->threshold * $userCount,
```
nanaya, line 66: maybe call it max_rank?

app/Models/Season.php
```diff
@@ -45,6 +48,28 @@ public static function latestOrId($id)
...
 
+    public function divisions(): HasMany
+    {
+        return $this->hasMany(Division::class);
+    }
+
+    public function divisionsWithAbsoluteThresholds(): array
+    {
+        $divisions = $this->divisions()->orderBy('threshold')->get();
+        $userCount = $this->userScores()->count();
```
nanaya, line 59: I believe this should have `forRanking` filter just like the actual table

also the value is already queried by the pagination thing...

resources/views/seasons/_rankings_table.blade.php
```diff
@@ -0,0 +1,79 @@
...
+                        </a>
+                    </div>
+                </td>
+                <td class="ranking-page-table__column ranking-page-table__column--dimmed">
+                    {!! i18n_number_format($score->total_score) !!}
+                </td>
+                <td class="ranking-page-table__column ranking-page-table__column--division">
+                    @if (isset($currentDivision))
+                        {{ $currentDivision['division']->name }}
+                    @endif
```
nanaya, line 74: just use `?->name` and skip the null check

resources/views/seasons/_rankings_table.blade.php
```diff
@@ -0,0 +1,79 @@
...
+                            class="ranking-page-table__user-link-text js-usercard"
+                            data-user-id="{{ $score->user_id }}"
+                            data-tooltip-position="right center"
+                        >
+                            {{ $score->user->username }}
+                        </a>
+                    </div>
+                </td>
+                <td class="ranking-page-table__column ranking-page-table__column--dimmed">
+                    {!! i18n_number_format($score->total_score) !!}
```
nanaya, line 69: should be escaped (`{{`)

resources/views/seasons/_rankings_table.blade.php
```diff
@@ -0,0 +1,79 @@
...
+                            break;
+                        }
+
+                        // this shouldn't ever happen if divisions are set up properly :)
+                        $currentDivision = null;
+                    }
+                }
+            @endphp
+
+            <tr class="ranking-page-table__row{{$score->user->isActive() ? '' : ' ranking-page-table__row--inactive'}}">
```
nanaya, line 42: use `class_with_modifiers` helper

app/Models/UserSeasonScoreAggregate.php
```diff
@@ -62,8 +64,23 @@ public function calculate(bool $muteExceptions = true): void
         $this->total_score = $total;
     }
 
+    public function scopeForRanking($query): Builder
+    {
+        return $query
+            ->whereHas('user', function ($userQuery) {
+                $userQuery->default();
+            })
```
nanaya, line 72: use the shorter form

nanaya (changes requested):

resources/views/seasons/_rankings_table.blade.php
venix12: that will throw error on trying to access division key on null, won't it? 🤔

nanaya: oh I was trying it on a different thing before where the variable is assigned actual model. an `?? ''` at the end works

database/migrations/2025_01_06_203319_create_season_divisions_table.php
```diff
@@ -0,0 +1,37 @@
+<?php
```
nanaya, line 1: migration date needs an update

database/migrations/2025_01_06_203319_create_season_divisions_table.php
```diff
@@ -0,0 +1,37 @@
...
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('season_divisions', function (Blueprint $table) {
+            $table->id();
+            $table->integer('season_id')->unsigned();
+            $table->string('name');
+            $table->string('image_url');
+            $table->decimal('threshold', 4, 3);
```
nanaya, line 24: this results in string (including when casted as decimal) which isn't wanted. Operating on it properly is also a bit verbose. And then the result is rounded anyway...

A double is probably fine (or cast to double)

app/Models/SeasonDivision.php
```diff
@@ -0,0 +1,20 @@
...
+declare(strict_types=1);
+
+namespace App\Models;
+
+/**
+ * @property int $id
+ * @property string $image_url
+ * @property string $name
+ * @property int $season_id
+ * @property int $threshold
```
nanaya, line 15: not int

app/Models/Season.php
```diff
@@ -45,6 +48,28 @@ public static function latestOrId($id)
...
+    {
+        $divisions = $this->divisions()->orderBy('threshold')->get();
+        $userCount ??= $this->userScores()->forRanking()->count();
+
+        $divisionsWithMaxRanks = [];
+
+        foreach ($divisions as $division) {
+            $divisionsWithMaxRanks[] = [
+                'division' => $division,
+                'max_rank' => $division->threshold * $userCount,
```
nanaya, line 66: can use an explicit rounding and casting to int

app/Models/SeasonDivision.php
```diff
@@ -0,0 +1,20 @@
...
+declare(strict_types=1);
+
+namespace App\Models;
+
+/**
+ * @property int $id
+ * @property string $image_url
+ * @property string $name
+ * @property int $season_id
+ * @property double $threshold
```
nanaya, line 15: it's all float in php

app/Models/Season.php
```diff
@@ -45,6 +48,28 @@ public static function latestOrId($id)
...
+    {
+        $divisions = $this->divisions()->orderBy('threshold')->get();
+        $userCount ??= $this->userScores()->forRanking()->count();
+
+        $divisionsWithMaxRanks = [];
+
+        foreach ($divisions as $division) {
+            $divisionsWithMaxRanks[] = [
+                'division' => $division,
+                'max_rank' => get_int($division->threshold * $userCount),
```
nanaya, line 66: it works but it's not quite what get_int is for... ಠ_ಠ (it's more for validating unknown variable/string, not a known float)

resources/views/seasons/_rankings_table.blade.php
```diff
@@ -0,0 +1,69 @@
...
+                        if ($rank <= $currentDivision['max_rank']) {
+                            break;
+                        }
+                    }
+                }
+            @endphp
+
+            <tr class="{{ class_with_modifiers('ranking-page-table__row', ['inactive' => !$score->user->isActive()]) }}">
+                <td class="ranking-page-table__column ranking-page-table__column--rank">
+                    #{{ $rank }}
```
nanaya, line 36: can use i18n formatting

resources/views/seasons/_rankings_table.blade.php
```diff
@@ -0,0 +1,69 @@
...
+            <th class="ranking-page-table__heading ranking-page-table__heading--focused">
+                {{ osu_trans('rankings.stat.division') }}
+            </th>
+        </tr>
+    </thead>
+
+    <tbody>
+        @foreach ($scores as $index => $score)
+            @php
+                $rank = $scores->firstItem() + $index;
```
nanaya, line 23: minor thing but maybe store the `firstItem` value beforehand

resources/views/seasons/show.blade.php
```diff
@@ -34,6 +34,10 @@ class="btn-osu-big btn-osu-big--rounded-thin"
     <div class="js-react--seasons-show"></div>
 @endsection
 
+@section('scores')
+    @include('seasons._rankings_table', [$divisions, $scores])
```
nanaya, line 38: that's not how passing variable works. It should be `['var' => $thing]` to access it as `$var` in the included view. It only seemed to work because all variables from current context are passed through.

nanaya (approved):

PR #11809, opened 2025-01-22
Adjust username change cost

### LINKED ISSUE #11790: Change the way name changes are priced
> Hi, I would like to share my opinion on the name change feature and a way I think it should be changed.
> 
> Right now, this is how the pricing works. However, this price will always stay the same whenever you want to > change your username, no matter how long a time it has been since you bought a name change last.
> 
> It says this change was made to avoid people cyling through names quickly, but 100$ to change your username 6 > times is a lot if you have played the game for a long time.
> 
> I think that, if you haven’t changed your username for a certain period of time (my suggestion is a year, but it  could be wherever is fair), the “counter” you see above pricing should decrease by 1, and reaching a minimum of 2. For example, if that counter was on 4 ($32), and I didn’t change my name for a year, it would go down to $16, and after another year it goes down to $8.
>
> I understand that username changes are a part of what funds osu, and if it would be too costly to change this then I understand, but I would appreciate if you could give it some thought.

I've heard this a few times over the last couple of months and tend to agree with it.

Let's change name changes to reduce the pricing tier for each year that has passed since the last name change (but never return to free).

notbakaneko:
closes #11790

The new logic is the price goes up 1 tier for every paid change, and down 1 tier for every year since the last paid change, to a minimum of 1.

### DIFF
--- a/tests/Models/UserTest.php
+++ b/tests/Models/UserTest.php
@@ -17,6 +17,17 @@ use Tests\TestCase;
 
 class UserTest extends TestCase
 {
+    public static function dataProviderForAttributeTwitter(): array
+    {
+        return [
+            ['@hello', 'hello'],
+            ['hello', 'hello'],
+            ['@', null],
+            ['', null],
+            [null, null],
+        ];
+    }
+
     public static function dataProviderForUsernameChangeCost()
     {
         return [
@@ -68,6 +79,30 @@ class UserTest extends TestCase
         ];
     }
 
+    public static function dataProviderValidDiscordUsername(): array
+    {
+        return [
+            ['username', true],
+            ['user_name', true],
+            ['user.name', true],
+            ['user2name', true],
+            ['u_sernam.e1337', true],
+            ['username#', false],
+            ['u', false],
+            ['morethan32characterinthisusername', false], // 33 characters
+
+            // old format
+            ['username#1337', true],
+            ['ユーザー名#1337', true],
+            ['username#1', false],
+            ['username#13bb', false],
+            ['username#abcd', false],
+            ['user@name#1337', false],
+            ['user#name#1337', false],
+            ['user:name#1337', false],
+        ];
+    }
+
     /**
      * @dataProvider dataProviderForAttributeTwitter
      */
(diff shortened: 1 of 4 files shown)


tests/Models/UserTest.php
```diff
@@ -65,24 +173,64 @@ public function testUsernameAvailableAtForNonDefaultGroup()
...
+
+        $user = User::factory()
+            ->has(UsernameChangeHistory::factory()->count(2))
+            ->create();
+
+        $this->travelBack();
+
+        UsernameChangeHistory::factory()->state(['type' => $type, 'user_id' => $user])->create();
+
+        $this->assertSame($user->usernameChangeCost(), $cost);
```
nanaya, line 233: the function signature is `$expected, $actual` so it should be `$cost, $user->...` here (and the rest of this file).

nanaya (changes requested):

nanaya (approved):

Phippe:
Once someone has changed their name 5 times or more, the default cost will be $100.

Now assume they wait for 4 years and then pay $8 for another name change. Shortly afterwards, they want to change their name again, but - if I understand the code correctly - now the price is back up to $100, not $16.

When I first read the proposal, I interpreted the cooldown to be more "permanent". Like once you've waited long enough, you would start at tier 1 again and then "work your way back up" for each name change after that.

I just wanted to ask if this is the intended behaviour.

nanaya:
yeah I think it's an oversight. A simpler logic would probably be to just set the tier to be the number of name changes over the last 4-5 years. @notbakaneko

peppy:
I guess we should wait for the oversight to be addressed before deploying to stop people getting confused.

notbakaneko:
After going through it again, it's a lot more complicated to handle if the cost tier gets 'reset' to a new level each time.
Consider the case where the username is changed once a year:
If you go to the highest tier then the cost is $100 (tier 5), after a year the price drops to $64 (tier 4) moving the next change to tier 5 but if you wait a year again, it drops to tier 4.
The same applies to the other tiers where it'll just stay the same price if you wait a year before every change.
The $8 tier also requires an exception for handling it since it can't go below $8.

It doesn't seem practical without going through every username change in the user's history and calculating the current tier the name change should put it at, or storing the current tier.

I think it might make more sense if it was based on username changes in the past X years and maybe a reduction based on when the last change was made.

peppy:
Just basing it on number of changes in last 3 years is fine IMO.

PR #11810, opened 2025-01-22
Add ability to join team

nanaya:
I think it's in relatively good state now 🤔 

~~- [ ] better notification message and everything else~~ improvements for the future 👀


notbakaneko:
I think the team leader needs some kind of indicator that there are pending join requests because you don't see anything unless you go into the member management page...

resources/views/teams/show.blade.php
```diff
@@ -90,8 +92,42 @@ class="{{ class_with_modifiers('team-action-button', 'part', ['disabled' => !$ca
...
+            @elseif ($currentUser?->teamApplication?->team_id === $team->getKey())
+                <form
+                    action="{{ route('teams.applications.destroy', ['team' => $team, 'application' => $currentUser->getKey()]) }}"
+                    data-turbo-confirm="{{ osu_trans('common.confirmation') }}"
+                    data-reload-on-success="1"
+                    method="POST"
+                >
+                    <input type="hidden" name="_method" value="DELETE" />
+                    <button
+                        class="team-action-button team-action-button--join"
```
notbakaneko, line 104: There's no css for `join`, though?

notbakaneko (changes requested):

app/Singletons/OsuAuthorize.php
```diff
@@ -1905,6 +1906,46 @@ public function checkScorePin(?User $user, ScoreBest|Solo\Score $score): string
...
+        $this->ensureLoggedIn($user);
+
+        $team = $application->team;
+
+        if ($team->leader_id !== $user->getKey()) {
+            return null;
+        }
+        if ($team->emptySlots() < 1) {
+            return 'team.member.store.full';
+        }
```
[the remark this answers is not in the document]
nanaya: 🤷

app/Jobs/Notifications/TeamApplicationBase.php
```diff
@@ -0,0 +1,51 @@
...
+use App\Models\Team;
+use App\Models\TeamApplication;
+use App\Models\User;
+
+abstract class TeamApplicationBase extends BroadcastNotificationBase
+{
+    const DELIVERY_MODE_DEFAULTS = ['mail' => true, 'push' => true];
+
+    private Team $team;
+    private int $userId;
```
notbakaneko, line 18:
```suggestion
    protected Team $team;
    protected int $userId;
```

notbakaneko (changes requested):

resources/lang/en/authorization.php
```diff
@@ -192,6 +192,15 @@
     ],
 
     'team' => [
+        'application' => [
+            'store' => [
+                'already_member' => "You're already part of the team.",
+                'already_other_member' => "You're already part of a different team.",
+                'currently_applying' => 'You have pending team join request.',
+                'team_closed' => 'The team is currently not accepting any join requests.',
+                'team_full' => "The team is full and can't accept any more members.",
```
notbakaneko, line 201: I forgot to mention, a (long) while ago it was discussed about always using single quotes for all the localization files and escaping as necessary? I forget the specific reason for going with single quotes only @peppy

app/Singletons/OsuAuthorize.php
```diff
@@ -1905,6 +1906,46 @@ public function checkScorePin(?User $user, ScoreBest|Solo\Score $score): string
...
+    {
+        $this->ensureLoggedIn($user);
+
+        $team = $application->team;
+
+        if ($team->leader_id !== $user->getKey()) {
+            return null;
+        }
+        if ($team->emptySlots() < 1) {
+            return 'team.member.store.full';
```
notbakaneko, line 1919: wrong/missing key

Also not sure if admin should be skipping the team slots check.

notbakaneko (changes requested):

app/Singletons/OsuAuthorize.php
nanaya: I think it's fine for admin to be able to skip that 🤔 it's not like some breaking/validation thing (and it's not even a static number in the first place)

notbakaneko (approved):

notbakaneko:
I though I mentioned it somewhere but I can't find it so maybe I didn't add the message? 🤨 

Should there be a cooldown with how frequently a user can request to join the same team on getting rejected?

nanaya:
🤷 a blocklist would probably be more useful instead of cooldown, I think?

PR #11811, opened 2025-01-23
Add support for parting rooms via interop + adjust responses

smoogipoo:
Builds on top of https://github.com/ppy/osu-web/pull/11786

`osu-server-spectator`:
- Will call the new endpoint to part rooms.
- Only needs the room ID when creating rooms, as it does a DB lookup of its own. I want to avoid deserialising osu!web models here because that stuff is messy osu!-side.
- Does not need a resultant value returned from joining/parting rooms.


app/Http/Controllers/Multiplayer/RoomsController.php
```diff
@@ -138,12 +138,14 @@ public function leaderboard($roomId)
 
     public function part($roomId, $userId)
     {
+        $currentUser = \Auth::user();
         // this allows admins/host/whoever to remove users from games in the future
-        if (get_int($userId) !== auth()->user()->user_id) {
+        if (get_int($userId) !== $currentUser->getKey()) {
             abort(403);
```
smoogipoo, line 144: Adjusted to follow the same structure as `join()`, which I assume was adjusted for a reason in https://github.com/ppy/osu-web/pull/11786.

nanaya (approved):

PR #11817, opened 2025-01-26
put header-nav-v4 in front of blurred menu image

WebFreak001:
This adds a `z-index: 1` to the `dashboard   friends   watchlists   account settings` menu at the top on the start page, so it's in front of the blurred background image that is currently on top of the menu.

This makes the colors pop more as well as making the border color consistent on the active item as well as just the rest of the bar.

Before:

image

After:

image

---

<details>
<summary>Artificial purple image</summary>

The effect is most notable when the menu image has a lot of dark purple / areas of color close to the menu bar color (example using an image of just a solid purple block):

Before:

image

After:

image

</details>

---

<details>
<summary>Artificial white image</summary>

However it might not look as good with very light / white headers: (border color now matches the big blob though)

Before:

image

After:

image

</details>

### DIFF
--- a/resources/css/bem/header-nav-v4.less
+++ b/resources/css/bem/header-nav-v4.less
@@ -14,6 +14,7 @@
   padding: 0;
   font-size: @font-size--normal;
   position: relative;
+  z-index: 1;
 
   @media @desktop {
     --header-nav-item-padding-y: 15px;


peppy:
I'm on board with this change visually 👍

nanaya:
it's inconsistent with the the below the image though. If anything I'd rather just remove the extended highlight or reduce its size (or some other adjustments).

PR #11823, opened 2025-01-28
Support ruleset variants in "best of" contests

nanaya:
Maybe? ಠ_ಠ

Add a variant field in the best_of object.


notbakaneko (approved):
Seems to work 🤔 
...also noticed it only considers existence of the playcount row and not if the value is actually `> 0` but that's a different problem 👀

PR #11825, opened 2025-01-28
Fix docs nav button resetting scroll position

nanaya:
I wonder if selecting anything in the navigation should also close it instead of having to close manually.


notbakaneko (changes requested):
This resets the scroll position and doesn't open the menu for me ?_?

nanaya:
what's the error? have you rebuild the docs after assets rebuild?

notbakaneko:
oh it loads nothing without a rebuild ( ﾟ ヮﾟ)

notbakaneko (approved):

PR #11827, opened 2025-01-28
Team member leaderboard

nanaya:
I think this is pretty close to the design?

Added a lone `info` link in the header to put the ruleset selector on (mainly for mobile). Will add link to member listing and settings page for admin later (and then remove the "back" link in respective pages).

depends on
- [x] #11826
- [x] #11837

### DIFF
--- a/app/Http/Controllers/TeamsController.php
+++ b/app/Http/Controllers/TeamsController.php
@@ -7,7 +7,10 @@ declare(strict_types=1);
 
 namespace App\Http\Controllers;
 
+use App\Exceptions\InvariantException;
+use App\Models\Beatmap;
 use App\Models\Team;
+use App\Models\User;
 use App\Transformers\UserCompactTransformer;
 use Symfony\Component\HttpFoundation\Response;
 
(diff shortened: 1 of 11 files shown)


notbakaneko:
I feel like the leaderboard should be its own tab; the members section in the middle being variable height based on the window height seems like a weird behavior now... 🤔

nanaya:
drafting this to move to its own tab (and thus depend on #11837)

app/Http/Controllers/TeamsController.php
```diff
@@ -66,6 +74,28 @@ public function edit(string $id): Response
...
+    {
+        $team = Team::findOrFail($id);
+        $ruleset ??= Beatmap::modeStr($team->default_ruleset_id);
+        $statisticsRelationName = User::statisticsRelationName($ruleset);
+        if ($statisticsRelationName === null) {
+            throw new InvariantException(osu_trans('beatmaps.invalid_ruleset'));
+        }
+        $leaderboard = $team
+            ->members
+            ->loadMissing("user.{$statisticsRelationName}")
```
notbakaneko, line 87: also need to load/set the team and userGroups

notbakaneko (changes requested):

app/Http/Controllers/TeamsController.php
nanaya: more like too much copypaste 👀

notbakaneko (changes requested):
I think it needs some adjustment for mobile 👀 
image

notbakaneko (approved):
eta someone with 800px with complains about the space 👀

PR #11830, opened 2025-01-30
More usable team settings page on mobile

nanaya:
I forgot it's a thing.


notbakaneko (changes requested):
I think the description textarea needs to be a bit more than 2 lines tall and the preview doesn't need to be so big when there's nothing in it 👀

PR #11831, opened 2025-01-31
Support non-4k/7k mania variant in "best of" contests

nanaya:
As requested.

The variant would be "nk"


notbakaneko (approved):
seems right 🤔

PR #11834, opened 2025-02-02
Add missing localisation to beatmap pack listing

TPGPL:
maybe would be better to do something like `by <strong>:author</strong>` but /shrug

### DIFF
--- a/resources/views/packs/index.blade.php
+++ b/resources/views/packs/index.blade.php
@@ -27,7 +27,7 @@
                         <div class="beatmap-pack__name">{{ $pack->name }}</div>
                         <div class="beatmap-pack__details">
                             <span class="beatmap-pack__date">{{ json_date($pack->date) }}</span>
-                            <span>{!! osu_trans('beatmappacks.show.created_by', ['author' => "<strong>{$pack->author}</strong>"]) !!}</span>
+                            <span>{!! osu_trans('beatmappacks.show.created_by', ['author' => "<strong>{make_blade_safe($pack->author)}</strong>"]) !!}</span>
                         </div>
                     </a>
                     <div class="beatmap-pack__body js-accordion__item-body"></div>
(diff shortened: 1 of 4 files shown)


notbakaneko (changes requested):
make it `by :author` and pass the bolded text through the variable

resources/views/packs/index.blade.php
```diff
@@ -27,7 +27,7 @@
                         <div class="beatmap-pack__name">{{ $pack->name }}</div>
                         <div class="beatmap-pack__details">
                             <span class="beatmap-pack__date">{{ json_date($pack->date) }}</span>
-                            <span>{{ osu_trans('beatmappacks.show.created_by', ['author' => "<strong>{$pack->author}</strong>"]) }}</span>
+                            <span>{!! osu_trans('beatmappacks.show.created_by', ['author' => "<strong>{$pack->author}</strong>"]) !!}</span>
```
notbakaneko, line 30: `author` should be escaped and wrapped in `HtmlString`, use `make_blade_safe`; same for the other file

notbakaneko (changes requested):

resources/views/packs/index.blade.php
```diff
@@ -27,8 +27,7 @@
                         <div class="beatmap-pack__name">{{ $pack->name }}</div>
                         <div class="beatmap-pack__details">
                             <span class="beatmap-pack__date">{{ json_date($pack->date) }}</span>
-                            <span class="beatmap-pack__author">by </span>
-                            <span class="beatmap-pack__author beatmap-pack__author--bold">{{ $pack->author }}</span>
+                            <span>{!! osu_trans('beatmappacks.show.created_by', ['author' => "<strong>{make_blade_safe($pack->author)}</strong>"]) !!}</span>
```
notbakaneko, line 30: This is just putting the function call into the string....

Also looking at it again, it can just use `e` instead of `make_blade_safe`/wrapping in `HtmlString` since it's being used in the blade itself and not being passed anywhere; just use the `tag` helper `tag('stong', content: e($pack->author))`

notbakaneko (changes requested):

notbakaneko (approved):

PR #11841, opened 2025-02-06
Reject submission of score 0

nanaya:
I think this may be confusing if client is still sending such scores? Note that this applies to solo scores as well.

Reference: https://github.com/ppy/osu/issues/31808#issuecomment-2639579715


peppy:
We'll make client side changes to also not submit in this case, but it's good to have this on osu-web as well.

notbakaneko (approved):

PR #11842, opened 2025-02-07
Update Shopify orders with more details

notbakaneko:
Separated to isolate the non-sdk related changes.
Draft because the Shopify dev environment doesn't work with the deprecated shopify sdk at the moment.

### DIFF
--- a/app/Http/Controllers/Payments/ShopifyController.php
+++ b/app/Http/Controllers/Payments/ShopifyController.php
@@ -70,9 +70,10 @@ class ShopifyController extends Controller
                 // TODO?
                 // admin_graphql_api_id (gid://shopify/Order/xxxxx)
                 // order_status_url
-                $orderNumber = $this->getOrderNumber();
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/Payments/ShopifyController.php
```diff
@@ -100,6 +104,29 @@ private function getOrderId()
...
+            if ($value === null) {
+                app('sentry')->getClient()->captureMessage(
+                    "Missing {$key} in Shopify webhook.",
+                    new Severity(Severity::WARNING),
+                    (new Scope())->setExtra('order_id', $this->getOrderId())
+                );
+            }
+        }
+
+        // Don't overwrite existing values with null/empty string later.
```
nanaya, line 126: or string '0' 👀

notbakaneko: 🤔

PR #11844, opened 2025-02-07
Add ruleset_id column to seasons table

venix12:

Prerequisite for displaying season division/rank on user profile.


nanaya:
how's it going to be queried exactly?

venix12:
for now, probably something like ``Season::active()->where('ruleset_id', $rulesetId)->first()``

nanaya:
The question was actually for all the way for whatever that needs to be displayed on the user page, not just this specific model. But I guess it's not too bad if just one.

venix12:
```php
public static function forSeason(User $user, Season $season): ?array {
    $score = $season->userScores()
        ->whereBelongsTo($user)
        ->first();
    
    if ($score === null) {
        return null;
    }
    
    $rank = $season->userScores()
        ->where('user_id', '<>', $score->user_id)
        ->where('total_score', '>=', $score->total_score)
        ->count() + 1;
    
    foreach ($season->divisionsWithMaxRanks() as $division) {
        if ($rank <= $division['max_rank']) {
            $userDivision = $division['division'];
            break;
        }
    }
      
    if (!isset($userDivision)) {
        return null;
    }
    
    return [
        'division' => json_item($userDivision, new SeasonDivisionTransformer()),
        'rank' => $rank,
        'season' => json_item($season, new SeasonTransformer()),
        'total_score' => $score->total_score,
    ];
}
```
Got something barebones like this rolling at the very moment, would also be cached somehow probably.

Some seasonal rank history display would be nice in the future, though probably out of scope at the very moment.

database/migrations/2025_02_07_172121_add_ruleset_id_to_seasons.php
```diff
@@ -0,0 +1,33 @@
+<?php
```
nanaya, line 1: please update the file date

nanaya (changes requested):

nanaya (approved):

PR #11845, opened 2025-02-09
Convert changelog chart to TypeScript

agatemosu:
~~The tooltip text now shows "null" instead of being empty in the dev env, matching the header nav item.~~

I also renamed the tooltip methods to start with the same word.


resources/js/charts/changelog-chart.ts
```diff
@@ -0,0 +1,307 @@
...
+import * as d3 from 'd3';
+import { first, groupBy, kebabCase, last, map, sumBy } from 'lodash';
+import moment from 'moment';
+import { fadeIn, fadeOut } from 'utils/fade';
+import { formatNumber } from 'utils/html';
+import { parseJson } from 'utils/json';
+
+interface BuildHistory {
+  created_at: string;
+  label: string | null;
```
nanaya, line 13: `null` label isn't supposed to happen. If it does, the data is wrong in the first place.

nanaya (changes requested):

resources/js/charts/changelog-chart.ts
```diff
@@ -0,0 +1,308 @@
...
+  normalized: number;
+  user_count: number;
+}
+
+interface ChartConfig {
```
nanaya, line 18: what I meant was calling something that contains actual data "config" isn't quite right

resources/js/charts/changelog-chart.ts
```diff
@@ -0,0 +1,308 @@
...
+
+    const pos = d3
+      .bisector<Datum, Date>((d) => d.data.date)
+      .left(this.data[0], x);
+
+    let dataRow = 0;
+    let currentLabel = '';
+    let labelModifier = '';
+
+    for (const [i, el] of this.data.entries()) {
```
nanaya, line 270: if you look closer, the `i` isn't needed; just assign the `el[pos]` to something directly.

nanaya (changes requested):

nanaya (approved):
the chart thing can use some actual refactor but this will do for now

PR #11846, opened 2025-02-10
Update Shopify related checkouts to use Shopify Storefront API

notbakaneko:
The Shopify Checkout API is deprecated and being removed, so we have to update to the Storefront API instead. The Storefront API has more integration features, but it's also more inconvenient and less suitable for our current usage 🤷 .

Part of the update involves updating the existing checkout ids the old ids are probably not going to work anymore once the Checkout API is removed.

The `@shopify/hydrogen-react` dependency is entirely only for the types.

- [x] command to get order details from Storefront API
- [x] apparently there are webhook versions too 🤨  (seems fine)

`store:migrate-shopify-checkouts` needs to be run to migrate the old `Checkout` ids to the new ones used by the Storefront API.


resources/js/store.ts
```diff
@@ -62,15 +63,42 @@ export class Store {
...
+            }
+          }
+        }
+      }
+    `;
+
+    // create shopify checkout.
+    // error returned will be a JSON string in error.message
+    const response = await storefrontClient().request(operation, { variables: this.shopifyCartInput(orderId) });
+    const data = response.data as { cartCreate: CartCreatePayload };
```
notbakaneko, line 99: casting the property of the response because I'm not about to go through the whole graphql type generation process for 2 queries right now.

nanaya: you mean like every single json api responses here? 👀

app/Models/Store/Order.php
```diff
@@ -184,6 +185,23 @@ public function scopeWithPayments($query)
         return $query->with('payments');
     }
 
+    #region Shopify convenience setters for webhook updates
```
nanaya, line 188: can't this be done in the `getShopifyParams` itself instead?

```php
return array_filter([
    'reference' => $params['admin_...'],
    '...' => '...',
], presence(...));

File: app/Models/Store/Order.php [hunk 3]
```diff
@@ -184,6 +185,23 @@ public function scopeWithPayments($query)
...
+    public function setAdminGraphqlApiIdAttribute(string $value)
+    {
+        $this->reference = $value;
+    }
+
+    public function setOrderNumberAttribute(string $value)
+    {
+        $this->transaction_id = static::PROVIDER_SHOPIFY."-{$value}";
```
nanaya, line 196: `.'-'.$value` isn't all that much different and more "uniform" in term of the string concat method 🙃

app/Console/Commands/StoreUpdateShopifyCheckoutId.php
```diff
@@ -0,0 +1,193 @@
...
+                    // nodes appear to be returned in order of values queried, including nulls set for not found
+                    if ($node !== null) {
+                        $orderId = static::getOrderIdFromNode($node);
+                        if ($orderId !== null) {
+                            $order = $ordersById[$orderId];
+                            $order->getConnection()->transaction(function () use ($node, $order) {
+                                $orderNode = $node['order'];
+
+                                $params = [
+                                    'transaction_id' => Order::PROVIDER_SHOPIFY.'-'.$orderNode['orderNumber'],
```
nanaya, line 108: this seems to not exist for incomplete (`processing`) orders...?

resources/views/store/orders/index.blade.php
```diff
@@ -46,24 +46,26 @@
...
-                            class="js-store-resume-checkout btn-osu-big btn-osu-big--rounded-thin"
-                            data-order-id="{{ $order->getKey() }}"
-                            data-provider="{{ $order->getPaymentProvider() }}"
-                            data-provider-reference="{{ $order->getProviderReference() }}"
-                            data-status="{{ $order->status }}"
-                        >
-                            {{ $order->isPaymentRequested() ? osu_trans('store.order.resume') : osu_trans('store.order.invoice') }}
-                        </button>
+                        @if ($order->shopify_url !== null)
+                            <a class="btn-osu-big btn-osu-big--rounded-thin" href="{{ $order->getShopifyUrl() }}">
```
nanaya, line 50: point to invoice page instead and have the action redirect to the keyed url for shopify orders?

nanaya (changes requested):

app/Console/Commands/StoreUpdateShopifyCheckoutId.php
```diff
@@ -0,0 +1,193 @@
...
+        foreach ($this->progress as $progress) {
+            $progress->setFormat('[%bar%] %current% %message%');
+            $progress->start();
+        }
+
+        Order::where('provider', 'shopify')->whereNotNull('reference')->chunkById(1000, function (Collection $chunk) {
+            $ordersById = $chunk->keyBy('order_id');
+            $this->progress['orders']->advance(count($ordersById));
+
+            $ids = $chunk->map(fn ($order) => $this->getCheckoutId($order->reference))->filter();
```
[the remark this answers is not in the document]
notbakaneko: those aren't valid gids and would be filtered out anyway by `getCheckoutId`

app/Models/Store/Order.php
```diff
@@ -266,6 +270,15 @@ public function getProviderReference(): ?string
         return static::splitTransactionId($this->transaction_id)[1] ?? null;
     }
 
+    /**
+     * Returns shopify_url without any of the querystring params
+     */
+    public function getShopifyUrl(): ?string
```
nanaya, line 276: not used?

app/Console/Commands/StoreMigrateShopifyCheckouts.php
```diff
@@ -0,0 +1,150 @@
...
+                    // nodes appear to be returned in order of values queried, including nulls set for not found
+                    if ($node !== null) {
+                        $orderId = static::getOrderIdFromNode($node);
+                        if ($orderId !== null) {
+                            $order = $ordersById[$orderId];
+                            if (isset($node['order'])) {
+                                $order->updateOrderWithGql($node['order']);
+                            } else {
+                                // orders that haven't completed checkout.
+                                $order->order->update(['shopify_url' => 'webUrl']);
```
nanaya, line 92: is actually intended to be a static `webUrl` string

resources/js/store.ts
```diff
@@ -62,15 +63,42 @@ export class Store {
...
+              }
+            }
+          }
+        }
+      }
+    `;
+
+    // create shopify checkout.
+    // error returned will be a JSON string in error.message
+    const response = await storefrontClient().request(operation, { variables: this.shopifyCartInput(orderId) });
```
nanaya, line 98: is it not weird the `cartInput` variable name is implicitly agreed in the function being called instead of it just returning the CartInput model? `variables: { cartInput: fun(...) } }`?

database/migrations/2025_04_06_000001_add_shopify_url_to_store_orders.php
```diff
@@ -0,0 +1,27 @@
+<?php
```
nanaya, line 1: what's that migration date

nanaya (changes requested):

nanaya (approved):

PR #11847, opened 2025-02-10
Remove legacy score user customization impact on user score bests

tsunyoku:
RFC - not totally sure if this is a direction that the team may agree with.

For someone who is exclusively playing osu!(lazer), it's a pretty confusing experience for other players who may not be mainly playing osu!(lazer) to not see your scores at all. For those who are interested in making the move from osu!stable to osu!(lazer), this may even be a deterrent as it can radiate the feeling that the scores aren't "important" if they're going to be less seen (I have seen this sentiment mentioned numerous times).

That's not to say that the "lazer mode" preference shouldn't exist, but it makes sense that user bests should display everything regardless of the setting (this is already the case for pinned(?) and recent scores) when statistics are also combined.


tsunyoku:
@peppy bumping this one. Still seeing a lot of people confused by this, and if this isn't the desired way forward then I'd still want to make progress on a different way of addressing this.

Some recent examples of confusion:

- https://osu.ppy.sh/community/forums/topics/2076574
- https://www.reddit.com/r/osugame/comments/1khedcr/what_happend_here_1k_pp_in_pinned_but_not_showing/

peppy (dismissed):

nanaya:
This will result in - when in non-lazer mode - lazer and classic scores being shown together without the `CL` mod indicator for the classic scores. Maybe just always show the mod for scores on profile page?

tsunyoku:
Yeah, I did notice that. This is already the case for pinned scores so I just left it be as I assumed a choice was already made on that front.

I can make it always display `CL` on the profile page if that's what we want.

nanaya (approved):

MrTerrorM:
Hello. I don't know where should I post this, so I will do it here.
From my perspective it would be better to get rid of the "CL" on scores, if the "lazer mode" is not ticked.

tsunyoku:
How do you determine the difference between legacy and non-legacy scores in that case? It's an important distinction.

MrTerrorM:
Well I don't really know. But if there isn't a CL on a score details, then why it should be on top plays?
I understand the push, but as a stable only player, it just creates clutter in the mod section.
Mayby there could be a different way of showing this? Like a purple stripe on the left of the grade?

{D3993635-345C-45FA-96E4-B46CBAAE5E47}

tsunyoku:
Score screen shows extended statistics for lazer, and also uses a different scoring system (unless you have lazer mode with classic scoring enabled), so there's plenty enough visual distinction I think.

I would not be against having CL display on the score screen at all, but it's out of scope of this change and also doesn't feel anywhere near as much of an issue as what's being solved here.

I really don't think a single mod is "clutter", but that's only my opinion.

MrTerrorM:
Well if that's the case, then I would remove the "Lazer mode" tickbox alltogether and only leave "Classic Scoring", because it is pretty useless right now since ticking it doesn't really change anything.

And well, about the "clutter". From a stable only viewpoint it is clutter, because it wasn't there in the first place. My purple stripe idea could be applied to the "lazer mode" (so it doesnt show the CL, but still shows the distinction between legacy and non-legacy scores).
\/ It could look like this (after hovering over the stripe it could say "Classic" like it is with the mod rn) \/
image

tsunyoku:
It changes a lot. Look at beatmap leaderboards for example. Until lazer mod multipliers go through several rounds of improvements, I'm not comfortable removing lazer mode. That is the long term goal though.

MrTerrorM:
Ok. So untill that time could there be mayby another checkbox, which would remove the cl mod from being shown?

tsunyoku:
I really don't think that's necessary.

richp2k:
isnt the obvious solution to have a lazer mod icon to differentiate lazer scores when the lazer mode tickbox is unchecked

PR #11848, opened 2025-02-11
Allow user mods in multiplayer freestyle

smoogipoo:
I considered two paths here - the first is to list every allowed mod in `allowed_mods`, and the second is to make freestyle act as a bypass. Listing every mod gets messy and wastes DB space (every playlist item is retained forever), so I've gone down the second path.

A future consideration may be that freemods is usually an all-or-nothing selection. We may want to change `freestyle` into a flags enum `{ ruleset | mods | beatmap }` and invert `allowed_mods` as `disallowed_mods`.

If you (@nanaya, @peppy) believe that's better to be done now, I can look down that path, but I'm hoping it's just as easily done in a future change.


peppy:
Probably fine for now

nanaya (approved):

PR #11855, opened 2025-02-12
Move team members section to its own page

nanaya:
With the current card design doesn't quite make sense for it to be on the left side of the info page.


resources/css/bem/team-members.less
```diff
@@ -2,21 +2,36 @@
...
-    color: hsl(var(--hsl-orange-4));
+  &__type {
+    border-radius: @border-radius--large;
+    display: grid;
+    gap: 5px;
+    padding: 5px;
+    background-color: hsl(var(--hsl-b2));
+    grid-template-columns: var(--grid-template-columns);
+    --grid-template-columns: 1fr;
+    --grid-template-columns-desktop: 1fr 1fr 1fr;
```
notbakaneko, line 16: `repeat(auto-fill, minmax(@user-card-width-min, 1fr))`
Needs a min width to stop clipping too many icons on card:
image

notbakaneko (changes requested):

resources/css/bem/team-members.less
notbakaneko: maybe get rid of `@user-card-width-min` and update the friends page as well since there's only 10px difference with `@user-card-width` and there's more icons than ever on the card now 🤔

notbakaneko (approved):

PR #11865, opened 2025-02-14
Limit contest entry name display length

### LINKED ISSUE #9846: no text wrapping for contests entries
image

nanaya:
This is a bit annoying with no fixed sizing/grid size because it means the whole available width isn't being used up optimally, leaving some extra white space which the name could fit but instead truncated.

Resolves #9846 maybe.


notbakaneko (approved):
It makes narrow entries able to contain less text but that's probably not a problem for us? 🤷

PR #11873, opened 2025-02-18
Show more stuff on team info section

nanaya:
Just so it's not too empty.

### DIFF
--- a/resources/views/teams/show.blade.php
+++ b/resources/views/teams/show.blade.php
@@ -124,14 +124,14 @@
                             </div>
                             @if (present($team->url))
                                 <div class="team-info-entry">
-                                    <div class="team-info-entry__title">{{ osu_trans('teams.edit.settings.url') }}</div>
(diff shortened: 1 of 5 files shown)


resources/views/teams/show.blade.php
```diff
@@ -124,14 +124,28 @@ class="team-action-button team-action-button--join js-login-required--click"
                             </div>
                             @if (present($team->url))
                                 <div class="team-info-entry">
-                                    <div class="team-info-entry__title">{{ osu_trans('teams.show.info.website') }}</div>
-                                    <div class="team-info-entry__value">
-                                        <span class="u-ellipsis-overflow">
-                                            <a href="{{ $team->url }}">{{ $team->url }}</a>
-                                        </span>
+                                    <div class="team-info-entry__title">{{ osu_trans('teams.edit.settings.url') }}</div>
```
notbakaneko, line 127: If we're sharing keys between edit/show/whatever, should they even be in that namespace?

notbakaneko (approved):

PR #11875, opened 2025-02-18
Check validity of mods wrt exclusivity on score submission

bdach:


app/Singletons/Mods.php
```diff
@@ -135,7 +135,22 @@ public function idsToBitset($ids): int
         );
     }
 
-    public function assertValidExclusivity(int $rulesetId, array $requiredIds, array $allowedIds): bool
+    public function assertValidExclusivity(int $rulesetId, array $modAcronyms): bool
```
bdach, line 138: Much of this is copied from what is now `assertValidMultiplayerExclusivity()`. I would have liked to have one call the other but the annoying part is how `assertValidMultiplayerExclusivity()` relies on collecting all of the `$disallowedIds` to then check `$allowedIds`. I'd have to make this return a `Set` rather than a `bool` and at that point it just kinda feels like a weird thing to do.

nanaya: I think you can simplify the other one with something like

```php
$this->assert...(...);
$disallowedIds = ...;
foreach ($rIds => $rId) {
    $disallowedIds->add($rId, ...$this->mods[...][...])
}
...;
```

It's a bit of duplicated work but probably not too bad...?

bdach: Fair, applied in https://github.com/ppy/osu-web/pull/11875/commits/7ce774bba0c64557fe1c28a03f25ce405b56ea88, I think.

I was a touch worried about `array_pop()` semantics, because it takes in the array as a reference, so I was sorta worried that calling the first assert was gonna empty the array fully rendering the following checks useless, but from what I can gather PHP has copy-on-write semantics on arrays when passing them as function parameters so it should be fine. (The tests also cover this off, which I tested by passing stuff by ref all the way down and seeing them break.)

app/Singletons/Mods.php
```diff
@@ -135,19 +135,31 @@ public function idsToBitset($ids): int
...
     {
+        while (($modAcronym = array_pop($modAcronyms)) !== null) {
+            $mod = $this->mods[$rulesetId][$modAcronym];
+            $incompatibleIds = $mod['IncompatibleMods'];
```
nanaya, line 142: I'm not actually sure why it's done this way originally instead of just

```suggestion
            $incompatibleIds = $this->mods[$rulesetId][$modAcronym]['IncompatibleMods'];
```

app/Singletons/Mods.php
```diff
@@ -135,19 +135,31 @@ public function idsToBitset($ids): int
...
+
+    public function assertValidMultiplayerExclusivity(int $rulesetId, array $requiredIds, array $allowedIds): bool
+    {
+        $this->assertValidExclusivity($rulesetId, $requiredIds);
+
         $disallowedIds = new Set();
 
         while (($requiredId = array_pop($requiredIds)) !== null) {
             $mod = $this->mods[$rulesetId][$requiredId];
             $incompatibleIds = $mod['IncompatibleMods'];
```
nanaya, line 161: as mentioned in other comment,
```suggestion
        foreach ($requiredIds as $requiredId) {
            $incompatibleIds = $this->mods[$rulesetId][$requiredId]['IncompatibleMods'];
```

app/Singletons/Mods.php
```diff
@@ -135,19 +135,31 @@ public function idsToBitset($ids): int
...
+            $mod = $this->mods[$rulesetId][$modAcronym];
+            $incompatibleIds = $mod['IncompatibleMods'];
+
+            $invalidIds = $incompatibleIds->intersect(new Set($modAcronyms));
+            if ($invalidIds->count() > 0) {
+                throw new InvariantException("incompatible mods: {$modAcronym}, {$invalidIds->join(', ')}");
+            }
+        }
+
+        return true;
```
nanaya, line 150: this and the other function probably shouldn't return anything. They're assertions.

nanaya (changes requested):

app/Models/Solo/Score.php
```diff
@@ -308,6 +308,11 @@ public function assertCompleted(): void
         if ($this->data->statistics->isEmpty()) {
             throw new InvariantException("field cannot be empty: 'statistics'");
         }
+
+        app('mods')->assertValidExclusivity(
+            $this->ruleset_id,
+            array_column($this->data->mods, 'acronym')
+        );
```
nanaya, line 315: oh right, so the mod ruleset is currently validated somewhere else (in arguably weird place) but I think mods validations should at least be grouped together (`validateSelection` in the Mods singleton).

bdach: Thing is, I specifically didn't want to put this specific exclusivity check there, because `validateSelection()` is called by `parseInputArray()`, and `parseInputArray()` is used in contexts where its only purpose is to determine whether the mods actually exist in the given ruleset at all and aren't duplicated, and *not* to check whether the set of mods is compatible with each other. Specifically, I'm talking about this usage:

https://github.com/ppy/osu-web/blob/551a11a1ecd80088d0aecc71668a048a7449f248/app/Models/Multiplayer/PlaylistItem.php#L73-L76

The list of "allowed mods" can contain mods that are incompatible with each other by design. That's because it's not a list of *mods that the player will play the game with*, but instead a list of *mods that the player can choose to play with* - therefore, the compatibility check there is not only not required, but actively wrong.

app/Singletons/Mods.php
bdach: I can't see a reason why both loops can't use `foreach` so I did both.

app/Singletons/Mods.php
bdach: Applied in https://github.com/ppy/osu-web/pull/11875/commits/47289670275a8639a48b4c3f4d41ade428e82a65; this affected tests too because they were checking the retval.

app/Singletons/Mods.php
nanaya: hmm, I wonder if this one can just gather all `$incompatibleIds` first and then compare with `$modAcronyms`. The `$modAcronym` info will be gone but it shouldn't happen in the first place if using client...?

I think the original idea was the modAcronyms array gets smaller as it goes

bdach: Applied in https://github.com/ppy/osu-web/pull/11875/commits/c81810e1701be5df2e507b2fcca7babd3a56372a, I think? I don't see any reason why it wouldn't work.

app/Singletons/Mods.php
```diff
@@ -135,28 +135,37 @@ public function idsToBitset($ids): int
...
+            $disallowedIds->add(...$incompatibleIds);
+        }
 
-            $invalidRequiredIds = $incompatibleIds->intersect(new Set($requiredIds));
-            if ($invalidRequiredIds->count() > 0) {
-                throw new InvariantException("incompatible mods: {$requiredId}, {$invalidRequiredIds->join(', ')}");
-            }
+        $invalidIds = $disallowedIds->intersect(new Set($modAcronyms));
+        if ($invalidIds->count() > 0) {
+            throw new InvariantException("incompatible mods: {$modAcronym}, {$invalidIds->join(', ')}");
```
nanaya, line 149: `$modAcronym` shouldn't be there

nanaya (approved):

app/Singletons/Mods.php
```diff
@@ -146,7 +146,7 @@ public function assertValidExclusivity(int $rulesetId, array $modAcronyms): void
 
         $invalidIds = $disallowedIds->intersect(new Set($modAcronyms));
         if ($invalidIds->count() > 0) {
-            throw new InvariantException("incompatible mods: {$modAcronym}, {$invalidIds->join(', ')}");
```
nanaya: it still works btw, just using the last value from the foreach above 👀

PR #11876, opened 2025-02-19
Allow admin to view all scores

nanaya:
Dunno if it is worth creating an auth for...


ELDOKSON:
if you want to be safe you have to

notbakaneko (approved):

PR #11879, opened 2025-02-19
Make deleted Beatmapset visible to moderators

notbakaneko:
Requested by NAT

mostly a copy paste from `show()`


resources/js/beatmap-discussions/header.tsx
```diff
@@ -64,6 +65,13 @@ export class Header extends React.Component<Props> {
   render() {
     return (
       <>
+        {this.beatmapset.deleted_at != null ? (
```
nanaya, line 68: `&&`?

app/Http/Controllers/BeatmapsetsController.php
```diff
@@ -129,7 +129,9 @@ public function search()
 
     public function discussion($id)
     {
-        $beatmapset = Beatmapset::findOrFail($id);
+        $beatmapset = priv_check('BeatmapsetShowDeleted')->can()
```
nanaya, line 132: I wonder if this should at least check for the whereHas thing just like the other (and maybe create a function specifically for it)

nanaya (changes requested):

app/Http/Controllers/BeatmapsetsController.php
notbakaneko: Discussions always `allBeatmaps` although I guess this should be fine if it's just an exists check 🤔

nanaya (approved):

peppy (approved):

PR #11880, opened 2025-02-19
Add team chat channel

nanaya:
Existing team will need to have the chat manually created. Something like `Team::where('channel_id', 0)->get()->each(function ($t) { $ch = $t->createChannel(); $t->save(); $t->members->each(fn ($m) => $ch->addUser($m->user)); })`

Also the team name is copied as channel name because it's shown in conversation list and stuff and doing additional query just for the team name would be a bit silly.

### DIFF
--- a/app/Http/Controllers/Teams/MembersController.php
+++ b/app/Http/Controllers/Teams/MembersController.php
@@ -9,7 +9,6 @@ namespace App\Http\Controllers\Teams;
 
 use App\Http\Controllers\Controller;
 use App\Models\Team;
-use App\Models\TeamMember;
 use Symfony\Component\HttpFoundation\Response;
 
 class MembersController extends Controller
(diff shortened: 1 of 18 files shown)


app/Models/Team.php
```diff
@@ -181,6 +225,33 @@ public function maxMembers(): int
...
+        $this->getConnection()->transaction(function () use ($member) {
+            $member->delete();
+            $user = $member->user;
+            if ($user !== null) {
+                $this->channel->removeUser($user);
+            }
+        });
+    }
+
+    public function resetChannelUsers(): void
```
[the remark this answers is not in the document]
nanaya: what indeed ಠ_ಠ

app/Models/Team.php
```diff
@@ -80,13 +87,26 @@ public function setUrlAttribute(?string $value): void
...
-            : bbcode((new BBCodeForDB($description))->generate());
+        (new TeamApplicationAccept($application, $this->leader))->dispatch();
+    }
+
+    public function createChannel(): Chat\Channel
+    {
+        $channel = Chat\Channel::create([
+            'type' => Chat\Channel::TYPES['team'],
+            'name' => $this->name,
+        ]);
```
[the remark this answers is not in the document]
nanaya: oh crap 😦

app/Http/Controllers/Teams/MembersController.php
```diff
@@ -9,7 +9,6 @@
 
 use App\Http\Controllers\Controller;
 use App\Models\Team;
-use App\Models\TeamMember;
```
notbakaneko, line 12: import still being used 👀

notbakaneko (changes requested):

app/Http/Controllers/Teams/MembersController.php
notbakaneko: also leaving team should also remove from channel?

nanaya: oh yeah forgot about that after merging 🤔

notbakaneko (approved):

PR #11881, opened 2025-02-19
Add general handler for image processor exception

nanaya:
Removing the same handler... which is already what the default handler does.

Also fixes error 500 when thrown somewhere else currently not having explicit handler like team settings.


notbakaneko (changes requested):
`AccountController::cover()` and `TopicCoversController::update()` as well?

peppy (approved):

PR #11920, opened 2025-02-21
Fix chat channel not displaying some new messages

notbakaneko:
Fixes the case where after chat and a channel was loaded and then chat was navigated away from, messages sent to that channel, would not be loaded when navigating back to that channel.

Basically, messages sent to a previously loaded channel, during the period chat is not the active page, would not be loaded. (This becomes more obvious with the upcoming Team chat link)

### DIFF
--- a/resources/js/chat/chat-state-store.ts
+++ b/resources/js/chat/chat-state-store.ts
@@ -100,6 +100,7 @@ export default class ChatStateStore implements DispatchListener {
       } else {
         this.pingService.stop();
         dispatch(new SocketMessageSendAction({ event: 'chat.end' }));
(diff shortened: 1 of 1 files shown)


nanaya:
what are the reproducer steps exactly ?_?

notbakaneko:
1. Open a chat channel.
2. Navigate anywhere not chat.
3. Other user sends a message to the channel.
4. Navigate back to the channel, message sent by other user is not there.
5. Other user sends another message, new message is there, but not the one from 3.

The easiest way is to 
1. open team chat
2. go back to team page
3. different user sends message to team chat
4. click team chat link

nanaya (approved):

PR #11931, opened 2025-02-23
Fix grid variables not being correctly applied in team description settings

### LINKED ISSUE #11928: Css bug in team settings
Image

If too much is written before the end of the team-settings-description-preview js-post-preview--preview block, then it goes beyond the scope of this block.

agatemosu:
Fixes #11928


nanaya (commented):
I think the actual problem is container's grid row for mobile being incorrectly applied on desktop

nanaya (approved):

Issue #11903, opened 2025-02-20
allow duplicate short team names


Italicc:
I understand the idea that these should be unique for gameplay purposes but as far as I know no competitive game really requires abbreviations to be unique anyway. I think it makes a lot more sense for those to be non-exclusive since there's only so many to go around that people would wanna use

Utiba:
Was about to make a post myself about this, but it seems like you already did. Definitely agree with this, I think it's a bit stupid to not allow duplicate short team names as I haven't seen any other team system in any game that has this in place. Teams are also tracked by their ID too, so I don't really see the point in not allowing duplicate short names if the ID of the team is already being tracked.

peppy:
We'll consider this one.

Meruru-chan:
Having only 4 characters AND unique after a while makes me literally unable to come up with anything I like to make a team now. We really need to allow duplicate tags honestly..

PR #11934, opened 2025-02-25
Support specifying short name for team url

nanaya:
It's kinda crap with `[...]` though since a lot of things don't auto url the closing bracket unless it's percent encoded. I added support for just opening bracket but it's even weirder.

A plus/minus/at/tilde sign or some other prefix would probably work better.


notbakaneko:
Well, considering `[` `]` are supposed to be reserved characters... 👀

nanaya:
yeah, `[]` is only because that's how it's shown on team page.

Also if we're going with supporting short name for url, #11903 wouldn't make sense anymore.

nanaya:
Drafting until we finish working out on the details.

- whether or not allowing duplicate
- what symbol to use (if going ahead)

RockRoller01:
> what symbol to use (if going ahead)

As mentioned on this Discord, this should probably just be `@` to stay consistent with username lookups. Having a different syntax for every aspect of the game would be annoying

nanaya:
use `@` sign and we're going ahead with unique short name

PR #11938, opened 2025-02-26
Display osu! username for changelog entry user link

### LINKED ISSUE #11935: Bug report page shows GitHub username instead of osu! name
I believe that Lazer uses the correct implementation.

nanaya:
Also link to osu! profile page instead of github. This kinda removes direct link to the user's github but it's still visible from the PR link so maybe it's fine...?

Resolves #11935.


notbakaneko (approved):
seems fine

PR #11949, opened 2025-02-28
Unify team and user profile header specification

nanaya:
1000x250 is the current actual display size (I don't even remember when 1200x320 was used) and 4MB matches the current team one.


resources/lang/en/users.php
```diff
@@ -225,7 +225,7 @@
                     'button' => 'Upload image',
                     'dropzone' => 'Drop here to upload',
                     'dropzone_info' => 'You can also drop your image here to upload',
-                    'size_info' => 'Cover size should be 2400x640',
+                    'size_info' => 'Cover size should be 2000x500',
```
nanaya, line 228: this really needs to be derived from the const instead of hardcoded 😦

notbakaneko (approved):

PR #11950, opened 2025-02-28
Fix error when creating team with blank name

nanaya:
<del>The actual validation happens before saving so this is probably the simplest fix (and let it error out during channel name validation).</del>

Includes refactoring of model saving process.


nanaya:
remove `required` attribute in name field, fill in short name, and submit

notbakaneko:
Doesn't it make more sense to flip the lines when creating the team so it errors first and returns the validation errors on `$team` instead of `$channel`?
```php
                    $team->saveOrExplode();
                    $channel = $team->createChannel();
```

app/Models/Team.php
```diff
@@ -280,10 +280,31 @@ public function save(array $options = [])
...
+                });
+            });
+        } else {
+            $saved = parent::save($options);
+        }
+
+        if ($saved) {
+            $this->flag()->updateFile();
+            $this->header()->updateFile();
+        }
```
notbakaneko, line 305: new filename is set here and doesn't get saved 👀

notbakaneko (changes requested):

app/Models/Team.php
```diff
@@ -281,30 +281,30 @@ public function save(array $options = [])
...
+
+                        $channel = $this->createChannel();
                         $channel->addUser($this->leader);
+
+                        $this->flag()->updateFile();
+                        $this->header()->updateFile();
                     }
 
-                    return $saved;
+                    return parent::save($options);
```
notbakaneko, line 299: Should probably only need to call `parent::save` an extra time only if some the files are updated; in the case of it returning `false` it's just calling twice in a row to return `false` but that can be fixed later

notbakaneko (approved):

app/Models/Team.php
nanaya: it's also to save the real channel id

PR #11951, opened 2025-02-28
Include custom beatmap tags set by own user in response

notbakaneko:
- [x] shouldn't query every beatmap separately


app/Http/Controllers/BeatmapsetsController.php
```diff
@@ -396,6 +398,7 @@ private function showJson($beatmapset)
             "{$beatmapRelation}.baseMaxCombo",
             "{$beatmapRelation}.failtimes",
             "{$beatmapRelation}.beatmapOwners.user",
+            "{$beatmapRelation}.ownBeatmapTags" => fn ($q) => $user !== null ? $q->where('user_id', $user->getKey()) : $q->none(),
```
nanaya, line 401: so this kinda works but I don't know if better 🤔 less fake `ownBeatmapTags` at least, and maybe call it `current_user_tag_ids` and always use `Auth::user()` in the scope.

Or maybe `JSON_OBJECTAGG($userId, JSON_ARRAYAGG(...))` and pass the current user all the way again.

```php
public function scopeWithUserTagIds($query, $userId)
{
    if ($userId === null) {
        $tagQuery = \DB::query()->selectRaw('null');
    } else {
        $tagQuery = BeatmapTag
            ::where('user_id', $userId)
            ->whereColumn('beatmap_id', $this->qualifyColumn('beatmap_id'));
        $tagQuery->selectRaw("JSON_ARRAYAGG({$tagQuery->qualifyColumn('tag_id')})");
    }

    return $query->addSelect(['user_tag_ids' => $tagQuery]);
}
```

app/Transformers/BeatmapCompactTransformer.php
```diff
@@ -12,12 +12,15 @@
 
 class BeatmapCompactTransformer extends TransformerAbstract
 {
+    public ?User $user = null;
```
nanaya, line 15: not used?

app/Http/Controllers/BeatmapsetsController.php
nanaya: can use at least a comment on this...

nanaya (changes requested):

app/Http/Controllers/BeatmapsetsController.php
notbakaneko: I haven't gotten around to checking it yet 😔

nanaya (approved):

PR #11954, opened 2025-02-28
Show session verification box for turbo requests

nanaya:
Also changed how turbo redirect works.


notbakaneko (changes requested):
if going to somewhere directly like `community/chat#join` with an account or env required to verify on login, it just spins instead of showing the verification

nanaya:
> if going to somewhere directly like `community/chat#join` with an account or env required to verify on login, it just spins instead of showing the verification

what's exactly the steps and env?

notbakaneko:
I just went directly to `/community/chat#join` and logged in as an admin 🤔

nanaya:
is it still a problem? I can't seem to reproduce it.

notbakaneko (approved):
Seems to be working now

PR #11959, opened 2025-03-03
Fix team channel operation not being rolled back on error

nanaya:
Isn't dealing with multiple databases fun.


notbakaneko (approved):
Oh right, this is a thing... 🤔

PR #11961, opened 2025-03-03
Delete team channel if there's no message

nanaya:
Seems weird keeping it around in such cases.


notbakaneko (changes requested):
Probably can do with a test as well

notbakaneko (approved):

PR #11966, opened 2025-03-04
Add interop endpoint for channel creation and member management

nanaya:
Related to ppy/osu#31908

- [ ] validation/type restriction?
- [ ] what response is needed for this
- [ ] is deleting channel also needed?

### DIFF
--- a/app/Http/Controllers/InterOp/Chat/ChannelsController.php
+++ b/app/Http/Controllers/InterOp/Chat/ChannelsController.php
@@ -1,0 +1,45 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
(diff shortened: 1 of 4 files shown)


nanaya:
do we want to delete messages and everything when deleting channel?

smoogipoo:
For spectator channels yeah. For multiplayer channels we have a discussion recently (can't remember where) about multiplayer events, and peppy wanted to see chat displayed as well, so maybe not.

That said, maybe we don't need to delete channels at all and just get clients to part them?

nanaya:
> Correct me if I'm wrong but are players already able to join arbitrary multiplayer channels right now, assuming they have the channel id? If so, that may be something we don't want, and for spectator too. If so, then maybe we want validation that clients can't manually join these types of channels and for the realtime server to be authoritative on that (i.e. skipping that validation).

no, only public and tournament channels can be joined through general chat endpoint

smoogipoo:
For any reviewers: probably best to let us get the spectator channel in the game before merging this.

PR #11968, opened 2025-03-04
Fix discussion page error with no visible beatmap

nanaya:
If there's discussion on the general section but none on any specific beatmaps, the store's `beatmaps` property ended up empty and crashes on discussions-state:402 (finding current beatmap id).

- deleted beatmapset
- deleted beatmaps (all of them)
- no discussion on any specific beatmaps
- there's discussion on general all section (not sure if actually needed to trigger the error)


notbakaneko (approved):
Probably a good enough workaround for the limited scope of users and cases that will actually see it, without making everything need to support a nullable selected beatmap. (I remember thinking surely this isn't likely to happen and left it 👀 )

PR #11970, opened 2025-03-04
Add season stats user profile display

venix12:
Part of #8736.

Decided to put it in the header after some playing around because it felt most fitting there, any further design improvement suggestions are more than welcome.

Also adds ``colour_tier`` column to the divisions table, for the accent line display.

## Preview

image
image


resources/js/profile-page/season-stats.tsx
```diff
@@ -0,0 +1,125 @@
...
+        my: 'bottom center',
+        viewport: $(window),
+      },
+      show: {
+        delay: 200,
+        event: event.type,
+        ready: true,
+      },
+      style: {
+        classes: 'qtip qtip--daily-challenge',
```
venix12, line 115: could probably unify the name somehow, though didn't really have any naming idea

resources/css/bem/season-stats-popup.less
```diff
@@ -0,0 +1,76 @@
...
+    padding-left: calc(@_gap * 2);
+    border-left: 2px solid hsl(var(--hsl-c1));
+  }
+
+  &__img {
+    height: 35px;
+    border-radius: @border-radius-small;
+  }
+
+  &__main {
```
nanaya, line 42: I find the alignment weird. It's centered but the separator in the middle make it look like the separator should be exactly on the center except it's probably not.

Maybe have the ranking on the left as small as possible with `auto auto 1fr` (with actual line div so a simple 15px gap can be used to match the padding), and centered image section.

Or maybe even move the rank to be another row (but with larger value size) along with total score

resources/css/bem/season-stats-popup.less
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+
+  &__division {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
+    padding-left: calc(@_gap * 2);
+    border-left: 2px solid hsl(var(--hsl-c1));
```
nanaya, line 34: maybe match the other separator color on the page

nanaya (changes requested):

resources/css/bem/season-stats.less
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+.season-stats {
+  @_img-height: 35px;
```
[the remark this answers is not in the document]
venix12: generally, currently used assets use the badge designs (so 40px height), but plans are to replace it with some custom thing in the future, guess setting it for 40px for now could make more sense, not very sure where i took the 35 px from... :thinking:

resources/css/bem/season-stats-popup.less
venix12: not sure what's the advantage of using ``auto auto 1fr`` than just using flex, unless i am missing what you mean exactly?

nanaya: with flex you need to specify `flex: none` to each relevant nodes

venix12: looking at it now, yeah i didn't really get what you mean exactly in the first place...

wondering if ``1fr auto 1fr`` wouldn't maybe be better so that the line just always stays in the center?

nanaya: I think the rank number area on the left looks way too empty compared to its right if the line is on exact center

venix12: went with ``auto auto 1fr`` for now then

nanaya (changes requested):
The location may overlap with user info on the left



and the popup may go off screen

resources/js/profile-page/season-stats.tsx
```diff
@@ -0,0 +1,128 @@
...
+          <Img2x
+            className='season-stats-popup__img'
+            src={stats.division.image_url}
+          />
+          <div className='season-stats-popup__name'>
+            {stats.division.name}
+          </div>
+          <div className='season-stats-popup__threshold'>
+            {trans('users.show.season_stats.division_top_percentage', {
+              value: formatNumber(Number(stats.division.threshold), 0, { style: 'percent' }),
```
nanaya, line 44: what's this `Number()` casting for?

nanaya (dismissed):

nanaya (approved):

venix12:
think it got overlooked :thinking:

PR #11973, opened 2025-03-05
Filter out statistics of deleted teams

nanaya:
Forgot about this case of deleted team on ranking page...

<del>Either this or just delete the statistics when the team is deleted 🤔 That one sounds better</del> how about both


tests/Models/TeamTest.php
```diff
@@ -30,6 +31,7 @@ public function testDelete(): void
         $team->members()->create(['user_id' => User::factory()->create()->getKey()]);
         Chat\Message::factory()->create(['channel_id' => $team->channel, 'user_id' => $team->leader_id]);
         $team->applications()->create(['user_id' => User::factory()->create()->getKey()]);
+        $team->statistics()->create(['ruleset_id' => 0]);
```
notbakaneko, line 34: other team in the test should also have `statistics` (and probably add `applications` as well)

notbakaneko (changes requested):

nanaya:
I'll add that tomorrow...

(also this kinda breaks the ranking page btw)

notbakaneko (approved):

PR #11975, opened 2025-03-06
Do additional sorting of user best scores

nanaya:
┐( ･_･)┌

(see inline comment)


app/Models/Traits/UserScoreable.php
```diff
@@ -68,6 +68,8 @@ public function beatmapBestScores(string $mode, int $limit, int $offset, array $
         }
 
         $results->load($with);
+        // make outdated index less obvious
+        $results = $results->sortBy('pp');
```
notbakaneko, line 72: Did you mean
```suggestion
        $results = $results->sortByDesc('pp');
```
👀

notbakaneko (approved):

PR #11979, opened 2025-03-07
Show statistics on team page

nanaya:
Maybe? 🤔 

Kinda took the original design but the stats are different and there isn't enough stuff to fill the section so I moved the existing info there and made team custom page its own section just like profile page.

Also matched section title and stuff from profile page.

<sub>the page is kinda long and sparse now</sub>


app/Models/TeamStatistics.php
```diff
@@ -27,6 +34,14 @@ public function members(): HasMany
         return $this->hasMany(TeamMember::class, 'team_id', 'team_id');
     }
 
+    public function getRank(): int
```
notbakaneko, line 37: should teams even have a rank if they have no plays or score on a ruleset?

notbakaneko (changes requested):

notbakaneko (approved):

PR #11983, opened 2025-03-08
Always require ruleset and beatmap id parameters for multiplayer plays

nanaya:
Reference: https://github.com/ppy/osu/pull/31882#discussion_r1983402441

- [x] confirm client always sends those two parameters


bdach:
> confirm client always sends those two parameters

It does; see https://github.com/ppy/osu/blob/775cdc087eda5c1525d763c6fa3d422db0e93f66/osu.Game/Online/Rooms/CreateRoomScoreRequest.cs#L33-L36.

In full-stack testing with https://github.com/ppy/osu/pull/31882 pre-fix I can confirm that this catches the fail case here, so seems good to me.

bdach (approved):

notbakaneko (approved):

notbakaneko:
Does this deploy need to be synced with anything? @bdach

bdach:
Well, the client has been sending these parameters since freestyle multiplayer went online, so I don't see any reason to sync the deploy with anything...?

PR #11988, opened 2025-03-10
Allow failed scores to display on results pages in realtime multiplayer

bdach:
Would close https://github.com/ppy/osu/issues/29263.

Opening as draft as I have a few doubts in my mind regarding this as written:

- I don't know very much about laravel ORM but the way this reaches 2 relations across into the room seems... expensive... but also is the simplest way to cover off all usages (one of them is `UserScoreAggregate::recalculate()` which is maybe not that relevant, but also needs to work correctly?)

- When we're talking about realtime "failed scores", there are actually two types of those: one being "user played the beatmap to end, but actually failed HP-wise midway through", and one being "the user just quit out". This way catches both. I'm not sure what to think of that; it's probably fine, at least for the intended usage of tournament play (as we'll want the aborted play to be logged there), but is it fine for casual play? (cc @peppy @smoogipoo)


peppy:
> When we're talking about realtime "failed scores", there are actually two types of those: one being "user played the beatmap to end, but actually failed HP-wise midway through"

Probably fine to rely on replay to discern between these for now. If we need it, we could store further metadata in the json in the future.

nanaya:
> * I don't know very much about laravel ORM but the way this reaches 2 relations across into the room seems... expensive... but also is the simplest way to cover off all usages (one of them is `UserScoreAggregate::recalculate()` which is maybe not that relevant, but also needs to work correctly?)

There's way to not incur the one extra room query (playlistItem should already be loaded).

Also, with this change, the aggregate score would include failed plays...

bdach:
> There's way to not incur the one extra room query (playlistItem should already be loaded).

What would that be? Do I add some include somewhere?

> Also, with this change, the aggregate score would include failed plays...

In realtime rooms only, which should only ever have one score per playlist item, so I'm not sure it's a problem...?

nanaya:
> What would that be? Do I add some include somewhere?

I think it's use setRelation on score token somewhere in the pipeline (and change the score creation process to use existing relation instead of requerying in app/Models/Multiplayer/ScoreLink.php:34.

> In realtime rooms only, which should only ever have one score per user per playlist item, so I'm not sure it's a problem...?

There's also the room level aggregate which sums over all the playlist item plays.

bdach:
> There's also the room level aggregate which sums over all the playlist item plays.

I see. I'm not sure that's something anyone will care about really though. Like in client "failing" lets you complete the play anyway, so it makes sense to me that fails should show on results and that they should count into total aggregates.

It's another part of the bigger problem that playlists and realtime multiplayer share a lot of code but conceptually they diverge in different directions because they're basically different game modes at their core.

bdach:
> I think it's use setRelation on score token somewhere in the pipeline (and change the score creation process to use existing relation instead of requerying in app/Models/Multiplayer/ScoreLink.php:34.

Would this be along the right lines?

```diff
diff --git a/app/Models/Multiplayer/Room.php b/app/Models/Multiplayer/Room.php
index 95347cd19b..10a78f12d3 100644
--- a/app/Models/Multiplayer/Room.php
+++ b/app/Models/Multiplayer/Room.php
@@ -456,6 +456,10 @@ class Room extends Model
         $this->assertValidCompletePlay();
 
         return $this->getConnection()->transaction(function () use ($params, $scoreToken) {
+            $playlistItem = $scoreToken->playlistItem()->firstOrFail();
+            $playlistItem->setRelation('room', $this);
+            $scoreToken->setRelation('playlistItem', $playlistItem);
+
             $scoreLink = ScoreLink::complete($scoreToken, $params);
             $user = $scoreLink->user;
             $agg = UserScoreAggregate::new($user, $this);
diff --git a/app/Models/Multiplayer/ScoreLink.php b/app/Models/Multiplayer/ScoreLink.php
index 42dbae1ce7..eca35a7dbc 100644
--- a/app/Models/Multiplayer/ScoreLink.php
+++ b/app/Models/Multiplayer/ScoreLink.php
@@ -31,7 +31,7 @@ class ScoreLink extends Model
             // multiplayer scores are always preserved.
             $score = Score::createFromJsonOrExplode([...$params, 'preserve' => true]);
 
-            $playlistItem = $token->playlistItem()->firstOrFail();
+            $playlistItem = $token->playlistItem;
             $requiredMods = array_column($playlistItem->required_mods, 'acronym');
             $mods = array_column($score->data->mods, 'acronym');
             $mods = app('mods')->excludeModsAlwaysValidForSubmission($score->ruleset_id, $mods);

```

With this applied, watching mysql general logs, there does appear to be one less query for the room, but I'm going to be real, I have very little idea of what I'm doing here (the only docs I can see for `setRelation()` is some third party blogpost, and phpstorm is yelling at me that `$token->playlistItem` is "property access via magic method" and I have no idea if that's bad or not).

nanaya:
> \<codes\>

seems about right (although playlistItem on the first one isn't strictly needed, can just set room relation directly on `$token->playlistItem`)

> With this applied, watching mysql general logs, there does appear to be one less query for the room, but I'm going to be real, I have very little idea of what I'm doing here (the only docs I can see for `setRelation()` is some third party blogpost, 

it's a bit underdocumented but it's in the laravel api docs

> and phpstorm is yelling at me that `$token->playlistItem` is "property access via magic method" and I have no idea if that's bad or not).

the attribute is missing in the model's phpdocs.

bdach:
> seems about right (although playlistItem on the first one isn't strictly needed, can just set room relation directly on `$token->playlistItem`)

I'm sort of confused by this. My crude inference as to what `setRelation()` does is that it'll put the model instance in like a dictionary internal to the entity. The reason why I did that thing with setting the `playlistItem` relation on `$scoreToken` is because I don't want to just *remove* the `$playlistItem` refetch from its original location (because it's there [for a reason apparently](https://github.com/ppy/osu-web/pull/11369), and I'm not sure what it is), and secondly, if I don't set the relation on the token, isn't it just going to bypass the set relation on the playlist item? As in, it'll refetch the playlist item because it's not preloaded/relation-set, and then the room preload is worse than nothing.

In light of that I applied the patch verbatim as I had it, but if I'm misunderstanding stuff (which I may well be), I'll change further.

> the attribute is missing in the model's phpdocs.

Added in https://github.com/ppy/osu-web/pull/11988/commits/68ec27803e0fa4c1cdc7e6320e1c3ed1786d81e5 I think.

> more like by adding preload on the query

Done in https://github.com/ppy/osu-web/pull/11988/commits/5fea5fce0e60dd9870245992325a94705e0c2cae I think, although I really only tested that this doesn't kill the query outright. Not sure how to tell if this does anything beneficial. mysql general log looks about the same to me.

nanaya:
> > seems about right (although playlistItem on the first one isn't strictly needed, can just set room relation directly on `$token->playlistItem`)
> 
> I'm sort of confused by this. My crude inference as to what `setRelation()` does is that it'll put the model instance in like a dictionary internal to the entity. The reason why I did that thing with setting the `playlistItem` relation on `$scoreToken` is because I don't want to just _remove_ the `$playlistItem` refetch from its original location (because it's there [for a reason apparently](https://github.com/ppy/osu-web/pull/11369), and I'm not sure what it is), and secondly, if I don't set the relation on the token, isn't it just going to bypass the set relation on the playlist item? As in, it'll refetch the playlist item because it's not preloaded/relation-set, and then the room preload is worse than nothing.

Right, I forgot about the playlist item somehow gone missing in the process 😦

Accessing `$token->playlistItem` will set the relation as long it's accessed that way but there's no simple way to handle missing playlist through that at the moment (unlike `firstOrFail`) so, eh

also, maybe add some tests...

bdach:
> also, maybe add some tests...

Would https://github.com/ppy/osu-web/pull/11988/commits/4b0bd1d15912b8e135c883cc5684f29bb0d3d850 work for you?

tests/Models/Multiplayer/UserScoreAggregateTest.php
```diff
@@ -279,18 +287,63 @@ public function testRecalculate(): void
...
+        $agg = UserScoreAggregate::new($user, $room);
+
+        $this->assertSame(1, $agg->completed);
+        $this->assertSame(1000, $agg->total_score);
+
+        $userHighScore = PlaylistItemUserHighScore::where([
+            'playlist_item_id' => $playlistItem->getKey(),
+            'user_id' => $user->getKey(),
+        ])->first();
+        $this->assertNotNull($userHighScore->score_id);
```
nanaya, line 313: maybe compare with score id returned by `roomAddPlay` instead?

nanaya (changes requested):

tests/Models/Multiplayer/UserScoreAggregateTest.php
```diff
@@ -279,18 +295,55 @@ public function testRecalculate(): void
...
+        $agg = UserScoreAggregate::new($user, $room);
+
+        $this->assertSame(1, $agg->completed);
+        $this->assertSame(1000, $agg->total_score);
+
+        $userHighScore = PlaylistItemUserHighScore::where([
+            'playlist_item_id' => $playlistItem->getKey(),
+            'user_id' => $user->getKey(),
+        ])->first();
+        $this->assertEquals($scoreLink->score_id, $userHighScore->score_id);
```
nanaya, line 321:
```suggestion
        $this->assertSame($scoreLink->score_id, $userHighScore->score_id);
```

```php
// this passes
$this->assertEquals(0, null);
```

PR #11989, opened 2025-03-11
Add per-beatmap team ranking

nanaya:
🤷

### DIFF
--- a/app/Http/Controllers/BeatmapsController.php
+++ b/app/Http/Controllers/BeatmapsController.php
@@ -36,7 +36,7 @@ class BeatmapsController extends Controller
     private static function assertSupporterOnlyOptions(?User $currentUser, string $type, array $mods): void
     {
         $isSupporter = $currentUser !== null && $currentUser->isSupporter();
-        if (!in_array($type, ScoreSearchParams::FREE_TYPES, true) && !$isSupporter) {
+        if (in_array($type, ScoreSearchParams::SUPPORTER_TYPES, true) && !$isSupporter) {
             throw new InvariantException(osu_trans('errors.supporter_only'));
         }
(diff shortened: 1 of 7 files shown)


app/Http/Controllers/BeatmapsController.php
```diff
@@ -36,7 +36,7 @@ public function __construct()
     private static function assertSupporterOnlyOptions(?User $currentUser, string $type, array $mods): void
     {
         $isSupporter = $currentUser !== null && $currentUser->isSupporter();
-        if ($type !== 'global' && !$isSupporter) {
+        if (!in_array($type, ScoreSearchParams::FREE_TYPES, true) && !$isSupporter) {
```
notbakaneko, line 39: wonder if it makes more sense to make it `SUPPORTER_TYPES` now if the condition is no longer simply `!== 'global'`

resources/js/beatmapsets-show/scoreboard/controller.ts
```diff
@@ -65,13 +65,17 @@ export default class Controller {
...
-    if (!core.currentUser?.is_supporter && (this.currentType !== 'global' || this.enabledMods.size > 0)) {
+    if (!core.currentUser?.is_supporter && this.requiresSupporter) {
       return 'supporter_only';
     }
 
     return this.xhrState;
   }
 
+  get requiresSupporter() {
+    return !['global', 'team'].includes(this.currentType) || this.enabledMods.size > 0;
```
notbakaneko, line 76: can be a `const Set<ScoreboardType>` (also maybe make it match against supporter required instead of not free)

notbakaneko (changes requested):

notbakaneko (approved):

peppy:
I'm adding client-side support for this.

Issue #3471, opened 2018-07-09
Add playcount sorting to score ranking


akari-ds:
Currently the new site's score ranking is sorted by ranked score with no option to sort by playcount which is available on the old site.

PR #11998, opened 2025-03-13
Add sort by score option for team ranking page

nanaya:
Just a tiny section above the table.


Joehuu:
Was hoping for carets like in old site, so that element doesn't need to exist and relist column labels if say sorting for more columns was added.

firefox_dNLAgp6lrn

Also, there's a long-standing issue for play count sorting in score ranking: https://github.com/ppy/osu-web/issues/3471
Another PR for sure, but just mentioning as ordering was implemented here.

notbakaneko (approved):

PR #11999, opened 2025-03-13
Update setting beatmap tags to use path param

notbakaneko:
`PUT /api/v2/beatmaps/{beatmap_id}/tags/{tag_id}` instead of body param

and expire the cache on update

### DIFF
--- a/app/Models/Beatmap.php
+++ b/app/Models/Beatmap.php
@@ -251,6 +251,11 @@ class Beatmap extends Model implements AfterCommit
         return $this->playmode === static::MODES['osu'] || $this->playmode === $rulesetId;
     }
 
+    public function expireTopTagIds()
+    {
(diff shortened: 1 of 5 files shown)


app/Models/Beatmap.php
```diff
@@ -251,6 +251,11 @@ public function canBeConvertedTo(int $rulesetId)
         return $this->playmode === static::MODES['osu'] || $this->playmode === $rulesetId;
     }
 
+    public function expireTopTagIds()
+    {
+        cache_expire_with_fallback("beatmap_top_tag_ids:{$this->getKey()}", $GLOBALS['cfg']['osu']['tags']['beatmap_tags_cache_duration']);
```
nanaya, line 256: is there reason why the tags are cached with fallback (which means it may potentially return empty array)? while the query isn't simple, I doubt it's slow enough to warrant the downside.

nanaya (changes requested):

nanaya (approved):
eh, can be done later I suppose

PR #12005, opened 2025-03-14
Adjust accuracy rounding for display

### LINKED ISSUE #12004: Accuracy displays need to floor to 0.01% rather than round to nearest 0.01%
[To match the client.](https://github.com/ppy/osu/blob/01828931a211fd4de31cacedbd73a247c1e22bd2/osu.Game/Utils/FormatUtils.cs#L20-L24)

Related to https://github.com/ppy/osu/issues/32371; https://osu.ppy.sh/scores/4510215027 is displayed as 97.27% in client, but 97.28% on the website.

nanaya:
Resolves #12004.

### DIFF
--- a/resources/js/utils/score-helper.ts
+++ b/resources/js/utils/score-helper.ts
@@ -12,9 +12,13 @@ import { trans } from './lang';
 import { legacyAccuracyAndRank } from './legacy-score-helper';
 
 export function accuracy(score: SoloScoreJson) {
-  return shouldReturnLegacyValue(score)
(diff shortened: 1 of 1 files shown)


bdach:
Source looks pretty correct to me, I mainly logged this as issue rather than try and fix myself because (a) in the middle of other work and (b) I wasn't sure how many places were gonna require fixing, so dunno if my approval here means anything.

notbakaneko (approved):

PR #12012, opened 2025-03-17
Add country filter for score ranking

nanaya:
The final goal is to allow same filters between performance and score sorting option just like on team ranking page (and reducing number of menu items). The index is similar to the one used for performance so performance characteristics should be the same.

The columns are also added for variant statistics but the process for updating it isn't in place (yet?) so it's still one step short from being able to have sort buttons.

Migration entry is `2025_03_18_065831_add_country_ranked_score_index_to_stats_tables`.

### DIFF
--- a/app/Http/Controllers/RankingController.php
+++ b/app/Http/Controllers/RankingController.php
@@ -94,7 +94,7 @@ class RankingController extends Controller
                 abort(404);
             }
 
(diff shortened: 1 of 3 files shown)


notbakaneko (approved):

nanaya:
pinging @peppy to confirm db change and migration entry

peppy (approved):

PR #12014, opened 2025-03-17
Add standardised scoring for judged contests

### LINKED ISSUE #11390: Add an option for standardized scoring in contest judging system
Currently most of the beatmapping contests use standardized scoring for a more proper score weighting across the judges, with the formula that goes as following and applies to final scores of the contest judging:

```judge X's final score = (judge X's raw score - judge X's average raw score) / judge X's standard deviation```

Should likely be added as an option to be enabled in ``extra_options`` on ``Contest``.

### PR DESCRIPTION
**Pull request by notbakaneko**
closes #11390

the option is `"is_score_standardised": true`.
I'm not sure how/when the scoring should be calculated (interop/admin page, etc), so I added a a command `contest:calculate-scores` for now.

- [x] store judge's standard deviation so it doesn't need to get recalculated every time
- [x] order by standardised score
- [x] better display in results?
- [x] add something to trigger calculation

### CODE REVIEW

File: resources/lang/en/contest.php [hunk 1]
```diff
@@ -30,6 +30,7 @@
         '_' => 'Judging results',
         'creator' => 'creator',
         'score' => 'Score',
+        'score_std' => 'Standardised Score',
         'total_score' => 'total score',
```
Walavouchey, line 34: casing?

app/helpers.php
```diff
@@ -965,6 +965,17 @@ function ujs_redirect($url, $status = 200)
     }
 }
 
+function std_dev(array $values): array
+{
+    $size = count($values);
+    $mean = array_sum($values) / $size;
+
+    return [
+        sqrt(array_sum(array_map(fn ($value) => pow($value - $mean, 2), $values)) / $size),
```
Walavouchey, line 974: looks correct

app/Models/ContestJudge.php
```diff
@@ -27,4 +28,17 @@ public function user(): BelongsTo
     {
         return $this->belongsTo(User::class, 'user_id');
     }
+
+    public function stdDev()
+    {
+        // TODO: treat missing scores as 0?
```
Walavouchey, line 34: all judges are supposed to judge all entries and they're poked if they haven't
default of 0 could make sense if it's to indicate that something is wrong, though ideally we'd know by counting the number of votes

resources/lang/en/contest.php
notbakaneko: this one of part of `Score` not `total score`

app/Models/ContestJudge.php
```diff
@@ -27,4 +30,24 @@ public function user(): BelongsTo
...
+    }
+
+    public function stdDev(): array
+    {
+        // TODO: treat missing scores as 0?
+        $entryScores = ContestJudgeScore::scoresByEntry()
+            ->whereHas(
+                'vote',
+                fn (Builder $q) => $q->where('user_id', $this->user_id)
+                    ->whereHas('entry', fn (Builder $qq) => $qq->where('contest_id', $this->contest_id))
```
nanaya, line 48:
```suggestion
            // maybe adjust the scope to accept ids instead
            ->scoresFrom($this->user, Contest::find($this->contest_id)
```

resources/js/contest-voting/entry.coffee
```diff
@@ -77,6 +70,23 @@ export class Entry extends React.Component
             target: '_blank'
             i className: 'fas fa-fw fa-lg fa-external-link-alt'
 
+
+  renderScore: ->
+    score = if @props.entry.results.score_std != null
```
nanaya, line 75:
```suggestion
    score = if @props.entry.results.score_std?
```

resources/js/contest-voting/entry.coffee
```diff
@@ -77,6 +70,23 @@ export class Entry extends React.Component
...
+      @props.entry.results.score_std.toFixed(2)
```
nanaya, line 76: transChoice requires the number to be a number, not a string. It'll also formatNumber it for display already.

(or alternatively the string must be passed as `count_delimited` replacement... which I noticed there's no way to manually override it for the js version)

resources/js/contest-judge-results/vote.tsx
```diff
@@ -27,7 +27,15 @@ export default function Vote(props: Props) {
         </div>
       )}
 
-      <div className='contest-judge-results-vote__total-score'>
+      <div className='contest-judge-results-vote__scores-header'>
+        {props.vote.total_score_std != null && (
+          <ValueDisplay
+            label={trans('contest.judge_results.score_std')}
+            modifiers={'score'}
+            value={props.vote.total_score_std.toFixed(2)}
```
nanaya, line 35: formatNumber

resources/js/contest-judge-results/vote.tsx
```diff
@@ -27,7 +27,15 @@ export default function Vote(props: Props) {
         </div>
       )}
 
-      <div className='contest-judge-results-vote__total-score'>
+      <div className='contest-judge-results-vote__scores-header'>
+        {props.vote.total_score_std != null && (
+          <ValueDisplay
+            label={trans('contest.judge_results.score_std')}
+            modifiers={'score'}
```
nanaya, line 34: extra `{}`

resources/js/contest-judge-results/header.tsx
```diff
@@ -29,6 +29,7 @@ export default class Header extends React.PureComponent<Props> {
 
   render() {
     const totalScore = `${this.props.entry.results.votes}/${this.props.contest.max_total_score}`;
+    const totalScoreStd = this.props.entry.judge_votes.reduce((total, vote) => (vote.total_score_std ?? NaN) + total, 0);
```
nanaya, line 32: shouldn't the default when missing be 0 instead?

...or maybe just actually include and use the `this.props.entry.results.score_std` value

resources/js/contest-judge-results/header.tsx
```diff
@@ -39,15 +40,23 @@ export default class Header extends React.PureComponent<Props> {
         />
 
         <div className='contest-judge-results-header__values'>
+          {!Number.isNaN(totalScoreStd) && (
+            <ValueDisplay
+              label={trans('contest.judge_results.total_score_std')}
+              modifiers='judge-results'
+              value={totalScoreStd.toFixed(2)}
```
nanaya, line 47: formatNumber

resources/css/bem/contest-judge-results-scores.less
```diff
@@ -4,14 +4,15 @@
 .contest-judge-results-scores {
   display: grid;
   align-items: center;
-  grid-gap: 5px;
+  gap: 5px;
   // category-name value
   grid-template-columns: minmax(auto, 120px) auto;
```
nanaya, line 9: unrelated but this probably should be `auto 1fr` otherwise this happens:

resources/js/contest-voting/entry.coffee
nanaya, line 1: `relativeVotePercentage` should be adjusted to be something like `100 * (entry.results.score_std - min_score_std) / (max_score_std - min_score_std)`

resources/js/contest-judge-results/vote.tsx
```diff
@@ -39,17 +47,17 @@ export default function Vote(props: Props) {
         {props.vote.scores.map((score) => {
           const category = props.contest.scoring_categories.find((x) => x.id === score.contest_scoring_category_id);
 
-          return (
+          return category != null && (
             <div key={score.id} className='contest-judge-results-scores__row'>
               <div className='contest-judge-results-scores__col'>
-                <div className='contest-judge-entry__description-icon' title={category?.description}>
+                <div className='contest-judge-entry__description-icon' title={category.description}>
```
nanaya, line 53: unrelated but the class doesn't really do anything useful apart of being a random element of an entirely different block

nanaya (changes requested):

resources/js/contest-voting/entry.coffee
notbakaneko: occurred to me that `1.0` is `one point zero points`, not the singular, but I don't know how that rule applies in other languages...

app/Http/Controllers/ContestEntriesController.php
```diff
@@ -19,7 +20,12 @@ class ContestEntriesController extends Controller
 {
     public function judgeResults($id)
     {
-        $entry = ContestEntry::with('contest')->findOrFail($id);
+        $entry = ContestEntry
+            ::with('contest')
+            // TODO: replace with casted extra_options.
+            // Assume hitting this endpoint is always for judged contests.
+            ->withScore(new Contest(['extra_options' => ['judged' => true, 'is_score_standardised' => true]]))
```
nanaya: just fetch the contest first? `Contest::whereHas('entries', fn ($q) => $q->whereKey($id))`

although looking at the url maybe it should contain the contest id in the first place... `/contests/{contest}/entries/{entry}/results`

app/Http/Controllers/ContestEntriesController.php
```diff
@@ -45,15 +51,16 @@ public function judgeResults($id)
...
             'judge_votes.scores',
             'judge_votes.total_score',
+            'judge_votes.total_score_std',
             'judge_votes.user',
             'results',
             'user',
         ]);
 
-        $entriesJson = json_collection($entry->contest->entries, 'ContestEntry');
+        $entriesJson = json_collection($entry->contest->entries, new ContestEntryTransformer());
```
nanaya: The `entries` should have `contest` loaded

app/Http/Controllers/ContestEntriesController.php
```diff
@@ -19,7 +20,12 @@ class ContestEntriesController extends Controller
 {
     public function judgeResults($id)
     {
-        $entry = ContestEntry::with('contest')->findOrFail($id);
+        $entry = ContestEntry
+            ::with('contest')
```
nanaya: this got re-queried by the `load` below

database/migrations/2025_05_28_000001_add_total_score_std_to_contest_judge_votes.php
nanaya, line 1: can use newer date

resources/css/bem/contest-judge-results-scores.less
```diff
@@ -4,14 +4,17 @@
 .contest-judge-results-scores {
   display: grid;
   align-items: center;
-  grid-gap: 5px;
+  gap: 5px;
```
nanaya, line 7: should probably be more like `5px 10px` (or more). The horizontal gap seems a bit small otherwise.

app/Models/ContestEntry.php
```diff
@@ -52,6 +53,53 @@ public function votes()
...
+                };
+                $beatmapsetIdsQuery->where('diff_size', $keys);
+            }
+        }
+
+        return $query->whereIn('entry_url', $beatmapsetIdsQuery);
+    }
+
+    // TODO: switch Contest param with casted extra_options object.
+    public function scopeWithScore(Builder $query, Contest $contest): Builder
```
nanaya, line 78: looking again I don't know if it's better here compared to inside `Contest` as it depends on passing `$contest` which could very well be unrelated. It also looks weird when accessed from `$contest` - `$contest->entries()->withScore($contest)`.

I think I'd rather have `entriesWithScore()` in the `Contest` itself...

notbakaneko: Not all entries in the contest need the score when viewing a single entry, and I'd rather split all the types of scores/votes into separate scopes/functions instead of overloading everything into the same fields, but that's not happening until  I get around to replacing the existing contests coffeescript.

nanaya: as long it returns the builder just do `find(...)` on it...?

Issue #10455, opened 2023-08-11
Lazer multiplayer room pages do not list per-beatmap results


Walavouchey:
lazer lobbies, e.g. https://osu.ppy.sh/multiplayer/rooms/399751, do not show results on individual beatmaps, unlike stable lobbies e.g. https://osu.ppy.sh/community/matches/108242105

this is important for tournaments because these match result pages are saved for later auditing, which is a [requirement for getting badged](https://osu.ppy.sh/wiki/en/Tournaments/Official_support#tournaments)

found while archiving match results for https://github.com/ppy/osu-wiki/pull/9875

nanaya:
This may be problematic after moving multiplayer scores to solo score table since the values could change when there's scoring update. @peppy

bdach:
Bumping because this was requested for the purposes of running further tournaments on lazer.

As far as I can tell there are a few potential disparate ways forward here:

- Have the "new" multiplayer pages (`/multiplayer/rooms/{id}`) show history information in some new way that isn't there in place right now. This is a rather open avenue, you could do many things here, so I'm not even entirely sure where to start with that.
- Do something to reuse the "old" multiplayer pages (`community/matches/{id}`) such that they can work with lazer multiplayer rooms. I see two subpaths here:
  - Doing some magic web-side to transmogrify the data that we have into the shape that that page / endpoint expects
  - Having `osu-server-spectator` insert rows in `osu_mp` database to simulate operation of stable and thus create a match page automagically.

Things to note:

- Right now, for lazer mp rooms, out of all things on that match history page, we only really have scores in the database. We don't have user history (joins/leaves), we don't have host changes, we don't have kicks. I'm unsure of how much this matters and how much the tournament people rely on it (will probably ask around).
- One thing that popped to my mind is that this match history page is supposed to be "archival" in nature, i.e. it should be immutable. Which makes me partial to the "manually insert rows in `osu_mp`" approach, because that way, we can do stuff on the scores tables and it won't accidentally break archived match logs on wiki pages or whatever.

Open questions:

- Do we want this history scored for every single multiplayer room ever, or do we want to limit it to specific rooms?

Interested in opinions from all potential stakeholders (@peppy @smoogipoo @ppy/team-web)

peppy:
> Do we want this history scored for every single multiplayer room ever, or do we want to limit it to specific rooms?

For the legacy tables, we keep history around for n months, and have a `persist` flag for tournament matches  specifically.

[...]

PR #12030, opened 2025-03-21
Convert multiplayer match history event component to typescript

bdach:
This is preparatory work for https://github.com/ppy/osu-web/issues/10455.

The reason why I want the entire page converted to typescript first is that the path I have forward to resolving the issue relies on doing some data shape transformation of the backing json which is much easier to do (and probably review) when typings are in place.

The plan is to PR components one-by-one (should take like 6 PRs, including this one). This is somewhat guided by what I could gather is "the general precedent" (i.e. preferring function-based components rather than class-based if possible, usage of `classWithModifiers()`, etc.), but is more or less a transcription of the previous logic guided by the typings and jslint. I'm still not sure on some things, e.g. when to `null` or to `undefined` so I would appreciate patience as I get my bearings in that regard.

The typing included here will cover the entire page already and should not get changed anymore until the typescript conversion is complete, unless I didn't rearrange my commits properly.


resources/js/interfaces/score-json.ts
```diff
@@ -21,7 +21,7 @@ interface ScoreCurrentUserAttributesJson {
 interface Match {
   pass: boolean;
   slot: number;
-  team: number;
+  team: 'blue' | 'none' | 'red';
```
bdach, line 24: Rationale for this change is that in the actual serialised json this appears to be a string, probably thanks to

https://github.com/ppy/osu-web/blob/026c39d05218ce2034693c46bceab7479895acc9/app/Models/LegacyMatch/Score.php#L87-L90

I wasn't sure whether I should be changing the attribute definition on the php model because it's confusing (in `artisan tinker` when I fetch the entire model the object dump shows an integer inside, but when I access the actual property it magically stringifies itself) so I didn't.

resources/js/mp-history/event.tsx
```diff
@@ -0,0 +1,56 @@
...
+  'player-left': ['fas fa-arrow-left', 'far fa-circle'],
+};
+
+export default function Event(props: Props) {
+  const user = props.event.user_id != null ? props.users[props.event.user_id] : undefined;
+
+  const event_type = props.event.detail.type;
+  if (event_type === 'other') {
+    return null;
+  }
```
bdach, line 33: Explicitly handled even though it wasn't before because otherwise ts yells at me that I can't use this to index into `icons` later. And none of this basically can handle the `other` type anyway (translations included).

resources/js/mp-history/event.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  let userLink = '';
+
+  if (user != null && event_type !== 'match-disbanded') {
+    userLink = linkHtml(route('users.show', { user: user.id }), user.username, { classNames: ['mp-history-event__username'] });
+  }
+
+  return (
+    <div className={'mp-history-event'}>
```
nanaya, line 42: just a quick comment here but `{}` isn't needed for plain string

resources/js/interfaces/score-json.ts
bdach: I made an attempt in https://github.com/ppy/osu-web/pull/12030/commits/32b1e29c6d7337eb1482d868c59b6b3aaffd28b7, although I'm not sure as to whether you were expecting it to change to `string` or `mixed` (I did the latter, because of its weird stringy-numbery duality as described above).

resources/js/mp-history/event.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <div className={'mp-history-event__time'}>
+        <TimeWithTooltip dateTime={props.event.timestamp} format={'LTS'} />
+      </div>
+      <div className={classWithModifiers('mp-history-event__type', [event_type])}>
+        {icons[event_type].map((m) => <i key={m} className={m} />)}
+      </div>
+      <div
+        dangerouslySetInnerHTML={{
+          __html: trans(`matches.match.events.${event_type}${user != null ? '' : '-no-user'}`, { user: userLink }),
```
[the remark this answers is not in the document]
bdach: https://github.com/ppy/osu-web/pull/12030/commits/230ee03590cb5bb1eb277f33a30e2f0c75a3a277 should do it I think? As a side effect the user links gained cards on hover, not sure how (un)desirable that is.

nanaya (changes requested):

resources/js/interfaces/match-json.ts
```diff
@@ -0,0 +1,70 @@
...
+import UserJson from './user-json';
+
+export default interface Match {
+  current_game_id?: number;
+  events: MatchEvent[];
+  first_event_id: number;
+  latest_event_id: number;
+  match: MatchDetails;
+  users: UserJson[];
+}
```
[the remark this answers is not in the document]
bdach: Have removed for now, will bring back for the last pull in conversion series when the relevant component is typescriptified.

resources/js/interfaces/match-json.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export interface MatchDetails {
```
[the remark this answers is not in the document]
bdach: I'm not sure how long the legacy prefix is going to live since in my dirty draft I didn't end up changing the shape of any of this *that* much - new multiplayer is much more malleable into the old since it... sort of has less features available right now.

However, applied for now in https://github.com/ppy/osu-web/pull/12030/commits/bed08781530a64db4645e0c7d814b75f71121ff1, we can cross that bridge when we get to it - I feel like that part is likely to get rewritten entirely because you guys will tell me what are the more proper ways to do everything that I did in the dirty draft..

resources/js/mp-history/event.tsx
```diff
@@ -0,0 +1,62 @@
...
+          user == null
+            ? trans(`matches.match.events.${eventType}-no-user`)
+            : <StringWithComponent
+              mappings={{
+                user: (<UserLink
+                  className={'mp-history-event__username'}
+                  user={{
+                    id: user.id,
+                    username: user.username,
+                  }}
```
nanaya, line 54:
```suggestion
                  className='mp-history-event__username'
                  user={user}
```

nanaya (approved):

PR #12041, opened 2025-03-24
Update to Laravel 11

nanaya:
The main major changes here are Carbon's `diffIn*` now defaults to non-absolute value, the same function returning float instead of int, and unsigned floats db column datatype isn't a thing anymore (and deprecated by mysql).

- [x] depends on #12032


notbakaneko (changes requested):
lockfile conflict

nanaya:
but the dependent pr hasn't even merged 🤔

notbakaneko (approved):
seems fine..... (famous last words)

Issue #12033, opened 2025-03-21
Separate user tags from beatmapset tags on the beatmap info page


felipemarins:
I really like the new user tag feature, but I think it's kind of confusing that it currently appears together with the "normal" beatmapset tags. I think it would be better if there was a proper new section dedicated to them (maybe above the beatmapset tags) with the possibility of showing a description of each tag on mouse hover and the number of votes each tag got.

PR #12042, opened 2025-03-24
Separate user tags on beatmapset info

tsunyoku:
Resolves https://github.com/ppy/osu-web/issues/12033.

image


resources/lang/en/beatmapsets.php
```diff
@@ -148,8 +148,9 @@
             'source' => 'Source',
             'storyboard' => 'This beatmap contains storyboard',
             'success-rate' => 'Success Rate',
-            'tags' => 'Tags',
```
notbakaneko, line 151: use a new key

PR #12044, opened 2025-03-25
Fix parsing multiple quoted search options

nanaya:
mm regexp. This disallows `artist="something "weird""`; inner quotes must be escaped `\"`.

The behaviour for `tag="hello world"aa` is a bit weird ┐( ･_･)┌


notbakaneko (approved):
Does it need any client updates for escaping?

nanaya:
I guess so but it's kind of a rare case anyway

PR #12047, opened 2025-03-25
Use fully quoted tags for Beatmapset user tag links

notbakaneko:


resources/js/beatmapsets-show/info.tsx
```diff
@@ -198,7 +198,7 @@ export default class Info extends React.Component<Props> {
                   <React.Fragment key={`${tag.name}-${i}`}>
                     <a
                       className='beatmapset-info__link'
-                      href={route('beatmapsets.index', { q: tag.name })}
+                      href={route('beatmapsets.index', { q: makeSearchQueryOption('tag', tag.name) })}
```
nanaya, line 201: the function initially has something like `esExact` parameter but then removed because not yet needed but I suppose it's needed now?

Or update the search library to remove the extra quotes when doing db lookup

resources/js/beatmapsets-show/info.tsx
```diff
@@ -198,7 +198,7 @@ export default class Info extends React.Component<Props> {
                   <React.Fragment key={`${tag.name}-${i}`}>
```
nanaya, line 198: unrelated but the `${i}` isn't needed for this one

or maybe can even use tag.id instead

nanaya (changes requested):

PR #12050, opened 2025-03-26
Convert game header component to typescript

bdach:


resources/js/mp-history/game-header.tsx
```diff
@@ -0,0 +1,80 @@
...
+                endTime,
+                startTime,
+              }}
+              pattern=':startTime - :endTime' />
+            : <StringWithComponent
+              mappings={{
+                inProgress: trans('matches.match.in-progress'),
+                startTime,
+              }}
+              pattern=':startTime :inProgress' />}
```
bdach, line 60: Oh, I forgot to mention - I'm kinda inventing with this one because I don't think anything else does ad-hoc patterns like this (at least I can't grep any) but maybe this is better than the previous code which was kinda hard to parse?

nanaya: a plain `<>{startTime} - {endTime}</>` and `<>{startTime} {trans...}</>` would probably be just as readable?

resources/js/mp-history/game-header.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { trans } from 'utils/lang';
+
+interface Props {
+  beatmap: BeatmapJson;
+  beatmapset: BeatmapsetJson;
+  game: LegacyMatchGameJson;
+}
+
+export default function GameHeader(props: Props) {
+  const timeFormat = 'LTS';
```
nanaya, line 24: can be outside of the function

nanaya (approved):

PR #12054, opened 2025-03-27
Allow changing team leader

### LINKED ISSUE #11914: Option to change team leader
Dunno if planned, but would like the ability to swap who is the team leader in the newly added osu teams feature

nanaya:
Resolves #11914.


app/Http/Controllers/Teams/MembersController.php
```diff
@@ -48,4 +48,24 @@ public function index(string $teamId): Response
...
+        ])->firstOrFail();
+        $team = $teamMember->team;
+
+        priv_check('TeamUpdate', $team)->ensureCan();
+
+        $team->update(['leader_id' => $teamMember->user_id]);
+
+        \Session::flash('popup', osu_trans(
+            'teams.members.set_leader.success',
+            ['user' => $teamMember->userOrDeleted()->username],
```
notbakaneko, line 66: does it even make sense to transfer to a deleted or restricted user?

notbakaneko (changes requested):

app/Http/Controllers/Teams/MembersController.php
nanaya: I'm not quite sure where to put the validation/authorization (which one is it even) so it's inlined in the controller action for now ಠ_ಠ

notbakaneko (approved):
Can probably also disable the button for changing to "deleted" users later

PR #12056, opened 2025-03-27
Add sentry tracking for slow requests

nanaya:
This requires enabling tracing with sampling rate of at least 0... which I vaguely remember still impacts performance? ┐( ･_･)┌


notbakaneko (approved):
let's find out :shipit:

nanaya:
@peppy will need to set the env to enable this

PR #12059, opened 2025-03-28
Add per ruleset tag support

### LINKED ISSUE #12057: Adding per ruleset tag support
- Add nullable `ruleset_id` column to tags table
- Only allow tags to be specified for `null` or current ruleset id

@notbakaneko this should be a pretty simple one, but also quite high priority. we already have data to [populate and update](https://docs.google.com/spreadsheets/d/13THVRdUPziX9oJxNHnW31NEmqZLs306tq8lZYwidc9w/edit?gid=0#gid=0) the tags table when this is ready.

notbakaneko:
closes #12057

### DIFF
--- a/app/Http/Controllers/BeatmapTagsController.php
+++ b/app/Http/Controllers/BeatmapTagsController.php
@@ -7,6 +7,7 @@ declare(strict_types=1);
 
 namespace App\Http\Controllers;
 
+use App\Exceptions\InvariantException;
 use App\Models\Beatmap;
 use App\Models\BeatmapTag;
 use App\Models\Tag;
@@ -41,6 +42,10 @@ class BeatmapTagsController extends Controller
         priv_check('BeatmapTagStore', $beatmap)->ensureCan();
 
         $tag = Tag::findOrFail($tagId);
+
+        if ($tag->ruleset_id !== null && $tag->ruleset_id !== $beatmap->playmode) {
+            throw new InvariantException();
+        }
 
         $tag
             ->beatmapTags()
--- a/app/Http/Controllers/BeatmapTagsController.php
+++ b/app/Http/Controllers/BeatmapTagsController.php
@@ -44,7 +44,7 @@ class BeatmapTagsController extends Controller
         $tag = Tag::findOrFail($tagId);
 
         if ($tag->ruleset_id !== null && $tag->ruleset_id !== $beatmap->playmode) {
-            throw new InvariantException();
+            throw new InvariantException(osu_trans('beatmap_tags.update.invalid_ruleset'));
         }
 
         $tag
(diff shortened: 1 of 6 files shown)


database/migrations/2025_03_28_000001_add_ruleset_id_to_tags.php
```diff
@@ -0,0 +1,31 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('tags', function (Blueprint $table) {
+            $table->smallInteger('ruleset_id')->nullable()->unsigned()->after('name');
+            $table->dropIndex('tags_name_unique');
+            $table->unique(['name', 'ruleset_id'], 'tags_name_ruleset_id_unique');
```
nanaya, line 19: rows with null ruleset id won't have the unique constraint...

nanaya: I wonder if this would make it confusing on whether or not there's database level constraint on uniqueness 🤔

notbakaneko: Apparently not needed since they're inserted manually 🤷

app/Http/Controllers/BeatmapTagsController.php
```diff
@@ -42,6 +43,10 @@ public function update($beatmapId, $tagId)
 
         $tag = Tag::findOrFail($tagId);
 
+        if ($tag->ruleset_id !== null && $tag->ruleset_id !== $beatmap->playmode) {
+            throw new InvariantException();
```
nanaya, line 47: an actual message...?

tests/Controllers/BeatmapTagsControllerTest.php
```diff
@@ -44,22 +54,27 @@ public function testStoreNoScore(): void
             ->assertForbidden();
     }
 
-    public function testDestroy(): void
+    public function testUpdateWrongRulesetId(): void
```
nanaya, line 57: also test with correct ruleset id?

nanaya (changes requested):

resources/lang/en/beatmap_tags.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+return [
+    'update' => [
+        'invalid_ruleset' => 'This ag is not valid with this Beatmap.',
```
nanaya, line 8:
```suggestion
        'invalid_ruleset' => 'This tag is not valid with this Beatmap.',
```

nanaya (approved):

peppy:
cc @bdach for visibility

PR #12062, opened 2025-03-31
Fix testUsernameChangeCostWindow setup

notbakaneko:
Use a known fixed number instead of  `while < $now`

### DIFF
--- a/tests/Models/UserTest.php
+++ b/tests/Models/UserTest.php
@@ -236,7 +236,7 @@ class UserTest extends TestCase
     public function testUsernameChangeCostWindow(int $years, int $cost)
     {
         $now = CarbonImmutable::now();
-        $this->travelTo(CarbonImmutable::now()->subYears(3));
+        $this->travelTo(CarbonImmutable::now()->subYearsNoOverflow(3));
 
         $user = User::factory()->create();
         // every 6 months for 3 years = 6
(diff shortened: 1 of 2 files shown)


tests/Models/UserTest.php
```diff
@@ -239,14 +239,16 @@ public function testUsernameChangeCostWindow(int $years, int $cost)
         $this->travelTo(CarbonImmutable::now()->subYears(3));
 
         $user = User::factory()->create();
-        while (CarbonImmutable::now()->isBefore($now)) {
+        // every 6 months for 3 years = 6
+        // using isBefore to setup adds too many when run at month boundaries.
+        for ($i = 0; $i < 6; $i++) {
```
nanaya, line 244: although it's not for another three years, there's a vaguely similar problem when run on feb 29

nanaya (changes requested):

tests/Models/UserTest.php
notbakaneko: it ironically works fine because `usernameChangeCost` doesn't check if the date is in the future

nanaya: still failing with `$now = CarbonImmutable::now()->subYears(1)->startOfMonth()->subDays(1);`...

nanaya (approved):

PR #12066, opened 2025-04-02
Fix error when checking blocked user ids for notification

nanaya:
That's a thing.

### DIFF
--- a/app/Jobs/Notifications/BroadcastNotificationBase.php
+++ b/app/Jobs/Notifications/BroadcastNotificationBase.php
@@ -248,7 +248,7 @@ abstract class BroadcastNotificationBase implements ShouldQueue
                 ...UserRelation
                     ::where('zebra_id', $this->source->getKey())
                     ->where('foe', true)
-                    ->whereIn('user_id', $userIds)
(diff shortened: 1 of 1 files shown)


app/Jobs/Notifications/BroadcastNotificationBase.php
```diff
@@ -241,13 +242,18 @@ private function excludeBlockedUserIds(array $userIds): array
...
-            ->whereIn('user_id', $userIds)
-            ->pluck('user_id')
-            ->all();
+        $excludedReceiverIds = new Set();
+        foreach (array_chunk($userIds, 10000) as $chunkedUserIds) {
+            $excludedReceiverIds->add(
+                ...UserRelation
+                    ::where('zebra_id', $this->source->getKey())
+                    ->where('foe', true)
+                    ->whereIn('user_id', $userIds)
```
notbakaneko, line 251: `$chunkedUserIds`?

Also maybe make `$userIds` into a `Set` and remove from there?

notbakaneko (changes requested):

notbakaneko (approved):

PR #12068, opened 2025-04-02
Convert game score component to typescript

bdach:


resources/js/mp-history/score.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+const firstRow = ['combo', 'accuracy', 'score'];
+const secondRow: ScoreStatisticsAttribute[] = ['count_geki', 'count_300', 'count_katu', 'count_100', 'count_50', 'count_miss'];
+
+export default function Score(props: Props) {
+  const user = props.users[props.score.user_id];
+
+  if (user == null) {
+    return null;
+  }
```
bdach, line 29: Not yet sure what the accepted solution is for handling "impossible" situations like these. Previous code didn't care about this missing, so maybe it's fine?

Note that a similar situation happens below with `props.score.match` being possibly undefined twice, but in those cases there are *somewhat sane* fallbacks available, so I just handled locally. Maybe a `props.score.match == null` check should be here too.

notbakaneko: `throw new Error`

resources/js/mp-history/score.tsx
```diff
@@ -0,0 +1,103 @@
...
+              let modifier = 'medium';
+              let value;
+
+              switch (m) {
+                case 'combo':
+                  value = formatNumber(props.score.max_combo);
+                  break;
+
+                case 'accuracy':
+                  value = `${formatNumber(props.score.accuracy * 100, 2)}%`;
```
nanaya, line 67:
```suggestion
                  value = formatNumber(props.score.accuracy, 2, { style: 'percent' });
```

resources/js/mp-history/score.tsx
```diff
@@ -0,0 +1,103 @@
...
+      <div
+        className='mp-history-player-score__shapes'
+        style={{ backgroundImage: `url(/images/layout/mp-history/shapes-team-${props.score.match.team ?? 'none'}.svg)` }} />
+      <div className='mp-history-player-score__main'>
+        <div className={classWithModifiers('mp-history-player-score__info-box', ['user'])}>
+          <div className='mp-history-player-score__username-box'>
+            <a className='mp-history-player-score__username' href={route('users.show', { user: user.id })}>{user.username}</a>
+            {!props.score.match.pass
+              ? <span className='mp-history-player-score__failed'>{trans('matches.match.failed')}</span>
+              : null}
```
nanaya, line 46: `false` is [special cased](https://react.dev/learn/conditional-rendering#logical-and-operator-) in react so if there's nothing to render when false, `{testCondition && <Thing />}` works (apparently also for `true`/`||` but doesn't seem to be documented?)

nanaya (changes requested):

resources/js/mp-history/score.tsx
```diff
@@ -0,0 +1,101 @@
...
+              let modifier = 'medium';
+              let value;
+
+              switch (m) {
+                case 'combo':
+                  value = formatNumber(props.score.max_combo);
+                  break;
+
+                case 'accuracy':
+                  value = formatNumber(props.score.accuracy * 100, 2, { style: 'percent' });
```
nanaya, line 65: percent style already handles the `* 100` internally 👀

bdach: 🤦 I even opened my test match page and checked on the failed thing but like completely forgot to double check the percent formatting

nanaya (approved):

PR #12071, opened 2025-04-03
Filter chat messages from block users from being displayed

notbakaneko:
Went with just not showing the message instead of a placeholder blocked message


nanaya (approved):
unread message shouldn't be shown on incoming blocked message but it can be dealt with separately I guess

PR #12072, opened 2025-04-04
Fix chat messages not being received in some cases

notbakaneko:
- [x] ppy/osu-notification-server#60
Waits for notification server to mark socket as ready before sending messages.
Currently, it's possible for the `chat.start` message to be sent before the notification server has attached the event listeners, causing the socket state to not correctly mark chat being active.
- [x] the above being deployed


resources/js/socket-worker.ts
```diff
@@ -97,11 +103,6 @@ export default class SocketWorker {
 
     const token = tokenEl.getAttribute('content');
     this.ws = new WebSocket(`${this.endpoint}?csrf=${token}`);
-    this.ws.addEventListener('open', action(() => {
-      this.retryDelay.reset();
```
nanaya, line 101: just gone? 🤔

nanaya (approved):

peppy:
Deploy is done now.

peppy (approved):

PR #12074, opened 2025-04-04
Validate mods against freestyle allowance

smoogipoo:
- [ ] Depends on https://github.com/ppy/osu-tools/pull/257

Freestyle items are allowed to include any allowed/required mods that are also allowed via `ValidForFreestyle`.


nanaya (approved):
Seems correct

smoogipoo:
I've removed the dependency on osu-tools since I've included `mods.json` here.

nanaya (approved):
Seems correct (2)

Just remember that if osu-tools gets updated separately without the change for this, it'll create a pr undoing this.

PR #12076, opened 2025-04-04
Combine performance and score ranking pages (+ cleanup)

nanaya:
The action function is still too big 🤨 

- [x] depends on #12065


notbakaneko (changes requested):
A quick one, but switching between score and performance should probably preserve the country filter?

app/Http/Controllers/RankingController.php
```diff
@@ -364,51 +378,97 @@ public function kudosu()
         return ext_view('rankings.kudosu', compact('scores'));
     }
 
-    public function spotlight($mode)
+    private function maxResults(int $rulesetId, ?CountryStatistics $countryStats, Builder $stats, array $params): int
     {
-        $chartId = get_int($this->params['spotlight'] ?? null);
+        switch ($params['type']) {
```
notbakaneko, line 383: missing `default` return/throw

notbakaneko (changes requested):

app/Http/Controllers/RankingController.php
nanaya: 🤔 php will throw magically due to missing return?

notbakaneko: good point 🤔
guess we can deal with static analysis problem later (so never)

notbakaneko (approved):

PR #12080, opened 2025-04-07
Prevent invalid image output size

nanaya:
Weird edge case which has been adding noise to sentry.


notbakaneko (approved):
that's an interesting image size ( ﾟ ヮﾟ)

PR #12081, opened 2025-04-07
Convert game component to typescript

bdach:


resources/js/mp-history/game-header.tsx
```diff
@@ -24,7 +24,7 @@ const timeFormat = 'LTS';
 export default function GameHeader(props: Props) {
   let title = getTitle(props.beatmapset);
   const version = props.beatmap.version;
-  if (version != null) {
+  if (version !== '') {
```
bdach, line 27: This is forced by changes to `deletedBeatmap()` in `Game` - previous code would put a version of `null` when using the fallback to a deleted beatmap, typings be damned, while now I've had to make the function obey to the typings, and `null` is no longer permitted in the `version` field, so empty string is the next best thing.

resources/js/mp-history/game.tsx
```diff
@@ -0,0 +1,147 @@
...
+import { classWithModifiers } from 'utils/css';
+import { formatNumber } from 'utils/html';
+import { trans, transExists } from 'utils/lang';
+import GameHeader from './game-header';
+import Score from './score';
+
+interface TeamScores {
+  blue: number;
+  red: number;
+}
```
bdach, line 22: This is a temporary place for this interface; it will be lifted up from here as the typescript conversion goes further.

resources/js/mp-history/game.tsx
```diff
@@ -0,0 +1,147 @@
...
+};
+
+interface Props {
+  event: LegacyMatchEventJson;
+  teamScores: TeamScores;
+  users: Partial<Record<number, UserJson>>;
+}
+
+export default function Game(props: Props) {
+  const game = props.event.game!;
```
nanaya, line 35:
```suggestion
  const game = props.event.game ?? fail('game field is missing');
```

resources/js/mp-history/game.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  if (game.scores.some((score) => score.match == null)) {
+    throw new Error('scores are missing match data');
+  }
```
nanaya, line 39: `!` is generally avoided and I think `scores` field should have match specific type instead.

resources/js/mp-history/game.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const winningTeam = props.teamScores.blue > props.teamScores.red ? 'blue' : 'red';
+  const difference = Math.abs(props.teamScores.blue - props.teamScores.red);
+
+  let sortedScores = game.scores.map((m) => {
+    const sortedScore = m as SortedScore;
+    sortedScore.teamRank = m.match!.team === winningTeam ? 1 : 2;
+    return sortedScore;
+  });
+
+  sortedScores = _.orderBy(sortedScores, ['teamRank', 'score'], ['asc', 'desc']);
```
nanaya, line 52: using plain sort with relevant callback function will remove the need for the `& { teamRank: ...}` type

resources/js/mp-history/game.tsx
```diff
@@ -0,0 +1,147 @@
...
+            key={m.match!.slot}
+            mode={game.mode}
+            score={m}
+            users={props.users} />
+        ))}
+      </div>
+      {showTeams && props.event.game?.end_time != null &&
+        <div>
+          <div className='mp-history-game__team-scores'>
+            {['red', 'blue'].map((m) => (
```
nanaya, line 72:
```suggestion
            {(['red', 'blue'] as const).map((m) => (
```
and remove the casting inside

resources/js/mp-history/game.tsx
```diff
@@ -0,0 +1,147 @@
...
+  });
+
+  sortedScores = _.orderBy(sortedScores, ['teamRank', 'score'], ['asc', 'desc']);
+
+  return (
+    <div className='mp-history-game'>
+      <GameHeader
+        beatmap={game.beatmap ?? deletedBeatmap(game.mode)}
+        beatmapset={game.beatmap?.beatmapset ?? deletedBeatmapset()}
+        game={game} />
```
nanaya, line 59: might as well update GameHeader to only need `game` props and get the beatmap(set) from it

nanaya (changes requested):

resources/js/mp-history/game.tsx
nanaya: actually, just pass non-null `game` as props directly

resources/js/mp-history/game.tsx
bdach: > should have match specific type instead.

Not sure where you want to have that type live. I put it next to `LegacyMatchGameJson` in https://github.com/ppy/osu-web/pull/12081/commits/1c2bd50bf2116ef7e683da383381379b1cf69df3, but that required an extra export that may be considered a touch ugly.

nanaya: I think it's usually (or sometimes) in the same file as the base json (score-json in this case)

resources/js/interfaces/legacy-match-game-json.ts
```diff
@@ -13,12 +13,15 @@ export default interface LegacyMatchGameJson {
...
   mods: string[]; // TODO: use ModJson
-  scores: ScoreJson[]; // TODO: use SoloScoreJson
+  scores: LegacyMatchScoreJson[]; // TODO: use SoloScoreJson
   scoring_type: LegacyMatchScoringType;
   start_time: string;
   team_type: LegacyMatchTeamType;
 }
 
+// used to ensure presence of match
+export type LegacyMatchScoreJson = ScoreJson & { match: Match };
```
nanaya, line 23: I think this should work better in term of automatically matching type

```suggestion
export type LegacyMatchScoreJson = ScoreJson & Require<Pick<ScoreJson, 'match'>>
```

nanaya (approved):

PR #12083, opened 2025-04-07
Fix overlapping scrollable Beatmap info sections

### LINKED ISSUE #12051: "User Tags" and "Mapper Tags" overlap
I don't know if this issue exists on PC, but I'm using a tablet and it's happening

Image

notbakaneko:
The entire box is scrollable instead.
Should fix #12051

### DIFF
--- a/resources/css/bem/beatmapset-info.less
+++ b/resources/css/bem/beatmapset-info.less
@@ -81,5 +81,20 @@
 
   &__scrollable {
     overflow-y: auto;
+
+    &::after {
+      content: "";
+      position: absolute;
+      bottom: 0;
+      left: 0;
+      width: 100%;
+      height: 10px;
+      background-image: linear-gradient(
+        to bottom,
+        hsla(var(--hsl-b4), 0),
+        hsl(var(--hsl-b4))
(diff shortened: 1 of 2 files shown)


resources/css/bem/beatmapset-info.less
```diff
@@ -74,24 +64,36 @@
   }
 
   &__header {
+    position: sticky;
```
nanaya, line 67: this effect is kinda weird for all but tags since they usually only have single value row

resources/js/beatmapsets-show/info.tsx
```diff
@@ -111,127 +111,129 @@ export default class Info extends React.Component<Props> {
...
-                    <a
-                      className='beatmapset-info__link'
-                      href={route('beatmapsets.index', { q: tag })}
-                    >
-                      {tag}
-                    </a>
-                    {' '}
-                  </React.Fragment>
-                ))}
+            {this.controller.tags.userTags.length > 0 &&
```
nanaya, line 192: not directly related to this pr but apparently user tags should still be shown for converts

PR #12086, opened 2025-04-08
Convert multiplayer room history content component to typescript

bdach:


resources/js/mp-history/content.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  getSnapshotBeforeUpdate(prevProps: Readonly<Props>, _: Readonly<any>): Snapshot {
+    const snapshot: Snapshot = {
+      scrollToLastEvent: prevProps.isAutoloading && this.props.isAutoloading && bottomPageDistance() < 10,
+    };
+
+    if (!snapshot.scrollToLastEvent) {
+      const hadEvents = prevProps.events != null && prevProps.events.length > 0;
+      const hasEvents = this.props.events != null && this.props.events.length > 0;
+      if (hadEvents && hasEvents) {
```
bdach, line 63: Bit verbose, but this is the only way that I could find that shuts the compiler up about being able to access the first element inside the block...

resources/js/mp-history/content.tsx
```diff
@@ -0,0 +1,179 @@
...
+            }
+
+            return (
+              <React.Fragment key={event.id}>
+                {(() => {
+                  if (inEvent) {
+                    inEvent = false;
+                    return eventsGroupClose;
+                  }
+                })()}
```
bdach, line 108: There's probably a better way to do this, but I don't know it 😅

resources/js/mp-history/content.tsx
```diff
@@ -0,0 +1,175 @@
...
+interface Props {
+  currentGameId?: number;
+  events?: LegacyMatchEvent[];
+  hasNext: boolean;
+  hasPrevious: boolean;
+  isAutoloading: boolean;
+  loadingNext: boolean;
+  loadingPrevious: boolean;
+  loadNext: (state: any) => void; // TODO: type is temporary, will be specified better when `main.coffee` is converted to TS
+  loadPrevious: (state: any) => void; // TODO: type is temporary, will be specified better when `main.coffee` is converted to TS
```
nanaya, line 25: or just `() => void`, its actual type (as expected by ShowMoreLink)

resources/js/mp-history/content.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+export interface TeamScores {
+  blue: number;
+  red: number;
+}
+
+export default class Content extends React.PureComponent<Props> {
+  scoresCache: Partial<Record<number, TeamScores>> = {};
+
+  componentDidUpdate(prevProps: Readonly<Props>, prevState: Readonly<any>, snapshot?: Snapshot) {
```
nanaya, line 44: better use `unknown` instead of `any`

resources/js/mp-history/content.tsx
```diff
@@ -0,0 +1,175 @@
...
+      $(window).stop().scrollTo(document.body.scrollHeight, 500);
+    } else if (snapshot?.referenceFunc) {
+      const referenceCurrent = snapshot.referenceFunc();
+      const documentScrollTopCurrent = window.scrollY;
+      const documentScrollTopTarget = documentScrollTopCurrent + referenceCurrent - snapshot.referencePrev!;
+      window.scrollTo(window.scrollX, documentScrollTopTarget);
+    }
+  }
+
+  getSnapshotBeforeUpdate(prevProps: Readonly<Props>, _: Readonly<any>): Snapshot {
```
nanaya, line 55: `_` kinda shadows global lodash. Maybe call it its actual name prevState just like in the other function (and use `unknown` too)

resources/js/mp-history/content.tsx
```diff
@@ -0,0 +1,175 @@
...
+            <ShowMoreLink
+              callback={this.props.loadNext}
+              hasMore
+              loading={this.props.isAutoloading || this.props.loadingNext} />
+          </div>}
+      </div>
+    );
+  }
+
+  teamScores(game: LegacyMatchGame): TeamScores {
```
nanaya, line 152: mark private

nanaya (changes requested):

resources/js/mp-history/content.tsx
bdach: > ...or just remove it altogether as it's not actually used

I'm very not used to just being able to do that 🤯 I figured since it was a react method it had to have a matching signature but I guess that's just not how js rolls is it

nanaya (changes requested):

PR #12089, opened 2025-04-09
Allow fetching up to 200 best scores

### LINKED ISSUE #7674: Allow retrieving more user's best scores
Currently `/users/.../scores/best` allows only the first 100 scores to be retrieved. Since it already supports pagination, it would be good to allow access to more scores, while limiting single query to 100 (if #7673 goes through) scores.

nanaya:
Resolves #7674.

Maybe should look into using the scores table directly again instead of through es (ref: #3363).


peppy:
> Maybe should look into using the scores table directly again instead

I didn't realise we added an index which can cover this. 🤔

peppy (approved):

notbakaneko (approved):
thought I checked this already 🤨

PR #12091, opened 2025-04-10
Fix username availability time display

nanaya:
Carbon 3's `diff()->days` is currently broken (briannesbitt/Carbon#3018). It should've only affected the display, not actual availability.


notbakaneko (approved):

notbakaneko:
(also how was this not a breaking change)

peppy:
😥

PR #12094, opened 2025-04-10
Add passed flag to playlist item attempts json

nanaya:
To be used for display by client. @smoogipoo


smoogipoo (dismissed):
Looks to work as expected

peppy (approved):

PR #12095, opened 2025-04-10
Use correct check for room aggregate completed value

nanaya:
Recent change to failing plays on realtime mode still being recorded (including the score_id) means its existence doesn't imply passing anymore.

- [ ] depends on #12094


nanaya:
nnnnever mind (calling the field `passed` instead and `completed` can stay being whatever it currently means)

PR #12115, opened 2025-04-16
Update Beatmap info user tag style

notbakaneko:
Displays it more like the game client
image

Some tags are kind of long and don't really fit in the existing box, though:
image

Now with page theme colours instead:
image
image

`h1` colour if user voted for that tag
image

I don't know what to do with the on-hover colours since multi-colour background buttons were never a thing on web before 👀

### DIFF
--- a/resources/css/bem/user-tag.less
+++ b/resources/css/bem/user-tag.less
@@ -4,24 +4,23 @@
 .user-tag {
   .default-border-radius();
   .link-plain();
-  .link-white();
   display: flex;
   overflow: hidden;
+  color: hsl(var(--hsl-c1));
 
   &__item {
     padding: 3px 6px;
 
     &--category {
-      background: #444;
-      color: #aaa;
+      background: hsl(var(--hsl-b3));
     }
 
     &--count {
-      background: #222;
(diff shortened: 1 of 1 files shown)


nanaya:
where did these colors come from...

bdach:
> where did these colors come from...

My ass, quite frankly. The entire results screen in client is a mess of bleh grays.

Maybe you can get @arflyte to find better ones? 🤷

nanaya:
also it's kinda long with pretty much just one tag per "line"

bdach:
The place where this design is used in client (results screen) has much more width to work with 🤷

Me just a grug programmer I ship feature I not into good UX design.

notbakaneko:
The entire space for beatmap metadata is kind of narrow, even on the tentative-not-quite-new-anymore design, everything is crammed into too narrow columns...

Hiviexd:
seems like this slipped between the cracks?

IMO for longer tags you could truncate the text and show the full tag text in a tooltip as a workaround until the beatmap page has better structure for this.

notbakaneko:
I'll have another look after finishing some other stuff on the beatmap page

PR #12131, opened 2025-04-22
Don't include user tags from general query search

### LINKED ISSUE #12084: Excluding map tags from search doesn't work

nanaya:
Simplest fix I can think of 🤷 Resolves #12084.

currently, given map with tag "what" and no user tag, with search query `-what`:

- ``->should(['term' => ['_id' => ['value' => $this->params->queryString, 'boost' => 100]]])``
  no match -> no hit (correct)
- ``->should(QueryHelper::queryString($this->params->queryString, $partialMatchFields, 'or', 1 / count($terms)))``
  matches excluded term -> no hit (correct)
- ``->should(QueryHelper::queryString($this->params->queryString, [], 'and'))``
  matches excluded term -> no hit (correct)
- ``->should(['nested' => ['path' => 'beatmaps', ...]])``
  doesn't match excluded term -> hit (incorrect)

See https://github.com/ppy/osu-web/pull/12131#issuecomment-3645279424

### DIFF
--- a/app/Libraries/Search/BeatmapsetSearch.php
+++ b/app/Libraries/Search/BeatmapsetSearch.php
@@ -61,12 +61,6 @@ class BeatmapsetSearch extends RecordSearch
                     ->should(['term' => ['_id' => ['value' => $this->params->queryString, 'boost' => 100]]])
                     ->should(QueryHelper::queryString($this->params->queryString, $partialMatchFields, 'or', 1 / count($terms)))
                     ->should(QueryHelper::queryString($this->params->queryString, [], 'and'))
-                    ->should([
-                        'nested' => [
-                            'path' => 'beatmaps',
-                            'query' => QueryHelper::queryString($this->params->queryString, ['beatmaps.top_tags'], 'or', 0.5 / count($terms)),
-                        ],
-                    ])
             );
         }
 


peppy:
@notbakaneko can you give this 15 mins and see if we can get this in? been hanging for a long time now 😅

notbakaneko (commented):
We really shouldn't keep trying to shoehorn our query syntax into lucene's `query_string`/`simple_query_string` and handle it properly instead #12364, even more so when handling term exclusions

nanaya:
well, the goal for this pr was just a quick fix for the new bug introduced by the user tag and nothing else 🤷

nanaya:
I wonder if user tag should just not be searchable with general query but require the tag= filter instead. It matches being a difficulty specific search and you probably don't want it to match unrelated difficulty when combined with other filters

notbakaneko:
I think that depends on whether users are treating it like another plain text field.

I think that at least for exclusions in the general query, they should not be filtering based on fields you don't really see on the main results, and need an explicit filter to remove them, not just for user tags.

nanaya:
short of switching the entire search to be based on beatmap (difficulty), it's probably better to keep the general search limited to beatmapset level attributes

notbakaneko (approved):

PR #12134, opened 2025-04-22
Better beatmap search query parsing

### LINKED ISSUE #10916: Beatmap search exclusion only looks at tags
If you search using the minus operator e.g. `-anime` you will still get beatmaps that are categorised as "Anime" in genres, but don't have "anime" in their tags. 

Apparently including song genre and language is required as per ranking criteria and this issue shouldn't exist, but it does.

Making the search look at the genre and language would also help with discoverability of graveyarded maps or maps in development

### LINKED ISSUE #12084: Excluding map tags from search doesn't work

notbakaneko:
Addresses the issue that the Lucene query string parsing doesn't fit our requirements for handling terms that should be excluded. Parses and extracts quoted and excluded terms so we can build a better query.

`a b -c -d` should match Beatmapsets with Beatmaps that contain `a` or `b`, with greater weight to those containing both, and exclude sets containing beatmaps that contain `c` or `d`

fixes #12084
fixes #10916

- [ ] apply parsing to other filters
- [ ] tests

Excluding terms from only tags (or other filters) currently doesn't work completely due to parsing issue #12135, so excluding results still has to be done like `tag="meta/featured artist" -style`, the issue being it will exclude *everything* that has `style` in the text fields used for search (which is a reason we didn't explictly support this directly and it was just a side effect of `simple_query_string`'s parsing.

### DIFF
--- a/app/Libraries/Search/BeatmapsetSearch.php
+++ b/app/Libraries/Search/BeatmapsetSearch.php
@@ -55,28 +55,38 @@ class BeatmapsetSearch extends RecordSearch
 
         if (present($this->params->queryString)) {
             // the subscoping is not necessary but prevents unintentional accidents when combining other matchers
-            $query->must(
-                (new BoolQuery())
-                    // results must contain at least one of the terms and boosted by containing all of them,
-                    // or match the id of the beatmapset.
-                    ->shouldMatch(1)
+            $bool = (new BoolQuery())
+                // results must contain at least one of the terms and boosted by containing all of them,
+                // or match the id of the beatmapset.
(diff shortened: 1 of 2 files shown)


nanaya:
considering `"hello -world"` and `"hello +-world"` don't make difference (as far as I can tell), I wonder if a dumb replacement of `(^| )-` into ` +-` before passed to es would work

notbakaneko:
#12148 instead

nanaya:
the other one doesn't change the general keyword exclusion though, does it? 🤔 (and I think the exclusion is still kinda not quite work as expected atm)

PR #12142, opened 2025-04-24
Fix type of beatmapset id column on downloads table

nanaya:
Exploded on dev/staging...


notbakaneko (approved):

nanaya:
@peppy may want to check if this needs manual updating. It took quite a bit locally here.

PR #12146, opened 2025-04-24
Add team chat notification

### LINKED ISSUE #12000: Add the option to enable/disable notification for Team channel on settings
Just like private messages, getting mapping related notifications and so on, it would be nice to add that kind of feature, sometimes it's annoying to check the team channel without any notification...

nanaya:
I think this works...?

Resolves #12000.


resources/lang/en/notifications.php
```diff
@@ -122,6 +122,12 @@
                     'channel_message_compact' => ':title',
                     'channel_message_group' => 'from :username',
                 ],
+
+                'team' => [
+                    'channel_team' => ':username says ":title" on team chat',
+                    'channel_team_compact' => ':username says ":title"',
+                    'channel_team_group' => 'team chat',
+                ],
```
notbakaneko, line 130: In hindsight, maybe the notification grouping for channel messages should have structured differently. 🤔 
It makes more sense for team messages to have the title be `New team message` and the group display be the most recent message instead of `team chat`?

notbakaneko (changes requested):

PR #12148, opened 2025-04-25
Support exclusion filters in beatmap search query

### LINKED ISSUE #10916: Beatmap search exclusion only looks at tags

### LINKED ISSUE #12084: Excluding map tags from search doesn't work

notbakaneko:
instead of #12134

fixes #12084
fixes #10916

Support exclusion filters. (`-tag="blah blah"`)
Tags only exclude beatmapsets if all the beatmaps in the beatmapset contain the matching tag(s)

Keyword exclusions are full word only; `bric` will not exclude `brick`. Keyword exclusions also only excludes the more visible parts of the search results: title, artist, creator; it doesn't exclude fields that are likely to result in a high chance of unintended filtering.

Removes blindly feeding the keywords portion of the search query into `simple_query_string` and use proper match query that matches our query syntax, instead of trying to contort it into `simple_query_string`.

Moves the filters into a separate class.
Fix missing type hints and `strict_types`

- [x] rebased off #12345
- [ ] #12364
- [x] migrate relevant exclusion stuff from #12134
Chose not to include the keyword parsing parts, as a query like `cat -free` might have unintended consequences of excluding all maps that have `free` any where in the searchable fields. Exclusions should explicitly use the `-tag=free -title=free` syntax to filter them out from the fields where they are not wanted.

Only tags support multiple include/exclusion at the moment; will look at updating the others later


nanaya:
Also mentioned in #12134 but I think #12084 also includes excluding map tags which this PR doesn't handle?

notbakaneko:
I recall it being mention mapper tags are usually used for titles, genres, etc; basically duplicating the existing fields but with extra variations on the names, so maybe exclude options should duplicate into the mapper tags as well 🤔

nanaya:
general exclusion query currently doesn't work quite as expected anyway... which is supposed to be fixed by the closed pr (or my workaround pr).

app/Libraries/Search/BeatmapsetSearch.php
```diff
@@ -129,140 +158,160 @@ public function records()
...
     {
-        if ($this->params->featuredArtist !== null) {
-            $trackIds = ArtistTrack::where('artist_id', $this->params->featuredArtist)->pluck('id');
-            $query->filter(['terms' => ['track_id' => $trackIds]]);
+        if ($this->includes->featuredArtist !== null) {
+            $trackIds = ArtistTrack::where('artist_id', $this->includes->featuredArtist)->pluck('id');
+            $this->query->filter(['terms' => ['track_id' => $trackIds]]);
+        }
+
+        if ($this->excludes->featuredArtist !== null) {
```
nanaya, line 218: I'm not sure where it went wrong but this doesn't seem to work quite right (168 is a-one):





include filter works as expected

nanaya (changes requested):
still checking

notbakaneko:
Updating this with tests, etc on a different branch so this PR doesn't fill with random commits

notbakaneko:
(more) rebasing and stuff

PR #12155, opened 2025-05-07
Use bpm data from beatmap when searching

### LINKED ISSUE #12152: Search results show filtered out BPM
### Type

Game behaviour

### Bug description

When filtering out specific BPM from the search result (for example, in this case BPM that is bigger than 100), the results occasionally shows maps with the filtered out BPM, the rest usually matches my search query.

### Screenshots or videos

https://github.com/user-attachments/assets/eaac6748-3d17-42cb-89b3-898df3f66b1d

### Version

2025-424.0-lazer

### Logs

[compressed-logs.zip](https://github.com/user-attachments/files/19977969/compressed-logs.zip)

nanaya:
Resolves #12152

Doesn't fix the problem of there's no way tell exactly the matching beatmaps but that's a different, more general issue.


notbakaneko (approved):
will need a reindex

PR #12156, opened 2025-05-08
Add team search

### LINKED ISSUE #12090: Add a search bar for the team page
I've seen several people complain about the lack of a search bar to find a team to join based on similar interests (specific anime/game etc), I think it'd be a pretty good QoL feature overall.

nanaya:
🤷 

Resolves #12090.

- [x] preferably after #12158


resources/views/home/_search_result_team.blade.php
```diff
@@ -0,0 +1,71 @@
+{{--
```
notbakaneko, line 1: actually, why isn't this React and `leader` added to the transformer include?

nanaya: 🤷 it's just simpler/faster for initial work. maybe later when it's needed somewhere else

app/Models/Traits/Es/TeamSearch.php
```diff
@@ -0,0 +1,34 @@
...
+    {
+        return static::query();
+    }
+
+    public function toEsJson()
+    {
+        $document = [];
+        foreach (static::esMappings() as $field => $mapping) {
+            $document[$field] = match ($field) {
+                'description' => BBCodeFromDB::removeBBCodeTags(new BBCodeForDB($this->description ?? '')->generate()),
```
notbakaneko, line 27: need to decode html entities as well, or apply the `html_strip` char_filter to the field. Also maybe `removeBlockQuotes` as well unless we still want the content in the quotes to be searchable, then maybe stripping the html tags that get added

notbakaneko (changes requested):

app/Models/Traits/Es/TeamSearch.php
nanaya: the quotes stay because there's no parent

notbakaneko (approved):

PR #12165, opened 2025-05-12
Don't display contest entries when judging period is over but results not available yet

notbakaneko:
I think that should cover that specific state 🤔 

more comprehensive fix when contests gets refactored 🤨

### DIFF
--- a/resources/views/contests/_voting-entrylist.blade.php
+++ b/resources/views/contests/_voting-entrylist.blade.php
@@ -2,22 +2,24 @@
     Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
     See the LICENCE file in the repository root for full licence text.
 --}}
-@if ($contest->hasThumbnails())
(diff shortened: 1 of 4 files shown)


resources/views/contests/_voting-entrylist.blade.php
```diff
@@ -2,22 +2,24 @@
...
 --}}
-@if ($contest->hasThumbnails())
-    <div class="js-react--contestArtList" data-src="contest-{{$contest->id}}"></div>
-@else
-    <div class="js-react--contestList" data-src="contest-{{$contest->id}}"></div>
-@endif
-<script id="contest-{{$contest->id}}" type="application/json">
-    {!! $contest->defaultJson(Auth::user()) !!}
-</script>
+@if (!$contest->isJudged() && $contest->isVotingEnded() && !$contest->show_votes))
```
nanaya, line 5: I think this should set `noVoteReason` in the `voting` template or controller instead.

nanaya (changes requested):

nanaya (approved):

PR #12166, opened 2025-05-12
Simplify user channel notification mail key

nanaya:
I can't tell why they're all in separate keys.

### DIFF
--- a/app/Jobs/Notifications/ChannelAnnouncement.php
+++ b/app/Jobs/Notifications/ChannelAnnouncement.php
@@ -7,16 +7,9 @@ declare(strict_types=1);
 
 namespace App\Jobs\Notifications;
 
-use App\Models\Notification;
-
 class ChannelAnnouncement extends ChannelMessageBase
 {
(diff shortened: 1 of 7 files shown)


app/Jobs/Notifications/ChannelAnnouncement.php
```diff
@@ -7,17 +7,10 @@
 
 namespace App\Jobs\Notifications;
 
-use App\Models\Notification;
-
 class ChannelAnnouncement extends ChannelMessageBase
 {
     const DELIVERY_MODE_DEFAULTS = ['mail' => true, 'push' => true];
 
-    public static function getBaseKey(Notification $notification): string
```
notbakaneko, line 16: The mail key groups are supposed to match the push notification keys, otherwise the function should be renamed

notbakaneko (changes requested):

app/Jobs/Notifications/ChannelAnnouncement.php
nanaya: That plan definitely has failed a long time ago.

PR #12170, opened 2025-05-13
Move Beatmapset SearchResponse interface to own file

### LINKED ISSUE #12169: Displayed default sort order is incorrect when searching beatmaps using query string options
Due to additional querystring parsing server-side that's not client side, the displayed default sort order is not correct (result order is still correct), i.e. shows `Relevance` when it's actually `Ranked`.

notbakaneko:
Separated from trying to fix #12169 since it turned out to be less simple that I thought....

### DIFF
--- a/resources/js/interfaces/beatmapset-search-response.ts
+++ b/resources/js/interfaces/beatmapset-search-response.ts
@@ -1,0 +1,19 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import BeatmapsetExtendedJson from 'interfaces/beatmapset-extended-json';
+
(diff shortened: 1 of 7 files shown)


resources/js/interfaces/beatmapset-search-response.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import BeatmapsetExtendedJson from 'interfaces/beatmapset-extended-json';
+
+type SortField = 'artist' | 'difficulty' | 'favourites' | 'plays' | 'ranked' | 'rating' | 'relevance' | 'title';
```
nanaya, line 6: combine with `sortNames` in https://github.com/ppy/osu-web/blob/ae31a218b68fbb2657de45999227348979553bad/resources/js/beatmaps/search-sort.tsx#L14 ?

nanaya (changes requested):

nanaya (approved):

PR #12172, opened 2025-05-14
Simplify score transformer format parameter

nanaya:
Maybe? 🤔

### DIFF
--- a/app/Http/Controllers/Multiplayer/Rooms/Playlist/ScoresController.php
+++ b/app/Http/Controllers/Multiplayer/Rooms/Playlist/ScoresController.php
@@ -64,7 +64,7 @@ class ScoresController extends BaseController
             ->limit($limit)
             ->getWithHasMore();
 
-        $transformer = ScoreTransformer::newSolo();
+        $transformer = new ScoreTransformer(false);
         $scoresJson = json_collection(
             $highScores->pluck('scoreLink'),
             $transformer,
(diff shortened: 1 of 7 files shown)


app/Http/Controllers/Multiplayer/Rooms/Playlist/ScoresController.php
```diff
@@ -64,7 +64,7 @@ public function index($roomId, $playlistId)
             ->limit($limit)
             ->getWithHasMore();
 
-        $transformer = ScoreTransformer::newSolo();
+        $transformer = new ScoreTransformer(false);
```
nanaya, line 67: this probably doesn't need explicit specification if we just add multiplayer score link support for legacy format 🤔

app/Transformers/ScoreTransformer.php
```diff
@@ -61,38 +58,19 @@ class ScoreTransformer extends TransformerAbstract
         'current_user_attributes',
     ];
 
-    private string $transformFunction;
+    private bool $legacyFormat;
 
-    public static function newSolo(): static
```
notbakaneko, line 66: `PlaylistItemTransformer` still has a reference to this

notbakaneko (changes requested):

app/Transformers/ScoreTransformer.php
nanaya: how did I miss it ಠ_ಠ

notbakaneko (approved):

PR #12173, opened 2025-05-14
Simplify legacy best score attribute getter

nanaya:
Not sure why I didn't do this when doing the `getAttribute` for base score model.


nanaya:
Actually might as well get rid of the class altogether.

PR #12175, opened 2025-05-15
Use new score format for legacy match score

nanaya:
a bunch of things at once mainly due to removal of legacy score typescript interface
- update legacy match score to support new score json format
- use new statistics interface for legacy match leaderboard
- remove legacy score typescript interface
- remove support for transforming legacy score

also
- [x] depends on #12172


app/Http/Controllers/ScoresController.php
```diff
@@ -57,21 +57,20 @@ public function download($rulesetOrSoloId, $id = null)
...
-
-            $soloScore = SoloScore::firstWhere(['legacy_score_id' => $score->getKey(), 'ruleset_id' => $score->ruleset_id]);
+            $soloScore = SoloScore::where([
+                'has_replay' => true,
+                'legacy_score_id' => $id,
+                'ruleset_id' => Beatmap::MODES[$rulesetOrSoloId],
+            ])->firstOrFail();
         }
 
+        $score = $soloScore->legacyScore() ?? $soloScore;
```
notbakaneko, line 72: Just noticed the lock key later on is not including the ruleset for legacy scores

notbakaneko (changes requested):

notbakaneko (approved):

PR #12183, opened 2025-05-21
Rename match to legacy match (controller, react component)

nanaya:
Just making it clearer. I think. Also for some reason the view is called index despite being for show action...


notbakaneko (changes requested):
rename `MatchesControllerTest` as well?

notbakaneko (approved):

PR #12184, opened 2025-05-21
Fix mobx array helper on negative index

nanaya:
I think mobx just considers negative index to be normal prop access.


notbakaneko:
mobx doesn't seem to have a problem `observable([])[-1]` 🤷

nanaya:
huh right 🤔 

also apparently the out of bound thing is no longer relevant since [mobx 6.7.0](https://github.com/mobxjs/mobx/blob/09703d6a3dc7a6214bf2e97cf7e53c690f53d56a/packages/mobx/CHANGELOG.md#670).

Maybe just update to it, replace the negative index usage with lodash `last` or something, and remove the function altogether.

PR #12187, opened 2025-05-23
Truncate long team names in quick search results

notbakaneko:
and usernames as well


resources/css/bem/user-search-card.less
```diff
@@ -71,10 +72,16 @@
 
     &--username {
       .link-plain();
+
+      display: inline-block;
```
nanaya, line 76: it ended up just being `block` because flex container can only contain block layout elements

resources/css/bem/user-search-card.less
```diff
@@ -71,10 +72,16 @@
...
+      white-space: preserve nowrap;
+      text-overflow: ellipsis;
```
nanaya, line 78: this doesn't trigger tooltip to show the whole thing...

resources/css/bem/user-search-card.less
```diff
@@ -71,10 +72,16 @@
 
     &--username {
       .link-plain();
+
+      display: inline-block;
+      white-space: preserve nowrap;
```
nanaya, line 77: apparently just `pre` is enough? (that's also what `.u-ellipsis-pre-overflow` does)

nanaya (changes requested):

resources/css/bem/user-search-card.less
notbakaneko: weird, it didn't truncate before, then again I did more stuff after it 🤔

nanaya (approved):

PR #12192, opened 2025-05-27
Adjust multiplayer json

nanaya:
- move beatmaps and beatmapsets to their own arrays to handle freestyle
- move details for game started event to playlist item as otherwise it'll be pretty difficult to deal with

depends on 
- [ ] #12191


notbakaneko:
Does the response structure change have any effect on your end @bdach ?

bdach:
Basically nobody other than web itself should be using the events endpoint, yeah.

PR #12193, opened 2025-05-27
Convert legacy match frontend to use multiplayer interface

nanaya:
I think this is the minimum set of changes for this specific update...

Relatively straightforward with a bunch of placeholder values.

- [x] depends on #12192


notbakaneko (approved):
probably fine 🤔

PR #12195, opened 2025-05-28
Update match page score styling

nanaya:
Gaps and stuff and align the numbers.

### DIFF
--- a/resources/css/bem/mp-history-game.less
+++ b/resources/css/bem/mp-history-game.less
@@ -107,6 +107,8 @@
   &__player-scores {
     padding: 10px;
     background-color: @osu-colour-b3;
+    display: grid;
+    gap: 5px;
 
     &--no-teams,
     &:last-child {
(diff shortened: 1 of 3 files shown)


nanaya:
...which part?

notbakaneko:
image
It looks like it's just going anywhere

nanaya:
not quite sure which "it" but if you're talking about the score details' bottom row, there are variable number of columns there (3-6)

(also that's the closest thing I can get them to align at all without hardcoding the length - which even then still have to be applied to the entire column and not just the number, otherwise the rightmost column would look wrong)

nanaya:
there will be more changes to the styling but it'll be after the rest of current multiplayer related prs.

assuming it's about the row, the alternative for now is to revert back to minimum size and have no alignment at all or have fixed size for the values which pushes the last value from being right aligned.

notbakaneko:
I think it's better it they were just all right aligned; the difference in spacing between the top and bottom rows from the items filling up the grid columns, and the different text lengths for the same item in different score details, just makes it look like it's floating all over the place.

notbakaneko (approved):

PR #12197, opened 2025-05-29
Add realtime multiplayer event history page

nanaya:
🤷 

Resolves the realtime part of #10455. I thought there's request for similar thing but for playlist but I couldn't find it.

<del>I think the main component can use a bit more code sharing with the legacy one but it'll be a bit involved and slightly annoying to type so maybe later if I feel like it when rebasing and stuff.</del> Done, but the components should probably be moved to `components` (maybe later because I don't want to deal with renames while keeping track of history here)

depends on:
- [x] #12215
- [x] #12193

### DIFF
--- a/resources/js/legacy-match/content.tsx
+++ b/resources/js/legacy-match/content.tsx
@@ -191,6 +191,7 @@ export default class Content extends React.PureComponent<Props> {
     this.autoloadTimeout = setTimeout(this.autoload, refreshTimeout);
   }
 
+  @action
   private readonly loadNext = () => {
     if (!this.hasNext || this.loadingNext) {
       return;
@@ -205,6 +206,7 @@ export default class Content extends React.PureComponent<Props> {
     }));
   };
 
+  @action
   private readonly loadPrevious = () => {
     if (!this.hasPrevious || this.loadingPrevious) {
       return;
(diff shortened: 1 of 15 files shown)


resources/js/legacy-match/content.tsx
```diff
@@ -102,35 +152,71 @@ export default class Content extends React.PureComponent<Props> {
...
       this.inEvent = false;
       return <div className={classWithModifiers('mp-history-content__item', ['event', 'event-close'])} />;
     }
   }
 
+  private delayedAutoload() {
+    this.autoloadTimeout = setTimeout(this.autoload, refreshTimeout);
+  }
+
+  private readonly loadNext = () => {
```
notbakaneko, line 195: needs `@action`

resources/js/legacy-match/content.tsx
```diff
@@ -102,35 +152,71 @@ export default class Content extends React.PureComponent<Props> {
...
+    clearTimeout(this.autoloadTimeout);
+    this.loadingNext = true;
+
+    this.props.loadNext().always(action(() => {
+      this.loadingNext = false;
+      this.delayedAutoload();
+    }));
+  };
+
+  private readonly loadPrevious = () => {
```
notbakaneko, line 210: needs `@action` here too

PR #12204, opened 2025-06-02
Remove incorrect user rank count adjustment logic

nanaya:
It can't be blindly applied due to qualified maps no longer increment the count.

It's also wrong anyway because it doesn't account solo/new scores.

Reference: ppy/osu-infrastructure#32


nanaya:
this is actually kinda higher priority than the other score model cleanups...

notbakaneko (approved):

PR #12207, opened 2025-06-03
Simplify score pin parameters

nanaya:
As I noticed the whole controller still has leftover from dealing with multiple types of score.

Dunno if `order1`/`order3` should be changed to something different like `before`/`after` or `below`/`above` although in those case it needs a bit of thinking whether which is relative to which - whether `below_score_id` means the specified `below_score_id` score should be above or below `scoreId`. The original idea is that the target score (`scoreId`) is always `order2`...

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -110,9 +110,11 @@ Route::group(['middleware' => ['web']], function () {
         Route::get('{rulesetOrScore}/{score?}', 'ScoresController@show')->name('show');
     });
 
-    Route::delete('score-pins', 'ScorePinsController@destroy')->name('score-pins.destroy');
-    Route::put('score-pins', 'ScorePinsController@reorder')->name('score-pins.reorder');
-    Route::resource('score-pins', 'ScorePinsController', ['only' => ['store']]);
+    Route::group(['prefix' => 'score-pins/{score}', 'as' => 'score-pins.'], function () {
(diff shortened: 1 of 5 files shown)


notbakaneko:
Just update it, I don't see `after`/`before` meaning anything other than putting the current score's ordinal position after or before the other score and order{1,2,3} is just confusing for API params

routes/web.php
```diff
@@ -110,9 +110,11 @@
         Route::get('{rulesetOrScore}/{score?}', 'ScoresController@show')->name('show');
     });
 
-    Route::delete('score-pins', 'ScorePinsController@destroy')->name('score-pins.destroy');
-    Route::put('score-pins', 'ScorePinsController@reorder')->name('score-pins.reorder');
-    Route::resource('score-pins', 'ScorePinsController', ['only' => ['store']]);
+    Route::group(['prefix' => 'score-pins/{score}', 'as' => 'score-pins.'], function () {
+        Route::put('reorder', 'ScorePinsController@reorder')->name('reorder');
+        Route::delete('/', 'ScorePinsController@destroy')->name('destroy');
+        Route::post('/', 'ScorePinsController@store')->name('store');
```
notbakaneko: Actually, shouldn't this be `put` (outcome is always the same) and `reorder` be `post`? (outcome using same params depends on current position of scores)

PR #12209, opened 2025-06-03
Add hits per play statistic to profile page

### LINKED ISSUE #2780: Add "hits per play" under user stats
An osu plus feature that I thought was pretty nifty, it adds the following to the stats panel:
Total hits / Total plays = Hits per play

This stat lets us know a bit more about the user's play style. Eg. Emilia is a tv size farmer with 227 hits per play, Rohulk is an endurance player at 615 hits per play.
In theory, we could manually divide total play count by total hits to find this number, but no one would actually do that frequently.

Example
screen shot 2018-03-23 at 7 38 51 pm

tsunyoku:
Resolves #2780.

Opted for flooring as it looks better alongside other statistics in my opinion, but displaying up to 2 decimals would probably also work. Happy with either.

image

### DIFF
--- a/resources/js/profile-page/stats.tsx
+++ b/resources/js/profile-page/stats.tsx
@@ -14,6 +14,7 @@ const entryKeys = [
   'play_count',
   'total_score',
   'total_hits',
+  'hits_per_play',
   'maximum_combo',
   'replays_watched_by_others',
 ] as const;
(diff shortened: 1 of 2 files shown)


nanaya (changes requested):
I wonder if this should just be calculated by the frontend

tsunyoku:
That was my first thought, but with `entryKeys` existing the way it did I didn't see a good way to include it whilst still calculating on the frontend

nanaya:
Handling it separately works:

```
    const val = key === 'hits_per_play'
      ? getHitsPerPlay(this.props.stats)
      : this.props.stats[key];
```

PR #12217, opened 2025-06-05
Update Contest Admin page styling

notbakaneko:
separated from #12014, going to add extra buttons to calculate the scoring and maybe extra messages.

- Use flexbox/grid instead
- stop using the `contest` class
- rename variable with confusing `s` position

### DIFF
--- a/resources/css/bem/admin-contest.less
+++ b/resources/css/bem/admin-contest.less
@@ -2,9 +2,37 @@
 // See the LICENCE file in the repository root for full licence text.
 
 .admin-contest {
-  &__meta-row {
-    margin-bottom: 5px;
-    color: @osu-colour-f1;
+  display: flex;
+  flex-direction: column;
+  gap: 15px;
+
+  &__description {
+    background-color: hsl(var(--hsl-b4));
+    font-size: @font-size--title-small;
+    padding: 10px;
+
+    @media @narrow-up {
+      padding: 20px;
+    }
+  }
+
+  &__meta-title {
+    color: hsl(var(--hsl-f1));
+    font-weight: bold;
+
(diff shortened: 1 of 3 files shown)


resources/css/bem/admin-contest.less
```diff
@@ -2,9 +2,38 @@
...
+    }
+  }
+
+  &__meta {
+    display: grid;
+    grid-template-columns: auto;
+    gap: 5px 10px;
+
+    @media @narrow-up {
+      grid-template-columns: 160px auto;
```
nanaya: why specifically 160px and not `auto`?

resources/css/bem/admin-contest.less
```diff
@@ -2,9 +2,38 @@
...
+    font-size: @font-size--title-small;
+    padding: 10px;
+
+    @media @narrow-up {
+      padding: 20px;
+    }
+  }
+
+  &__meta-title {
+    .u-ellipsis-overflow();
```
nanaya: shouldn't be needed? is this related to the 160px? which label is too long? (and even then it should have tooltip at least?)

nanaya (changes requested):
may also want to get rid of `osu-table--with-handle` in the entry list table as it adds vertical scroll in mobile

it will end up looking like off aligned but if anything I think the background should cover the whole row anyway... (or alternatively just remove the right margin)

resources/css/bem/admin-contest.less
notbakaneko: just using the old value 🤷

notbakaneko:
I'm not bothering with the user-entry-list yet, it's not related to anything I'm touching atm

PR #12223, opened 2025-06-10
Update user rank count on score removal

nanaya:
Only for when going from scoreable (except qualified) to pending (or lower).

I think it works...? Attempted to write test for this but it's taking a bit too long ( ﾟ ヮﾟ)a will retry later (probably on different pr) with reduced scope


notbakaneko (approved):
Seems fine, I think 🤔

PR #12234, opened 2025-06-16
Improve support for multiple update streams on changelog entries

nanaya:
Changelog entry and update stream are supposed to be connected through a different table, `repository_update_stream` but never implemented. The `stream_id` in `repositories` is only to specify the default stream when there's no stream name in the tag (`1111-streamname`).

- `repository_update_stream` needs to be filled in with existing repository/stream and also tachyon
- existing changelog entries (that's linked to tachyon related repositories) need to be linked with a tachyon build entry otherwise next tachyon release build will link all of them
  - the build will need `private` flag enabled (new column) as it'll have like thousands of changelog entries
  - or maybe not, just set the tachyon stream's parent to lazer (new column `streams.parent_id`)

### DIFF
--- a/app/Models/Build.php
+++ b/app/Models/Build.php
@@ -82,8 +82,7 @@ class Build extends Model implements Commentable
         $changelogEntry = new ChangelogEntry();
 
         $newChangelogEntryIds = $stream
-            ->changelogEntries()
-            ->orphans($stream->getKey())
+            ->orphanChangelogEntries()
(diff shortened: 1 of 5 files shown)


notbakaneko (approved):

notbakaneko:
btw does this need anything else before deploy? @peppy @nanaya

peppy:
I'll adjust things post reply, should be fine.

PR #12240, opened 2025-06-18
Simplify handling of missing switch cases

nanaya:
Seems better? Also added item name as part of the message.


nanaya:
why wouldn't it throw on unexpected data?

notbakaneko:
I'm saying in every case it might have to handle unexpected data like in `parse-event` it doesn't throw (and shouldn't); in the cases where it does throw for the type checking, they're all cases where the compile error should be handled, so the message isn't really useful.

nanaya:
but it's for when the typing is wrong...

nanaya:
I guess the message can just be made generic

PR #12244, opened 2025-06-19
Better daily challenge day selector

### LINKED ISSUE #12028: [suggestion] Replace Daily Challenge dropdown on site with calendar
Image

this endless dropdown would make significantly more sense as a calendar view :) even just using built-in browser's date picker could work pretty well

nanaya:
Resolves #12028.


notbakaneko:
I feel like a shortcut that goes to today or latest would be useful...

app/Http/Controllers/Ranking/DailyChallengeController.php
```diff
@@ -8,63 +8,76 @@
...
+        $room = Room::dailyChallengeFor($date);
+        if ($room === null) {
+            $playlist = null;
+            $scores = null;
+            $userScore = null;
+        } else {
+            $playlist = $room->currentPlaylistItem;
 
-        $scores = $room->topScores()->paginate();
+            $scores = $room->topScores()->paginate();
```
notbakaneko, line 44: not this directly, but have we been doing the order without an index this whole time? 🤔

nanaya: seems so

resources/css/bem/daily-challenge-date.less
```diff
@@ -0,0 +1,41 @@
...
+.daily-challenge-date {
+  padding: 10px;
+  display: grid;
+  gap: 10px;
+  border-radius: @border-radius-large;
+  background-color: hsl(var(--hsl-d4));
+
+  &__header {
+    display: grid;
+    grid-template-columns: auto auto auto 150px auto auto;
```
notbakaneko: should be `minmax(100px, 150px)` or something to scale down to fit on narrower screens; 100px seems to fit on iphone SE screens
image

notbakaneko (changes requested):

PR #12246, opened 2025-06-20
Only show team colour shape on team matches

nanaya:
Seems weird showing all blue flags when it's not a team match.


resources/js/legacy-match/score.tsx
```diff
@@ -47,7 +48,7 @@ export default observer(function Score(props: Props) {
   const team = props.playlistItem.details.teams?.[props.score.user_id] ?? 'none';
```
notbakaneko: probably don't coalesce to `none` or try loading if the shape is being removed?

notbakaneko (approved):

PR #12247, opened 2025-06-20
Attempt to render game detail with no start event

nanaya:
Maybe? ┐(･_･)┌

Not quite sure how it happened but it'll be either this or retroactively adding the past event which probably isn't even possible due to id-based sorting

### DIFF
--- a/app/Transformers/Multiplayer/PlaylistItemTransformer.php
+++ b/app/Transformers/Multiplayer/PlaylistItemTransformer.php
@@ -6,6 +6,7 @@
 namespace App\Transformers\Multiplayer;
 
 use App\Models\Multiplayer\PlaylistItem;
+use App\Models\Multiplayer\Room;
 use App\Transformers\BeatmapCompactTransformer;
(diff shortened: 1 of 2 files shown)


notbakaneko (approved):

app/Transformers/Multiplayer/PlaylistItemTransformer.php
```diff
@@ -47,7 +48,7 @@ public function includeBeatmap(PlaylistItem $item)
     public function includeDetails(PlaylistItem $item)
     {
         return $this->primitive($item->detailEvent->event_detail ?? [
-            'room_type' => $item->room->type,
+            'room_type' => Room::REALTIME_DEFAULT_TYPE,
```
notbakaneko, line 51: wait, what's this change for?

nanaya: otherwise it may be shown as team match which wouldn't make sense without the team member data.

PR #12253, opened 2025-06-23
Use correct timestamp for game start time

### LINKED ISSUE #12252: Lazer realtime match link map duration is wrong and sometimes overlap other maps
https://osu.ppy.sh/multiplayer/rooms/1438149/events
Image

nanaya:
Playlist may be created way before it's actually played.

Resolves #12252


bdach (approved):
Yeah the creation date of the playlist item isn't reliable at all. Most notably in playlists, where they'll all invariably get created along with the room.

Issue #11030, opened 2024-02-24
Text on a leaderboard starts breaking when too many mods are used


aert5:
Here is an example beatmap - https://osu.ppy.sh/beatmapsets/325158#osu/741477
![Screenshot from 2024-02-24 20-24-30](https://github.com/ppy/osu-web/assets/157501646/e3e542cc-877c-4efa-87a2-e0ba104456e9)
Maybe limit the number of mods shown and then show the rest of them in a tooltip?

arflyte:
Maigad.

I think we had a design for stacking when there's too many.

bdach:
Previously: https://github.com/ppy/osu/issues/7923, https://github.com/ppy/osu/issues/13952.

nanaya:
I'll make it show up to 4 and if it's more than that, show first 3 followed by "+x", and the mods area (or just the "+x") can be hovered to show a popup listing the rest of them

nanaya:
Fixed in #12755

PR #12255, opened 2025-06-23
Prevent overflowing mods display

### LINKED ISSUE #12251: Lazer realtime match link sometimes doesn't display usernames on map leaderboard
Also seems to be breaking css when this happens.
https://osu.ppy.sh/multiplayer/rooms/1375826/events

Image

nanaya:
Resolves #12251. Or more like "resolves" until a proper mod display is implemented (see #11030).


notbakaneko (approved):
I guess that works...

PR #12266, opened 2025-06-26
Add API endpoint for getting user's passed beatmaps on beatmapsets

notbakaneko:
Edit: It returns the list of beatmaps in the beatmapset that were passed now, because paging/cursoring through multiple possible scores per completed beatmap for the highest one was a bit nebulous...

For supporting vote checking and scoring in Locus https://discord.com/channels/90072389919997952/1141756208638861333/1387361319056838699

`api/v2/users/{user_id}/beatmaps-passed?beatmapset_ids[]={beatmapset_id1}&beatmapset_ids[]={beatmapset_id2}`

Returns a list of Beatmaps with passed plays in the Beatmapset; I don't know about how the criteria is handled if there are scores on multiple diffs on the same Beatmapset, so it just returns for all Beatmap diff (also beatmapset_id isn't in the search index, anyway). Scores with difficulty reduction mods are automatically excluded.

### DIFF
--- a/app/Http/Controllers/Users/BeatmapsController.php
+++ b/app/Http/Controllers/Users/BeatmapsController.php
@@ -15,7 +15,10 @@ use App\Models\Beatmap;
 use App\Models\User;
 use App\Transformers\BeatmapCompactTransformer;
 
-class ScoresController extends Controller
+/**
+ * @group Users
+ */
+class BeatmapsController extends Controller
 {
     private const LIMIT = 50;
 
@@ -24,6 +27,28 @@ class ScoresController extends Controller
         $this->middleware('require-scopes:public');
     }
 
+    /**
+     * Search Beatmaps Passed
+     *
+     * Searches for the Beatmaps a User has passed by Beatmapset.
+     *
+     * ---
+     *
+     * ### Response format
+     *
+     * Returns the list of [Beatmaps](#beatmap) completed matching the criteria.
+     *
+     * Field           | Type
+     * --------------- | ----
+     * beatmaps_passed | [Beatmap](#beatmap)[]
+     *
+     * @urlParam user integer required The id of the user.
+     * @queryParam beatmapset_ids integer[] The list of Beatmapset. Example: [1,2]
+     * @queryParam exclude_converts bool Whether or not to exclude converts. No-example
+     * @queryParam is_legacy bool Whether or not to consider legacy scores. Leave empty for all scores. No-example
+     * @queryParam no_diff_reduction bool Whether or not to exclude diff reduction mods. Defaults to true. No-example
+     * @queryParam ruleset_id int The [Ruleset](#ruleset) ID. Leave empty for all rulesets. No-example
+     */
     public function beatmapsPassed($userId)
     {
         $user = User::findOrFail($userId);
@@ -40,7 +65,7 @@ class ScoresController extends Controller
         $count = count($beatmapsetIds);
 
         if ($count === 0) {
-            return ['completed_beatmaps' => []];
+            return ['beatmaps_passed' => []];
         }
         if ($count > self::LIMIT) {
             throw new RequestTooLargeException('beatmapset_ids', self::LIMIT);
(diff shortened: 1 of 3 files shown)


app/Libraries/Search/BeatmapsPassedSearch.php
```diff
@@ -0,0 +1,76 @@
...
+    {
+        if ($this->completedBeatmapIds === null) {
+            $response = $this->search->response();
+            $this->search->assertNoError();
+
+            $this->completedBeatmapIds = array_map(
+                fn (array $hit): int => (int) $hit['key'],
+                $response->aggregations(self::AGG_NAME)['buckets'],
+            );
+        }
```
nanaya: might as well assign in constructor...? this seems like just an extra part of the one in there

app/Http/Controllers/Users/ScoresController.php
```diff
@@ -0,0 +1,60 @@
...
+        if ($count === 0) {
+            return response()->noContent();
+        }
+        if ($count > self::LIMIT) {
+            throw new RequestTooLargeException('beatmapset_ids', self::LIMIT);
+        }
+
+        RequestCost::setCost($count);
+
+        $beatmaps = Beatmap::whereIn('beatmapset_id', array_slice($params['beatmapset_ids'], 0, 10))->get();
```
nanaya: what's the slice for

app/Http/Controllers/Users/ScoresController.php
```diff
@@ -0,0 +1,60 @@
...
+class ScoresController extends Controller
+{
+    private const LIMIT = 50;
+
+    public function __construct()
+    {
+        $this->middleware('require-scopes:public');
+    }
+
+    public function beatmapsetCompletion($userId)
```
nanaya: The parameter is beatmapset but it returns beatmap, not beatmapset?

app/Http/Controllers/Users/ScoresController.php
```diff
@@ -0,0 +1,60 @@
...
+        $user = User::findOrFail($userId);
+
+        $params = get_params(\Request::all(), null, [
+            'beatmapset_ids:int[]',
+        ], ['null_missing' => true]);
+
+        $count = count($params['beatmapset_ids'] ?? []);
+
+        if ($count === 0) {
+            return response()->noContent();
```
nanaya: shouldn't it be `['completed_beatmaps'=>[]]`? Either that or return error of missing required parameter?

app/Libraries/Search/BeatmapsPassedSearch.php
```diff
@@ -0,0 +1,76 @@
...
+                        'top_hits' => [
+                            'size' => 1,
+                        ],
+                    ],
+                ],
+            ]]);
+        }
+    }
+
+    public function completedBeatmapIds(): array
```
nanaya: the class already has Beatmap as its main subject so maybe just `completedIds`?

app/Libraries/Search/BeatmapsPassedSearch.php
```diff
@@ -0,0 +1,76 @@
...
+
+    private ?array $completedBeatmapIds = null;
+    private ScoreSearch $search;
+
+    public function __construct(
+        int $userId,
+        array $beatmapIds,
+        bool $noDiffReduction = true,
+        ?int $rulesetId = null,
+        ?bool $isLegacy = null
```
nanaya: feels like these should be explicitly specified every time...

app/Http/Controllers/Users/ScoresController.php
```diff
@@ -0,0 +1,60 @@
...
+        $this->middleware('require-scopes:public');
+    }
+
+    public function beatmapsetCompletion($userId)
+    {
+        $user = User::findOrFail($userId);
+
+        $params = get_params(\Request::all(), null, [
+            'beatmapset_ids:int[]',
+        ], ['null_missing' => true]);
```
nanaya: might as well just get_arr directly if just one parameter...?

app/Http/Controllers/Users/ScoresController.php
notbakaneko: It's returning which beatmaps have been played in the beatmapset 🤷 
(and completion is probably not the right name for it)

app/Http/Controllers/Users/ScoresController.php
notbakaneko: wait, I forgot I was going to add options to this

nanaya (approved):

PR #12270, opened 2025-06-30
Remove imagemin

notbakaneko:
I don't think this actually gets used and is an ancient leftover when we used gulp? 🤔


peppy (approved):
YOLO

PR #12273, opened 2025-06-30
Update yargs

notbakaneko:
Update to latest `yargs` to unblock an update to `yargs-parser`


notbakaneko:
needs more package updates to remove the stripAnsi error during linting so updating all of them together in #12274 instead

PR #12274, opened 2025-06-30
Update dependencies

notbakaneko:
Updates a bunch of dependencies that Dependabot wants to update and also update/fix some update blockers

### DIFF
--- a/app/Exceptions/RequestTooLargeException.php
+++ b/app/Exceptions/RequestTooLargeException.php
@@ -1,0 +1,16 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+declare(strict_types=1);
+
+namespace App\Exceptions;
+
+class RequestTooLargeException extends InvariantException
+{
+    public function __construct(string $name, int $limit)
+    {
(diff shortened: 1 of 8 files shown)


resources/css/functions.less
```diff
@@ -118,7 +118,8 @@
 }
 
 .inner-shadow-top() {
-  box-shadow+: inset 0 (@box-shadow-radius - 1px) @box-shadow-radius -1px @box-shadow-color;
+  box-shadow: inset 0 (@box-shadow-radius - 1px) @box-shadow-radius -1px
```
nanaya: that's not right... (missing the `+`)

notbakaneko: whoops deleted it while testing something else 👀

package.json
```diff
@@ -124,6 +124,7 @@
     "webpack-build-notifier": "^2.3.0"
   },
   "resolutions": {
-    "is-plain-object": "^5.0.0"
+    "is-plain-object": "^5.0.0",
+    "strip-ansi": "^6.0.1"
```
notbakaneko: it works on my machine without this (22 LTS) even after clearing everything but it doesn't seem to work on GHA even with node 22 🤷 

`7` onwards changes the export format

notbakaneko: ~~and it works now~~ 🤷 

Apparently the version of yarn I was on is dumb 🤷

PR #12276, opened 2025-07-03
Fix OAuth token delegation test

notbakaneko:
Tests were failing/passing for the wrong reason.

Also move the dataProviders to the current format

### DIFF
--- a/tests/Models/OAuth/TokenTest.php
+++ b/tests/Models/OAuth/TokenTest.php
@@ -33,7 +33,7 @@ class TokenTest extends TestCase
     {
         return Passport::scopes()
             ->pluck('id')
-            ->filter(fn ($id) => !in_array($id, ['chat.write', 'delegate'], true))
+            ->filter(fn ($id) => !in_array($id, ['chat.write', 'chat.write_manage', 'delegate'], true))
             ->map(fn ($id) => [['delegate', $id]])
             ->values();
     }
@@ -43,6 +43,8 @@ class TokenTest extends TestCase
         return [
             'chat.write requires delegation' => [['chat.write'], InvalidScopeException::class],
             'chat.write delegation' => [['chat.write', 'delegate'], null],
+            'chat.write_manage requires delegation' => [['chat.write_manage'], InvalidScopeException::class],
+            'chat.write_manage delegation' => [['chat.write_manage', 'delegate'], null],
         ];
     }
 
@@ -146,7 +148,7 @@ class TokenTest extends TestCase
      */
     public function testDelegationNotAllowedScopes(array $scopes)
     {
-        $user = User::factory()->create();
+        $user = User::factory()->withGroup('bot')->create();
         $client = Client::factory()->create(['user_id' => $user]);
 
         $this->expectException(InvalidScopeException::class);
--- a/tests/Models/OAuth/TokenTest.php
+++ b/tests/Models/OAuth/TokenTest.php
@@ -17,7 +17,7 @@ use Tests\TestCase;
 
 class TokenTest extends TestCase
 {
-    public static function authCodeChatWriteRequiresBotGroupDataProvider()
+    public static function dataProviderForTestAuthCodeChatWriteRequiresBotGroup()
     {
         return [
             [null, InvalidScopeException::class],
@@ -29,7 +29,7 @@ class TokenTest extends TestCase
         ];
     }
 
-    public static function delegationNotAllowedScopesDataProvider()
+    public static function dataProviderForTestDelegationNotAllowedScopes()
     {
         return Passport::scopes()
             ->pluck('id')
@@ -38,7 +38,7 @@ class TokenTest extends TestCase
             ->values();
     }
 
-    public static function delegationRequiredScopesDataProvider()
+    public static function dataProviderForTestDelegationRequiredScopes()
     {
         return [
             'chat.write requires delegation' => [['chat.write'], InvalidScopeException::class],
@@ -46,7 +46,7 @@ class TokenTest extends TestCase
         ];
     }
 
-    public static function delegationRequiresChatBotDataProvider()
+    public static function dataProviderForTestDelegationRequiresChatBot()
     {
         return [
             [null, InvalidScopeException::class],
@@ -58,7 +58,7 @@ class TokenTest extends TestCase
         ];
     }
 
-    public static function scopesDataProvider()
+    public static function dataProviderForTestScopes()
     {
         return [
             'null is not a valid scope' => [null, InvalidScopeException::class],
@@ -67,7 +67,7 @@ class TokenTest extends TestCase
         ];
     }
 
-    public static function scopesClientCredentialsDataProvider()
+    public static function dataProviderForTestScopesClientCredentials()
     {
         return [
             'null is not a valid scope' => [null, InvalidScopeException::class],
@@ -89,7 +89,7 @@ class TokenTest extends TestCase
     }
 
     /**
-     * @dataProvider authCodeChatWriteRequiresBotGroupDataProvider
+     * @dataProvider dataProviderForTestAuthCodeChatWriteRequiresBotGroup
      */
     public function testAuthCodeChatWriteRequiresBotGroup(?string $group, ?string $expectedException)
     {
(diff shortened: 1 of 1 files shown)


tests/Models/OAuth/TokenTest.php
```diff
@@ -83,22 +187,12 @@ public function testClientCredentialResourceOwnerPublic()
...
-        $this->createToken(null, $scopes, $client);
-    }
-
-    /**
-     * @dataProvider delegationRequiredScopesDataProvider
+     * @dataProvider dataProviderForTestDelegationRequiredScopes
      */
     public function testDelegationRequiredScopes(array $scopes, ?string $expectedException)
     {
+        $this->scopesTested($scopes);
```
notbakaneko: not sure if should just not use a dataprovider and do all the scopes in the same run

tests/Models/OAuth/TokenTest.php
```diff
@@ -11,28 +11,143 @@
...
+            $data[] = [$scope, 'bng', InvalidScopeException::class];
+            $data[] = [$scope, 'bot', null];
+            $data[] = [$scope, 'gmt', InvalidScopeException::class];
+            $data[] = [$scope, 'nat', InvalidScopeException::class];
+        }
+
+        return $data;
+    }
+
+    public static function dataProviderForTestDelegationRequiredScopes()
```
notbakaneko: Not sure how I feel about using loops and filter to setup these tests, it strikes me more as "Is it outputting what I put in? Yes."

tests/Models/OAuth/TokenTest.php
```diff
@@ -11,28 +11,143 @@
...
+    public static function dataProviderForTestAuthCodeChatScopesAllowsSelf()
+    {
+        return static::chatScopes()->map(fn ($scope) => [$scope]);
+    }
+
+    public static function dataProviderForTestAuthCodeChatScopesRequiresBotGroup()
+    {
+        $data = [];
+        foreach (static::chatScopes() as $scope) {
+            $data[] = [$scope, null, InvalidScopeException::class];
```
nanaya: is there point in specifying the exact exception for all the data providers when it's just one same class?

tests/Models/OAuth/TokenTest.php
```diff
@@ -11,28 +11,143 @@
...
+        return $data;
+    }
+
+    public static function dataProviderForTestDelegationRequiredScopes()
+    {
+        $scopes = static::allPassportScopeIds();
+
+        $filterScopes = new Set(Token::SCOPES_REQUIRE_DELEGATION);
+        $filterScopes->remove('delegate');
+        $delegationScopes = $scopes->intersect($filterScopes);
```
nanaya:
```suggestion
        $delegationScopes = array_diff(Token::SCOPES_REQUIRE_DELEGATION, ['delegate']);
```

tests/Models/OAuth/TokenTest.php
```diff
@@ -11,28 +11,143 @@
...
+        $noDelegationScopes = $scopes->filter(fn ($id) => !in_array($id, Token::SCOPES_REQUIRE_DELEGATION, true));
```
nanaya: can use `->diff` just like above

...or maybe just loop `static::allPassportScopeIds()` to generate the `$data` instead so it's obvious all of the scopes are tested as opposed to parsing separately looped arrays. It should also remove the need for the weird test on teardown thing (which isn't even obvious it's only for one specific test).

```
foreach (...) {
    if (...) {
        $data[...] = ...;
    } elseif (...) {
        $data[...] = ...;
    } else {
        $data[...] = ...;
    }
}
```

tests/Models/OAuth/TokenTest.php
```diff
@@ -11,28 +11,143 @@
...
+
+        $data = [
+            'delgate scope alone' => [['delegate'], null],
```
nanaya: del`e`gate

nanaya (changes requested):

tests/Models/OAuth/TokenTest.php
```diff
@@ -12,47 +12,139 @@
...
+            $data[] = [$scope, 'gmt', true];
+            $data[] = [$scope, 'nat', true];
+        }
+
+        return $data;
+    }
+
+    public static function dataProviderForTestDelegationNotAllowedScopes()
+    {
+        $delgationExceptDelegate = array_diff(Token::SCOPES_REQUIRE_DELEGATION, ['delegate']);
```
nanaya: all the "delgation"s...

tests/Models/OAuth/TokenTest.php
```diff
@@ -12,47 +12,139 @@
...
+
+        $data = [];
+        foreach (static::allPassportScopeIds() as $scope) {
+            match (true) {
+                $scope === 'delegate' => null,
+                in_array($scope, $delgationExceptDelegate, true) => $data[] = [[$scope], 'delegate_required'],
+                !in_array($scope, Token::SCOPES_REQUIRE_DELEGATION, true) => $data[] = [[$scope], null],
+                default => throw new \Exception("{$scope} was not handled."),
```
nanaya: `match` already does this by default when there's no explicit default?

tests/Models/OAuth/TokenTest.php
```diff
@@ -12,47 +12,139 @@
...
+    public static function dataProviderForTestDelegationNotAllowedScopes()
+    {
+        $delegationExceptDelegate = array_diff(Token::SCOPES_REQUIRE_DELEGATION, ['delegate']);
+
+        $data = [];
+        foreach (static::allPassportScopeIds() as $scope) {
+            match (true) {
+                $scope === 'delegate' => null,
+                in_array($scope, $delegationExceptDelegate, true) => $data[] = [[$scope], 'delegate_required'],
+                !in_array($scope, Token::SCOPES_REQUIRE_DELEGATION, true) => $data[] = [[$scope], null],
```
nanaya: why not

```
$scope === 'delegate' => ..., // (or part of the other one with ternary check)
in_array($scope, Token::SCOPE...) => ...,
default => ...,
```

?

tests/Models/OAuth/TokenTest.php
```diff
@@ -12,47 +12,139 @@
...
+        foreach (static::allPassportScopeIds() as $scope) {
+            match (true) {
+                $scope === 'delegate' => $data[] = [['delegate'], null],
+                in_array($scope, $delegationExceptDelegate, true) => $data[] = [[$scope, 'delegate'], null],
+                !in_array($scope, Token::SCOPES_REQUIRE_DELEGATION, true) => $data[] = [[$scope, 'delegate'], 'delegate_invalid_combination'],
+                default => throw new \Exception("{$scope} was not handled."),
+            };
+        }
+
+        return $data;
```
nanaya: can be `return array_map()`? (and for the other one maybe with additional `array_reject_null`)

tests/TestCase.php
```diff
@@ -50,7 +51,7 @@ public static function regularOAuthScopesDataProvider()
     {
         $data = [];
 
-        foreach (Passport::scopes()->pluck('id') as $scope) {
+        foreach (static::allPassportScopeIds() as $scope) {
             // just skip over any scopes that require special conditions for now.
             if (in_array($scope, ['chat.read', 'chat.write', 'chat.write_manage', 'delegate'], true)) {
```
nanaya: part of the array can be replaced with the new token model const?

tests/TestCase.php
```diff
@@ -87,6 +88,16 @@ protected static function fileList($path, $suffix)
         );
     }
 
+    protected static function chatScopes()
```
nanaya: return type? (mainly noticed because the other new function here, but then also noticed a few other new functions are missing return type as well)

tests/Models/OAuth/TokenTest.php
```diff
@@ -12,47 +12,139 @@
...
+        return [
+            'null is not a valid scope' => [null,'empty'],
+            'empty scope should fail' => [[],'empty'],
+            'all scope is not allowed' => [['*'], 'all_scope_no_client_credentials'],
+        ];
+    }
+
+    /**
+     * @dataProvider dataProviderForTestAuthCodeChatScopesAllowsSelf
+     */
```
nanaya: I wonder if new dataProvider usages should all use the attribute thing instead `#[DataProvider(...)]`

notbakaneko: Might as well since the providers are being moved/renamed, I guess? 🤷

nanaya (approved):

Issue #11622, opened 2024-11-03
Beatmap leaderboards don't show rate change settings clearly


SupDos:
Currently you have to hover over each mod to see the specific settings it has, compared to in game leaderboards showing the specific change for common settings

| web | lazer |
| ----- | ----- | 
|![firefox_2024-11-03_20-29-22](https://github.com/user-attachments/assets/e7fb43ac-39e2-473e-8b9f-6fa68734b296)|![osu!_2024-11-03_20-30-06](https://github.com/user-attachments/assets/c22369fa-e8c2-449f-aef4-57b54ffb257b)|

Couldn't find an issue for this specifically, only for adding the tooltip

kjyuu:
To add to that, this is especially a problem on web for rate changed DT scores (for example x1.01) where people also added CL.
At a glance there is no differentiation between one described and actual stable DTCL score.

cl8n:
it's part of #10931 but I guess I'll leave this open too since it's not obvious that's tracking the same thing.

Issue #10931, opened 2024-01-31
Bring web icons in line with lazer


mikaib:
Not sure if this is the correct place to post this, however, I personally find the mod icons on the site to be quite difficult to see because the icons are very light/thin. Especially when zoomed out its hard to read.

![afbeelding](https://github.com/ppy/osu-web/assets/70071615/8394aa2b-494d-44f2-b99c-7336f8eada06)

I'm very sorry if this is not the correct place to post this.

arflyte:
Noted. Already tested on some alternatives but no good replacement so far. Will keep on trying.

mikaib:
Heya!

I saw the changes in the youtube video published today on the "osu!" youtube channel.
Should I close this issue?

With kind regards,
mikaib

PR #12282, opened 2025-07-08
Update mod icon appearance to match client

bdach:
- Closes https://github.com/ppy/osu-web/issues/10931
- Closes https://github.com/ppy/osu-web/issues/11622

| before | after |
| :-: | :-: |
| Screenshot 2025-07-08 at 12 51 45 | Screenshot 2025-07-08 at 12 58 55 |

Notable changes / additions:
- updated colours to match client (most notably difficulty increase is now red not yellow)
- added the small cog icon for showing whether a mod has been customised
- added the extended display showing most relevant adjustments in case of rate-adjusting mods and difficulty adjust
- added support for font awesome mod icons

Assets taken from [here](https://www.figma.com/design/QuGqnnM2zyhDaz2TyCsON1/New-Mods?node-id=0-1&p=f&t=VGBuInOS674oKmUD-0) with some manual processing. Random and Touch Device don't match what client currently is using but I kinda decided I don't care.

This PR does a significant amount of reinventing how the icons work, in that it entirely uses vector assets. I'm not going to pretend that that looks absolutely amazing on every browser, especially at small sizes, but given the degree to which the redesigned mod icons require control over their individual pieces, I find that the best approach going forward.

To that end, if this approach is accepted, all old rasterised assets in [here](https://github.com/ppy/osu-web/tree/master/public/images/badges/mods) could be deleted, but I'm not doing that before there's consensus that current approach is fine to go forward with.


notbakaneko:
iirc one of the reasons we avoided going crazy with using svg everywhere on web was the lack of rendering quality consistency across browsers. In this case the aliasing looks terrible on Chrome
image

but I also remember an earlier case where some icon looked terrible in Safari. (hopefully the rendering performance got better... ...right..?)

bdach:
I'll concede that in my testing chrome looked the worst, but all three major browsers looked *passable* to me on my machine.

It's all a question of how far you're willing to put the tradeoff point. This has the biggest ease of synchronising with client. I didn't even start with svgs, I wanted to start with rasterized blanks like [the ones client uses](https://github.com/ppy/osu-resources/blob/master/osu.Game.Resources/Textures/Icons/BeatmapDetails/mod-icon.png) but I couldn't figure out how to use them without post-processing (client applies multiplicative fill over these assets; `mix-blend-mode` supports multiplicative blending but does not respect the alpha channel of the asset, which means that it fills a square box around the asset; `mask` doesn't work because it uses *only* the alpha channel so the result has a weird blur from the drop shadow on the asset; `hue-rotate()` probably could work but it requires the blank to *have* an initial hue for it to work, etc.)

You could start with blanks like the ones currently in web source (one per each mod type) but it'd probably result in 36 separate pngs (main icon body, the extender thing, plus the cog indicator for changed setting, times 6 mod types, times 2 sizes per asset), plus probably specific assets in 2 sizes per every mod that currently isn't using the generic acronym variant. And you'd still probably want to be using the generic acronym variant which also rasterizes... not great, IMO, already.

resources/css/bem/mod.less
```diff
@@ -2,89 +2,182 @@
...
-      --generic-display: none;
-    }
+  &__sizing-container {
+    // the entire mod component is expected to scale to a given height.
+    // there are two options for that: either specify every size here against a fixed "reference size" and then scale the entire component in post,
+    // or bake a scale into every single size specification of every element of the component.
+    // this is attempting to implement the first approach via `zoom`
+    // (`scale` cannot be used because it does not affect the layout size of the element, leaving undesirable gaps)
+    // however, if `zoom` is specified on the root element it breaks tooltips, therefore zoom is applied to a child div instead.
+    zoom: calc(var(--mod-height, @mod-height-normal) / @mod-height);
```
nanaya: maybe try using `em` unit instead?

bdach: I considered `em` actually but wasn't sure if that was too hacky to live or not given it's relative to `font-size`.

Will give it a shot, though, because at this point the alternative is not much more appealing.

bdach: Seems to work pretty alright (https://github.com/ppy/osu-web/pull/12282/commits/0884e6110cb10d4a9b6ed30bd92735f086bff23a).

The only snag is that text elements need to be nested in either pseudo- or just straight up nested elements because a `font-size` spec on an element affects `em` on *both* itself and its descendants, so sizing gets real wonky.

nanaya: quite a few places still use `px`... they all need to use `em`

bdach: The ones I've left in pixels I left somewhat intentionally as they are arbitrary adjustments to make stuff look a little less wonky on small sizes. I can try moving those to `em` if you think that's better.

nanaya: at least I don't think those work quite right... (too much bottom padding)

maybe it's fine with some adjustments?

resources/js/components/mod.tsx
```diff
@@ -40,18 +41,84 @@ function settingsLabel(modJson: NonNullable<typeof modNames[string]>, scoreModJs
...
+function getExtendedContent(scoreModJson: ScoreModJson): string | null {
+  const settings = Object.entries(scoreModJson.settings ?? {});
+  switch (scoreModJson.acronym) {
+    case 'HT':
+    case 'DC':
+    case 'DT':
+    case 'NC':
+    {
+      const speedChange = settings.find((s) => s[0] === 'speed_change');
+      return speedChange != null ? `${formatNumber(speedChange[1] as number, 2)}x` : null;
```
nanaya: I guess this was copied from the other function but a simple `const speedChange = scoreModJson.settings?.speed_change` would work.

also the other one has `×`, not `x` 👀

resources/css/bem/mod.less
```diff
@@ -2,89 +2,182 @@
...
-    content: attr(data-acronym);
-    display: var(--generic-display);
+  &__customised-indicator {
+    .fas();
+    .center-content();
+    width: 0.35em;
+    height: 0.35em;
+    border-radius: 50%;
+    background-color: var(--type-fg-colour);
+    z-index: 1;
```
nanaya: maybe put the element last instead... (dealing with zindex isn't fun)

resources/js/components/mod.tsx
```diff
@@ -40,18 +41,84 @@ function settingsLabel(modJson: NonNullable<typeof modNames[string]>, scoreModJs
...
+      for (const [setting, value] of settings) {
+        if (setting in displayCandidates) {
+          if (displayCandidate != null) {
+            return null;
+          }
+
+          displayCandidate = displayCandidates[setting as keyof typeof displayCandidates];
+          displayValue = value as number;
+        }
+      }
```
nanaya: similar to above, and a bit less castings

```suggestion
      let displayCandidate: typeof displayCandidates[keyof typeof displayCandidates] | undefined; // or maybe a better typing...
      let displayValue: number | undefined;
      for (const [key, setting] of Object.entries(displayCandidates)) {
          // is this to not display anything if there are multiple display candidates for the score?
          if (displayCandidate != null) {
            return null;
          }

          const nextDisplayValue = scoreModJson.settings?.[key];
          if (typeof nextDisplayValue === 'number') {
            displayValue = nextDisplayValue;
            displayCandidate = setting;
          }
      }
```

resources/css/bem/mod.less
bdach: See whether you find https://github.com/ppy/osu-web/pull/12282/commits/75aee51f9d0a802a89c03029287f40b4975582f6 any better.

resources/css/bem/mod.less
```diff
@@ -2,89 +2,184 @@
...
+      }
+    }
+
+    .mod-icon-fa(@acronym, @glyph) {
+      display: flex;
+      content: @glyph;
+      font-size: 0.8em;
+      // slight adjustments for small sizes
+      padding-left: 1px;
+      padding-bottom: 1px;
```
nanaya: I don't think it works for small sizes...



compared to no padding-bottom:

resources/js/components/mod.tsx
```diff
@@ -40,18 +41,84 @@ function settingsLabel(modJson: NonNullable<typeof modNames[string]>, scoreModJs
...
+
+      if (displayCandidate != null && displayValue != null) {
+        return `${displayCandidate.acronym}${formatNumber(displayValue, displayCandidate.significantDigits)}`;
```
nanaya: I wonder if this (and the other one) should always use english number formatting...

nanaya (changes requested):

resources/js/components/mod.tsx
bdach: > also the other one has `×`, not `x` 👀

Yeah the client uses plain `x` not the multiply sign. I've changed to use the multiply sign because 🤷

resources/js/components/mod.tsx
bdach: Applied with minor alterations (an actual typing, and also the `displayCandidate != null` guard needs to be after checking that the mod has a non-default setting value, i.e. inside the `typeof` guard).

> `// is this to not display anything if there are multiple display candidates for the score?`

Yeah you got that correct. Compare [client](https://github.com/ppy/osu/blob/b4502d3d90e3263f35b5cc8c9bfc8d91ec5b6d73/osu.Game/Rulesets/Mods/ModDifficultyAdjust.cs#L75-L76).

resources/css/bem/mod.less
```diff
@@ -2,89 +2,182 @@
 // See the LICENCE file in the repository root for full licence text.
 
 .mod {
+  // the entire mod component is expected to scale to a given height.
+  // there are two options for that: either specify every size here against a fixed "reference size" and then scale the entire component in post,
+  // or bake a scale into every single size specification of every element of the component.
+  // this is implementing the second approach via usage of `font-size` and `em`.
+  font-size: var(--mod-height, @mod-height-normal);
   height: var(--mod-height, @mod-height-normal);
```
nanaya: weird, I thought I commented on this 🤔 

```suggestion
  height: 1em;
```

nanaya (changes requested):

Issue #11416, opened 2024-08-07
Add mod leaderboard support for new lazer mods


bdach:
Coming from https://osu.ppy.sh/community/forums/topics/1944960?n=1, but the issue encompasses more mods than just accuracy challenge.

Tenexxt:
_(I pasted this from a reply I made on the forum for the sake of clarity)_
Kinda off topic, but the overall select panel could be improved to be easier to navigate on the web - some mod icons are different from the ones in game, and if all mods will be added to the select it will be kinda crowded, so it could be done in a way that you can pull down the list of mods you want Based on their category (Difficulty increase, Fun etc. Also this could be done for the in-game select itself)

PR #12283, opened 2025-07-09
Allow filtering beatmap leaderboards using lazer mods when lazer mode is active

bdach:
| | lazer mode off | lazer mode on |
| :-: | :-: | :-: |
| osu! | Screenshot 2025-07-09 at 11 34 53 | Screenshot 2025-07-09 at 11 35 34 |
| taiko | Screenshot 2025-07-09 at 11 35 02 | Screenshot 2025-07-09 at 11 35 41 |
| catch | Screenshot 2025-07-09 at 11 35 10 | Screenshot 2025-07-09 at 11 35 48 |
| mania (non-convert) | Screenshot 2025-07-09 at 11 35 24 | Screenshot 2025-07-09 at 11 44 00 |
| mania (convert) | Screenshot 2025-07-09 at 11 35 17 | Screenshot 2025-07-09 at 11 35 57 |

---

Closes https://github.com/ppy/osu-web/issues/11416.

In theory this a self-contained change but I would say there is a soft prerequisite on https://github.com/ppy/osu-web/pull/12282 because things might look a little silly without the redesign:

Screenshot 2025-07-09 at 11 36 32


resources/js/cli/mod-names-generator.js
nanaya, line 1: the `modSorter` needs fixing so `10K` is in correct place

```
return a.Acronym.localeCompare(b.Acronym, 'en', { numeric: true });
```

resources/js/beatmapsets-show/scoreboard/main.tsx
```diff
@@ -6,26 +6,28 @@ import BeatmapExtendedJson from 'interfaces/beatmap-extended-json';
 import { route } from 'laroute';
 import { computed, makeObservable, observable } from 'mobx';
 import { observer } from 'mobx-react';
+import modNames from 'mod-names.json';
+import core from 'osu-core-singleton';
 import * as React from 'react';
 import { classWithModifiers } from 'utils/css';
 import { trans } from 'utils/lang';
 import { SwitchError } from 'utils/switch-never';
+import ModJson from '../../interfaces/mod-json';
```
nanaya: (also will need to be moved)

```suggestion
import ModJson from 'interfaces/mod-json';
```

nanaya (changes requested):

resources/js/beatmapsets-show/scoreboard/main.tsx
bdach: I wish there was a way to tell phpstorm this so I don't have to double check it every time, but it looks like I've got it configured as correctly as it can be in this respect 😖

PR #12287, opened 2025-07-11
Add basic test for mania user stats count adjustment on score delete

nanaya:
The factory part was fun 🤷 

Also >testing the factory

### DIFF
--- a/tests/Jobs/RemoveBeatmapsetSoloScoresDbTest.php
+++ b/tests/Jobs/RemoveBeatmapsetSoloScoresDbTest.php
@@ -1,0 +1,50 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+declare(strict_types=1);
+
+namespace Tests\Jobs;
+
+use App\Jobs\RemoveBeatmapsetSoloScores;
+use App\Models\Beatmap;
+use App\Models\Beatmapset;
+use App\Models\Solo\Score;
+use App\Models\User;
+use PHPUnit\Framework\Attributes\DataProvider;
+use Tests\TestCase;
+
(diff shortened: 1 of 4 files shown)


tests/Models/BeatmapTest.php
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    #[DataProvider('dataProviderForTestFactoryConvertsToManiaKeys')]
+    public function testFactoryConvertsToManiaKeys(int $keys): void
+    {
+        $beatmap = Beatmap::factory()->convertsToManiaKeys($keys)->make();
+        $beatmap->convert = true;
+        $beatmap->playmode = Beatmap::MODES['mania'];
+
+        $this->assertSame($keys, $beatmap->diff_size);
```
nanaya, line 28: actually now I think about it, I'm not even sure if it's testing the `getDiffSize` method or the factory ( ﾟ ヮﾟ)a

tests/Jobs/RemoveBeatmapsetSoloScoresDbTest.php
```diff
@@ -0,0 +1,50 @@
...
+
+    #[DataProvider('dataProviderForTestHandleUserStatsCountConvertedMania')]
+    public function testHandleUserStatsCountConvertedMania(int $keys): void
+    {
+        $beatmapset = Beatmapset::factory()->ranked()->create();
+        $beatmap = Beatmap::factory()->ranked()->convertsToManiaKeys($keys)->create([
+            'beatmapset_id' => $beatmapset,
+        ]);
+        $user = User::factory()->create();
+        $score = Score::factory()->create([
```
notbakaneko: unused variable

tests/Models/BeatmapTest.php
notbakaneko, line 28: needs to test some numbers that aren't just from the factory... or does it? 🤔

notbakaneko (changes requested):

PR #12290, opened 2025-07-14
Show more data on daily challenge leaderboard

nanaya:
This adds two more columns (combo/perfect and rank). There's still more that can be improved like more columns (mainly mods and ruleset-specific stats) and link to the score itself...

The first one requires either changing the table into react or porting the column generation back to php.

The second one I'd rather convert the table into grids instead of doing the link-per-column thing.

### DIFF
--- a/resources/views/rankings/daily_challenge.blade.php
+++ b/resources/views/rankings/daily_challenge.blade.php
@@ -152,6 +152,81 @@
         @include('rankings._beatmapsets', ['beatmapsets' => [$playlist->beatmap->beatmapset], 'modifiers' => 'daily-challenge'])
     @endsection
     @section('scores')
-        @include('multiplayer.rooms._rankings_table')
+        <table class="ranking-page-table">
+            <thead>
+                <tr>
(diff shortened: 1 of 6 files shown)


resources/views/rankings/daily_challenge.blade.php
```diff
@@ -152,6 +152,81 @@ class="{{ class_with_modifiers('daily-challenge-day', [
...
+                    @php
+                        $score = $agg->score;
+                    @endphp
+                    @if ($score === null)
+                        @continue
+                    @endif
+                    <tr class="{{ class_with_modifiers('ranking-page-table__row', ['inactive' => !$score->user->isActive()]) }}">
+                        <td class="ranking-page-table__column">
+                            #{{ i18n_number_format($loop->first
+                                ? $score->userRank()
```
notbakaneko: this gets the `userRank` from `Solo\Score` instead of the playlist?

notbakaneko (changes requested):

PR #12291, opened 2025-07-15
Mark perfect combo on match play leaderboard

nanaya:
Interestingly there's no lint for the `any` type variable declaration? 🤔


notbakaneko:
Well, there's `no-explicit-any` which is currently off because there's some stuff where it's just passing through arguments we don't care the type of but haven't gotten around to fix...

PR #12298, opened 2025-07-17
Restore voted state on contest entry when navigating through history

### LINKED ISSUE #11854: Contest votes get visually reset when going back
When going back from a beatmap info page, the contests page visually reverts all the votes I selected during the session. Updating it by voting again fixes the visual bug.

https://github.com/user-attachments/assets/7dde79aa-fd74-48cb-9ded-7282a94bdddd

notbakaneko:
closes #11854
Also persists the show voted only flag.
Also discovered a bug if show voted only is enabled and removing a vote through the gallery view but that's not being fixed here.

~~Yes, it's saving more state than needed because it's easier than digging for the specific properties~~

### DIFF
--- a/resources/js/contest-voting/base-entry-list.coffee
+++ b/resources/js/contest-voting/base-entry-list.coffee
@@ -49,19 +49,14 @@ export class BaseEntryList extends React.Component
       contest: response.contest
       selected: response.userVotes
       waitingForResponse: false
-      callback
+      () =>
+        @saveState()
+        callback?()
 
 
   componentDidMount: ->
     $.subscribe "contest:vote:click.#{@eventId}", @handleVoteClick
     $.subscribe "contest:vote:done.#{@eventId}", @handleUpdate
-
-
-  componentDidUpdate: (_, prevState) ->
-    if !isEqual(prevState.selected, @state.selected) || prevState.showVotedOnly != @state.showVotedOnly
-      @props.container.dataset.state = JSON.stringify
-        selected: @state.selected
-        showVotedOnly: @state.showVotedOnly
 
 
   componentWillUnmount: ->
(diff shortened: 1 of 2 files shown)


resources/js/contest-voting/base-entry-list.coffee
```diff
@@ -63,4 +73,6 @@ export class BaseEntryList extends React.Component
 
 
   onToggleShowVotedOnlyClick: =>
-    @setState showVotedOnly: !@state.showVotedOnly
+    @setState
+      showVotedOnly: !@state.showVotedOnly
+      () => @props.container.dataset.state = JSON.stringify(@state)
```
nanaya: can't this just be run in `componentDidUpdate`?

nanaya (changes requested):

resources/js/contest-voting/base-entry-list.coffee
```diff
@@ -44,10 +51,19 @@ export class BaseEntryList extends React.Component
...
       callback
 
+
   componentDidMount: ->
     $.subscribe "contest:vote:click.#{@eventId}", @handleVoteClick
     $.subscribe "contest:vote:done.#{@eventId}", @handleUpdate
 
+
+  componentDidUpdate: (_, prevState) ->
+    if !isEqual(prevState.selected, @state.selected) || prevState.showVotedOnly != @state.showVotedOnly
```
nanaya: looking again, the previous one is probably more correct in term of dealing with `waitingForResponse` state which affects the `selected` array 👀 

Or maybe add the state check (which prevents saving incorrect `selected` state on toggling `showVotedOnly` while also waiting for response)

Or don't mix selected state with inflight one (probably requires larger change)... And I just noticed there's nothing handling failed voting request?

nanaya (changes requested):

resources/js/contest-voting/base-entry-list.coffee
notbakaneko: I forgot that's why I put it in the callback in the first place 👀

nanaya (approved):

PR #12300, opened 2025-07-18
Apply phpbb acl check to editing Topic title and posts

notbakaneko:
As discussed internally, you shouldn't be able to update the title of a forum topic if you can't post to it.

Also consolidate some tests and add a few more checks.

### DIFF
--- a/app/Singletons/OsuAuthorize.php
+++ b/app/Singletons/OsuAuthorize.php
@@ -1413,14 +1413,15 @@ class OsuAuthorize
     {
         $prefix = 'forum.post.edit.';
 
-        if ($this->doCheckUser($user, 'ForumModerate', $post->forum)->can()) {
+        $forum = $post->forum;
+        if ($this->doCheckUser($user, 'ForumModerate', $forum)->can()) {
             return 'ok';
         }
 
         $this->ensureLoggedIn($user);
         $this->ensureCleanRecord($user);
 
-        if (!$this->doCheckUser($user, 'ForumView', $post->forum)->can()) {
+        if (!$this->doCheckUser($user, 'ForumView', $forum)->can()) {
             return $prefix.'no_forum_access';
         }
 
@@ -1438,6 +1439,10 @@ class OsuAuthorize
 
         if ($post->post_edit_locked) {
             return $prefix.'locked';
+        }
+
+        if (!ForumAuthorize::aclCheck($user, 'f_post', $forum)) {
+            return 'forum.topic.edit.no_permission';
         }
 
         return 'ok';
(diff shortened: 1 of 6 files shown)


app/Singletons/OsuAuthorize.php
```diff
@@ -1493,14 +1493,24 @@ public function checkForumTopicDelete(?User $user, Topic $topic): string
...
+        if ($this->doCheckUser($user, 'ForumModerate', $topic->forum)->can()) {
+            return 'ok';
+        }
+
+        $postEditPermission = $this->doCheckUser($user, 'ForumPostEdit', $topic->firstPost);
+        if (!$postEditPermission->can()) {
+            return $postEditPermission->rawMessage();
+        }
+
+        if (!ForumAuthorize::aclCheck($user, 'f_post', $topic->forum)) {
```
nanaya: shouldn't this also apply to the post editing auth itself?

app/Singletons/OsuAuthorize.php
```diff
@@ -1493,14 +1498,20 @@ public function checkForumTopicDelete(?User $user, Topic $topic): string
...
     /**
-     * @param User|null $user
-     * @param Topic $topic
-     * @return string
      * @throws AuthorizationCheckException
      */
     public function checkForumTopicEdit(?User $user, Topic $topic): string
     {
-        return $this->checkForumPostEdit($user, $topic->firstPost);
+        if ($this->doCheckUser($user, 'ForumModerate', $topic->forum)->can()) {
```
nanaya: is there specific reason adding this when it's also in `ForumPostEdit`?

database/factories/Forum/ForumFactory.php
```diff
@@ -29,4 +30,23 @@ public function closed(): static
...
+        return $this->state([
+            'moderator_groups' => array_map(fn ($group) => app('groups')->byIdentifier($group)->getKey(), $groups),
+        ]);
+    }
+
+    public function withAuthorize(?string $type): static
+    {
+        return $type === null
+            ? $this
+            : $this->afterCreating(function (Forum $forum) use ($type) {
```
nanaya: the `use` wouldn't have been needed if using `fn ()` syntax

nanaya (changes requested):

app/Singletons/OsuAuthorize.php
notbakaneko: Actually, thinking about it again, the reason I had it separated originally was that posting a new topic uses `f_post` but posting a reply to the topic is `f_reply`, so editing posts other than the first should have been `f_reply`.

Issue #11188, opened 2024-05-01
Recent event doesn't appear when setting a high ranking score in lazer


cl8n:
(this thing) 
![image](https://github.com/ppy/osu-web/assets/9954349/18a695d6-8cab-4da5-ad4e-e84c12a1accb)

opening the issue here, but I'm not sure if this is supposed to be the responsibility of osu-web or the score queue.

nanaya:
The calculation itself is somewhere else but the event row should probably be created by osu-web through interop endpoint? 🤔

cl8n:
makes sense. I won't be implementing this but some more random thoughts on briefly looking into it:

- this is directly related to the "lost first place" event, which I think should be re-implemented at the same time
  - and also that one has some special behavior to be deleted if the user who caused it turns out to be a cheater, no idea how that works though.
- would be cool if the stable and lazer version of these events could be distinguished, and filtered for display based on the viewer's "lazer mode" setting
  - if this is done, then setting a stable score would actually need to create one of each event I think? because the rankings won't be the same

NeofetchNpc:
when is it implemented?

peppy:
> when is it implemented?

I can understand you are excited to see this happen, but please know that we are doing our best to get everything working how everyone wants it as fast as we can. There's no ETA yet but it's on my high priority list.

PR #12303, opened 2025-07-21
Add supporting pieces for displaying lazer leaderboard rank achieved events on user profiles

bdach:
First step toward https://github.com/ppy/osu-web/issues/11188.

PRing early without the rest of the pieces to discuss general direction first.

## Baseline assumptions

- We want to show users achieving top leaderboard positions on user profiles for lazer to match stable.
- We want to respect the user's "lazer mode" setting because leaderboard positions on "lazer" mode and on "classic" mode are different.

## How this proposes to achieve the above

`osu_events` gets a new nullable boolean column called `legacy_score_event` which also gets added to the primary lookup index on the table. The intention is thus:

- Events that have nothing to do with leaderboards have this flag left to `NULL`.
- Events that pertain to non-lazer-mode leaderboards, or in other words, events emitted by `osu-web-10`, should have this flag set to `TRUE`.
- Events that pertain to lazer-mode leaderboards, which `osu-queue-score-statistics` will emit via web, should have this flag set to `FALSE`.

This PR adds a shared interop endpoint that is intended to be used by `osu-queue-score-statistics` to facilitate the last point of this.

## What will need to happen next, if this approach is accepted

1. Database schema migration is deployed.
2. `osu-web-10` is updated so that when it inserts rank events, it inserts them with `legacy_score_event = TRUE`.
3. On production, something along the lines of

    **⚠️⚠️⚠️ WARNING UNTESTED DOUBLE CHECK ⚠️⚠️⚠️**
    ```sql
    UPDATE `osu_events` SET `legacy_score_event` = TRUE WHERE (`text` LIKE "%achieved rank%" OR `text` LIKE "%has lost first place%")
    ```
    is ran to back-migrate past events. This is admittedly a bit dodgy because of relying on text content but not much better can be done quickly because the table doesn't have any explicit enum column for event types.

    In theory you could skip this if you're fine with things looking a little wonky for 90 days with lazer mode on.
5. `osu-queue-score-statistics` is updated to start emitting these events for lazer mode leaderboards using the shared interop endpoint from this PR.

I'm also not doing anything about first place ranks here because that will require extra work (in particular, another db table).

### DIFF
--- a/app/Http/Controllers/Ranking/DailyChallengeController.php
+++ b/app/Http/Controllers/Ranking/DailyChallengeController.php
@@ -41,7 +41,11 @@ class DailyChallengeController extends Controller
         } else {
             $playlist = $room->currentPlaylistItem;
 
-            $scores = $room->topScores()->paginate();
+            $scores = $playlist
+                ->highScores()
+                ->forRanking()
+                ->with(['score.user.team'])
+                ->paginate();
         }
 
         $currentUser = \Auth::user();
@@ -67,7 +71,7 @@ class DailyChallengeController extends Controller
                 ->get()
                 ->keyBy(fn ($agg) => $dateHelper::makeId(parse_db_time($agg->getAttribute('room_starts_at'))));
 
-            $userScore = $room?->topScores()->whereBelongsTo($currentUser)->first();
+            $userScore = $playlist?->highScores()->whereBelongsTo($currentUser)->first();
         }
 
         return ext_view('rankings.daily_challenge', compact(
--- a/app/Models/Multiplayer/PlaylistItemUserHighScore.php
+++ b/app/Models/Multiplayer/PlaylistItemUserHighScore.php
@@ -119,6 +119,15 @@ class PlaylistItemUserHighScore extends Model
         return $this->belongsTo(User::class, 'user_id');
     }
 
+    public function scopeForRanking(Builder $query): Builder
+    {
+        return $query
+            ->passing()
+            ->whereHas('user', fn ($q) => $q->default())
+            ->orderByDesc('total_score')
+            ->orderBy('score_id');
+    }
+
     public function scopePassing(Builder $query): Builder
     {
         return $query->where('total_score', '>', 0);
@@ -146,5 +155,18 @@ class PlaylistItemUserHighScore extends Model
             'score_id' => $score->getKey(),
             'total_score' => $score->total_score,
         ])->save();
+    }
+
+    public function userRank(): ?int
+    {
+        if ($this->total_score === null || $this->score_id === null) {
+            return null;
+        }
+
+        $query = static::where('playlist_item_id', $this->playlist_item_id)
+            ->forRanking()
+            ->cursorSort('score_asc', $this);
+
+        return 1 + $query->count();
     }
 }
--- a/resources/css/bem/ranking-page-table.less
+++ b/resources/css/bem/ranking-page-table.less
@@ -62,11 +62,25 @@
     }
   }
 
+  &__combo {
+    &--perfect {
+      color: hsl(@beatmap-score--hsl-perfect);
+    }
+  }
+
   &__heading {
     font-weight: normal;
     color: hsl(var(--hsl-f1));
-    padding: 5px 10px;
+    padding: 6px 5px;
     text-align: center;
+
+    &:first-child {
+      padding-left: 10px;
+    }
+
+    &:last-child {
+      padding-right: 10px;
+    }
 
     &--focused {
       color: hsl(var(--hsl-c1));
--- a/resources/css/bem/score-rank.less
+++ b/resources/css/bem/score-rank.less
@@ -29,6 +29,11 @@
   .all(~"D", "D");
   .all(~"F", "F");
 
+  &--daily-challenge {
+    display: inline-block;
+    font-size: 16px; // icon size
+  }
+
   &--full {
     .full-size();
   }
--- a/resources/js/contest-voting/base-entry-list.coffee
+++ b/resources/js/contest-voting/base-entry-list.coffee
@@ -1,6 +1,7 @@
 # Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
 # See the LICENCE file in the repository root for full licence text.
 
+import { isEqual } from 'lodash'
 import { button, span } from 'react-dom-factories'
 import { trans } from 'utils/lang'
 import { nextVal } from 'utils/seq'
@@ -10,7 +11,10 @@ export class BaseEntryList extends React.Component
     super props
 
     @eventId = "contests-show-voting-#{nextVal()}"
-    @state =
+    # TODO: combine with count in GalleryContestVoteProgress when converting to typescript
+    state = JSON.parse(@props.container.dataset.state) if @props.container.dataset.state?
+
+    @state = Object.assign
       waitingForResponse: false
       contest: @props.contest
       selected: @props.selected
@@ -18,7 +22,9 @@ export class BaseEntryList extends React.Component
       options:
         showPreview: @props.options.showPreview ? false
         showLink: @props.options.showLink ? false
-        linkIcon: @props.options.linkIcon ? false
+        linkIcon: @props.options.linkIcon ? false,
+      state ? {}
+
 
   handleVoteClick: (_e, {contest_id, entry_id, callback}) =>
     return unless contest_id == @state.contest.id
@@ -35,6 +41,7 @@ export class BaseEntryList extends React.Component
       waitingForResponse: true
       callback
 
+
   handleUpdate: (_e, {response, callback}) =>
     return unless response.contest.id == @state.contest.id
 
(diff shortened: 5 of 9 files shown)


stanriders:
Wouldn't it make more sense to do this _after_ mod multipliers (and therefore leaderboard ordering) are stabilised? Otherwise the the rank achieved information is pretty much irrelevant and would just be a profile spam for every lazer player (since lazer scores are getting higher leaderboard places than stable by default right now)

peppy:
> Wouldn't it make more sense to do this _after_ mod multipliers (and therefore leaderboard ordering) are stabilised? Otherwise the the rank achieved information is pretty much irrelevant and would just be a profile spam for every lazer player (since lazer scores are getting higher leaderboard places than stable by default right now)

The starting point in my eyes is to get data flowing. I do not think it will result in the level of "profile spam" that you imagine, but if it does, I see this as beneficial to reprioritise concerns over adjusting problematic multipliers sooner rather than later.

To be clear: this concern was raised from the outset and I am dismissing (deferring) it for now.

app/Http/Controllers/InterOp/UsersController.php
```diff
@@ -32,6 +34,25 @@ public function achievement($id, $achievementId, $beatmapId = null)
...
+            request()->all(),
+            null,
+            [
+                'position_after:int',
+                'rank:string',
+                'legacy_score_event:bool',
+            ]
+        );
+        $params['beatmap'] = Beatmap::findOrFail($beatmapId);
+        $params['ruleset'] = Ruleset::from($rulesetId);
```
nanaya: this whole enum thing hasn't been working very well... (it causes more annoyance than the typing help it provides)

app/Models/Event.php
```diff
@@ -495,4 +525,31 @@ private static function beatmapsetParams($beatmapset)
...
+        $url = e(route('beatmaps.show', ['beatmap' => $beatmap, 'm' => $ruleset], false));
+        return [
+            'title' => e($beatmap->beatmapset->artist.' - '.$beatmap->beatmapset->title.' ['.$beatmap->version.']'),
+            'url' => $url,
+            'url_clean' => $GLOBALS['cfg']['app']['url'].$url,
+        ];
+    }
+
+
+    public static function modeToString(Ruleset $mode)
```
nanaya: there's already one in translation `beatmaps.mode.{$rulesetName}` (the translation is locked and not translated to other languages)

app/Models/Event.php
```diff
@@ -470,9 +498,11 @@ public function parse()
         return $this;
     }
 
-    public function scopeRecent($query)
+    public function scopeRecent($query, null | true $legacyOnly)
     {
-        return $query->orderBy('event_id', 'desc')->limit(5);
+        return $query->orderBy('event_id', 'desc')
+            ->whereIn('legacy_score_event', [null, $legacyOnly === true])
```
nanaya: `null` value can only be matched using `is [not] null`. this won't match columns with `null` value.

app/Models/Event.php
```diff
@@ -173,6 +176,31 @@ public static function generate($type, $options)
 
                 break;
 
+            case 'rank':
+                $beatmap = $options['beatmap'];
+                $ruleset = $options['ruleset'];
+                $rulesetName = self::modeToString($ruleset);
+                $beatmapParams = static::beatmapParams($beatmap, $ruleset);
```
nanaya: at least decide on either `self::` or `static::`...

app/Models/Event.php
```diff
@@ -495,4 +525,31 @@ private static function beatmapsetParams($beatmapset)
             'url_clean' => $GLOBALS['cfg']['app']['url'].$url,
         ];
     }
+
+    private static function beatmapParams($beatmap, $ruleset)
+    {
+        $url = e(route('beatmaps.show', ['beatmap' => $beatmap, 'm' => $ruleset], false));
+        return [
+            'title' => e($beatmap->beatmapset->artist.' - '.$beatmap->beatmapset->title.' ['.$beatmap->version.']'),
```
nanaya: probably easier to read with `"{$var}"` string interpolation

app/Http/Controllers/InterOp/UsersController.php
```diff
@@ -32,6 +34,25 @@ public function achievement($id, $achievementId, $beatmapId = null)
...
+        $params['user'] = User::findOrFail($userId);
+
+        Event::generate('rank', $params);
```
nanaya, line 56: missing parameter should be handled so it doesn't return throw exception/return 500 when accessed.

app/Models/Event.php
```diff
@@ -173,6 +176,31 @@ public static function generate($type, $options)
...
+                $beatmapParams = static::beatmapParams($beatmap, $ruleset);
+                $user = $options['user'];
+                $userParams = static::userParams($user);
+                $positionAfter = $options['position_after'];
+                $positionText = $positionAfter <= 50 ? "<b>rank #{$positionAfter}</b>" : "#{$positionAfter}";
+                $rank = $options['rank'];
+                $legacyScoreEvent = $options['legacy_score_event'];
+
+                $params = [
+                    'text' => "<img src='/images/{$rank}_small.png'/> <b><a href='{$userParams['url']}'>{$userParams['username']}</a></b> achieved rank {$positionText} on <a href='{$beatmapParams['url']}'>{$beatmapParams['title']}</a> ({$rulesetName})",
```
nanaya: `rank` in `achieved rank` is redundant with `$positionText` in some cases

bdach:
> and what about legacy score events which also the case for lazer mode...? (for example, first place legacy score that's also the first for lazer mode)

The events will have to be duplicated (one for lazer, one for stable). I don't really see any other way around it. The processes are just too distributed to avoid that.

Or I guess you could *maybe avoid* duplication by having `osu-web-10` stop issuing the events and move that to `osu-queue-score-processor` and have it de-duplicate locally (emit an event with NULL mode if it happens to be good for both modes) but I don't know how much that's worth the effort.

Either way that's not something web should be worried about.

app/Http/Controllers/InterOp/UsersController.php
bdach: Even though this is essentially an internal interop endpoint?

nanaya: yeah, giving hint what went wrong would be useful...

app/Http/Controllers/InterOp/UsersController.php
bdach: If I'm to use `beatmaps.mode.{$rulesetName}` instead of my own helper as per https://github.com/ppy/osu-web/pull/12303#discussion_r2221997607 *and* stop using `Ruleset` I'm going to need to either find a replacement for

https://github.com/ppy/osu-web/blob/1e61e7045d4049d686319e63fba82598415dc6c7/app/Enums/Ruleset.php#L38-L43

or just accept the ruleset directly as string or something.

nanaya: there's `Beatmap::modeStr()`

app/Models/Event.php
bdach: I knew that but then must have forgot and decided to not test properly or fooled myself into thinking I tested properly but hadn't really.

app/Http/Controllers/InterOp/UsersController.php
```diff
@@ -32,6 +34,30 @@ public function achievement($id, $achievementId, $beatmapId = null)
...
+        abort_unless(isset($params['rank']), 422, 'missing rank parameter');
+        abort_unless(isset($params['legacy_score_event']), 422, 'missing legacy_score_event parameter');
+
+        $params['beatmap'] = Beatmap::findOrFail($beatmapId);
+        $params['ruleset'] = Beatmap::modeStr($rulesetId);
+        $params['user'] = User::findOrFail($userId);
+
+        Event::generate('rank', $params);
+        // TODO: also emit lost first place rank event when relevant
+    }
```
nanaya, line 59: weird, I guess I didn't actually post the comment on there should be an explicit return value (a 204 response in this case)

app/Models/Event.php
```diff
@@ -495,4 +524,14 @@ private static function beatmapsetParams($beatmapset)
             'url_clean' => $GLOBALS['cfg']['app']['url'].$url,
         ];
     }
+
+    private static function beatmapParams($beatmap, $ruleset)
+    {
+        $url = e(route('beatmaps.show', ['beatmap' => $beatmap, 'm' => $ruleset], false));
```
nanaya: `m` requires `rulesetId`, not name

bdach: ~~please take me back to static typing, stat~~

nanaya:
@peppy may want to take care when deploying this one's migration as it took quite a bit when I tried it locally.

peppy:
Have applied the migration manually

```sql
ALTER TABLE osu_events ADD COLUMN legacy_score_event BOOLEAN NULL DEFAULT NULL, ADD INDEX user_id_legacy_score_event (user_id, event_id, legacy_score_event), DROP INDEX user_id;

insert into migrations (migration, batch) select '2025_07_21_101446_add_legacy_score_event_flag_to_events', max(batch) + 1 FROM migrations;
```

PR #12307, opened 2025-07-22
Force rooms with no end time to be the first

### LINKED ISSUE #12306: Active multiplayer rooms show incorrectly at the bottom of history
`https://osu.ppy.sh/users/<user_id>/realtime`

Image

They should be at the top.

nanaya:
Was reviewing something else and figured to play around with dealing with nulls in db.

I haven't done much testing on this (namely hitting other room listing endpoint) so stuff may break 🤷 (or just slow)

And there may be better way of fixing this...?

Resolves #12306. Drafting because I realized there's also problem when the cursor value itself is null... ಠ_ಠ It maybe better to just set the end time to be the silly date instead after all.


nanaya:
See #12312.

Issue #7441, opened 2021-04-08
Add API route to moderate forum topic


cl8n:
continuation of #6072

"moderate" as in pin, lock, move, delete. should require a new scope, and probably only available to client owner, as "forum.write" is not quite the same level of permission as managing a forum

for reference my use case is pinning/unpinning on Project Loved forum each round and locking voting threads when they are complete

cl8n:
need some opinion for how to handle the permissions here -- precedent so far is that API routes ignore any permissions gained from groups/etc, but in this case I want to explicitly allow it. I'm hesitant to make a new scope like `forum.moderate` because that's a lot of permission to give away when users like GMT or Admin authorize with it...

peppy:
The only way I could see this working is if it was limited to a bot account class. Seems like we would never want to be giving out any kind of moderation abilities via the API in any normal scenario. Also there should be some warning message on bot accounts around the oauth section if we go with that direction.

PR #12308, opened 2025-07-22
Add OAuth scope for pinning and locking topics

notbakaneko:
- `forum.write_manage` scope added; scope is `client_credentials` only, requires `delegate` and requires the bot account to be in the groups listed in the forum's `moderator_groups`.
- `forum.write` can now be `delegate`d to a bot account
- Added `group_permissions` (can use some name suggestion) scope for enabling forum moderation group permissions
- ~~The OAuth scopes documentation is out of date and still needs updating~~ just moved them here instead of a larger doc update
- ~~things that support group permissions with OAuth have been changed to apply receive the group permission if the token supports delegation or is an `ownToken`.~~ The `ownToken` thing should only be only for chat.

related #7441 but this doesn't include moving or deleting

- [x] #12300

I thought of making a separate bot group for this to separate the "chat bots" and bots used for other things, but decided to use the existing bot group for now.

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -1045,10 +1045,13 @@ class User extends Model implements AfterCommit, AuthenticatableContract, HasLoc
         return $this->inGroupWithPlaymode('bng', $mode);
     }
 
+    /**
+     * Check if the User is explicitly assigned Forum permissions via moderator_groups.
+     *
+     * Permission set directly though moderator_groups are available through OAuth with the group_permissions scope.
+     */
     public function isForumModerator(Forum\Forum $forum): bool
     {
-        // permission set directly though moderator_groups should behave like checks
-        // using findUserGroup directly instead of isGroup and are available with OAuth.
         $token = $this->token();
 
         return ($token === null || $token->delegatesOwner() && $token->can('group_permissions'))
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -1042,6 +1042,13 @@ class User extends Model implements AfterCommit, AuthenticatableContract, HasLoc
         return $this->inGroupWithPlaymode('bng', $mode);
     }
 
+    public function isForumModerator(Forum\Forum $forum): bool
+    {
+        // permission set directly though moderator_groups should behave like checks
+        // using findUserGroup directly instead of isGroup and are available with OAuth.
+        return $forum->moderator_groups !== null && !empty(array_intersect($this->groupIds()['active'], $forum->moderator_groups));
+    }
+
     public function isLimitedBN(?string $mode = null)
     {
         return $this->inGroupWithPlaymode('bng_limited', $mode);
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -1051,9 +1051,9 @@ class User extends Model implements AfterCommit, AuthenticatableContract, HasLoc
         // using findUserGroup directly instead of isGroup and are available with OAuth.
         $token = $this->token();
 
-        return $token !== null && !($token->delegatesOwner() && $token->can('forum.delegate')) && !$token->isOwnToken()
-            ? false
-            : $forum->moderator_groups !== null && !empty(array_intersect($this->groupIds()['active'], $forum->moderator_groups));
+        return ($token === null || $token->delegatesOwner() && $token->can('group_permissions'))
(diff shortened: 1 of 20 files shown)


nanaya:
Quite a few forum related auth checks still rely on `User#groupIds()` which doesn't follow token-based user group restrictions...

app/Models/User.php
```diff
@@ -1023,7 +1023,10 @@ public function isAdmin()
 
     public function isChatAnnouncer()
     {
-        return $this->findUserGroup(app('groups')->byIdentifier('announce'), true) !== null;
+        $token = $this->token();
+        return $token !== null && !$token->delegatesOwner() && !$token->isOwnToken()
+            ? false
+            : $this->findUserGroup(app('groups')->byIdentifier('announce'), true) !== null;
```
notbakaneko, line 1030: This change may cause problems if people have been using lazer to send additional messages to announcement channels

nanaya (commented):
I think this still needs a new scope to grant access to user's group?

otherwise access token of moderators may give access to stuff like deleted posts and non public forums even with just `public` scope? (haven't actually tested)

notbakaneko:
I'll get to that at some point 👀

peppy:
This feels like it should be a 1 day task, but here we are a month later...

Hiviexd:
any progress on this?

notbakaneko:
Missed the last notification...

app/Models/User.php
```diff
@@ -1042,6 +1045,17 @@ public function isFullBN(?string $mode = null)
         return $this->inGroupWithPlaymode('bng', $mode);
     }
 
+    public function isForumModerator(Forum\Forum $forum): bool
+    {
+        // permission set directly though moderator_groups should behave like checks
+        // using findUserGroup directly instead of isGroup and are available with OAuth.
+        $token = $this->token();
+
+        return $token !== null && !($token->delegatesOwner() && $token->can('forum.delegate')) && !$token->isOwnToken()
```
nanaya: the chat announce is easier to control (as those involve chat scopes) but auth code for own client giving out forum moderator permission on public scope (reading deleted posts for example) is probably not an expected outcome (and it's currently already problematic in even worse way).

which is why I suggested adding scope which simply enable all user group access instead of adding more specific exceptions on specific places (which potentially gives more permission than expected)

notbakaneko: come to think of it, the whole own token thing was for the whole applying to run a chat bot rules thing :|

tests/Controllers/Forum/TopicsControllerTest.php
```diff
@@ -152,6 +277,55 @@ public function testPin(?string $group, array $forumGroups, bool $success): void
         }
     }
 
+    #[DataProvider('dataProviderForAuthCodeDelegateOnlyGroupTests')]
+    public function testPinAuthCode(array $groups, array $forumGroups): void
```
nanaya: the test doesn't even touch the pin action...

the same with `testLockAuthCode` and `testLockClientCredentialsWithGroupPermission` (and maybe others)

tests/Controllers/Forum/TopicsControllerTest.php
```diff
@@ -7,17 +7,73 @@
...
+            // standalone group
+            [['bot'], [], false, false],
+            [['loved'], [], true, false],
+            [['loved'], ['loved'], true, false],
+            [['loved'], ['gmt'], true, false],
+            [['gmt'], [], true, false],
+            [['gmt'], ['loved'], true, false],
+            [['gmt'], ['gmt'], true, false],
+
+            // with bot group, bot needs to be last for the factory.
```
nanaya: it's missing comment on the why...

app/Models/User.php
```diff
@@ -1042,6 +1045,16 @@ public function isFullBN(?string $mode = null)
         return $this->inGroupWithPlaymode('bng', $mode);
     }
 
+    public function isForumModerator(Forum\Forum $forum): bool
+    {
+        // permission set directly though moderator_groups should behave like checks
+        // using findUserGroup directly instead of isGroup and are available with OAuth.
```
nanaya: the comment can use an update (notably the scope requirement)

PR #12312, opened 2025-07-28
Separate active and ended room listing

### LINKED ISSUE #12306: Active multiplayer rooms show incorrectly at the bottom of history

nanaya:
How about just give the room with no end time its own section instead.

Resolves #12306.

The empty wordings can use adjustment.


notbakaneko (changes requested):
The text for no active rooms can be `Not in currently in any (in-progress?) osu!(lazer) multiplayer games!` and ended could be `Not in any completed osu!(lazer) multiplayer games yet!`

notbakaneko (approved):

Issue #12281, opened 2025-07-08
The tooltip for favorites is missing on the website


oxamera:
https://osu.ppy.sh/community/forums/topics/2102094?n=1

PR #12323, opened 2025-08-01
Add default tooltip for favourites count on `beatmapset` page

diquoks:
- closes https://github.com/ppy/osu-web/issues/12281

| master | pr |
|-|-|
| image | image |

### DIFF
--- a/resources/js/beatmapsets-show/header.tsx
+++ b/resources/js/beatmapsets-show/header.tsx
@@ -231,19 +231,11 @@ export default class Header extends React.Component<Props> {
 
   @action
   private readonly onEnterFavouriteIcon = () => {
-    if (this.filteredFavourites.length < 1) {
-      if (this.favouritePopupDisposer != null) {
-        this.favouritePopupDisposer();
-        $(this.favouriteIconRef.current ?? []).qtip('destroy', true);
-      }
-
-      return;
-    }
-
     this.favouritePopupDisposer ??= createTooltip(
       () => this.favouriteIconRef.current,
       () => ({
         count: this.controller.beatmapset.favourite_count,
+        title: this.filteredFavourites.length < 1 ? trans('beatmapsets.show.stats.no_favourites') : undefined,
         users: this.filteredFavourites,
       }),
       'right center',
--- a/resources/js/components/user-list-popup.tsx
+++ b/resources/js/components/user-list-popup.tsx
@@ -76,7 +76,7 @@ export default function UserListPopup(props: Props) {
           ))}
         </div>
       }
-      {props.count > props.users.length &&
+      {props.count > props.users.length && props.users.length > 0 &&
         <div className='user-list-popup__remainder-count'>
           {transChoice('common.count.plus_others', props.count - props.users.length)}
         </div>
--- a/resources/lang/en/beatmapsets.php
+++ b/resources/lang/en/beatmapsets.php
@@ -224,6 +224,7 @@ return [
             'rating-spread' => 'Rating Spread',
             'nominations' => 'Nominations',
             'playcount' => 'Playcount',
+            'no_favourites' => 'No favourites yet!',
         ],
 
         'status' => [


nanaya (changes requested):
this conflicts with existing tooltip

diquoks:
> this conflicts with existing tooltip

strange, it was okay for me
will take another look later

nanaya:
try adding to favorite, hover on the icon, remove from favorite, and then hover again.

diquoks:
> strange, it was okay for me

I mean this tooltip did not appear if `this.controller.beatmapset.favourite_count > 0`

nanaya:
the correct fix is updating the existing tooltip to also show on empty list

(see [this](https://github.com/ppy/osu-web/blob/e2a95a080194f577bcba59ecfa147567a0a1d1b4/resources/js/beatmap-discussions/discussion-vote-buttons.tsx#L91-L94) for reference)

diquoks:
can't check right now, but should work

nanaya:
no it doesn't (there's a different check on the line above). also the count should be accessed inside the function due to how mobx works.

diquoks:
sry, was overconfident :)
will come back when understand better

diquoks:
it seems to be working correctly now, but tooltip's style is different

image
image
image

nanaya:
it should be positioned the same as when there's any favorites and the tooltip should say something about being no favorites yet

also please rebase and squash the commits

diquoks:
smth like that?
image image

resources/js/beatmapsets-show/header.tsx
```diff
@@ -236,15 +236,14 @@ export default class Header extends React.Component<Props> {
...
     }
 
-    this.favouritePopupDisposer ??= createTooltip(
+    this.favouritePopupDisposer = createTooltip(
       () => this.favouriteIconRef.current,
       () => ({
         count: this.controller.beatmapset.favourite_count,
-        users: this.filteredFavourites,
+        title: this.filteredFavourites.length < 1 ? trans('beatmapsets.show.stats.no_favourites') : undefined,
+        users: this.filteredFavourites.length < 1 ? [] : this.filteredFavourites,
```
nanaya: array of length of 0 is already `[]`?
```suggestion
        users: this.filteredFavourites,
```

resources/js/beatmapsets-show/header.tsx
```diff
@@ -236,15 +236,14 @@ export default class Header extends React.Component<Props> {
         this.favouritePopupDisposer();
         $(this.favouriteIconRef.current ?? []).qtip('destroy', true);
       }
-
-      return;
```
nanaya, line 240: this whole check should be removed as the tooltip is changed to be always shown

resources/js/beatmapsets-show/header.tsx
```diff
@@ -236,15 +236,14 @@ export default class Header extends React.Component<Props> {
...
     }
 
-    this.favouritePopupDisposer ??= createTooltip(
+    this.favouritePopupDisposer = createTooltip(
```
nanaya: the `??=` should stay...

resources/js/components/user-list-popup.tsx
```diff
@@ -76,7 +76,7 @@ export default function UserListPopup(props: Props) {
           ))}
         </div>
       }
-      {props.count > props.users.length &&
+      {props.count > props.users.length && props.users.length > 0 &&
```
nanaya, line 79: what's this for?

resources/js/beatmapsets-show/header.tsx
diquoks: actually yeah, overlooked it

resources/js/beatmapsets-show/header.tsx
diquoks: ooops, i got mixed it up with another language

resources/js/components/user-list-popup.tsx
diquoks: | with | without |
|-|-|
| image | image |

nanaya: how did you get that state?

resources/js/beatmapsets-show/header.tsx
diquoks: reverted in https://github.com/ppy/osu-web/pull/12323/commits/4d2a8e2a9b8b569ff58eea54d8683d32429deee0

resources/js/beatmapsets-show/header.tsx
diquoks: deleted in https://github.com/ppy/osu-web/pull/12323/commits/4d2a8e2a9b8b569ff58eea54d8683d32429deee0

resources/js/beatmapsets-show/header.tsx
diquoks: reverted in https://github.com/ppy/osu-web/pull/12323/commits/4d2a8e2a9b8b569ff58eea54d8683d32429deee0

diquoks:
very sry for such a stupid pr with a bunch of mistakes, it may seem that i started coding only yesterday 😅

resources/js/components/user-list-popup.tsx
diquoks: other 52 users cannot be displayed because they are missing on my local `osu-web` btw
image

nanaya: broken data aside, how does it make sense to have the number says 52 but the tooltip says no favorites...

diquoks: > broken data aside, how does it make sense to have the number says 52 but the tooltip says no favorites...

they cannot be displayed because they're missing in database, everything will be fine in real `osu-web`

diquoks:
i'd better close it and stay out of ts

PR #12324, opened 2025-08-01
Update class name for verification mail state

nanaya:
Preparing for a different `State` class 😄 

Things may explode when php tries to deserialise existing class...? I guess I can create a blank class inheriting the new class name but not sure if worth it.


notbakaneko (approved):
seems safe 👀

PR #12325, opened 2025-08-01
Calculate multiplayer room difficulty range based only on non-expired items when the room is open

bdach:
- Fixes the other part of https://github.com/ppy/osu/issues/34379.
- See also: https://github.com/ppy/osu/pull/34464.

Note the caveat that this only applies ***when the room is open***. After the room has ended expired items are taken into account again. This is to avoid showing an empty [0.00*, 0.00*] range on ended rooms, since after all ended rooms should invariably have all of their playlist items expired.


peppy:
Logic looks about what I'd expect 👍

app/Models/Multiplayer/Room.php
```diff
@@ -399,12 +403,18 @@ public function difficultyRange()
         }
 
         if ($extraQuery) {
+            $beatmapIds = $this->playlist();
```
nanaya: the `->select` can go here

nanaya (changes requested):
is it possible to have open room with all items expired? I think it'll currently show a 0-0 range in such case...?

peppy:
> is it possible to have open room with all items expired? I think it'll currently show a 0-0 range in such case...?

I wonder if it should use the most recent non-expired beatmap in this scenario... or the previous method of including expired items? Both seem better than undefined.

bdach:
yes, it is possible to have all items expired. I forgot the non-host queue modes don't clone the last-played playlist item after running out.

unsure what you would have me do in this case. falling back to all items played ever would be simplest. probably.

nanaya:
just pick one I guess...

app/Models/Multiplayer/Room.php
```diff
@@ -376,13 +376,15 @@ public function difficultyRange()
         $extraQuery = true;
 
         if ($this->relationLoaded('playlist')) {
+            $anyNonExpiredItems = $this->playlist->where('expired', '=', false)->isNotEmpty();
```
nanaya: is this supposed to be used inside the `$extraQuery` check down there? (otherwise this should be set right before the loop it's used instead)

app/Models/Multiplayer/Room.php
```diff
@@ -403,11 +405,11 @@ public function difficultyRange()
         }
 
         if ($extraQuery) {
-            $beatmapIds = $this->playlist();
-            if (!$this->hasEnded()) {
-                $beatmapIds = $beatmapIds->where('expired', '=', false);
-            }
-            $beatmapIds = $beatmapIds->select('beatmap_id');
+            $playlistItems = $this->playlist()->select(['beatmap_id', 'expired'])->get();
```
nanaya: maybe set this inside the relationLoaded check above as well (with `$this->playlist`) and assign it here only if unset.

app/Models/Multiplayer/Room.php
```diff
@@ -403,11 +405,11 @@ public function difficultyRange()
...
+
+            $beatmapIds = $playlistItems->where('expired', '=', false)->isNotEmpty()
+                ? $playlistItems->where('expired', '=', false)->pluck('beatmap_id')
+                : $playlistItems->pluck('beatmap_id');
```
nanaya: the `where()` here is executed immediately so it may loop the same thing twice

```php
$activePlaylistItems ??= $playlistItems->where('expired', '=', false);
$beatmapIds = (
  $activePlaylistItems->isNotEmpty()
    ? $activePlaylistItems
    : $playlistItems
)->pluck('beatmap_id')
```
maybe?

Similarly for the loop inside `relationLoaded` above.

app/Models/Multiplayer/Room.php
```diff
@@ -376,12 +376,20 @@ public function difficultyRange()
...
             if ($this->playlist->count() > 0) {
                 $firstItem = $this->playlist[0];
+                $anyNonExpiredItems = $activePlaylistItems->isNotEmpty();
 
                 if ($firstItem->relationLoaded('beatmap')) {
                     $extraQuery = false;
                     foreach ($this->playlist as $item) {
+                        if ($anyNonExpiredItems && $item->expired) {
+                            continue;
+                        }
```
nanaya:
```suggestion
                    foreach ($anyNonExipiredItems ? $activePlaylistItems : $playlistItems as $item) {
```
but maybe don't inline the conditional. and looking again, maybe
```php
...
$activePlaylistItems = ...;
// or some better named variable
$displayPlaylistItems = $activePlaylistItems->isNotEmpty() ? $activePlaylistItems : $playlistItems;
...

...
if ($extraQuery) {
    if (!isset($displayPlaylistItems)) {
        $playlistItems = ...;
        ...
        $displayPlaylistItems = ...;
    }
    // maybe short enough to be inlined
    $beatmapIds = $displayPlaylistItems->pluck('beatmap_id');
    ...
}
```

nanaya (changes requested):
also it would be nice if the commit message says more specifically the what and/or why of the changes...?

bdach:
> also it would be nice if the commit message says more specifically the what and/or why of the changes...?

I would love to provide better commit messages but to be quite honest I'm at the stage of barely understanding why I'm doing these changes. I am *guessing* some of this is being suggested for performance purposes but with laravel magic it is very difficult for me to tell what anything is going to do. laravel api docs on `Collection` are 404ing on me, phpstorm code navigation to look at what any method does doesn't work half the time because of php dynamic magic, and I haven't figured out how to attach a debugger to docker, so all ways that I usually use to learn how anything works are of no use here.

nanaya:
yeah at least here it's just mostly performance

(it's probably at microoptimization level but the entire base framework is already built on pile of inefficiencies so I'd take any small wins left as long it's not too verbose/confusing)

peppy:
Is this blocked on something at the moment?

nanaya:
apparently I missed the latest update

PR #12327, opened 2025-08-03
Add ability to hide judges from contest results

Hiviexd:
It's somewhat common for contests to anonymize judges from their results to protect them from harassment, so I feel such setting is needed for our contest system.

This adds a `show_judges` attribute to `extra_options`, it's `true` by default to not affect past contests, so actual usage would be adding `show_judges: false` to `extra_options`.

### DIFF
--- a/app/Libraries/SessionVerification/Controller.php
+++ b/app/Libraries/SessionVerification/Controller.php
@@ -63,13 +63,13 @@ class Controller
 
         $key = strtr(get_string(\Request::input('verification_key')) ?? '', [' ' => '']);
         $user = Helper::currentUserOrFail();
-        $state = State::fromSession($session);
+        $mailState = MailState::fromSession($session);
 
         try {
-            if ($state === null) {
+            if ($mailState === null) {
                 throw new UserVerificationException('expired', true);
             }
-            $state->verify($key);
+            $mailState->verify($key);
         } catch (UserVerificationException $e) {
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/ContestEntriesController.php
```diff
@@ -46,14 +51,19 @@ public function judgeResults($contestId, $id)
...
             'judge_votes.total_score_std',
-            'judge_votes.user',
             'results',
             'user',
-        ]);
+        ];
+
+        if (!$contest->anon_judges) {
+            $includes[] = 'judge_votes.user';
+        }
```
notbakaneko: move this up together with `$relationships`

app/Models/Contest.php
```diff
@@ -157,6 +157,11 @@ public function isJudgingActive(): bool
         return $this->isJudged() && $this->isVotingOpen() && !$this->show_votes;
     }
 
+    public function getAnonJudgesAttribute()
+    {
+        return $this->getExtraOptions()['anon_judges'] ?? false;
```
notbakaneko: I think this should be called `show_judges` to be inline with the other `show_x` options

notbakaneko (changes requested):

notbakaneko (approved):
the test is kind of not useful but w/e; `showEntryUser` has one because it's a state combination

PR #12329, opened 2025-08-04
Adjust user group display on mobile

### LINKED ISSUE #11851: Usergroups can become untappable on mobile if a user has too many
Visiting [Maxus' profile page](https://osu.ppy.sh/users/4335785) on mobile and trying to tap the "TC" usergroup makes it extremely difficult to do so without zooming in due to the placement of the usergroup next to the display banner button.

Image

nanaya:
Resolves #11851 maybe? 🤷 

The cover display toggle still may occasionally block thing but that's sort of a different problem. As for the spotlight badge, there's a different open issue for that.

Also fixes a subtle bug of extra padding on country and team flags.


resources/css/bem/profile-info.less
```diff
@@ -89,24 +93,29 @@
   &__details {
     --gutter-size: @gutter-v2;
     --gutter-size-desktop: @gutter-v2-desktop;
-    padding: 0 var(--gutter-size);
+    --details-vertical-padding: var(--vertical-padding);
+    padding: var(--details-vertical-padding) var(--gutter-size);
     display: flex;
     align-items: center;
     position: relative;
-    height: calc(var(--content-height) + var(--vertical-padding) * 2);
```
nanaya, line 96: I can't remember if there's reason to specify this on mobile 🤔

notbakaneko (changes requested):
Team with long name doesn't wrap like it should anymore

Image

PR #12333, opened 2025-08-05
Remove remember login option

nanaya:
I don't think there's option for that on the frontend anymore? Or ever?

And actually enabling this option on login adds a 400 bytes of cookies thanks to how efficient laravel's cookies are.


notbakaneko (approved):
I don't even remember this option (or it being a good idea)

PR #12334, opened 2025-08-06
Clean up login action parameter handling

nanaya:
Also fixes passing `null` to `trim()`.


app/Http/Controllers/SessionsController.php
```diff
@@ -29,9 +29,14 @@ public function store()
     {
         $request = request();
 
-        $params = get_params($request->all(), null, ['username:string', 'password:string', 'cf-turnstile-response:string']);
-        $username = presence(trim($params['username'] ?? null));
-        $password = presence($params['password'] ?? null);
+        $params = get_params(
+            $request->all(),
+            null,
+            ['username:string', 'password', 'cf-turnstile-response'],
```
nanaya, line 35: weirdly enough, `:string` does `get_string()` while unspecified does `presence(get_string())` 👀

notbakaneko (approved):

PR #12337, opened 2025-08-06
Add supporting pieces for displaying lazer leaderboard first place-related content on user profiles

bdach:
First step toward https://github.com/ppy/osu-infrastructure/issues/27.

You'd think the first step would be to create a table, but it seems one already exists since https://github.com/ppy/osu-web/pull/9127, it just fell down the back of the sofa so to speak.

The general intention is that `osu-queue-score-statistics` will be updating `beatmap_leaders` as it processes scores.

Also this PR intentionally isn't *reading* anything from `beatmap_leaders` yet; I'm splitting that part of the task off to sequence it being merged *after* the table is alive (read: the `osu-queue-score-statistics` change goes live) so that people don't complain that they have no first place ranks visible on their profile with lazer mode on.

## [Add shared interop endpoint for emitting "first place lost" profile events](https://github.com/ppy/osu-web/commit/04378a532ac5d9b184c57d13ee00be7a1d0078dd)

To be used by `osu-queue-score-statistics`.

## [Clear beatmap leaders on score deletion](https://github.com/ppy/osu-web/commit/6a880f79c10d7b0757c83e4b6abdecd84ee8a620)

Covers scenarios such as moving from qualified to ranked (and vice versa).

This is being intentionally done in this PR *before* anything is actually written to this table to make sure nothing will need to be *deleted* from it manually once it actually goes alive.


app/Http/Controllers/InterOp/UsersController.php
```diff
@@ -54,7 +54,28 @@ public function rankAchieved($userId, $beatmapId, $rulesetId)
...
+    }
+
+    public function firstPlaceLost($userId, $beatmapId, $rulesetId)
+    {
+        $params = get_params(
+            request()->all(),
+            null,
+            [
+                'position_after:int',
+                'rank:string',
```
nanaya: neither are used?

(and if it's just one, there's shorter way with `get_bool(request('...'))`)

tests/Jobs/RemoveBeatmapsetSoloScoresTest.php
```diff
@@ -51,8 +53,17 @@ public function testHandle()
 
         static::reindexScores();
 
+        foreach ($beatmapset->beatmaps as $beatmap) {
+            foreach (array_values(Beatmap::MODES) as $ruleset) {
+                BeatmapLeader::sync($beatmap->getKey(), $ruleset);
```
nanaya: more like `$rulesetId`

tests/Jobs/RemoveBeatmapsetSoloScoresTest.php
```diff
@@ -51,8 +53,17 @@ public function testHandle()
...
+            }
+        }
+        $this->assertNotEmpty(BeatmapLeader::whereIn('beatmap_id', $beatmapset->beatmaps->pluck('beatmap_id'))->where('score_id', '<=', $maxScoreId)->get());
```
nanaya: this line can be a bit shorter if `$beatmapset->beatmaps->pluck('beatmap_id')` is assigned to `$beatmapIds` before the leader sync (and it can be used in the loop as well instead of repeating `->getKey()`).

also maybe `->assertFalse(...->empty())` (and `assertTrue` down there) dunno if any better apart of slightly lighter query 🤷‍♀️

tests/Jobs/RemoveBeatmapsetSoloScoresTest.php
```diff
@@ -68,6 +79,7 @@ public function testHandle()
                 User::truncate();
                 UserGroup::truncate();
                 UserGroupEvent::truncate();
+                BeatmapLeader::truncate();
```
nanaya: this model truncation section is alphabetically sorted

app/Models/Event.php
```diff
@@ -198,6 +198,28 @@ public static function generate($type, $options)
...
                 break;
 
+            case 'rankLost':
+                $beatmap = $options['beatmap'];
+                $ruleset = $options['ruleset'];
+                $rulesetName = trans("beatmaps.mode.{$ruleset}");
+                $beatmapLink = static::beatmapLink($beatmap, $ruleset);
+                $user = $options['user'];
+                $userLink = static::userLink($user);
+                $legacyScoreEvent = $options['legacy_score_event'];
```
nanaya: taking out options to separate variables don't do much anymore. I think it was originally for the inline html generation so it's slightly shorter. but it's quite different now and just adding lines unnecessarily.

although I guess if the options are taken out first instead of being interleaved with `$rulesetName` etc, it can double as required parameter docs.

tests/Jobs/RemoveBeatmapsetSoloScoresTest.php
```diff
@@ -51,8 +53,17 @@ public function testHandle()
 
         static::reindexScores();
 
+        foreach ($beatmapset->beatmaps as $beatmap) {
+            foreach (array_values(Beatmap::MODES) as $ruleset) {
```
nanaya: no need to `array_values`, it's already doing that just with `Beatmap::MODES as $rulesetId`

app/Http/Controllers/InterOp/UsersController.php
bdach: ^C^V moment...

tests/Jobs/RemoveBeatmapsetSoloScoresTest.php
bdach: I've applied the first half of this but

> also maybe `->assertFalse(...->empty())` (and `assertTrue` down there) dunno if any better apart of slightly lighter query 🤷‍♀️

for this I'm not sure what you'd want me to do. If it's 

```diff
diff --git a/tests/Jobs/RemoveBeatmapsetSoloScoresTest.php b/tests/Jobs/RemoveBeatmapsetSoloScoresTest.php
index 7ab918a4f99..41ab3e4613d 100644
--- a/tests/Jobs/RemoveBeatmapsetSoloScoresTest.php
+++ b/tests/Jobs/RemoveBeatmapsetSoloScoresTest.php
@@ -60,11 +60,11 @@ class RemoveBeatmapsetSoloScoresTest extends TestCase
                 BeatmapLeader::sync($beatmapId, $rulesetId);
             }
         }
-        $this->assertNotEmpty(BeatmapLeader::whereIn('beatmap_id', $beatmapIds)->where('score_id', '<=', $maxScoreId)->get());
+        $this->assertFalse(BeatmapLeader::whereIn('beatmap_id', $beatmapIds)->where('score_id', '<=', $maxScoreId)->empty());
 
         $job->handle();
 
-        $this->assertEmpty(BeatmapLeader::whereIn('beatmap_id', $beatmapIds)->where('score_id', '<=', $maxScoreId)->get());
+        $this->assertTrue(BeatmapLeader::whereIn('beatmap_id', $beatmapIds)->where('score_id', '<=', $maxScoreId)->empty());
 
         $search = new ScoreSearch();
         // this also makes sure the job deletes scores from index


```

then that fails for me because (I think???) that's not a laravel collection yet but still an eloquent query or something. I couldn't find anything in eloquent that would look `->empty()`-shaped.

app/Models/Event.php
bdach: > although I guess if the options are taken out first instead of being interleaved with `$rulesetName` etc, it can double as required parameter docs.

Have done just this for now.

tests/Jobs/RemoveBeatmapsetSoloScoresTest.php
nanaya: whoops I meant `->exists()`

app/Models/Event.php
```diff
@@ -198,6 +198,29 @@ public static function generate($type, $options)
                 ];
                 break;
 
+            case 'rankLost':
+                $beatmap = $options['beatmap'];
+                $ruleset = $options['ruleset'];
+                $user = $options['user'];
+                $legacyScoreEvent = $options['legacy_score_event'];
+
+                $rulesetName = trans("beatmaps.mode.{$ruleset}");
```
nanaya: I keep forgetting (and not really relevant here since it should always be run in english locale) but in general `osu_trans(...)` should be used instead of `trans(...)` because there's some additional behaviours needed to handle crowdin output...

```
> trans('beatmaps.mode.osu', locale: 'ja')
= ""

> osu_trans('beatmaps.mode.osu', locale: 'ja')
= "osu!"
```

app/Http/Controllers/InterOp/UsersController.php
```diff
@@ -54,7 +54,20 @@ public function rankAchieved($userId, $beatmapId, $rulesetId)
...
 
         Event::generate('rank', $params);
-        // TODO: also emit lost first place rank event when relevant
+        return response([], 204);
+    }
+
+    public function firstPlaceLost($userId, $beatmapId, $rulesetId)
+    {
+        $params = request()->all();
+        abort_unless(isset($params['legacy_score_event']), 422, 'missing legacy_score_event parameter');
```
nanaya: this check isn't quite correct because you can pass value `x` and it'll pass this check but then the value ends up being `null` when casted.

```suggestion
        $legacyScoreEvent = get_bool(request('legacy_score_event'));
        abort_unless(isset($legacyScoreEvent), 422, 'missing legacy_score_event parameter');
```

(although now I look again, the message is kinda wrong and similar thing can be said about `$rulesetId`...)

PR #12338, opened 2025-08-06
Add retry logic to the commonly failing dusk test

nanaya:
I think this works...? Should probably extract it somewhere but I'm inlining it for now for testing purpose.

And maybe just doing the reset every time in the `setUp` could also work...? Instead of the current cookie reset.

<sub>wouldn't it be funny if the test fails right in this pr</sub>


notbakaneko:
And then it fails after merge on a different test 🙈

nanaya:
...well fuck 🚮

PR #12343, opened 2025-08-07
Fix incorrect html entity escaping on `text_clean` in events

bdach:
As mentioned in https://github.com/ppy/osu-web/pull/12337#pullrequestreview-3096222523

I can remove the test if it's too much, just found it the most convenient way to check that this is doing the right thing.

### DIFF
--- a/resources/css/bem/mod.less
+++ b/resources/css/bem/mod.less
@@ -77,24 +77,29 @@
       }
     }
 
-    .mod-icon-fa(@acronym, @glyph) {
-      display: flex;
-      content: @glyph;
-      font-size: 0.8em;
-      // slight adjustments for small sizes
-      padding-left: 1px;
+    .mod-icon-fa() {
+      content: "";
+      --margin: 0.32em;
+      margin: var(--margin);
+      width: calc(100% - var(--margin) * 2);
+      height: calc(100% - var(--margin) * 2);
+
+      background-color: var(--type-fg-colour);
+      mask-size: contain;
+      mask-position: center;
+      mask-repeat: no-repeat;
     }
 
     .mod-icon-fas(@acronym, @glyph) {
       &--@{acronym}::after {
-        .fas();
-        .mod-icon-fa(@acronym, @glyph);
+        .mod-icon-fa();
+        mask-image: url("~@fortawesome/fontawesome-free/svgs/solid/@{glyph}.svg");
       }
     }
     .mod-icon-far(@acronym, @glyph) {
       &--@{acronym}::after {
-        .far();
-        .mod-icon-fa(@acronym, @glyph);
+        .mod-icon-fa();
+        mask-image: url("~@fortawesome/fontawesome-free/svgs/regular/@{glyph}.svg");
       }
     }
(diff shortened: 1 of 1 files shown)


app/Models/Event.php
```diff
@@ -528,8 +532,10 @@ private static function beatmapsetParams($beatmapset)
     private static function beatmapParams($beatmap, $ruleset)
     {
         $url = e(route('beatmaps.show', ['beatmap' => $beatmap, 'ruleset' => $ruleset], false));
```
nanaya: probably doesn't really matter much but this shouldn't be escaped for the `_clean` variant either...?

app/Models/Event.php
```diff
@@ -528,8 +532,10 @@ private static function beatmapsetParams($beatmapset)
...
+        $title = "{$beatmap->beatmapset->artist} - {$beatmap->beatmapset->title} [{$beatmap->version}]";
         return [
-            'title' => e("{$beatmap->beatmapset->artist} - {$beatmap->beatmapset->title} [{$beatmap->version}]"),
+            'title' => e($title),
+            'title_clean' => $title,
```
nanaya: or actually, considering how they end up being used just to generate links and that none of these being reused due to differing escaping, just return `'link' => tag('a', ...), 'link_clean' => "[{$url} {$title}]"` for all those params thing instead...? and call the functions `thingLink` or something

each of these functions can just first generate the title and url and pass it to a function which generate both link variants

one weird exception looks like the username change which uses a different name but accepting an additional parameter for the function may be workable (or some other methods)

there are a few places where `$user['id']` is being used but it's easily available through a different mean anyway.

(also this whole thing is kinda what web-10 used to do...)

nanaya (changes requested):

app/Models/Event.php
```diff
@@ -504,34 +507,33 @@ public function scopeRecent($query, null | true $legacyOnly)
...
+    private static function beatmapLink($beatmap, $ruleset)
     {
-        $url = e(route('beatmaps.show', ['beatmap' => $beatmap, 'ruleset' => $ruleset], false));
+        $url = route('beatmaps.show', ['beatmap' => $beatmap, 'ruleset' => $ruleset], false);
+        $title = "{$beatmap->beatmapset->artist} - {$beatmap->beatmapset->title} [{$beatmap->version}]";
         return [
-            'title' => e("{$beatmap->beatmapset->artist} - {$beatmap->beatmapset->title} [{$beatmap->version}]"),
-            'url' => $url,
-            'url_clean' => $GLOBALS['cfg']['app']['url'].$url,
+            'html' => tag('a', ['href' => $url], e($title)),
```
nanaya: oh crap, I just remembered this won't parse correctly because the function returns attributes wrapped in `"` but the regexp expects `'` 😅  fun php string <del>interpolation</del> concatenation time, I guess (or `sprintf`?)

maybe good idea to add `$this->assertArrayNotHasKey('parse_error', $event->parse()->details);` to the tests... (or some other sort of parsing check)

PR #12345, opened 2025-08-12
Add Beatmapset search tests

notbakaneko:
The keyword inclusion filter part of the tests being added for #12148
So that we can add test cases of what results are expected to match or not. 

`php test phpunit --filter=BeatmapsetSearch` to run just the search tests

No querystring or tag search tests yet because I'm still working out that out...


tests/Libraries/Search/BeatmapsetSearch/TestCase.php
```diff
@@ -0,0 +1,78 @@
...
+
+        $this->assertEqualsCanonicalizing($beatmapsetIds, new BeatmapsetSearch(new BeatmapsetSearchRequestParams($params))->response()->ids());
+    }
+
+    #[DataProvider('dataProvider')]
+    public function testSearch(array $params, array $expected): void
+    {
+        $anyParams = new BeatmapsetSearchParams();
+        $anyParams->status = 'any';
+        $this->assertCount(count(static::$beatmapsets), new BeatmapsetSearch($anyParams)->response()->ids());
```
nanaya: this can be its own test instead of being run every time?

tests/Libraries/Search/BeatmapsetSearch/TestCase.php
```diff
@@ -0,0 +1,78 @@
...
+
+abstract class TestCase extends BaseTestCase
+{
+    protected static iterable $beatmapsets;
+
+    abstract public static function dataProvider(): array;
+
+    public static function setUpBeforeClass(): void
+    {
+        static::deleteAllEsBeatmapsets();
```
nanaya: seems weird doing cleanup for es (assuming dirty state) but not for the database entries...

tests/Libraries/Search/BeatmapsetSearch/SimpleFilterTest.php
```diff
@@ -0,0 +1,119 @@
...
+                            'diff_approach' => 2 + $sequence->index,
+                            'diff_drain' => 2 + $sequence->index,
+                            'diff_overall' => 2 + $sequence->index,
+                            'diff_size' => 2 + $sequence->index,
+                            'difficultyrating' => 2 + $sequence->index,
+                    ]))
+                )
+                    ->create([
+                        'favourite_count' => 200,
+                    ]),
```
nanaya: what kind of style is this...
```suggestion
                )->create([
                    'favourite_count' => 200,
                ]),
```
or may even just make the whole function call single lined

tests/Libraries/Search/BeatmapsetSearch/SimpleFilterTest.php
```diff
@@ -0,0 +1,119 @@
...
+                            'hit_length' => 2 + $sequence->index,
+                            'countNormal' => 2 + $sequence->index,
+                            'countSlider' => 2 + $sequence->index,
+                            'bpm' => 2 + $sequence->index,
+                            'diff_approach' => 2 + $sequence->index,
+                            'diff_drain' => 2 + $sequence->index,
+                            'diff_overall' => 2 + $sequence->index,
+                            'diff_size' => 2 + $sequence->index,
+                            'difficultyrating' => 2 + $sequence->index,
+                    ]))
```
nanaya: while at it
```suggestion
                        ]))
```

tests/Libraries/Search/BeatmapsetSearch/RankedFilterTest.php
```diff
@@ -0,0 +1,57 @@
...
+
+    public static function setUpBeforeClass(): void
+    {
+        parent::setUpBeforeClass();
+
+        static::withDbAccess(function () {
+            $factory = Beatmapset::factory()->withBeatmaps();
+            $helper = fn (DateTimeInterface $date) => $factory->ranked($date)->state([
+                'last_update' => $date,
+                'submit_date' => $date,
```
nanaya: this means the test won't check if it's testing the correct field (since same date applies to all three fields) but maybe not really needed...? 🤷

tests/Libraries/Search/BeatmapsetSearch/ManiaKeysFilterTest.php
```diff
@@ -0,0 +1,42 @@
...
+    public static function setUpBeforeClass(): void
+    {
+        parent::setUpBeforeClass();
+
+        static::withDbAccess(function () {
+            $beatmapsetFactory = Beatmapset::factory()->ranked();
+            $beatmapFactory = Beatmap::factory()->ruleset('mania')->state(['approved' => Beatmapset::STATES['ranked']]);
+
+            static::$beatmapsets = [
+                $beatmapsetFactory->withBeatmaps('osu')->create(),
```
nanaya: shouldn't this have `diff_size => 7` as well?

tests/Libraries/Search/BeatmapsetSearch/SimpleFilterTest.php
```diff
@@ -0,0 +1,119 @@
...
+
+        static::withDbAccess(function () {
+            $factory = Beatmapset::factory()->ranked();
+            $beatmapFactory = Beatmap::factory()->ranked()->ruleset('osu');
+            static::$beatmapsets = [
+                $factory->has(
+                    $beatmapFactory
+                        ->count(2)
+                        ->state(new Sequence(fn (Sequence $sequence) => [
+                            'total_length' => 1 + $sequence->index,
```
nanaya: this has same problem as the date related tests

tests/Libraries/Search/BeatmapsetSearch/TestCase.php
```diff
@@ -0,0 +1,78 @@
...
+
+    protected function searchAndAssert(array $params, array $expected): void
+    {
+        $beatmapsetIds = array_map(fn (int $index) => static::$beatmapsets[$index]->getKey(), $expected);
+
+        $this->assertEqualsCanonicalizing($beatmapsetIds, new BeatmapsetSearch(new BeatmapsetSearchRequestParams($params))->response()->ids());
+    }
+
+    #[DataProvider('dataProvider')]
+    public function testSearch(array $params, array $expected): void
```
nanaya: before protected functions?

tests/Libraries/Search/BeatmapsetSearch/TestCase.php
```diff
@@ -0,0 +1,78 @@
...
+    {
+        $anyParams = new BeatmapsetSearchParams();
+        $anyParams->status = 'any';
+        $this->assertCount(count(static::$beatmapsets), new BeatmapsetSearch($anyParams)->response()->ids());
+
+        $this->searchAndAssert($params, $expected);
```
nanaya: why is this separate function? If anything the general state check should be its own function instead...

tests/Libraries/Search/BeatmapsetSearch/FeaturedArtistFilterTest.php
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public static function setUpBeforeClass(): void
+    {
+        parent::setUpBeforeClass();
+
+        static::withDbAccess(function () {
+            $artists = Artist::factory()
+                ->count(2)
+                ->state(new Sequence(fn (Sequence $sequence) => ['id' => $sequence->index + 1]))
```
nanaya: seems weird to have a function with typing for the parameters but not the return...?

tests/Libraries/Search/BeatmapsetSearch/TestCase.php
```diff
@@ -0,0 +1,78 @@
...
+            Genre::truncate();
+            Language::truncate();
+            User::truncate();
+            UserGroup::truncate();
+        });
+
+        static::$beatmapsets = [];
+    }
+
+    protected static function refresh(): void
```
nanaya: can use a bit more descriptive name

nanaya (changes requested):

tests/Libraries/Search/BeatmapsetSearch/TestCase.php
nanaya: I think doing this in set up where they're created is ok... just compare and throw generic exception or something. it's just setup check after all

tests/Libraries/Search/BeatmapsetSearch/TestCase.php
notbakaneko: It's a bit of a problem because we don't want to be doing and waiting for a synchronous delete all after every test or class, and it also triggers conflict errors in test cases that do update the index asynchronously, so it is kind of opt-in at the moment. 😐

nanaya: which thing is opt-in?

I'm not sure I understand what you're saying here...

notbakaneko: clearing out the index only for tests that check it.

Unless you mean you want to truncate the tables before and after the TestCase?

nanaya: I'd probably only clear out the index together with clearing out the tables (only after the test case). It doesn't seem to make sense to just clean one but not the other.

nanaya: oh you mean the tests other than this whole beatmapset search?

It doesn't seem to actually index beatmaps sometimes though. I can't quite tell what the condition is. Seems like something with transaction but not sure what exactly.

some comments noting the situation would be useful - mainly tests involving beatmaps may add to index but not clean up afterwards.

notbakaneko: ~~Now that you mentions it, there was something like that, so maybe I can fix it here~~ 🤔 
It turns out everything should be getting indexed but model factories with `afterCreating` that involve extra calls that have transactions anywhere have been preventing the final commit from happening (And since `User` is practically everywhere, it's been preventing the indexing commit from triggering)

tests/Libraries/Search/BeatmapsetSearch/SimpleFilterTest.php
```diff
@@ -0,0 +1,103 @@
...
+        static::withDbAccess(function () {
+            $factory = Beatmapset::factory()->ranked();
+            $beatmapFactory = Beatmap::factory()->ranked()->ruleset('osu');
+
+            $props = array_values(static::KEYS);
+            $baseValues = [];
+            foreach ($props as $offset => $prop) {
+                $baseValues[$prop] = $offset;
+            }
+            $baseValues['hit_length'] = $baseValues['total_length']; // don't use the rand in factory.
```
nanaya: but why not use the rand (if it causes error then maybe fix the factory instead...? if it's some other reason then can use explanation)

tests/Libraries/Search/BeatmapsetSearch/SimpleFilterTest.php
```diff
@@ -0,0 +1,103 @@
...
+
+            $offsetValues = function (int $offset) use ($baseValues): array {
+                $props = [];
+                foreach ($baseValues as $key => $value) {
+                    $props[$key] = $value + $offset;
+                }
+
+                return $props;
+            };
```
nanaya: this?
```suggestion
            $offsetValues = fn (int $offset): array => array_map(
                fn (int $value): int => $value + $offset,
                $baseValues,
            );
```

tests/Libraries/Search/BeatmapsetSearch/TestCase.php
```diff
@@ -0,0 +1,90 @@
...
+
+abstract class TestCase extends BaseTestCase
+{
+    protected static iterable $beatmapsets = [];
+
+    abstract public static function dataProvider(): array;
+
+    /**
+     * @afterClass
+     * @beforeClass
```
nanaya: use attributes?

tests/Libraries/Search/BeatmapsetSearch/TestCase.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        new BeatmapsetSearch()->deleteAll();
+    }
+
+    public static function setUpBeforeClass(): void
+    {
+        $count = count(static::$beatmapsets);
+        if ($count === 0) {
+            throw new \Exception('No beatmapsets added to test setup.');
+        }
```
nanaya, line 50: seems like a weird thing to check...? what situation is this guarding against?

tests/Libraries/Search/BeatmapsetSearch/CreatorFilterTest.php
```diff
@@ -0,0 +1,41 @@
...
+    public static function setUpBeforeClass(): void
+    {
+        static::withDbAccess(function () {
+            $factory = Beatmapset::factory()->ranked();
+            $mapper1 = User::factory()->create(['username' => 'mapper']);
+            $mapper2 = User::factory()->create(['username' => 'another_mapper']);
+
+            static::$beatmapsets = [
+                $factory->withBeatmaps()->withBeatmaps(guestMapper: $mapper1)->create(['creator' => 'someone']),
+                $factory->withBeatmaps()->create(['user_id' => $mapper1]),
```
nanaya: can move the `->withBeatmaps()` to the initial `$factory` assignment?

tests/Libraries/Search/BeatmapsetSearch/RankedFilterTest.php
```diff
@@ -0,0 +1,60 @@
...
+
+declare(strict_types=1);
+
+namespace Tests\Libraries\Search\BeatmapsetSearch;
+
+use App\Models\Beatmapset;
+use Carbon\CarbonImmutable;
+
+// Includes created and updated date tests.
+class RankedFilterTest extends TestCase
```
nanaya: `SimpleDateFilterTest`? To match `SimpleFilterTest`

tests/Libraries/Search/BeatmapsetSearch/SimpleFilterTest.php
```diff
@@ -0,0 +1,103 @@
...
+        $data = [];
+        foreach (array_keys(static::KEYS) as $offset => $key) {
+            $value = $offset + 2;
+            $otherValue = $offset + 1;
+            $data[] = [['q' => "{$key}={$value}"], [0, 1]];
+            $data[] = [['q' => "{$key}>{$value}"], [1, 2]];
+            $data[] = [['q' => "{$key}>={$value}"], [0, 1, 2]];
+            $data[] = [['q' => "{$key}<{$value}"], [0]];
+            $data[] = [['q' => "{$key}<={$value}"], [0, 1]];
+            $data[] = [['q' => "{$key}>2 {$key}={$otherValue}"], [0]]; // same key overrides
```
nanaya: "overrides" isn't exactly correct. `stars>=1 stars<4` results in `0.995 <= stars <= 3.995`, not `stars <= 3.995` (also various behaviour on this is already tested by the query parser test)

tests/Libraries/Search/BeatmapsetSearch/TitleFilterTest.php
```diff
@@ -0,0 +1,39 @@
...
+namespace Tests\Libraries\Search\BeatmapsetSearch;
+
+use App\Models\Beatmapset;
+
+class TitleFilterTest extends TestCase
+{
+    public static function dataProvider(): array
+    {
+        return [
+            [['q' => 'title=best'], [0, 1, 2]],
```
nanaya, line 17: I think the test is missing something here... (a todo comment at least?)

On prod, I believe the exact "I will" titled maps should have highest relevance but it's currently not?

tests/Libraries/Search/BeatmapsetSearch/SimpleFilterTest.php
notbakaneko: I keep forgetting `array_map` just preserves the keys...

tests/Libraries/Search/BeatmapsetSearch/TestCase.php
notbakaneko: The (admittedly weird) case where `0 = 0` not because the index refresh didn't complete but because nothing was set up or set into the wrong place and it still passed.

tests/Libraries/Search/BeatmapsetSearch/TitleFilterTest.php
notbakaneko: All the ordering and updated query handling stuff will be in #12364

nanaya: The test still passes even without `consistent()` option...

```
diff --git a/tests/Libraries/Search/BeatmapsetSearch/TitleFilterTest.php b/tests/Libraries/Search/BeatmapsetSearch/TitleFilterTest.php
index 3f415ebdae..f5af178dd9 100644
--- a/tests/Libraries/Search/BeatmapsetSearch/TitleFilterTest.php
+++ b/tests/Libraries/Search/BeatmapsetSearch/TitleFilterTest.php
@@ -14,12 +14,12 @@ class TitleFilterTest extends TestCase
     public static function dataProvider(): array
     {
         return [
-            [['q' => 'best'], [0, 1, 2, 3, 4]],
-            [['q' => 'best beatmap'], [3, 1, 2, 0, 4], true],
-            [['q' => '"best beatmap"'], [3, 1, 2], true],
-            [['q' => '-best'], []],
-            [['q' => '-best -beatmap'], []],
-            [['q' => '-"best beatmap"'], [0, 4]],
+            //[['q' => 'best'], [0, 1, 2, 3, 4]],
+            //[['q' => 'best beatmap'], [3, 1, 2, 0, 4], true],
+            //[['q' => '"best beatmap"'], [3, 1, 2], true],
+            //[['q' => '-best'], []],
+            //[['q' => '-best -beatmap'], []],
+            //[['q' => '-"best beatmap"'], [0, 4]],

             [['q' => 'title=best'], [0, 1, 2, 3]],
             [['q' => 'title="best beatmap"'], [1, 2, 3]],
@@ -32,7 +32,7 @@ class TitleFilterTest extends TestCase
     public static function setUpBeforeClass(): void
     {
         static::withDbAccess(function () {
-            $factory = Beatmapset::factory()->consistent()->ranked()->withBeatmaps();
+            $factory = Beatmapset::factory()->ranked()->withBeatmaps();
             static::$beatmapsets = [
                 $factory->create(['title' => 'best']),
                 $factory->create(['title' => 'the best beatmap']),
```

nanaya: also, shouldn't the ordering be part of this pr...? or even why is it not applied for all tests in the first place

notbakaneko: `favourite_count` affects scoring enough to change ordering if there is enough difference so it should be isolated away

nanaya: but with just `title=`, the `queryString` is empty and thus doesn't affect anything? I can't find anywhere else the favourite count affecting the result...

https://github.com/ppy/osu-web/blob/0b569ed14ab6dce57852625727dbd7768e7be2f5/app/Libraries/Search/BeatmapsetSearch.php#L107-L113

nanaya: (as per the first comment, I think there's something weird with the `title=` search and the test should have at least a `TODO` comment about it - unless you figure out what's happening there)

notbakaneko: #12373 may be related ಠ_ಠ

nanaya: I guess this, artist, and creator filter search tests need randomised dates...? (also I think the order test should still be part of this pr and applied to all tests)

nanaya (changes requested):
still waiting on the result order test btw 🤷 

(to fix the failing case can maybe either recreate index where appropriate or update the test case so it isn't so ambiguous that a slight weighting change impacts the result)

PR #12346, opened 2025-08-13
Wrap another randomly failing test with retries 

nanaya:
Continues off #12344, now with actually reusable logic.

### DIFF
--- a/tests/Browser/LoginTest.php
+++ b/tests/Browser/LoginTest.php
@@ -38,18 +38,15 @@ class LoginTest extends DuskTestCase
     {
         $user = User::factory()->create();
 
-        static::withRetries(function () use ($user) {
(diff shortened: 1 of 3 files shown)


tests/Browser/LoginTest.php
```diff
@@ -39,32 +38,18 @@ public function testLogout(): void
...
-                static::closeAll();
-                if ($attempts++ > 5) {
-                    throw $e;
-                }
-            }
-            if (!$timedOut) {
-                break;
-            }
-        }
+        static::withRetries(function () use ($user) {
```
notbakaneko: Can use shorthand instead (the other one too, I guess)

notbakaneko (changes requested):
I look forward to `testLogin` timing out next 🤔

notbakaneko (approved):

PR #12355, opened 2025-08-18
User session verification using totp

### LINKED ISSUE #5163: Implement token-based 2FA authentication
Branching this off #2862, as it's now related to the client-specific stuff.

We should have an option for users to use a TOTP-based two-factor authentication method, in addition for the current, email-based one.

nanaya:
Just the verification part.

Enroll user and get the qr code with
```
App\Models\UserTotpKey::createOrFirstForUser(App\Models\User::find($id))->getQrCodeCli()
```

(the qr generator library also to be used to generate svg version for frontend)


routes/web.php
```diff
@@ -224,6 +224,7 @@
             Route::resource('sessions', 'Account\SessionsController', ['only' => ['destroy']]);
             Route::get('verify', 'AccountController@verifyLink');
             Route::post('verify', 'AccountController@verify')->name('verify');
+            Route::post('verify/mail-fallback', 'AccountController@verificationMailFallback')->name('verify-mail-fallback');
```
notbakaneko:
```suggestion
            Route::post('verify/mail-fallback', 'AccountController@verificationMailFallback')->name('verify.mail-fallback');
```

app/Http/Controllers/AccountController.php
```diff
@@ -312,6 +312,14 @@ public function updatePassword()
         }
     }
 
+    public function verificationMailFallback()
```
notbakaneko, line 315: I think this needs to skip verification or you can't fallback to mail? 🤔

notbakaneko (commented):
Still checking

app/Models/OAuth/Token.php
```diff
@@ -156,6 +162,17 @@ public function setScopesAttribute(?array $value)
         $this->attributes['scopes'] = $this->castAttributeAsJson('scopes', $value);
     }
 
+    public function setVerificationMethod(string $method): void
```
notbakaneko, line 165: missing on `SessionVerificationInterface`?

app/Models/OAuth/Token.php
```diff
@@ -99,6 +100,11 @@ public function getResourceOwner(): ?User
         return $this->user;
     }
 
+    public function getVerificationMethod(): ?string
```
notbakaneko, line 103: missing on `SessionVerificationInterface`?

app/Models/UserTotpKey.php
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public static function digits(): int
+    {
+        return TOTP::DEFAULT_DIGITS;
+    }
+
+    public static function generateUri(User $user): string
+    {
+        $totp = TOTP::generate();
```
notbakaneko: Should pass in `ClockInterface` to avoid the deprecation.

...which just defaults to 

```
public function now(): DateTimeImmutable
{
    return new DateTimeImmutable();
}
```

or use `Carbon\FactoryImmutable` or something

app/Models/UserTotpKey.php
```diff
@@ -0,0 +1,91 @@
...
+        $totp->setIssuer('osu!');
```
notbakaneko: not really important but could use some env variable so I don't have production and dev authenticator all defaulting to the same name

app/Models/UserTotpKey.php
```diff
@@ -0,0 +1,91 @@
...
+
+        return $totp->getProvisioningUri();
+    }
+
+    public static function isValidKey(string $uri, string $key): bool
+    {
+        return Factory::loadFromProvisioningUri($uri)->verify($key, null, 10);
+    }
+
+    public function user(): BelongsTo
```
notbakaneko, line 67: missing use `Illuminate\Database\Eloquent\Relations\BelongsTo`

app/Models/UserTotpKey.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        return $this->belongsTo(User::class, 'user_id');
+    }
+
+    public function totp()
```
notbakaneko: `: OTPInterface`

nanaya:
Rebased with no change

notbakaneko (changes requested):

app/Models/UserTotpKey.php
nanaya: it's not even used anymore 🤨

app/Libraries/SessionVerification/State.php
```diff
@@ -0,0 +1,79 @@
...
+
+    public function getMethod(): string
+    {
+        $currentMethod = $this->session->getVerificationMethod();
+
+        if ($currentMethod === null) {
+            LoginAttempt::logAttempt(\Request::getClientIp(), $this->user, 'verify');
+
+            // force mail to prevent client without totp support from showing wrong message
+            $currentMethod = is_api_request() && api_version() < 20250818
```
notbakaneko: should probably move all these `api_version()` magic numbers somewhere so we know what they are

notbakaneko: also `is_api_request() && api_version() < 20250818 || $this->user->userTotpKey === null ? 'mail' : 'totp'`?

app/Libraries/SessionVerification/Controller.php
```diff
@@ -34,58 +69,69 @@ public static function initiate()
...
-            if ($mailState === null) {
-                throw new UserVerificationException('expired', true);
+            if ($state->getMethod() === 'totp') {
+                $totp = $state->user->userTotpKey;
+
+                if ($totp === null) {
+                    // erased between verification start and here
+                    return static::mailFallback($state, static::FALLBACK_MODES['totp_gone']);
+                }
+                $state->user->userTotpKey->assertValidKey($key);
```
notbakaneko: `$totp->assertValidKey($key)`

notbakaneko (changes requested):

app/Libraries/SessionVerification/Controller.php
```diff
@@ -12,20 +12,55 @@
...
+            'messageKey' => 'user_verification.errors.totp_gone',
+            'statusCode' => 422,
+        ],
+        'user_initiated' => [
+            'messageKey' => null,
+            'statusCode' => 200,
+        ],
+    ];
+
+    public static function mailFallback(State $state, array $options)
```
notbakaneko, line 26: Missed this earlier, but can't `$options` be just the key string; I don't see anything but `FALLBACK_MODES[key]` being passed in, and the contents of the array aren't being used anywhere else?

nanaya: the idea is the caller passes a const value instead of any string

notbakaneko (approved):

PR #12362, opened 2025-08-22
Add pinned flag & description field to multiplayer rooms

bdach:
- `pinned` is specifically for https://github.com/ppy/osu/issues/34537.
  - I went for the flag rather than a new category of room because categories have stuff like filtering attached to them that I think doesn't make sense if a separate "pinned" room category was to be created.
  - I expect that going forward all spotlights / FA rooms should also specify `pinned = true`. This would affect / require updating ASS automation (cc @peppy).
  - Pinned rooms will be prioritised order-wise when listing rooms using the default `created` sort. This affects both the room listing endpoint (`GET /api/v2/rooms`), as well as the website's user playlist listing (`/users/{user_id}/playlists`). The second is arguable and could be considered as a side effect but I think it's probably fine and not worth creating a separate sort mode for.
- `description` is for currently-unspecified further use.

  @Hiviexd previously asked for it [in terms of daily challenge to describe the 'theme' of the challenge](https://discord.com/channels/90072389919997952/1237133068116955308/1364531588255449108) but I don't know that themed challenges are still a thing. Maybe it could be used for the special pinned rooms.

  I'm reluctant to make this available to the general game populace because as I've learned over time, the general game populace cannot be trusted with a freeform textbox.

### DIFF
--- a/app/Models/Multiplayer/Room.php
+++ b/app/Models/Multiplayer/Room.php
@@ -35,12 +35,14 @@ use Illuminate\Database\Eloquent\SoftDeletes;
  * @property int|null $channel_id
  * @property \Carbon\Carbon|null $created_at
  * @property \Carbon\Carbon|null $deleted_at
+ * @property string|null $description
  * @property \Carbon\Carbon $ends_at
  * @property User $host
  * @property int $id
  * @property int|null $max_attempts
  * @property string $name
  * @property int $participant_count
+ * @property bool $pinned
  * @property \Illuminate\Database\Eloquent\Collection $playlist PlaylistItem
  * @property \Illuminate\Database\Eloquent\Collection $scoreLinks ScoreLink
  * @property-read Season $season
--- a/app/Models/Multiplayer/Room.php
+++ b/app/Models/Multiplayer/Room.php
@@ -94,6 +94,7 @@ class Room extends Model
         'auto_skip' => 'boolean',
         'ends_at' => 'datetime',
         'password' => PresentString::class,
+        'pinned' => 'boolean',
         'starts_at' => 'datetime',
     ];
     protected array $macros = [
--- a/app/Models/Multiplayer/Room.php
+++ b/app/Models/Multiplayer/Room.php
@@ -63,6 +63,7 @@ class Room extends Model
             ['column' => 'id', 'order' => 'DESC', 'type' => 'int'],
         ],
         'created' => [
+            ['column' => 'pinned', 'order' => 'DESC', 'type' => 'bool'],
             ['column' => 'id', 'order' => 'DESC', 'type' => 'int'],
         ],
     ];
--- a/database/migrations/2025_08_21_103748_add_pinned_description_to_multiplayer_rooms.php
+++ b/database/migrations/2025_08_21_103748_add_pinned_description_to_multiplayer_rooms.php
@@ -19,6 +19,7 @@ return new class extends Migration
         Schema::table('multiplayer_rooms', function (Blueprint $table) {
             $table->boolean('pinned')->default(false);
             $table->string('description')->nullable();
+            $table->index('pinned');
         });
     }
 
@@ -28,6 +29,7 @@ return new class extends Migration
     public function down(): void
     {
         Schema::table('multiplayer_rooms', function (Blueprint $table) {
+            $table->dropIndex(['pinned']);
             $table->dropColumn(['pinned', 'description']);
         });
     }
--- a/database/migrations/2025_08_21_103748_add_pinned_description_to_multiplayer_rooms.php
+++ b/database/migrations/2025_08_21_103748_add_pinned_description_to_multiplayer_rooms.php
@@ -1,0 +1,34 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+declare(strict_types=1);
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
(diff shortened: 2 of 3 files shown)


nanaya:
I wonder if it's better for client to make two separate requests, one for pinned categories and another one for the rest...? (although there's currently no way to filter multiple categories)

bdach:
That's missing part of the point which is that there are cases where we want a room to be at the top of the listing that may not fit one of the current categories ("featured artist" / "spotlights"). The recent cases of this include [monthly beatmap contest winners](https://osu.ppy.sh/multiplayer/rooms/1424656) (although this one is *maaaaybe* borderline appropriate) and, more appropriately, [locus](https://osu.ppy.sh/multiplayer/rooms/1539920). Also Lazer Grand Arena (tournament) qualifiers a while back.

Whether that means we want to do "contests" and "tournaments" categories or whatever instead of the pinned flag is up for further discussion which I half-expected.

Hiviexd:
re: description, daily challenge themes are still a thing, albeit rarer than before. would be nice to make descriptions available to daily challenges via ASS tooling.

notbakaneko:
I feel like `pinned` is an odd thing to sort by, more so when it involves cursoring or pagination; shouldn't being pinned also imply it should always be there?

bdach:
> I feel like `pinned` is an odd thing to sort by, more so when it involves cursoring or pagination; shouldn't being pinned also imply it should always be there?

I'm not attached to it being included in the sorting. The only thing I'm looking to do is to fix the current dodgy way that client is using to pin FA / spotlights playlists to the top locally, which basically only works if those playlists are returned on the first page of the listing. If you specify an alternative way for me to achieve the same end result I am happy to oblige.

nanaya:
I think it's fine for now. If there's ever need to split the request/response it shouldn't be too difficult of a change (at least from the web side)

app/Models/Multiplayer/Room.php
```diff
@@ -35,12 +35,14 @@
...
  * @property \Carbon\Carbon|null $created_at
  * @property \Carbon\Carbon|null $deleted_at
+ * @property string|null $description
  * @property \Carbon\Carbon $ends_at
  * @property User $host
  * @property int $id
  * @property int|null $max_attempts
  * @property string $name
  * @property int $participant_count
+ * @property bool $pinned
```
nanaya, line 45: also need to be added to the `$casts` array

nanaya (changes requested):

notbakaneko:
Sorting by `pinned` first would also need an index to cover it, wouldn't it?

database/migrations/2025_08_21_103748_add_pinned_description_to_multiplayer_rooms.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('multiplayer_rooms', function (Blueprint $table) {
+            $table->boolean('pinned')->default(false);
+            $table->string('description')->nullable();
+            $table->index('pinned');
```
nanaya, line 23: it should also add the column to existing indexes otherwise it may not actually be useful...

bdach: You think all five of them or no?

```
  KEY `multiplayer_rooms_user_id_index` (`user_id`),
  KEY `multiplayer_rooms_category_ends_at_index` (`category`,`ends_at`),
  KEY `multiplayer_rooms_category_user_id_index` (`category`,`user_id`),
  KEY `multiplayer_rooms_ends_at_index` (`ends_at`),
  KEY `multiplayer_rooms_type_category_ends_at_index` (`type`,`category`,`ends_at`),
```

It's a bit of a finger-in-the-air type discussion when I don't have a representative db to `EXPLAIN ANALYZE` on...

nanaya: yeah I think so.

bdach: Looking at all the indices again I'm not sure `pinned` makes sense to include in every single index. The existing indices with `ends_at` are pretty clearly there to cover off the `ended` sort mode, which doesn't look at the pinned flag.

With https://github.com/ppy/osu-web/pull/12362/commits/3dadb548a8d310b05a09b12699b0e5fb4c4f5751 the indices go from

```
KEY `multiplayer_rooms_user_id_index` (`user_id`),
KEY `multiplayer_rooms_category_ends_at_index` (`category`,`ends_at`),
KEY `multiplayer_rooms_category_user_id_index` (`category`,`user_id`),
KEY `multiplayer_rooms_ends_at_index` (`ends_at`),
KEY `multiplayer_rooms_type_category_ends_at_index` (`type`,`category`,`ends_at`)
```

to

```
KEY `multiplayer_rooms_category_ends_at_index` (`category`,`ends_at`),
KEY `multiplayer_rooms_ends_at_index` (`ends_at`),
KEY `multiplayer_rooms_type_category_ends_at_index` (`type`,`category`,`ends_at`),
KEY `multiplayer_rooms_pinned_index` (`pinned`),
KEY `multiplayer_rooms_user_id_pinned_index` (`user_id`,`pinned`),
KEY `multiplayer_rooms_category_user_id_pinned_index` (`category`,`user_id`,`pinned`)
```

which I'd hope to be enough.

nanaya: figuring out the indexing has been a bit of hell but at least for `mode=all` (not actually exposed in client but available as api), this index is required (also speeds up other modes etc but most notably that mode)

```
multiplayer_rooms_high (room_id,in_room,user_id,updated_at)
```

brings down the specific query on `multiplayer_rooms` table for the request down from 0.4s to 0.1s with the `pinned` order added (it's 0.1s without the new column)

(it also seems to speed up another queries so that's nice)

nanaya: also fwiw `active` filter does check `ends_at` column so there's that...

peppy:
migrations have been applied ✅

PR #12363, opened 2025-08-25
Scale the text colour on the difficulty badge.

AndyShiue:
Scale text colour on the difficulty badge for maps with really high SR. The below images explain all:

before:
old badges

after:
new badges

I cannot really play maps with SR>6.5 so this is not really useful for me. However it's really cool so I did it :)
I guess directly specifying the `color` attribute of the difficulty badge would not be a good idea but I don't know what's the right way yet. It's so welcome if someone can point out how to fix it.
If this idea eventually become appreciated, it's also fine to bikeshed on how the colour spectrum should be.


bdach:
Judging by the fact that @peppy has historically been on record saying that star rating should basically never be allowed to exceed 10, I don't think we want any of this.

AndyShiue:
Ok then because I just did it for fun

peppy:
This doesn't look bad on paper, but I'm not sure we should be supporting this based on how few maps fit this criteria.

Let's close for now and revisit if it ever becomes relevant.

stanriders:
Keep in mind that with difficulty increasing mods the amount of maps in that range is way bigger. 9+ stars is the norm for the top players in pretty much all rulesets, and while they're a tiny percentage of the playerbase they have the biggest reach aka the rest of the playerbase sees 9+ stars more often

Star rating not exceeding 10 is kinda not possible anyway unless we want to rescale everything as soon as a new hardest map drops

peppy:
Dunno then, I'm not against adding colour if there's valid usage for it.

nanaya (approved):

Givikap120:
@nanaya after this got merged - I think a colors for high SR are just too dim
Look on 9 star, how it's glowing with yellow
When 12 star is dark purple on the black background
I'm supporting the idea, but I think colors should be altered to be much brighter
Something like yellow on 9 to bright red on 10, to hot pink on 11, to blue on 12

AndyShiue:
Currently we reuse the spectrum of the badge background for maps with sr>9. Changing the spectrum will sacrifice this simplicity.

Givikap120:
> Currently we reuse the spectrum of the badge background for maps with sr>9. Changing the spectrum will sacrifice this simplicity.

I don't understand what is the point of simplicity (it's just a numbers in the gradient formula) if colors don't look good enough, as they were created for a different purpose (to be a background, not text color).
I think 9 star colors look way more prestigious, intimidating and aggressive than 12 star colors, what is not good.

AndyShiue:
Good point, so now let's start the bikeshedding :)

Givikap120:
This is my attempt and preserving similar color scheme, but making it more bright and vibrant (the only difference is moving orange to 9.5, and placing red on 10).

Old:
image

New:
image

I still don't sure about purplish blue at 12.5+, I think it looks quite underwhelming, maybe it's better to stop at 12 star pinkish purple.

fbasken:
Having looked at these new high star text colors for months now, I'm really inclined to agree with @Givikap120 , the dull lavender text color for 12.5*+ is both difficult to read, and very underwhelming. The text color should indicate that these are the absolute hardest maps in the game. Admittedly not a huge priority but at some point it would be good to have the color spectrum for this text redesigned.

I agree the simplest fix for now would just be to set the text color for 12.00*+ to stop at the existing 12.00* purple-pink color, but ultimately I think the text color spectrum should terminate at ~12.50* to some sort of bright red/hot pink.

PR #12364, opened 2025-08-26
Updated Beatmap search query parsing

notbakaneko:
Beatmap search query string parsing portion of #12148
Removes all usages of `QueryHelper::queryString`/`simple_query_string` in `BeatmapsetSearch` which isn't suitable for our supported syntax without more contortions and parsing to make it fit, at which point we might as well just parse it ourselves.

- [x] `tokenise` test
- [x] depends on #12345

### DIFF
--- a/app/Libraries/Elasticsearch/QueryHelper.php
+++ b/app/Libraries/Elasticsearch/QueryHelper.php
@@ -40,9 +40,8 @@ class QueryHelper
             $mode = str_starts_with($token, '-') ? 'exclude' : 'include';
             $word = ltrim($token, '-');
             if (str_starts_with($word, '"')) {
-                $phraseStart = ltrim($word, '"');
                 $token = strtok('"');
-                $parts[$mode][] = $phraseStart.' '.$token;
+                $parts[$mode][] = $word.' '.$token.'"';
             } else {
                 $parts[$mode][] = $word;
             }
--- a/app/Libraries/Elasticsearch/QueryHelper.php
+++ b/app/Libraries/Elasticsearch/QueryHelper.php
@@ -27,6 +27,32 @@ class QueryHelper
         throw new Exception('$clause should be associative array or Queryable.');
     }
 
+    public static function tokenize(string $query): array
+    {
+        $parts = [
+            'exclude' => [],
+            'include' => [],
+        ];
+
+        $mode = 'include';
+        $token = strtok($query, ' ');
+        while ($token !== false) {
+            $mode = str_starts_with($token, '-') ? 'exclude' : 'include';
+            $word = ltrim($token, '-');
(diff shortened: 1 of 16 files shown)


app/Libraries/Elasticsearch/QueryHelper.php
```diff
@@ -27,6 +27,31 @@ public static function clauseToArray($clause): array
         throw new Exception('$clause should be associative array or Queryable.');
     }
 
+    public static function tokenize(string $query): array
```
nanaya: I don't think this is the expected result? (also, `tokenise`?)

```
> QueryHelper::tokenize('"hello" "world"')
= [
    "exclude" => [],
    "include" => [
      ""hello" world"",
    ],
  ]
```

app/Libraries/Elasticsearch/QueryHelper.php
```diff
@@ -27,6 +27,31 @@ public static function clauseToArray($clause): array
...
+    {
+        $parts = [
+            'exclude' => [],
+            'include' => [],
+        ];
+
+        $mode = 'include';
+        $token = strtok($query, ' ');
```
nanaya, line 47: should probably at least replace all kind of spaces to this normal space first. Full width space `　` is usually expected to act as normal separator in this context

nanaya (changes requested):
just a quick check

notbakaneko:
Making test updates

app/Libraries/Search/BeatmapsetSearch.php
```diff
@@ -20,48 +20,92 @@
...
+
     public function __construct(?BeatmapsetSearchParams $params = null)
     {
         parent::__construct(
             Beatmapset::esIndexName(),
             $params ?? new BeatmapsetSearchParams(),
             Beatmapset::class
         );
+
+        $this->tokens = QueryHelper::tokenise($params->queryString ?? '');
```
nanaya, line 38: move to the (request) params class?

why most of the logic are in the request params one...

tests/Libraries/Search/BeatmapsetSearch/NegativeBoostTest.php
```diff
@@ -0,0 +1,38 @@
...
+class NegativeBoostTest extends TestCase
+{
+    protected array $defaultExpectedSort = ['_score', 'id'];
+
+    public static function dataProvider(): array
+    {
+        return [
+            [['q' => 'apple'], [2, 1, 0]],
+            [['q' => 'apple -too'], [1, 2, 0]],
+            [['q' => '-too'], [3, 2, 1, 0]], // no effect because all the scores are 0
```
nanaya, line 21: what's with this negative boost thing? shouldn't this just exclude the ones with `too` tag...?

PR #12367, opened 2025-08-27
Fix wiki search completion auto-fill on mouse over

### LINKED ISSUE #12318: Wiki search shouldn't prefill result if the cursor is there and not moving
In the video, let's say I want to search "osu!mania" slowly, but it prefills "osu!monthly" because the cursor is where it was but I wasn't moving it:

https://github.com/user-attachments/assets/e0ed9505-78fe-45bc-83e6-6df5e21116e3

I only see this prefill behavior in Safari, but it prefills only when the cursor is moved and selects the prefilled text so I can keep writing what I intended:

https://github.com/user-attachments/assets/a8253981-54f3-45b9-8545-e9852110ba73

notbakaneko:
Separate select and highlight tracking to behave more like how other autocomplete drop downs handle autocomplete; 
arrow keys select and fill, mouse only highlights. 
 
(I am not doing the faded fill text, or bing's mouse over select on right arrow since I find not being able to move the cursor in the input annoying)

fixes #12318

### DIFF
--- a/resources/css/bem/wiki-search.less
+++ b/resources/css/bem/wiki-search.less
@@ -27,12 +27,12 @@
 
   &__input {
     .reset-input();
-    color: @osu-colour-c1;
+    color: hsl(var(--hsl-c1));
     flex: 1;
     padding: 13px 0 13px @_gutter;
 
     &::placeholder {
-      color: @osu-colour-c2;
+      color: hsl(var(--hsl-c2));
     }
   }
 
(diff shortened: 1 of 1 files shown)


nanaya:
I wonder if mouse hover action should just be completely decoupled from keyboard navigation just like in various autocomplete boxes (at least in firefox and chrome address bar, and github autocompletes).

Also the current one, hover item &rarr; press up/down key (mouse stays) &rarr; move mouse (still in the same item area) &rarr; doesn't reselect the item under the mouse cursor until leaving the item area and back.

nanaya (approved):

PR #12368, opened 2025-08-28
Add matchmaking room type and stats table

smoogipoo:
- Only BanchoBot should be able to create these rooms.
- These rooms start off with a large (50+) number of items in their playlist. osu-server-spectator will likely clean these up when the room is closed.
- The stats table is very low-effort. Will need more tables here when ELO is in place.

### DIFF
--- a/app/Models/Multiplayer/Room.php
+++ b/app/Models/Multiplayer/Room.php
@@ -428,6 +428,11 @@ class Room extends Model
         return $realtimeTypes->contains($this->type);
     }
 
+    public function isMatchmaking()
+    {
+        return $this->type === static::MATCHMAKING_TYPE;
+    }
+
     public function isScoreSubmissionStillAllowed()
     {
         // TODO: move grace period to config or use the beatmap's duration
@@ -667,14 +672,17 @@ class Room extends Model
 
         $playlistItemsCount = count($playlistItems);
 
-        if ($this->isRealtime() && $playlistItemsCount !== 1) {
-            if ($this->type !== static::MATCHMAKING_TYPE) {
-                throw new InvariantException('realtime room must have exactly one playlist item');
-            }
-        }
-
         if ($playlistItemsCount < 1) {
             throw new InvariantException('room must have at least one playlist item');
+        }
+
+        if ($this->isMatchmaking()) {
+            $banchoBotId = $GLOBALS['cfg']['osu']['legacy']['bancho_bot_user_id'];
+            foreach ($playlistItems as $item) {
+                $item->owner_id = $banchoBotId;
+            }
+        } elseif ($this->isRealtime() && $playlistItemsCount !== 1) {
+            throw new InvariantException('realtime room must have exactly one playlist item');
         }
 
         if (mb_strlen($this->name) > 100) {
@@ -763,9 +771,18 @@ class Room extends Model
 
     private function assertHostRoomAllowance()
     {
+        $banchoBotId = $GLOBALS['cfg']['osu']['legacy']['bancho_bot_user_id'];
+
+        if ($this->host->getKey() == $banchoBotId) {
+            // BanchoBot can always create rooms.
+            return;
+        }
+
         $query = static::active()->startedBy($this->host);
 
-        if ($this->isRealtime()) {
+        if ($this->isMatchmaking()) {
+            throw new InvariantException('matchmaking rooms cannot be created');
+        } else if ($this->isRealtime()) {
             $query->whereIn('type', static::REALTIME_TYPES);
             $max = 1;
         } else {
--- a/app/Models/Multiplayer/Room.php
+++ b/app/Models/Multiplayer/Room.php
@@ -74,6 +74,7 @@ class Room extends Model
     ];
 
     const PLAYLIST_TYPE = 'playlists';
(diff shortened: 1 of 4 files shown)


database/migrations/2025_08_26_000000_create_matchmaking_stats_table.php
```diff
@@ -0,0 +1,29 @@
...
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('matchmaking_stats', function (Blueprint $table) {
+            $table->unsignedMediumInteger('user_id')->nullable(false);
+            $table->unsignedMediumInteger('first_placements')->default(0);
```
notbakaneko: are we still using `mediumint` even when it doesn't have to match a pre-existing column size?

smoogipoo: Changed it to `unsignedInteger`. I did this because I saw prior examples, but I found even more recent examples that don't specify medium size, e.g.:

https://github.com/ppy/osu-web/blob/a917eb9ee33f66f204925472615cf3449618e53b/database/migrations/2025_04_17_113658_create_multiplayer_room_events_table.php#L17-L28

smoogipoo:
Just for a sense of expectations here, I may keep expanding on this PR but it should be considered 'complete' at any point. It should be able to merge independently of osu/osu-server-spectator changes/merges.

app/Models/Multiplayer/Room.php
```diff
@@ -74,8 +74,9 @@ class Room extends Model
     ];
 
     const PLAYLIST_TYPE = 'playlists';
+    const MATCHMAKING_TYPE = 'matchmaking';
     const REALTIME_DEFAULT_TYPE = 'head_to_head';
-    const REALTIME_TYPES = ['head_to_head', 'team_versus'];
+    const REALTIME_TYPES = ['head_to_head', 'team_versus', 'matchmaking'];
```
nanaya: maybe just this instead

```suggestion
    const USER_REALTIME_TYPES = ['head_to_head', 'team_versus'];
    const REALTIME_TYPES = [...self::USER_REALTIME_TYPES, self::MATCHMAKING_TYPE];
```

smoogipoo: `osu-server-spectator` was recently refactored and the "standard" naming was approved for these types, so I've gone with that here rather than "user" (https://github.com/ppy/osu-server-spectator/tree/master/osu.Server.Spectator/Hubs/Multiplayer/Standard).

database/migrations/2025_08_26_000000_create_matchmaking_stats_table.php
```diff
@@ -0,0 +1,32 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('matchmaking_stats', function (Blueprint $table) {
```
peppy: `matchmaking_user_stats`

database/migrations/2025_08_26_000000_create_matchmaking_stats_table.php
```diff
@@ -0,0 +1,32 @@
...
+            $table->unsignedInteger('user_id')->nullable(false);
```
peppy: probably should have a `ruleset_id`?

database/migrations/2025_08_26_000000_create_matchmaking_stats_table.php
```diff
@@ -0,0 +1,32 @@
...
+            $table->unsignedInteger('first_placements')->default(0);
```
peppy: as discussed in discord, should have a `score` column for basic score tallying.

i'd also add a standard updated at column in line with other laravel tables for bookkeeping.

nanaya (commented):
seems fine apart of migration files need to have the date adjusted

smoogipoo:
What are their dates expected to be?

nanaya:
after the latest one in master branch

smoogipoo:
Have updated migration dates.

database/migrations/2025_09_04_000002_create_matchmaking_pools_table.php
```diff
@@ -0,0 +1,35 @@
...
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('matchmaking_pools', function (Blueprint $table) {
+            $table->increments('id');
+            $table->smallInteger('ruleset_id');
+            $table->string('name');
+            $table->boolean('active');
+            $table->timestamps();
```
nanaya, line 22: the columns generated by this has no default value so whatever doing the insert will have to set them explicitly

smoogipoo: Is that an issue? These entries are expected to always be created manually.

nanaya: not really, just need to remember to actually set them (it is automatically set when it's done through laravel)

nanaya (approved):

PR #12370, opened 2025-08-29
 Interface to add/remove totp verification 

nanaya:
I'm pretty sure the whole thing needs more words 🤷 (explanation etc)

- [x] depends on #12355

account settings page

setup page 1

setup page 2

setup done

unsetup page

unsetup done

### DIFF
--- a/app/Http/Controllers/UserTotpController.php
+++ b/app/Http/Controllers/UserTotpController.php
@@ -19,7 +19,7 @@ class UserTotpController extends Controller
         parent::__construct();
 
         $this->middleware('auth');
-        $this->middleware('throttle:20,1440,user-totp:');
+        $this->middleware('throttle:60,10,user-totp:');
         $this->middleware('verify-user');
 
         $this->middleware(function ($request, $next) {
--- a/app/Http/Controllers/UserTotpController.php
+++ b/app/Http/Controllers/UserTotpController.php
@@ -103,7 +103,6 @@ class UserTotpController extends Controller
                 $message = osu_trans('user_totp.store.ok');
             } catch (\Throwable $e) {
                 if (is_sql_unique_exception($e)) {
-                    $existingTotpKey = $currentUser->userTotpKey()->first();
                     $message = osu_trans('user_totp.store.existing');
                 } else {
                     throw $e;
--- a/app/Http/Controllers/UserTotpController.php
+++ b/app/Http/Controllers/UserTotpController.php
@@ -27,7 +27,7 @@ class UserTotpController extends Controller
 
             if ($this->currentUser->userTotpKey !== null) {
                 $session = \Session::instance();
-                $session->flash('popup', osu_trans('user_totp.index.already_setup'));
+                $session->flash('popup', osu_trans('user_totp.store.existing'));
                 \Cache::forget($this->cacheKey());
 
                 return ujs_redirect(route('account.edit').'#authenticator-app');
@@ -97,34 +97,29 @@ class UserTotpController extends Controller
 
         $currentUser = \Auth::user();
 
-        $existingTotpKey = $currentUser->userTotpKey;
-        if ($existingTotpKey === null) {
-            if (UserTotpKey::isValidKey($totpUri, $key)) {
-                try {
-                    $totpKey = $currentUser->userTotpKey()->create([
-                        'uri' => $totpUri,
-                    ]);
-                } catch (\Throwable $e) {
-                    if (!is_sql_unique_exception($e)) {
-                        throw $e;
-                    }
-                    $existingTotpKey = $currentUser->userTotpKey()->first();
+        if (UserTotpKey::isValidKey($totpUri, $key)) {
+            try {
+                $totpKey = $currentUser->userTotpKey()->create([
+                    'uri' => $totpUri,
+                ]);
+                $message = osu_trans('user_totp.store.ok');
+            } catch (\Throwable $e) {
+                if (!is_sql_unique_exception($e)) {
+                    throw $e;
                 }
-                \Cache::forget($cacheKey);
-            } else {
-                return response(['form_error' => [
-                    'key' => [osu_trans('user_verification.errors.incorrect_key')],
-                ]], 422);
+                $existingTotpKey = $currentUser->userTotpKey()->first();
+                $message = osu_trans('user_totp.store.existing');
             }
+            \Cache::forget($cacheKey);
+
+            $session->flash('popup', $message);
+
+            return ujs_redirect(route('account.edit').'#authenticator-app');
         }
 
-        $message = $existingTotpKey !== null && $existingTotpKey->uri !== $totpUri
-            // this also handles race condition between key existence check and creation
-            ? osu_trans('user_totp.store.existing')
-            : osu_trans('user_totp.store.ok');
-        $session->flash('popup', $message);
-
-        return ujs_redirect(route('account.edit').'#authenticator-app');
+        return response(['form_error' => [
+            'key' => [osu_trans('user_verification.errors.incorrect_key')],
+        ]], 422);
     }
 
     private function cacheKey(): string
--- a/app/Http/Controllers/UserTotpController.php
+++ b/app/Http/Controllers/UserTotpController.php
@@ -97,7 +97,7 @@ class UserTotpController extends Controller
 
         if (UserTotpKey::isValidKey($totpUri, $key)) {
             try {
-                $totpKey = $currentUser->userTotpKey()->create([
+                $currentUser->userTotpKey()->create([
                     'uri' => $totpUri,
                 ]);
                 $message = osu_trans('user_totp.store.ok');
--- a/app/Http/Controllers/UserTotpController.php
+++ b/app/Http/Controllers/UserTotpController.php
@@ -102,11 +102,12 @@ class UserTotpController extends Controller
                 ]);
                 $message = osu_trans('user_totp.store.ok');
             } catch (\Throwable $e) {
-                if (!is_sql_unique_exception($e)) {
+                if (is_sql_unique_exception($e)) {
+                    $existingTotpKey = $currentUser->userTotpKey()->first();
+                    $message = osu_trans('user_totp.store.existing');
+                } else {
                     throw $e;
                 }
-                $existingTotpKey = $currentUser->userTotpKey()->first();
-                $message = osu_trans('user_totp.store.existing');
             }
             \Cache::forget($cacheKey);
 
--- a/app/Http/Controllers/UserTotpController.php
+++ b/app/Http/Controllers/UserTotpController.php
@@ -19,7 +19,7 @@ class UserTotpController extends Controller
         parent::__construct();
 
         $this->middleware('auth');
-        //$this->middleware('throttle:20,1440,user-totp:');
+        $this->middleware('throttle:20,1440,user-totp:');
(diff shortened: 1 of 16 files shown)


app/Http/Controllers/UserTotpController.php
```diff
@@ -0,0 +1,140 @@
...
+                return response(['form_error' => [
+                    'key' => [osu_trans('user_verification.errors.incorrect_key')],
+                ]], 422);
+            }
+        }
+
+        // this also handles race condition between key existence check and creation
+        if ($existingTotpKey !== null && $existingTotpKey !== $totpUri) {
+            // pretend setup is complete
+            throw new InvariantException('you already have a different totp key setup!');
```
Userlog2:
```suggestion
            throw new InvariantException('you already have a different totp key set up!');
```

resources/lang/en/user_totp.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+return [
+    'create' => [
+        'finish' => 'Finish',
+        'key' => 'Scan the QR code with the authenticator app and enter the verification key',
+        'password' => 'To setup app based two-factor authentication, please enter your current password',
```
Userlog2:
```suggestion
        'password' => 'To set up app based two-factor authentication, please enter your current password',
```

resources/lang/en/user_totp.php
```diff
@@ -0,0 +1,28 @@
...
+        'start' => 'Continue',
+    ],
+
+    'destroy' => [
+        'missing' => 'You don\'t have authenticator app verification setup.',
```
Userlog2:
```suggestion
        'missing' => 'You don\'t have authenticator app verification set up.',
```

resources/lang/en/user_totp.php
```diff
@@ -0,0 +1,28 @@
...
+        'ok' => 'Authenticator app verification removed.',
+    ],
+
+    'edit' => [
+        'password' => 'Please enter your current password to remove verification using authenticator app.',
+        'start' => 'Remove',
+    ],
+
+    'store' => [
+        'ok' => 'Authenticator based verification has been setup',
```
Userlog2:
```suggestion
        'ok' => 'Authenticator-based verification has been set up',
```

resources/lang/en/accounts.php
```diff
@@ -147,6 +147,20 @@
         'update' => 'update',
     ],
 
+    'user_totp' => [
+        'title' => 'Authenticator App',
+
+        'button' => [
+            'remove' => 'Remove',
+            'setup' => 'Setup',
```
Userlog2:
```suggestion
            'setup' => 'Set up',
```

(assumes this is referring to "Set up an authenticator app")

Userlog2 (changes requested):
Minor issue: "To *setup* an app" is incorrect. "To set up an app" is the correct form.
("setup" as one word is correct when "setup" is being used to mean "configuration", not when it's being used to mean "configure").

resources/lang/en/accounts.php
peppy: For this specific instance I'd probably change to "Add authenticator app" or something, rather than "Set up".

resources/views/accounts/_edit_user_totp.blade.php
```diff
@@ -0,0 +1,53 @@
+{{--
+    Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+    See the LICENCE file in the repository root for full licence text.
+--}}
+<div class="account-edit" id="totp">
```
notbakaneko: I think the user-facing side should all use the `authenticator` or `OTP` naming and not the specific type of OTP

app/Http/Controllers/UserTotpController.php
```diff
@@ -0,0 +1,136 @@
...
+        $this->middleware('auth');
+        //$this->middleware('throttle:20,1440,user-totp:');
+        $this->middleware('verify-user');
+
+        $this->middleware(function ($request, $next) {
+            $this->currentUser = \Auth::user();
+
+            if ($this->currentUser->userTotpKey !== null) {
+                $session = \Session::instance();
+                $session->flash('popup', osu_trans('user_totp.index.already_setup'));
```
notbakaneko: missing key

app/Http/Controllers/UserTotpController.php
```diff
@@ -0,0 +1,136 @@
...
+    public function __construct()
+    {
+        parent::__construct();
+
+        $this->middleware('auth');
+        $this->middleware('throttle:60,10,user-totp:');
+        $this->middleware('verify-user');
+
+        $this->middleware(function ($request, $next) {
+            $this->currentUser = \Auth::user();
```
notbakaneko: assigning to non-existent property?

also the other `$currentUser = \Auth::user()`s only to just call a single line...

resources/views/user_totp/create.blade.php
```diff
@@ -0,0 +1,60 @@
...
+                <div class="password-reset__input-group">
+                    <div class="qr-svg">
+                        {!! qr_svg($uri) !!}
+                    </div>
+                </div>
+                <div class="password-reset__input-group">
+                    {{ osu_trans('user_totp.create.key') }}
+                </div>
+                <label class="password-reset__input-group">
+                    <input autofocus class="password-reset__input" name="key">
```
notbakaneko: `autocomplete="one-time-code" inputmode="numeric"`?

nanaya:
Rebased to move the extra backend related changes to the correct pr

resources/lang/en/accounts.php
```diff
@@ -147,6 +147,21 @@
         'update' => 'update',
     ],
 
+    'user_totp' => [
+        'title' => 'Authenticator App',
+        'usage_note' => 'Use authenticator app instead of email for session verification. Fallback to email verification is still possible. Verification using authenticator app is not yet available for osu!(lazer) - it will still be done through email.',
```
notbakaneko: `Use authenticator app instead of email for verification. Email verification will still be available as a fallback/alternative. ...` I don't think we mention session in any of the verification text anywhere.

notbakaneko (changes requested):

app/Http/Controllers/UserTotpController.php
```diff
@@ -0,0 +1,129 @@
...
+            $session->flash('popup', osu_trans('user_totp.store.restart'));
+
+            return ujs_redirect(route('authenticator-app.create'));
+        }
+
+        $currentUser = \Auth::user();
+
+        if (UserTotpKey::isValidKey($totpUri, $key)) {
+            try {
+                $totpKey = $currentUser->userTotpKey()->create([
```
notbakaneko: unused variable

app/Http/Controllers/UserTotpController.php
```diff
@@ -0,0 +1,129 @@
...
+                    'uri' => $totpUri,
+                ]);
+                $message = osu_trans('user_totp.store.ok');
+            } catch (\Throwable $e) {
+                if (!is_sql_unique_exception($e)) {
+                    throw $e;
+                }
+                $existingTotpKey = $currentUser->userTotpKey()->first();
+                $message = osu_trans('user_totp.store.existing');
```
notbakaneko: is this missing an `if` or something? Never mind that the unique constraint exception means if should already exist unless there's another race that deleted it after trying to create it 🤔

app/Http/Controllers/UserTotpController.php
```diff
@@ -0,0 +1,129 @@
...
+                'password' => [osu_trans('user_totp.issue_uri.invalid_password')],
+            ]], 422);
+        }
+
+        \Cache::put($this->cacheKey(), UserTotpKey::generateUri($currentUser), static::TIMEOUT);
+
+        return ujs_redirect(route('authenticator-app.create'));
+    }
+
+    public function store()
```
notbakaneko, line 81: wonder if setting up should have a test as well...

notbakaneko (changes requested):
It occurs to me there's no way to set it up on mobile from the device with the authenticator app and camera for scanning the qr code 🤔

app/Http/Controllers/UserTotpController.php
nanaya: the if is right above this part

notbakaneko: what is `->first()` and assigning to variable for? 🤔

nanaya: that was some leftover...

resources/views/user_totp/create.blade.php
```diff
@@ -0,0 +1,69 @@
...
+                method="POST"
+                data-remote
+                data-skip-ajax-error-popup="1"
+            >
+                <div class="password-reset__input-group">
+                    <div class="qr-svg">
+                        {!! qr_svg($uri) !!}
+                    </div>
+                    <div>
+                        <a href="{{ $uri }}">{{ osu_trans('user_totp.create.key_link') }}</a>
```
notbakaneko, line 26: are there any sites that actually let you copy the link? 🤔 
(I'm also not sure how well this works for apps that aren't the default password manager on iOS since they want a code instead)

nanaya: I'd rather tell to take screenshot of the page and pass it to the app instead of explaining all the parameters

notbakaneko: I don't know if authenticators even let you use screenshots instead of the camera... (is it time to print out the qr code and put it on a wooden table?)

notbakaneko: problem for the future™️

notbakaneko (approved):

resources/views/user_totp/create.blade.php
Userlog2: > I don't know if authenticators even let you use screenshots instead of the camera...

Mine (Google's app) doesn't let you do that. Only QR or manual entry.

app/Http/Controllers/UserTotpController.php
```diff
@@ -0,0 +1,129 @@
...
+    {
+        return ext_view('user_totp.create', [
+            'uri' => \Cache::get($this->cacheKey()),
+        ]);
+    }
+
+    public function destroy(): Response
+    {
+        $currentUser = \Auth::user();
+        $currentUser->userTotpKey?->delete();
```
notbakaneko:
```suggestion
        \Auth::user()->userTotpKey?->delete();
```

app/Http/Controllers/UserTotpController.php
```diff
@@ -0,0 +1,129 @@
...
+
+        \Session::flash('popup', osu_trans('user_totp.destroy.ok'));
+
+        return ujs_redirect(route('account.edit').'#authenticator-app');
+    }
+
+    public function edit(): Response
+    {
+        $currentUser = \Auth::user();
+        if ($currentUser->userTotpKey === null) {
```
notbakaneko:
```suggestion
        if (\Auth::user()->userTotpKey === null) {
```

?

notbakaneko (changes requested):
before I forget again

notbakaneko (approved):

PR #12374, opened 2025-09-01
Ignore tag for topic title length check

### LINKED ISSUE #12353: Allow Forum Tags to Exceed Maximum Forum Title Length
When adding forum tags to certain posts with long titles, the tag cannot be added due to the forum title maximum length being reached. I was wondering if injecting the tags could supercede this rather than just increasing the forum title character count. 

My reasoning behind this is if we allow for more characters to be used then they will just be used up with people making the titles longer and longer until it reaches the maximum count.

nanaya:
Resolves #12353

- [ ] fix table charset on prod (not adding this to migration)
  - `alter table phpbb_topics charset=utf8mb4 collate=utf8mb4_0900_ai_ci;`
  - I think most of other tables' collation has different (older) one but this one is the latest mysql default

### DIFF
--- a/app/Models/Forum/Topic.php
+++ b/app/Models/Forum/Topic.php
@@ -116,7 +116,7 @@ class Topic extends Model implements AfterCommit
     ];
 
     const MAX_FIELD_LENGTHS = [
-        'topic_title' => 100,
+        'topic_title' => 255, // matching db, there's separate validation for title_normalised
     ];
 
     const VIEW_COUNT_INTERVAL = 86400; // 1 day
@@ -251,6 +251,11 @@ class Topic extends Model implements AfterCommit
         return $this->attributes['poll_length'] / 86400;
     }
 
+    public function getTitleNormalisedAttribute(): string
+    {
+        return $this->titleNormalized();
+    }
+
     public function getTopicFirstPosterColourAttribute($value)
     {
         if (present($value)) {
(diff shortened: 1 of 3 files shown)


notbakaneko:
Doesn't that still leave the existing columns in the old charset and collation?

nanaya:
the type change also resets the charset etc (which failed the migration for me because the table charset is mb3 and the title column has row with emoji)

(most of the columns are already explicitly set to mb4)

notbakaneko (approved):

nanaya:
@peppy please check the alter table above. The column change migration itself should be quick...

```sql
-- table charset update (no migration entry)
alter table phpbb_topics charset=utf8mb4 collate=utf8mb4_0900_ai_ci;

-- migration entry 2025_09_01_140349_adjust_topic_title_length
alter table `phpbb_topics` modify `topic_title` varchar(255) not null default '';
```

peppy:
```sql
Database changed
mysql> alter table phpbb_topics charset=utf8mb4 collate=utf8mb4_0900_ai_ci;
Query OK, 0 rows affected (0.10 sec)
Records: 0  Duplicates: 0  Warnings: 0

mysql> alter table `phpbb_topics` modify `topic_title` varchar(255) not null default '';
Query OK, 0 rows affected (0.01 sec)
Records: 0  Duplicates: 0  Warnings: 0
```

mysql smart

added migration entry too

PR #12375, opened 2025-09-03
Add beatmapset version and related models

nanaya:
Just the models. Mainly so it's easier to check the structures and docs (which I don't use).

The tables are already there.


bdach (approved):

notbakaneko (approved):
looked at this 2 weeks ago 😅

PR #12377, opened 2025-09-04
Add `variant` column to `matchmaking_pools`

smoogipoo:


database/migrations/2025_09_04_000004_add_variant_to_matchmaking_pools.php
```diff
@@ -0,0 +1,31 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('matchmaking_pools', function (Blueprint $table) {
+            $table->smallInteger('variant')->after('ruleset_id')->default(0);
```
nanaya: `variant_id`? currently in web it's only expressed as string though - `4k`/`7k`

smoogipoo: Renamed, though I think int is better? Dunno if we need to make much consideration for custom rulesets that may want strings, at this point.

smoogipoo:
Also adjusted the index, it probably doesn't matter at the end of the day though so let me know if it should just be dropped instead (I don't expect this to have more than 100 rows any time soon).

database/migrations/2025_09_04_000004_add_variant_to_matchmaking_pools.php
nanaya: what are the values here? 0 means... no variant? and then 1 and 2 for 4k and 7k?

smoogipoo:
@nanaya I'm going to hijack this PR a bit...

image

:upside_down_face:

database/migrations/2025_09_04_000004_add_variant_to_matchmaking_pools.php
smoogipoo: Yeah I guess you could think of 0 as no variant, or the default variant. Is using nullables preferred for DB stuff? It's a bit easier from o-s-s this way where a value is always provided.

For mania it'd be 4 and 7.

smoogipoo: image

database/migrations/2025_09_04_000004_adjust_matchmaking_tables.php
```diff
@@ -0,0 +1,47 @@
...
+        Schema::table('matchmaking_pools', function (Blueprint $table) {
+            $table->smallInteger('variant_id')->after('ruleset_id')->default(0);
+
+            $table->dropIndex(['ruleset_id', 'active']);
+            $table->index(['ruleset_id', 'variant_id', 'active']);
+        });
+
+        Schema::table('matchmaking_pool_beatmaps', function (Blueprint $table) {
+            $table->dropColumn('beatmap_id');
+            $table->unsignedInteger('beatmap_id')->after('pool_id');
```
nanaya: no need to drop the column
```suggestion
            $table->unsignedMediumInteger('beatmap_id')->change();
```
(and apparently it's currently mediumint everywhere)

(also for the `down` one)

smoogipoo: Without the drop columns I get:

```
SQLSTATE[42S21]: Column already exists: 1060 Duplicate column name 'beatmap_id' (Connection: mysql, SQL: alter table `matchmaking_pool_beatmaps` add `beatmap_id` int unsigned not null after `pool_id`)
```

nanaya: you need the `->change()` at the end

smoogipoo: Oh, I missed the `->change()`.

smoogipoo: Should be done now.

nanaya (approved):

database/migrations/2025_09_04_000004_adjust_matchmaking_tables.php
```diff
@@ -0,0 +1,45 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('matchmaking_pools', function (Blueprint $table) {
+            $table->smallInteger('variant_id')->after('ruleset_id')->default(0);
```
peppy, line 18: I don't agree with this, unless there's another table storing variants...

Should be called `variant` and likely be nullable (although see below, can be non-nullable with `0` default if it works better). Adding an `_id` suffix suggests it's referencing something else.

Following client side example implementations:

https://github.com/ppy/osu/blob/f8830c6850128456266c82de83273204f8b74ac0/osu.Game/Configuration/RealmRulesetSetting.cs#L14-L15

https://github.com/ppy/osu/blob/2a0e4c364d11c681b4a8ada723f56b0f1687666d/osu.Game/Input/Bindings/RealmKeyBinding.cs#L22

nanaya: osu-web refers 4k and 7k for "variant". and looking at how playmode/ruleset/ruleset_id went (is it 0 or osu or osu!), I'd call it id unless that's how it's displayed (table/display name)

nanaya: also if we're adding more variants (please don't) I doubt int would work for its naming apart of for more mania key count...

peppy: i don't foresee it being used ever again, and as discussed, if anything we'd split mania into multiple rulesets and nuke this from existence.

will just go with what we have as i don't have the energy 🤷

PR #12378, opened 2025-09-04
Add strict to migration files and create better default

nanaya:
Initially just creating the better defaults but then figured might as well update all the migrations while at it.

<del>Moved it to before licence because that's what the linter supports 🤔</del> Back to the original place. I don't suppose we want to use it blindly.

Alternatively maybe change the licence to docblock comment instead?

### DIFF
--- a/database/migrations/2015_01_01_133337_base_tables.php
+++ b/database/migrations/2015_01_01_133337_base_tables.php
@@ -2,6 +2,9 @@
 
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
 // See the LICENCE file in the repository root for full licence text.
+
+declare(strict_types=1);
+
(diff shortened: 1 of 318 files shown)


notbakaneko:
I've noticed it being inlined as just `<?php declare(strict_types=1)`, dunno if that's cleaner...

nanaya:
maybe but there doesn't seem any way to do that

and psr-12 explicitly disallows that anyway (except for one-liner with closing php tag) https://www.php-fig.org/psr/psr-12/#3-declare-statements-namespace-and-import-statements

nanaya:
actually, apparently file-level docblock comment style is the one that's supposed to be used for licence header...?

that said, the linter accepts the existing style (licence then declare), it just doesn't add it at the correct place when auto fixing it but I suppose we don't want it to be automatically fixed anyway?

notbakaneko (approved):

PR #12379, opened 2025-09-05
Combine beatmapset related routes

nanaya:
I noticed this when trying to add a new route under beatmapset. Dunno why they're not prefixed and all 🤷


notbakaneko (approved):
I should have sworn I had a branch with something like this somewhere 🤔

PR #12380, opened 2025-09-05
Change `matchmaking_pool_beatmaps` primary to index

smoogipoo:


nanaya (commented):
so there's no unique constraint at all for this table?

smoogipoo:
This table should have no unique constraints.

Do you have any feelings/preferences/standards towards order? I don't think we particularly care about order but maybe timestamps are appropriate (even if unused)?

nanaya:
I have no idea how this is used so there's nothing much I can say...

smoogipoo:
Then its current state is enough for our usages. It's just a one-to-many mapping from `matchmaking_pools`.

database/migrations/2025_09_05_000000_remove_matchmaking_pool_beatmaps_pk.php
```diff
@@ -0,0 +1,29 @@
...
+        Schema::table('matchmaking_pool_beatmaps', function (Blueprint $table) {
+            $table->dropPrimary('pool_id');
+            $table->index('pool_id');
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
```
nanaya, line 26: I was about to let this pass but noticed if I do a rollback (like on dev) and then rerun the migration, it'll fail due to missing primary key to be dropped and duplicated index to be created

nanaya (changes requested):

database/migrations/2025_09_05_000000_remove_matchmaking_pool_beatmaps_pk.php
smoogipoo: I'm not sure how to do this one... The simple:

```php
        Schema::table('matchmaking_pool_beatmaps', function (Blueprint $table) {
            $table->dropIndex('pool_id');
            $table->primary('pool_id');
        });
```

Results in:

```
SQLSTATE[42000]: Syntax error or access violation: 1091 Can't DROP 'pool_id'; check that column/key exists (Connection: mysql, SQL: alter table `matchmaking_pool_beatmaps` drop index `pool_id`)
```

Plus, adding the primary would result in a potential PK violation when rolling back with items in the table, so I thought just not doing anything would be better...

Would appreciate guidance.

smoogipoo: Hmm, it works if the index is `['pool_id']` (array) instead.

nanaya: it's one weird laravel thing but you need to put the column name in an array when dropping (but not when creating even though it's accepted as well)

```
$table->dropIndex(['pool_id']);
```

yeah if the table is populated then it's a bit weird but if you create the primary key before dropping the index, they can fix the content and then rerun the migration

database/migrations/2025_09_05_000000_remove_matchmaking_pool_beatmaps_pk.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+        Schema::table('matchmaking_pool_beatmaps', function (Blueprint $table) {
+            $table->dropIndex(['pool_id']);
+            $table->primary('pool_id');
```
nanaya, line 30: the previous one creating primary key first was better - as I mentioned before, an early failed primary key creation means the migration can be rerun (re-rolled back) but if the index is dropped first and then failed at creating the primary key, subsequent rerun will fail dropping the nonexistent index 😅

PR #12381, opened 2025-09-05
Add minimum threshold for displaying beatmap user tags

### LINKED ISSUE #12015: Adjust user tag display thresholds
There needs to be some minimum votes before the tag is displayed and the maximum number of tags is currently a bit high.

The exact values for either are TBD 🤷 (or maybe just make it configurable)

notbakaneko:
The search index doesn't include the count which needs a new schema so I'm excluding that part and making it filter out on index for now.

I feel like there should be a better place to pass the config variables but I just put it there unless there's a better suggestion 🤷 

closes #12015


Userlog2:
Agree with 5 votes being the threshold - 10 is too high

config/osu.php
```diff
@@ -206,7 +206,7 @@
     ],
     'tags' => [
         'beatmap_tags_cache_duration' => 60 * (get_int(env('BEATMAP_TAGS_CACHE_DURATION')) ?? 60), // in minutes, converted to seconds
-        'tags_cache_duration' => 60 * (get_int(env('TAGS_CACHE_DURATION')) ?? 60), // in minutes, converted to seconds
+        'min_votes_display' => get_int(env('BEATMAP_TAGS_MIN_VOTES_DISPLAY') ?? 5),
```
nanaya: so the first one is `beatmap_tags_` prefixed, both the config name and the env var, meanwhile the new config is `beatmap_tags_` prefixed but only the env. (there's also the top tag count with just `beatmap_` prefix)

maybe make it a bit more consistent...? (and maybe rename the parent key to `beatmap_tags` and remove the prefix from the config name)

also the `?? 5` should be outside the `get_int` otherwise it may return `null`

app/Models/Beatmap.php
```diff
@@ -406,6 +406,10 @@ public function slowTopTagIds(): array
                 ? $a['tag_id'] - $b['tag_id']
                 : $b['count'] - $a['count']);
 
+            // slowTopTagIds is only used by indexing so it should be fine to filter out tags under the threshold for now.
+            $minVotes = $GLOBALS['cfg']['osu']['tags']['min_votes_display'];
+            $countById = array_filter($countById, fn ($count) => $count >= $minVotes);
```
nanaya: filtering before sorting is probably more useful

app/Models/Beatmap.php
```diff
@@ -402,11 +402,15 @@ public function slowTopTagIds(): array
                 $countById[$vote->tag_id] ??= ['tag_id' => $vote->tag_id, 'count' => 0];
                 $countById[$vote->tag_id]['count']++;
             }
+            // slowTopTagIds is only used by indexing so it should be fine to filter out tags under the threshold for now.
+            $minVotes = $GLOBALS['cfg']['osu']['tags']['min_votes_display'];
```
nanaya: um

(can use a name update)

notbakaneko: did I not save it 🤔

PR #12383, opened 2025-09-08
Add ELO data column to `matchmaking_user_stats`

smoogipoo:
The data here is varying, but in general it will look like this:

```json
{
  "contest_count": 0,
  "normal_factor":
  {
    "mu": 1500.0,
    "sig": 350.0
  },
  "logistic_factors": [],
  "last_contest_time": null,
  "approximate_posterior":
  {
    "mu": 1300.0,
    "sig": 350.0
  }
}
```

I don't think we will, but if we ever want to display the user's ELO, we can use `approximate_posterior->mu ± (3 * (approximate_posterior.sig - 80))`.

`logistic_factors` will contain up to 30 members each with the signature:

```json
{
  "mu" : 0.0,
  "warg": 0.0,
  "wout": 0.0
}
```


database/migrations/2025_09_08_000000_add_matchmaking_elo_columns.php
```diff
@@ -0,0 +1,33 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('matchmaking_user_stats', function (Blueprint $table) {
```
hburn7, line 19: Would this table need refactoring should a competitive option be added in the future? Maybe `quickplay_user_stats` is ideal? See https://github.com/ppy/osu-server-spectator/pull/313#discussion_r2334667598 for similar thoughts.

smoogipoo: I think we can do that when we get to that point, I feel like I've already abused osu!web at this point with changes to these tables 😅

PR #12397, opened 2025-09-17
Update android requirement to 9+ for osu!(lazer)

Joehuu:
Companion to https://github.com/ppy/osu/pull/35043.


smoogipoo:
There have been further developments in the issue thread so closing this for now.

PR #12401, opened 2025-09-18
Extend signature editor size

### LINKED ISSUE #8618: Signature Editor Size too small
image
image

The signature editor should be the exact same size as the size of the signature on the forums, else its quite useless for previewing signatures that are more than just some text or a single image

nanaya:
Fill up the entire section width (to its right) as nothing else is using the specified width and it can use the extra width.

I suppose this resolves #8618. It's probably as wide as it can get and as I mentioned there, there's no such thing as signature size (or at least the width).


notbakaneko (approved):
I'm not sure if it looks sillier if it extends across the whole width, or only up to the right side and a gap on the left 🤷

PR #12402, opened 2025-09-18
Wrap long mods list in score detail

### LINKED ISSUE #12348: More than 10+ mods combination in mobile version will break the screen.
Seems fine if enabled the Desktop Version. 
But in Mobile Version, it's breaking the screen.

https://github.com/user-attachments/assets/2ff76888-23cb-4620-9a31-9dedf78b08ae

notbakaneko:
fixes #12348

Adds a wrapper than (ab)uses `display: contents` to wrap the mods which keeping the rest of the layout.
The other option was to render another copy of pp in the mods list...

### DIFF
--- a/resources/css/bem/play-detail.less
+++ b/resources/css/bem/play-detail.less
@@ -141,7 +141,6 @@
   &__mods {
     display: flex;
     margin-top: 10px;
-    padding-right: 0;
     flex-wrap: wrap;
 
     @media @desktop {
@@ -154,9 +153,10 @@
     }
   }
 
-  &__mods_pp {
+  &__mods-pp {
     display: flex;
     justify-content: space-between;
+    gap: 5px;
 
     @media @desktop {
       display: contents;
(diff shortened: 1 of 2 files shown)


resources/css/bem/play-detail.less
```diff
@@ -139,7 +139,28 @@
...
+      max-width: 400px; // avoid squashing too much of the left side
+      align-items: center;
+      padding: @_padding-vertical-desktop 10px;
+      margin-top: 0;
+      order: -1;
+      background-color: var(--bg-main);
+    }
+  }
+
+  &__mods_pp {
```
nanaya: `mods-pp`?

resources/css/bem/play-detail.less
```diff
@@ -139,7 +139,28 @@
...
+    display: flex;
+    justify-content: space-between;
```
nanaya, line 158: can probably use a `gap`

resources/css/bem/play-detail.less
```diff
@@ -139,7 +139,28 @@
   }
 
   &__mods {
-    flex: 1;
+    display: flex;
+    margin-top: 10px;
+    padding-right: 0;
```
nanaya: not needed?

notbakaneko:
oh I didn't push it 🙈

nanaya (approved):

PR #12405, opened 2025-09-18
Sort score mods in profile page

### LINKED ISSUE #12404: Inconsistent mod ordering between categories on the profile page (only for some scores)
_Initially mentioned in https://github.com/ppy/osu-web/issues/11839, but probably makes more sense as a new issue, before causing confusion._

Noticed while taking a look at https://github.com/ppy/osu-web/issues/12348.
Check the green mods order mixing up with the red ones.

At first, I thought it was related to only the Lazer mode being enabled, but it happens even when disabling it.

Can be seen on the pinned scores of @Aikyuu-Chan's [profile](https://osu.ppy.sh/users/3932190) (with scores from ~3 years ago)

Image

A more recent score has the correct ordering (with customization taking place)... so not really sure what's happening 🤷🏻‍♂️ 
Maybe it's related to "classic"/non-lazer scores?
(Also check the "Spun Out" mod position)

Image

nanaya:
Resolves #12404. Regressed when removing the classic option from the page.

<del>The function arguments don't feel quite right though...</del> Maybe just extend the original function instead.


notbakaneko (approved):

notbakaneko (changes requested):
now it conflicts

notbakaneko (approved):

PR #12408, opened 2025-09-19
Fix own maps search

### LINKED ISSUE #11926: map won't count towards "my maps" for some users in collabs
not sure how to word this properly but just as an example, me (velamy) and meiqth have a couple collabs ranked but using the "My Maps" filter on the website gives inconsistent results, and we arent able to see the same map with this filter (the o i i a map should be the second one in meiqth's image, while neppa does not show up for me at all)

Image

Image

Image

Image

nanaya:
I think this was missed when the multi owner thing was added. Fixes #11926.


notbakaneko (approved):

notbakaneko:
I feel like the `beatmapOwner` isn't being filled in on upload somehow 🤔

PR #12411, opened 2025-09-19
Fix missing translation

nanaya:
I think it's safe to reuse this one...?


notbakaneko (approved):
I like how putting in the wrong password is the one thing I didn't do 🙈

Issue #12412, opened 2025-09-19
Search always shows 's at the end of strings in Russian


und3rlife:
![Image](https://github.com/user-attachments/assets/8bb1efe4-c0e8-4bfd-871f-d5c756a1ad05)

notbakaneko:
this should be fixed in #12395 (and #12413 ...and the localisations updated) when it's deployed

PR #12413, opened 2025-09-19
Fix placeholder shown in quick search when no results

notbakaneko:
Change the key, otherwise un-updated localisations show the `:modes` placeholder instead. #12395

Noticed while checking #12412


nanaya (approved):
it's more like a lang.js bug - any string parent key is returned immediately and never actually traversed to the end

```
window.Lang.get('quick_search.result.empty_for.asdad')
":modes の結果なし" 
```

PR #12414, opened 2025-09-21
Fix incorrect translation key

nanaya:
...actually tested this time.


notbakaneko (approved):
...did I not save the file

PR #12417, opened 2025-09-22
Cache json-formatted medals data

nanaya:
I don't think there's anything user-specific in here 🤔 

<del>...or there may be problem if adding the achieved count</del> any attempts at passing fresh data ends up being very annoying so might as well use potentially stale data instead


notbakaneko (approved):
I'd hope there isn't anything user-specific in there 🤔

PR #12418, opened 2025-09-23
Allow pulling active matches in /matches endpoint

### LINKED ISSUE #11636: Allow pulling only open multiplayer lobbies from ``/matches`` endpoint

someonecantcode:
Closes #11636 

Gives the option for users to get active rooms via active query by checking if the `end_time` is null. When active query is false, it does the same default behavior of returning active & unactive rooms like when normally getting /matches. It wasn't really clear in the issue if when `active=false`, it should show inactive rooms, so I just made it have the same behavior of showing active & unactive matches.

Do I need to add or do anything for this page also? 
link: https://osu.ppy.sh/docs/index.html#matches

endpoint examples:
```bash
/matches?active=1  
/matches?active=true
```

api examples
```bash
teawsl@egghead:~/osu-web$ curl -H "Authorization: Bearer $TOKEN"   \ --get http://localhost:8080/api/v2/matches
{
  "matches": [
    {
      "id": 2,
      "start_time": "2025-09-23T00:52:33+00:00",
      "end_time": "2025-09-23T01:47:33+00:00",
      "name": "inactive 1"
    },
    {
      "id": 1,
      "start_time": "2025-09-23T01:42:33+00:00",
      "end_time": null,
      "name": "active 1"
    }
  ],
  "params": {
    "limit": 50,
    "sort": "id_desc",
    "active": null
  },
  "cursor": null,
  "cursor_string": null
}

teawsl@egghead:~/osu-web$ curl -H "Authorization: Bearer $TOKEN"  \ --get  http://localhost:8080/api/v2/matches?active=1
{
  "matches": [
    {
      "id": 1,
      "start_time": "2025-09-23T01:42:33+00:00",
      "end_time": null,
      "name": "active 1"
    }
  ],
  "params": {
    "limit": 50,
    "sort": "id_desc",
    "active": true
  },
  "cursor": null,
  "cursor_string": null
}


```

### DIFF
--- a/app/Libraries/Search/BeatmapsetSearch.php
+++ b/app/Libraries/Search/BeatmapsetSearch.php
@@ -17,7 +17,6 @@ use App\Models\Solo;
 use App\Models\Tag;
 use App\Models\User;
 use Ds\Set;
-use Illuminate\Database\Eloquent\Builder;
 
 class BeatmapsetSearch extends RecordSearch
 {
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/LegacyMatchesController.php
```diff
@@ -65,16 +65,22 @@ public function index()
         $params = request()->all();
         $limit = \Number::clamp(get_int($params['limit'] ?? null) ?? 50, 1, 50);
         $cursorHelper = LegacyMatch::makeDbCursorHelper($params['sort'] ?? null);
+        $active = get_bool($params['active'] ?? null);
 
         [$matches, $hasMore] = LegacyMatch
             ::where('private', false)
+            ->when(request('active'), fn($q) => $q->whereNull('end_time'))
```
nanaya: use the `$active` variable and do something when `active=0` is explicitly passed instead of still returning active rooms... (I don't think the current one makes sense)

someonecantcode: alright, ty for the clarification. i'll use the `$active` and add the behavior for `active=0` to only pull inactive rooms.

PR #12426, opened 2025-09-30
Allow username capitalization change

godkama:
This is not a big need, but the only way to change username capitalization so far has been only to message support (i think?) or to change username twice (temporary username and then other capitalization). This needs to be checked further since i'm not the best at Laravel but should work!


notbakaneko:
Why are you mixing unrelated PR contents?

godkama:
> Why are you mixing unrelated PR contents?

thats mb i made pr with my main branch and sometimes forget to switch branches between changes. no clue how to go back on those commits on the web tho

godkama:
i'll git rebase when home

godkama:
reopening with correct branch

PR #12427, opened 2025-10-01
Add documentation to /matches for added query params

someonecantcode:
Added documentation on the `active` parameter for the `/matches` endpoint to cover changes in #12418. 

Apologies for not covering this beforehand in the previous pr.

### DIFF
--- a/app/Http/Controllers/LegacyMatchesController.php
+++ b/app/Http/Controllers/LegacyMatchesController.php
@@ -41,7 +41,7 @@ class LegacyMatchesController extends Controller
      * @usesCursor
      * @queryParam limit integer Maximum number of matches (50 default, 1 minimum, 50 maximum). No-example
      * @queryParam sort string `id_desc` for newest first; `id_asc` for oldest first. Defaults to `id_desc`. No-example
-     * @queryParam active boolean `true` for active matches only; `false` for inactive matches; Defaults to `null` returning both. No-example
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/LegacyMatchesController.php
```diff
@@ -36,10 +36,12 @@ public function __construct()
      * matches       | [Match](#match)[]             | |
      * params.limit  | integer                       | |
      * params.sort   | string                        | |
+     * params.active | boolean                       | |
      *
      * @usesCursor
      * @queryParam limit integer Maximum number of matches (50 default, 1 minimum, 50 maximum). No-example
      * @queryParam sort string `id_desc` for newest first; `id_asc` for oldest first. Defaults to `id_desc`. No-example
+     * @queryParam active boolean `true` for active matches only; `false` for inactive matches; Defaults to `null` returning both. No-example
```
nanaya: while passing string `null` works, it's not the correct value for the parameter. It should be either blank or not specified.

nanaya (changes requested):

app/Http/Controllers/LegacyMatchesController.php
someonecantcode: Ok, changed to defaults to not specified.

I am exceeding 120 chars phpcs psr-2 standard but I hope that is ok.

someonecantcode:
Something weird is happening with the tests when I rebased. I'm going to close this to pr and send another on a clean branch with the same changes.

PR #12428, opened 2025-10-01
Reduce chance of model factory collision

nanaya:
I tried some fancier ideas (generating actual random bytes) but it just doesn't work quite right without creating the file ahead of time (or storing the data in the model itself but that'd instead mean having test-specific things in the model).

Maybe I'll do it if this pr ends up not being sufficient.


notbakaneko:
Was there anything relying on the test file being all `a`s 🤔

nanaya:
<del>nope, it's just for correctness and that there's a file at all</del>

...actually one of the tests test the file hash. it doesn't need to be all `a` as long it's valid.

notbakaneko (approved):

Issue #8888, opened 2022-05-10
Better tracking of API usage


peppy:
Right now we don't have enough visibility over API usage. It would be good to have this tracked in the database, against tokens.

See `osu_apikeys` for reference, which does similar tracking for the v1 API.

PR #12430, opened 2025-10-01
Record token usage count

nanaya:
Dunno if it's enough for #8888 🤷 

This also currently tracks api usage from client

(maybe also the miss count at another time...?)

### DIFF
--- a/app/Models/Model.php
+++ b/app/Models/Model.php
@@ -19,7 +19,7 @@ use Illuminate\Database\Eloquent\Model as BaseModel;
 
 abstract class Model extends BaseModel
 {
-    use HasFactory, Traits\FasterAttributes, Validatable;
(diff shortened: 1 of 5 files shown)


app/Models/Model.php
```diff
@@ -19,7 +19,7 @@
 
 abstract class Model extends BaseModel
 {
-    use HasFactory, Traits\FasterAttributes, Validatable;
+    use HasFactory, Traits\FasterAttributes, Traits\IncrementInstance, Validatable;
```
nanaya, line 22: not very related but one of the more recent coding styles says to import one trait at a time

> Each individual trait that is imported into a class MUST be included one-per-line and each inclusion MUST have its own `use` import statement.

https://www.php-fig.org/per/coding-style/#42-using-traits

notbakaneko (approved):

PR #12435, opened 2025-10-02
Hide realtime events with invalid playlist item id

nanaya:
Either this or make a placeholder playlist item...

### DIFF
--- a/app/Http/Controllers/Multiplayer/RoomsController.php
+++ b/app/Http/Controllers/Multiplayer/RoomsController.php
@@ -51,12 +51,16 @@ class RoomsController extends Controller
         ], ['null_missing' => true]);
         $params['limit'] = \Number::clamp($params['limit'] ?? 100, 1, 101);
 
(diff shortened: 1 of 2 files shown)


bdach:
What prompted this? I thought I fixed playlist items with associated events getting deleted... (https://github.com/ppy/osu-server-spectator/pull/291)

nanaya:
I guess the old broken one(s?) was never deleted...

bdach:
That may be it yeah. So long as there are no new orphaned rows after that PR's deploy date (August 1), maybe could "just" clean up the orphans and call it a day?

nanaya:
that could work too yeah

nanaya:
here's the query to clean them up I guess

```sql
delete from multiplayer_realtime_room_events re where playlist_item_id is not null and not exists (select 1 from multiplayer_playlist_items pi where re.playlist_item_id = pi.id);
```

nanaya:
Removed the problematic data instead.

PR #12437, opened 2025-10-02
Add length limit to team description field

nanaya:
It's not really supposed to be that long...

(there are some teams close to this length)


notbakaneko (approved):
60000 makes more sense than 63000 but it they're already that long, then 🤷

PR #12439, opened 2025-10-03
Set correct limit for team description

nanaya:
Apparently I never ended up committing the correct number for some reason ಠ_ಠ

(the longest is a bit over 63k)


notbakaneko (approved):
🤔

PR #12440, opened 2025-10-03
Limit return type of presence function

nanaya:
It currently includes `undefined` as possible return even though it's impossible (but not inferred due to `present` not being a type guard).

I wonder if the return should be `undefined` instead to match getInt (or the other one to return `null`?)


notbakaneko (approved):
I think it makes more sense to return `null` instead of nebulous `undefined` which is allowed to be assigned to `void` ಠ_ಠ

PR #12442, opened 2025-10-03
Save Beatmapset nomination level

notbakaneko:
When performing nomination checks, for nominators that are no longer nominators, uses the nomination level at the time they made the nomination. For current nominators, it still uses their current level.

I thought to adding another column but decided against it and just reuse the existing column and shim the calls.
Also going to rename calls to `modes` to `rulesets` either in another PR or just add on to this one.


app/Libraries/Beatmapset/NominateBeatmapset.php
```diff
@@ -64,16 +64,18 @@ public function handle()
         $nomination = $this->beatmapset->beatmapsetNominations()->current()->where('user_id', $this->user->getKey());
         if (!$nomination->exists()) {
             $this->beatmapset->getConnection()->transaction(function () {
+                $modes = array_filter($this->user->nominationModes(), fn ($ruleset) => $this->nominatedRulesets->contains($ruleset), ARRAY_FILTER_USE_KEY);
+
                 $eventParams = [
```
nanaya, line 61: might as well inline with the create

app/Libraries/Beatmapset/NominateBeatmapset.php
```diff
@@ -64,16 +64,18 @@ public function handle()
         $nomination = $this->beatmapset->beatmapsetNominations()->current()->where('user_id', $this->user->getKey());
         if (!$nomination->exists()) {
             $this->beatmapset->getConnection()->transaction(function () {
+                $modes = array_filter($this->user->nominationModes(), fn ($ruleset) => $this->nominatedRulesets->contains($ruleset), ARRAY_FILTER_USE_KEY);
```
nanaya: I wonder if `$this->nominatedRulesets` should just be this instead so the other checks don't need to refer back to user instance every time

app/Models/BeatmapsetNomination.php
```diff
@@ -10,7 +10,7 @@
  * @property int $beatmapset_id
  * @property \Carbon\Carbon|null $created_at
  * @property int $id
- * @property array $modes
+ * @property ?array $modes
```
nanaya, line 13: the rest uses `|null`...?

nanaya (changes requested):

app/Models/BeatmapsetNomination.php
notbakaneko: it's valid phpdoc (that doesn't work with union types) 🤷

nanaya: sounds like how the typing in php itself works

nanaya (approved):

PR #12448, opened 2025-10-06
Fix empty gap when viewing Beatmapset Discussion history

notbakaneko:
Caused by the spacer element expanding vertically when the history section is too short. The other way is to apply `osu-page--full` here but I expect there'll be comments on why history has a big blank space and the other sections don't. https://github.com/ppy/osu-web/blob/7e6e7361a423474f664a354b3a0c35ee6f00cc25/resources/js/beatmap-discussions/events.tsx#L23

Before:
image

After:
image


resources/css/bem/page-extra-tabs-before.less
```diff
@@ -3,7 +3,7 @@
 
 .page-extra-tabs-before {
   height: 15px;
-  margin: auto;
+  margin: 0 auto 0 auto;
```
nanaya: that's a funny way to write `0 auto`

PR #12451, opened 2025-10-07
Allow viewing unlisted group pages when in group

Hiviexd:
Supersedes #10445

### DIFF
--- a/app/Http/Controllers/GroupsController.php
+++ b/app/Http/Controllers/GroupsController.php
@@ -12,7 +12,8 @@ class GroupsController extends Controller
     public function show($id)
     {
         $group = app('groups')->byIdOrFail($id);
-        abort_unless($group->hasListing(), 404);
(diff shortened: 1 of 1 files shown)


peppy:
@Hiviexd what' the use case for this? one thing i'm worried about is that this presents a new security issue (we usually don't expose some groups, and this would be the first time we are).

Hiviexd:
I included this alongside #12450 when porting clayton's PRs #10445 and #10446.

Original PR states it makes sense as it matches permissions from the group history page, and I think one useful case from this is being able to easily clean-up the tournament staff and announce groups regularly, as those are the only unlisted groups that are [publicly documented](https://osu.ppy.sh/wiki/en/People/User_groups) and have regular additions/removals.

I don't feel strongly about this so I'll leave this entirely up to you whether you want this to be a thing or not.

peppy:
I think this fits better in an ASS command TBQH. Let's not.

PR #12458, opened 2025-10-08
Update styling for dashboard sidebar

nanaya:
More gaps and less negative margins (partial).

### DIFF
--- a/resources/css/bem/user-home.less
+++ b/resources/css/bem/user-home.less
@@ -22,15 +22,14 @@
   }
 
   &__beatmapsets {
-    margin: -10px 0;
-  }
-
-  &__button {
-    margin: 10px 0;
+    margin: 0 calc(var(--padding-x) * -1);
+    display: grid;
+    gap: 10px;
   }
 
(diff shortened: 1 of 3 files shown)


resources/css/bem/user-home.less
```diff
@@ -22,15 +22,14 @@
...
-    margin: 10px 0;
+    margin: 0 calc(var(--padding-x) * -1);
+    display: grid;
+    gap: 10px;
   }
 
   &__buttons {
-    margin: -10px 0;
+    display: grid;
+    gap: 10px;
```
notbakaneko, line 32: I think the gap is actually 7px on the current one because of the margin collapsing with the `a` margin :|

notbakaneko (approved):
gap between the home buttons is slightly larger but should be fine..? (or maybe the smaller one was wrong all along 🤔 )

resources/css/bem/user-home.less
nanaya: yeah I don't think that was intentional

PR #12459, opened 2025-10-08
Fix error on qualifying map with existing old disqualification

nanaya:
Managed to hit this when randomly testing nominating/qualifying a map.


notbakaneko (approved):
I assume it works because it's exactly the same as the one I didn't PR 👀

PR #12462, opened 2025-10-09
Clean up user online status chart

nanaya:
- weird margin/padding all over
  - which includes missing bottom margin
- a lot of extraneous svg handling
  - including leftover things when it's also used for profile page stuff


notbakaneko:
...wait since when did the hover stop working

notbakaneko (approved):

nanaya:
I don't think it's ever a thing for the chart on dashboard

PR #12471, opened 2025-10-14
Add anime cover flag to beatmapset

nanaya:
The database flag, user preference backend, and application of said preference on cover renderer.


notbakaneko (approved):
seems fine (assuming the frontend for setting options is separate)

PR #12474, opened 2025-10-15
Better random filename for test

nanaya:
Tried faker's unique word but it exploded at 500 so might as well generate completely random names? 🤷


nanaya:
also I wonder if the diff function should ignore case change 🤔

notbakaneko:
I feel like same filename with only difference in case should not be a thing, even if the filesystem originally used is case-sensitive...? @peppy

notbakaneko (approved):

Issue #12416, opened 2025-09-22
Show percentage of users that have unlocked individual medals


peppy:

glassive:
What would the percentage be based on? If I'm not mistaken, there are over 35 million user accounts, and simply dividing the amount of medal unlocks by such an amount would result in extremely small values for the rarer medals, to the point where it would be indistinguishable. The idea of displaying the total amount of unlocks for each medal sounded more sensible in that regard.

pompydev:
I have the same concern as @glassive.

<img width="244" height="432" alt="Image" src="https://github.com/user-attachments/assets/aab01b27-5879-425b-9fab-3c508f3463de" />

The number of people who passed a 1 star map not even being 60% is absurd.

glassive:
> What would the percentage be based on? If I'm not mistaken, there are over 35 million user accounts, and simply dividing the amount of medal unlocks by such an amount would result in extremely small values for the rarer medals, to the point where it would be indistinguishable. The idea of displaying the total amount of unlocks for each medal sounded more sensible in that regard.

Displaying the total amount of clears would not only function in giving players a sense of how common / rare a medal is, it would also help distinguish between rare medals in a more precise manner. Say two medals have 300 and 20 clears respectively ; when expressed as a percentage, its hard to tell them apart, but with the real value, it becomes obvious that one is even more difficult than the other.
Maybe there is a best of both worlds where we can both see the % of active players who unlocked and achievement, alongside the total clears across all accounts?

peppy:
Yeah, that sounds like it might work better.

PR #12479, opened 2025-10-16
Include achieved count in the achievement tooltip

nanaya:
This?

Reference: https://github.com/ppy/osu-web/issues/12416#issuecomment-3409473152


notbakaneko (changes requested):
conflicts

notbakaneko (approved):

PR #12480, opened 2025-10-16
Clean up source beatmap image file after processing

nanaya:
The fullsize one is still being used for moderation. It was originally the original file but using optimised one should be fine...?

### DIFF
--- a/resources/views/admin/beatmapsets/cover.blade.php
+++ b/resources/views/admin/beatmapsets/cover.blade.php
@@ -32,11 +32,14 @@
                 <i class="fas fa-fw fa-trash"></i>
                 {{osu_trans('admin.beatmapsets.covers.remove')}}
             </button>
-            @foreach (array_merge(['raw', 'fullsize'], $beatmapset->coverSizes()) as $size)
-                <h3>{{$size}}</h3>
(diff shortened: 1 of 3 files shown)


resources/views/admin/beatmapsets/cover.blade.php
```diff
@@ -32,11 +32,14 @@ class="btn-osu-big btn-osu-big--rounded-thin"
...
                 {{osu_trans('admin.beatmapsets.covers.remove')}}
             </button>
-            @foreach (array_merge(['raw', 'fullsize'], $beatmapset->coverSizes()) as $size)
-                <h3>{{$size}}</h3>
-                <a href="{{$beatmapset->coverURL($size)}}">
-                    <div>{{$beatmapset->coverURL($size)}}</div>
-                    <img class="beatmapset-cover-admin__img" src="{{$beatmapset->coverURL($size)}}">
+            @foreach (['fullsize', ...$beatmapset->coverSizes()] as $size)
+                @php
+                    $url = beatmapset->coverURL($size);
```
notbakaneko:
```suggestion
                    $url = $beatmapset->coverURL($size);
```

notbakaneko (changes requested):

notbakaneko (approved):

PR #12488, opened 2025-10-20
Switch to sampling rate for sentry metrics thing

nanaya:
It seems to be only ever so slightly faster than sentry's native sampling (which has more details) 🤷 (if at all)

(the whole thing, including the previous threshold version)


app/Http/Middleware/SentryMetrics.php
```diff
@@ -16,10 +16,10 @@ public function handle(Request $request, \Closure $next): Response
 
         $response = $next($request);
 
-        $end = microtime(true);
-        $duration = $end - $start;
+        $sampleRateInversed = $GLOBALS['cfg']['osu']['sentry']['log_sample_rate_inversed'];
+        if ($sampleRateInversed !== null && rand(1, $sampleRateInversed) < 100) {
```
notbakaneko: shouldn't this be `<= 100` otherwise a sample rate of 1 isn't 100%?

notbakaneko (approved):

PR #12489, opened 2025-10-21
Add option to require verification for all users

nanaya:
May be needed in some cases.

The indentation looks a bit weird...


notbakaneko (changes requested):
also the `.env.example`

notbakaneko (approved):

PR #12491, opened 2025-10-21
Use different source for checking score processing status

### LINKED ISSUE #12463: Scores with unranked mods always show that calculation is still in progress
Image
This also applies to the game client:

Image
(i'm not exactly sure I'm opening an issue in the right place but at least you'll know about it)

nanaya:
The history table may get occasionally truncated so it's not quite reliable especially for older scores.

Inversely the value in the count table may be a bit too optimistic but it should only last a few seconds.

- [x] depends on #12490

Resolves #12463.

<sub>this silly collection overriding thing is quite confusing but also convenient</sub>


app/Models/Solo/Score.php
```diff
@@ -404,7 +411,9 @@ public function isProcessed(): bool
             return $this->attributes['processed_version'] !== null;
         }
 
-        return $this->processHistory !== null;
+        $this->collection ??= new ScoreCollection([$this]);
+
+        return $this->getKey() <= $this->collection->lastProcessedScoreId();
```
nanaya, line 416: fwiw I tried using the request attribute directly here but it ends up quite a bit slower.

notbakaneko (dismissed):

notbakaneko (approved):

PR #12492, opened 2025-10-21
Fix BBCodeFromDB deprecation warning

notbakaneko:
noticed when reviewing #12487

I wonder if it's better to just initialize the `$text` attribute as empty string instead but I didn't go as far as checking if anything relied on it being null but not empty string, yet. 🤔 
(Also most of the methods and attributes can all probably be made private?)

### DIFF
--- a/app/Http/Middleware/RequireScopes.php
+++ b/app/Http/Middleware/RequireScopes.php
@@ -68,13 +68,10 @@ class RequireScopes
      */
     private function requestHasScopedMiddleware(Request $request): bool
     {
-        $value = $request->attributes->get('requestHasScopedMiddleware');
-        if ($value === null) {
-            $value = $this->containsScoped($request);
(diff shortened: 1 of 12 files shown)


app/Libraries/BBCodeFromDB.php
```diff
@@ -405,6 +405,9 @@ public function toHTML()
     public function toEditor()
     {
         $text = $this->text;
+        if ($text === null) {
+            return null;
```
nanaya: probably want to return empty string instead to keep existing behaviour? (and if we ever check whether or not the source may ever be null)

notbakaneko: wait str_replace returns `''` with`null` ಠ_ಠ

nanaya: I think most of the text replacement functions return empty string when passed null...

PR #12494, opened 2025-10-23
Exclude restricted users when counting the more rare achievements

nanaya:
🤷 

Counting those on more common achievements with 100k+ (esp on the 1M+ range) achieved count takes forever so just the rarer ones for now.

(already updated on live db)

note that we don't have scheduled count refresh job at the moment...


peppy:
Is there a reason the job isn't scheduled? Is that something I have to do?

nanaya:
because it's not originally supposed to have the number decreases (the increment is done in real time)

notbakaneko (approved):

notbakaneko:
(not sure if we want the cutoff to be configurable but that can be done later)

PR #12498, opened 2025-10-28
Fix user agent detection deprecation

notbakaneko:
Switch to `device-detector` instead; the main difference is for platforms, `OS X` is now just `Mac` and `Linux` is `GNU/Linux`. Also updated the comment about iPadOS because it just uses a desktop Safari UA...

### DIFF
--- a/app/Jobs/RemoveBeatmapsetBestScores.php
+++ b/app/Jobs/RemoveBeatmapsetBestScores.php
@@ -5,7 +5,6 @@
 
 namespace App\Jobs;
 
-use App\Libraries\Elasticsearch\BoolQuery;
 use App\Models\Beatmap;
 use App\Models\Beatmapset;
 use App\Models\Score\Best\Model;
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -20,7 +20,6 @@
     "graham-campbell/github": "*",
     "guzzlehttp/guzzle": "*",
     "itsgoingd/clockwork": "^5.1",
-    "jenssegers/agent": "*",
```
nanaya, line 23: also used in `app/Libraries/Session/Store.php`

app/Libraries/Agent.php
```diff
@@ -0,0 +1,37 @@
...
+
+        if (!isset($rules)) {
+            $rules = array_merge(
+                static::$browsers,
+                static::$operatingSystems,
+                static::$phoneDevices,
+                static::$tabletDevices,
+                static::$additionalOperatingSystems,
+            );
+        }
```
nanaya: I think you can just assign it directly now
```suggestion
        static $rules = array_merge(
            static::$browsers,
            static::$operatingSystems,
            static::$phoneDevices,
            static::$tabletDevices,
            static::$additionalOperatingSystems,
        );
```

nanaya:
I wonder if it's better to just use a different library (like [device-detector](https://packagist.org/packages/matomo/device-detector)) which actually supports all type of devices instead of shoehorning desktop browser data into mobile-only library.

nanaya (approved):

PR #12502, opened 2025-10-29
Add top plays page

nanaya:
I guess. <del>I'll rename the page (and maybe the controller) to something better later™.</del>

There's no mobile specific layout (yet?). It just scrolls vertically like other ranking pages.

(not actual/current data)


minisbett:
Maybe an "x" after the combo to reflect how leaderboards on beatmap pages look like?

And also, what will the page look like if a score has many mods? afaik that's also an issue with beatmap leaderboards currently, leading to scuffed table layouts at times

tsunyoku:
The placement of the grade feels a bit weird considering user profiles and beatmap leaderboards both tend to put the grade as far to the left as possible. Could it be moved left to the beatmap title or something like that?

minisbett:
> The placement of the grade feels a bit weird considering user profiles and beatmap leaderboards both tend to put the grade as far to the left as possible. Could it be moved left to the beatmap title or something like that?

What if the grade is.. left away? In the scope of beatmap leaderboards etc. they absolutely make sense, but here the pages main intention is designed to be PP, and i feel like grades don't matter as much as info because various maps are compared.

It'd also allow more room for mods.

tsunyoku:
I suppose grades matter less in this context. At bare minimum I think put miss count in its place. Even if we're keeping it that should probably exist.

notbakaneko (commented):
Looks fine, other than the 15+KB of numbers sent to the db with every request...

notbakaneko (approved):
Probably going to need our own length aware pagination interface or the ranking page template not rely on it to remove the growing list of ids sent, so let's go with this for now...

PR #12509, opened 2025-10-31
Adjust class name for legacy replay file library

nanaya:
The plan is to add another one for the new one and have the main Score class return either class accordingly. And remove the new replay file related functions in the Score class.


notbakaneko (approved):
looks correct 🤔

PR #12510, opened 2025-10-31
Separate new score replay file download to its own class 

nanaya:
I was about to adjust the controller download but it ended up a lot bigger than expected so just this for now.

- [x] depends on #12509

### DIFF
--- a/app/Models/Solo/Score.php
+++ b/app/Models/Solo/Score.php
@@ -10,6 +10,7 @@ namespace App\Models\Solo;
 use App\Enums\Ruleset;
 use App\Enums\ScoreRank;
 use App\Exceptions\InvariantException;
+use App\Libraries\Score\ReplayFile;
 use App\Libraries\Score\ScoringMode;
 use App\Libraries\Score\UserRank;
 use App\Libraries\Search\ScoreSearchParams;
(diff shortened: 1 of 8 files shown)


app/Models/Solo/Score.php
```diff
@@ -518,6 +503,13 @@ public function queueForProcessing(): void
         ]));
     }
 
+    public function replayFile(): ?ReplayFile
+    {
+        return !$this->isLegacy() && $this->has_replay
```
notbakaneko, line 508: Would it make sense to return `LegacyReplayFile` instead of `null` if it's a legacy score?

Looking at where `legacyScore()` is currently used, this seems to be the only other thing in the controller that makes use of it being a legacy score; the others should have the same value in `$soloScore`: 
https://github.com/ppy/osu-web/blob/23b7f66fb2563ad6b5bc09fe519fa30f6b463512/app/Http/Controllers/ScoresController.php#L98-L102

...although it's probably less confusing if it doesn't pull files from a different location...

nanaya: it's part of the controller adjustment which I excluded here...

although it may end up as part of removing the function in legacy score which involves updating logic on the score deletion on beatmap qualification instead (and moving the replay view count as well)

notbakaneko (approved):

PR #12514, opened 2025-11-04
Define macro functions directly

nanaya:
Instead of using an intermediate function generator.


app/Models/Store/Order.php
```diff
@@ -627,39 +627,37 @@ public static function cart($user)
             ->first();
     }
 
-    public function macroItemsQuantities(): \Closure
```
nanaya, line 630: this one doesn't seem to be used so I wonder if I should just delete it instead...?

notbakaneko: can probably remove it; I feel like it was for one of the order displays that's long gone

notbakaneko (approved):

notbakaneko (approved):

PR #12520, opened 2025-11-05
Use custom collection to handle room participant preload

nanaya:
Not sure if actually better 🤔


notbakaneko:
not sure if putting preloads into a collection is better 🤔 
Also it's repopulating every time `recentParticipants()` is called 👀

nanaya:
well technically all laravel preloads are done through collection. the mapping shouldn't take too long since it's just a map (and only called once)

notbakaneko (approved):

PR #12521, opened 2025-11-05
Adjust quick play user stats indexes

nanaya:
<del>It's not yet deployed so might as well just edit the existing migration ( ﾟ ヮﾟ)a</del> lol


bdach:
It'll have been deployed to staging tho so you might need to correct manually there. Although it's an index so *maaaaybe* not a huge deal.

notbakaneko (approved):

PR #12523, opened 2025-11-06
Updated download page with new default client

notbakaneko:
The guide video will be updated after a new guide is made with the new page.
The placeholders also need to be filled with images.

I tried fiddling with the taglines layout and alignment to make it handle different length texts without looking silly or lopsided, but it's hard to tell without the localisations. Ideally you'd want both lines to be about the same length or it starts looking odd. (We're currently assuming the backgrounds have something on the left to balance out the buttons on the right)

edit: added images/video
edit 2: updated images and video (video doesn't have some text/image updates)

https://github.com/user-attachments/assets/e1fafd88-ef9f-4751-a64f-596555fcfa11

<details>
<summary>Desktop:</summary>
image
image

</details>
<details>
<summary>Mobile:</summary>
image
</details>


tadatomix:
May I ask something? Is it possible to look at the change, without running website locally? And may I ask what was meant by the guide in this case? Thank you

resources/views/home/download.blade.php
```diff
@@ -9,102 +9,107 @@
...
-                    <h3 class="download-page-header__action-title">
-                        {{ osu_trans('home.download.action_lazer_title') }}
-                    </h3>
-
-                    <div>
-                        <div>
-                            {{ osu_trans('home.download.action_lazer_description') }}
+            <div class="download-page__banner-tail">
+                <div class="download-page__banner-content download-page__banner-content--tail">
+                    <div class="download-page__text download-page__text--download-stable">if you're looking for the older one</div>
```
nanaya: translation? also can probably use a link to describe the difference and stuff (the same as the existing one?)

resources/css/bem/download-page.less
```diff
@@ -2,57 +2,190 @@
...
+
     @media @desktop {
-      max-width: 220px;
+      padding: 40px;
     }
   }
 
-  &__steps {
+  &__tagline-1 {
+    position: relative;
```
nanaya: not needed?

app/Http/Controllers/HomeController.php
```diff
@@ -68,25 +69,51 @@ public function getDownload()
...
-            'Windows' => 'windows_x64',
-            // current iPadOS declares itself as a desktop browser.
-            'iOS' => 'ios',
-            // FIXME: Figure out a way to differentiate Intel and Apple Silicon.
-            'Mac' => 'macos_as',
-            'Android' => 'android',
-            'GNU/Linux' => 'linux_x64',
-            default => 'windows_x64',
-        };
+        $platform = request('platform');
```
nanaya: can use a `get_string` so the following check doesn't explode when accessed with `platform[]=x`

resources/css/bem/select-options.less
```diff
@@ -77,6 +81,10 @@
 
     --option-hover-bg: hsla(var(--hsl-b6), 0.6);
     --option-padding: var(--option-select-padding);
+
+    .@{_top}--download & {
+      border-radius: @border-radius--extra;
```
nanaya, line 87: the popup can probably also use some more rounded corners

also the text and icon should probably be smaller

app/Http/Controllers/HomeController.php
```diff
@@ -68,25 +69,51 @@ public function getDownload()
...
+        if (!array_key_exists($platform, $lazerPlatformNames)) {
+            $platform = null;
+        }
+
+        if ($platform === null) {
```
nanaya: might as well just do it in one step?

```suggestion
        if (!array_key_exists($platform, $lazerPlatformNames)) {
```

resources/css/bem/download-page.less
```diff
@@ -2,57 +2,190 @@
...
 
   &__step-number {
-    .circle(@number-width);
     .center-content();
+    width: 60px;
+    height: 24px;
+    border-radius: $height;
 
-    border: solid @osu-colour-h1 2px;
+    border: solid #fff 2px;
```
nanaya: `currentColor` for the colour probably makes sense here

resources/css/bem/download-page.less
```diff
@@ -2,57 +2,190 @@
...
     color: #fff;
```
nanaya: might as well remove this redundant styling while at it

resources/css/bem/btn-osu-big.less
```diff
@@ -143,35 +143,55 @@
...
-    --bg-colour: hsl(var(--hsl-blue-3));
-    --bg: var(--bg-colour) center/150% url("~@images/backgrounds/button.svg");
-    --hover-bg: var(--bg-colour) calc(50% - 20px) 50%/150%
-      url("~@images/backgrounds/button.svg");
-  }
-
-  &--download-lazer {
-    --bg-colour: hsl(var(--hsl-purple-3));
+  &--download-stable {
+    --bg-colour: var(--hsl-yellow-1);
```
nanaya: where did this come from? shouldn't it be orange? (and where did the yellow definition come from?)

resources/css/bem/btn-osu-big.less
```diff
@@ -143,35 +143,55 @@
...
+          hsla(var(--bg-colour), 0.9)
+        )
+        border-box;
+
+    --bg-colour: var(--hsl-pink-1);
+    --cookie: url("~@images/backgrounds/cookie_v3.svg");
+    --overlay-colour: rgba(0, 0, 0, 0.75);
+
+    .default-text-shadow();
+    position: relative;
```
nanaya: doesn't do anything?

resources/css/bem/download-page.less
```diff
@@ -2,57 +2,190 @@
...
+      margin-bottom: 0;
+      padding-top: 20px;
+      padding-bottom: 20px;
+    }
+  }
+
+  &__banner-tail {
+    display: flex;
+    justify-content: flex-end;
+    gap: 10px;
```
nanaya: not needed?

resources/css/bem/download-page.less
```diff
@@ -2,57 +2,190 @@
...
+    height: 120px;
+    max-width: 280px;
+  }
+
+  &__extra-link {
+    .link-white();
+    font-size: @font-size--title-small;
+
+    &--separator {
+      pointer-events: none;
```
nanaya: `::before` can't be selected anyway...?

resources/css/bem/download-page.less
```diff
@@ -2,57 +2,190 @@
...
     color: #fff;
     font-weight: 700;
-    margin-right: @number-margin-right;
-    flex: none;
     padding-bottom: 2px; // more center the number
   }
 
   &__text {
-    color: @osu-colour-c1;
+    color: hsl(var(--hsl-c1));
```
nanaya: not needed

resources/css/bem/btn-osu-big.less
```diff
@@ -482,21 +507,27 @@
     }
   }
 
+  &__logo {
```
nanaya: not used

nanaya (changes requested):

resources/css/bem/btn-osu-big.less
notbakaneko: Orange is less orange than I thought 🤔 
It looked yellower on the design but, come to think of it, that's probably because the background there is green 🙈

resources/views/home/download.blade.php
```diff
@@ -9,102 +9,116 @@
...
+                        <span class="download-page__step-number">1</span>
                         <div class="download-page__text download-page__text--title">
-                            <span class="download-page__step-number">1</span>
                             {{ osu_trans("home.download.steps.download.title") }}
                         </div>
                         <div class="download-page__text download-page__text--description">
                             {{ osu_trans("home.download.steps.download.description") }}
                         </div>
+                        <div class="download-page__example">
+                            {{-- TODO: placeholder --}}
```
nanaya: just run the installer/client/etc and take the needed screenshots maybe

resources/views/home/download.blade.php
```diff
@@ -9,102 +9,116 @@
...
-                        {{ osu_trans('home.download.action_lazer_title') }}
-                    </h3>
-
-                    <div>
+            <div class="download-page__banner-tail">
+                <div class="download-page__banner-content download-page__banner-content--tail">
+                    <div class="download-page__text download-page__text--download-stable">
                         <div>
-                            {{ osu_trans('home.download.action_lazer_description') }}
+                            {{ osu_trans('home.download.stable_text') }}
```
nanaya: the wordings and the lack of punctuation makes it read like one sentence

nanaya (changes requested):

resources/views/home/download.blade.php
notbakaneko: Am I the only one that thinks having screenshots like 
image
or 
image
resized and cropped to fit the aspect ratio (and somehow still be visibly legible) of the design looks silly? 🤔

nanaya: I don't think it needs to be actually legible, just the general vibe. And maybe post in design channel as well... (the actual page with screenshots filled in)

app/Http/Controllers/HomeController.php
```diff
@@ -64,29 +65,52 @@ public function getDownload()
             'android' => osu_trans('home.download.os_version_or_later', ['os_version' => 'Android 5']),
             'ios' => osu_trans('home.download.os_version_or_later', ['os_version' => 'iOS 13.4']),
             'linux_x64' => 'Linux (x64)',
-            'macos_as' => osu_trans('home.download.os_version_or_later', ['os_version' => 'macOS 10.15']).' (Apple Silicon)',
-            'windows_x64' => osu_trans('home.download.os_version_or_later', ['os_version' => 'Windows 8.1']).' (x64)',
+            'macos_as' => osu_trans('home.download.os_version_or_later', ['os_version' => 'macOS 12']).' (Apple Silicon)',
+            'macos_intel' => osu_trans('home.download.os_version_or_later', ['os_version' => 'macOS 12']).' (Intel)',
```
nanaya, line 69: can use a config entry for this...

notbakaneko: for which part >_>

nanaya: the lazer_dl thing so the download button actually does something when clicked...

Issue #12506, opened 2025-10-30
Remove usage of legacy score tables


nanaya:
The `osu_scores{,_{taiko,fruits,mania}}{,_high}` table. The data should all be available in the new table with two exceptions (non-best score id and mania perfect).

Update steps
- [x] 1. use new score model for replay file download
- [x] 2. use new score model for legacy view count relation (`osu_replays{,_{taiko,fruits,mania}}`)
  - this includes moving handling of replay file deletion
  - there may be existing bug of `has_replay` in new model not being updated correctly
- [ ] 3. ??? something else I probably forgot about
- [ ] 4. replace legacy score models with plain class (mainly for legacy transformer)
- [ ] 5. remove most of legacy score models logic
  - simplify just enough for it to be plainly deleted by the removal job

peppy:
@nanaya i think some pieces of this may have been completed (hopefully?)

nanaya:
I think number 1 is only halfway done atm. Number 2 ended up requiring a few test updates. Number 4 is partially done but no pr yet.

PR #12532, opened 2025-11-10
Move handling of legacy replay file to main score model

nanaya:
Still need checking.

(addresses part of #12506)


nanaya:
Done in a different way.

PR #12536, opened 2025-11-11
Clean up score controller

nanaya:
- unify replay file name
- use new score model for all operations on replay download
- skip Score model import alias

depends on
- [x] #12535

(addresses part of #12506)


app/Http/Controllers/ScoresController.php
```diff
@@ -95,27 +92,23 @@ public function download($rulesetOrSoloId, $id = null)
...
-        ];
-
-        return response()->streamDownload(function () use ($file) {
-            echo $file;
-        }, $this->makeReplayFilename($score), $responseHeaders);
+        return response()->streamDownload(
+            function () use ($file) {
+                echo $file;
+            },
+            "replay-{$score->getMode()}_{$score->beatmap_id}_{$score->getKey()}.osr",
```
notbakaneko: I don't think we should be using exactly the same naming format as legacy replay when the score ids will be different

notbakaneko (changes requested):

app/Http/Controllers/ScoresController.php
nanaya: the id here is always the new one

(yeah it'll be kinda confusing with older downloads but the previous name for this sucked

notbakaneko: that's what I'm saying, the new `replay-osu_1_1.osr` will be different from the previous `replay-osu_1_1.osr`; although it *should* be unlikely to conflict due to beatmap ids (I look forward to being proven wrong 🙈 )

nanaya: updating to a better filename is a bit more involved so see you later in a different pr ಠ_ಠ

notbakaneko (dismissed):
just a note that this seems fine, but I'd rather wait for the filename format update before this goes in instead of the filename changing now and again later.

nanaya:
the format rename with current controller is a bit annoying because two different models are involved...

nanaya:
Or I guess I can just revert to the current filename instead.

notbakaneko (approved):

PR #12537, opened 2025-11-12
 Remove custom collection for models

nanaya:
Any map attempts will break this so never mind 😐

The cached request thing doubles the time but should still be in microsecond range.


notbakaneko (approved):
custom collection so useful until it wasn't 😬

PR #12538, opened 2025-11-12
Add `matchmaking_room_events` table

smoogipoo:
Similar to `multiplayer_realtime_room_events` with one important distinction: the playlist item ID can be -1 (indicating a "random" user selection).

### DIFF
--- a/database/migrations/2025_11_12_073551_create_matchmaking_room_events.php
+++ b/database/migrations/2025_11_12_073551_create_matchmaking_room_events.php
@@ -17,7 +17,8 @@ return new class extends Migration
             $table->bigIncrements('id');
             $table->unsignedBigInteger('room_id');
             $table->string('event_type');
-            $table->bigInteger('playlist_item_id')->nullable();
+            $table->bigInteger('playlist_item_id')->nullable()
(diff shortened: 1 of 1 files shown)


nanaya:
can't the random playlist item id thing be 0 instead? the column type may get confusing if it's randomly referenced for the type of the column when adding another table etc.

smoogipoo:
The -1 was decided on internally in discussion starting [here](https://discord.com/channels/90072389919997952/1327149041511043134/1432979680431964180). At this point I'd pull in @bdach to comment on why -1 was better than 0.

bdach:
I don't like ascribing special meanings to default values of data types. Leads to bugs wherein *forgetting to set* the value results in the same thing as *purposefully using the special value*.

I would have preferred to use `INT NULL`, for whatever it's worth, but it would have been annoying due to having to do the song and dance of migrating client contracts between spectator server and client, so `-1` was the next best thing.

If you're horribly offended by it maybe something like `int.MaxValue` could be used instead but like... it's still kind of dangerous to use that if we ever approach the limit of int with respect to beatmap IDs. Or we grit teeth and do the `NULL` thing, I dunno. Just anything other than zero...

nanaya:
maybe add `->comment(...)` to the column to indicate it's not quite the right type and hopefully nothing explodes due to the slight type mismatch

PR #12541, opened 2025-11-12
Add special handling for season room ranking page

nanaya:
First step in reorganising playlist ranking page.


notbakaneko (changes requested):
So if the latest featured Room has a season, it highlights the season tab, shows the list of rooms of the season in the drop down but you can't actually select the multiplayer tab and select a different room?

nanaya:
the featured scope has been changed to only return featured_artist rooms

if the featured_artist room is part of a season, we have a different problem

notbakaneko (approved):
(assuming nothing goes where it shouldn't)

PR #12545, opened 2025-11-13
 Group playlist related ranking pages

nanaya:
<del>Waiting on design (maybe).</del> Matches `/follows` page design. There's a newer design but it'll be done separately.

- [x] depends on #12544


notbakaneko (changes requested):
ruleset selector on old spotlights page also needs some updating

notbakaneko (approved):

PR #12547, opened 2025-11-14
Reduce number checks when mounting react component

nanaya:
Rather than looping by the ever-growing `reactTurbolinks.components`, check for the actually existing elements on the page instead.


notbakaneko (changes requested):
conflicts

notbakaneko (approved):
looks correct

PR #12553, opened 2025-11-18
Better shorthand name for month

nanaya:
`m` just looks like minute. A few other languages already have longer text anyway.

(discussed offline)


notbakaneko (approved):

bdach:
Client [was doing this already](https://github.com/ppy/osu/blob/be77257ddb4d819001e999be20962d99fed4ae88/osu.Game/Extensions/TimeDisplayExtensions.cs#L97-L98) so no adjustments required.

PR #12555, opened 2025-11-18
Add ranking page for quickplay

nanaya:
now with sort option and pools

- [x] depends on #12545
- [x] depends on #12562
- [x] rewrite to use the updated pool based listing

### DIFF
--- a/app/Http/Controllers/Ranking/MatchmakingController.php
+++ b/app/Http/Controllers/Ranking/MatchmakingController.php
@@ -36,7 +36,7 @@ class MatchmakingController extends Controller
             ->allUserStats()
             ->with('user.team')
             ->whereHas('user', fn ($q) => $q->default())
-            ->where('elo_data->contest_count', '>=', 5);
+            ->where('plays', '>=', 5);
 
         $sort = get_string(request('sort'));
         if (!array_key_exists($sort, static::SORTS)) {
--- a/app/Http/Controllers/Ranking/MatchmakingController.php
+++ b/app/Http/Controllers/Ranking/MatchmakingController.php
@@ -11,7 +11,7 @@ use App\Http\Controllers\Controller;
 use App\Models\Beatmap;
 use App\Models\MatchmakingPool;
 
-class QuickplayController extends Controller
+class MatchmakingController extends Controller
 {
     const SORTS = [
         'points' => [['total_points', 'DESC'], ['rating', 'DESC']],
@@ -26,7 +26,7 @@ class QuickplayController extends Controller
         if ($poolId === null) {
             $pool = MatchmakingPool::where(['ruleset_id' => $rulesetId])->firstOrFail();
 
-            return ujs_redirect(route('rankings.quickplay', ['mode' => $rulesetName, 'pool' => $pool->getKey()]));
+            return ujs_redirect(route('rankings.matchmaking', ['mode' => $rulesetName, 'pool' => $pool->getKey()]));
         }
 
         $pools = MatchmakingPool::where(['ruleset_id' => $rulesetId])->get();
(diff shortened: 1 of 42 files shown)


resources/views/rankings/matchmaking.blade.php
```diff
@@ -0,0 +1,111 @@
...
+    use App\Http\Controllers\Ranking\MatchmakingController;
+
+    $params = [
+        'mode' => $rulesetName,
+        'sort' => $sort,
+    ];
+    $poolSelectOptionTransformer = fn ($pool) => [
+        'id' => $pool->getKey(),
+        'text' => $pool->getDisplayName(),
+    ];
```
nanaya: I wonder if the SelectOptionTransformer should be updated to this instead (and have dedicated interface etc)

nanaya:
I think that's all the additional updates I want to do here ಠ_ಠ (short of actual errors)

external name is kept as quick play for now.

app/Models/Beatmap.php
```diff
@@ -74,6 +74,23 @@ class Beatmap extends Model implements AfterCommit
...
+        ],
+        self::MODES['fruits'] => [
+            0 => '',
+        ],
+        self::MODES['mania'] => [
+            0 => '',
+            4 => '4k',
+            7 => '7k',
+        ],
+    ];
```
nanaya, line 92: oh yeah, this feels vaguely silly especially considering we won't be adding any more variants but this sort of also work as variant validation so maybe fine? also dunno what the value really should be (currently it's the display-ish name)

also I don't even want to think how this goes in typescript...

app/Models/MatchmakingUserStats.php
```diff
@@ -0,0 +1,37 @@
...
+ * @property array $elo_data
+ * @property int $first_placements
+ * @property int $pool_id
+ * @property int $rating
+ * @property int $total_points
+ * @property \Carbon\Carbon|null $updated_at
+ * @property User $user
+ * @property int $user_id
+ */
+class MatchmakingUserStats extends Model
```
notbakaneko, line 23: where do these stats come from?

app/Http/Controllers/Ranking/MatchmakingController.php
```diff
@@ -0,0 +1,58 @@
...
+use App\Http\Controllers\Controller;
+use App\Models\Beatmap;
+use App\Models\MatchmakingPool;
+
+class MatchmakingController extends Controller
+{
+    const SORTS = [
+        'points' => [['total_points', 'DESC'], ['rating', 'DESC']],
+        'rating' => [['rating', 'DESC'], ['total_points', 'DESC']],
+    ];
```
notbakaneko, line 19: These seem to prefer the `(pool_id, first_placements)` index (because the 2nd order column isn't in the respective indexes?)

notbakaneko (changes requested):

app/Models/MatchmakingUserStats.php
```diff
@@ -0,0 +1,37 @@
...
+{
+    public $incrementing = false;
+
+    protected $casts = [
+        'elo_data' => 'array',
+    ];
+    protected $primaryKey = ':composite:';
```
notbakaneko:
```suggestion
    protected $primaryKey = ':composite';
```

notbakaneko: also `ModelCompositePrimaryKeysTest`

notbakaneko (changes requested):

app/Http/Controllers/Ranking/MatchmakingController.php
notbakaneko: maybe a helper that is like
```
function json_options(Interface $current, Interface[] $items): array
{
  return [
    'currentItem' => ...,
    'items' => ...,
  ]
}
```
would be useful since it's almost always the case that both are wanted

notbakaneko (approved):

PR #12559, opened 2025-11-19
 Add `pool_id` to `matchmaking_user_stats`

smoogipoo:
Primarily, replaces the `ruleset_id` column with `pool_id` in `matchmaking_user_stats`.

Migration is somewhat complex due to primary key uniqueness, and has a caveat documented in code.

Secondarily, removes the `rating` column from the same table, which will/has been replaced with a virtual column.


smoogipoo:
Hmmm now that I think about it again, we might want to keep the `ruleset_id` column around temporarily (just unused) for server-spectator.

PR #12570, opened 2025-11-26
Add `comment` column to `matchmaking_pools`

smoogipoo:
I think we may want this just to describe the pools for our own uses, not actually used by any service.

### DIFF
--- a/database/migrations/2025_11_26_081715_add_comment_to_matchmaking_pools.php
+++ b/database/migrations/2025_11_26_081715_add_comment_to_matchmaking_pools.php
@@ -1,0 +1,27 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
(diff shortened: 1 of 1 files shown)


database/migrations/2025_11_26_081715_add_comment_to_matchmaking_pools.php
```diff
@@ -0,0 +1,27 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('matchmaking_pools', function (Blueprint $table) {
+            $table->string('comment')->nullable()->after('name');
```
peppy, line 17: Probably call it `description` because I think it'd be useful to show this as tooltip or subtext.

(and maybe preemptively include in api response while we're here)

smoogipoo:
On further discussion this isn't going to be required as we'll be displaying the name as a subheading in the client.

PR #12571, opened 2025-11-27
Show highest matchmaking rank on profile page

nanaya:
Details tooltip will be added later™ (in a different pr)

- [x] depends on #12555

### DIFF
--- a/resources/js/profile-page/matchmaking.tsx
+++ b/resources/js/profile-page/matchmaking.tsx
@@ -21,7 +21,7 @@ export default class Matchmaking extends React.Component<Props> {
     }
 
     for (const stats of this.props.allStats) {
-      if (stats.rank > highestRank.rank) {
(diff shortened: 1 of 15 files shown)


resources/js/profile-page/matchmaking.tsx
```diff
@@ -0,0 +1,42 @@
...
+export default class Matchmaking extends React.Component<Props> {
+  render() {
+    let highestRank = this.props.allStats[0];
+
+    if (highestRank == null) {
+      return null;
+    }
+
+    for (const stats of this.props.allStats) {
+      if (stats.rank > highestRank.rank) {
```
notbakaneko: shouldn't this be the other way for higher rank?

notbakaneko (changes requested):

notbakaneko (approved):

Issue #12556, opened 2025-11-18
Filters for top plays page


AJT127:
https://osu.ppy.sh/rankings/top-plays/osu?page=1#scores 
Really cool feature. Was wondering about the scope for:

1) Allowing to filter by country / mod combination (maybe even friend ranking as well)
2) Adding a button to only show the highest pp score of a map, allowing for a lot more depth in displayed plays

PR #12573, opened 2025-11-28
Add support for country-specific top plays

tsunyoku:
image

Addresses part of #12556.

There's probably a non-zero chance that the time required to recalculate top plays for each country is too much on a production-sized index, but I don't have a good way to measure this nor make assumptions. In the worst case, country-specific top plays could fetch a lower score count, or maybe the country recalculation could be split into a separate command that runs less frequently?

### DIFF
--- a/app/Console/Commands/RankingsRecalculateTopPlays.php
+++ b/app/Console/Commands/RankingsRecalculateTopPlays.php
@@ -9,6 +9,7 @@ namespace App\Console\Commands;
 
 use App\Libraries\Score\TopPlays;
 use App\Models\Beatmap;
+use App\Models\Country;
 use Illuminate\Console\Command;
 
 class RankingsRecalculateTopPlays extends Command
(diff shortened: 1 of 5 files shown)


nanaya:
yeah this takes 30 minutes compared to 30 seconds for global only...

I doubt reducing count would help - it's most likely the sorting and filtering that is taking the most time.

tsunyoku:
How about a separate command that updates less frequently then?

nanaya:
there's probably a way to get the whole thing in one go

also the plan is to increase the frequency, not reducing it...

tsunyoku:
The idea behind updating less frequently was to prevent updates of the global top plays just because the country one is slower. I agree reducing frequency is not ideal.

I'm not immediately sure of any ways we can do the whole thing in one go, at least if you want the country top plays to be equally populated, but if you have a direction to point me in then I can try figure it out.

nanaya:
something with this maybe? https://www.elastic.co/guide/en/elasticsearch/reference/7.17/search-aggregations-metrics-top-hits-aggregation.html

PR #12574, opened 2025-11-28
Show miss count on top plays rankings

tsunyoku:
image

I think miss count is probably the most important score related metric alongside accuracy, so it should be shown. Especially with no grade/rank shown (I think it should remain that way), at bare minimum miss count can be an indicator of FC.


nanaya:
full combo is already indicated by the colour of the max combo value though

tsunyoku:
It doesn't indicate how far away from an FC it is and we have nothing which indicates that either, it's really hard to deduce any information from these cards for non-perfect combos.

Max combo is also not a _perfect_ indicator of FC because of dropped slider ends, but I don't really think that's relevant here.

nanaya:
should the column be after max combo as they're closely related?

(also isn't it the miss count that can't be used as full combo indicator? or am I misunderstanding the terms here? the perfect combo indicator is based on actual achievable max combo)

peppy:
Positioning looks correct if we're matching the beatmap information page. But the case of all these headers doesn't match (all caps over there), so maybe we should match that while we're here?

nanaya:
all other ranking page headers are title cased

also if we're comparing to beatmap leaderboard, the accuracy column is somewhere else completely, not between max combo and miss

peppy:
That's fine, let's choose an ordering and go with it, it's not important.

PR #12576, opened 2025-12-01
Don't return error when receiving non ECheck pending IPN from Paypal

notbakaneko:
So Paypal doesn't keep retrying it.

Paypal sends more than `echeck` as the reason for `Pending` now, but the IPN payload itself doesn't have enough information to determine what is going on (cancellation, bank/card provider delays/confirmation, etc) without digging through the transaction history.

We don't really treat pending differently other than recording information since it all looks the same from our end.

### DIFF
--- a/app/Libraries/Payments/PaymentProcessor.php
+++ b/app/Libraries/Payments/PaymentProcessor.php
@@ -12,7 +12,6 @@ use App\Models\Store\Payment;
 use App\Traits\Memoizes;
 use App\Traits\Validatable;
 use DB;
-use Sentry\Severity;
 use Sentry\State\Scope;
 
 abstract class PaymentProcessor implements \ArrayAccess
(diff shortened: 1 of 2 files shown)


app/Libraries/Payments/PaymentProcessor.php
```diff
@@ -211,9 +212,20 @@ public function pending()
...
+            // Log warning if non echeck pending status will overwrite the echeck field.
+            if ($order->tracking_code === Order::PENDING_ECHECK && $order->tracking_code !== $reason) {
+                app('sentry')->getClient()->captureMessage(
+                    "Trying to overwrite pending echeck with {$reason}",
+                    new Severity(Severity::WARNING),
+                    (new Scope())->setExtra('order_id', $order->getKey())
+                );
+            } else {
+                $order->tracking_code = $reason;
+            }
```
nanaya: I wonder if this should be its own function in paypal processor instead (`updatePendingTrackingCode` or something)

app/Libraries/Payments/PaymentProcessor.php
```diff
@@ -211,9 +212,20 @@ public function pending()
...
+            $reason = $this['pending_reason'] === 'echeck' ? Order::PENDING_ECHECK : $this['pending_reason'];
+            // Log warning if non echeck pending status will overwrite the echeck field.
+            if ($order->tracking_code === Order::PENDING_ECHECK && $order->tracking_code !== $reason) {
+                app('sentry')->getClient()->captureMessage(
+                    "Trying to overwrite pending echeck with {$reason}",
+                    new Severity(Severity::WARNING),
+                    (new Scope())->setExtra('order_id', $order->getKey())
+                );
+            } else {
+                $order->tracking_code = $reason;
```
nanaya: redundant with the assignment after this

app/Libraries/Payments/PaymentProcessor.php
```diff
@@ -211,9 +212,20 @@ public function pending()
...
             $order = $this->getOrder()->lockSelf();
-            // Only supported by Paypal processor atm, so assume eCheck.
-            // Change if the situation changes.
-            $order->tracking_code = Order::PENDING_ECHECK;
+
+            $reason = $this['pending_reason'] === 'echeck' ? Order::PENDING_ECHECK : $this['pending_reason'];
+            // Log warning if non echeck pending status will overwrite the echeck field.
+            if ($order->tracking_code === Order::PENDING_ECHECK && $order->tracking_code !== $reason) {
+                app('sentry')->getClient()->captureMessage(
+                    "Trying to overwrite pending echeck with {$reason}",
```
nanaya: more like "overwriting" rather than "trying to overwrite"

app/Libraries/Payments/PaymentProcessor.php
notbakaneko: more like the other line isn't supposed to be there 🙃

PR #12577, opened 2025-12-02
Add job to archive support related forum topics

nanaya:
From `update_forums.php`.

Changed the tag to be case insensitive.


app/Models/Forum/Topic.php
```diff
@@ -329,11 +331,12 @@ public function titleNormalized()
             return $title;
         }
 
+        $tags = [];
```
notbakaneko: I feel like this should be `static`...

notbakaneko (changes requested):

notbakaneko (approved):

PR #12582, opened 2025-12-04
Add ordering to quickplay pools listing

nanaya:
( ﾟ ヮﾟ)a

### DIFF
--- a/app/Http/Controllers/Ranking/MatchmakingController.php
+++ b/app/Http/Controllers/Ranking/MatchmakingController.php
@@ -23,13 +23,15 @@ class MatchmakingController extends Controller
         $rulesetName ??= default_mode();
         $rulesetId = Beatmap::MODES[$rulesetName] ?? abort(422, 'invalid ruleset parameter');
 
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Ranking/MatchmakingController.php
```diff
@@ -23,13 +23,15 @@ public function show(?string $rulesetName = null, ?string $poolId = null)
         $rulesetName ??= default_mode();
         $rulesetId = Beatmap::MODES[$rulesetName] ?? abort(422, 'invalid ruleset parameter');
 
+        $poolsQuery = MatchmakingPool::where(['ruleset_id' => $rulesetId])->orderByDesc('active')->orderByDesc('id');
```
notbakaneko, line 26: this isn't going to use the index because there's `variant_id` in the index and no `variant_id` in the query, is it? 🤔

nanaya: probably won't matter for a while since there's not supposed to be many pools at all

also it still uses the index... probably for the ruleset_id filter

notbakaneko (approved):

PR #12583, opened 2025-12-04
Fix News post notification not sent when directly syncing the page

notbakaneko:
Turns out that `afterCommit()` __does not__ run after `save` and runs immediately when not explicitly in a transaction, which causes the job to dispatch before it's saved.

`NewsPost::syncAll()` creates the record first and doesn't have this issue.

### DIFF
--- a/app/Models/NewsPost.php
+++ b/app/Models/NewsPost.php
@@ -414,11 +414,14 @@ class NewsPost extends Model implements Commentable, Wiki\WikiObject
 
     public function save(array $options = [])
     {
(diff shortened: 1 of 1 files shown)


nanaya:
wouldn't wrapping the save in transaction work then?

notbakaneko:
assuming we remember to do it everywhere...

Also going to be a case of "order of logic in save depends on thing from context of caller"; I'd rather have it not run if that's the case

nanaya:
I mean inside the overridden save function

notbakaneko:
I guess... 
wrapping `parent::save()` in transaction seemed weird to me so I change it to this 🤷

PR #12595, opened 2025-12-09
⚡️ Move deployment image to AlmaLinux and frankenphp

ThePooN:
Seeing great performance uplift in synthetic benchmarks with frankenphp against swoole in my production-like deployment.  

I have not run them on dedicated resources so I won't provide precise numbers, but I'm never seeing any less than 30% more req/sec, often more.  
Memory usage during the synthetic benchmarks is also twice lower, but the usage pattern is so different than real production usage, I'm not considering it any more than *encouraging*.

frankenphp requires PHP compiled with ZTS. There is only one repository providing such builds, that is FrankenPHP's own repos, but only their RPM repos allow us to bind ourselves to a specific PHP version. So I switched the image to be based on AlmaLinux 10 instead, which is very close to RHEL, with strong community adoption and ships all the packages we need, including multiple versions of Node.

Both AlmaLinux and the PHP repo support arm64. The image builds fine but I haven't tested it as I don't possess hardware on this platform. The build is so much faster though, thanks to dnf being so much faster than apt especially under ARM emulation.  
The image is also 190MB lighter so we'll see overall faster production deployments.

notbakaneko mentioned on Discord merging this may currently be blocked by some swoole-specific calls:
> note that only swoole supports Octane::tasks() that's being used by chat to send messages without blocking the worker on waiting for response

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -62,7 +62,7 @@ module.exports = function(config) {
       ChromeCustom: {
         base: 'ChromeHeadless',
         // chrome sandboxing does not work in every container runtime
-        flags: '--no-sandbox',
+        flags: ['--no-sandbox'],
       },
     },
     exclude: [],


ThePooN:
I have ensured that no loaded extension is missing other than swoole.  
```
# docker.io/pppy/osu-web:2025.1204.0
php > print_r(get_loaded_extensions());
Array
(
    [0] => Core
    [1] => date
    [2] => libxml
    [3] => openssl
    [4] => pcre
    [5] => zlib
    [6] => filter
    [7] => hash
    [8] => json
    [9] => pcntl
    [10] => random
    [11] => Reflection
    [12] => SPL
    [13] => session
    [14] => standard
    [15] => sodium
    [16] => mysqlnd
    [17] => PDO
    [18] => xml
    [19] => calendar
    [20] => ctype
    [21] => curl
    [22] => dom
    [23] => mbstring
    [24] => FFI
    [25] => fileinfo
    [26] => ftp
    [27] => gd
    [28] => gettext
    [29] => iconv
    [30] => igbinary
    [31] => intl
    [32] => exif
    [33] => mysqli
    [34] => pdo_mysql
    [35] => pdo_sqlite
    [36] => Phar
    [37] => posix
    [38] => readline
    [39] => shmop
    [40] => SimpleXML
    [41] => sockets
    [42] => sqlite3
    [43] => sysvmsg
    [44] => sysvsem
    [45] => sysvshm
    [46] => tokenizer
    [47] => xmlreader
    [48] => xmlwriter
    [49] => xsl
    [50] => zip
    [51] => redis
    [52] => swoole
    [53] => ds
    [54] => Zend OPcache
)

# ghcr.io/thepoon/osu-web:frankenphp-deployment-gha-test
php > print_r(get_loaded_extensions());
Array
(
    [0] => Core
    [1] => date
    [2] => libxml
    [3] => openssl
    [4] => pcre
    [5] => zlib
    [6] => calendar
    [7] => ctype
    [8] => curl
    [9] => dom
    [10] => json
    [11] => filter
    [12] => hash
    [13] => iconv
    [14] => SPL
    [15] => mbstring
    [16] => pcntl
    [17] => session
    [18] => standard
    [19] => Phar
    [20] => posix
    [21] => random
    [22] => readline
    [23] => Reflection
    [24] => exif
    [25] => SimpleXML
    [26] => sockets
    [27] => sodium
    [28] => tokenizer
    [29] => trader
    [30] => xml
    [31] => xmlreader
    [32] => xmlwriter
    [33] => FFI
    [34] => apcu
    [35] => ds
    [36] => fileinfo
    [37] => ftp
    [38] => gd
    [39] => gettext
    [40] => intl
    [41] => msgpack
    [42] => mysqlnd
    [43] => PDO
    [44] => shmop
    [45] => sqlite3
    [46] => sysvmsg
    [47] => sysvsem
    [48] => sysvshm
    [49] => xsl
    [50] => zip
    [51] => igbinary
    [52] => mysqli
    [53] => pdo_mysql
    [54] => pdo_sqlite
    [55] => redis
    [56] => Zend OPcache
)
```

~~As far as testing goes, I haven't done more than test whether~~ it builds, runs, can display some pages from an unseeded database and browse wiki/news on my production-like instance (can provide a link privately).  
I intend to rather use staging as a seeded test env.

edit: all phpunit and browser tests seem to pass using the analogous development image https://github.com/ppy/osu-web/pull/12596

peppy:
Is there any reading material available on how it achieves the 30% boost? I'd kinda want to know what's the deal there before going full steam ahead in this direction.

ThePooN:
FrankenPHP's author made a talk 3 years back when they announced the project, in which they mention they're heavily leveraging Go's very optimized net/http libraries, Goroutines, and PHP's ZTS mode.  
Swoole, despite using the similar worker model, is not using ZTS.

image

image

image

image

source: https://dunglas.dev/2022/10/frankenphp-the-modern-php-app-server-written-in-go/

ThePooN:
Have replaced jhead by exiv2 and jpegtran to allow upgrading to AlmaLinux 10.

We're building assets with Node.js 22 instead of 20 now, which seems fine as far as I can tell.

ThePooN:
Making frankenphp the default Octane server in config broke CI, as it wasn't running off the Docker environment. It failed as the `frankenphp` binary shipped by default misses critical extensions.

I refactored the workflow to use the Docker dev set-up instead, which should also make it much easier to maintain going forward, along side other fixes.

002love:
Our current Debian-based images are flagged for [CVE-2023-45853](https://security-tracker.debian.org/tracker/CVE-2023-45853) (Integer Overflow in zlib). I've audited `BeatmapsetArchive.php` to check for reachability; while our `libzip` layer currently mitigates this, staying on a vulnerable system library is a long-term risk 🤷‍♂️ 
Supporting this change

app/Libraries/ImageProcessor.php
```diff
@@ -63,7 +63,34 @@ public function purgeExif()
             return;
         }
 
-        exec('jhead -autorot -purejpg -q '.escapeshellarg($this->inputPath));
+        $exifRotation = exec('exiv2 --key Exif.Thumbnail.Orientation -P v '.escapeshellarg($this->inputPath));
```
nanaya: can use php's exif related function instead

app/Libraries/ImageProcessor.php
```diff
@@ -63,7 +63,34 @@ public function purgeExif()
...
+                break;
+            case 7:
+                $args = '-transverse';
+                break;
+            case 8:
+                $args = '-rotate 270';
+                break;
+        }
+
+        exec('jpegtran '.$args.' -trim -outfile '.escapeshellarg($this->inputPath).' '.escapeshellarg($this->inputPath));
```
nanaya: missing exif tag removal

Dockerfile.deployment
```diff
@@ -86,6 +96,7 @@ RUN yarn production
 
 RUN php artisan ip2asn:update
 RUN php artisan scribe:generate
+RUN php artisan octane:install
```
nanaya: this should just be run separately and have the resulting php file committed

.github/workflows/tests.yml
```diff
@@ -5,183 +5,83 @@ concurrency:
...
-
 jobs:
   tests:
     strategy:
       matrix:
         php: ['8.4']
     name: Tests
     runs-on: ubuntu-latest
+    env:
+      PHP_VERSION: ${{ matrix.php }}
```
nanaya: I don't think we'll ever support multiple php version so just drop this and the matrix and other related things

.github/workflows/tests.yml
```diff
@@ -5,183 +5,83 @@ concurrency:
...
-          tools: composer:v2
-          php-version: ${{ matrix.php }}
-          coverage: none
-
-      - name: Get composer cache directory
-        id: composercache
-        run: echo "dir=$(composer config cache-files-dir)" >> $GITHUB_OUTPUT
+      - name: Create .env
+        run: |
+          cat > .env <<EOF
```
nanaya, line 20: maybe make a `.env.github` or something instead

app/Libraries/ImageProcessor.php
```diff
@@ -63,7 +63,34 @@ public function purgeExif()
             return;
         }
 
-        exec('jhead -autorot -purejpg -q '.escapeshellarg($this->inputPath));
+        $exifRotation = exec('exiv2 --key Exif.Thumbnail.Orientation -P v '.escapeshellarg($this->inputPath));
+
+        $args = '';
+        switch ((int) $exifRotation) {
```
nanaya: use `$args = match (...) { };`

.github/workflows/tests.yml
```diff
@@ -5,183 +5,83 @@ concurrency:
...
+          APP_ENV=testing
+          APP_URL=http://nginx
+          GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }}
```
nanaya: this one needs to be on actual shell env, not just the .env (used for composer settings which doesn't use the file)

docker-compose.yml
```diff
@@ -15,6 +15,8 @@ x-web: &x-web
   build:
     context: ./docker/development
     dockerfile: ../../Dockerfile.development
+    args:
+      PHP_VERSION: "${PHP_VERSION}"
```
notbakaneko: this needs a default or something

nanaya: more like just get rid of it and set the version directly. including the gha matrix which I don't think we use at all anymore 

(a single source of version number could also work)

app/Libraries/ImageProcessor.php
nanaya: or maybe wait for this... https://bugzilla.redhat.com/show_bug.cgi?id=2432745

ThePooN: I sincerely doubt Red Hat will be interested in taking over security maintenance of someone's abandoned pet project long-term, when multiple better, widely-available and well-maintained alternatives exist. I see no point in sticking to jhead long-term.

.github/workflows/tests.yml
ThePooN: I find `.env.github` confusing. What do you think of `.env.ci`? or `.env.actions`?

Also, personally I'm not fond of adding a top-level file given it would serve no other purpose than this workflow. I see no harm in defining environment here as long as it's short and not duplicate.
Any opinion @notbakaneko?

docker-compose.yml
ThePooN: We can define PHP_VERSION in .env, and add an extra job in the tests workflow to define the matrix.  
See example in https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#using-an-output-to-define-two-matrices.

What do you think?

nanaya: if it's to be defined somewhere, it shouldn't be in env file since you can't expect people to figure out if the file need update at random times. also the version definition file should be used everywhere and not manually re-defined in multiple different places like how it currently is

and I just remembered the gha matrix thing has been pretty annoying in combination with branch protection rules whenever the version was updated (the matching job is gone and some manual fuckery is needed)

app/Libraries/ImageProcessor.php
ThePooN: Done.

IDEA says "'ext-exif' is missing in composer.json". I'm only seeing extensions not-typically-built-by-default in composer.json atm though, unlike exif. Let me know if I should add it.

app/Libraries/ImageProcessor.php
ThePooN: > missing exif tag removal

Added `-copy none`.

Dockerfile.deployment
ThePooN: Done. The artisan command adds it to gitignore though, which implies to me it could be changed in the future. Can revert.

app/Libraries/ImageProcessor.php
ThePooN: Done.

.github/workflows/tests.yml
ThePooN: Done.

docker-compose.yml
ThePooN: Ok, `.php-version` then? I've seen `.node-version` before :shrug:

Getting rid of the GHA matrix stuff.

nanaya: yeah if it works and simple enough

and might as well add the mention in the readme file

ThePooN: I have this for defining PHP version in a single place using `.php-version`, but honestly I'm not sure it's worth it given it's only in both Dockerfiles now: https://github.com/ThePooN/osu-web/commit/8eeb9f0d70bcbd8008f5a8871b7ac1975c59ad1f

nanaya: there's another one in gha lint file...

ThePooN:
I have rewritten the lint workflow to be Docker-based, just like the new test workflow: https://github.com/ppy/osu-web/commit/ead760e3ece4ccf2938f369e2fdaaf02e87c951b

This brings a more consistent environment, at the cost of speed, it now takes +90s to run (3mn total) due to many unnecessary dependencies. I don't think this is a big deal since the test workflow will always be slower.

docker-compose.yml
ThePooN: Nice catch...

Addressed with https://github.com/ppy/osu-web/pull/12595#issuecomment-3808013111 anyway. Solves the same concern for node version too.

nanaya:
I don't know about others but early lint fail means not wasting time waiting on something that can be fixed in seconds...

ThePooN:
That is true, is this recurrent though? Personally my IDE set up everything without any input from my end, so it's most likely I'd catch these before pushing.

I can revert this, but then I don't have a solution for having a single-source of node version. AlmaLinux 10 doesn't offer the option, unless we're switching to a 3rd-party repo. NodeSource has an RPM repo so I'll give it a try.

karma.conf.js
```diff
@@ -56,9 +56,15 @@ module.exports = function(config) {
...
     autoWatch: true,
     basePath: '.',
-    browsers: ['ChromeHeadless'],
     colors: true,
     concurrency: Infinity,
+    customLaunchers: {
+      ChromeCustom: {
+        base: 'ChromeHeadless',
+        // chrome sandboxing does not work in every container runtime
+        flags: process.env.CHROME_NO_SANDBOX ? ['--no-sandbox'] : [],
```
nanaya: don't bother making it configurable. there's none for the other browser test either

Dockerfile.development
```diff
@@ -1,44 +1,51 @@
...
+    (curl -fsSL https://rpm.nodesource.com/setup_$(cat /.node-version).x | bash -) && \
+    microdnf install -y \
+      chromedriver \
+      chromium \
+      frankenphp \
+      git \
+      libjpeg-turbo-utils \
+      mysql8.4 \
+      nodejs \
+      npm \
```
nanaya: it's just an alias for nodejs...?

nanaya:
oh yeah, maybe throw that TERMINFO thing in as well... (which doesn't seem to make sense)

karma.conf.js
ThePooN: Done

Dockerfile.development
ThePooN: Looks like that is the case in NodeSource indeed. Removed

nanaya (approved):

PR #12596, opened 2025-12-09
♻️ Move development image to AlmaLinux and frankenphp for consistency with deployment image

ThePooN:
To match with #12595.

~~Completely untested as I don't have an osu-web Docker-based dev environment set-up.~~  
~~Just opening this draft PR as a base for osu-web devs using such Docker set-up to pick up if #12595 gets merged.~~

Undrafted. I have used this image to bring up a dev environment, run PHPUnit and browser tests, both 100% passed effortlessly.


ThePooN:
Merged in #12595 as the scope of these PRs have expanded a bit.

PR #12599, opened 2025-12-10
Send news notifications on a schedule

### LINKED ISSUE #12597: News post notifications go out immediately ahead of scheduled news post release time
https://discord.com/channels/188630481301012481/1003689172420276246/1448047579093925999

notbakaneko:
Adds a command and schedule to send news notifications because sending them when they're posted doesn't quite work for those scheduled to be published in the future.

closes #12597

### DIFF
--- a/app/Models/NewsPost.php
+++ b/app/Models/NewsPost.php
@@ -27,9 +27,6 @@ use Illuminate\Database\Eloquent\Builder;
  * @property string|null $tumblr_id
  * @property \Carbon\Carbon|null $updated_at
  * @property string|null $version
- * @method static Builder default()
- * @method static Builder published()
- * @method static Builder year(?int $year)
  */
 class NewsPost extends Model implements Commentable, Wiki\WikiObject
 {
@@ -173,21 +170,21 @@ class NewsPost extends Model implements Commentable, Wiki\WikiObject
         }
     }
 
-    public function scopeDefault(Builder $query): Builder
+    public function scopeDefault(Builder $query): void
     {
-        return $query->published()->orderBy('published_at', 'DESC');
+        $query->published()->orderBy('published_at', 'DESC');
     }
 
-    public function scopePublished(Builder $query): Builder
+    public function scopePublished(Builder $query): void
     {
-        return $query->whereNotNull('published_at')
+        $query->whereNotNull('published_at')
             ->where('published_at', '<=', Carbon::now());
     }
 
-    public function scopeYear(Builder $query, ?int $year): ?Builder
+    public function scopeYear(Builder $query, ?int $year): void
     {
         if ($year === null) {
-            return null;
+            return;
         }
 
         $baseStart = Carbon::create($year);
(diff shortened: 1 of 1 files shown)


app/Models/NewsPost.php
```diff
@@ -170,21 +173,21 @@ public static function syncAll()
         }
     }
 
-    public function scopeDefault($query)
+    public function scopeDefault(Builder $query): Builder
```
nanaya: apparently scopes are supposed to return void (or more like the return isn't used)

database/migrations/2025_12_10_000000_add_count_last_news_post_published_at_notified.php
```diff
@@ -0,0 +1,28 @@
...
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        $last = Count::where('name', 'last_news_post_published_at_notified')->first();
+
+        if ($last === null) {
+            $lastestTimestamp = NewsPost::published()->last()?->published_at->timestamp;
+            Count::create(['name' => 'last_news_post_published_at_notified', 'count' => $lastestTimestamp ?? 0]);
```
nanaya: do this in the function itself? `::first([...]) ?? ...::create([...]);` (or some longer form)

tests/Models/NewsPostTest.php
```diff
@@ -86,10 +89,31 @@ public function testNotificationDelivery(array $options, ?string $series, bool $
...
+    {
+        $user = User::factory()->create();
+        $user->notificationOptions()->create([
+            'name' => UserNotificationOption::NEWS_POST,
+            'details' => ['push' => true, 'series' => null],
+        ]);
+
+        $this->expectCountChange(fn () => Notification::count(), 0);
+
+        NewsPost::factory()->create();
```
nanaya: explicitly set the published_at field?

app/Models/NewsPost.php
```diff
@@ -27,6 +27,9 @@
  * @property string|null $tumblr_id
  * @property \Carbon\Carbon|null $updated_at
  * @property string|null $version
+ * @method static Builder default()
```
nanaya: fwiw laravel 12 has [a new method for defining scopes](https://laravel.com/docs/12.x/eloquent#local-scopes) which probably isn't compatible with this...

database/migrations/2025_12_10_000000_add_count_last_news_post_published_at_notified.php
```diff
@@ -0,0 +1,26 @@
...
+use App\Models\NewsPost;
+use Illuminate\Database\Migrations\Migration;
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        Count::firstOrCreate(
+            ['name' => 'last_news_post_published_at_notified'],
+            ['count' => NewsPost::default()->first()?->published_at->timestamp ?? 0],
```
nanaya: I mean in the `Count` model function itself instead of creating a dedicated migration 🤨 

seems a bit silly adding more migration just for this

PR #12600, opened 2025-12-11
Allow renaming of teams' name & tag

### LINKED ISSUE #12548: teams: allow renaming existing team's name & tag
it seems we can't rename both name and tag of an existing team, forcing us to delete and re-create a new team if we wanna rename

is there any reasons behind this? would love to see this getting implemented

iwa:
fixes #12548 

i tried to tackle this one with basic features:
- allow renaming of both name & tag, with their own fields
- add a `name_changed_at` timestamp field in team's model
- adjustable cooldown (currently set to 7 days) to avoid rename spam
  - prevent renaming during cooldown
  - shows the remaining time before next rename
- add a `input-text--disabled` css class

<table>
<tr>
 <td>Before renaming</td>
 <td>After renaming</td>
<tr>
 <td>image</td>
 <td>
image
</td>
</table>


iwa:
one thing i'm not sure about is style, when on cooldown i've added a disabled attribute on both inputs, but idk if there's a specific style convention you use to visually determine a field as disabled?

iwa:
i've added a `--disabled` subclass to `input-text`, taking inspiration on what's been done in other parts of the project

haven't bound to `[disabled]` cause i don't want to cause side effects anywhere else

(updated pr's desc to show the new result)

peppy:
If we're even considering this, the cooldown should be more like 1 year. Or maybe a store purchase.

iwa:
if so then either cooldown or store purchase should also apply to creating a team to make it fair, right? but this doesn't really seem like a good idea considering teams are only a social feature
also team name doesn't feel as important as a username so i'm not sure ppl would be willing to pay exclusively for it

i guess 2 possible options could be:
- make a long cooldown as you suggested, applying both for creating & renaming
- lock managing a team (create & update) as a supporter perk, and keep a low or even no cooldown

PR #12602, opened 2025-12-12
Remove play count check for quick play ranking

nanaya:
It's now just a hint instead.


notbakaneko (approved):
the `*` doesn't look very visible, though..

PR #12610, opened 2025-12-17
Add yearly playcount table

nanaya:
This will be used for calculating percentile of a user's playcount for the year.

The actual data will be inserted separately, probably calculated in a different server and then inserted directly. Slight inaccuracies shouldn't affect the result too much.

The query:
```sql
insert into yearly_playcount (year,playcount)
  select 2025, sum(playcount)
  from osu_user_month_playcount
  where `year_month` between '2501' and '2512' group by user_id;
```

### DIFF
--- a/app/Models/YearlyPlaycount.php
+++ b/app/Models/YearlyPlaycount.php
@@ -1,0 +1,40 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+declare(strict_types=1);
+
+namespace App\Models;
+
+/**
+ * @property int $playcount
+ * @property int $year
+ */
+class YearlyPlaycount extends Model
+{
+    public $timestamps = false;
+    protected $table = 'yearly_playcount';
+
+    public static function getPosition(int $year, int $userId): array
+    {
+        $users = \Cache::remember(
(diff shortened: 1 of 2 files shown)


notbakaneko (approved):

app/Models/YearlyPlaycount.php
```diff
@@ -0,0 +1,40 @@
...
+            3600,
+            fn () => max(1, static::where('year', $year)->count()),
+        );
+
+        $yearString = substr((string) $year, 2, 2);
+        $playcount = (int) UserMonthlyPlaycount
+            ::where('user_id', $userId)
+            ->whereBetween('year_month', ["{$yearString}01", "{$yearString}12"])
+            ->sum('playcount');
+        $pos = static::where('year', $year)->where('playcount', '>', $playcount)->count();
```
notbakaneko, line 32: wait, is this going to use the table instead?

notbakaneko (changes requested):
I think the query to insert in the description is missing setting the user id?

app/Models/YearlyPlaycount.php
nanaya: the new table is just for the overall count. I suppose it's also possible to use it for per-user count but that means more columns and index and it's not all that difficult to re-count for single user anyway 🤷

notbakaneko: 🤔

notbakaneko (approved):

PR #12614, opened 2025-12-22
 Add user summary model

nanaya:
Still missing multiplayer stats which may or may not happen.

- [x] depends on #12610


Userlog2:
Just a thought: Maybe the amount of plays of Featured Artist maps could be displayed as a percentage (compared to all plays in total)?

app/Models/UserSummary.php
```diff
@@ -0,0 +1,392 @@
...
+            'combo' => 0,
+            'pp' => 0,
+            'score' => 0,
+        ];
+        foreach ($scores as $score) {
+            if ($summary['combo'] < $score->max_combo) {
+                $summary['combo'] = $score->max_combo;
+            }
+            if ($summary['score'] < $score->total_score) {
+                $summary['score'] = $score->total_score;
```
WitherFlower, line 290: I assume the intent here is to show the highest score with the standardized scoring, but I feel like for some players it's just gonna pull out a map with a low star rating that they happened to FC with a bunch of mods.

Maybe it would also be interesting to add the highest classic scoring play, as that's what counts towards the ranked score leaderboard.

peppy: Brought this up in internal feedback, feels like a very weird stat to show that doesn't carry much meaning.

notbakaneko (approved):
it works™️  and probably isn't going to remain in the same form next time anyway

PR #12618, opened 2025-12-24
Add API endpoint for uploading screenshots on lazer

LiquidPL:
Web side PR for ppy/osu#10718.

This implementation matches the behaviour of the respective endpoint in legacy web, since the screenshot retrieval endpoint is currently being handled by that. Ability to retrieve said screenshots using osu-web will come in a separate PR.

Steps to deploy:

- [ ] Configure the screenshot s3 bucket in the `S3_SCREENSHOT_BUCKET` env var
- [ ] Set the screenshot secret in the `SCREENSHOTS_SHARED_SECRET` env var


app/Http/Controllers/ScreenshotsController.php
```diff
@@ -0,0 +1,51 @@
...
+                'required',
+                File::types(['jpeg']),
+            ],
+        ]);
+
+        datadog_increment('osu.screenshots');
+
+        $screenshot = new Screenshot();
+        $screenshot->user_id = auth()->user()->getKey();
+        $screenshot->save();
```
nanaya: can be done in one step with `::create`. And use `\Auth::user()` instead of `auth()->user()`

app/Http/Controllers/ScreenshotsController.php
```diff
@@ -0,0 +1,51 @@
...
+
+        $this->storage()->putFileAs('/', $validated['screenshot'], "{$screenshot->getKey()}.jpg");
```
nanaya: should be done in the model

app/Http/Controllers/ScreenshotsController.php
```diff
@@ -0,0 +1,51 @@
...
+
+        return response()->json([
+            'url' => route('screenshots.show', [
+                'screenshot' => $screenshot->getKey(),
+                'hash' => substr(md5($screenshot->getKey().config('osu.screenshots.shared_secret')), 0, 4),
```
nanaya: should be done in the model. And use `$GLOBALS['cfg'][...]` instead of `config(...)`

database/factories/ScreenshotFactory.php
```diff
@@ -0,0 +1,20 @@
...
+
+use App\Models\Screenshot;
+
+class ScreenshotFactory extends Factory
+{
+    protected $model = Screenshot::class;
+
+    public function definition(): array
+    {
+        return []; // nothing in this model to fake
```
nanaya: user_id field?

app/Http/Controllers/ScreenshotsController.php
```diff
@@ -0,0 +1,51 @@
...
+            ]),
+        ], 201);
```
nanaya: 201 requires returning `location:` header when it's not the same url as the post request.

Or just return 200.

app/Http/Controllers/ScreenshotsController.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
```
nanaya, line 5: can use declare strict (also for all other files)

app/Http/Controllers/ScreenshotsController.php
```diff
@@ -0,0 +1,51 @@
...
+use Storage;
+
+class ScreenshotsController extends Controller
+{
+    public function store(Request $request)
+    {
+        $validated = $request->validate([
+            'screenshot' => [
+                'required',
+                File::types(['jpeg']),
```
nanaya: need a size limit of maybe 10MB (or `App\Libraries\Uploader::DEFAULT_MAX_FILESIZE`)

config/filesystems.php
```diff
@@ -97,6 +97,12 @@
             'visibility' => 'public',
         ],
 
+        'local-screenshot' => [
```
nanaya: before solo-replay (also for the other one)

config/filesystems.php
```diff
@@ -122,6 +128,12 @@
             ...$s3Default,
             'bucket' => presence(env('S3_SOLO_REPLAY_BUCKET')) ?? 'solo-scores-replays',
         ],
+
+        's3-screenshot' => [
+            ...$s3Default,
+            'bucket' => presence(env('S3_SCREENSHOT_BUCKET')) ?? 'screenshots',
```
nanaya, line 129: add to `.env.example` (also for the other one)

app/Http/Controllers/ScreenshotsController.php
```diff
@@ -0,0 +1,51 @@
...
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace App\Http\Controllers;
+
+use App\Models\Screenshot;
+use Illuminate\Contracts\Filesystem\Filesystem;
+use Illuminate\Http\Request;
+use Illuminate\Validation\Rules\File;
+use Storage;
```
nanaya: just use directly with `\Storage` instead of adding more lines

nanaya (changes requested):

tests/Browser/SanityTest.php
```diff
@@ -486,6 +489,9 @@ private function bindParams(LaravelRoute $route)
             'wiki.sitemap' => [
                 'locale' => 'en',
             ],
+            'screenshots.show' => [
+                 'hash' => md5(self::$scaffolding['screenshot']->getKey().config('osu.screenshots.shared_secret')),
```
nanaya: if only the model has a `urlHash()` function...

config/filesystems.php
```diff
@@ -118,6 +124,15 @@
             'region' => env('S3_CENTRAL_BUCKET_REGION'),
         ],
 
+        's3-screenshot' => [
+            ...$s3Default,
+            'bucket' => presence(env('S3_SCREENSHOT_BUCKET')) ?? 'screenshots',
+            'key' => env('S3_SCREENSHOT_KEY'),
+            'region' => env('S3_SCREENSHOT_REGION'),
+            'secret' => env('S3_SCREENSHOT_SECRET'),
```
nanaya: key and secret will use the same one as the others so they shouldn't need to be specified separately

LiquidPL: i wasn't sure if that's the case since i believe the screenshots bucket predates this repo

app/Http/Controllers/ScreenshotsController.php
```diff
@@ -0,0 +1,41 @@
...
+
+use App\Models\Screenshot;
+use Illuminate\Http\Request;
+use Illuminate\Validation\Rules\File;
+
+class ScreenshotsController extends Controller
+{
+    public function store(Request $request)
+    {
+        $validated = $request->validate([
```
nanaya, line 18: the exception thrown needs handling for correct status code in the App\Exceptions\Handler

nanaya (changes requested):

app/Http/Controllers/ScreenshotsController.php
LiquidPL: wait, I just looked through the repo and I realized that the project isn't using laravel's native validation anywhere, this PR would be the only instance of it. maybe in that case it's better to just handle the exception here and abort with 422, rather than try to plug laravel validation in there?

nanaya: just add the class to the status code handler

it used to be used somewhere else but nothing else uses it atm yeah

nanaya (approved):

PR #12624, opened 2025-12-27
Wrapped UI

notbakaneko:
I had to redo (more like make up) a lot of stuff upon discovering the page flow of original design didn't quite work once proper data starts getting displayed and the window isn't a fixed width _or_ height, and apparently there isn't a css way to "fit as many columns as possible in a grid but have the columns variable size and still line up and not have contents inside wrap/overflow" 

Most of it are in the same files so it's easier to remove.
There are a lot of css variables to make it easier to me to adjust the sizes ಠ_ಠ

- [x] #12614
- [x] #12619
- [x] whatever thing that actually outputs the json
- [x] better images for daily challenge and beatmapping pages (without repeating the same image too many times); the covers don't look very good - or maybe just roll up all the stats into the statistics page?

### DIFF
--- a/app/Http/Controllers/WrappedController.php
+++ b/app/Http/Controllers/WrappedController.php
@@ -61,7 +61,7 @@ class WrappedController extends Controller
                 new BeatmapCompactTransformer(),
                 ['beatmapset.user', 'owners'],
             ),
-            'share_link' => route('wrapped', ['userId' => $summary->user_id, 'share' => $summary->share_key]),
+            'share_link' => route('wrapped', ['user' => $summary->user_id, 'share' => $summary->share_key]),
             'summary' => $data,
             'user_id' => $summary->user_id,
         ];
--- a/routes/web.php
+++ b/routes/web.php
@@ -412,7 +412,7 @@ Route::group(['middleware' => ['web']], function () {
 
     Route::get('/', 'HomeController@index')->name('home');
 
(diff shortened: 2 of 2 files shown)


app/Http/Controllers/WrappedController.php
```diff
@@ -0,0 +1,30 @@
...
+    {
+        parent::__construct();
+    }
+
+    public function show($userId)
+    {
+        $user = User::findOrFail($userId);
+
+        // TODO: stuff actually needs to go in here from https://github.com/ppy/osu-web/pull/12614
+        $json = [
```
notbakaneko, line 53: actual json needs to go here

resources/js/entrypoints/wrapped-show.tsx
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import core from 'osu-core-singleton';
+import React from 'react';
+import { parseJson } from 'utils/json';
+import WrappedShow from 'wrapped-show';
+import sampleData from 'wrapped-show/sums/summary-8447637.json';
+
+const userId = 8447637;
```
notbakaneko: directly imported data for testing that isn't included in the PR

PR #12626, opened 2025-12-27
Adjust "account not in good standing" display thresholds

### LINKED ISSUE #12533: Make the "Account not in good standing" text appear only when recent infringements are bad enough
### Introduction

The "username's account is not in a good standing" text appears when a user has a silence of any length within the last 28 days. But does a chat silence lasting 5 minutes, which are used with the impact of warnings in practice, issued 22 days ago, mean that a user should think: "I'd better watch out, a couple more offences and my account is at serious risk..."? It doesn't, because this isn't how things operate in practice. 
Being implicitly told that your account is at risk for any minor infringement is [a bit much](https://github.com/ppy/osu-web/issues/12533#issuecomment-3569192311).

For non-moderators, this text is redundant, since the presence of the Account Standing tab indicates the same thing, with the exception of some tournament-banned users. The message is a hold-over from the old website design, where the "not good standing" warning would be displayed along with the actual amount of infringements until restriction.

However, this warning can be used as an additional factor of deterrence instead of being removed completely. Here's how:

### Proposal

The "not good standing" warning functions as a middle step between "any silences at all within the last month" (accomplished by the Account Standing tab's presence) and "official final warning" (accomplished manually by the GMT).

An example pseudocode implementation would be the following (the conditions are, of course, just an example and may be adjusted as seen fit):
```
if (recentInfringements >= 4 || (recentInfringements >= 3 && longestRecentSilenceMins >= 160) || longestRecentSilenceMins >= 2560) || activeUser.isModerator:
displayAccountStandingWarning()
```

### What good would implementing this do?

First and foremost, adding these qualifications required would **make the bad standing warning _actually mean something_ - and therefore, be taken more seriously**, since it wouldn't just appear for any minor infringement. This alone would be a worthwhile reason to make this change.

In terms of the practical benefits this would have, this would give users an extra warning that their infringement history is on the severe side, which may have a significant effect since it appears right where a user would go from "having made a mistake or two" to "being a persistent offender" - or for users with silences long enough, who should be taking their account standing seriously regardless of infringement count.

Notice how big the gap between "having been silenced within the last 28 days" and "receiving a formal final warning from the GMT" is otherwise - this would mean it isn't a small enough step to be insignificant, and putting a small "hey, watch out" sign there would be likely to have an effect on at least some users.

### Addressing potential issues and criticisms

This proposal **does not** make moderators have to act in a different way or alter GMT workflow, since it's something the user will see on the UI/UX side. **It is from the point of view of a silenced user, _not_ the point of view of a GMT**.

This also **does not make things significantly more complicated** for anyone after its implementation, since it would most likely involve adding only a few lines of code and change the conditions for showing one UI element, even if it wouldn't be quite as simple in practice as in my example. 

It wouldn't reveal any info the GMT doesn't want revealed either.

002love:
Resolves #12533

This PR implements logic to hide the "bad standing" warning for minor infractions, as discussed in the issue

Key changes:
- warning is hidden for regular users if recent silences are <24h and inactive
- moderators still see the warning if any history exists (preserving visibility for staff)
- threshold of 3+ infringements triggers the warning regardless of length

Includes PHPUnit tests covering all logic branches


app/Models/User.php
```diff
@@ -1075,6 +1075,43 @@ public function isModerator()
         return $this->isGMT() || $this->isNAT();
     }
 
+    /**
+     * Determine if the user has a "bad standing" based on recent infringements.
+     *
+     * @param User|null $viewer The user viewing the profile.
+     * @return bool
+     */
+    public function hasBadStanding(?User $viewer = null): bool
```
LiquidPL: change the name to `inBadStanding` to be more in line with wording on the website

app/Models/User.php
```diff
@@ -1075,6 +1075,43 @@ public function isModerator()
...
+        if ($silences->isEmpty()) {
+            return false;
+        }
+
+        if ($viewer?->isModerator()) {
+            return true;
+        }
+
+        foreach ($silences as $silence) {
+            if ($silence->endTime()->isAfter()) {
```
LiquidPL: `isAfter()` should have the timestamp passed into it

tests/Models/UserTest.php
```diff
@@ -269,4 +270,94 @@ public function testValidDiscordUsername(string $username, bool $valid)
             $this->assertArrayHasKey('user_discord', $user->validationErrors()->all());
         }
     }
+
+    public function testHasBadStandingForModeratorWithHistory()
+    {
+        config_set('osu.user.ban_persist_days', 28);
```
LiquidPL: i don't think these need to be set explicitly in the tests, `28` is already provided by a default value in `config/osu.php`.

app/Models/User.php
```diff
@@ -1075,6 +1075,43 @@ public function isModerator()
...
+            ->where('ban_status', UserAccountHistory::TYPES['silence'])
+            ->get();
+
+        if ($silences->isEmpty()) {
+            return false;
+        }
+
+        if ($viewer?->isModerator()) {
+            return true;
+        }
```
LiquidPL: handling the moderator case in the model is a bad idea - the model just passes data from the database, it shouldn't be aware of the request state. IMO this method should always return the bad standing state based on the silence length/count, and the mod override should be handled in the presentation layer.

`UserTransformer` already returns the account history list to the frontend, so you could determine if the good standing blurb should be displayed over there.

tests/Models/UserTest.php
```diff
@@ -269,4 +270,94 @@ public function testValidDiscordUsername(string $username, bool $valid)
...
+
+    public function testHasBadStandingForModeratorCleanUser()
+    {
+        config_set('osu.user.ban_persist_days', 28);
+        $viewer = User::factory()->withGroup('gmt')->create();
+        $user = User::factory()->create();
+
+        // no history
+        $this->assertFalse($user->hasBadStanding($viewer));
+    }
```
LiquidPL: if moderator case is removed from `hasBadStanding` the respective tests should be removed as well

app/Models/User.php
002love: fixed

app/Models/User.php
002love: fixed

tests/Models/UserTest.php
002love: fixed

app/Models/User.php
002love: fixed

tests/Models/UserTest.php
002love: fixed

002love:
Thanks for the feedback, applied all requested changes:
- renamed `hasBadStanding` to `inBadStanding` in the model and moved the moderator logic to the UserTransformer 
- refactored `isAfter()` to `isFuture()`
- cleaned up `config_set` calls in unit tests

Userlog2:
Thanks for taking the time to implement my idea!

One minor point of criticism, though: It's redundant (and in my opinion, slightly counterproductive) to show the "bad standing" warning if a user has an active silence because:

1. There is another equally obvious info message showing that the user is silenced,
2. The user *is silenced* and feeling the full effects of it; an extra red text isn't going to make it much more noticeable when they already can't chat, comment, etc.
3. It makes the purpose of the "bad standing" warning less clear. It's intended for severe/repeat offenders, so flashing it at a minor chat offender for 5 minutes could be a bit confusing.

app/Models/User.php
```diff
@@ -1075,6 +1075,38 @@ public function isModerator()
...
+        foreach ($silences as $silence) {
+            if ($silence->endTime()->isFuture()) {
+                return true;
+            }
+        }
+
+        $recentCount = $silences->count();
+        $longestPeriod = $silences->max('period') ?? 0;
+
+        if ($longestPeriod >= 1440 * 60 || $recentCount >= 3) {
```
Userlog2: Currently, silences are factors of 5 minutes. Any silence longer than a day would, in practice, be at least 2560 minutes (42h40m) long.
```suggestion
        if ($longestPeriod >= 2560 * 60 || $recentCount >= 3) {
```

app/Models/User.php
```diff
@@ -1075,6 +1075,38 @@ public function isModerator()
...
+        $silences = $this->accountHistories()
+            ->recent()
+            ->where('ban_status', UserAccountHistory::TYPES['silence'])
+            ->get();
+
+        if ($silences->isEmpty()) {
+            return false;
+        }
+
+        foreach ($silences as $silence) {
```
Userlog2: It's redundant and not necessary to always display this warning if the user has an active silence, as explained in [my comment](https://github.com/ppy/osu-web/pull/12626#issuecomment-3699119806).

app/Models/User.php
002love: Fixed

app/Models/User.php
002love: Fixed

Userlog2: If I'm interpreting this code properly, this makes the warning *never* be shown if the user has an active silence. I meant that the "bad standing" warning should be independent of whether the silence is active or not.

Also, you may want to update the "key changes" list in the main page.

LiquidPL: Yeah, I just realized that an active silence will always hide the warning, even if it's length would normally make it visible.

LiquidPL (changes requested):

app/Models/User.php
```diff
@@ -1075,6 +1075,38 @@ public function isModerator()
         return $this->isGMT() || $this->isNAT();
     }
 
+    /**
+     * Determine if the user has a "bad standing" based on recent infringements.
+     *
+     * @return bool
+     */
+    public function inBadStanding(): bool
```
nanaya: this should be done in the frontend instead of doing redundant query (which happens on all page access due to inclusion in transformer for current user json)

nanaya (changes requested):

app/Models/User.php
002love: fixed

Userlog2:
An active silence shouldn't *hide* the account standing warning either, since the warning should be independent of whether the silence is active. Please remove lines 97 to 102.

resources/js/profile-page/account-standing.tsx
```diff
@@ -78,6 +79,31 @@ const content: Record<Column, (props: ColumnProps) => JSX.Element | null> = {
...
 @observer
 export default class AccountStanding extends React.Component<ExtraPageProps> {
+  @computed
+  get inBadStanding() {
+    const user = this.props.controller.state.user;
+    const history = user.account_history;
+
+    if (core.currentUser?.is_moderator && history.length > 0) {
+      return true;
+    }
```
nanaya: pretty sure this should check for silence, not entire history?

resources/js/profile-page/account-standing.tsx
```diff
@@ -78,6 +79,31 @@ const content: Record<Column, (props: ColumnProps) => JSX.Element | null> = {
...
+    const silences = history.filter((d) => d.type === 'silence');
+
+    if (silences.length === 0) {
+      return false;
+    }
+
+    const recentCount = silences.length;
+    const longestPeriod = Math.max(...silences.map((s) => s.length));
+
+    if (longestPeriod >= 2560 * 60 || recentCount >= 3) {
```
nanaya: where did this 2560 come from? seems like a weird number to have (in term of time).

And just return the check instead of doing the equivalent of `if (condition) return true; return false`

nanaya (changes requested):

resources/js/profile-page/account-standing.tsx
Userlog2: 2560 minutes = 42 hours 40 minutes ≈ 43 hours, the length of a "standard" long silence. Silence lengths are calculated with a simple exponential function, which in this case is 5min * 2⁹. [Further context here.](https://osu.ppy.sh/community/forums/topics/2138362?n=12)

nanaya: that should be put in the comment

resources/js/profile-page/account-standing.tsx
002love: yep, fixed

resources/js/profile-page/account-standing.tsx
002love: fixed

resources/js/profile-page/account-standing.tsx
```diff
@@ -78,6 +79,27 @@ const content: Record<Column, (props: ColumnProps) => JSX.Element | null> = {
 
 @observer
 export default class AccountStanding extends React.Component<ExtraPageProps> {
+  @computed
+  get inBadStanding() {
+    const user = this.props.controller.state.user;
+    const history = user.account_history;
+    const silences = history.filter((d) => d.type === 'silence');
+
+    if (core.currentUser?.is_moderator && silences.length > 0) {
```
nanaya: can use an is_admin too

resources/js/profile-page/account-standing.tsx
```diff
@@ -6,6 +6,7 @@ import StringWithComponent from 'components/string-with-component';
 import TimeWithTooltip from 'components/time-with-tooltip';
 import UserLink from 'components/user-link';
 import UserAccountHistoryJson from 'interfaces/user-account-history-json';
+import core from 'osu-core-singleton';
```
nanaya: need some lint fix both php and js

resources/js/profile-page/account-standing.tsx
002love: fixed

resources/js/profile-page/account-standing.tsx
002love: fixed

nanaya (approved):

PR #12628, opened 2025-12-29
Improve search usability for guest users

### LINKED ISSUE #12620: Search while logged out has some usability concerns
Image

002love:
Fixes #12620

PR improves the search experience for non-authenticated users by making the "login required" notices more prominent and extending this restriction to team search

Before | After 
-- | --
image | Screenshot 2025-12-29 at 01 30 37


resources/css/bem/search-result.less
```diff
@@ -95,6 +95,16 @@
     &--notice {
       font-size: @font-size--title-small;
       text-align: center;
+
+      .textual-button {
```
nanaya: that's not how bem works. also, just use plain `<a>` instead of `<button>`.

002love: fixed

resources/css/bem/search-result.less
```diff
@@ -109,4 +109,14 @@
...
   }
+
+  &__login-link {
+    .link-default();
+    .link-plain();
+    font-weight: 700;
+
+    .link-hover({
+      text-decoration: underline;
+    });
```
nanaya: most of these are already the default for `a` (and some are even conflicting with each other). And I don't think the font-weight is needed either so this whole styling can be removed.

nanaya (changes requested):

resources/css/bem/search-result.less
002love: fixed

resources/views/home/_search_results.blade.php
```diff
@@ -5,9 +5,9 @@
 <div class="search-result search-result--{{ $mode }}">
     @if (!auth()->check() && $search->isLoginRequired())
         <div class="search-result__row search-result__row--notice">
-            <button class="textual-button textual-button--inline js-user-link">
+            <a href="{{ route('login') }}" class="js-user-link">
```
nanaya: the route isn't usable in this way. use `#` like in other places

nanaya (changes requested):

PR #12629, opened 2025-12-29
Fix blackout height issue on most mobile browsers

### LINKED ISSUE #12497: Height of the backdrop for the mobile navigation menu is broken.
While the nav dropdown menu is open, if I scroll down, due to chrome's own url navbar going off the screen, the height of the mobile navbar menu is too short to cover the whole screen and only recalculates after the scroll animation is done.

Chrome version: `141.0.7390.122`

https://github.com/user-attachments/assets/64ddc088-8fbf-4e3e-9b23-47723a7f327f

UnhappyF:
Resolves #12497 

Overrides blackout height to prevent jumpy effect after viewport size change when browser searchbar goes off screen. Аffects blackout element only for mobile and only when its visible.

However, there are still issues with viewport size on Safari ios 26, but seems like we cant do anything with it until Apple fixes their browser

| Before        | After         |
|---------------|---------------|
| <video src="https://github.com/user-attachments/assets/6210ff1c-d7bb-4166-ab3d-54316f508386"/>   | <video src="https://github.com/user-attachments/assets/078d9a8b-f85c-493d-bbeb-b9ac62539858"/>   |


resources/css/bem/blackout.less
```diff
@@ -13,4 +13,11 @@
     z-index: @z-index--overlay;
     opacity: 0.4;
   }
+
+  //prevent jumpy height resize when search bar hides on mobile browsers
+  @media @mobile {
+    &[data-visibility="visible"] {
+      height: 100lvh;
+    }
+  }
```
LiquidPL: it might be better to edit the `full-size()` helper to use the dynamic viewport units, since the blackout overlay is fullscreen on desktop as well.

besides, i think that `dvh`/`dvw` should be used instead, since the overlay should always conform to the current viewport size, and not the largest one.

LiquidPL (changes requested):

resources/css/bem/blackout.less
UnhappyF: We can't apply this to `.full-size` because it uses `position: absolute`. It stretches the element to the parent size, and viewport units would break the site. 
`.blackout` overrides it to `position: fixed`, so we can safely override the height using viewport units. 

`dvh` doesn't solve the problem, as most browsers don't seem to recalculate the viewport size before the search bar animation ends, or do it fast enough. Yes, it seems like not an ideal solution, because it’s not an ideal problem. In the future, we can change it to `dvh` when mobile browsers solve their issues.

LiquidPL: yeah, that's a good point, also i failed to realize that `full-size` is being used in contexts where an element doesn't span the entire viewport so this wouldn't be possible anyway. let's just leave it as it is for now.

LiquidPL (approved):

resources/css/bem/blackout.less
nanaya: put it in the main block style instead of weirdly specific to mobile/visible?

UnhappyF: I didn’t want to change the element more than the issue requires.
`.blackout` is always present on the page, even when it’s not visible. On mobile devices with the address bar open, there will always be an invisible fixed element extending beyond the viewport. This doesn’t affect anything, but it feels a bit odd. That’s why I’d prefer to adjust the height only when it’s actually needed.

If you think it’s fine, I can move it to the main block.

nanaya: that's a lot of codes for something that's not visible. and if anything, it currently changes size whenever the address bar changes state which sounds worse...? don't bother

UnhappyF: done

nanaya (approved):

nanaya:
@UnhappyF please allow maintainer edit or merge/rebase with master.

PR #12631, opened 2025-12-29
Fix contradiction in API documentation regarding Client Credentials scopes

### LINKED ISSUE #12611: The osu!web Documentation regarding the Client Credentials Grant might need to be updated
The [Scopes](https://osu.ppy.sh/docs/index.html#scopes) documentation mentions that some scopes are exclusive to the Client Credentials Grant, but the [Client Credentials Grant](https://osu.ppy.sh/docs/index.html#client-credentials-grant) documentation states that the `scope` must be `public` as other scopes have no meaningful effect.

002love:
Fixes #12611

Documentation currently claims that `public` is the only meaningful scope for Client Credentials, which is misleading

In reality, scopes like `delegate`, `forum.write_manage`, and `group_permissions` are supported and used by bots. I've updated the description to list these and added a note that delegatable scopes can't be mixed with `public` in a single request

This was also recently discussed in [openiddict-core #2403](https://github.com/openiddict/openiddict-core/pull/2403)


resources/views/docs/auth.blade.php
```diff
@@ -382,7 +382,7 @@
                 'example' => 'client_credentials',
             ],
             'scope' => [
-                'description' => 'Must be `public`; other scopes have no meaningful effect.',
+                'description' => 'A space-delimited string of [scopes](#scopes). Supports `public` and [delegation](#client-credentials-delegation) scopes (like `delegate` or `forum.write_manage`). Note: delegatable scopes cannot be combined with non-delegatable ones (like `public`).',
```
notbakaneko:
```suggestion
                'description' => 'A space-delimited string of [scopes](#scopes). Only `public` and scopes that allow [delegation](#client-credentials-delegation) are supported.
```
Extra repeating of the scopes in the linked section is unnecessary

resources/views/docs/auth.blade.php
```diff
@@ -478,7 +478,8 @@
 
 <p>
     Client Credentials Grant tokens may be allowed to act on behalf of the owner of the OAuth client (delegation) by requesting the {{ ApidocRouteHelper::scopeBadge('delegate') }} scope, in addition to other scopes supporting delegation.
-    When using delegation, scopes that support delegation cannot be used together with scopes that do not support delegation.
+    When using the Client Credentials grant with scopes that support delegation, all requested scopes must support delegation.
+    You cannot mix delegatable and non-delegatable scopes in a single request.
```
notbakaneko: Prefer using `delegation` as we're using it as a specific terminology and not `delegatable` (and `delegable` is the correct spelling, anyway).

...this change also seems unnecessary when it's directly under `Client Credentials Delegation` and the context of the line immediately before it.

notbakaneko (changes requested):

resources/views/docs/auth.blade.php
002love: fixed

notbakaneko (approved):

Issue #9864, opened 2023-02-15
Proposal: Giving "osu! original" tracks the attention they deserve.


MChecaH:
## Context
In the last few years, there has been a surge of commissioned tracks specifically made to be used for osu! beatmapsets. This includes but is not limited to [exclusive Featured Artist releases](https://osu.ppy.sh/wiki/en/osu%21_originals#featured-artist-exclusive-releases), [official tournament commissioned works](https://osu.ppy.sh/wiki/en/osu%21_originals#official-tournament-releases), and even [community ran Mapping tournaments](https://osu.ppy.sh/wiki/en/osu%21_originals#community-run-contest-releases). A full list of tracks can be found in [here](https://osu.ppy.sh/wiki/en/osu%21_originals#list-of-osu!-originals).

https://github.com/ppy/osu-wiki/commit/5762eb8e08168710205db325dd0f06fb390f5bbe By the end of 2022, there have been active efforts by wiki maintainers and The Team™ to update this listing. But I believe that only mentioning the tracks in a separate wiki article is not giving the proper credit that these tracks should have. It feels disjointed with the ecosystem at large. They should be given a far better spotlight being content made **by** the game **for** the game. And the current best approach would be a more integrated system in the Beatmap Listing to highlight this kind of content [a la Featured Artist](https://osu.ppy.sh/beatmapsets?c=featured_artists).

## Proposal
This proposal wants to start discussion for a potential three-pronged approach to this problem:

### __osu!web__ - [[Beatmap Listing]](https://osu.ppy.sh/beatmapsets?c=featured_artists&s=any)
- **New `osu! original` label:** Exactly the same as existing labels like Explicit, Spotlight, and Featured Artist. These would quickly highlight those beatmapsets that are using this content. Eye candy that might also catch player's attention! I'll get to this later, but this label should **not** be a replacement to the current Featured Artist label depending on the definition given to this new category.
- **New `osu! originals` filter:** Exactly the same as the Featured Artist implementation. Quick button to quickly sort all beatmapsets using this content.

<p align="center">
<img src="https://mch.s-ul.eu/f6ayYIV6" width=85% height=85%>
</p>

### __osu!wiki__ - [[Ranked Criteria]](https://osu.ppy.sh/wiki/en/Ranking_Criteria#metadata)
[...]

PR #12634, opened 2025-12-30
Add "osu! original" beatmap badge and search filter

Hiviexd:
image

---

Resolves https://github.com/ppy/osu-web/issues/9864.

The osu! original badge is prioritized over the FA one in terms of rendering because it'd be redundant given all original tracks are FA, alongside the limited beatmap card space.

Also made the explicit badge link to its relevant wiki page.

### DIFF
--- a/app/Libraries/Search/BeatmapsetSearch.php
+++ b/app/Libraries/Search/BeatmapsetSearch.php
@@ -68,6 +68,7 @@ class BeatmapsetSearch extends RecordSearch
         $this->addFeaturedArtistFilter($query);
         $this->addFeaturedArtistsFilter($query);
         $this->addFollowsFilter($query);
+        $this->addExclusiveTracksFilter($query);
         $this->addGenreFilter($query);
         $this->addLanguageFilter($query);
         $this->addExtraFilter($query);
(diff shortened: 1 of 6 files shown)


ghost:
Not sure if this is also necessary but what if  you add osu! originals button next to Featured Artist button as shown on the screenshot below



Maybe its easier for players to try to find their favorite osu! original song

app/Transformers/BeatmapsetCompactTransformer.php
```diff
@@ -82,6 +82,7 @@ public function transform(Beatmapset $beatmapset)
             'title' => $beatmapset->title,
             'title_unicode' => $beatmapset->title_unicode,
             'track_id' => $beatmapset->track_id,
+            'is_exclusive_track' => $beatmapset->track?->exclusive ?? false,
```
nanaya, line 85: this just can't be here as it'll cause one additional query for each beatmapset

app/Libraries/Search/BeatmapsetSearch.php
```diff
@@ -174,6 +175,14 @@ private function addFeaturedArtistsFilter($query)
         }
     }
 
+    private function addExclusiveTracksFilter($query)
+    {
+        if ($this->params->showExclusiveTracks) {
+            $trackIds = ArtistTrack::where('exclusive', true)->pluck('id');
```
nanaya, line 181: maybe add another field to the es index instead of passing hundreds of id every time

Issue #12617, opened 2025-12-22
Display upload requirements in contest submission box


Walavouchey:
should display
- max file size
- allowed file extensions
- required image resolution
- anything else that would error

this is an issue especially in art contests. artists don't know about the 8 mb upload limit and the contest page only let's them know about it in an error popup after trying to upload an entry exceeding the limit (we sometimes get distressed complaints about this when people try and fail submitting this way minutes before the deadline)

peppy:
I think we can start by putting them in the contest description (if that's not already the case), since that displays right above the upload box.

PR #12635, opened 2025-12-31
Rework allowed extensions handling in contest submission

### LINKED ISSUE #12616: Ability to whitelist file extensions for contest submissions
for mapping contests, people submitting `.osu` files instead of `.osz` files is a constant pain point, needing manual fixing each time so they can work with our contest tooling (to anonymise entries)

`.osu` file submissions are problematic because if the metadata isn't perfect, what audio file the submission is supposed to be paired up with can become ambiguous

sometimes, mapping contests have required multiple beatmaps for single submissions ([example](https://osu.ppy.sh/community/contests/248)). the optimal way to handle this would be `.zip` files with multiple `.osz` files inside (enforced by external scripts as always), but since `.zip` submissions aren't a thing, the workaround has been template `.osz` archives containing multiple audio files and beatmap difficulties

an option to configure which file extensions the submission box allows would remove this jank

Hiviexd:
Resolves #12616 via:

1. Removing `.osu` from default extensions, as that's practically never wanted as a type of submission.
2. Adding ability to override allowed file extensions via `allowed_extensions_override` extra option.


app/Http/Controllers/ContestEntriesController.php
```diff
@@ -145,26 +145,29 @@ public function store()
 
         priv_check('ContestEntryStore', $contest)->ensureCan();
 
-        $allowedExtensions = [];
+        $defaultExtensions = [];
```
notbakaneko: I'm not sure it makes sense calling this default with it being repopulated all the time. We should probably move this all to a `const` as well since each case is always the same value... (then it can be default)

resources/js/contest-entry/uploader.coffee
```diff
@@ -34,16 +34,19 @@ export class Uploader extends React.Component
...
+        defaultExtensions = ['osz']
+        maxSize = 32 * 1024 * 1024
 
       when 'music'
-        allowedExtensions = ['.mp3']
-        maxSize = 16*1024*1024
+        defaultExtensions = ['mp3']
+        maxSize = 16 * 1024 * 1024
+
+    rawExtensions = @props.contest.allowed_extensions_override ? defaultExtensions
```
notbakaneko: same deal as the php, check `allowed_extensions` first

app/Http/Controllers/ContestEntriesController.php
```diff
@@ -145,26 +145,29 @@ public function store()
...
                 $maxFilesize = 32 * 1024 * 1024;
                 break;
             case 'music':
-                $allowedExtensions[] = 'mp3';
+                $defaultExtensions[] = 'mp3';
                 $maxFilesize = 16 * 1024 * 1024;
                 break;
         }
 
+        $allowedExtensions = $contest->getAllowedExtensionsOverride() !== null
```
notbakaneko: `getAllowedExtensions` first, then defaults if `null`

app/Models/Contest.php
```diff
@@ -444,6 +444,11 @@ public function getForcedHeight()
         return $this->getExtraOptions()['forced_height'] ?? null;
     }
 
+    public function getAllowedExtensionsOverride()
```
notbakaneko: just `allowed_extensions` is enough

app/Http/Controllers/ContestEntriesController.php
```diff
@@ -145,26 +145,29 @@ public function store()
...
+            ? array_map('strtolower', $contest->getAllowedExtensionsOverride())
```
notbakaneko: also the `strtolower` on an option we're setting ourselves seems unnecessary

app/Models/Contest.php
```diff
@@ -444,6 +444,11 @@ public function getForcedHeight()
         return $this->getExtraOptions()['forced_height'] ?? null;
     }
 
+    public function getAllowedExtensions()
+    {
+        return $this->getExtraOptions()['allowed_extensions'] ?? null;
```
notbakaneko: I think just return the defaults from here instead

tests/Models/ContestTest.php
```diff
@@ -230,4 +239,18 @@ public function testShowJudges(?bool $showJudgesOption, bool $result): void
         ]);
         $this->assertSame($result, $contest->show_judges);
     }
+
+    /**
+     * @dataProvider dataProviderForTestAllowedExtensionsOverride
+     */
+    public function testAllowedExtensionsOverride(?array $allowedExtensionsOverride, ?array $result): void
```
notbakaneko: this test looks like it's just testing if `A = A` 🤔

app/Models/Contest.php
Hiviexd: done, also refactored max filesize in the same manner so I can use it for https://github.com/ppy/osu-web/issues/12617

tests/Models/ContestTest.php
Hiviexd: refactored the test so it wouldn't be useless

resources/lang/en/contest.php
```diff
@@ -73,7 +73,7 @@
 
         'wrong_type' => [
             'art' => 'Only .jpg and .png files are accepted for this contest.',
-            'beatmap' => 'Only .osu files are accepted for this contest.',
+            'beatmap' => 'Only .osz files are accepted for this contest.',
```
nanaya: should be updated to handle the configurable allowed extensions instead

resources/lang/en/contest.php
```diff
@@ -70,13 +70,7 @@
...
         'drop_here' => 'Drop your entry here',
         'download' => 'Download .osz',
-
-        'wrong_type' => [
-            'art' => 'Only .jpg and .png files are accepted for this contest.',
-            'beatmap' => 'Only .osz files are accepted for this contest.',
-            'music' => 'Only .mp3 files are accepted for this contest.',
-        ],
-
+        'wrong_type' => 'Only :types files are accepted for this contest.',
```
notbakaneko: use a different key, iirc, recycling the key with a different format causes problems with lang.js

PR #12636, opened 2025-12-31
Add transformer and endpoint for team details

LiquidPL:
Prerequisite for ppy/osu#32584.

I've noticed that there was already a `TeamStatisticsTransformer` that is to my knowledge used in the team rankings pages. It is inconsistent with the API conventions in some places (specifically using ruleset IDs instead of string values). I figure it's not part of the public API, but I didn't want to refactor it before confirming first.


app/Http/Controllers/TeamsController.php
```diff
@@ -153,7 +180,20 @@ public function show(string $id, ?string $ruleset = null): Response
...
-        return ext_view('teams.show', compact('rulesetId', 'statistics', 'team'));
+        if (is_api_request()) {
+            $jsonTeam = [
+                ...json_item($team, new TeamExtendedTransformer()),
+                'statistics' => json_item($statistics, new TeamStatisticsTransformer(), ['rank']),
+            ];
+
+            // remove stuff from the statistics object since it's already in the main team object
+            unset($jsonTeam['statistics']['ruleset_id']);
+            unset($jsonTeam['statistics']['team_id']);
```
nanaya: I don't think this is done anywhere. Also ruleset_id doesn't exist in the main team object.

Don't bother.

app/Http/Controllers/TeamsController.php
```diff
@@ -153,7 +180,20 @@ public function show(string $id, ?string $ruleset = null): Response
 
         $team->loadMissing(prefix_strings('members.user.', UserCompactTransformer::CARD_INCLUDES_PRELOAD));
 
-        return ext_view('teams.show', compact('rulesetId', 'statistics', 'team'));
+        if (is_api_request()) {
+            $jsonTeam = [
+                ...json_item($team, new TeamExtendedTransformer()),
+                'statistics' => json_item($statistics, new TeamStatisticsTransformer(), ['rank']),
```
nanaya: use transformer includes?

app/Transformers/TeamStatisticsTransformer.php
```diff
@@ -6,10 +6,11 @@
 namespace App\Transformers;
 
 use App\Models\TeamStatistics;
+use League\Fractal\Resource\Primitive;
 
 class TeamStatisticsTransformer extends TransformerAbstract
 {
-    protected array $availableIncludes = ['member_count', 'team'];
+    protected array $availableIncludes = ['rank', 'member_count', 'team'];
```
nanaya: alphabetical sort (also in various other places)

app/Transformers/TeamTransformer.php
```diff
@@ -20,4 +28,19 @@ public function transform(Team $team): array
...
     }
+
+    public function includeLeader(Team $team): Item
+    {
+        return $this->item($team->leader, new UserCompactTransformer());
+    }
+
+    public function includeMembersCount(Team $team): Primitive
+    {
+        return $this->primitive($team->members()->count());
```
nanaya: should exclude restricted users. also the empty slots thing already loads the `members` relation and this will just do another one

tests/api_routes.json
```diff
@@ -1685,5 +1685,21 @@
             "App\\Http\\Middleware\\RequireScopes"
         ],
         "scopes": []
+    },
+    {
+        "uri": "api/v2/teams/{team}/{ruleset?}",
```
nanaya: use `artisan route:convert --to-json=tests/api_routes.json` to update this file

nanaya (changes requested):

app/Transformers/TeamTransformer.php
```diff
@@ -8,9 +8,27 @@
...
 class TeamTransformer extends TransformerAbstract
 {
+    protected array $availableIncludes = [
+        'empty_slots',
+        'leader',
+        'members_count',
+        'statistics',
+    ];
+
+    protected int $rulesetId = 0; // osu
```
nanaya: this should probably be the team's default ruleset id?

app/Http/Controllers/TeamsController.php
```diff
@@ -127,6 +131,29 @@ public function part(string $id): Response
...
+     * Get Team
+     *
+     * This endpoint returns the details of a specified team.
+     *
+     * ---
+     *
+     * ### Response format
+     *
+     * Returns the [TeamExtended](#teamextended) object.
+     * The following [optional attributes on Team](#team-optionalattributes) are included:
```
nanaya: the link doesn't work

also the exact same includes is always applied to the extended object so it's redundant here

nanaya (changes requested):

app/Http/Controllers/TeamsController.php
LiquidPL: it works on the docs page (unless there's some way to make it work in the docstring as well)

i'll remove the includes though

nanaya: I mean it doesn't work in the docs page because there's no corresponding id for the section

LiquidPL: image

i was under the assumption that one is created automatically based on markdown headings

nanaya: That's not the line I commented on (`team-optionalattributes`)

app/Transformers/TeamTransformer.php
```diff
@@ -8,16 +8,56 @@
...
 class TeamTransformer extends TransformerAbstract
 {
+    protected array $availableIncludes = [
+        'empty_slots',
+        'leader',
+        'members_count',
+        'statistics',
+    ];
+
+    protected ?int $rulesetId;
```
nanaya: now I think about it, it doesn't really make sense to use team default if not set (although can still work by making a `getRulesetId` function and requiring team as parameter).

Maybe make it non-nullable here and don't set anything by default (it'll just error on first read-before-set which is probably fatal anyway)

app/Transformers/TeamTransformer.php
```diff
@@ -8,16 +8,56 @@
...
+    public function setRulesetId(int $rulesetId): static
+    {
+        $this->rulesetId = $rulesetId;
+
+        return $this;
+    }
+
     public function transform(Team $team): array
     {
+        $this->rulesetId = $team->default_ruleset_id ?? 0;
```
nanaya: this is very wrong...

LiquidPL: yeah i didn't want to do this in the first place, there's not really a place in the transformer to set a default like this (especially since they can be reused)

nanaya: this line overrides whether already set. as I mentioned elsewhere it could be done correctly but probably not very useful

Issue #10241, opened 2023-05-24
re-vitalizing the beatmap packs system


Hiviexd:
currently, packs are currently maintained by only one person who regularly creates standard packs through a UI from 13 years ago. and with said person being on the verge of resigning, alongside other pack categories being basically unused/abandoned, i'd like to propose a set of changes that will revive this system and actually make it usable:

- expose the .create-pack command to more users 
   - this can be accomplished by creating a #packs channel and role in internal
- add a reliable way to edit/delete packs so we can correct any mistakes/problems that occur during the creation process
   - an example use case is correcting a pack upload that used an invalid id/id of an undownloadable beatmap (e.g. DMCA'd)
- form a small but dedicated group of volunteers who will be creating/managing beatmap packs
   - i'm willing to step up and form/manage said group
   - this group will be responsible of creating new packs for most current categories, handle pack requests, create medal packs on a regular schedule, with more tasks that can be determined as we go
   - if the group establishes a certain level of permanence, they could be granted a display usergroup (like spotlights) as an incentive/form of recognition, which has basically proven integral for osu!'s volunteer groups
- add searching and filtering to beatmap packs page
   - ability to assign tags to beatmap packs upon creation/editing
   - search bar and filters similar to beatmap/featured artist listing
- add a section in homepage that displays the latest 5 created packs
   - section would be somewhere under the "new ranked beatmaps" one
   - this would be a key factor in shining more light towards the packs system

peppy:
I dunno, if we're going to build on it any more than what it currently is, changes need to be made to how it works.

- Pack downloads should be done by the client
- They should not require any special server storage - it would just be a collection of set IDs that the client is downloading
- Users should be able to make them and share them on their profile.

peppy:
A lot of this has already been completed.

Could be good to convert the OP to checklist items and check off all the completes ones.

Hiviexd:
ive made this into a proper tasklist + added issues that were already mentioned in the past for better tracking

PR #12637, opened 2025-12-31
Show beatmap pack tags in beatmap info page

Hiviexd:
image

---

Field was added in https://github.com/ppy/osu-web/pull/9822 but never actually used, and showing pack tags on the beatmap info page seems like a simple change that makes sense to have.

Relevant to https://github.com/ppy/osu-web/issues/10241 I think?


resources/js/interfaces/beatmapset-json.ts
```diff
@@ -116,6 +116,7 @@ interface BeatmapsetJsonDefaultAttributes {
   id: number;
   nsfw: boolean;
   offset: number;
+  pack_tags: string[];
```
nanaya: this belongs in available includes, not here.

resources/js/beatmapsets-show/info.tsx
```diff
@@ -228,6 +228,27 @@ export default class Info extends React.Component<Props> {
...
+                <div>
+                  {this.controller.beatmapset.pack_tags.map((pack, i) => (
+                    <React.Fragment key={`${pack}-${i}`}>
+                      <a
+                        className='beatmapset-info__link'
+                        href={route('packs.show', pack)}
+                      >
+                        {pack}
+                      </a>
+                      {i < this.controller.beatmapset.pack_tags.length - 1 && ', '}
```
nanaya: a bit simpler check if moved to before the link with `i !== 0`

resources/js/beatmapsets-show/info.tsx
```diff
@@ -228,6 +228,27 @@ export default class Info extends React.Component<Props> {
                 </div>
               </div>
             }
+
+            {this.controller.beatmapset.pack_tags && this.controller.beatmapset.pack_tags.length > 0 &&
```
nanaya: update the BeatmapsetJsonForShow typing instead

nanaya (changes requested):

resources/js/interfaces/beatmapset-json.ts
```diff
@@ -148,6 +149,7 @@ export function deletedBeatmapset(): BeatmapsetJson {
     id: 0,
     nsfw: false,
     offset: 0,
+    pack_tags: [],
```
nanaya: not needed
```suggestion
```

PR #12638, opened 2025-12-31
Validate Shopify redirection hosts

002love:
Ensures that redirects to external Shopify URLs are validated against a whitelist to prevent open redirect vulnerabilities

Via:
- Adding host validation in `StoreController@getInvoice` (restricting to `*.ppy.sh` and `*.myshopify.com`)
- Implementing `validateAndRedirect` helper in `store.ts` to sanitize frontend checkout redirects

### DIFF
--- a/resources/js/store.ts
+++ b/resources/js/store.ts
@@ -111,7 +111,7 @@ export class Store {
     };
 
     await $.post(route('store.checkout.store'), params);
-    window.location.href = data.cartCreate.cart.checkoutUrl;
+    this.validateAndRedirect(data.cartCreate.cart.checkoutUrl);
   }
 
   resumeCheckout(event: ClickEvent) {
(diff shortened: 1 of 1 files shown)


peppy:
I don't understand why this is required. Is it assuming that we don't trust shopify at this point? Because the URLs are coming from signed shopify requests.

002love:
Fair enough. My concern was mainly about preventing the endpoint from being used as a *bouncer* for arbitrary redirects, regardless of how much we trust the source

If you feel the signed requests from Shopify provide sufficient security and this adds unnecessary complexity, feel free to close it. I just thought it would be a solid *safe design* check to ensure the domain is never used for off-platform redirection 🤷‍♂️

peppy:
will defer to @ppy/team-web to merge or close this. i'm fine with either.

nanaya:
I don't think the specific domain is even documented...?

(and guarding against shopify doesn't seem useful when they also control the allowed redirect domain)

PR #12648, opened 2026-01-05
Standardise display order of playcount / favourites

### LINKED ISSUE #12644: Beatmap favorites and playcount order inconsistency
Beatmap cards have favorites first and then playcount, whereas it's the opposite for beatmap info pages. This is mostly noticeable on the mobile site where these stats are always next to each other. On desktop, they are on top of each other in beatmap cards, but technically still in reverse order.

Image
Image

someonecantcode:
Closes #12644 

I think it'd be best to change the beatmapset-panel order in the listing instead as when you hover the tooltip on the beatmap, it puts a pop up with people the right.

beatmapset-panel listing solution:
Image

how'd it look if we switched the tooltips + hover (not implemented):

image


peppy:
While we're here, what's with this icon alignment?

Safari 2026-01-05 at 04 47 23

someonecantcode:
hi, I'm not really sure what you meant by that. i think it's just how the heart in font awesome looks like? The heart is smaller than the circles.  i was able to align them if i scaled up the heart by 1.1 though; it's very hacky though.

i looked back at the image posted in the issue, current website, and testing and it seems that there is a small gap with the favourite icon in general.

website:
image

testing:
image

issue provided image:
image

someonecantcode:
@nanaya I wasn't sure what order you wanted it. 

The current website seems to link up favourites + hype and plays + nominations. It also seems that there should be two rows.
 I'm just going to make it plays + favourites + date for the first row and hype + nominations to make it consistent with the small panel order.
current website order:
image

image

current implementation 1 row example:
image

nanaya:
maybe just make it the same order as compact but with date on the second row instead?

alternatively maybe something like this
```
(as per compact item order)
1 3 5
2 4
```

someonecantcode:
kk. I did the alternative one. 

| master | this commit | 
|---------|--------------|
| image | image

someonecantcode:
alr changed to this
image

nanaya (approved):

PR #12650, opened 2026-01-05
Show pinned livestream on home dashboard

### LINKED ISSUE #10309: show pinned streams in `/community/livestreams` on the home page
https://osu.ppy.sh/community/livestreams has this feature where moderators can pin streams which gives them extra space in that page

thing is, this feature is super under-used right now, i found out about this randomly and i reckon most moderators aren't aware of it as well.

this feature would be EXTREMELY useful if pinned streams were also displayed on the homepage, so streams can actually be highlighted instead of them staying hidden in a low-traffic page.

use cases for this would be pinning world cup streams, official and unofficial special events/contests, etc...

Hiviexd:
image

Resolves #10309

Right now any moderator can pin/unpin streams ([source](https://github.com/Hiviexd/osu-web/blob/96da5b1a908a40867ff72916d69f9688b3fa899e/app/Singletons/OsuAuthorize.php#L1825-L1834)), not sure if we want that to remain the case..


app/Models/LivestreamCollection.php
```diff
@@ -12,13 +12,28 @@
...
+        }
+
+        $collection = new static();
+
+        foreach ($collection->all() as $stream) {
+            if ($stream->data['id'] === (string) $id) {
+                Cache::forever(static::FEATURED_CACHE_KEY, (string) $id);
+                Cache::put(static::FEATURED_DATA_CACHE_KEY, $stream->data, 300);
+                return;
+            }
```
nanaya: this whole part doesn't seem needed?

resources/css/bem/user-home-livestream.less
```diff
@@ -0,0 +1,21 @@
...
+  background-color: @osu-colour-b4;
+  box-sizing: border-box;
+
+  &__player {
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+  }
```
nanaya: not needed

resources/css/bem/user-home.less
```diff
@@ -43,6 +42,31 @@
...
+    align-items: flex-end;
+
+    @media @desktop {
+      padding-left: 0;
+    }
+  }
+
+  &__livestream-section {
+    width: 100%;
+    box-sizing: border-box;
```
nanaya: what's this for? also the other one

resources/views/home/_user_livestream.blade.php
```diff
@@ -0,0 +1,15 @@
+{{--
+    Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+    See the LICENCE file in the repository root for full licence text.
+--}}
+<?php
+    $playerId = 'home-livestream-'.time().'-'.rand();
```
nanaya: can be inlined

app/Models/LivestreamCollection.php
```diff
@@ -12,13 +12,28 @@
...
+            Cache::forget(static::FEATURED_DATA_CACHE_KEY);
+            return;
+        }
+
+        $collection = new static();
+
+        foreach ($collection->all() as $stream) {
+            if ($stream->data['id'] === (string) $id) {
+                Cache::forever(static::FEATURED_CACHE_KEY, (string) $id);
+                Cache::put(static::FEATURED_DATA_CACHE_KEY, $stream->data, 300);
```
nanaya: just let the `featured` function do this instead? this specific function can just
```
Cache::forever(KEY, value);
Cache::forget(DATA);
```

app/Models/LivestreamCollection.php
```diff
@@ -74,13 +89,24 @@ public function featured()
...
-                    return $stream;
-                }
+        if ($featuredStreamId === null) {
+            return null;
+        }
+
+        $cachedStreamData = Cache::get(static::FEATURED_DATA_CACHE_KEY);
+        if ($cachedStreamData !== null) {
+            return new Twitch\Stream($cachedStreamData);
+        }
```
nanaya: this should probably be the first thing in this function

app/Models/LivestreamCollection.php
```diff
@@ -74,13 +89,24 @@ public function featured()
...
-                if ($stream->data['id'] === $featuredStreamId) {
-                    return $stream;
-                }
+        if ($featuredStreamId === null) {
+            return null;
+        }
+
+        $cachedStreamData = Cache::get(static::FEATURED_DATA_CACHE_KEY);
+        if ($cachedStreamData !== null) {
+            return new Twitch\Stream($cachedStreamData);
```
nanaya: why not store the object itself?

app/Models/LivestreamCollection.php
```diff
@@ -12,13 +12,28 @@
 class LivestreamCollection
 {
     const FEATURED_CACHE_KEY = 'featuredStream:arr:v2';
+    const FEATURED_DATA_CACHE_KEY = 'featuredStream:data:v2';
```
nanaya: :v2 not needed here

resources/css/bem/user-home.less
```diff
@@ -43,6 +42,31 @@
...
+  }
+
+  &__news {
+    padding: 0 0 20px;
```
nanaya: looks like this belongs in the `__left-section`? And both `__news` and `__livestream-section` don't seem to be doing anything useful

resources/css/bem/user-home.less
Hiviexd: remnant from an older attempt >.>

will cleanup (same as the other one)

resources/js/core/twitch-player.tsx
```diff
@@ -34,6 +34,7 @@ export default class TwitchPlayer {
     const options = {
       channel: div.dataset.channel ?? fail(`twitch player ${div.id} is missing channel data`),
       height: '100%',
+      muted: div.dataset.muted === 'true',
```
nanaya: should be autoplay option instead. I think it's fine to not mute by default.

and need typing update for the option. Maybe either inline the call (`new Player(id, { ... }`) or explicitly type the assignment to make the check actually work.

app/Models/LivestreamCollection.php
```diff
@@ -12,13 +12,28 @@
...
+
+        $collection = new static();
+
+        foreach ($collection->all() as $stream) {
+            if ($stream->data['id'] === (string) $id) {
+                Cache::forever(static::FEATURED_CACHE_KEY, (string) $id);
+                Cache::forget(static::FEATURED_DATA_CACHE_KEY);
+                return;
+            }
+        }
```
nanaya: it's fine to assume the id is correct and just do both right away

app/Models/LivestreamCollection.php
```diff
@@ -72,15 +87,26 @@ public function download($api)
 
     public function featured()
     {
+        $cachedStream = Cache::get(static::FEATURED_DATA_CACHE_KEY);
```
nanaya: looking again, this should be `Cache::many([FEATURED_KEY, FEATURED_DATA])` instead since there probably won't be any featured livestream most of the time

resources/views/home/_user_livestream.blade.php
```diff
@@ -0,0 +1,12 @@
+{{--
+    Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+    See the LICENCE file in the repository root for full licence text.
+--}}
+<div class="user-home-livestream">
+    <div
```
nanaya: this whole extra div should be removed

resources/css/bem/user-home.less
```diff
@@ -32,8 +32,7 @@
     gap: 10px;
   }
 
-  &__news {
-    padding: 0 0 20px;
+  &__left-section {
```
nanaya, line 35: where did this part go https://github.com/ppy/osu-web/pull/12650#discussion_r2707274058
```suggestion
  &__left-section {
    padding-bottom: 20px;
```

PR #12653, opened 2026-01-05
Add normalized accuracy field to UserStatistics

### LINKED ISSUE #12444: Two different accuracy formats coexist in the API
It is far from new so I figure it is already well-known, but I find it enough of an annoyance to open an issue

It's possible through the API to get the `accuracy` of `Scores` through several routes, it will always be in a format where "1" is the maximum accuracy (displayed as 100% in-game/web)
It's also possible through the API to get the `hit_accuracy` of `UserStatistics` through several routes, it will always be in a format where "100" is the maximum accuracy

As someone who's in charge of an API wrapper, I'd definitely prefer if there was only a single format, preferably the one used by `Scores`

Would making any change on that front cause too many things to break, even if the benefits of such a change were to be locked behind a new `x-api-version` version?

UnhappyF:
Resolves #12444 

Currently, the API exposes accuracy in two different formats:
- [0..100] for `hit_accuracy` in `UserStatistics`
- [0..1] for `accuracy` in Score endpoints

Added a normalized [0..1] `accuracy` field to UserStatistics for consistency with the Score endpoints
Marked `hit_accuracy` as deprecated in documentation.


app/Models/UserStatistics/Model.php
```diff
@@ -57,6 +57,13 @@ public function getHitAccuracyAttribute($value)
         return $this->accuracy_new ?? round($this->accuracy * 100, 2);
     }
 
+    public function getAccuracyNormalizedAttribute()
```
nanaya: I think the spelling is supposed to be "normalised"

resources/js/interfaces/user-statistics-json.ts
```diff
@@ -25,6 +25,7 @@ interface UserStatisticsBaseJson {
   global_rank: number | null;
   grade_counts: Record<Grade, number>;
   hit_accuracy: number;
+  accuracy: number;
```
nanaya: might as well update the usage in profile page stats and rankings table (use proper percent formatter)

app/Models/UserStatistics/Model.php
UnhappyF: It looks like the project uses "normalize" a lot more than "normalise"
image

nanaya: but which ones are more recent

resources/js/interfaces/user-statistics-json.ts
UnhappyF: Done

For the profile page, it would be nice to change the translation key `users.show.stats.hit_accuracy` to `users.show.stats.accuracy`, but I'm not sure I can do this without conflicting with Crowdin.

app/Models/UserStatistics/Model.php
UnhappyF: Fixed

PR #12654, opened 2026-01-06
Limit wrapped link to relevant* users

nanaya:
\* only checks for registration date instead of actual playcount for the year.

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -10,6 +10,7 @@ use App\Http\Controllers\RankingController;
 use App\Libraries\Base64Url;
 use App\Libraries\LocaleMeta;
 use App\Models\LoginAttempt;
+use App\Models\UserSummary;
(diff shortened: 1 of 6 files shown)


app/helpers.php
```diff
@@ -2004,7 +2005,11 @@ function migration(string $name): Migration
     return require database_path("migrations/{$name}.php");
 }
 
-function has_viewed_wrapped(int $userId): bool
+function has_viewed_wrapped(): bool
 {
-    return request_attribute_remember('wrapped', fn () => App\Models\UserSummary::hasViewed($userId));
```
nanaya, line 2009: this one wouldn't even return correct value if called multiple times with different parameter 🤨

notbakaneko: ...what 🤨
(nvm I get it)

notbakaneko (approved):

PR #12623, opened 2025-12-26
Add endpoint for showing screenshots

LiquidPL:
Depends on #12618.

Steps to deploy:

- [ ] Set cutoff screenshot ID for "legacy" screenshots without hash in the `SCREENSHOTS_LEGACY_ID_CUTOFF` env var.
- [ ] [OPTIONAL] Switch production to use these endpoints instead of legacy web if everything works correctly.

PR #12664, opened 2026-01-07
Add endpoint for showing screenshots

LiquidPL:
Depends on #12618. Recreated the original PR (#12623) to pull in changes from base PR code review and some cleanup.

Steps to deploy:

- [x] Set cutoff screenshot ID for "legacy" screenshots without hash in the `SCREENSHOTS_LEGACY_ID_CUTOFF` env var.
- [x] [OPTIONAL] Switch production to use these endpoints instead of legacy web if everything works correctly.


app/Http/Controllers/ScreenshotsController.php
```diff
@@ -33,9 +34,34 @@ public function store(Request $request)
...
-        // TODO: move this logic over from legacy web
-        //       this empty method is left as a placeholder for the route used above to work
+        abort_if(Screenshot::urlHash(intval($id)) !== $hash, 404);
+
+        return $this->showBase($id);
+    }
+
+    public function showLegacy(int $id)
+    {
+        abort_if(!Screenshot::isLegacyId(intval($id)), 404);
```
nanaya: the routes can be combined with a lookup function in the model which does something like

```
if hash == null
  if id > legacy_id
    return null
else
  expectedHash = ...
  if !hash_equals(expectedHash, hash)
    return null

return findOrFail(...)
```

app/Http/Controllers/ScreenshotsController.php
```diff
@@ -33,9 +34,34 @@ public function store(Request $request)
...
+        return $this->showBase($id);
+    }
+
+    private function showBase($id)
+    {
+        $screenshot = Screenshot::findOrFail($id);
+        $screenshot->update([
+            'hits' => $screenshot->hits + 1,
+            'last_access' => Carbon::now(),
+        ]);
```
nanaya: should use `incrementInstance('hits', 1, ['last_access' => ...])`

app/Http/Controllers/ScreenshotsController.php
```diff
@@ -33,9 +34,34 @@ public function store(Request $request)
...
+
+        $file = $screenshot->fetch();
+
+        abort_if(!$file, 404);
+
+        return response()->stream(function () use ($file) {
+            echo $file;
+        }, 200, ['Content-Type' => 'image/jpeg']);
```
nanaya: need cache-control header too (it's missing in the current one)

app/Http/Controllers/ScreenshotsController.php
```diff
@@ -33,9 +34,34 @@ public function store(Request $request)
...
+    {
+        $screenshot = Screenshot::findOrFail($id);
+        $screenshot->update([
+            'hits' => $screenshot->hits + 1,
+            'last_access' => Carbon::now(),
+        ]);
+
+        $file = $screenshot->fetch();
+
+        abort_if(!$file, 404);
```
nanaya: doesn't really matter here but generally should avoid casting php string to boolean (explicitly check with `=== null` instead)

tests/Controllers/ScreenshotsControllerTest.php
```diff
@@ -7,22 +7,104 @@
...
+        self::assertEquals($screenshot->last_access, $after->last_access);
+    }
+
+    public function testShowLegacy()
+    {
+        \Storage::fake('local-screenshot');
+
+        $user = User::factory()->create();
+        $screenshot = Screenshot::factory()->create(['user_id' => $user->getKey()]);
+        \Storage::disk('local-screenshot')->putFileAs('/', UploadedFile::fake()->image('ss.jpg'), "{$screenshot->getKey()}.jpg");
```
nanaya: use the existing store function?

tests/Controllers/ScreenshotsControllerTest.php
```diff
@@ -7,22 +7,104 @@
...
+        ]))->assertOk()->assertHeader('Content-Type', 'image/jpeg');
+    }
+
+    public function testShowLegacyIdAboveCutoff()
+    {
+        config_set('osu.screenshots.legacy_id_cutoff', 727000);
+
+        $this->get(route('screenshots.show-legacy', [
+            'screenshot' => 727001,
+        ]))->assertNotFound();
```
nanaya, line 105: this doesn't really tell whether it's because of the cutoff or the screenshot is just missing

tests/Controllers/ScreenshotsControllerTest.php
```diff
@@ -7,22 +7,104 @@
...
+
+        $this->get(route('screenshots.show', [
+            'screenshot' => $screenshot->getKey(),
+            'hash' => Screenshot::urlHash($screenshot->getKey()),
+        ]))->assertOk();
+
+        $screenshot = $screenshot->refresh();
+
+        self::assertEquals(1, $screenshot->hits);
+        self::assertEqualsUpToOneSecond(Carbon::now(), $screenshot->last_access);
```
nanaya: make it part of the general show test

tests/Controllers/ScreenshotsControllerTest.php
```diff
@@ -7,22 +7,104 @@
...
+    {
+        $user = User::factory()->create();
+        $screenshot = Screenshot::factory()->create(['screenshot_id' => 727, 'user_id' => $user->getKey()]);
+
+        $this->get(route('screenshots.show', [
+            'screenshot' => 727,
+            'hash' => 'asdf',
+        ]))->assertNotFound();
+
+        $after = $screenshot->refresh();
```
nanaya: `refresh()` updates and returns the same object itself so the test below will always true regardless. and anyway use `expectCountChange` instead

tests/Controllers/ScreenshotsControllerTest.php
```diff
@@ -7,22 +7,104 @@
...
-        ])->assertOk()->assertJson(['url' => 'http://localhost/ss/1/9784']); // md5("1"."1234567890abcd")
+        ])->assertOk()->assertJson(fn (AssertableJson $json) =>
+            $json->whereType('url', 'string'))->json('url');
+
+        $matches = [];
+
+        $this->assertGreaterThan(0, preg_match('/https?:\/\/.+\/(\d+)\/(.{4})/', $url, $matches));
+
+        \Storage::disk('local-screenshot')->assertExists("{$matches[1]}.jpg");
+        $this->assertEquals(Screenshot::urlHash(intval($matches[1])), $matches[2]);
```
nanaya: use assertSame

tests/Controllers/ScreenshotsControllerTest.php
```diff
@@ -7,22 +7,104 @@
...
+
+        $url = $this->postJson(route('api.screenshots.store'), [
             'screenshot' => UploadedFile::fake()->image('screenshot.jpg'),
-        ])->assertOk()->assertJson(['url' => 'http://localhost/ss/1/9784']); // md5("1"."1234567890abcd")
+        ])->assertOk()->assertJson(fn (AssertableJson $json) =>
+            $json->whereType('url', 'string'))->json('url');
+
+        $matches = [];
+
+        $this->assertGreaterThan(0, preg_match('/https?:\/\/.+\/(\d+)\/(.{4})/', $url, $matches));
```
nanaya: use a different regexp delimiter so `/` doesn't need to be escaped. No need to match `https?` (or explicitly match the start), but the end should probably be matched explicitly with `$`

tests/Controllers/ScreenshotsControllerTest.php
```diff
@@ -7,22 +7,104 @@
...
+
+        \Storage::disk('local-screenshot')->assertExists("{$matches[1]}.jpg");
```
nanaya: or get the object and check its fetch result instead of relying on specific file structure

app/Models/Screenshot.php
```diff
@@ -31,24 +31,34 @@ class Screenshot extends Model
...
     public function url(): string
     {
         return route('screenshots.show', [
             'screenshot' => $this->getKey(),
-            'hash' => $this->urlHash(),
+            'hash' => self::urlHash($this->getKey()),
         ]);
     }
 
-    public function urlHash(): string
```
nanaya, line 47: there's only one place where this is used without the object itself so maybe keep this while calling the static function with a different name

(if anything, the `url()` function is called exactly once although I suppose it could've been used in tests as well)

.env.example
```diff
@@ -90,6 +90,7 @@ SLACK_ENDPOINT=https://myconan.net/null/
 
 # S3_SCREENSHOT_BUCKET=
 # SCREENSHOTS_SHARED_SECRET=
+# SCREENSHOTS_LEGACY_ID_CUTOFF=
```
nanaya: I forgot in the other pr but nowadays the example should have the default value

tests/Controllers/ScreenshotsControllerTest.php
LiquidPL: i don't think it's supposed to care about a screenshot (not) existing under that id, legacy code instantly bails if `id > cutoff` (and there is no hash)

LiquidPL: this is more to test that the legacy code path exits out immediately in that specific case, and i'm not sure if we should give any sign to the user that there is any screenshot under that id.

app/Models/Screenshot.php
LiquidPL: removed all uses of the static method from outside the model and made it private

tests/Controllers/ScreenshotsControllerTest.php
nanaya: without the actual model, removing the cutoff check in the model will still pass this test

tests/Controllers/ScreenshotsControllerTest.php
```diff
@@ -7,22 +7,99 @@
...
+
+        $this->get(route('screenshots.show', [
+            'screenshot' => $matches[1],
+            'hash' => $matches[2],
+        ]))->assertOk();
+    }
+
+    public function testShow()
+    {
+        $user = User::factory()->create();
```
nanaya: apart for the store test, this doesn't do anything

app/Models/Screenshot.php
```diff
@@ -31,22 +31,54 @@ class Screenshot extends Model
...
+        return substr(md5($id.$GLOBALS['cfg']['osu']['screenshots']['shared_secret']), 0, 4);
+    }
+
+    public static function lookup(int $id, ?string $hash): ?self
+    {
+        if ($hash === null) {
+            if (!self::isLegacyId($id)) {
+                return null;
+            }
+        } else if (self::hashForId($id) !== $hash) {
```
nanaya: use `hash_equals` when comparing secrets

app/Models/Screenshot.php
```diff
@@ -31,22 +31,54 @@ class Screenshot extends Model
...
-            'hash' => $this->urlHash(),
+            'hash' => $this->hash(),
         ]);
     }
 
-    public function urlHash(): string
+    public function hash(): ?string
     {
-        return substr(md5($this->getKey().$GLOBALS['cfg']['osu']['screenshots']['shared_secret']), 0, 4);
+        if (self::isLegacyId($this->getKey())) {
```
nanaya: I don't think this is needed

tests/Controllers/ScreenshotsControllerTest.php
```diff
@@ -7,22 +7,99 @@
...
-    public function testScreenshot()
+    public function testStore()
     {
-        Storage::fake('local-screenshot');
-
         $user = User::factory()->create();
         $this->actAsScopedUser($user);
 
-        $this->postJson(route('api.screenshots.store'), [
+        \Storage::fake('local-screenshot');
```
nanaya: might as well put this in `setUp`

nanaya (approved):

PR #12669, opened 2026-01-11
Remove user profile cover from watchlist pages

### LINKED ISSUE #12572: Some watchlist pages use the profile banner instead of the generic asset
Just a small issue that I noticed.

For some reason, the "[mapper](https://osu.ppy.sh/home/follows/mapping)" and "[comment](https://osu.ppy.sh/home/follows/comment)" watchlist pages use the logged-in profile banner.

Doesn't seem intentional since all of the other pages use the same default banner ("forum topic", "beatmap discussion", even the account settings page). Or maybe it is...?

(The friends page also uses the profile banner, but that looks fine and intentional.)

UnhappyF:
Resolves #12572

Removes profile cover from "mapper" and "comment" watchlists
All watchlist pages now use a generic image.

### DIFF
--- a/resources/js/entrypoints/follows-comment.tsx
+++ b/resources/js/entrypoints/follows-comment.tsx
@@ -7,5 +7,5 @@ import * as React from 'react';
 import { parseJson } from 'utils/json';
 
 core.reactTurbolinks.register('follows-comment', () => (
-  <Main follows={parseJson('json-follows-comment')} user={core.currentUserOrFail} />
+  <Main follows={parseJson('json-follows-comment')} />
 ));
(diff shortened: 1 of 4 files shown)


notbakaneko:
Actually, I think it might be supposed to be the other way; the newer personalized settings-related pages have the banner and the older ones haven't been updated.

UnhappyF:
Do we need to remove profile covers from watchlist pages, or add them where they are missing? @peppy

UnhappyF:
I found that Blade pages in this section already have `js-current-user-cover` in the header, but the cover is not set on initial render. It also affects the "dashboard" and "account settings" pages. I'm not sure if the "dashboard" should use the user cover. Could someone who is working on the project comment on this, please? @notbakaneko @nanaya

nanaya:
they all probably used user cover at one point but something broke along the way

UnhappyF:
Superseded by #12702

PR #12671, opened 2026-01-13
Add 'ranked_play' column types

smoogipoo:
Adds values for the upcoming ranked play, and an extra change that allows MM rooms to be created with no playlist, because ranked play is going to be optimised to not require playlist items prepared.

### DIFF
--- a/app/Http/Controllers/ContestEntriesController.php
+++ b/app/Http/Controllers/ContestEntriesController.php
@@ -145,25 +145,8 @@ class ContestEntriesController extends Controller
 
         priv_check('ContestEntryStore', $contest)->ensureCan();
 
-        $allowedExtensions = [];
-        $maxFilesize = 0;
-        switch ($contest->type) {
-            case 'art':
-                $allowedExtensions[] = 'jpg';
-                $allowedExtensions[] = 'jpeg';
-                $allowedExtensions[] = 'png';
-                $maxFilesize = 8 * 1024 * 1024;
-                break;
-            case 'beatmap':
-                $allowedExtensions[] = 'osu';
-                $allowedExtensions[] = 'osz';
-                $maxFilesize = 32 * 1024 * 1024;
-                break;
-            case 'music':
-                $allowedExtensions[] = 'mp3';
-                $maxFilesize = 16 * 1024 * 1024;
-                break;
-        }
+        $allowedExtensions = $contest->getAllowedExtensions();
+        $maxFilesize = $contest->getMaxFilesize();
 
         if (!in_array(strtolower($file->getClientOriginalExtension()), $allowedExtensions, true)) {
             abort(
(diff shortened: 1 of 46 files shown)


app/Models/Multiplayer/Room.php
```diff
@@ -74,14 +74,14 @@ class Room extends Model
     const TYPE_GROUPS = [
         'playlists' => [self::PLAYLIST_TYPE],
         'realtime' => self::REALTIME_STANDARD_TYPES,
-        'quickplay' => [self::MATCHMAKING_TYPE],
+        'quickplay' => self::MATCHMAKING_TYPES,
     ];
 
     const PLAYLIST_TYPE = 'playlists';
-    const MATCHMAKING_TYPE = 'matchmaking';
+    const MATCHMAKING_TYPES = ['matchmaking', 'ranked_play'];
```
smoogipoo, line 81: Leaving `matchmaking` as the type for "quick play" for now for backwards compatibility.

app/Models/Multiplayer/Room.php
```diff
@@ -457,7 +457,11 @@ public function isRealtime()
 
     public function isMatchmaking()
     {
-        return $this->type === static::MATCHMAKING_TYPE;
+        static $matchmakingTypes;
+
+        $matchmakingTypes ??= new Set(static::MATCHMAKING_TYPES);
```
nanaya: the current php supports this syntax...
```suggestion
        static $matchmakingTypes = new Set(static::MATCHMAKING_TYPES);
```

database/migrations/2026_01_13_000001_add_ranked_play_type_to_multiplayer_rooms.php
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    public function down(): void
+    {
+        DB::statement("ALTER TABLE `multiplayer_rooms` MODIFY `type` ENUM(
+            'playlists',
+            'head_to_head',
+            'team_versus',
+            'matchmaking'
+        ) NOT NULL");
```
nanaya: current laravel supports updating enum columns (also for `up`)
```suggestion
        Schema::table('multiplayer_rooms', function (Blueprint $table) {
            $table->enum('type', [
                'playlists',
                'head_to_head',
                'team_versus',
                'matchmaking',
            ])->change();
        });
```

nanaya (changes requested):
is the playlist check change for all matchmaking types or just `ranked_play`?

smoogipoo:
> is the playlist check change for all matchmaking types or just `ranked_play`?

The way I've implemented it it'll affect all MM types, because I believe quick play can do something similar. I still worry that adding 20 playlist items per room is creating unreasonable DB load (I would be curious how big the `multiplayer_playlist_items` table is right now).

That said, quick play will be temporarily (permanently?) disabled with ranked play.

nanaya (approved):

PR #12673, opened 2026-01-13
Remove unused spotlight scores relation

nanaya:
I can't find where it's actually used. Removing this so there's one less thing to be worried about when simplifying the underlying score models.

### DIFF
--- a/app/Models/Spotlight.php
+++ b/app/Models/Spotlight.php
@@ -5,7 +5,6 @@
 
 namespace App\Models;
 
-use App\Models\Score\Best as ScoreBest;
 use DB;
 use Illuminate\Database\Schema\Blueprint;
 use Schema;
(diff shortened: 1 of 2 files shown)


app/Models/Spotlight.php
```diff
@@ -157,7 +128,6 @@ public function createTables()
             }
 
             foreach ($modes as $mode) {
-                static::createBestScoresTable($this->bestScoresTableName($mode));
```
notbakaneko, line 160: it should still create all the spotlight-related tables (and also probably move all the creation to `SpotlightSeeder`); it seems odd to create only 2/3 of the spotlight tables. 

I think at some point we just decided not to have score handling for spotlights or something...

notbakaneko (changes requested):

app/Models/Spotlight.php
nanaya: I don't think it matters. if anything I'd rather make a new table containing all the spotlight score data than this nightmare

notbakaneko (approved):
since we're going to put those tables in the bin 🗑️

PR #12675, opened 2026-01-13
Hide reply button if comments are disabled

### LINKED ISSUE #12640: ``Reply`` button shouldn't be shown for normal users when comments are disabled
So that this UX issue doesn't happen:

Image

Should still be shown for any user who can post comments despite comments being disabled, such as moderators.

UnhappyF:
Resolves #12640 

Prevent rendering the `reply` button when comments are disabled and the user does not have required permissions. 

I assume we want to keep the `reply` button visible when comments are enabled but the user cannot comment at that moment (for example, when an unauthenticated user clicks `reply` and is prompted to log in). To support this behavior, a `locked` field was added to `CommentableMeta` to separate disabled comments from permission-related restrictions.


resources/js/models/comment.ts
```diff
@@ -63,6 +63,24 @@ export default class Comment {
     return this.canModerate || (this.isOwner && !this.isDeleted);
   }
 
+  @computed
+  get canReply() {
+    const meta = this.controller.getCommentableMeta(this);
+
+    if (this.isDeleted) {
+      return false;
+    }
```
LiquidPL, line 70: this guard should be placed at the beginning of the method

resources/js/models/comment.ts
```diff
@@ -63,6 +63,24 @@ export default class Comment {
...
+
+    if (
+      'current_user_attributes' in meta &&
+      meta.current_user_attributes.can_new_comment_reason == null
+    ) {
+      return true;
```
LiquidPL, line 79: this field and `locked` check for the [same](https://github.com/ppy/osu-web/blob/master/app/Singletons/OsuAuthorize.php#L1149) [thing](https://github.com/ppy/osu-web/blob/master/app/Transformers/CommentableMetaTransformer.php#L23) so there's not really a point in including the latter on the transformer

UnhappyF: But `current_user_attributes.can_new_comment_reason` returns a translated string, so we can’t check the exact reason using this field

LiquidPL: my bad. i've overlooked the other stuff being checked in `OsuAuthorize`.

LiquidPL: however, this still somehow doesn't sit well with me. 

`can_new_comment_reason` being null effectively implies `locked` being false, so one of the checks is being redundant. besides this doesn't need to check for anything but the locked state, all the other cases (not logged in/silenced/etc.) are already handled elsewhere (and also the `canReply` name implies this is responsible for more than it actually is).

i think it's better to check if a comment is locked directly in `renderReplyButton()`.

UnhappyF: There is a case where `can_new_comment_reason == null` and `locked == true`, when the user is allowed to comment on locked threads (currently only admin can do this). I chose this approach instead of checking the admin role directly, to keep the privilege logic in `priv_check`.

resources/js/models/comment.ts
```diff
@@ -63,6 +63,24 @@ export default class Comment {
...
+  @computed
+  get canReply() {
+    if (this.isDeleted) {
+      return false;
+    }
+
+    const meta = this.controller.getCommentableMeta(this);
+
+    if (
+      'current_user_attributes' in meta &&
```
nanaya: check this early and return false immediately if missing so the rest of the function can work on just the expected commentable type without needing to check the field existence multiple times

resources/views/docs/_structures/commentable_meta.md
```diff
@@ -21,6 +21,7 @@ owner_title             | string?
 title                   | string                                                             | display title
 type                    | string                                                             | the type of the object
 url                     | string                                                             | url of the object
+locked                  | boolean                                                            | `true` if new comments are disabled
```
nanaya: sort alphabetically

UnhappyF: fixed

resources/js/models/comment.ts
UnhappyF: fixed

nanaya (approved):

PR #12679, opened 2026-01-15
Clean up livestream featured player

nanaya:
- use aspect ratio instead of padding hack
- remove need of absolute positioning
- remove third party cookie workaround as it's no longer needed
  - remove placeholder as well
- better typing maybe


resources/views/livestreams/_featured.blade.php
```diff
@@ -2,22 +2,11 @@
...
-    $playerId = 'livestream-'.time().'-'.rand();
-?>
 <div class="livestream-featured">
-    <div
-        id="{{ $playerId }}"
-        class="js-twitch-player livestream-featured__content hidden"
-        data-channel="{{ $featuredStream->data['user_name'] }}"
-    ></div>
-
     <a
```
nanaya: I feel this can just be removed with the third party cookie thing being fixed

notbakaneko (commented):
Should the featured stream be auto playing? Asking this because the first thing I got while testing it was a twitch ad 🤔

nanaya:
it's been autoplaying but yeah idk...

notbakaneko:
I guess it makes sense to autoplay if the featured stream is a tournament or something

notbakaneko (approved):

PR #12680, opened 2026-01-15
Add count for lazer multiplayer games

### LINKED ISSUE #12667: Number of multiplayer rooms in homepage only accounts for stable client.
Image

Image

Image

nanaya:
Resolves #12667.

- [x] migration entry `2026_01_16_114540_add_multiplayer_games_lazer_to_banchostats` (has been run in prod)

### DIFF
--- a/app/Models/BanchoStats.php
+++ b/app/Models/BanchoStats.php
@@ -5,8 +5,6 @@
 
 namespace App\Models;
 
-use Illuminate\Support\Facades\DB;
-
 /**
  * @property int $banchostats_id
  * @property \Carbon\Carbon $date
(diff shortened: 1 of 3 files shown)


app/Models/BanchoStats.php
```diff
@@ -26,7 +24,7 @@ public static function stats()
     {
         return array_reverse(
             static::whereRaw('banchostats_id mod 10 = 0')
-                ->select(['users_irc', DB::raw('(users_osu + users_lazer) AS users'), 'multiplayer_games', 'date'])
+                ->selectRaw('(users_osu + users_lazer) AS users, multiplayer_games, multiplayer_games_lazer')
```
nanaya, line 27: it feels kinda silly getting all the multiplayer counts just to end up using one single value...

peppy: do the same as with users and total them in the sql maybe?

nanaya: yeah but that'd feel even sillier summing up a hundred ish numbers just to use one

the user column is used for historical chart so it makes sense there

which also makes me wonder if it should just count the rooms in real time...

peppy (approved):

PR #12682, opened 2026-01-15
Display more info in contest entry input field

### LINKED ISSUE #12617: Display upload requirements in contest submission box

Hiviexd:
https://github.com/user-attachments/assets/a021060e-daf0-4607-bc86-5f9030444598

Resolves #12617

Shows allowed filetypes and max size in field and adds a loading state when uploading entries

Also displays dimensions when `forced_width` and `forced_height` are defined (which is the case for art contests)

image


resources/css/bem/contest-userentry.less
```diff
@@ -69,12 +69,30 @@
   &__uploader {
     margin: 0;
     width: 100%;
-    height: 100%;
+    min-height: 100%;
```
LiquidPL: i think this can still be `height`, given that it's just going to span all of its parent anyway

notbakaneko: yeah this should just be `height`, it makes no sense for it to go beyond 100%. (Actually, a bunch of the other attributes aren't even necessary, either, if the parent is changed to flexbox and this has `flex: 1`

Hiviexd: did a cleanup based on what you said at the end

resources/css/bem/contest-userentry.less
```diff
@@ -67,14 +69,29 @@
   }
 
   &__uploader {
-    margin: 0;
-    width: 100%;
-    height: 100%;
+    flex: 1;
+    padding: 10px;
     align-items: center;
```
notbakaneko: the `align-items` and `overflow` aren't necessary either, and also update the `width` and `height` in `fileupload.less` to 100%

notbakaneko (approved):
hope I don't stumble upon another Safari problem ಠ_ಠ

PR #12684, opened 2026-01-15
Migrate forum post tag management to separate popup

LiquidPL:
Per request of the forum support team (see https://discord.com/channels/90072389919997952/1460408526697402579).

|desktop|mobile|
|-------|------|
|image|image|

Since now the UX expectation is to have the moderator select more than one tag while the dropdown is open, I have removed the confirmation popups that were previously there - I believe there's enough visual confirmation of success in the dropdown as it is.

I've also removed translation strings for said confirmations, since the old nav toolbar buttons are removed, I don't think these are used anywhere else. Should I also remove them in other languages, or will that be handled by Crowdin down the line?

This also adds some new tags to the interface, again per request of the forum support team.


resources/views/forum/topics/_edit_tags.blade.php
```diff
@@ -0,0 +1,67 @@
...
+@php
+$icons = [
+    'added' => 'fas fa-cogs',
+    'archived' => 'fas fa-archive',
+    'assigned' => 'fas fa-user',
+    'confirmed' => 'fas fa-exclamation-triangle',
+    'duplicate' => 'fas fa-copy',
+    'invalid' => 'far fa-times-circle',
+    'resolved' => 'far fa-check-circle',
+];
```
notbakaneko: unused

resources/views/forum/topics/_edit_tags.blade.php
```diff
@@ -0,0 +1,67 @@
...
+        data-menu-id="tag-editor"
+        data-visibility="hidden"
+    >
+        @foreach ($topic::ISSUE_TAGS as $category => $types)
+            <span class="simple-menu__heading">{{ $category }}</span>
+            @foreach ($types as $type)
+                @php
+                    $state = $topic->hasIssueTag($type);
+                @endphp
+                <button
```
notbakaneko: these should be debounced like `AccountEdit`

notbakaneko (changes requested):

resources/views/forum/topics/_edit_tags.blade.php
LiquidPL: ~~do they need to though? the issue tag endpoint doesn't allow to change multiple tags in one request, and the buttons themselves get disabled for the duration of the request, so the user can't spam click them~~

nanaya (commented):
the buttons don't seem to do anything in my setup

even if it does (by readding the removed data-remote etc), there's no logic to update the url so it won't actually toggle anything the second time

also, `<button>` may not contain `<input>` or `<label>`

LiquidPL:
i've cleaned up templates and brought back the `_issue_tag_*` ones, also reworked how the requests are made

feel free to squash when merging

notbakaneko:
The point of the `debounce` in account edit is to batch multiple changes into a one request and only send it at a stable state.

The current one doesn't handle errors and behaves weirdly when latency is invoved:

https://github.com/user-attachments/assets/2a07bb86-004d-49f0-8aed-e8c8ba8f4a0f



`6ecac3e` sends separate requests, leaves the spinner, and doesn't check/uncheck the box until after the operation is done:

https://github.com/user-attachments/assets/34cdcb7c-99af-4681-bb87-d295edd738d8



We expect feedback on the box being checked/unchecked, radio buttons switching, etc, to be immediate and not after:

https://github.com/user-attachments/assets/78e851a3-6d6e-4085-9ba1-f848c287c625

LiquidPL:
thanks for pointing this out. i somehow got completely confused implementing this.

app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -166,21 +166,24 @@ public function issueTag($id)
 
         priv_check('ForumModerate', $topic->forum)->ensureCan();
 
-        $issueTag = presence(Request::input('issue_tag'));
-        $state = get_bool(Request::input('state'));
-        $type = 'issue_tag_'.str_slug($issueTag);
+        $tags = presence(get_arr(Request::input('tags'), 'get_bool')) ?? [];
```
notbakaneko: `get_arr` turns it into an interger-indexed array which makes the next check not work

notbakaneko: the tags don't need anything special like multi-state or anything, it's probably simplier to just send all the selected tags as an array of strings and do a diff to add/remove tags

resources/js/core/forum/forum-topic-tag-editor.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  private xhr?: JQuery.jqXHR;
+
+  constructor() {
+    this.debouncedUpdate = debounce(this.update, 1000);
+
+    $(document).on('click', '.js-forum-topic-tag-editor-tag', this.onClick);
+  }
+
+  private readonly getChangedTags = (container: HTMLElement): { [key: string]: boolean } => {
```
notbakaneko: Move the selector to the `input`, make the data attribute the name, and just send all the selected tags
`[...container.querySelectorAll<HTMLInputElement>('.js-forum-topic-tag-editor-tag')].filter(((input) => input.checked)).map((elem) => elem.dataset.issueTag)`

resources/views/forum/topics/_issue_tag.blade.php
```diff
@@ -6,26 +6,17 @@
...
-    data-remote="1"
-    data-method="post"
+<div
+    class="simple-menu__item js-forum-topic-tag-editor-tag"
+    data-issue-tag="{{ $issueTag }}"
 >
-    <span class="btn-circle__content">
-        {!! issue_icon($issueTag) !!}
+    <label class="osu-switch-v2">
+        <input type="checkbox" class="osu-switch-v2__input" {{ $state ? 'checked' : '' }}>
```
notbakaneko: `input` elements should have a `name`

resources/js/core/forum/forum-topic-tag-editor.ts
notbakaneko: Also this isn't used as a callback and doesn't need to be an arrow function

resources/js/core/forum/forum-topic-tag-editor.ts
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import { debounce } from 'lodash';
+import { route } from '../../laroute';
+import { onError } from '../../utils/ajax';
```
notbakaneko: `import` with absolute path

resources/js/core/forum/forum-topic-tag-editor.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    this.xhr = $.post(route('forum.topics.issue-tag', {
+      tags: this.getEnabledTags(editor),
+      topic: editor.dataset.topicId,
+    }));
+
+    this.xhr.catch(onError)
+      .done(this.setComplete);
+  };
```
notbakaneko: we use `.done` and `.fail` for requests unless there's something we explicitly want to do with promise chaining, and we also don't mix `then`/`catch` with `done`/`fail`

resources/css/bem/simple-menu.less
```diff
@@ -182,6 +189,14 @@
     padding: 10px 0;
   }
 
+  &__heading {
+    padding: 5px;
+    font-size: @font-size--normal;
+    font-weight: 900;
+    text-transform: uppercase;
+    color: #ffffff;
```
notbakaneko: `#fff`

resources/js/core/forum/forum-topic-tag-editor.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    if (editor == null) {
+      throw new Error('could not find editor element');
+    }
+
+    this.setLoading();
+
+    const tagToApply = target.dataset.issueTag;
+
+    if (tagToApply === null || tagToApply === undefined) {
```
notbakaneko: use `== null` for null or undefined check

resources/views/forum/topics/_issue_tag.blade.php
```diff
@@ -6,26 +6,17 @@
...
+        <input type="checkbox" name="{{ $issueTag }}" class="osu-switch-v2__input js-forum-topic-tag-editor-checkbox" {{ $state ? 'checked' : '' }}>
+        <span class="osu-switch-v2__content"></span>
+    </label>
+    <span class="simple-menu__item-icon">{!! issue_icon($issueTag) !!}</span>
+    <span>{{ $issueTag }}</span>
+    <span class="simple-menu__item-loading-spinner">
+        {!! spinner() !!}
     </span>
-</button>
+</div>
```
notbakaneko: extra `js-` class and data shouldn't be needed here anymore

`label` should should be the top level and enclose the whole clickable area.
Also `spinner` isn't needed here anymore.

resources/js/core/forum/forum-topic-tag-editor.ts
```diff
@@ -0,0 +1,75 @@
...
+    $(document).on('click', '.js-forum-topic-tag-editor-tag', this.onClick);
+  }
+
+  private getEnabledTags(container: HTMLElement) {
+    return [...container.querySelectorAll<HTMLInputElement>('.js-forum-topic-tag-editor-checkbox')]
+      .filter((input) => input.checked)
+      .map((el) => el.name);
+  }
+
+  private readonly onClick = (e: Event) => {
```
notbakaneko: or
```typescript
constructor() {
...
  $(document).on('change', '.js-forum-topic-tag-editor-checkbox', this.onChange);
}

private readonly onChange = (e: JQuery.TriggeredEvent<unknown, unknown, HTMLInputElement, unknown>) => {
  const target = e.currentTarget;
  const container = target.closest<HTMLElement>('.js-forum-topic-tag-editor');
  if (container == null) {
    throw new Error('missing container');
  }

  this.xhr?.abort();

  // sync the checkbox state on both desktop and mobile checkboxes
  $(`.js-forum-topic-tag-editor-checkbox[name="${target.name}"]`).prop('checked', target.checked);

  this.setLoading();
  this.debouncedUpdate(container);
};
```

resources/css/bem/simple-menu.less
```diff
@@ -224,6 +239,17 @@
...
 
+    .@{_top}--forum-topic-tag-editor & {
+      display: grid;
+      grid-template-columns: min-content 35px min-content min-content;
+      padding: @_gutter @_gutter * 2;
+      align-items: center;
+
+      &::before {
+        display: none;
+      }
```
notbakaneko: the other menu on the page having the stripe and this one not seems odd to me... 🤔

app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -166,21 +166,30 @@ public function issueTag($id)
 
         priv_check('ForumModerate', $topic->forum)->ensureCan();
 
-        $issueTag = presence(Request::input('issue_tag'));
-        $state = get_bool(Request::input('state'));
-        $type = 'issue_tag_'.str_slug($issueTag);
+        $tags = presence(get_arr(Request::input('tags'), 'get_string')) ?? [];
```
notbakaneko: `presence` isn't useful here

app/Models/Forum/Topic.php
```diff
@@ -139,6 +147,14 @@ class Topic extends Model implements AfterCommit
         'topic_time' => TimestampOrZero::class,
     ];
 
+    /**
+     * @return string[]
+     */
+    public static function getIssueTagsFlat(): array
+    {
+        return array_merge(...array_values(static::ISSUE_TAGS));
```
notbakaneko:
```suggestion
        return array_flatten(static::ISSUE_TAGS);
```

app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -166,21 +166,30 @@ public function issueTag($id)
...
         }
 
-        $this->logModerate('LOG_ISSUE_TAG', compact('issueTag', 'state'), $topic);
+        foreach ($topic->issueTags() as $existingTag) {
+            if (in_array($existingTag, $tags, true)) {
+                continue;
+            }
 
-        $method = $state ? 'setIssueTag' : 'unsetIssueTag';
+            $this->logModerate('LOG_ISSUE_TAG', ['tag' => $existingTag, 'state' => false], $topic);
```
notbakaneko: this needs to be `issueTag` for the serialization/deserialization to work

app/Http/Controllers/Forum/TopicsController.php
```diff
@@ -166,21 +166,30 @@ public function issueTag($id)
...
+            $topic->unsetIssueTag($existingTag);
+        }
 
-        $topic->$method($issueTag);
+        foreach ($tags as $tag) {
```
notbakaneko: This is readding unchanged tags to the log. Split the changes first

```php
$currentTags = $topic->issueTags();
$removed = array_diff($currentTags, $tags);
$added = array_diff($tags, $currentTags);

foreach ($removed as $tag) {
    $this->logModerate('LOG_ISSUE_TAG', ['issueTag' => $tag, 'state' => false], $topic);
    $topic->unsetIssueTag($tag);
}

foreach ($added as $tag) {
    $this->logModerate('LOG_ISSUE_TAG', ['issueTag' => $tag, 'state' => true], $topic);
    $topic->setIssueTag($tag);
}
```

resources/css/bem/simple-menu.less
LiquidPL: yeah true, i initially removed it because it was getting too cluttered but you've got a point

resources/views/forum/topics/_issue_tag.blade.php
```diff
@@ -6,26 +6,12 @@
...
-    data-remote="1"
-    data-method="post"
->
-    <span class="btn-circle__content">
-        {!! issue_icon($issueTag) !!}
-    </span>
-</button>
+
+<label class="simple-menu__item">
+    <label class="osu-switch-v2">
```
notbakaneko: btw why isn't this using the `objects._switch` include? 🤔

LiquidPL: i forgot about it :upside_down_face:

PR #12687, opened 2026-01-16
Add endpoint for awarding client-side medals

bdach:


app/Http/Controllers/UsersController.php
```diff
@@ -683,6 +685,27 @@ public function show($id, $mode = null)
...
+        $request = \Request::instance();
+        $achievement = app('medals')->byIdOrFail($achievementId);
+
+        abort_unless($achievement->client_side, 422, 'achievement cannot be unlocked');
+
+        try {
+            ClientCheck::parseToken($request);
+        } catch (HttpException $e) {
+            abort(403);
+        }
```
nanaya, line 700: this isn't needed unless we want to hide the specific message (although it'll be a 422)

app/Models/Achievement.php
```diff
@@ -48,6 +50,7 @@ public function getAttribute($key)
         return match ($key) {
             'achieved_count',
             'achievement_id',
+            'client_side',
```
nanaya: should be in the boolean casting section

routes/web.php
```diff
@@ -596,6 +596,8 @@
         Route::get('me/download-quota-check', 'HomeController@downloadQuotaCheck')->name('download-quota-check');
         //  GET /api/v2/me
         Route::get('me/{mode?}', 'UsersController@me')->name('me');
+        //  PUT /api/v2/me/achievements/:achievementId
+        Route::put('me/achievements/{achievementId}', 'UsersController@unlockClientSideAchievement')->name('unlock-client-side-achievement');
```
nanaya, line 600: use `artisan route:convert --to-json=tests/api_routes.json` to generate the json thing

nanaya (changes requested):
just a quick look. will check more later

app/Http/Controllers/UsersController.php
bdach: I want a different response code to be able to distinguish between recoverable and non-recoverable failures client side. As in a 403 is intended to signal to the client to stop trying to unlock medals.

I guess I could also be using the message itself but this seemed more proper.

nanaya (approved):

PR #12688, opened 2026-01-16
Remove unused legacy score reporting support

nanaya:
There has been no score best report since 2024 so it should be fine to remove (mainly because `best_id` has been null due to using `Solo\Score` model).

Apparently I should've kept the `scoreUrl` thing because I missed the legacy match scores (to be done separately later).


notbakaneko (changes requested):
..now it conflicts

notbakaneko (approved):

PR #12692, opened 2026-01-19
"Remove" all 0 entries in profile beatmap section

### LINKED ISSUE #12632: Remove "Nominated maps" list from common users userpages
We don't think that common user can nominate anything :( You need to filter player status and, if player is beatmap nominator, automatically add "Nominated maps" list into beatmap part on userpage

someonecantcode:
Closes #12632 

image

> dont render when there are 0 items/count.

image
> When empty, the section can be hidden. This can also apply to all other sections if we want to tidy things up. I'm fine with that. - peppy. Using the same message when Recent section is empty.

Question: is there any difference between `state.count` and `state.items.length`?


nanaya (commented):
maybe always render the favourite maps section when it's completely empty instead?

`.count` is the total count, `.items` only contains loaded items

peppy:
I think always showing favourites but hiding the others sounds like a good compromise in discoverability and cleanliness 👍.

someonecantcode:
alr. this is how'd it look always showing favourites even when 0 count.

image

resources/js/profile-page/beatmapsets.tsx
```diff
@@ -67,8 +71,9 @@ export default class Beatmapsets extends React.Component<ExtraPageProps> {
   };
 
   private readonly renderBeatmapsets = (section: typeof sectionKeys[number]) => {
-    const state = this.props.controller.state.lazy.beatmaps?.[section.key];
+    const state = this.data?.[section.key];
     if (state == null) return;
+    if(state.count <= 0 && section.key !== 'favourite') return;
```
nanaya: kinda weird seeing a count check for below zero instead of just exactly zero. Also there should be space between `if (`

resources/js/profile-page/beatmapsets.tsx
```diff
@@ -44,9 +44,13 @@ const sectionKeys = [
 
 @observer
 export default class Beatmapsets extends React.Component<ExtraPageProps> {
+  private get data() {
```
nanaya: this doesn't do much anymore, does it

nanaya (changes requested):

resources/js/profile-page/beatmapsets.tsx
someonecantcode: yeah not anymore. slightly less redundancy. ill just remove it

PR #12693, opened 2026-01-19
Unify 204 response generation

nanaya:
With some `sed` and stuff.

```
sed -i -e 's/response(\[], 204);/response()->noContent();/' app/Http/Controllers/*.php app/Http/Controllers/*/*.php
```
(and a bit more for null and other paths)


notbakaneko (approved):
looks correct

PR #12695, opened 2026-01-19
Update beatmap cover placeholder

nanaya:
Recycling wrapped fallback backgrounds. Keeping the old one because who knows what else is using it atm (in addition to the one in wrapped).


notbakaneko (approved):
Loading the full size image for the cover thumbnail seems silly but they should be cached already on the next use, so maybe fine? 🤔

PR #12696, opened 2026-01-20
Simplify legacy score model

nanaya:
I suppose I could've kept the `scoresBest*` relations in user and beatmap models but the verbose way isn't too annoying.

<sub>I should probably test I didn't break the seeder</sub>


app/Models/Score/Best/Model.php
```diff
@@ -64,71 +46,4 @@ public function getAttribute($key)
             'user' => $this->getRelationValue($key),
         };
     }
-
-    public function replayFile(): ?LegacyReplayFile
-    {
-        return $this->replay ? new LegacyReplayFile($this) : null;
```
nanaya, line 70: apparently I forgot to remove this in the previous pr 👀

notbakaneko (approved):

PR #12703, opened 2026-01-22
Use css variable to set current user avatar/cover

nanaya:
I think it's better than updating every single elements...?


notbakaneko:
Would it be a good idea to also switch the react components to this? 🤔 
I think only the user profile page cover needs a direct setting for previews...?

notbakaneko (approved):
seems correct

PR #12704, opened 2026-01-22
Paginate team members

LiquidPL:
Needed for ppy/osu#32584 and also because we shouldn't fetch 200+ user rows with every page load.

Didn't update the manage members page yet since this PR is already large enough and besides I'll probably port the entire thing to React so that the empty slot counter can update when removing team members and also so that changing things around doesn't incur a page load every time.

image

### DIFF
--- a/app/Http/Controllers/Teams/MembersController.php
+++ b/app/Http/Controllers/Teams/MembersController.php
@@ -15,6 +15,9 @@ use App\Transformers\TeamMemberTransformer;
 use App\Transformers\UserCompactTransformer;
 use Symfony\Component\HttpFoundation\Response;
 
+/**
+ * @group Teams
+ */
 class MembersController extends Controller
 {
     const int PAGE_SIZE = 50;
@@ -43,6 +46,22 @@ class MembersController extends Controller
         return response()->noContent();
     }
 
+    /**
+     * Get Team Members
+     *
+     * This endpoint returns the members of a specified team.
+     *
+     * ---
+     *
+     * ### Response format
+     *
+     * Returns the [TeamMember](#teammember) object.
+     *
+     * @urlParam team integer required ID of the team. Example: 1
+     * @queryParam limit integer Maximum number of results. Example: 50
+     * @queryParam offset integer Result offset for pagination. Example: 100
+     * @queryParam withLeader boolean Whether the response should include the team leader. Example: 0
+     */
     public function index(string $teamId): Response
     {
         $team = Team::findOrFail($teamId);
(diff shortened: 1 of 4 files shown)


app/Transformers/TeamMemberTransformer.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    protected array $availableIncludes = ['user'];
+
+    protected array $defaultIncludes = ['user'];
+
+    public function transform(TeamMember $member): array
+    {
+        return [
+            'created_at' => json_time($member->created_at),
+        ];
```
LiquidPL, line 23: this is here even though it's not used anywhere because I want to use it for the manage members page in a future PR and I'd rather it be here from the beginning than later deal with changing the schema once lazer depends on it

app/Http/Controllers/Teams/MembersController.php
```diff
@@ -39,15 +46,59 @@ public function destroy(string $teamId, string $userId): Response
...
+     * @queryParam limit integer Maximum number of results. Example: 50
+     * @queryParam offset integer Result offset for pagination. Example: 100
+     * @queryParam withLeader boolean Whether the response should include the team leader. Example: 0
+     */
     public function index(string $teamId): Response
     {
         $team = Team::findOrFail($teamId);
 
-        priv_check('TeamUpdate', $team)->ensureCan();
+        if (!is_json_request()) {
```
LiquidPL, line 69: I'm not sure about this. On one hand, the `index` resource seems like the right place for returning a paginated list of items, but on the other hand I'm shoehorning two different things into one controller method here.

resources/css/bem/show-more-link.less
```diff
@@ -87,6 +87,10 @@
     }
   }
 
+  &--team-members {
+    grid-column: 1 / -1;
+  }
```
LiquidPL, line 92: I didn't change the color of the button here since I wasn't sure which one should I use. Both the button and the list background are `@osu-colour-b2` so I can't go two more down to `-b0` so they are reasonably separated like for instance on the user profile.

app/Transformers/TeamMemberTransformer.php
LiquidPL: as well as the `withLeader` param

nanaya:
I'm not sure if this is needed. It's not like the count is unbounded.

LiquidPL:
What's the current upper bound? Even if it's not particularly large, it's still going to load upwards of 200 database records every time the page is loaded in the worst case scenario (unless that is still a negligible amount).

nanaya:
it's 256. And yeah it seems fine.

LiquidPL:
Alright, I'll simplify this down to just returning the member list in an include on the team transformer then. Or just open that in a fresh PR cuz there's no reason to include this all in the history.

PR #12711, opened 2026-01-26
Adjust password reset form

nanaya:
- per-user rate limiter
- shorter lifetime for reset key
- shorter delay between mails (but also subject to the new limiter)


app/Libraries/User/PasswordResetData.php
```diff
@@ -24,6 +25,23 @@ private function __construct(
...
 
+    /**
+     * Rate limiter for password reset attempts
+     *
+     * This applies to both new reset request and resend mail request.
+     */
+    public static function attempt(User $user): bool
+    {
+        $limiterKey = "password_reset_user:{$user->getKey()}";
+        if (\RateLimiter::remaining($limiterKey, 3) < 1) {
```
LiquidPL: this could use `\RateLimiter::tooManyAttempts()` instead

app/Libraries/User/PasswordResetData.php
```diff
@@ -97,8 +119,7 @@ public function hasMoreTries(): bool
 
     public function isActive(): bool
     {
-        return $this->attrs['expiresAt'] > time()
-            && hash_equals($this->attrs['authHash'], static::authHash($this->user));
+        return hash_equals($this->attrs['authHash'], static::authHash($this->user));
     }
```
LiquidPL, line 123: maybe the function name could be changed since this no longer checks for `expiresAt` and only for hash equality?

nanaya: the hash check is whether or not the user changed the password since requesting the reset

LiquidPL (approved):

peppy (approved):

PR #12717, opened 2026-01-28
Hide user command chat messages

### LINKED ISSUE #12689: `!report` is visible to other users if sent from web chat
Found this out the hard way recently. Reported someone at 18:23 UTC today via web chat, and another user (also on web chat) could see it and acknowledged it. BanchoBot DMed me to say that chat moderators were alerted.

**This seems like a pretty important issue.**

nanaya:
Resolves #12689

I don't know if there's a valid indicator when the message isn't a valid command...

It'll at least still show messages that start with two exclamation marks `!!` or followed by a space `! thing`

### DIFF
--- a/app/Models/Chat/Message.php
+++ b/app/Models/Chat/Message.php
@@ -122,11 +122,7 @@ class Message extends Model implements ReportableInterface
 
     public function isUserCommand(): bool
     {
-        $content = $this->content;
-
-        return strlen($content) > 1
-            && $content[0] === '!'
-            && ($content[1] !== ' ' && $content[1] !== '!');
+        return preg_match('/^![^ !]/', $this->content) === 1;
     }
 
     public function reportableAdditionalInfo(): ?string
--- a/app/Models/Chat/Message.php
+++ b/app/Models/Chat/Message.php
@@ -28,6 +28,14 @@ class Message extends Model implements ReportableInterface
 {
     use Reportable;
 
+    public ?string $uuid = null;
+
+    protected $primaryKey = 'message_id';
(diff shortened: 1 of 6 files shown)


LiquidPL:
I wonder what impact would this have on someone using non-moderation BanchoBot commands in a public channel. There's [a few](https://osu.ppy.sh/wiki/en/BanchoBot) commands aside from !report that a user would be able to send, and right now I assume the command itself would not show but the result would.

peppy:
> I wonder what impact would this have on someone using non-moderation BanchoBot commands in a public channel. There's [a few](https://osu.ppy.sh/wiki/en/BanchoBot) commands aside from !report that a user would be able to send, and right now I assume the command itself would not show but the result would.

is a valid point which was discussed IRL. should show user commands to the local user where feasible.

notbakaneko (changes requested):
probably can do with some filtering tests

app/Models/Chat/Message.php
```diff
@@ -126,7 +126,7 @@ public function isUserCommand(): bool
 
         return strlen($content) > 1
             && $content[0] === '!'
-            && ($content[1] !== ' ' || $content[1] !== '!');
+            && ($content[1] !== ' ' && $content[1] !== '!');
```
notbakaneko: unneeded brackets?

notbakaneko (changes requested):

notbakaneko:
> simplify
> regex

ಠ_ಠ

nanaya:
=]

notbakaneko (approved):

PR #12727, opened 2026-01-30
Use preinstalled yarn

nanaya:
One less thing to (redundantly) install.


notbakaneko:
I feel like we should be setting `packageManager` as well... (also funny that yarn 1 is still the known good release)

nanaya:
looks like it requires specifying exact version (and downloading it on version mismatch)

doesn't seem like an interesting idea

notbakaneko (approved):

PR #12730, opened 2026-01-30
Move news updates to separate section in mail notification digest

### LINKED ISSUE #12598: News post notifications read badly
Image

Probably need their own template. It reads as if there's been an update / comment post / something rather than highlighting something new.

notbakaneko:
I originally had something more structured and also broke up the other notifications into more categories but reverted them in the end because:
1. mail notification uses a different grouping to summarize notifications
2. notification digest currently sends frequently enough that there shouldn't be that many news posts in the mail to need grouping them into series.

Also thought of adding a short abstract in the mail but I think that's better done in the notification data itself and not querying it for every mail...?
(also fake markdown to make it stand out more?)

closes #12598

### DIFF
--- a/Dockerfile.deployment
+++ b/Dockerfile.deployment
@@ -34,7 +34,7 @@ RUN echo "install_weak_deps=False" >> /etc/dnf/dnf.conf && \
       unzip && \
     microdnf clean all
 
-RUN npm install -g yarn
+RUN corepack enable yarn
 
 WORKDIR /app
 
--- a/Dockerfile.development
+++ b/Dockerfile.development
@@ -39,7 +39,7 @@ RUN echo "install_weak_deps=False" >> /etc/dnf/dnf.conf && \
       util-linux && \
     microdnf clean all
 
-RUN npm install -g yarn
+RUN corepack enable yarn
 
 RUN curl -L "https://getcomposer.org/download/latest-2.x/composer.phar" > /usr/local/bin/composer && chmod 755 /usr/local/bin/composer
 
--- a/app/Http/Controllers/HomeController.php
+++ b/app/Http/Controllers/HomeController.php
@@ -14,6 +14,7 @@ use App\Models\BeatmapDownload;
 use App\Models\Beatmapset;
 use App\Models\Build;
 use App\Models\Forum\Post;
+use App\Models\LivestreamCollection;
 use App\Models\Multiplayer\Room;
 use App\Models\NewsPost;
 use App\Models\UserDonation;
@@ -133,12 +134,16 @@ class HomeController extends Controller
 
             $dailyChallenge = Room::dailyChallengeFor(CarbonImmutable::now());
 
+            $livestream = new LivestreamCollection();
+            $featuredStream = $livestream->featured();
+
             return ext_view('home.user', compact(
                 'menuImages',
                 'newBeatmapsets',
                 'news',
                 'popularBeatmapsets',
                 'dailyChallenge',
+                'featuredStream',
             ));
         } else {
             $news = json_collection($news, 'NewsPost');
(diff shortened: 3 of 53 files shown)


app/Mail/UserNotificationDigest.php
```diff
@@ -77,7 +87,11 @@ public function build()
         $user = $this->user;
 
         return $this
-            ->text('emails.user_notification_digest', compact('groups', 'user'))
+            ->text('emails.user_notification_digest', [
+                'groups' => array_values($this->groups),
```
nanaya, line 92: the two variables before this end up being useless

app/Mail/UserNotificationDigest.php
```diff
@@ -33,11 +41,17 @@ private function addToGroups(Notification $notification)
             $baseKey = 'notifications.mail.'.$class::getMailBaseKey($notification);
             $key = $class::getMailGroupingKey($notification);
 
+            if ($class === NewsPostNew::class) {
```
nanaya, line 38: move the key variables after this section?

app/Mail/UserNotificationDigest.php
```diff
@@ -15,7 +16,14 @@
 // Paired with App\Jobs\UserNotificationDigest
 class UserNotificationDigest extends Mailable
 {
-    private $groups = [];
+    private array $groups = [];
+    private array $news = [];
+
+    private static function getDetails(Notification $notification)
```
nanaya: is static function after or before constructor

and return type?

app/Mail/UserNotificationDigest.php
```diff
@@ -15,7 +16,14 @@
...
+    {
+        // remove anything not a string because trans doesn't like it.
+        return array_filter($notification->details ?? [], fn ($value) => is_string($value));
```
nanaya: `is_string(...)`

nanaya:
Not sure if any better...

```
Hi testuser10,

Just letting you know that there have been new updates on items you are watching.

# News Updates

## Community Mentorship Program Winter 2025 Signups Now Open
https://on.nanaya.net/home/news/2026-01-05-community-mentorship-program-winter-2025-signups-now-open

---

Change email notification preferences:
https://on.nanaya.net/home/account/edit#notifications

--
osu! | https://on.nanaya.net
```

maybe @peppy can confirm if it's ok.

peppy:
It's definitely better than what we had, but dunno if I'd say it's "good". Maybe fine to merge as a stepping stone.

FWIW I'm not against using rich text emails to make these look better these days.

resources/views/emails/_news.blade.php
```diff
@@ -0,0 +1,12 @@
+{{--
+    Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+    See the LICENCE file in the repository root for full licence text.
+--}}
+# {!! osu_trans('notifications.mail.news_updates') !!}
+
+@foreach ($news as $details)
+## {!! $details['title'] !!}
```
nanaya: no more series/category...?

app/Mail/UserNotificationDigest.php
notbakaneko: did I not commit the deleted lines....no I didn't 🙈

app/Mail/UserNotificationDigest.php
```diff
@@ -26,18 +29,31 @@ public function __construct(private array $notifications, private User $user)
...
+    }
+
     private function addToGroups(Notification $notification)
     {
         try {
             $class = BroadcastNotificationBase::getNotificationClassFromNotification($notification);
+
+            if ($class === NewsPostNew::class) {
+                // group news by category
+                $details = static::getDetails($notification);
```
nanaya: I just noticed, the filtering doesn't really do anything in this context does it 🤔

app/Mail/UserNotificationDigest.php
```diff
@@ -26,18 +29,31 @@ public function __construct(private array $notifications, private User $user)
...
+                $details['link'] = NewsPostNew::getMailLink($notification);
```
nanaya: and this can be combined with the `$link` variable setting far down below

nanaya (changes requested):

PR #12731, opened 2026-01-30
Add `multiplayer.write_manage` scope for multiplayer refereeing API

bdach:
Scope to be used by `osu-server-spectator` for https://github.com/ppy/osu-server-spectator/issues/406.

I initially named it `multiplayer.write` but `write_manage` seems closer in line with the chat usage sooo 🤷 

The access matrix here is the same as it is for chat scopes, but I will restate it in plain language, because it is confusing, to the point that I was confused by it at several points myself, which is why this took so long to get out because I wanted to 100% confirm it:

- Non-bot owned client + `client_credentials` grant:
  - ⛔ "Delegation with Client Credentials is only available to bots."
- Non-bot owned client + `authorization_code` grant + client owner tries to authorize self:
  - ✅ Client acts as client owner / self.
- Non-bot owned client + `authorization_code` grant + some other user tries to authorize self:
  - ⛔ "This scope is only available for bots or your own clients."
- Bot-owned client + `client_credentials` grant:
  - ✅ Client acts as client owner / the bot account.
- Bot-owned client + `authorization_code` + any user tries to authorize self:
  - ✅ Client acts as the authorized user (real delegation).

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -81,11 +81,6 @@ class TestCase extends BaseTestCase
         );
     }
 
-    protected static function ownClientOrBotScopes(): Collection
-    {
-        return static::allPassportScopeIds()->filter(fn ($scope) => in_array($scope, Token::SCOPES_OWN_CLIENT, true));
-    }
-
     protected static function allPassportScopeIds(): Collection
     {
         return Passport::scopes()->pluck('id');
--- a/tests/Models/OAuth/TokenTest.php
+++ b/tests/Models/OAuth/TokenTest.php
@@ -18,13 +18,13 @@ class TokenTest extends TestCase
 {
     public static function dataProviderForTestAuthCodeOwnClientOrBotScopesAllowsSelf()
     {
-        return static::ownClientOrBotScopes()->map(fn ($scope) => [$scope]);
+        return array_map(fn ($scope) => [$scope], Token::SCOPES_OWN_CLIENT);
     }
 
     public static function dataProviderForTestAuthCodeOwnClientOrBotScopesRequiresBotGroup()
     {
         $data = [];
-        foreach (static::ownClientOrBotScopes() as $scope) {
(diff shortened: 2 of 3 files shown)


tests/TestCase.php
```diff
@@ -81,9 +81,9 @@ protected static function fileList($path, $suffix)
         );
     }
 
-    protected static function chatScopes(): Collection
+    protected static function ownClientOrBotScopes(): Collection
     {
-        return static::allPassportScopeIds()->filter(fn ($scope) => str_starts_with($scope, 'chat.'));
+        return static::allPassportScopeIds()->filter(fn ($scope) => in_array($scope, Token::SCOPES_OWN_CLIENT, true));
```
notbakaneko: at this point we might as well use `Token::SCOPES_OWN_CLIENT` directly since the previous assumption of not having this variable and adding more items to it doesn't hold true, anymore.

notbakaneko:
Also is there anything web-side going to be doing anything with this scope, or is it just authing and storing it?

bdach:
> Also is there anything web-side going to be doing anything with this scope, or is it just authing and storing it?

I don't foresee any web use of the scope.

One thing I will eventually have to figure out is that this refereeing API will probably permit its users to create multiple realtime rooms at a time. Currently a user can only host one realtime room at a time:

https://github.com/ppy/osu-web/blob/a404d7d61fce782abcf4e6cc0ce48594e420a325/app/Models/Multiplayer/Room.php#L810

But I don't foresee the token taking part in this, because the relevant room creation in the realtime multiplayer case happens via an interop endpoint:

https://github.com/ppy/osu-web/blob/f94704239f175f771db64409e8b82ef89c34e358/app/Http/Controllers/InterOp/Multiplayer/RoomsController.php#L35-L43

so I don't think it will involve the user's token, it will probably be an explicit boolean argument in the server-to-server call.

notbakaneko (approved):
I'm just going to leave this open for changes until it needs to go in, then

PR #12737, opened 2026-02-02
Fix TOTP missing clock deprecation

notbakaneko:


notbakaneko:
There's also symfony's `NativeClock` which... defaults to `now` and default timezone

PR #12738, opened 2026-02-02
Adjust profile page play detail on mobile

nanaya:
Accuracy and menu button should be centre aligned to the rank/grade icon...?


notbakaneko (approved):
makes more sense to me

PR #12747, opened 2026-02-04
Skip middleware for screenshot viewing endpoint

nanaya:
Seems like a good idea.

### DIFF
--- a/app/Http/Controllers/ScreenshotsController.php
+++ b/app/Http/Controllers/ScreenshotsController.php
@@ -37,10 +37,10 @@ class ScreenshotsController extends Controller
     public function show($id, ?string $hash = null)
     {
         $screenshot = Screenshot::lookup(intval($id), $hash);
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/ScreenshotsController.php
```diff
@@ -37,10 +37,10 @@ public function store(Request $request)
     public function show($id, ?string $hash = null)
     {
         $screenshot = Screenshot::lookup(intval($id), $hash);
-        abort_if($screenshot === null, 404);
-
-        $file = $screenshot->fetch();
-        abort_if($file === null, 404);
+        $file = $screenshot?->fetch();
+        if ($screenshot === null || $file === null) {
```
notbakaneko: doesn't `$file` null already imply `$screenshot` is null with `?->`?

notbakaneko (changes requested):

notbakaneko (approved):

PR #12749, opened 2026-02-05
Fix nested label checkboxes

notbakaneko:
Before I forget about it again..

nested `label` isn't supposed to be valid; most of the switch includes are already inside a `label` anyway, so I changed it to `div`.
Also removed the `form` in `p` since it was outputing an empty `p` outside of the `form`...

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -999,7 +999,7 @@ class UsersController extends Controller
         $ip = Request::ip();
 
         if (IpBan::where('ip', '=', $ip)->exists()) {
-            return error_popup('Banned IP', 403);
+            return error_popup('Account registration currently unavailable', 403);
         }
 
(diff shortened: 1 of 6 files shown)


resources/views/user_cover_presets/index.blade.php
```diff
@@ -113,16 +114,16 @@ class="btn-osu-big btn-osu-big--rounded-small"
...
+                            </p>
+                        </form>
+
+                        <form
+                            action="{{ route('user-cover-presets.update', $item) }}"
+                            enctype="multipart/form-data"
+                            method="POST"
+                            data-reload-on-success="1"
+                        >
+                            <div class="user-cover-preset-replace">
```
nanaya: the `div` isn't needed? also if the class is adjusted to have margin, the `p` in the form above can be removed

nanaya (approved):

PR #12751, opened 2026-02-05
Adjust mod icon's extended indicator position

nanaya:
Closer to how it looks like in lazer...?


nanaya:
Superseded by #12752.

PR #12754, opened 2026-02-06
Use proper assets url for beatmap search empty result image

nanaya:
Unfortunately the syntax is a bit crap and `pretty` is making it even worse but otherwise it's better-ish.

Alternative includes importing both versions and explicitly specifying it.
```
import imgUrl from '@images/img.png'
import imgUrl2x from '@images/img@2x.png'

...
  <Img2x src={imgUrl} src2x={imgUrl2x} ... />
...
```
Which seems worse? Or is it.


notbakaneko (dismissed):

notbakaneko:
actually thinking about it again, maybe the import is better for more static urls 🤔

nanaya:
the annoying part is the two imports per file as I don't think there's alternative to it short of adding another preprocessor layer...

notbakaneko:
I think that's fine since it doesn't appear often.....yet

notbakaneko (approved):

PR #12755, opened 2026-02-06
 Limit displayed mods on score tables

nanaya:
The simplest possible styling and stuff.

- [x] depends on #12750


stanriders:
This should probably prioritize mods that affect scoring and try to hide only the ones that don't

notbakaneko:
I wonder if there should be a maximum height for the popup, although this scenario seems unlikely (at least 15 mods overflow **and** nothing making the page taller) 🤔
image

resources/js/components/mods.tsx
```diff
@@ -0,0 +1,124 @@
...
+interface Props {
+  maxIcons?: number;
+  modifiers?: Modifiers;
+  mods: ScoreModJson[];
+}
+
+@observer
+export default class Mods extends React.Component<Props> {
+  private overflowMods: Props['mods'] = [];
+  private tooltipDisposer: undefined | (() => void);
```
notbakaneko: `tooltipDisposer?:`?

resources/js/beatmapsets-show/scoreboard/table-row.tsx
```diff
@@ -157,7 +157,7 @@ export default class ScoreboardTableRow extends React.Component<Props> {
 
         <TdLink href={this.scoreUrl} modifiers='mods'>
           <div className={`${bn}__mods`}>
-            {displayMods(score).map((mod) => <Mod key={mod.acronym} mod={mod} />)}
+            <Mods mods={displayMods(score)} />
```
notbakaneko: Also need something to cancel the navigation ontouch, otherwise it'll just load the score page.

notbakaneko (changes requested):

resources/js/beatmapsets-show/scoreboard/table-row.tsx
nanaya: I wonder if the entire section should just not be a link because you can't see the mod tooltip as well...

notbakaneko (approved):

Meruru-chan:
Honestly I would raise the limit of how much is displayed by 1, Because currently any 4mod SS'es done with +SD / PF in addition to it will end up with HardRock being hidden behind the +2 thingy... So you won't be able to tell if the play is actually a 4mod play (with lazer DT rate change) at a glance without hovering...
image

axiomatism:
It would be awesome if this was a toggle thing (not necessary)
It would be awesome if the [+X] icon was the same size as the typical mod icon (please)
image
(i think it would also be cool if rate change took up the space of 2 mods))

PR #12759, opened 2026-02-08
Hide tag editor button on desktop for non-privileged users

LiquidPL:
oops

(shouldn't cause any issues since the edit tags endpoint is properly checking permissions)


notbakaneko (approved):
I could have sworn I checked this..

PR #12768, opened 2026-02-11
User beatmap tag picker

### LINKED ISSUE #12678: Make beatmap listing support searching by user tags in a way usable by humans
Currently searching in beatmap listing "works", for some definition of the word. The downsides are too numerous to even list:

- You have to know what the tags are. Currently the only available way to know what the tags are when interacting with the website is going to [this wiki page](https://osu.ppy.sh/wiki/en/Beatmap/Beatmap_tags). The fact that this wiki page exists is very stupid and [the fact that *it is outdated*](https://osu.ppy.sh/comments/3907098) is even stupider.
- If you're not clicking on the tag from another beatmap page, you have to know what the special syntax for user tag search is. Including that you have to enclose any user tag with spaces in double double quotes.

We should have a proper UI that people can click on and look and see a list of user tags and it will do this filtering instead of the abomination that the current flow is.

LiquidPL:
(Maybe?) resolves #12678.

This adds a component displaying the list of all beatmap tags, allowing users to add one to the search query by clicking it.

| Desktop | Mobile |
|--------|--------|
| image |image |

Kinda RFC, since I'm not sure what should be the way forward here, and also there's still some stuff I'd want to do here:

* Currently the picker is only available on the main (non-sticky) search box. I've tried a few ways of having it in both places, but at this point I couldn't find a way of doing it without straight up having two separate instances of the picker for each searchbox and manually syncing the list scroll position and other things, which is ugly. I decided to hold off doing that in hopes someone can point me towards some better solution here, or maybe it's fine to not have this on the sticky header searchbox anyway.
* The picker is displaying all tags from all rulesets - there is no filtering in place yet. Personally I think that the full list is already pretty long (130+ items on prod), and I'm assuming most people will only care about one ruleset at a time - having tags specific to other ones mixed in with the intended one is only going to make it confusing. Unsure if this should be dealt with through filtering or just adding a ruleset icon to appropriate list items.
* Right now selecting a tag will naively add it to the search query, regardless of whether it's already been added or not. I'll probably tackle that in a future PR, since I was planning to also add some form of autocomplete to the main searchbox. 

EDIT: Rulesets are now filtered based on the currently searched ruleset, and if filtering by all, there's an indicator next to tags that are available in multiple rulesets:

image


resources/js/beatmaps/user-tag-picker-controller.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export interface TagGroup {
+  name: string;
+  tags: BeatmapTag[];
+}
+
+export default class UserTagPickerController {
+  @observable query: string|null = null;
+  @observable showPicker = false;
+  @observable tags = observable.array<BeatmapTag>();
```
nanaya: use normal array?

resources/js/beatmaps/user-tag-picker-controller.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  constructor() {
+    makeObservable(this);
+    onBecomeObserved(this, 'tags', this.fetchTags);
+
+    autorun(() => {
+      if (this.showPicker) {
+        this.query = null;
+      }
```
nanaya: it's probably annoying when the popup is accidentally closed

resources/js/beatmaps/user-tag-picker-controller.ts
```diff
@@ -0,0 +1,61 @@
...
+    });
+  }
+
+  @computed
+  get groups() {
+    const queried = this.query !== null
+      ? this.tags.filter((tag) => tag.fullName.includes(this.query!))
```
nanaya: should ignore case. also store the query in a variable first so it doesn't need to use `!` (or use empty string instead)

resources/js/beatmaps/user-tag-picker-controller.ts
```diff
@@ -0,0 +1,61 @@
...
+      sorted,
+      (tag) => tag.group,
+    );
+    const groups: TagGroup[] = [];
+
+    for (const group of Object.keys(grouped)) {
+      groups.push({ name: group, tags: grouped[group] });
+    }
+
+    return groups;
```
nanaya:
```suggestion
    return Object.entries(grouped).map((name, tags) => ({ name, tags }));
```

resources/js/beatmaps/user-tag-picker.tsx
```diff
@@ -0,0 +1,70 @@
...
+import core from '../osu-core-singleton';
+import { trans } from '../utils/lang';
+import { TagGroup } from './user-tag-picker-controller';
+
+const controller = core.beatmapTagPickerController;
+const beatmapsetSearchController = core.beatmapsetSearchController;
+
+const UserTagPicker = observer(() => {
+  const inputRef = useRef<HTMLInputElement>(null);
+  const scrollViewRef = useRef<HTMLDivElement>(null);
```
nanaya: not used

resources/css/bem/user-tag-picker.less
```diff
@@ -0,0 +1,110 @@
...
+  background-color: @osu-colour-b4;
+
+  @media @mobile and (max-width: $width + @gutter-v2 * 2) {
+    width: calc(100vw - @gutter-v2 * 2);
+  }
+
+  &__search {
+    .default-border-radius();
+
+    width: calc(100% - $padding-top * 2);
```
nanaya: not needed

resources/css/bem/user-tag-picker.less
```diff
@@ -0,0 +1,110 @@
...
+  // to put the scrollbar on the far edge of the element
+  padding-top: 20px;
+  padding-bottom: $padding-top;
+  border-radius: @border-radius--large;
+
+  display: flex;
+  flex-direction: column;
+  gap: 10px;
+
+  color: @osu-colour-c1;
```
nanaya: use `hsl(var(...))` for new styles

resources/css/bem/user-tag-picker.less
```diff
@@ -0,0 +1,110 @@
...
+  background-color: @osu-colour-b4;
+
+  @media @mobile and (max-width: $width + @gutter-v2 * 2) {
+    width: calc(100vw - @gutter-v2 * 2);
+  }
```
nanaya: Note that it's still too wide because it doesn't account for the scrollbar (when viewed on small window desktop)
```suggestion
  width: calc(100vw - @gutter-v2 * 2);
  max-width: 500px;
```

resources/css/bem/user-tag-picker.less
```diff
@@ -0,0 +1,110 @@
...
+
+  &__search {
+    .default-border-radius();
+
+    width: calc(100% - $padding-top * 2);
+    height: 30px;
```
nanaya: use vertical padding instead

resources/css/bem/user-tag-picker.less
```diff
@@ -0,0 +1,110 @@
...
+
+    margin: 0 $padding-top;
```
nanaya: the `$padding-top` is way too far. use explicit variable instead

resources/css/bem/user-tag-picker.less
```diff
@@ -0,0 +1,110 @@
...
+    &::before {
+      content: "";
+      display: inline-block;
+      margin-right: 5px;
+
+      width: 6px;
+      height: 12px;
+      border-radius: 3px;
+
+      background-color: #00ffaa;
```
nanaya: where did this colour come from

resources/css/bem/user-tag-picker.less
```diff
@@ -0,0 +1,110 @@
...
+
+  &__tag {
+    .default-border-radius();
+    width: 100%;
+    min-height: 30px;
+    padding: 5px 10px;
+    flex-shrink: 0;
+
+    display: flex;
+    flex-direction: row;
```
nanaya: use grid instead with template column of `100px 1fr`

resources/js/beatmaps/user-tag-picker-container.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+    return () => {
+      document.removeEventListener('keydown', onKeyDown);
+      document.removeEventListener('mousedown', onDocumentMouseDown);
+      document.removeEventListener('click', onDocumentClick);
+    };
+  }, [onKeyDown, onDocumentMouseDown, onDocumentClick]);
+
+  return (
+    <div ref={container} className='user-tag-picker-button'>
```
nanaya: this should probably use `PopupMenu` instead

routes/web.php
```diff
@@ -623,7 +623,7 @@
         Route::get('wiki/{locale}/{path}', 'WikiController@show')->name('wiki.show')->where('path', '.+');
 
         // Tags
```
notbakaneko, line 625: this just makes the declared scopes hard to find, we don't want to be digging through the routes file to be finding specific middlewares.

resources/js/beatmaps/user-tag-picker-controller.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    autorun(() => {
+      if (this.showPicker) {
+        this.query = null;
+      }
+    });
+  }
+
+  @computed
+  get groups() {
```
nanaya, line 36: need filter by ruleset id (if applicable) and remove duplicate tags

resources/js/beatmaps/user-tag-picker.tsx
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import { action } from 'mobx';
+import { observer } from 'mobx-react';
+import React, { useEffect, useRef } from 'react';
+import BeatmapTag from '../models/beatmap-tag';
+import core from '../osu-core-singleton';
+import { trans } from '../utils/lang';
```
nanaya: only same directory and inner level should use relative path for import

resources/css/bem/user-tag-picker.less
LiquidPL: pulled this out of lazer and forgot to replace with a more appropriate value

resources/js/beatmaps/user-tag-picker-container.tsx
LiquidPL: I'll have to figure something out so that the popup is still aligned with the right edge of the searchbar since I understand that `PopupMenu` will align it with the button instead.

or I guess I could just move the tag button to the right (but that would look kinda bad)

LiquidPL: would it be okay to patch `PopupMenu` like this:

```patch
diff --git a/resources/js/components/popup-menu.tsx b/resources/js/components/popup-menu.tsx
--- a/resources/js/components/popup-menu.tsx	(revision 484a61129e6363603b8ba9a1bd5a5c3bc58bb963)
+++ b/resources/js/components/popup-menu.tsx	(date 1771072290903)
@@ -11,6 +11,7 @@
 
 export interface Props {
   children: Children;
+  customAlignRef?: React.RefObject<HTMLElement>;
   customRender?: (children: React.ReactNode, ref: React.RefObject<HTMLElement>, toggle: (event: React.MouseEvent<HTMLElement>) => void) => React.ReactNode;
   direction?: 'left' | 'right';
   onHide?: () => void;
@@ -149,7 +150,7 @@
   private readonly resize = () => {
     if (!this.state.active) return;
 
-    const buttonEl = this.buttonRef.current;
+    const buttonEl = this.props.customAlignRef ? this.props.customAlignRef.current : this.buttonRef.current;
     const menuEl = this.menuRef.current;
     const menuRootEl = this.menuRootRef.current;
     if (buttonEl == null || menuEl == null || menuRootEl == null) {
@@ -169,7 +170,7 @@
     let left = scrollX + buttonRect.right;
     // shift the menu right if it clips out of the window;
     // menuRect.x doesn't update until after layout is finished so the known position of buttonRect is used instead.
-    if (this.props.direction === 'right' || buttonRect.x - menuRect.width < 0) {
+    if (this.props.direction === 'right' || buttonRect.right - menuRect.width < 0) {
       left += menuRect.width - buttonRect.width;
     }
 
```

This way i could configure it to be aligned to the right side of the input container instead of the tag button.
(also changing `buttonRect.x` to `buttonRect.right` because otherwise it would shift to left before it reaches the edge of the screen)

nanaya: or just pass a different element for the button ref

although I don't know if it's when worth it. the popup is for the button, not the entire input box anyway

LiquidPL: true, on desktop it looks reasonably fine, but on mobile i think it'd look better if it spanned the entire screen

LiquidPL: especially since right now it doesn't reach the left edge anyway as i said

https://github.com/user-attachments/assets/a4ff65b2-a241-485d-aaf6-ad1612cbcb1d

LiquidPL: nevermind i guess i could just

```css
@media @mobile {
  transform: translateX(@button-width);
}
```

or something like that. it still doesn't resolve the issue with shifting left though...

LiquidPL: on the other hand, I guess this does look fine after all

resources/js/models/beatmap-tag.ts
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import { observable } from 'mobx';
+import Ruleset, { rulesetIdToName } from '../interfaces/ruleset';
+import TagJson from '../interfaces/tag-json';
+
+export default class BeatmapTag {
+  @observable description: string;
```
nanaya: pretty sure the decorator doesn't do anything without `makeObservable` call

(and does this even need one at all)

resources/js/beatmaps/user-tag-picker.tsx
```diff
@@ -0,0 +1,73 @@
...
+      ? currentQuery + ` tag="${tagString}"`
+      : `tag="${tagString}"`;
+
+    beatmapsetSearchController.filters.update('query', newQuery);
+  });
+
+  return (<div className='user-tag-picker__tag' onClick={onClick}>
+    <div>
+      {beatmapsetSearchController.filters.mode === null && tag.rulesets.map((ruleset) => (
+        <span key={ruleset} className={`user-tag-picker__tag-info user-tag-picker__tag-info--ruleset fal fa-extra-mode-${ruleset}`} />
```
nanaya: seems like it may be better in the description instead of taking up limited space for the name

resources/css/bem/user-tag-picker.less
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }
+
+  &__tag-info {
+    font-size: @font-size--title-small;
+
+    &--name {
+      width: 100px;
+      font-weight: 700;
+      flex-shrink: 0;
```
nanaya: a bunch of the flex- things and width not needed anymore

resources/css/bem/user-tag-picker.less
```diff
@@ -0,0 +1,113 @@
...
+
+    display: grid;
+    grid-template-columns: 100px 1fr;
+    gap: 10px;
+
+    align-items: center;
+
+    background-color: hsl(var(--hsl-b3));
+
+    overflow: hidden;
```
nanaya: what's this for

resources/js/beatmaps/user-tag-picker-button.tsx
```diff
@@ -0,0 +1,32 @@
...
+import PopupMenu from 'components/popup-menu';
+import { observer } from 'mobx-react';
+import React, { useCallback } from 'react';
+import { trans } from 'utils/lang';
+import UserTagPicker from './user-tag-picker';
+
+const UserTagPickerButton = observer(() => {
+  const renderButton = useCallback((children: React.ReactNode, ref: React.RefObject<HTMLDivElement>, toggle: (event: React.MouseEvent<HTMLElement>) => void) => (
+    <>
+      <div
```
nanaya: button should be button

resources/js/beatmaps/user-tag-picker-button.tsx
```diff
@@ -0,0 +1,32 @@
...
+        ref={ref}
+        className='beatmapsets-search__icon beatmapsets-search__icon--tags'
```
nanaya: the block should always be in the same file. otherwise make this its own block

resources/css/bem/user-tag-picker.less
```diff
@@ -0,0 +1,113 @@
...
+
+  height: 400px;
+
+  // left/right padding is applied on the list
+  // to put the scrollbar on the far edge of the element
+  padding: @padding 0;
+  border-radius: @border-radius--large;
+
+  display: flex;
+  flex-direction: column;
```
nanaya: can just use grid instead. also for the other one

resources/css/bem/user-tag-picker.less
```diff
@@ -0,0 +1,113 @@
...
+  &__list {
+    display: flex;
+    flex-direction: column;
+    gap: 5px;
+    overflow-y: scroll;
+    overscroll-behavior: contain;
+    padding: 0 @padding;
+  }
+
+  &__category {
```
nanaya, line 58: a `position: sticky` would probably be nice

resources/css/bem/user-tag-picker.less
```diff
@@ -0,0 +1,113 @@
...
+      border-radius: 3px;
+
+      background-color: hsl(var(--hsl-h1));
+    }
+  }
+
+  &__tag {
+    .default-border-radius();
+    width: 100%;
+    min-height: 30px;
```
nanaya: not needed?

resources/js/beatmaps/user-tag-picker-controller.ts
```diff
@@ -0,0 +1,117 @@
...
+    const query = this.query;
+    const ruleset = this.ruleset;
+
+    const filtered = ruleset !== null
+      ? this.tags.filter((tag) => tag.ruleset === ruleset || tag.ruleset === null)
+      : this.tags;
+
+    const queried = query !== null
+      ? filtered.filter((tag) => tag.fullName.toLowerCase().includes(query.toLowerCase()))
```
nanaya: do the query.tolowercase once when setting the variable...

resources/js/beatmaps/user-tag-picker-button.tsx
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import PopupMenu from 'components/popup-menu';
+import { observer } from 'mobx-react';
+import React, { useCallback } from 'react';
+import { trans } from 'utils/lang';
+import UserTagPicker from './user-tag-picker';
+
+const UserTagPickerButton = observer(() => {
```
nanaya: doesn't observe anything?

resources/js/beatmaps/user-tag-picker-button.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const renderButton = useCallback((children: React.ReactNode, ref: React.RefObject<HTMLDivElement>, toggle: (event: React.MouseEvent<HTMLElement>) => void) => (
```
nanaya: doesn't need to be defined inside this function (and thus doesn't need `useCallback`)

resources/js/beatmaps/user-tag-picker-button.tsx
```diff
@@ -0,0 +1,32 @@
...
+      >
+        <i className='fas fa-tag' />
+      </div>
+      {children}
+    </>
+  ), []);
+
+  return (
+    <PopupMenu customRender={renderButton} direction='left'>
+      {(_) => <UserTagPicker />}
```
nanaya: what's `(_)` doing there instead of just `()`

resources/js/beatmaps/user-tag-picker.tsx
```diff
@@ -0,0 +1,73 @@
...
+const controller = core.beatmapTagPickerController;
+const beatmapsetSearchController = core.beatmapsetSearchController;
+
+const UserTagPicker = observer(() => {
+  const inputRef = useRef<HTMLInputElement>(null);
+  const onChange = action((e: React.ChangeEvent<HTMLInputElement>) => controller.query = e.target.value);
+
+  useEffect(() => {
+    inputRef.current?.focus();
+  }, [inputRef]);
```
nanaya: I'm pretty sure the dependency should be an empty array here (ref doesn't change ever so it doesn't do anything for effect)

resources/js/beatmaps/user-tag-picker.tsx
```diff
@@ -0,0 +1,73 @@
...
+import React, { useEffect, useRef } from 'react';
+import { trans } from 'utils/lang';
+import { BeatmapTagManyRulesets, TagGroup } from './user-tag-picker-controller';
+
+const controller = core.beatmapTagPickerController;
+const beatmapsetSearchController = core.beatmapsetSearchController;
+
+const UserTagPicker = observer(() => {
+  const inputRef = useRef<HTMLInputElement>(null);
+  const onChange = action((e: React.ChangeEvent<HTMLInputElement>) => controller.query = e.target.value);
```
nanaya: useCallback

resources/js/beatmaps/user-tag-picker.tsx
```diff
@@ -0,0 +1,73 @@
...
+    {group.tags.map((tag) => <UserTag key={tag.id} tag={tag} />)}
+  </>
+));
+
+const UserTag = observer(({ tag }: { tag: BeatmapTagManyRulesets }) => {
+  const onClick = action(() => {
+    let tagString = tag.fullName;
+
+    if (/\s/g.test(tagString)) {
+      tagString = `"${tagString}"`;
```
nanaya: the quoting doesn't matter at the moment so just make them uniform

resources/js/beatmaps/user-tag-picker-controller.ts
```diff
@@ -0,0 +1,117 @@
...
+      (result, tag) => {
+        const existingTagIndex = findIndex(result, (t) => t.fullName === tag.fullName);
+
+        if (existingTagIndex === -1) {
+          const tagToAdd = {
+            ...tag,
+            rulesets: [],
+          } as BeatmapTagManyRulesets;
+
+          if (tag.ruleset !== null) tagToAdd.rulesets.push(tag.ruleset);
```
nanaya: I wonder if the BeatmapTag model itself should just have `ruleset(_ids)` instead

resources/js/models/beatmap-tag.ts
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import { observable } from 'mobx';
+import Ruleset, { rulesetIdToName } from '../interfaces/ruleset';
```
nanaya: no relative import if going up...

resources/js/beatmaps/user-tag-picker-controller.ts
```diff
@@ -0,0 +1,117 @@
...
+  private readonly fetchTags = () => {
+    if (this.tags.length > 0) {
+      return;
+    }
+
+    $.get(route('tags.index'))
+      .done(action((tags: { tags: TagJson[] }) => this.tags = tags.tags.map((tag) => new BeatmapTag(tag))));
+  };
+
+  private rulesetOrdering(ruleset: Ruleset|null) {
```
nanaya: wouldn't be needed if ruleset_id is kept

app/Http/Controllers/TagsController.php
```diff
@@ -13,7 +13,7 @@ public function __construct()
     {
         parent::__construct();
 
-        $this->middleware('require-scopes:public');
+        $this->middleware(['auth', 'require-scopes:public']);
```
nanaya: should not change the existing api

resources/js/beatmaps/user-tag-picker-controller.ts
```diff
@@ -0,0 +1,117 @@
...
+  name: string;
+  tags: BeatmapTagManyRulesets[];
+}
+
+export type BeatmapTagManyRulesets = Omit<BeatmapTag, 'ruleset'> & {
+  rulesets: Ruleset[];
+};
+
+export default class UserTagPickerController {
+  @observable query: string|null = null;
```
nanaya: being nullable doesn't do anything useful. and it turns empty string after filled in at least once anyway.

resources/js/beatmaps/user-tag-picker-controller.ts
```diff
@@ -0,0 +1,117 @@
...
+      ? this.tags.filter((tag) => tag.ruleset === ruleset || tag.ruleset === null)
+      : this.tags;
+
+    const queried = query !== null
+      ? filtered.filter((tag) => tag.fullName.toLowerCase().includes(query.toLowerCase()))
+      : filtered;
+
+    const sortPredicates: ((tag: BeatmapTag) => any)[] = [((tag) => tag.fullName)];
+
+    // Order tags by ruleset when all rulesets are displayed so that they are not
```
nanaya: I don't know if this is useful considering there are tags for multiple (but not all) rulesets which means they'll be only "grouped" and visible in the first ruleset section and missing in the other rulesets.

resources/js/beatmaps/user-tag-picker.tsx
```diff
@@ -0,0 +1,73 @@
...
+    const newQuery = currentQuery !== null
+      ? currentQuery + ` tag="${tagString}"`
+      : `tag="${tagString}"`;
+
+    beatmapsetSearchController.filters.update('query', newQuery);
+  });
+
+  return (<div className='user-tag-picker__tag' onClick={onClick}>
+    <div>
+      {beatmapsetSearchController.filters.mode === null && tag.rulesets.map((ruleset) => (
```
nanaya: there are some tags which have its own id for all rulesets. those should behave the same as if there's no rulesets specified

resources/js/beatmaps/user-tag-picker.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { observer } from 'mobx-react';
+import core from 'osu-core-singleton';
+import React, { useEffect, useRef } from 'react';
+import { trans } from 'utils/lang';
+import { BeatmapTagManyRulesets, TagGroup } from './user-tag-picker-controller';
+
+const controller = core.beatmapTagPickerController;
+const beatmapsetSearchController = core.beatmapsetSearchController;
+
+const UserTagPicker = observer(() => {
```
nanaya: this syntax results in nameless components when viewed in react dev tools. It needs to be `const UserTagPicker = observer(function UserTagPicker() { ... })` ([or other alternatives](https://mobx.js.org/react-integration.html#displayname))

also for other components

resources/css/bem/beatmapsets-search.less
```diff
@@ -92,40 +88,45 @@
     .@{_top}--sticky & {
       width: @_icon-width--sticky + 20px;
     }
+
+    &--tags {
```
nanaya, line 92: colour change on activation and hover would be useful

nanaya (changes requested):

resources/js/beatmaps/user-tag-picker-controller.ts
LiquidPL: image

i dunno, without at least some form of ordering this looks messy to me. 

however, according to [wiki](https://github.com/ppy/osu-wiki/blob/master/wiki/Beatmap/Beatmap_tags/en.md?plain=1#L25) these are supposed to have custom ordering eventually, so maybe that's fine for now? i'll remove this then.

resources/js/models/beatmap-tag.ts
LiquidPL: :upside_down_face: 

at this point i should just configure a linter to catch them for me since they slip through even if i try my best to fix them every time

resources/js/beatmaps/user-tag-picker-controller.ts
```diff
@@ -0,0 +1,96 @@
...
+        if (existingTagIndex === -1) {
+          result.push(tag);
+        } else {
+          result[existingTagIndex].rulesetIds.push(...tag.rulesetIds);
+        }
+
+        return result;
+      },
+      [],
+    );
```
nanaya: that's a lot of lines...
```
const tagByname = {}
for (const tagJson of tags) {
  const tag = tagByName[...]
  if (tag == null) {
    tagByName[...] = new BeatmapTag(...)
  } else if (tagJson.ruleset_id != null) {
    tag.rulesetIds.push(...)
  }
}
const unique = sortBy(Object.values(tagByName), 'fullName')
```

resources/js/beatmaps/user-tag-picker.tsx
```diff
@@ -0,0 +1,88 @@
...
+    .includes(tagString.toLowerCase());
+
+  return (<div className={classWithModifiers('user-tag-picker__tag', { active })} onClick={onClick}>
+    <span className='user-tag-picker__tag-info user-tag-picker__tag-info--name'>{tag.name}</span>
+    <div>
+      {beatmapsetSearchController.filters.mode === null && !hasAllRulesets && tag.rulesetIds.map((ruleset) => (
+        <span key={rulesetIdToName[ruleset]} className={`user-tag-picker__tag-info user-tag-picker__tag-info--ruleset fal fa-extra-mode-${rulesetIdToName[ruleset]}`} />
+      ))}
+      <span className='user-tag-picker__tag-info user-tag-picker__tag-info--description'>{tag.description}</span>
+    </div>
```
nanaya: this doesn't make structural sense.
```
span.__tag-info (name)
div
  span.__tag-info (rulesets)
  span.__tag-info (description)
```

it should just be
```
span.__tag-info (name)
span.__tag-info (rulesets, description (all inline))
```

resources/js/beatmaps/user-tag-picker.tsx
```diff
@@ -0,0 +1,88 @@
...
+  }, [tagString]);
+
+  const hasAllRulesets = reduce(
+    rulesetIds,
+    (hasAll, id) => hasAll && (tag.rulesetIds as readonly number[]).includes(id),
+    true,
+  );
+
+  const active = beatmapsetSearchController.filters.query?.toLowerCase()
+    .includes(tagString.toLowerCase());
```
nanaya: this should at least disable the button from adding redundant search query. Or alternatively clicking on an active one should remove it instead.

resources/js/beatmaps/user-tag-picker.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  return (
+    <div className='user-tag-picker'>
+      <input
+        ref={inputRef}
+        className='user-tag-picker__search'
+        name='tag-search'
+        onChange={onChange}
+        placeholder={trans('beatmaps.listing.search.tag_picker.prompt')}
+        value={controller.query ?? ''}
```
nanaya: already non nullable

resources/js/beatmaps/user-tag-picker.tsx
```diff
@@ -0,0 +1,88 @@
...
+      : tagString;
+
+    beatmapsetSearchController.filters.update('query', newQuery);
+  }, [tagString]);
+
+  const hasAllRulesets = reduce(
+    rulesetIds,
+    (hasAll, id) => hasAll && (tag.rulesetIds as readonly number[]).includes(id),
+    true,
+  );
```
nanaya: assuming it's unique, just compare the length?

resources/js/beatmaps/user-tag-picker.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  const active = beatmapsetSearchController.filters.query?.toLowerCase()
+    .includes(tagString.toLowerCase());
+
+  return (<div className={classWithModifiers('user-tag-picker__tag', { active })} onClick={onClick}>
+    <span className='user-tag-picker__tag-info user-tag-picker__tag-info--name'>{tag.name}</span>
+    <div>
+      {beatmapsetSearchController.filters.mode === null && !hasAllRulesets && tag.rulesetIds.map((ruleset) => (
+        <span key={rulesetIdToName[ruleset]} className={`user-tag-picker__tag-info user-tag-picker__tag-info--ruleset fal fa-extra-mode-${rulesetIdToName[ruleset]}`} />
```
nanaya: no need to map the `ruleset(Id)` for the `key`...

nanaya (changes requested):

resources/js/beatmaps/user-tag-picker-controller.ts
notbakaneko: if you're going to modify an array in `reduce` like this, you might as well use a more readable `for...of`

resources/js/beatmaps/user-tag-picker.tsx
notbakaneko: clicking again should remove it, it'd be annoying to click to activate but then be manually deleting it

resources/js/beatmaps/search-panel.tsx
```diff
@@ -47,13 +49,21 @@ const Filter = observer(({ multiselect = false, name, options, grid = false }: F
...
 
+  @computed
+  private get query() {
+    return this.controller.filters.query ?? '';
+  }
+
+  private set query(query: string) {
+    this.controller.filters.update('query', query);
+  }
+
```
notbakaneko, line 65: This breaks typing space into the search input, the display state and internal state are explicitly handled separately for this reason.

resources/js/beatmaps/user-tag-picker-controller.ts
```diff
@@ -0,0 +1,96 @@
...
+  @computed
+  get groups() {
+    const ruleset = this.ruleset;
+
+    const filtered = ruleset != null
+      ? this.tags.filter((tag) => tag.rulesetIds.length === 0 || tag.rulesetIds.includes(ruleset))
+      : this.tags;
+
+    const queried = filtered.filter(
+      (tag) => tag.fullName.toLowerCase().includes(this.query.toLowerCase()),
```
notbakaneko: move the matching to `BeatmapTag`; also, might as well keep a lower case copy of the name in BeatmapTag to compare

resources/js/beatmaps/user-tag-picker-controller.ts
```diff
@@ -0,0 +1,96 @@
...
+    return Object.entries(grouped).map(([name, tags]) => ({ name, tags }));
+  }
+
+  @action
+  private readonly fetchTags = () => {
+    if (this.tags.length > 0) {
+      return;
+    }
+
+    $.get(route('tags.index'))
```
notbakaneko: use `getJSON` when expecting json

resources/js/beatmaps/user-tag-picker-controller.ts
```diff
@@ -0,0 +1,96 @@
...
+  name: string;
+  tags: BeatmapTag[];
+}
+
+export default class UserTagPickerController {
+  @observable query: string = '';
+  @observable tags: BeatmapTag[] = [];
+
+  @computed
+  private get ruleset(): RulesetId|undefined {
```
notbakaneko: `rulesetId()`

resources/js/beatmaps/user-tag-picker.tsx
LiquidPL: yeah i was afraid that this would be not enough :sweat_smile:

resources/js/beatmaps/search-panel.tsx
LiquidPL: in this case I need a way to get the state update from the tag component to the display observable somehow. would having a `reaction()` or something of that sort in `SearchPanel` to watch for updates on the internal state like this be okay:

```ts
reaction(
  () => this.controller.filters.query,
  (newQuery) => {
    if (this.query === newQuery) return;

    this.query = newQuery ?? '';
  },
);
```

however if i understand mobx correctly this would make the updates when typing in the input to be fired twice so that's not good either...

resources/js/beatmaps/user-tag-picker.tsx
LiquidPL: if we go for removing then i guess i should move the query changing logic to `BeatmapsetSearchFilters` as well

but then again there's the problem of getting that changed query back to the `SearchPanel` input

resources/js/beatmaps/search-panel.tsx
LiquidPL: alternatively could i change `BeatmapsetSearchFilters` to trim the query only when the search request is being made? is `filters.query` used in some any other places that would prevent doing this?

nanaya: just move the field to the controller or the filter (as different name)

the trimmed query is at least also used for generating url

resources/js/beatmaps/search-panel.tsx
```diff
@@ -180,6 +185,28 @@ export class SearchPanel extends React.Component<Props> {
     );
   }
 
+  private renderTagPicker() {
+    return(
+      <PopupMenu customRender={this.renderTagPickerButton} direction='left'>
```
notbakaneko: this needs an update with the removal of `customRender` in #12785 and switch to using `skipButton`

app/Http/Controllers/TagsController.php
```diff
@@ -18,6 +20,10 @@ public function __construct()
 
     public function index()
     {
+        if (!is_api_request() && \Auth::check() === false) {
```
nanaya: this should check for advanced search priv instead

resources/css/bem/beatmapsets-search.less
nanaya: still need to be done

resources/js/beatmapset-search-filters.ts
```diff
@@ -131,7 +141,31 @@ export class BeatmapsetSearchFilters {
...
+
+  @action
+  tagRemove(tag: BeatmapTag) {
+    const currentQuery = this.queryClean;
+
+    if (currentQuery === null) {
+      return;
+    }
+
+    const newQuery = currentQuery.replace(tag.tagString(), '').trim();
```
nanaya: should also remove double space in case of removing tag query in the middle

resources/js/beatmapset-search-filters.ts
```diff
@@ -148,4 +182,10 @@ export class BeatmapsetSearchFilters {
 
     this[key] = value === this.getDefault(key) ? null : value;
   }
+
+  private getValue(key: FilterKey) {
+    if (key === 'query') return this.queryClean;
+
+    return this[key];
+  }
```
nanaya: I think I'd rather update the original intercept to do something like `this.queryRaw = change.newValue` than this

(or replace it with getter/setter with similar effect)

resources/js/beatmaps/user-tag-picker.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const controller = core.beatmapTagPickerController;
+const beatmapsetSearchController = core.beatmapsetSearchController;
+
+export default observer(function UserTagPicker() {
+  const inputRef = useRef<HTMLInputElement>(null);
+  const onChange = useCallback(
+    (e: React.ChangeEvent<HTMLInputElement>) => runInAction(() => controller.query = e.target.value),
+    [],
+  );
```
nanaya: can just be normal function outside the component

resources/js/beatmaps/user-tag-picker-controller.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  constructor() {
+    makeObservable(this);
+    onBecomeObserved(this, 'tags', this.fetchTags);
+  }
+
+  @computed
+  get groups() {
+    const ruleset = this.rulesetId;
```
nanaya: `rulesetId` (also other places)

resources/js/models/beatmap-tag.ts
```diff
@@ -0,0 +1,45 @@
...
+import { RulesetId } from 'interfaces/ruleset';
+import TagJson from 'interfaces/tag-json';
+
+export default class BeatmapTag {
+  description: string;
+  fullName: string;
+  fullNameLowercase: string;
+  group: string;
+  id: number;
+  name: string;
```
nanaya, line 11: this naming is confusing as it contains value different to the original json

resources/js/models/beatmap-tag.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    this.group = split[0];
+    this.name = split[1];
+  }
+
+  matchesFullName(match: string) {
+    return this.fullNameLowercase.includes(match.toLowerCase());
+  }
+
+  tagString() {
```
nanaya: maybe more like queryString or toQuery

resources/js/interfaces/ruleset.ts
```diff
@@ -8,13 +8,24 @@ export const rulesetIdToName = {
...
     return maybeRuleset as Ruleset;
   }
 }
 
+export function ensureRulesetId(maybeRulesetId: number|string): RulesetId | undefined {
+  if (typeof maybeRulesetId === 'string') {
+    maybeRulesetId = parseInt(maybeRulesetId, 10);
+  }
+
+  if ((rulesetIds as readonly number[]).includes(maybeRulesetId)) {
```
nanaya: probably simpler as `rulesetIdToName[maybeRulesetId as RulesetId] != null`

resources/js/beatmaps/user-tag-picker.tsx
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }, [tag, active]);
+
+  const hasAllRulesets = tag.rulesetIds.length === rulesets.length;
+
+  return (<div className={classWithModifiers('user-tag-picker__tag', { active })} onClick={onClick}>
+    <span className='user-tag-picker__tag-info user-tag-picker__tag-info--name'>{tag.name}</span>
+    <span className='user-tag-picker__tag-info user-tag-picker__tag-info--description'>
+      {beatmapsetSearchController.filters.mode === null && !hasAllRulesets && tag.rulesetIds.map((ruleset) => (
+        <span key={ruleset} className={`user-tag-picker__tag-ruleset fal fa-extra-mode-${rulesetIdToName[ruleset]}`} />
```
nanaya: style class shouldn't be mixed with icon classes.

although instead of css styling the space can be actual space (`{' '}`)

resources/js/interfaces/ruleset.ts
```diff
@@ -8,13 +8,24 @@ export const rulesetIdToName = {
...
 export const rulesets = Object.values(rulesetIdToName);
+export const rulesetIds = Object.keys(rulesetIdToName).map((id) => parseInt(id, 10));
 
 export function ensureRuleset(maybeRuleset: string): Ruleset | undefined {
   if ((rulesets as readonly string[]).includes(maybeRuleset)) {
     return maybeRuleset as Ruleset;
   }
 }
 
+export function ensureRulesetId(maybeRulesetId: number|string): RulesetId | undefined {
```
nanaya: `|` in ts is styled with space around it

resources/css/bem/beatmapsets-search.less
LiquidPL: i'm still not sure about the colors here

| normal | hover | active |
|--------|--------|--------|
| image | image| image |

the hover state feels not different enough to me :thinking:

app/Http/Controllers/TagsController.php
```diff
@@ -18,6 +20,10 @@ public function __construct()
 
     public function index()
     {
+        if (!is_api_request() && priv_check('BeatmapsetAdvancedSearch')->can() === false) {
+            throw new AuthenticationException('User is not logged in.');
+        }
```
nanaya: just use `ensureCan()`. it already handles api requests correctly

resources/css/bem/beatmapsets-search.less
```diff
@@ -92,40 +88,55 @@
...
     font-size: @font-size--large;
     margin: 20px 0;
     position: relative;
 
+    border: 1px solid hsl(var(--hsl-b4));
+    background-color: hsl(var(--hsl-b2));
+
+    display: flex;
+    flex-direction: row;
+    justify-content: stretch;
```
nanaya: these don't seem to do anything?

resources/js/beatmaps/user-tag-picker.tsx
```diff
@@ -0,0 +1,77 @@
...
+  }, [tag, active]);
+
+  const hasAllRulesets = tag.rulesetIds.length === rulesets.length;
+
+  return (<div className={classWithModifiers('user-tag-picker__tag', { active })} onClick={onClick}>
+    <span className='user-tag-picker__tag-info user-tag-picker__tag-info--name'>{tag.tagName}</span>
+    <span className='user-tag-picker__tag-info user-tag-picker__tag-info--description'>
+      {beatmapsetSearchController.filters.mode === null && !hasAllRulesets && tag.rulesetIds.map((ruleset) => (<>
+        <span key={ruleset} className={`user-tag-picker__tag-ruleset fal fa-extra-mode-${rulesetIdToName[ruleset]}`} />{' '}
+      </>))}
```
nanaya: the key needs to be in the React.Fragment (`<>`)

also the nonexistent __tag-ruleset class is still there

resources/js/beatmapset-search-filters.ts
```diff
@@ -54,6 +55,7 @@ export class BeatmapsetSearchFilters {
   @observable nsfw: FilterValueType = null;
   @observable played: FilterValueType = null;
   @observable query: FilterValueType = null;
+  @observable queryRaw: FilterValueType = null;
```
nanaya: it's more useful as a string

resources/js/interfaces/ruleset.ts
```diff
@@ -8,13 +8,24 @@ export const rulesetIdToName = {
   3: 'mania',
 } as const;
 export const rulesets = Object.values(rulesetIdToName);
+export const rulesetIds = Object.keys(rulesetIdToName).map((id) => parseInt(id, 10));
```
nanaya: not used?

resources/js/beatmaps/user-tag-picker.tsx
```diff
@@ -0,0 +1,77 @@
...
+import core from 'osu-core-singleton';
+import React, { useCallback, useEffect, useRef } from 'react';
+import { classWithModifiers } from 'utils/css';
+import { trans } from 'utils/lang';
+import { TagGroup } from './user-tag-picker-controller';
+
+const controller = core.beatmapTagPickerController;
+const beatmapsetSearchController = core.beatmapsetSearchController;
+
+const onChange = (e: React.ChangeEvent<HTMLInputElement>) => runInAction(() => controller.query = e.target.value);
```
nanaya: wrap in `action` instead of creating another function with `runInAction`

resources/js/beatmapset-search-filters.ts
```diff
@@ -134,6 +137,33 @@ export class BeatmapsetSearchFilters {
     return this[key] ?? this.getDefault(key);
   }
 
+  @action
+  tagAdd(tag: BeatmapTag) {
+    const currentQuery = this.query;
+    const tagString = tag.toQuery();
+
+    const newQuery = currentQuery !== null
+      ? currentQuery + ` ${tagString}`
```
nanaya: weird mixture of string interpolation method

resources/js/beatmapset-search-filters.ts
```diff
@@ -134,6 +137,33 @@ export class BeatmapsetSearchFilters {
     return this[key] ?? this.getDefault(key);
   }
 
+  @action
+  tagAdd(tag: BeatmapTag) {
+    const currentQuery = this.query;
+    const tagString = tag.toQuery();
+
+    const newQuery = currentQuery !== null
```
nanaya, line 150: adding a space at the of end of the new query string may be useful?

resources/css/bem/beatmapsets-search.less
nanaya: turn white (`c1`) instead? And I think it can use the same color between active and hover (where did the `h1` come from?)

resources/css/bem/user-tag-picker.less
```diff
@@ -0,0 +1,120 @@
...
+    .default-border-radius();
+
+    margin: 0 @padding;
+    padding: 8px 5px;
+
+    outline: none;
+    border: none;
+
+    background-color: hsl(var(--hsl-b5));
+    font-size: 14px;
```
nanaya: non-icon font size should use the `@font-size--` variable

also I think it needs to be 16px (`@font-size--phone-input`) for it to not zoom in ios?

resources/js/models/beatmap-tag.ts
```diff
@@ -0,0 +1,46 @@
...
+      this.rulesetIds.push(tag.ruleset_id);
+    }
+
+    const split = this.name.split('/');
+
+    this.categoryName = split[0];
+    this.tagName = split[1];
+  }
+
+  matchesFullName(match: string) {
```
nanaya: `matchesName`?

resources/css/bem/beatmapsets-search.less
LiquidPL: i was looking through the color system document and it just made sense for me at that point somehow :thinking:

PR #12770, opened 2026-02-12
Allow multiple 'tournament mode' realtime multiplayer rooms per user

bdach:
## [Rename per-user playlist count limit envvars to better reflect purpose](https://github.com/ppy/osu-web/commit/43c70ac6c4ffd172cb3cdd81588d0eff23c6e73f)

- `USER_MAX_MULTIPLAYER_ROOMS` -> `USER_MAX_PLAYLISTS`
- `USER_MAX_MULTIPLAYER_ROOMS_SUPPORTER` -> `USER_MAX_PLAYLISTS_SUPPORTER`

Was pretty confusing that it's a "max multiplayer rooms" envvar but also one that's only applied to playlists, IMO. Also I'm about to add more envvars and make it worse.

## [Allow multiple 'tournament mode' realtime multiplayer rooms per user](https://github.com/ppy/osu-web/commit/1b5fcad1d69ffa388423740cfa3f1e8dff383cce)

'Tournament mode' here referring to rooms created via the new refereeing API that I'm working on, namely https://github.com/ppy/osu-server-spectator/issues/406.

Of note, this is steered via an ephemeral flag passed via the shared interop room creation request. I can't *currently* think of any reason as to why we might want to persist that flag, but maybe it is something we wish to do. Not sure, open to discussion on that front. (@peppy you might have an opinion on this?)

Limits enforced here match bancho (4 simultaneous rooms for normal users, 50 for bots).


peppy:
> Of note, this is steered via an ephemeral flag passed via the shared interop room creation request. I can't currently think of any reason as to why we might want to persist that flag, but maybe it is something we wish to do. Not sure, open to discussion on that front. (@peppy you might have an opinion on this?)

Historically, we have inferred tournament matches to – as one example – persist history and avoid them being cleaned up as part of the one-month-multiplayer-history purging. I think persisting that flag into the rooms' databased metadata may be useful. Even for analytic or debug purposes it might be handy to have around.

If it's not too much effort, maybe shove it in the `multiplayer_rooms` table (unless I'm missing something and we already know the source of a room somehow).

bdach:
> If it's not too much effort, maybe shove it in the `multiplayer_rooms` table (unless I'm missing something and we already know the source of a room somehow).

https://github.com/ppy/osu-web/pull/12770/commits/349708c8607372c1227a535a38d07971fc789656 should do it I think.

app/Models/Multiplayer/Room.php
```diff
@@ -47,6 +47,7 @@
  * @property \Illuminate\Database\Eloquent\Collection $scoreLinks ScoreLink
  * @property-read Season $season
  * @property \Carbon\Carbon $starts_at
+ * @property bool $tournament_mode
```
nanaya, line 50: also needs an entry in the `$casts` array

tests/Controllers/InterOp/Multiplayer/RoomsControllerTest.php
```diff
@@ -98,4 +98,71 @@ public function testStore(): void
...
+    public function testUserCanStoreMoreRoomsIfTournamentMode(): void
+    {
+        $beatmap = Beatmap::factory()->create();
+        $params = [
+            ...static::startRoomParams(),
+            'user_id' => User::factory()->create()->getKey(),
+            'tournament_mode' => true,
+        ];
+
+        $this->expectCountChange(fn () => Room::count(), 4);
```
nanaya: this probably should use the config variable (`$GLOBALS['cfg'][...]` or the method in user model)...? same with the other test

app/Models/Multiplayer/Room.php
bdach: keep forgetting about these magic dingdongs...

tests/Controllers/InterOp/Multiplayer/RoomsControllerTest.php
bdach: Does https://github.com/ppy/osu-web/pull/12770/changes/e31ab1ce1b0afdecb1d8204b1df2fe9504fa4894 work for you?

nanaya: the bot and non-bot count don't really relate to each other and nothing checks one is higher than the other...?

maybe do a `config_set('osu.user...', $normalUser->maxTournamentRooms() + 2)` beforehand and work from there?

bdach: I half expected that response but wasn't generally sure how pedantic to be about this 😅 

https://github.com/ppy/osu-web/pull/12770/commits/490dfc6988d4f6dbd3752146e4da498241cc6f25 maybe?

nanaya (approved):

PR #12771, opened 2026-02-12
Fix overflow from mod custom indicator

nanaya:
Not sure how I missed this...

Also I don't know what's with the weird positioning if I don't change the display to grid.


notbakaneko:
It looks like the `svg` element itself is treated as `inline`, so not actually the full height of the container unless the container is `flex` or `grid` then it becomes `block`... (or `display: block` set directly on the svg 🤷 )

notbakaneko (approved):

PR #12772, opened 2026-02-13
Ensure assets are built on github action

nanaya:
I think this should work? There's probably a smarter way to do this.

For some reason the `logs` command on gha doesn't exit if the service already exited before running the command... Or something.

And the `wait` command also exit with error code 1 if the service already exited, thus the `|| true`

*ensure as in ensure it's built before running more tests that require the assets


peppy:
Is this fixing some test failures?

nanaya:
yes https://github.com/ppy/osu-web/actions/runs/21978237059/job/63494503823?pr=12755

notbakaneko (approved):

PR #12777, opened 2026-02-16
Add more details to contest judging results dropdown

### LINKED ISSUE #12775: Contest judging results page dropdown/breadcrumb should display the user name, not entry name
we can assume that users entering a contest are unique, but the entry name is not guaranteed to be unique

from https://osu.ppy.sh/community/contests/267/entries/66756/results:

Image

even when the entry names are unique, they're not that informative in the dropdown and breadcrumb here

Image

slightly related but it would be good if the window title displayed the user (and probably contest title) too

Image

notbakaneko:
Shows the username with the entry in the drop down.
Also adds the missing cache invalidation when calculating standardised scores and renames the cache keys for more consistency.
...and fix the actual json being missing from the response for `ContestEntryJsonForResults` type.
...and more results caching instead of summing the scores on every request.
...and then use the existing cached values.

This ended up being a rabbit hole of changes. 👀 

closes #12775

Doing the browser title update separately since I want to figure out if there's a better way to update it on navigation without the doing all the viewsMounted workarounds used by chat, so it can be applied to other react-side navigation history updates as well.

### DIFF
--- a/app/Http/Controllers/ContestEntriesController.php
+++ b/app/Http/Controllers/ContestEntriesController.php
@@ -44,20 +44,14 @@ class ContestEntriesController extends Controller
         $entries = $contest->entriesByType(null)->loadMissing($relationships);
         $entry = $entries->findOrFail($id);
 
-        $contestJson = json_item(
-            $contest,
-            new ContestTransformer(),
-            [
-                'max_judging_score',
-                'max_total_score',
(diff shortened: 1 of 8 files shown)


nanaya:
how about separating the prs ಠ_ಠ

nanaya:
This can probably be closed?

PR #12781, opened 2026-02-18
Invalidate cache after calculating Contest standardised scores

notbakaneko:
splitting/reordering #12777

Adds the missing cache invalidation when calculating standardised scores and renames the cache keys for more consistency. The preloads list is ignored because it doesn't make sense to cache results based on a variable without being keyed by it, and the eager loading for results is always the same, anyway.

### DIFF
--- a/app/Models/Contest.php
+++ b/app/Models/Contest.php
@@ -319,26 +319,17 @@ class Contest extends Model
         }
     }
 
-    public function entriesByType(?User $user, array $preloads = [])
+    public function entriesByType(?User $user)
     {
         if ($this->show_votes) {
-            $includes = [
-                'contest',
-                'user' => fn ($q) => $q->select('user_id', 'username'),
-            ];
-
-            if ($this->isJudged()) {
-                $includes[] = 'judgeVotes.scores';
-            }
-
             return \Cache::remember(
                 $this->entriesWithScoresCacheKey(),
                 300,
-                fn () => $this->entries()->with($includes)->withScore($this)->get()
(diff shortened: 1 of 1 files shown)


app/Models/Contest.php
```diff
@@ -470,4 +482,20 @@ public function showEntryUser(): bool
...
+        return "contest:{$this->getKey()}:entries_with_scores";
+    }
+
+    private function resetCache(): void
+    {
+        \Cache::forget($this->entriesWithScoresKey());
+        \Cache::forget($this->usersVotedCountCacheKey());
+    }
+
+    private function usersVotedCountCacheKey(): string
```
nanaya, line 488: one is suffixed with `CacheKey` and the other one is just `Key`?

app/Models/Contest.php
```diff
@@ -320,15 +321,26 @@ public function vote(User $user, ContestEntry $entry)
...
-            return Cache::remember(
-                "contest_entries_with_votes_{$this->id}",
+            $includes = [
+                'contest',
+                'user' => fn ($q) => $q->select('user_id', 'username'),
+            ];
+
+            if ($this->isJudged()) {
+                $includes[] = 'judgeVotes.scores';
+            }
```
nanaya: shouldn't they be inside the cache callback function instead?

although I wonder if those are just better preloaded later. the main thing cached here is the count and the `contest` include causes the cache to contain the same contest repeatedly

...heck, the `$preloads` parameter in this function can probably even just be removed and have the callers load them by themselves

nanaya (approved):

PR #12782, opened 2026-02-18
Wait for token validation result

TheDomco:
The validation key can be anything, just has to be unique per request.
The results are strings such as `invalid token`  which can be returned to the user (similar to the ones in `parseToken`).
The push happens earlier for registration now, no `user_id` needed anymore.


app/Libraries/ClientCheck.php
```diff
@@ -67,19 +68,38 @@ public static function parseToken(Request $request): array
...
             'token' => $tokenData['token'],
             'url' => $tokenData['url'],
-            'user_id' => $userId,
+            'validation_key' => $validationKey,
         ]));
+
+        $result = \LaravelRedis::blPop([$validationKey], $GLOBALS['cfg']['osu']['client']['token_validation_timeout']);
+        $timedOut = empty($result);
+
+        if ($timedOut) {
```
nanaya: is there any other possible fail value...?
(and it looks like it throws on connection failure but none of the existing codes handle it anyway)
```suggestion
        if ($result === null) {
```

app/Libraries/ClientCheck.php
```diff
@@ -89,6 +109,18 @@ private static function getKey(Build $build): string
...
+    private static function getMultipartData(Request $request): ?string
+    {
+        $contentType = $request->header('Content-Type', '');
+        $isMultipart = str_contains($contentType, 'multipart/form-data');
+
+        if (!$isMultipart) {
+            return null;
+        }
+
+        return json_encode($request->post());
```
nanaya: can just be a single return with ternary

also I suppose it's intentionally json-string-encoded twice? (here and when pushing to the queue)

app/Libraries/ClientCheck.php
```diff
@@ -89,6 +109,18 @@ private static function getKey(Build $build): string
             ?? '';
     }
 
+    private static function getMultipartData(Request $request): ?string
+    {
+        $contentType = $request->header('Content-Type', '');
+        $isMultipart = str_contains($contentType, 'multipart/form-data');
```
nanaya: `str_starts_with` is probably sufficient?

config/osu.php
```diff
@@ -118,6 +118,7 @@
         'token_keys' => $clientTokenKeys,
         'token_lifetime' => (get_float(env('CLIENT_TOKEN_LIFETIME_HOUR')) ?? 0.25) * 3600,
         'token_queue' => env('CLIENT_TOKEN_QUEUE') ?? 'token-queue',
+        'token_validation_timeout' => env('CLIENT_TOKEN_VALIDATION_TIMEOUT') ?? 0.05,
```
nanaya, line 121: should be added to the `.env.example`

app/Libraries/ClientCheck.php
TheDomco: changed to ternary
it was intentional, but will actually change the code on the other end, it can be done without encoding twice

PR #12783, opened 2026-02-18
Show username in Contest judging results list

### LINKED ISSUE #12775: Contest judging results page dropdown/breadcrumb should display the user name, not entry name

notbakaneko:
splitting/reordering #12777

Adds the contest entry's username for disambiguating beatmap entries with the same name, and fixes the declared json type being a lie and missing the supposed data

- [x] #12781

closes #12775


app/Http/Controllers/ContestEntriesController.php
```diff
@@ -44,34 +40,14 @@ public function judgeResults($contestId, $id)
...
-        );
-
-        $entryJson = json_item($entry, new ContestEntryTransformer(), $includes);
-
-        foreach ($contest->entries as $entry) {
-            $entry->setRelation('contest', $contest);
-        }
-
-        $entriesJson = json_collection($contest->entries, new ContestEntryTransformer());
+        // entriesByType includes required scores and user data.
```
nanaya: not anymore

resources/views/contest_entries/judge-results.blade.php
```diff
@@ -3,14 +3,21 @@
     See the LICENCE file in the repository root for full licence text.
 --}}
 
-@extends('master')
+@php
+    $title = "{$entryJson['title']} ({$entryJson['user']['username']})";
```
nanaya: I wonder if something like `[username] title` or `username / title` would be easier to read

notbakaneko:
I decided to do the thing I've been wanting to make `SelectOptions` do for a while now...

nanaya:
Seems a bit weird with `-`...?

resources/js/components/select-options.tsx
```diff
@@ -68,15 +68,19 @@ export default class SelectOptions<T extends Option> extends React.Component<Pro
...
             children: (
               <>
-                <div className='u-ellipsis-overflow'>
-                  {this.props.selected?.text}
-                </div>
+                {typeof text === 'string' ? (
+                  <div className='u-ellipsis-overflow'>
+                    {text}
+                  </div>
+                ) : text}
```
nanaya: I think this should also be done inside the `renderOptions`?

this whole render redirection thing is kind of hard to follow...

resources/js/contest-judge-results/header.tsx
```diff
@@ -25,18 +23,17 @@ interface Props {
   entry: ContestEntryJsonForResults;
 }
 
+function entryToOption(entry: ContestEntryJsonForResults) {
+  return {
+    contest_id: entry.contest_id,
+    id: entry.id,
+    text: <ValueDisplay label={entry.title} modifiers='select-option' value={entry.user.username} />,
```
nanaya, line 30: can't this be done inside the `renderOption`? Pass `entry` in the option and render this in addition to `{children}` (with empty text to keep the dropdown chevron)

or maybe it's a bit too much hacks

notbakaneko: I'd rather update the dropdown to have less callbacks and remove the render props instead but that'll be a separate thing since that turns out to be more steps than expected for this PR

nanaya (approved):

PR #12785, opened 2026-02-18
Add dedicated popup menu state

nanaya:
Instead of passing around specific state and functions.

This allows rendering custom button in a more sensible manner.

Theoretically passing a class function or useCallback for the menu may be better than creating a new function every render but that's for another time.


resources/js/components/popup-menu.tsx
```diff
@@ -47,83 +47,95 @@ export default class PopupMenu extends React.PureComponent<PropsWithDefaults, St
     return el == null ? null : $(el);
   }
 
+  private get mobxState() {
+    return this.props.state ?? this.internalMobxState;
```
notbakaneko: just `this.internalMobxState`? In fact, it's `private` and assigned in constructor, so just use it directly?

notbakaneko:
I can't help but think that at this point, if we're passing state objects with public properties and methods in the `render` callback, we might as well skip the whole RenderProp callback thing and just pass in a separate object as props to `PopupMenu`? 🤔

nanaya:
There's a version which passes the object all the time and have the menu rendered as plain child (https://github.com/ppy/osu-web/compare/master...nanaya:popup-menu-state-2?expand=1)

But then I noticed it'll render the menu all the time and adding the check everywhere seems rather clunky. Passing the menu component as props means it also needs to obtain the parents props through other methods.

At any rate those will involve bigger changes and just this one seems like a good small step.

notbakaneko (changes requested):

resources/js/components/popup-menu.tsx
nanaya: it should use the props if passed one

(as in it's not guaranteed to have same props all the time)

nanaya: although now I think again, it probably should never assign received props for the internal fallback state

notbakaneko (approved):

PR #12787, opened 2026-02-19
Fix contest judging page not eager loading correctly

notbakaneko:
- [x] #12781

Also cleaned up some of the contest calling transformer.
...maybe the shuffling should be moved to the transformer too instead of this `defaultJson` function ಠ_ಠ

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1525,22 +1525,22 @@ bn.js@^5.2.1:
   integrity sha512-v2YAxEmKaBLahNwE1mjp4WON6huMNeuDvagFZW+ASCuA/ku0bXR9hSMw0XpiqMoA3+rmnyck/tPRSFQkoC9Cuw==
 
 body-parser@^1.19.0:
-  version "1.20.3"
-  resolved "https://registry.yarnpkg.com/body-parser/-/body-parser-1.20.3.tgz#1953431221c6fb5cd63c4b36d53fab0928e548c6"
-  integrity sha512-7rAxByjUMqQ3/bHJy7D6OGXvx/MMc4IqBn/X0fcM1QUcAItpZrBEYhWGem+tzXH90c+G01ypMcYJBO9Y30203g==
+  version "1.20.4"
+  resolved "https://registry.yarnpkg.com/body-parser/-/body-parser-1.20.4.tgz#f8e20f4d06ca8a50a71ed329c15dccad1cdc547f"
+  integrity sha512-ZTgYYLMOXY9qKU/57FAo8F+HA2dGX7bqGc71txDRC1rS4frdFI5R7NhluHxH6M0YItAP0sHB4uqAOcYKxO6uGA==
   dependencies:
-    bytes "3.1.2"
+    bytes "~3.1.2"
     content-type "~1.0.5"
     debug "2.6.9"
     depd "2.0.0"
-    destroy "1.2.0"
-    http-errors "2.0.0"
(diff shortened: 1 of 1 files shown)


app/Models/ContestEntry.php
```diff
@@ -100,6 +100,12 @@ public function scopeWithScore(Builder $query, Contest $contest): Builder
         return $query->orderBy($orderValue, 'desc');
     }
 
+    public function getTitleAttribute()
+    {
+        // TODO: pass in Content instead?
```
nanaya: Contest? Content? What content?

If anything maybe it should just fail if contest isn't preloaded instead?

app/Transformers/ContestTransformer.php
```diff
@@ -22,6 +25,13 @@ class ContestTransformer extends TransformerAbstract
         'users_voted_count',
     ];
 
+    public ?int $seed = null;
+    public ?string $sort = null;
+
+    public function __construct()
+    {
+    }
```
nanaya: what's this for?

app/Models/Contest.php
```diff
@@ -343,56 +344,51 @@ public function entriesByType(?User $user)
...
-            $contestJson['entries'] = json_collection(
-                $this->entriesByType($user)->loadMissing($preloads),
-                new ContestEntryTransformer(),
-                $includes,
-            );
-        }
-
-        if (!empty($contestJson['entries'])) {
-            if (!$showVotes) {
+            if (!$this->show_votes) {
```
nanaya: `$showVotes`?

app/Http/Controllers/ContestsController.php
```diff
@@ -27,15 +27,19 @@ public function index()
...
             ->findOrFail($id);
 
         abort_if(!$contest->isJudged(), 404);
 
         priv_check('ContestJudgeShow', $contest)->ensureCan();
 
+        foreach ($contest->entries as $entry) {
+            $entry->setRelation('contest', $contest);
+        }
+        $contest->preloadedEntries = $contest->entries;
```
nanaya: I wonder if this is better done in the `entriesByType` instead

app/Models/Contest.php
```diff
@@ -343,56 +344,51 @@ public function entriesByType(?User $user)
...
+                $includes[] = 'entries.results';
+            }
+            if ($this->showEntryUser()) {
+                $includes[] = 'entries.user';
+            }
         }
 
+        $this->preloadedEntries = $this->entriesByType($user)->loadMissing($preloads);
+
         return json_encode([
```
nanaya, line 379: unrelated but this results in `ContestEntries@vote` to return text data type 🤨

app/Transformers/ContestTransformer.php
notbakaneko: I think I was test out passing the params during `new` at one point but ended up having to assign them later anyway 👀

app/Http/Controllers/ContestsController.php
notbakaneko: this doesn't use `entriesByType`? (and also why I moved the assignment out to where it's called)

nanaya: it originally did...? And still does if not set?

app/Models/Contest.php
notbakaneko: this method should be removed and moved to a transformer, if anything (which will happen...later)

PR #12799, opened 2026-02-25
Cleanup some discussions page layout

notbakaneko:
moving the `osu-page` element up instead of repeating it in the components, so I don't have to include it when shuffling things around for the user filter updates.


resources/css/bem/beatmapset-events.less
```diff
@@ -4,6 +4,8 @@
 .beatmapset-events {
   display: flex;
   flex-direction: column;
+  padding: 20px 50px;
```
nanaya: wrong on mobile

resources/css/bem/beatmapset-events.less
```diff
@@ -4,6 +4,12 @@
...
+  background-color: hsl(var(--hsl-b5));
+
+  @media @mobile {
+    padding: 20px 10px;
```
nanaya: looks like this should use `.default-gutter-v2()`? (and separate vertical padding)

PR #12800, opened 2026-02-25
Show replay view count on show score page

### LINKED ISSUE #12788: Add number of replays watched in score details
I think it'd be interesting to add a "replays watched" stat here so others can directly see how many views a score has, when it isn't easily visible on the profile page.

tsunyoku:
Resolves #12788.

image


nanaya:
that doesn't seem like the correct place for it...

tsunyoku:
Where would you put it?

nanaya:
that's a good question...

I'll let others decide

resources/js/scores-show/player.tsx
```diff
@@ -51,6 +51,12 @@ export default function Player(props: Props) {
         <strong>
           {props.score.user.username}
         </strong>
+        <span>
+          {trans('scores.show.player.watched')}
+        </span>
+        <strong>
+          {props.score.replay_views} times
```
nanaya: should be translatable with proper `:count_delimited`

resources/lang/en/scores.php
```diff
@@ -20,6 +20,8 @@
                 'country' => 'Country Rank',
                 'global' => 'Global Rank',
             ],
+
+            'watched' => 'Watched',
```
nanaya: string values are all at the top area

nanaya (changes requested):

nanaya (approved):

Issue #12709, opened 2026-01-25
Add profile pp/rank delta to recent scores


evilldev:
It would be interesting to see how much profile pp or rank a player gained when they set a top score. It would make it easier to keep track of good sessions and show progression better. The top of the recent plays section could potentially even show the total pp and rank gain for every score in the list.

Maybe something like this:

<img width="765" height="121" alt="Image" src="https://github.com/user-attachments/assets/7b6c3947-7506-4dbe-a5c6-2a76d84fd0f8" />

peppy:
Seems like a cool idea.

For this to work we'd need a rotating table (partition based) which stored metadata related to scores. Or we could decide to just store this in the `scores.data` field for cases where it's relevant (maybe the much simpler option).

stanriders:
Storing it permanently would likely be a waste since it would invalidate with every pp update - you won't be able to do anything meaningful with this information anymore

tsunyoku:
I'd be interested in taking this on, but I have some thoughts.

I don't think it should be permanently stored. PP reworks alone would make these values incorrect long term, so my idea would be that this would show specifically under the "Recent scores" view and any metadata would get cleaned up once the scores are too old to display there. If you were to show it on best performances then it will probably give the impression that the data will be there permanently when I don't see that being reasonable. The only way you could make the data permanent would be to try and re-simulate PP gain by running all of their scores in order of date after a rework, but this seems like a lot of compute for little gain. Maybe someone has better ideas.

I'm personally not convinced that `scores.data` is a good place for this given it should be non-permanent in nature. It seems like a good idea to have some sort of `scores_metadata` table (something like `(score_id, pp_delta, rank_delta)`, maybe could even be expanded to contain other metadata in future) and then have a simple clean-up job to clean these rows.

peppy:
I'm not against adding a new table for this kind of purpose, especially if there's other data we could consider moving (or adding) there.

It would use the standard partition cycling with n days of retention.

PR #12801, opened 2026-02-25
Add `score_metadata` table and show PP/rank deltas

tsunyoku:
RFC. Part of #12709.

This creates a new `score_metadata` table which intends to contain PP and rank deltas for each score. This should be short-lived data, and as such I've made it so that these deltas only appear on recent scores.

For this new table, we should use standard partition cycling (with a retention period that ideally matches the recent score scope, so 24h) as suggested [here](https://github.com/ppy/osu-web/issues/12709#issuecomment-3942857362). I cannot find any existing examples of this in migrations so currently it does not do that. Advice on this would be good.

This is not responsible for populating the table, that will need to happen in `osu-queue-score-statistics`. I am holding off on that until people are happy with this PR.

image

image

### DIFF
--- a/app/Models/Solo/Score.php
+++ b/app/Models/Solo/Score.php
@@ -165,7 +165,8 @@ class Score extends Model implements Traits\ReportableInterface
         return $this->morphTo(__FUNCTION__, 'legacy_score_type', 'legacy_best_id');
     }
 
-    public function metadata() {
+    public function metadata()
+    {
         return $this->hasOne(ScoreMetadata::class, 'score_id');
     }
 


nanaya:
how would the down delta work...?

tsunyoku:
So, you technically cannot lose ranks from setting a score as you'll always stay at the same PP or gain. You cannot directly lose PP for setting a score.

However, rank decay exists. There's 2 options here:

1. Keep it simple, rank deltas will include decay by simply calculating delta between their rank before recalculation and their rank after recalculation. This means negative deltas are possible. This would also mean that the rank delta shown on scores would be 1:1 with what they see in-game.
2. Calculate their rank twice - once without the score, and then again with the score. This would give an updated "before" rank that doesn't have decay factored into it so it'd be a more realistic explanation of how your rank _would_ move, but may be confusing  as it won't match what's displayed in-game.

Personally I lean towards the first one because I think matching the delta shown in-game is likely going to make the most sense in most cases, but I can understand why it might be unintuitive to put that directly against a score when the in-game deltas are more about the user.

Just depends which route we want to take.

nanaya:
I wonder if that'd result in the recent section ending up with mostly negatives

I think the actual details will need to be decided on first

tsunyoku:
Anecdotally I'd say it probably won't. The user is most likely to experience decay on their first score of the day, especially if it's been several days since their last, but ranks are not really decaying fast enough for every score after that to result in a loss of ranks. Even if it did, that's feedback they're already receiving on the results screen.

Another option is only displaying the rank delta if the PP delta is > 0, so if they're decaying between scores because they didn't gain PP then it just won't show anything.

I agree on figuring out details first, this is an RFC after all. I don't feel strongly about any particular direction.

PR #12806, opened 2026-02-27
Adjust token validation failures to set score flag correctly

peppy:
Matches
https://github.com/ppy/osu-queue-score-statistics/blob/f07b2be0f845d248ccfda9ed4110570899387d41/osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs#L247-L249.

Alternative would be deleting the scores outright.

### DIFF
--- a/app/Http/Controllers/TeamsController.php
+++ b/app/Http/Controllers/TeamsController.php
@@ -176,15 +176,21 @@ class TeamsController extends Controller
             return ujs_redirect(route('teams.show', compact('team', 'ruleset')));
         }
 
-        $team->loadMissing(prefix_strings('members.user.', UserCompactTransformer::CARD_INCLUDES_PRELOAD));
+        $team->loadMissing(prefix_strings(
+            'members.user.',
+            array_diff(UserCompactTransformer::CARD_INCLUDES_PRELOAD, ['team']),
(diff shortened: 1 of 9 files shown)


app/Http/Controllers/Multiplayer/Rooms/Playlist/ScoresController.php
```diff
@@ -219,7 +219,13 @@ public function update($roomId, $playlistItemId, $tokenId)
 
         $score = $scoreLink->score;
         if ($score->wasRecentlyCreated) {
-            ClientCheck::validateToken($clientTokenData, scoreId: $score->getKey());
+            $validationResult = ClientCheck::validateToken($clientTokenData, scoreId: $score->getKey());
+
+            if ($validationResult !== null && $validationResult !== ClientCheck::SUCCESS) {
```
nanaya, line 224: no need to null check beforehand...

oh nvm

peppy: The idea was that null response means a timeout / unknown state, which we don't want to abort here for. I think it is required for this?

I'm open for writing this in an easier to understand way though.

Issue #12630, opened 2025-12-17
Show the last update time of translated articles


MizuIro-H:
IMO this should be done to help peoples knows that how old what they're reading, instead of just show a hint of "this translate version is outdated".

The elements of this is already shown in codes, idk why they dont show in osu websites.

Walavouchey:
this info is shown in https://osu.wiki/status/fr but i guess it's not linked in enough places

peppy:
I think this would be nice in the official site header.

PR #12809, opened 2026-03-02
Add last edited / outdated timestamp to wiki pages

SisypheOvO:
## Previews

Learn from Wikipedia for placing things at the bottom.

common original articles:
image
up-to-date translations(es):
image
outdated translations(zh Mappers' Guild):
image
*Note that some articles like this can have updated time later than outdated time because of minor changes, typically among all articles or all translations for one article, without removing the `outdated_since` yaml front matter.*

Can be part of issue #12630, but I think the issue doesn't cover all cases (typically articles with different types of notices outlined in `resources\views\wiki\_notice.blade.php`), like stubs, outdated original articles and needsCleanups. Imo all original/translation articles with all types of notices may have these "last edited time" added to clarify it (also I think this is what Wikipedia does)

## How this works

common timestamps are based on the latest commit date. Timestamps for outdated translations are also based on their yaml front matter `outdated_since`.

### DIFF
--- a/app/Libraries/OsuWiki.php
+++ b/app/Libraries/OsuWiki.php
@@ -124,6 +124,19 @@ class OsuWiki
         }
     }
 
+    public static function fetchCommitDate(string $hash): ?string
+    {
+        try {
+            $commit = GitHub::repo()
+                ->commits()
+                ->show(static::user(), static::repository(), $hash);
+
+            return $commit['commit']['committer']['date'] ?? null;
+        } catch (GithubException $e) {
+            return null;
+        }
+    }
+
     public static function isImage($path)
     {
         return preg_match('/\.(?:jpe?g|gif|png)$/i', $path) === 1;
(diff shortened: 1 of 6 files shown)


nanaya:
we've been occasionally hitting github hit rate limit and this wouldn't help. there's also plan on serving a wiki checkout somewhere else as static files so to not use github at all and this wouldn't work either...

SisypheOvO:
> we've been occasionally hitting github hit rate limit and this wouldn't help. there's also plan on serving a wiki checkout somewhere else as static files so to not use github at all and this wouldn't work either...

I think yes, this will make it worse for the first problem which is sad because I didn't know that :( , but still works whether or not the second plan is implemented? The new system may still be based on version control (I suppose, maybe still github) and/so an update timestamp for each page is still wanted for many. And you can't avoid version control both for managing purpose and also for readers' experience.

What I agree is that maybe we do need a better thing (for example yaml headers) to track updated time because tiny changes that doesn't count for an "update" are still regarded as the latest updates for translation articles now.

notbakaneko:
yeah, I don't see this happening unless we can get the update times in the same request as the content.

PR #12811, opened 2026-03-02
Work around anchor links not correctly detected by turbo

### LINKED ISSUE #12807: Clicking on the table of contents in the wiki reloads the page
Clicking a link in any wiki page's table of contents causes a full page reload instead of scrolling to the corresponding section.

This occurs in all browsers: Firefox, Chromium and Epiphany.

nanaya:
Resolves #12807

<sub>I wouldn't be surprised if this causes some different bugs</sub>


notbakaneko (approved):
seems to work™️

PR #12812, opened 2026-03-02
Check ranking blocker issue across all difficulties

### LINKED ISSUE #12808: Pending issues warning on qualified maps isn't always visible
If a suggestion or problem is posted in a qualified map's discussion page, it is blocked from getting ranked until they are resolved, and there is a warning/message stating this. However, this is only visible if the difficulty where it was posted is selected, or if it's in the general (all difficulties) tab. Therefore it's hard to notice sometimes unless you open the difficulty dropdown menu where a "1" appears next to the difficulty in question.

For example, on [this map](https://osu.ppy.sh/beatmapsets/2381078/discussion) it only appears if you select Alumetri's Expert, but not on any other difficulty as shown below:
Image
Image

nanaya:
Resolves #12808

### DIFF
--- a/resources/js/beatmap-discussions/discussions-state.ts
+++ b/resources/js/beatmap-discussions/discussions-state.ts
@@ -277,6 +277,11 @@ export default class DiscussionsState {
   }
 
   @computed
+  get firstUnresolvedDiscussion() {
+    return this.nonDeletedDiscussions.find((discussion) => discussion.can_be_resolved && !discussion.resolved);
+  }
+
+  @computed
   get nominators() {
     const nominators: UserJson[] = [];
     for (let i = this.beatmapset.events.length - 1; i >= 0; i--) {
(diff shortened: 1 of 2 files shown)


resources/js/beatmap-discussions/discussions-state.ts
```diff
@@ -276,6 +276,11 @@ export default class DiscussionsState {
     return new Set(this.beatmapset.eligible_main_rulesets);
   }
 
+  @computed
+  get firstUnresolvedDiscussion() {
+    return this.nonDeletedDiscussions.find((discussion) => discussion.can_be_resolved && discussion.resolved);
```
notbakaneko:
```suggestion
    return this.nonDeletedDiscussions.find((discussion) => discussion.can_be_resolved && !discussion.resolved);
```

notbakaneko (changes requested):

notbakaneko (approved):

PR #12815, opened 2026-03-03
Skip priv check for batch message interop

nanaya:
Related to ppy/osu#36788

Need to make sure all the usage of batch chat interop have their own check already.

### DIFF
--- a/app/Http/Controllers/LegacyInterOpController.php
+++ b/app/Http/Controllers/LegacyInterOpController.php
@@ -256,7 +256,8 @@ class LegacyInterOpController extends Controller
                         $sender,
                         $pmTarget,
                         presence($messageParams['message'] ?? null),
-                        $messageParams['is_action'] ?? null
+                        $messageParams['is_action'] ?? null,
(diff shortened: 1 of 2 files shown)


notbakaneko:
can it not use `REQUEST_IS_INTEROP_KEY`?

nanaya:
it's easier to limit (or in this case allow) specific endpoint than checking all existing path and usage

app/Http/Controllers/LegacyInterOpController.php
```diff
@@ -269,6 +270,7 @@ public function userBatchSendMessage()
                         $channel,
                         presence($messageParams['message'] ?? null),
                         $messageParams['is_action'] ?? false
+                        false,
```
notbakaneko: I think a comma is missing...

notbakaneko (changes requested):

notbakaneko (approved):
Seems fine assuming the caller is checking all the other things it needs

PR #12819, opened 2026-03-04
Skip resolved check on deleted beatmaps

nanaya:
Another whoops.

### DIFF
--- a/resources/js/beatmap-discussions/discussions-state.ts
+++ b/resources/js/beatmap-discussions/discussions-state.ts
@@ -278,7 +278,11 @@ export default class DiscussionsState {
 
   @computed
   get firstUnresolvedDiscussion() {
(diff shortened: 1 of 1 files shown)


resources/js/beatmap-discussions/discussions-state.ts
```diff
@@ -278,7 +278,11 @@ export default class DiscussionsState {
 
   @computed
   get firstUnresolvedDiscussion() {
-    return this.nonDeletedDiscussions.find((discussion) => discussion.can_be_resolved && !discussion.resolved);
+    return this.nonDeletedDiscussions.find((discussion) => (
```
nanaya, line 281: I wonder if this should just exclude discussion from deleted maps in the first place

notbakaneko (approved):

PR #12820, opened 2026-03-04
Add helper for getting storage with correct type

nanaya:
Adding the prefix every time is getting old 🚮


notbakaneko (approved):
I guess it's fine it doesn't handle the unsuffixed ones since that's the default...

PR #12821, opened 2026-03-04
Add function to regenerate audio preview

nanaya:
Only accessible through ad hoc tinker etc.

depends on
- [x] #12820
- [x] #12814

for testing. drop osz files on top directory, create `public/test` directory, and then include the file in a tinker session.
```php
<?php

$t = time();
$html = 'public/test/index.html';
file_put_contents($html, '<!doctype html>');

foreach (glob('./*.osz') as $f) {
    $oszName = substr($f, 2);
    $bid = intval($oszName);
    $path = "{$t}-{$bid}.ogg";
    file_put_contents(
        "public/test/{$path}",
        (new BeatmapsetArchive($f, null))->generateAudioPreview(),
    );
    file_put_contents(
        $html,
        "<p>{$oszName}<br><audio src='{$path}' controls></audio></p>",
        FILE_APPEND,
    );
}

file_put_contents($html, '<script src="index.js"></script>', FILE_APPEND);
```

`index.js`
```javascript
const players = document.querySelectorAll('audio');

for (let i = 0; i < players.length; i++) {
  const player = players[i];
  const nextPlayer = players[(i + 1) % players.length];
  player.addEventListener('ended', () => nextPlayer.play());
}
```

### DIFF
--- a/app/Models/Beatmapset.php
+++ b/app/Models/Beatmapset.php
@@ -6,7 +6,6 @@
 namespace App\Models;
 
 use App\Enums\Ruleset;
-use App\Exceptions\BeatmapProcessorException;
 use App\Exceptions\ImageProcessorServiceException;
 use App\Exceptions\InvariantException;
 use App\Jobs\CheckBeatmapsetCovers;
@@ -454,6 +453,11 @@ class Beatmapset extends Model implements AfterCommit, Commentable, Indexable, T
         return $urls;
     }
 
+    public function archive(): ?BeatmapsetArchive
+    {
+        return $this->memoize(__FUNCTION__, fn () => BeatmapsetArchive::fetch($this));
+    }
+
     public function coverURL($coverSize = 'cover', $customTimestamp = null)
     {
         $timestamp = $customTimestamp ?? $this->defaultCoverTimestamp();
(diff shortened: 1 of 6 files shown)


notbakaneko:
I noticed the generated files are `ogg` but the preview urls are hardcoded to `mp3`..?

nanaya:
to keep the existing url 🙃

notbakaneko:
nvm, I was looking at something else

app/Models/Beatmapset.php
```diff
@@ -597,6 +562,21 @@ public function regenerateCovers(?array $sizesToRegenerate = null)
...
+        $preview = $this->archive()->generateAudioPreview();
+
+        if ($preview === null) {
+            return false;
+        }
+
+        return storage_disk('beatmapset')->put(
+            "preview/{$this->getKey()}.mp3",
+            $preview,
+            ['Content-Type' => 'audio/ogg'],
```
nanaya, line 576: I'm not actually sure this does anything at all

app/Models/BeatmapsetArchive.php
```diff
@@ -25,6 +27,91 @@ public function __construct(string $osz)
...
+        // archive file is gone, nothing to do for now
+        if ($statusCode === 302) {
+            return null;
+        }
+        if ($statusCode !== 200) {
+            throw new BeatmapProcessorException('Failed downloading osz: HTTP Error '.$statusCode);
+        }
+
+        try {
+            return new BeatmapsetArchive(get_stream_filename($oszFile), $beatmapset);
```
notbakaneko: `new static`? or else the return type should be `?self`

notbakaneko (changes requested):

notbakaneko (approved):

app/Models/Beatmapset.php
```diff
@@ -597,6 +562,21 @@ public function regenerateCovers(?array $sizesToRegenerate = null)
         $this->update(['cover_updated_at' => $this->freshTimestamp()]);
     }
 
+    public function regenerateAudioPreview(): mixed
```
notbakaneko: forgot to mention, does this return anything else besides `bool`?

app/Models/BeatmapsetArchive.php
```diff
@@ -69,33 +160,52 @@ public function hasFile(?string $filename)
         return $this->zip->locateName($filename, \ZipArchive::FL_NOCASE) !== false;
     }
 
-    // Parses given list (of .osu files) and finds background images referenced.
-    // If $performFallback is enabled, all .osu files in archive are scanned if $filelist yields no result.
-    // This allows beatmap order to dictate the priority (to match existing behaviour).
-    public function scanBeatmapsForBackground(array $filelist, bool $performFallback = false)
+    public function generateAudioPreview(): ?string
```
notbakaneko, line 165: missing `return null` if it falls through (and maybe declare `strict_types`)

notbakaneko (changes requested):

app/Models/Beatmapset.php
nanaya: oh huh the contract is bool 🤔

notbakaneko (approved):

PR #12832, opened 2026-03-10
Use different rank style for legacy scores

notbakaneko:
https://discord.com/channels/90072389919997952/1458132886946451497/1480765862016319589

The original icons are all different sizes, so I squished them down to the same height to keep the sizing and alignment simple. (The `S`s in `SS` and `S` are actually supposed to be the same size making `SS` slightly taller....)

### DIFF
--- a/app/Models/BeatmapsetArchive.php
+++ b/app/Models/BeatmapsetArchive.php
@@ -139,7 +139,8 @@ class BeatmapsetArchive
     {
         return $this->osuFileList ??= array_values(array_unique([
             // use db order
-            ...($this->beatmapset?->beatmaps->pluck('filename') ?? []),
+            // filename column in beatmaps table is nullable
+            ...array_reject_null($this->beatmapset?->beatmaps->pluck('filename') ?? []),
             ...preg_grep('/\.osu$/i', $this->fileList()),
         ]));
     }
--- a/app/Models/Multiplayer/Room.php
+++ b/app/Models/Multiplayer/Room.php
@@ -647,6 +647,10 @@ class Room extends Model
             'auto_skip:bool',
         ], ['null_missing' => true]);
 
+        if ($params['name'] === null) {
+            throw new InvariantException("field 'name' is required");
+        }
+
         $this->fill([
             'max_attempts' => $params['max_attempts'],
             'name' => app('chat-filters')->filter($params['name']),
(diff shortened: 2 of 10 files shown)


resources/js/scores-show/dial.tsx
```diff
@@ -29,37 +31,43 @@ export default function Dial(props: Props) {
...
-                d={arc({ innerRadius: 75, outerRadius: 100, ...d }) ?? undefined}
-              />
-            ))}
-          </g>
-        </svg>
-      </div>
+    <div className={classWithModifiers('score-dial')}>
+      {props.legacy ? (
+        <div className={classWithModifiers('score-dial__legacy-rank', props.rank)} />
+      ) : (
```
nanaya: might as well create a new component?

webpack.config.js
```diff
@@ -199,6 +199,7 @@ const rules = [
 const resolve = {
   alias: {
     '@fonts': path.resolve(__dirname, 'resources/fonts'),
+    '@image_resources': path.resolve(__dirname, 'resources/images'),
```
notbakaneko, line 202: this is so there aren't 2 copies of the same asset; we should probably separate the ones that no longer need to be in `public` at some point...

resources/images/scores/legacy-ranking-A.png
nanaya, line 1: maybe in a `ranking/` directory? And `legacy` somewhere in the path/filename.

resources/css/bem/legacy-rank.less
```diff
@@ -0,0 +1,28 @@
...
+  background-size: contain;
+  background-repeat: no-repeat;
+  background-position: center;
+  // makes the lettering about 150px tall accounting for the whitespace.
+  height: 160px;
+  // tak up the same space as score-dial
+  width: 200px;
+  margin: 20px 0;
+
+  .rank(~"XH");
```
nanaya: `each()` is now a thing in less

resources/js/scores-show/legacy-rank.tsx
```diff
@@ -0,0 +1,14 @@
...
+import Rank from 'interfaces/rank';
+import React from 'react';
+import { classWithModifiers } from 'utils/css';
+
+interface Props {
+  rank: Rank;
+}
+
+export default function LegacyRank({ rank }: Props) {
+  return <div className={classWithModifiers('legacy-rank', rank)} />;
```
nanaya: might as well inline? 🤨

nanaya (changes requested):

resources/js/scores-show/legacy-rank.tsx
notbakaneko: not sure if `const a = () => ...; export default const a;` is any better here...

nanaya: no I mean inline in where it's used instead of additional component since it's literally just single div with specific class

PR #12835, opened 2026-03-11
Keep showing user card menu button when activated

nanaya:
(and just in general "active" user card state)

It's been bothering me a bit.

Converting the states to mobx may also be a good idea? Or at least updating some of the stuff in there to use the existing mobx state (mainly the current user related stuff).


notbakaneko (changes requested):
I think the react state and `PureComponent` should be removed while we're at it; I recall things becoming silly due to them having different update and rerender cycles and having different snapshots of what they're actually rendering...

nanaya:
Check the #12837 for the state removal. As for Component/PureComponent, I've been seeing conflicting recommendation with the PureComponent one being [the one recommanded in readme for 7.6.0](https://github.com/mobxjs/mobx/tree/mobx-react%407.6.0/packages/mobx-react#class-components) (with the example still showing Component).

(apparently the recommendation is PureComponent since mobx-react 6.0.0)

notbakaneko (approved):

PR #12839, opened 2026-03-11
Make avatar cache purge function reusable

nanaya:
I'm just guessing how it works 🤷 Seems to work for both avatar and beatmap audio preview.

- [x] copy/rename `AVATAR_CACHE_PURGE_AUTHORIZATION_KEY` env to `CACHE_PROXY_PURGE_AUTHORIZATION_KEY`


ThePooN:
I think we cab get rid of the configurable method as well. The `GET` code is also obsolete. This was all temporary glue while we were switching infra.

`s3-nginx-proxy` always expects `DELETE`.

notbakaneko (approved):

ThePooN (approved):

PR #12843, opened 2026-03-12
Normalise audio preview volume

nanaya:
Might as well. -14 lufs as seemingly commonly used for internet stuff.


notbakaneko (approved):
looks (sounds) correct...

PR #12844, opened 2026-03-12
Add mapping for ruleset name to id and adjust things

nanaya:
Mainly so I can use `rulesetIds.osu` etc in https://github.com/nanaya/osu-web/blob/003a6f76fc2621474f35ac1f055e24061f33f4e9/resources/js/utils/score-helper.ts#L158


resources/js/interfaces/ruleset.ts
```diff
@@ -1,16 +1,20 @@
...
+  taiko: 1,
 } as const;
-export const rulesets = Object.values(rulesetIdToName);
+
+export const rulesetNames = invert(rulesetIds) as Record<RulesetId, Ruleset>;
+export const rulesets = Object.values(rulesetNames);
 
 export function ensureRuleset(maybeRuleset: string): Ruleset | undefined {
-  if ((rulesets as readonly string[]).includes(maybeRuleset)) {
+  if (rulesetIds[maybeRuleset as Ruleset] != null) {
```
notbakaneko: `maybeRuleset in rulesetIds`

could also make it a type guard and accept `unknown` and remove the explicit casting, but I don't think it's needed anywhere else...yet

resources/js/interfaces/ruleset.ts
```diff
@@ -20,13 +24,13 @@ export function ensureRulesetId(maybeRulesetId: number | string): RulesetId | un
     maybeRulesetId = parseInt(maybeRulesetId, 10);
   }
 
-  if (rulesetIdToName[maybeRulesetId as RulesetId] != null) {
+  if (rulesetNames[maybeRulesetId as RulesetId] != null) {
```
notbakaneko: `maybeRulesetId in rulesetNames`

notbakaneko (changes requested):

notbakaneko (approved):

PR #12846, opened 2026-03-13
Delete preview of maps with download disabled

nanaya:
As they should be.


notbakaneko:
so I noticed the previews for download disabled maps are still enabled? 🤔

nanaya:
yeah will be done separately

notbakaneko (approved):

PR #12847, opened 2026-03-13
Limit displayed accuracy meter to the achieved score rank

### LINKED ISSUE #12840: Circular accuracy meter doesn't capping at `A` grade limit when there is misses but acc is higher than `95.00`
As above.

Image

Lazer:

Image

nanaya:
Resolves #12840.

- [x] depends on #12844


resources/js/scores-show/dial.tsx
```diff
@@ -23,10 +19,29 @@ const displayRank: Record<Rank, string> = {
   XH: 'SS',
 };
 
-export default function Dial(props: Props) {
+// the index for maximum accuracy of the rank
+const cutoffIndex: Record<Rank, number> = {
```
notbakaneko: I think it makes more sense to declare these in the same place that makes the array?

notbakaneko (approved):

PR #12848, opened 2026-03-13
Regenerate audio preview as part of beatmapset processing

### LINKED ISSUE #2748: Map audio preview needs some improvements
https://osu.ppy.sh/beatmapsets/572387/#osu/1255280 vs https://osu.ppy.sh/beatmapsets/333426#fruits/737862 

https://osu.ppy.sh/beatmapsets/694107/#fruits/1468584 vs https://osu.ppy.sh/beatmapsets/458825/#fruits/983060

The first links are late, the second links are at the correct time. They all have the same preview points in game.
Probably a client-side issue

Was a thing on the old site too

Summary:
- broken preview (#10331, #11041)
- low quality (#6459)
- timing (this, #10775)

nanaya:
Mostly lifted from the job it replaced but with less logging (can be readded if actually needed).

Resolves #2748

### DIFF
--- a/app/Http/Controllers/LegacyInterOpController.php
+++ b/app/Http/Controllers/LegacyInterOpController.php
@@ -8,7 +8,7 @@ namespace App\Http\Controllers;
 use App\Exceptions\Handler as ExceptionHandler;
 use App\Jobs\EsDocument;
 use App\Jobs\Notifications\ForumTopicReply;
-use App\Jobs\RegenerateBeatmapsetCover;
+use App\Jobs\RegenerateBeatmapsetMedia;
 use App\Libraries\Chat;
 use App\Models\Beatmap;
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/LegacyInterOpController.php
notbakaneko, line 65: it occurs to me that because `regenerateCovers` calls `update`, it already queues a beatmapset index job (and also for removing), but if they're called by a running job or there's existing jobs in the queue, the unique constraint will handle the duplicates; otherwise, if they run immediately, it indexes multiple times 🤔 

(probably not so much an issue if there's always something in queue, and if the queue is empty... well, the queue doesn't have anything to do anyway 👀 )

notbakaneko (approved):

PR #12851, opened 2026-03-14
Fix incorrect check on beatmapset search query update

### LINKED ISSUE #12850: Beatmap search input does not allow backspace deletion of certain strings.
On the search beatmap page, if the user types in the input a string that contains "- " (hiphen followed by space) anywhere in it, it is not possible to erase that part of the string using backspace. It can be deleted by selecting manually via ctrl + a or highlighting with the mouse, and text after this pattern can be erased normally with backspace, but when it's time to delete "- " it just doesn't.

Image

nanaya:
`update` is already the entrypoint for updating any filter attributes so intercepting the `query` field doesn't really mean much - it still can't be updated directly regardless.

There's also another problem of debounce for query change stopped working because it monitors all fields.

Resolves #12850

### DIFF
--- a/resources/js/beatmapset-search-filters.ts
+++ b/resources/js/beatmapset-search-filters.ts
@@ -92,12 +92,11 @@ export class BeatmapsetSearchFilters {
 
   constructor(url: string) {
     const filters = filtersFromUrl(url);
+    this.queryRaw = filters.query ?? '';
     for (const key of keyNames) {
       const value = filters[key] ?? null;
-      this[key] = value === this.getDefault(key) ? null : value;
+      this[key] = this.sanitiseValue(key, value);
     }
-
-    this.queryRaw = this.query ?? '';
 
     makeObservable(this);
   }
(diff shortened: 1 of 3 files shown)


resources/js/beatmapset-search-filters.ts
```diff
@@ -92,21 +92,13 @@ export class BeatmapsetSearchFilters {
 
   constructor(url: string) {
     const filters = filtersFromUrl(url);
+    this.queryRaw = filters.query ?? '';
     for (const key of keyNames) {
       const value = filters[key] ?? null;
-      this[key] = value === this.getDefault(key) ? null : value;
+      this[key] = this.sanitiseValue(key, value);
```
nanaya, line 98: this here (previous and current) has only been accidentally working correctly because `sort` comes after `query`...

LiquidPL (approved):
yea, it's working now :sweat_smile:

notbakaneko (approved):

PR #12852, opened 2026-03-14
Port lazer's user tag filtering behaviour

### LINKED ISSUE #12849: Improve matching tags to match lazer behaviour
in lazer you can type: `sk ju` to get skillset jumps, but cant on website 

https://github.com/user-attachments/assets/c64daadd-9cff-4596-9f7c-652167597b40

https://github.com/user-attachments/assets/52d1f5bb-a322-40c5-a3b1-bc4821f84fa5

LiquidPL:
Resolves #12849.

This makes the tag filtering behaviour match the one in [lazer](https://github.com/ppy/osu/blob/39912e044a05480d7587b6c8d9dfd78f3f99dd0f/osu.Game/Screens/Ranking/UserTagControl.AddTagsPopover.cs#L249)/[framework](https://github.com/ppy/osu-framework/blob/2140bf7dae73a6389ffea01f81319a39392c8ef1/osu.Framework/Graphics/Containers/SearchContainer.cs#L162) - that is, each word in the query is matched separately against both the name and description.

### DIFF
--- a/app/Http/Controllers/LegacyInterOpController.php
+++ b/app/Http/Controllers/LegacyInterOpController.php
@@ -8,7 +8,7 @@ namespace App\Http\Controllers;
 use App\Exceptions\Handler as ExceptionHandler;
 use App\Jobs\EsDocument;
 use App\Jobs\Notifications\ForumTopicReply;
-use App\Jobs\RegenerateBeatmapsetCover;
+use App\Jobs\RegenerateBeatmapsetMedia;
 use App\Libraries\Chat;
 use App\Models\Beatmap;
 use App\Models\Beatmapset;
@@ -58,7 +58,7 @@ class LegacyInterOpController extends Controller
         $beatmapset = Beatmapset::withTrashed()->findOrFail($id);
 
         if (!$beatmapset->trashed()) {
-            $job = (new RegenerateBeatmapsetCover($beatmapset))->onQueue('beatmap_default');
+            $job = (new RegenerateBeatmapsetMedia($beatmapset))->onQueue('beatmap_default');
             $this->dispatch($job);
         }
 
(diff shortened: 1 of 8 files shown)


resources/js/models/beatmap-tag.ts
```diff
@@ -36,8 +38,16 @@ export default class BeatmapTag {
     this.tagName = split[1];
   }
 
-  matchesName(match: string) {
-    return this.nameLowercase.includes(match.toLowerCase());
+  match(match: string) {
+    const split = match.toLowerCase().split(/\s+/);
```
nanaya: should probably done by the caller. And might as well do the ruleset check here too.

and can use a `trim`

resources/js/beatmaps/user-tag-picker-controller.ts
```diff
@@ -34,15 +34,10 @@ export default class UserTagPickerController {
 
   @computed
   get groups() {
-    const rulesetId = this.rulesetId;
+    const filtered = this.tags.filter((tag) =>
+      tag.match(this.query.trim().split(/\s+/), this.rulesetId));
```
nanaya: store the trim/split result in a variable first...

also the toLowerCase is missing

resources/js/beatmaps/user-tag-picker-controller.ts
```diff
@@ -34,15 +34,12 @@ export default class UserTagPickerController {
 
   @computed
   get groups() {
-    const rulesetId = this.rulesetId;
+    const filtered = this.tags.filter((tag) => {
+      const querySplit = this.query.trim().split(/\s+/);
```
nanaya: there's no point if it's still run hundred of times. put this out of the loop. the same with lowercase thing; do it before the split

PR #12857, opened 2026-03-17
Don't reset guest user preferences on load

### LINKED ISSUE #12856: Preview volume resets after opening a new beatmap tab
Firefox 148.0

https://github.com/user-attachments/assets/5e6ba7dc-d33c-485c-a43e-133d77fd507c

nanaya:
Resolves #12856.

### DIFF
--- a/resources/js/core/user/user-preferences.ts
+++ b/resources/js/core/user/user-preferences.ts
@@ -54,7 +54,7 @@ export default class UserPreferences {
   setUser(user?: CurrentUserJson) {
     this.user = user;
 
-    if (!this.updatingOptions) {
+    if (this.user != null && !this.updatingOptions) {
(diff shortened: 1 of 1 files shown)


resources/js/core/user/user-preferences.ts
```diff
@@ -54,7 +54,7 @@ export default class UserPreferences {
   setUser(user?: CurrentUserJson) {
     this.user = user;
 
-    if (!this.updatingOptions) {
+    if (this.user != null && !this.updatingOptions) {
```
notbakaneko: can use `user != null` and remove the `?.` on the next line

notbakaneko (changes requested):

notbakaneko (approved):

nanaya:
wait, is the preference field even nullable 🤔

PR #12859, opened 2026-03-18
Set preview time at 40% if unset

nanaya:
Missed when looking at the old logic.


peppy:
Reference: https://github.com/peppy/osu-web-10/blob/5e900db7a889014fbf7302c7adea5d9d5ce9c5ed/www/web/mp3-osz2.php#L109

notbakaneko (approved):

PR #12868, opened 2026-03-23
Show join public channel page if not currently in any channels

nanaya:
Joining public channel has been added so use it instead of showing a page of nothingness.


notbakaneko (approved):
makes sense

PR #12869, opened 2026-03-23
Include cover and team in playlists leaderboard API response

LiquidPL:
I'm working on updating the leaderboard on the playlists screen to match the new ones on song select/daily challenge (see https://discord.com/channels/90072389919997952/1458132886946451497/1485771051429527644). This commit updates the endpoint for said leaderboard to return user teams and covers, which were previously missing.


app/Http/Controllers/Multiplayer/RoomsController.php
```diff
@@ -221,12 +221,12 @@ public function leaderboard($roomId)
             'leaderboard' => json_collection(
                 $room->topScores()->paginate($limit),
                 'Multiplayer\UserScoreAggregate',
-                ['user.country']
+                ['user.country', 'user.cover', 'user.team']
```
LiquidPL, line 224: Would it be okay to use `ScoreTransformer::MULTIPLAYER_BASE_INCLUDES` here even though it's a different transformer? (same below)

notbakaneko: no, they're different transformers whose includes happen to have the same name.
It might make sense if there was an interface everything implementing the includes have to implement but that's not what we're doing.

notbakaneko (approved):

Issue #11666, opened 2024-11-19
Sorting by multiple people on Beatmap's Discussion page


Toayhel:
Hi! Found out how to use GitHub, so I'm basically reposting what I wrote here: https://osu.ppy.sh/community/forums/topics/2004400?n=1

I'd like to suggest a QOL idea of adding a feature that improves sorting on Beatmap's Discussion page.

There are maps that had been modded by a lot of people and when you're a BN it might be hard (more like annoying) to check stuff like

https://toayhel.s-ul.eu/1agllXIX or https://toayhel.s-ul.eu/dCFt38FI

Instead of looking for your own or other BN's mods, you can use "sort by user" thing. However, when you want to nominate a beatmap, you need to do a lot of indepth checks of a beatmap, so it could get ranked without any problems. After a map is modded by both BNs and mods are applied, you need to recheck it and see if your suggestions/problem stamps are properly fixed.

It's not mandatory, but checking other BN's mods wouldn't hurt because you may find something that you didn't notice yourself, but mapper rejected it and you agree with the mod. It would make rechecking a little easier since you don't have to double check the map:

1. Rechecking your mods first and going through the set
2. Rechecking other BN's mod and going through the set again
Instead you could just sort by two or more users, so you could recheck everything at once.

Some BNs expressed their support to this idea on BN Server (may check out the replies below: https://discord.com/channels/316154420591067136/316154420591067136/1307673613390053520), in DMs (https://toayhel.s-ul.eu/XPzk1lcC) and promoted with priority stars on osu! forum I've linked in the first line. That's why it would be amazing if this could be added. I believe that this may help reducing the amount of disqualifications and unranks since rechecking will be a lot easier and take less time (since it may be kinda tedious to recheck a map you've spent a lot of time to mod on)

notbakaneko:
Do you mean being able to select multiple users when filtering?

Toayhel:
Yes, so people could choose not only themselves here, but other specific people too (sort by me and Spkz in this case) 

![image](https://github.com/user-attachments/assets/ccaa144e-5395-4a53-9e18-0fcb48b92a84)

PR #12870, opened 2026-03-24
Move discussion sort state into discussions state

notbakaneko:
So that the sort toggle doesn't have to be in the same component.

I put off PRing this in case I had to make more changes for #11666 but didn't end up making any 🤷

### DIFF
--- a/resources/js/beatmap-discussions/discussions.tsx
+++ b/resources/js/beatmap-discussions/discussions.tsx
@@ -4,52 +4,17 @@
 import IconExpand from 'components/icon-expand';
 import BeatmapsetDiscussionJson, { BeatmapsetDiscussionJsonForShow } from 'interfaces/beatmapset-discussion-json';
 import BeatmapsetDiscussionsStore from 'interfaces/beatmapset-discussions-store';
-import { action, computed, makeObservable, observable } from 'mobx';
+import { action, computed, makeObservable } from 'mobx';
 import { observer } from 'mobx-react';
 import * as React from 'react';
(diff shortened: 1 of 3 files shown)


resources/js/beatmap-discussions/discussions.tsx
```diff
@@ -4,81 +4,32 @@
...
-  }
-
   private get discussionsState() {
     return this.props.discussionsState;
   }
 
   @computed
   private get isTimelineVisible() {
-    return this.discussionsState.currentPage === 'timeline' && this.currentSort === 'timeline';
+    return this.discussionsState.currentPage === 'timeline' && this.discussionsState.currentSort === 'timeline';
```
nanaya, line 32: unrelated but the timeline sort is only possible in timeline page isn't it

resources/js/beatmap-discussions/sort.ts
```diff
@@ -0,0 +1,42 @@
...
+export const sortPresets = {
+  created_at: {
+    sort(a: BeatmapsetDiscussionJson, b: BeatmapsetDiscussionJson) {
+      return a.created_at === b.created_at
+        ? a.id - b.id
+        : Date.parse(a.created_at) - Date.parse(b.created_at);
+    },
+    text: trans('beatmaps.discussions.sort.created_at'),
+  },
+  // there's obviously no timeline field
```
nanaya, line 16: this comment isn't useful (anymore?) since there's no updated_at field either

nanaya (approved):

PR #12871, opened 2026-03-24
Attempt to type notification details

nanaya:
Just dump everything into one interface for now.

The camelcase thing is mildly annoying so just drop the json as is.


notbakaneko (changes requested):
conflicts

notbakaneko (approved):

PR #12872, opened 2026-03-25
Fix user cards re-rendering too often

notbakaneko:
seeing https://github.com/ppy/osu-web/pull/12837/changes#diff-beae6c5ff40beb4d5425b33c73d1a7ed16115fe27e3fbddc8e37e5d653805412R122 reminded me of the problem of unwanted re-renders triggering; in this case, `mergeModifiers` in the parent is always returning a new array, so the `props` are always changing...

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -55,7 +55,7 @@ jobs:
       - name: Install js dependencies
         run: yarn --frozen-lockfile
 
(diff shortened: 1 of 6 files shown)


nanaya:
I wonder if the modifiers function should split string instead and the merge modifiers function returns string instead of array.

notbakaneko:
looking at #12874 first

notbakaneko:
#12874 is probably a better idea instead of having to work around the props every time

PR #12873, opened 2026-03-25
Use observables instead of setState for activated popup key context

notbakaneko:
The context is still used to avoid passing state as props for the sake of passing them further down the chain, but maybe we can reduce it to needing a single context instead of 2?


nanaya:
how would it work with single context? 🤔

resources/js/components/user-cards.tsx
```diff
@@ -46,4 +68,9 @@ export class UserCards extends React.PureComponent<Props> {
       </ContainerContext.Provider>
     );
   }
+
+  @action
+  private readonly activeKeyDidChange = (key: number | null) => {
```
nanaya: I wonder if it's simpler to have the `ContainerContext` value to be an observable like `{ key: 1 }` instead

notbakaneko:
single context was not a great idea since it was just creating new objects to pass the existing reference with the key... so a context 🙈

nanaya:
I think this one is still pretty useful

notbakaneko:
going to rebase this one off the string modifiers later

resources/js/components/popup-menu-persistent.tsx
```diff
@@ -13,9 +13,9 @@ export function PopupMenuPersistent(props: Props) {
   const container = React.useContext(ContainerContext);
   const key = React.useContext(KeyContext);
 
-  const onHide = React.useCallback(() => container.activeKeyDidChange(null), [container]);
-  const onShow = React.useCallback(() => container.activeKeyDidChange(key), [container, key]);
-  React.useEffect(() => () => container.activeKeyDidChange(null), [container]);
+  const onHide = React.useCallback(() => container.setValue(null), [container]);
+  const onShow = React.useCallback(() => () => container.setValue(key), [container, key]);
```
nanaya: one too many `() =>`

nanaya (changes requested):

nanaya (approved):

PR #12874, opened 2026-03-25
Only allow string for Modifiers prop

nanaya:
Maybe a good idea? 🤷

### DIFF
--- a/resources/js/notification-widget/item.tsx
+++ b/resources/js/notification-widget/item.tsx
@@ -8,7 +8,7 @@ import NotificationDeleteButton from 'notifications/notification-delete-button';
 import NotificationReadButton from 'notifications/notification-read-button';
 import { NotificationContext } from 'notifications-context';
 import * as React from 'react';
-import { classWithModifiers, mergeModifiers, urlPresence } from 'utils/css';
+import { classWithModifiers, mergeModifiers, Modifiers, urlPresence } from 'utils/css';
 import { trans } from 'utils/lang';
 
 interface Props {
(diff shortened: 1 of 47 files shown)


notbakaneko:
I guess it could work 🤔

notbakaneko (changes requested):
there are a few `modifiers?: string[]` that could use an update as well

resources/js/notification-widget/item.tsx
```diff
@@ -21,7 +21,7 @@ interface Props {
   item: Notification;
   markRead?: () => void;
   message: string;
-  modifiers: string[];
+  modifiers: Modifiers;
```
notbakaneko:
```suggestion
  modifiers?: Modifiers;
```

resources/js/notifications/notification-read-button.tsx
```diff
@@ -4,11 +4,11 @@
 import { Spinner } from 'components/spinner';
 import { observer } from 'mobx-react';
 import * as React from 'react';
-import { classWithModifiers } from 'utils/css';
+import { classWithModifiers, Modifiers } from 'utils/css';
 
 interface Props {
   isMarkingAsRead: boolean;
-  modifiers: string[];
+  modifiers: Modifiers;
```
notbakaneko:
```suggestion
  modifiers?: Modifiers;
```

and delete the `defaultProps` (which is array 👀 )

notbakaneko (changes requested):
comment.tsx
achievement-badge-icon.tsx
achievement-badge.tsx
notification-read-button.tsx

should probably also be `modifiers?: Modifiers` for consistency?

there's also the `data-modifiers` in `_search_result_user.blade.php`

notbakaneko (approved):
syntax isn't any weirder than css `class` already being space delimited... (and solves the array being new props problem)

PR #12881, opened 2026-03-27
Show osu! ruleset map if there's only convert maps for the user's ruleset

### LINKED ISSUE #12117: Show beatmap's original ruleset when user has a non-standard default ruleset.
As stated. This only applies to beatmaps which have a standard ruleset difficulty, as it will switch to showing the convert from your default ruleset. I am not opening a standard set to view its mania converts. I am opening it to see the actual beatmap.

nanaya:
Just cleaning up my backlog.

Resolves #12117

### DIFF
--- a/resources/js/utils/beatmap-helper.ts
+++ b/resources/js/utils/beatmap-helper.ts
@@ -31,11 +31,13 @@ const difficultyTextColourSpectrum = d3.scaleLinear<string>()
   .range(['#F6F05C', '#FF8068', '#FF4E6F', '#C645B8', '#6563DE', '#18158E'])
   .interpolate(d3.interpolateRgb.gamma(2.2));
 
-interface FindDefaultParams<T> {
-  group?: Map<Ruleset, T[]>;
-  items?: T[];
+type FindDefaultParams<T> = {
+  group: Map<Ruleset, T[]>;
+  items?: undefined;
+} | {
+  items: T[];
   ruleset?: Ruleset;
-}
+};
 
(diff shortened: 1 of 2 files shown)


notbakaneko (commented):
~~does this part actually get triggered by anything? I can't find anything that passes `mode` but not `items`~~
typing into the wrong text box...

resources/js/utils/beatmap-helper.ts
```diff
@@ -57,10 +57,13 @@ export function findDefault<T extends BeatmapJson>(params: FindDefaultParams<T>)
 
   if (params.group == null) return null;
 
-  const findModes = params.mode == null ? userModes() : [params.mode];
+  if (params.mode != null) {
+    return findDefault({ items: params.group.get(params.mode) ?? [], mode: params.mode });
+  }
```
notbakaneko: does this part actually get triggered by anything? I can't find anything that passes `mode` but not `items`

nanaya: apparently if the hash is just ruleset like `#fruits`...

notbakaneko: doesn't that pass `items` but not `mode` from here? 

https://github.com/ppy/osu-web/blob/4f00c59aad9cdf246b6599e272ce8d7db9416916/resources/js/beatmapsets-show/controller.ts#L62

🤔

nanaya: right, although that's also differently buggy...

notbakaneko (approved):

PR #12883, opened 2026-03-27
Limit measured loudness statistics

nanaya:
God [why](https://osu.ppy.sh/beatmapsets/1001317) 😐

(also done by the normalize script)


notbakaneko (approved):
ಠ_ಠ

PR #12887, opened 2026-04-01
Streamline SelectOptions rendering

notbakaneko:
Removes the render callbacks and let other components using `SelectOptions` to have more control and directly specify what to render instead of callbacks. 

Extracted from the branch that is adding support for #11666 and other sorting and filtering enhancements, so the component already supports multi selection, but the other parts are not included in this PR.

### DIFF
--- a/resources/js/beatmap-discussions/user-filter.tsx
+++ b/resources/js/beatmap-discussions/user-filter.tsx
@@ -64,7 +64,7 @@ export class UserFilter extends React.Component<Props> {
     }
 
     const style = groupColour(this.getGroup(this.props.discussionsState.selectedUser));
-    return <span className='u-group-colour u-ellipsis-overflow' style={style}>{this.props.discussionsState.selectedUser?.username}</span>;
+    return <span className='u-group-colour u-ellipsis-overflow' style={style}>{this.props.discussionsState.selectedUser.username}</span>;
   }
 
   @computed
--- a/resources/js/beatmap-discussions/user-filter.tsx
+++ b/resources/js/beatmap-discussions/user-filter.tsx
@@ -2,7 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 import mapperGroup from 'beatmap-discussions/mapper-group';
-import SelectOptions, { OptionRenderProps } from 'components/select-options';
+import SelectOptions from 'components/select-options';
 import BeatmapsetDiscussionsStore from 'interfaces/beatmapset-discussions-store';
 import UserJson from 'interfaces/user-json';
 import { action, computed, makeObservable } from 'mobx';
@@ -12,22 +12,18 @@ import * as React from 'react';
 import { makeUrl, parseUrl } from 'utils/beatmapset-discussion-helper';
 import { groupColour } from 'utils/css';
 import { trans } from 'utils/lang';
+import { getInt } from 'utils/math';
 import DiscussionsState from './discussions-state';
 
 const allUsers = Object.freeze({
   id: null,
-  text: trans('beatmap_discussions.user_filter.everyone'),
-});
-
-const noSelection = Object.freeze({
-  id: null,
-  text: trans('beatmap_discussions.user_filter.label'),
+  username: trans('beatmap_discussions.user_filter.everyone'),
 });
 
 interface Option {
-  groups: UserJson['groups'];
+  groups?: UserJson['groups'];
   id: UserJson['id'] | null;
-  text: UserJson['username'];
+  username: UserJson['username'];
 }
 
 interface Props {
@@ -35,25 +31,10 @@ interface Props {
   store: BeatmapsetDiscussionsStore;
 }
 
-function mapUserProperties(user: UserJson): Option {
-  return {
-    groups: user.groups,
-    id: user.id,
-    text: user.username,
-  };
-}
-
 @observer
 export class UserFilter extends React.Component<Props> {
   private get ownerId() {
     return this.props.discussionsState.beatmapset.user_id;
-  }
-
-  @computed
-  private get selected() {
-    return this.props.discussionsState.selectedUser != null
-      ? mapUserProperties(this.props.discussionsState.selectedUser)
-      : noSelection;
   }
 
   @computed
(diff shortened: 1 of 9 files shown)


resources/js/components/select-options.tsx
```diff
@@ -102,50 +84,47 @@ export default class SelectOptions<T extends Option> extends React.Component<Pro
...
-    this.props.onChange?.(option);
-  };
-
-  private renderOption({ children, onClick, option, selected = false }: ComponentOptionRenderProps<T>) {
-    const cssClasses = classWithModifiers(`${bn}__option`, { selected });
 
-    if (this.props.renderOption != null) {
-      return this.props.renderOption({ children, cssClasses, onClick, option });
+    const id = event.currentTarget.dataset.id;
+    this.showingSelector = this.props.onSelect?.(id) ?? false;
```
nanaya: if there's no onSelect prop, maybe just hide the selector and let turbo handle the navigation? That'll remove the need for the props for like half of the current usages of the component.

something like
```
if (this.props.onSelect == null) {
  this.showingSelector = false;
} else {
  event.preventDefault(); // or also handled by the onSelect?
  ...
}
```

resources/js/components/select-options.tsx
```diff
@@ -1,44 +1,33 @@
...
+  // the callback should return the display state the selector should go into after the click, or undefined for the default.
+  // the typing of id is to match dataset, so the component doesn't handle the parsing.
+  onSelect?: (id?: string) => boolean | void;
+  options: Iterable<RenderableOption<T>>;
+  selected: RenderableOption<T>['id'] | Set<RenderableOption<T>['id']>;
 }
 
 @observer
-export default class SelectOptions<T extends Option> extends React.Component<Props<T>> {
+export default class SelectOptions<T extends string | number> extends React.PureComponent<React.PropsWithChildren<Props<T>>> {
```
nanaya: is the `PropsWithChildren` thing actually needed

resources/js/components/select-options.tsx
```diff
@@ -1,44 +1,33 @@
...
-import { classWithModifiers, Modifiers } from 'utils/css';
+import { classWithModifiers, Modifiers, ModifiersExtended } from 'utils/css';
 
 const bn = 'select-options';
 
-export interface Option {
-  id: string | number | null;
-  text: string | React.ReactNode;
+interface RenderableOption<T> {
+  children: string | React.ReactNode;
```
nanaya: just noticed, `string |` isn't actually needed...?

resources/js/interfaces/select-option-json.ts
```diff
@@ -1,8 +1,7 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
 // See the LICENCE file in the repository root for full licence text.
 
-import { Option } from 'components/select-options';
-
-export default interface SelectOptionJson extends Option {
-  id: number;
+export default interface SelectOptionJson {
+  id: string | number;
+  text: string;
```
nanaya, line 6: seems wrong calling this `SelectOptionJson` when the actual `SelectOption` prop uses `children` instead of `text`

resources/js/components/select-options.tsx
```diff
@@ -102,50 +84,47 @@ export default class SelectOptions<T extends Option> extends React.Component<Pro
...
+        ? this.props.selected.has(option.id)
+        : this.props.selected === option.id;
+
+      yield this.renderOption(option, { selected });
+    }
   }
 
-  private renderText(text: T['text']) {
-    return typeof text === 'string' ? (
+  private renderText(children: RenderableOption<T>['children']) {
```
nanaya: render"text" but the variable name is `children` 🤔 I wonder if the variable name and related fields should just stay as `text`

nanaya (changes requested):

resources/js/components/select-options.tsx
notbakaneko: it's so the prop will always match whatever it's supposed to be typed as, but then I notice the prop isn't declared as required ಠ_ಠ

resources/js/components/select-options.tsx
notbakaneko: I think there was some kind of type discrimination issue when the other ones were originally declared like this 🤔

resources/js/beatmap-discussions/user-filter.tsx
```diff
@@ -69,13 +50,28 @@ export class UserFilter extends React.Component<Props> {
...
   }
 
+  @computed
+  private get text() {
+    if (this.props.discussionsState.selectedUser == null) {
+      return trans('beatmap_discussions.user_filter.label');
+    }
+
+    const style = groupColour(this.getGroup(this.props.discussionsState.selectedUser));
+    return <span className='u-group-colour u-ellipsis-overflow' style={style}>{this.props.discussionsState.selectedUser?.username}</span>;
```
nanaya: `selectedUser` can't be null here?

resources/js/components/basic-select-options.tsx
```diff
@@ -24,31 +23,36 @@ type Props = PropsBase & ({
   type: 'matchmaking';
 });
 
+// TODO: stricter typing; require only one of string or number at a time, not both.
 export default class BasicSelectOptions extends React.PureComponent<Props> {
+  private get options() {
+    return this.props.items.map((item) => ({
+      href: this.href(item.id),
+      id: item.id,
+      text: item.text,
```
nanaya: can be `...item`...?

resources/js/components/select-options.tsx
```diff
@@ -102,45 +86,43 @@ export default class SelectOptions<T extends Option> extends React.Component<Pro
...
-        this.optionSelected(event, option);
-      },
-      option,
-      selected: this.props.selected?.id === option.id,
-    }));
+  private *renderOptions() {
+    for (const option of this.props.options) {
+      const selected = this.props.selected instanceof Set || isObservableSet(this.props.selected)
+        ? this.props.selected.has(option.id)
+        : this.props.selected === option.id;
```
nanaya: probably doesn't matter but seems a bit weird doing a constant type check in a loop

resources/js/components/select-options.tsx
```diff
@@ -102,45 +86,43 @@ export default class SelectOptions<T extends Option> extends React.Component<Pro
...
-      children: this.renderText(option.text),
-      onClick: (event: React.MouseEvent) => {
-        this.optionSelected(event, option);
-      },
-      option,
-      selected: this.props.selected?.id === option.id,
-    }));
+  private *renderOptions() {
+    for (const option of this.props.options) {
+      const selected = this.props.selected instanceof Set || isObservableSet(this.props.selected)
```
nanaya: probably doesn't matter but seems a bit weird doing a constant type check in a loop

resources/js/components/select-options.tsx
```diff
@@ -1,44 +1,35 @@
...
-  onChange: (option: T) => void;
-  options: T[];
-  renderOption?(props: OptionRenderProps<T>): React.ReactNode;
-  selected: T;
+  // the callback should return the display state the selector should go into after the click, or undefined for the default.
+  // the typing of id is to match dataset, so the component doesn't handle the parsing.
+  onSelect?: (id?: string) => boolean | void;
+  options: Iterable<RenderableOption<T>>;
+  selected: RenderableOption<T>['id'] | Set<RenderableOption<T>['id']>;
+  text: React.ReactNode;
```
nanaya: for this here I wonder if react does something different if using `children` prop

nanaya (changes requested):
also conflicts

resources/js/components/select-options.tsx
notbakaneko: `children` is just whatever is passed between the jsx tags, making it required makes the typecheck fail if it's missing, not having it just ignores it and `PropsWithChildren` has it optional which is...not useful

nanaya (approved):

PR #12888, opened 2026-04-03
Prepare support for filtering Beatmap Discussions by multiple users

notbakaneko:
part of adding support for #11666

This PR migrates the single user selection state and url handling to support multiple user selections (and additional filtering options) while keeping the current UI. I wanted this separate to make sure it wasn't magically breaking something 👀


nanaya (commented):
merge conflicts

resources/js/beatmap-discussions/discussion.tsx
```diff
@@ -103,10 +103,26 @@ export class Discussion extends React.Component<Props> {
...
+  private get replies() {
+    return this.props.discussionsState != null ? this.props.discussion.posts.slice(1) : [];
+  }
+
   private get users() {
     return this.props.store.users;
   }
 
+  @computed
+  private get visibleReplies() {
```
notbakaneko, line 116: this is for the extra toggles that will be added later; it doesn't account for deleted replies, but I wonder if it should 🤔

resources/css/bem/beatmap-discussion.less
```diff
@@ -107,6 +107,12 @@
     flex-direction: column;
   }
 
+  &__info {
+    color: hsl(var(--hsl-f1));
+    font-size: @font-size--normal;
+    padding: 5px @_top-padding-x;
```
nanaya: align with the rest of the section? and the top/bottom padding should be 10px instead?

resources/js/beatmap-discussions/discussions-state.ts
```diff
@@ -68,8 +68,10 @@ export default class DiscussionsState {
 
   @observable readPostIds = new Set<number>();
   @observable selectedNominatedRulesets: Ruleset[] = [];
-  @observable selectedUserId: number | null = null;
+  @observable readonly selectedUserIds = new Set<number>();
+  @observable selectedUserIncludeReplies = false;
```
nanaya: `selectedUsersIncludeReplies`? or instead `user(s?)FilterIncludeReplies`?

resources/js/beatmap-discussions/discussion.tsx
```diff
@@ -287,6 +280,31 @@ export class Discussion extends React.Component<Props> {
     );
   }
 
+  private renderReplies() {
+    if (this.props.discussionsState == null) return null;
+
+    let cssClasses = `${bn}__expanded`;
+    if (this.collapsed) {
```
nanaya, line 287: unrelated but I think this was to prevent losing composed message when collapsed but the addition of saving to local storage probably removes the need for this?

resources/js/beatmap-discussions/discussions-state.ts
```diff
@@ -337,9 +342,10 @@ export default class DiscussionsState {
     return null;
   }
 
+  // TODO: maybe get rid of this and use the ids directly?
   @computed
-  get selectedUser() {
-    return this.store.users.get(this.selectedUserId);
+  get selectedUsers() {
+    return [...this.selectedUserIds].map((id) => this.store.users.get(id)).filter((user): user is UserJson => user != null);
```
nanaya: can be dynamically casted with `user is NonNullable<typeof user>` but that's kinda long isn't it...

resources/js/utils/beatmapset-discussion-helper.ts
```diff
@@ -294,14 +302,20 @@ export function parseUrl(urlString?: string | null, discussions?: BeatmapsetDisc
...
+  }
 
   const ret: ParsedUrlParams = {
     beatmapId,
     beatmapsetId,
     filter: isFilter(filter) ? filter : defaultFilter,
     // empty path segments are ''
     mode: isDiscussionPage(mode) ? mode : defaultMode(beatmapId),
-    user: getInt(url.searchParams.get('user')),
+    users,
```
nanaya, line 314: sort and remove duplicates? (or maybe for the makeUrl instead)

resources/js/beatmap-discussions/type-filters.tsx
```diff
@@ -80,7 +80,10 @@ export default class TypeFilters extends React.Component<Props> {
 
   private readonly setFilter = (event: React.SyntheticEvent<HTMLElement>) => {
     event.preventDefault();
-    this.props.discussionsState.changeFilter(event.currentTarget.dataset.type);
+    event.currentTarget.blur(); // to unfocus re-selected filter
+    const type = event.currentTarget.dataset.type;
+    const nextFilter = type === this.props.discussionsState.currentFilter ? 'total' : type;
+    this.props.discussionsState.changeFilter(nextFilter);
```
nanaya: this isn't related to the user filter change, is this?

nanaya (changes requested):

resources/js/beatmap-discussions/discussion.tsx
notbakaneko: it also means large parts of the DOM tree don't keep getting added and removed every time it gets toggled, although it's questionable which is better for this...

nanaya: the default is open so it probably doesn't matter

also I just remembered there are more states than just the content itself... (the reply input box open/close, message edit etc)

resources/js/utils/beatmapset-discussion-helper.ts
```diff
@@ -181,6 +182,7 @@ export function makeUrl(options: MakeUrlOptions) {
     filter,
     post,
     user,
+    users,
```
nanaya, line 184: the `user` one isn't needed anymore, is it?

nanaya (approved):

PR #12899, opened 2026-04-08
Notify user on mention in chat

nanaya:
This is just the notification part. No autocomplete or link to jump to the message or anything fancy like that.

Limited to single mention (for now? or should it be configurable? or more mentions for certain groups?)

I wonder what a good delivery mode defaults should be for this (currently the same as others - web push only).

- [x] depends on #12871

### DIFF
--- a/tests/Libraries/ChatTest.php
+++ b/tests/Libraries/ChatTest.php
@@ -15,6 +15,7 @@ use App\Models\Chat\Channel;
 use App\Models\Chat\Message;
 use App\Models\OAuth\Client;
 use App\Models\User;
+use App\Models\UserNotificationOption;
 use Event;
 use Exception;
 use Mail;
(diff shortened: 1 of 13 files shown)


tests/Libraries/ChatTest.php
```diff
@@ -156,6 +157,35 @@ public function testSendMessage(bool $verified, ?string $expectedException)
...
+        $sender = User::factory()->create();
+        $mention = User::factory()->create();
+        $mention->notificationOptions()->create([
+            'name' => UserNotificationOption::CHANNEL_MENTION,
+            'details' => [
+                'mail' => true,
+                'push' => true,
+            ],
+        ]);
+        $mentionUsername = strtr($mention->username, [' ' => '_']);
```
notbakaneko: I think there should be a separate test specifically for this case as well 🤔

notbakaneko (changes requested):

nanaya:
being able to mention in public channel is a bit weird but there's no user listing and users don't even join channel in stable. and it's not like anyone can create random public channels

the main problem with this is the link will go nowhere. a join option would probably be nice at least

notbakaneko (approved):

PR #12905, opened 2026-04-10
Add required utility package

nanaya:
Apparently frankenphp updated itself to run the groupadd operation earlier? And shadow-utils got installed by something else (chromium) and everything were only worked by a coincidence since frankenphp itself doesn't pull in shadow-utils.

I wonder if it's better to just pull in `10` image instead of `10-minimal`...


notbakaneko (approved):
what fun

PR #12906, opened 2026-04-10
Update docker compose filename

nanaya:
Easier autocomplete/adjustment between compose.yml and compose.override.yml

the downside is it's a bit too similar to `composer.{json,lock}`


notbakaneko:
`docker-compose.override.yaml` works, though?

nanaya:
`docker-compose.y{,a}ml` is [the old filename](https://docs.docker.com/compose/intro/compose-application-model/#the-compose-file)

ThePooN (approved):

notbakaneko (approved):

PR #12909, opened 2026-04-14
Use ref for referencing current wiki autocomplete selection

nanaya:
Randomly noticed this when looking at existing autocomplete thing.

Seems a bit cleaner this way?


notbakaneko (approved):
I think reusing refs for different elements is generally not recommended but this should be fine? 🤔

PR #12910, opened 2026-04-14
Use video for download page background

notbakaneko:
Something like this? 🤔 

https://github.com/user-attachments/assets/1626b187-ca08-427d-8c2a-b239a166a825

...and of course there's a problem loading the video after navigating with Safari
Somehow setting `src` directly makes it play most of the time (instead of a empty block)


nanaya (commented):
I wonder if it should be left anchored instead

nanaya (approved):
I guess it's fine (until someone complains their pc dying because of the video)

PR #12911, opened 2026-04-15
Add typeguard to string.present

notbakaneko:
Something I found while cleaning up old branches, not sure how useful it actually is other than narrowing the type down (which doesn't seem to be relied on in the current cases it's used)

### DIFF
--- a/resources/js/utils/string.ts
+++ b/resources/js/utils/string.ts
@@ -5,6 +5,6 @@ export function presence(value?: string | null) {
   return value != null && value !== '' ? value : null;
 }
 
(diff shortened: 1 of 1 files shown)


nanaya (commented):
technically this has same problem with [typeguarding `Number.isFinite`](https://github.com/microsoft/TypeScript/issues/15048) in that a false `present(x)` call doesn't always mean `x` is null - it could be an empty string.

notbakaneko:
That's probably why I didn't add it after all since the guard also narrows down the inverse :|

oh dejavu #10561

PR #12913, opened 2026-04-16
Fix error on viewing contest as guest

nanaya:
┐(･.･)┌


notbakaneko (approved):
I could have sworn I added that already...

Issue #1622, opened 2017-10-24
Revisit image caching


peppy:
- Make async (best effort display if dimensions unavailable, rather than blocking request).
- Handle errors better (retry retrieval on failure x times before giving up?).

PR #12915, opened 2026-04-16
Limit bbcode image size fetching to 10

nanaya:
Maybe? Side effect may include broken gallery on first load and moving scroll position on image load (or maybe it's fixed now? it's kinda hard to test)

- [x] depends on #12914

Also will need queue worker adjustment to handle the new queue `imagesize` (and maybe remove `store-notifications` if it's still being used) @ThePooN 

(related to #1622)


notbakaneko:
10....what? 🤔

nanaya:
10 fetches

notbakaneko (dismissed):

notbakaneko (approved):

PR #12920, opened 2026-04-19
Clean up and update quick/ranked play rankings

LiquidPL:
* split ranked and quick play into separate tabs
* remove points sort/column for ranked play

notably this changes the tab on the ranking page from `quick play` to `matchmaking` while still being under the same translation key which should be fine cuz crowdin should catch that, while also changing the routes a bit.

image

image

~~the ranked play tab is showing no dropdown here since i only had one pool in the DB (which i believe is the current behaviour anyway). also sorry for not having any test data in here.~~ updated the screenshots


LiquidPL:
Drafting this for if we'll even need to bring quick play rankings back.

PR #12922, opened 2026-04-20
Cache oauth tokens

josermzte:
Inspired by [this thread](https://x.com/ThePooN/status/2045905428152713473), I took a look at the project and noticed OAuth tokens aren't cached. If there's a reason this isn't already cached that I'm missing, feel free to close.

`AuthApi::validTokenFromRequest` runs two OAuth queries on every authenticated API request. One to look up the client, one to look up the access token. The JWT is already cryptographically verified upstream, so these DB hits are only there to confirm nothing's been revoked or expired.

The change caches the validated token (with its client relation attached) in Redis, keyed by token id. On cache hit, skip the two queries. On miss or any Redis hiccup, fall back to the existing DB path.

Also threw in tests for the cache and each invalidation path. Ran the OAuth suite locally to confirm nothing else broke, all good.

### TTL

Default is 60 seconds, configurable via `OAUTH_TOKEN_CACHE_DURATION` (0 disables entirely). 60 felt like a reasonable starting point. Easy to tweak later if the real hit data suggests otherwise.

### Invalidation

Explicit forget on every revocation path so the TTL is just a safety net:

- `Token::revoke()`, `markVerified()`, `setVerificationMethod()` for single-token mutations
- `Client::revoke()` / `resetSecret()` / `revokeForUser()` for bulk revocations (these pluck affected token ids before the `UPDATE` and forget each)

One extra check in the middleware compares the cached `client_id` against the `oauth_client_id` claim from the JWT. If they don't match, the entry is treated as stale and the middleware falls through to DB validation.


notbakaneko:
I don't think we want to be making round trips to redis, serializing and deserializing entire model objects and their relations, cache invalidation, vs a single row DB lookup.
This is making the whole process of handling tokens more complicated, still makes separate network requests to check for token validity (plus another for cache miss), when one of them is by the underlying library itself. If there's anything to solve, it should be exposing the model from the library so it can be directly accessed and validated in the middleware.

josermzte:
Yeah this is kinda embarrassing — I went too hard for the cache stuff and forgot about the BearerTokenValidator check.

> If there's anything to solve, it should be exposing the model from the library so it can be directly accessed and validated in the middleware.

We can totally override Passport's `Bridge\AccessTokenRepository` so `isAccessTokenRevoked` loads the token just once with `client` eager-loaded and exposes it to `AuthApi`, so the middleware skips re-querying entirely.

Happy to take a crack at that tonight if you're open to it.

josermzte:
Closing this one in favor of reusing the token that's already loaded on the first validation check.

PR #12925, opened 2026-04-21
Use css variables for discussion type filter colours

notbakaneko:
One thing less to check later, I guess? (and also not directly related to functional change)

### DIFF
--- a/resources/css/bem/counter-box.less
+++ b/resources/css/bem/counter-box.less
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 .counter-box {
+  --counter-box-border-colour: hsla(var(--hsl-b5), 0.25);
   @top: counter-box;
 
   @color-deleted: black;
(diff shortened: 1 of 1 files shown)


resources/css/bem/counter-box.less
```diff
@@ -2,33 +2,35 @@
...
-  }
+  @top: counter-box;
 
-  .with-active(@color) {
-    &:hover,
-    &:focus,
-    &:active,
-    &.js-active {
-      .border(@color);
+  @types: deleted, pending, praises, mapper-notes, mine, resolved, total;
```
nanaya: `@types: { deleted: black; mapper-notes: ... }` probably easier to understand?

the `@key`/`@value` part of this section https://lesscss.org/functions/#list-functions-each

resources/css/bem/counter-box.less
```diff
@@ -108,34 +82,7 @@
...
-    }
-
-    .@{top}--resolved & {
-      background-color: @color-resolved;
-    }
-
-    .@{top}--total & {
-      background-color: @color-total;
-    }
+    background-color: var(--line-colour, initial);
```
nanaya: maybe declare a default at the block so no need to worry about accidentally inheriting the value

nanaya (changes requested):

resources/css/bem/counter-box.less
notbakaneko: good point 🙈

nanaya (approved):

PR #12927, opened 2026-04-21
Close blank forum reply box with escape key

nanaya:
As requested by smoogi

This currently requires the input box to be actually in focus which probably is fine?


notbakaneko (approved):
seems kind of niche with it also needing to be empty but 🤷

PR #12928, opened 2026-04-21
Remove schedule service from docker compose

nanaya:
Less messing around with data in unexpected way and just less work in general (especially on larger dev data).


notbakaneko (changes requested):
also should be removed from `SETUP.md`

ThePooN:
If that's removed from compose, there should be a line of documentation on how to run them instead as needed.

nanaya:
I don't think any of them are needed for dev. One already needs to know the `artisan` command itself and all the relevant commands are listed there.

(and unless you run the docker 24/7, you probably don't actually want to rely on the existing schedule anyway)

notbakaneko (approved):

PR #12931, opened 2026-04-22
Combine notification stack queries

nanaya:
This can't get much uglier ┐( ﾟ ヮﾟ)┌

It seems to work at least?


nanaya:
nvm, the category thing made it almost impossible to deal with correctly without pre-mapping it in the table itself ಠ_ಠ

PR #12932, opened 2026-04-22
Fix enchant widget

nanaya:
Apparently they changed the way the widget is loaded at some point.


resources/views/objects/_enchant.blade.php
```diff
@@ -3,5 +3,5 @@
     See the LICENCE file in the repository root for full licence text.
 --}}
 @if ($GLOBALS['cfg']['services']['enchant']['id'] !== null)
-    <div class="enchant-help-center" data-id="{{ $GLOBALS['cfg']['services']['enchant']['id'] }}"></div>
```
notbakaneko, line 6: isn't removing the class just going to make the load abort itself?

notbakaneko (approved):

PR #12938, opened 2026-04-23
Combine caching of selected channel

nanaya:
Seems silly caching it in multiple places. The check in `selected` doesn't seem worth caching so I removed it as well.

### DIFF
--- a/resources/js/chat/chat-state-store.ts
+++ b/resources/js/chat/chat-state-store.ts
@@ -63,6 +63,7 @@ export default class ChatStateStore implements DispatchListener {
     return new Set(this.channelStore.groupedChannels.PUBLIC.map((channel) => channel.channelId));
   }
 
+  @computed
   get selectedChannel() {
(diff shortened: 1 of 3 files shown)


notbakaneko (approved):
~~looks fine~~

resources/js/chat/conversation-list-item.tsx
```diff
@@ -22,15 +21,12 @@ export default class ConversationListItem extends React.Component<Props> {
     return this.props.channel.type !== 'TEAM';
   }
 
-  @computed
```
notbakaneko, line 25: ~~wait, this currently stops the every list item from rendering when selecting a new item~~
this change makes the entire channel list re-render every time a channel is selected

(reworded to make more sense)

notbakaneko (changes requested):

resources/js/chat/conversation-list-item.tsx
nanaya: stops the... what?

notbakaneko (approved):

PR #12939, opened 2026-04-23
Support filtering Beatmap Discussion by multiple users

### LINKED ISSUE #11666: Sorting by multiple people on Beatmap's Discussion page

notbakaneko:
Filter by multiple users, include/exclude replies, hide/include replies from other users within the discussion.
Also moves the filters down to be next to the discussions closer with all the other sorting and filter options, and pins them to the top for easier access.

image

with a user selected:
image

I did think about putting all the usernames selected as the main dropdown text but it just gets long and truncated so I went with `Multiple users selected` instead...

closes #11666

### DIFF
--- a/resources/css/bem/beatmapset-discussions-toolbar.less
+++ b/resources/css/bem/beatmapset-discussions-toolbar.less
@@ -1,0 +1,58 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+.beatmapset-discussions-toolbar {
+  --num-columns: 6;
+  display: flex;
+  flex-direction: column;
+  gap: 10px;
+  background-color: hsl(var(--hsl-b4));
+  padding: 10px;
+
+  &__group {
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: flex-end;
+    gap: 10px;
+
+    &--filters {
+      @media @desktop {
+        display: grid;
+        grid-template-columns: minmax(0, 500px) repeat(var(--num-columns), auto);
+      }
+
+      flex-direction: column;
+      flex-wrap: nowrap;
+    }
+  }
+
+  &__item {
+    .reset-input();
+    display: flex;
+    align-items: center;
+    gap: 5px;
+    flex: none;
+
+    &--link {
+      .link-plain();
+      .default-border-radius();
+      font-size: @font-size--normal;
+      color: white;
+      padding: 5px 10px;
+
+      &:hover {
+        background-color: hsl(var(--hsl-b3));
+      }
+    }
+  }
+
+  &__type-filters {
+    display: flex;
+    flex-wrap: wrap;
+    gap: 10px;
+
+    @media @desktop {
+      display: contents;
+    }
+  }
+}
--- a/resources/css/bem/beatmapset-discussions-toolbar.less
+++ b/resources/css/bem/beatmapset-discussions-toolbar.less
@@ -21,6 +21,7 @@
         grid-template-columns: minmax(0, 500px) repeat(var(--num-columns), auto);
       }
 
+      // force the elements to stack to avoid them shifting around if they resize.
       flex-direction: column;
       flex-wrap: nowrap;
     }
--- a/resources/css/bem/beatmapset-discussions-toolbar.less
+++ b/resources/css/bem/beatmapset-discussions-toolbar.less
@@ -27,21 +27,18 @@
 
   &__item {
     .reset-input();
+    .link-plain();
+    .default-border-radius();
     display: flex;
     align-items: center;
     gap: 5px;
+    padding: 5px 10px;
     flex: none;
+    font-size: @font-size--normal;
+    color: white;
 
-    &--link {
-      .link-plain();
-      .default-border-radius();
-      font-size: @font-size--normal;
-      color: white;
-      padding: 5px 10px;
-
-      &:hover {
-        background-color: hsl(var(--hsl-b3));
-      }
+    &:hover {
+      background-color: hsl(var(--hsl-b3));
     }
   }
 
@@ -49,9 +46,5 @@
     display: flex;
     flex-wrap: wrap;
     gap: 10px;
-
-    @media @desktop {
-      display: contents;
-    }
   }
 }
--- a/resources/css/bem/beatmapset-discussions-toolbar.less
+++ b/resources/css/bem/beatmapset-discussions-toolbar.less
@@ -2,7 +2,6 @@
 // See the LICENCE file in the repository root for full licence text.
 
 .beatmapset-discussions-toolbar {
-  --num-columns: 6;
   display: flex;
   flex-direction: column;
   gap: 10px;
@@ -17,8 +16,7 @@
 
     &--filters {
       @media @desktop {
-        display: grid;
-        grid-template-columns: minmax(0, 500px) repeat(var(--num-columns), auto);
+        flex-direction: row;
       }
 
       // force the elements to stack to avoid them shifting around if they resize.
--- a/resources/css/bem/counter-box.less
+++ b/resources/css/bem/counter-box.less
@@ -44,6 +44,10 @@
     color: @osu-colour-c1;
   }
 
+  &--beatmap-discussions {
+    min-width: 60px; // allows up to 3 digits without resizing.
+  }
+
   &--info {
     padding: 10px 0;
     background-color: transparent;
@@ -86,8 +90,8 @@
     font-size: @font-size--small;
     white-space: nowrap;
 
-    .@{top}--beatmap-discussions & {
-      margin-right: 10px;
+    &--beatmap-discussions {
+      margin-right: 0;
     }
 
     .@{top}--info & {
--- a/resources/css/bem/counter-box.less
+++ b/resources/css/bem/counter-box.less
@@ -44,11 +44,6 @@
     color: @osu-colour-c1;
   }
 
-  &--beatmap-discussions {
-    margin: 5px;
-    height: var(--box-height);
-  }
-
   &--info {
     padding: 10px 0;
     background-color: transparent;
--- a/resources/css/bem/counter-box.less
+++ b/resources/css/bem/counter-box.less
@@ -71,6 +71,7 @@
 
   &__content {
     line-height: 1;
+    min-width: 0; // prevents line from squishing in beatmapset discussions toolbar.
   }
 
   &__count {
--- a/resources/css/bem/counter-box.less
+++ b/resources/css/bem/counter-box.less
@@ -3,6 +3,8 @@
 
 .counter-box {
   --counter-box-border-colour: hsla(var(--hsl-b5), 0.25);
+  --counter-box-title-font-weight: 600;
+  --counter-box-title-margin-right: 30px;
   --line-colour: initial;
 
   @top: counter-box;
@@ -45,10 +47,13 @@
   }
 
   &--beatmap-discussions {
+    --counter-box-title-margin-right: 0;
     min-width: 60px; // allows up to 3 digits without resizing.
   }
 
   &--info {
+    --counter-box-title-font-weight: normal;
+    --counter-box-title-margin-right: 0;
     padding: 10px 0;
     background-color: transparent;
     border: none;
@@ -85,18 +90,9 @@
   }
 
   &__title {
-    margin-right: 30px;
-    font-weight: 600;
+    margin-right: var(--counter-box-title-margin-right);
+    font-weight: var(--counter-box-title-font-weight);
     font-size: @font-size--small;
     white-space: nowrap;
-
-    &--beatmap-discussions {
-      margin-right: 0;
-    }
-
-    .@{top}--info & {
-      font-weight: normal;
-      margin-right: 0;
-    }
   }
 }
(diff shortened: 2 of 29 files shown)


resources/js/beatmap-discussions/discussion.tsx
```diff
@@ -115,12 +115,12 @@ export class Discussion extends React.Component<Props> {
   @computed
   private get visibleReplies() {
     if (this.props.discussionsState == null
-      || this.props.discussionsState.selectedUserIds.size === 0
```
notbakaneko, line 118: there's an existing issue if the url param contains a user id that isn't in the discussions, I wonder if state initialization should sanitize and normalize the urls first 🤔

nanaya:
The floating bar seems like something people will be annoyed with as it takes up vertical space...? (and even worse on mobile)

notbakaneko:
I did have an idea of adding the filters into the floating toolbar as a popup but that would have been a next step.

Also realized the filter font sizes not shrinking down when the other tabbar sticky does looks odd...

nanaya:
maybe a toggle button on the page tab bar?

notbakaneko:
how about a collapsible bar? 🤔 
image

nanaya:
that looks weird and the expanded view takes even more space ಠ_ಠ

it being still collapsed when scrolled up seems a bit pointless as well (and it should not be floating by default either?)

notbakaneko:
I don't think stacking even more optionally pinable elements is the solution and looks even more weird, and having a toggle button in the tab bar that can also scroll sideways doesn't make sense either (and does nothing on the history tab).

nanaya:
how about just not make it floatable/pinnable for now...?

(also the new one cuts off the user list dropdown when it's too long instead of extending the page, although it really should just have its height limited and show scrollbar instead)

notbakaneko:
Removing the pinning for now so the feature can go out while I try something else out, but that has some refactoring involved and may have additional changes that conflict

resources/css/bem/page-extra-tabs.less
nanaya, line 1: should be removed/renamed?

resources/js/beatmap-discussions/user-filter.tsx
```diff
@@ -119,13 +134,29 @@ export class UserFilter extends React.Component<Props> {
...
+      : this.discussionsState.selectedUserIds.has(user.id);
+
     return {
       href: urlOptions != null ? makeUrl(urlOptions) : '#',
       id: user.id,
-      text: <span className='u-group-colour u-ellipsis-overflow' style={this.styleForUser(user)}>{user.username}</span>,
+      text: (
+        <>
+          <div className='osu-switch-v2'>
+            <input
```
nanaya: `input` can't be inside an `a`

nanaya (changes requested):

resources/js/beatmap-discussions/user-filter.tsx
notbakaneko: I'm just removing the link instead of adding more workarounds for this, since it also doesn't seem to make sense to open in new tab with multi selection options.

nanaya: but `button` can't have `input` either ಠ_ಠ (and it just doesn't make sense if you think about it - like, are you checking the box or pressing the button)

even worse, `div` can't be inside `button`.

maybe add `--checked`, `--checkbox` modifiers to the class and use divs instead? It's not like the checkbox itself have much meaning in term of actual input.

resources/js/components/icon-expand.tsx
```diff
@@ -10,7 +10,7 @@ interface IconProps {
 
 interface Props {
   expand: boolean;
-  parentClass: string;
+  parentClass?: string;
```
nanaya: where's this needed?

resources/js/components/select-options.tsx
```diff
@@ -97,15 +98,20 @@ export default class SelectOptions<T extends string | number> extends React.Pure
...
-        className={classWithModifiers(`${bn}__option`, modifiers)}
+        className={classWithModifiers(`${bn}__option`, { selected })}
         data-id={option.id ?? undefined}
         href={option.href}
         onClick={this.optionSelected}
       >
+        {this.props.useCheckmark && (
+          <span className={`${bn}__checkmark`}>
+            {selected && <span className='fas fa-check' />}
+          </span>
```
nanaya, line 113: technically can be done using the `fa-fw` class instead (``className={`fa-fw ${selected ? ... }`}``) although that'll ignore the decoration colour thing

but then again the `--colour`/`--decoration-colour` thing don't seem to be used anywhere at the moment...

resources/css/bem/beatmapset-discussions-toolbar.less
```diff
@@ -0,0 +1,59 @@
...
+  &__group {
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: flex-end;
+    gap: 10px;
+
+    &--filters {
+      @media @desktop {
+        display: grid;
+        grid-template-columns: minmax(0, 500px) repeat(var(--num-columns), auto);
```
nanaya: what does this do over just flex/row 🤔 (which would remove the need to specify the `num-columns`)

resources/lang/en/beatmap_discussions.php
```diff
@@ -113,5 +113,6 @@
     'user_filter' => [
         'everyone' => 'Everyone',
         'label' => 'Filter by user',
+        'multiple' => 'Multiple users selected',
```
nanaya: I wonder if showing the selected count would be useful

resources/js/beatmap-discussions/discussions-state.ts
```diff
@@ -70,7 +70,9 @@ export default class DiscussionsState {
   @observable repliesIncludeSelectedUsers = false;
```
nanaya, line 70: already mentioned somewhere else but this addition should've also updated the user listing to include replying users

resources/css/bem/counter-box.less
```diff
@@ -91,8 +90,8 @@
     font-size: @font-size--small;
     white-space: nowrap;
 
-    .@{top}--beatmap-discussions & {
-      margin-right: 10px;
+    &--beatmap-discussions {
+      margin-right: 0;
```
nanaya: or make it a variable and replace it in the block's modifier?

resources/js/beatmap-discussions/discussion.tsx
```diff
@@ -335,4 +334,19 @@ export class Discussion extends React.Component<Props> {
...
+      yield* this.replies;
+      return;
+    }
+
+    for (const reply of this.replies) {
+      if (!this.showDeleted && reply.deleted_at != null) continue;
+      // always include non-deleted system posts for the markers and state changes
+      if (reply.system) yield reply;
+      if (!this.props.discussionsState.showOtherReplies && !this.props.discussionsState.selectedUserIds.has(reply.user_id)) continue;
+      yield reply;
```
nanaya: `reply.system` may end up being yielded twice

(also the only usage of this generator does `...thisGenerator()` anyway so might as well return plain array instead?)

resources/js/beatmap-discussions/toolbar.tsx
```diff
@@ -0,0 +1,129 @@
...
+          <span>
+            <span className={this.discussionsState.repliesIncludeSelectedUsers ? 'fas fa-check-square' : 'far fa-square'} />
+          </span>
+          <span>
+            {trans('beatmaps.discussions.include_replies')}
+          </span>
+        </button>
+        <button
+          className={`${bn}__item ${bn}__item--link`}
+          onClick={this.toggleShowOtherReplies}
```
nanaya, line 101: probably fine as is for now but it's kind of weird the checkbox options don't persist on reload/modify the url like the other filters in the same section

resources/js/beatmap-discussions/user-filter.tsx
```diff
@@ -82,11 +89,13 @@ export class UserFilter extends React.Component<Props> {
   render() {
     return (
       <SelectOptions
-        href={this.props.discussionsState.url}
+        blackout={false} // css sticky elements render on a different stacking context and always get covered by the blackout.
```
nanaya: the comment may get confusing without it being sticky

nanaya (changes requested):

resources/js/components/icon-expand.tsx
```diff
@@ -23,7 +23,7 @@ function Icon({ position, icon }: IconProps) {
 
 export default function IconExpand({ expand = false, parentClass }: Props) {
   return (
-    <span className={`icon-stack ${parentClass}`}>
+    <span className={`icon-stack ${parentClass ?? ''}`}>
```
nanaya: not nullable atm 🤔

resources/js/beatmap-discussions/user-filter.tsx
```diff
@@ -125,7 +142,6 @@ export class UserFilter extends React.Component<Props> {
       text: <span className='u-group-colour u-ellipsis-overflow' style={this.styleForUser(user)}>{user.username}</span>,
     };
   };
-
```
nanaya, line 128: the spacing should probably stay

resources/js/beatmap-discussions/user-filter.tsx
```diff
@@ -34,44 +34,53 @@ interface Props {
 @observer
 export class UserFilter extends React.Component<Props> {
   private get ownerId() {
-    return this.props.discussionsState.beatmapset.user_id;
+    return this.discussionsState.beatmapset.user_id;
   }
 
   // TODO: add actual multi user selection.
```
nanaya: the comment can be removed?

resources/css/bem/beatmapset-discussions-toolbar.less
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  &__type-filters {
+    display: flex;
+    flex-wrap: wrap;
+    gap: 10px;
+
+    @media @desktop {
+      display: contents;
+    }
```
nanaya: seems fine without?

resources/css/bem/beatmapset-discussions-toolbar.less
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  &__item {
+    .reset-input();
+    display: flex;
+    align-items: center;
+    gap: 5px;
+    flex: none;
+
+    &--link {
```
nanaya: all of them have this modifier? might as well make this the default and remove the modifier?

(also they are buttons...?)

resources/js/beatmap-discussions/toolbar.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  constructor(props: Props) {
+    super(props);
+    makeObservable(this);
+  }
+
+  render() {
+    return (
+      <>
+        <div className='beatmapset-discussions-toolbar'>
```
nanaya: some others use `bn` variable, some others like this one use the name directly...

resources/js/beatmap-discussions/toolbar.tsx
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  constructor(props: Props) {
+    super(props);
+    makeObservable(this);
+  }
+
+  render() {
+    return (
+      <>
```
nanaya: this wrapper doesn't do anything

resources/js/beatmap-discussions/toolbar.tsx
```diff
@@ -0,0 +1,124 @@
...
+        </div>
+      </>
+    );
+  }
+
+  private renderShowDeletedToggle() {
+    if (!canModeratePosts()) return null;
+
+    return (
+      <button
```
nanaya: extracting this button to its own component/function may be nice?

nanaya (changes requested):

resources/js/beatmap-discussions/user-filter.tsx
notbakaneko: that's weird, I thought I accidentally wrote it without the space but I guess it had the space all along 🤔

nanaya (approved):

PR #12950, opened 2026-04-26
Skip duplicate OAuth token lookup

josermzte:
Follow-up to #12922 

`BearerTokenValidator` already loads the token for the revocation check, so we override it to eager-load the client and stash the Token on the request. `AuthApi` reuses it instead of re-querying. One less DB query per authed request.

Threw in some tests too.


nanaya:
Fixed in different way with #13233.

PR #12956, opened 2026-05-07
Update API usage and rate limit guidelines

peppy:
I've had more and more questions recently which really should be covered by this section. And even people claiming that the actual rate limit *is 1200 requests per minute* (it's not).

This is a rewrite to hopefully address these issues.

### DIFF
--- a/resources/views/docs/info.md.blade.php
+++ b/resources/views/docs/info.md.blade.php
@@ -14,9 +14,26 @@ Code examples are provided in the dark area to the right, you can use the tabs a
 
 # Terms of Use
 
-Use the API for good. Don't overdo it. If in doubt, ask before (ab)using :). _this section may expand as necessary_.
+Use the API for good. Don't overdo it. If in doubt, [ask](mailto:pe@ppy.sh) before serious long term use.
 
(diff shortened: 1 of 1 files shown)


resources/views/docs/info.md.blade.php
```diff
@@ -14,9 +14,26 @@
...
+
+Examples of incorrect / abusive usage:
+
+- Polling for new data every minute for every user.
+- Polling more than once a minute for the same user/beatmap.
+- Using the API to try to gain a competitive advantage.
+- Using the API as if it is your database, re-requesting the same data every time it is needed.
+- Harvesting mass score/user/beatmap data (consider using [data.ppy.sh](https://data.ppy.sh) instead if looking for seed or sample data).
+
+Please **limit your usage to 60 requests per minute (generally 1 request per second)**. The internal rate limits are higher than this and allow some degree of bursting, but exceeding this specified limit may lead to your API tokens being revoked, or in serious abuse cases your access to the API being restricted.
```
notbakaneko: `...limit your usage to no more than 60 requests per minute...`?

peppy: Addressed in 374f454.

notbakaneko (approved):

PR #12960, opened 2026-05-11
Remove unused chat input container text alignment

nanaya:
What even ಠ_ಠ


notbakaneko (approved):
(I don't remember it changing anything last time I checked)

PR #12961, opened 2026-05-12
Add user suggestion endpoint

nanaya:
Moving the wiki one to the same controller may be useful? 🤷 

Alternatively add the new action to users controller instead but it's already rather long.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -319,6 +319,9 @@ Route::group(['middleware' => ['web']], function () {
     Route::post('session', 'SessionsController@store')->name('login');
     Route::delete('session', 'SessionsController@destroy')->name('logout');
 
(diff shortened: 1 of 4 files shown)


routes/web.php
```diff
@@ -319,6 +319,9 @@
     Route::post('session', 'SessionsController@store')->name('login');
     Route::delete('session', 'SessionsController@destroy')->name('logout');
 
+    Route::get('suggestions/user', 'SuggestionsController@user')->name('suggestions.user');
+    Route::get('suggestions/wiki', 'SuggestionsController@wiki')->name('suggestions.wiki');
```
notbakaneko, line 323: group them?

nanaya: maybe if there's another one? seems like a lot of ceremonies for barely any benefits 🤷‍♀️

notbakaneko (approved):

PR #12962, opened 2026-05-12
Add optional maximum participant count to realtime multiplayer rooms

bdach:
- Part of https://github.com/ppy/osu-server-spectator/issues/405

This PR allows specifying a maximum participant count when creating a room. This specification is optional.

Depending on whether the maximum is specified, I intend for one of two things to happen:

- The maximum is not specified. The room size is theoretically uncapped as it is now. Realtime rooms in the lazer client continue to work the same way as they did until now.

- The maximum is specified. The room cannot have more users than that. The enforcement of the maximum is done by `osu-server-spectator`. In addition, the maximum being present introduces *slots* to the room.

  Slots are a concept from stable wherein a room with N maximum users has N slots for users to occupy. Slots have no effect on gameplay other than visual ordering of the list of room participants. In normal multiplayer rooms, a user can change the slot by themselves, while in tournament contexts, slots are usually referee-assigned and sometimes have ulterior meaning (such as representing players' seedings, location in the tournament bracket, etc.)

  Web will actually not be concerned with *any* of that going forward, except for maybe a change to the room history page to include the slots in the sorting criteria of users in a match.

  The maximum participant count has two caps: one technical (column type used is `unsigned tinyint`, which is [0..255]), and one practical (the lower bound is 2, because nothing below that makes sense, and the upper bound is configurable via the environment and defaults to 16 which matches stable).

---

@ppy/team-client please read through above and voice objections as required.


peppy (dismissed):
Spec seems fine to me

nanaya (commented):
yeah seems fine apart of the `.env.example` needing an update as well

nanaya (approved):

Issue #12885, opened 2026-03-28
Receive notifications for replies to modding discussions


Serizawa-Haruki:
If you post a suggestion or problem in a map's discussion page, you automatically get notifications when someone replies to it. However, if you only reply to a post while not being the original poster, you don't get notified if someone replies back to you. It would be useful to enable this so that you don't have to watch the entire discussion page or check manually. I also noticed there are no notifications if you hype a map and someone replies to your hype post (I can't remember about praises).

I think this should simply be part of the existing setting (I assume it's the "beatmap modding" one). On a side note, a tooltip explaining what exactly each of these options does would be really helpful because some are not very clear:
<img width="315" height="221" alt="Image" src="https://github.com/user-attachments/assets/66fc62b3-6b2e-48a8-adbe-9515e0320342" />

peppy:
Seems like an oversight to me, definitely should be a thing.

notbakaneko:
iirc the original thinking was to avoid notifying everyone that comments in a discussion simply because they commented in it.

Serizawa-Haruki:
I get that, but I think if someone comments in a specific discussion they are at least somewhat invested, especially when other people might be responding to them directly. It's similar to forum threads where you get an automatic bookmark if you post in one.

PR #12963, opened 2026-05-13
Add tooltips for explaining the types of notifications

notbakaneko:
I think these are correct 🤔 

Was mentioned in passing [here](https://github.com/ppy/osu-web/issues/12885)


resources/views/accounts/_edit_notifications.blade.php
```diff
@@ -128,13 +128,16 @@ class="account-edit-entry__checkbox account-edit-entry__checkbox--grid"
...
+                                    'modifiers' => 'grid',
                                     'name' => "user_notification_option[{$name}][details][{$mode}]",
                                 ]])
                             </label>
                         @endforeach
 
-                        <span class="account-edit-entry__checkbox-label account-edit-entry__checkbox-label--grid">
+                        <span
+                            class="account-edit-entry__checkbox-label account-edit-entry__checkbox-label--grid"
+                            title="{{ osu_trans("accounts.notifications.tooltips.{$name}") }}"
```
nanaya: the tooltip placement seems weird (maybe put in another span?) and having it at the right side instead of top may feel better.

also there may be a bit of discoverability problem? although I don't know how improve it that doesn't also make it a lot more busy (underlining the label like with `<abbr>` or placing question mark icon on the side)

nanaya (changes requested):

nanaya (approved):

PR #12633, opened 2025-12-30
Improve user search relevance by boosting staff, contributors, and active users

002love:
Fixes #12579

### Rationale
Currently, user search results are often dominated by inactive accounts, making it difficult to find active community members or staff. This PR introduces a boosting mechanism in Elasticsearch to prioritize relevant users without affecting the accuracy of string matching

### Changes
- **Elasticsearch Mapping:** added `groups` field to the `users` index mapping
- **Indexing:** updated `UserSearch` trait to include user group IDs in the search index
- **Search Logic:**
    - refactored `UserSearch::getQuery()` to use a nested `bool` query
    - string matching is now inside a `must` block to maintain strict `minimum_should_match: 1` requirements
    - added `should` clauses for boosting:
        - **Staff & BN** (PPY, GMT, NAT, BN, DEV): `boost: 5`
        - **Contributors** (Alumni, Project Loved): `boost: 2`
        - **Recent Activity** (Active within last 30 days): `boost: 1.5`
    - **Performance:** group ids are retrieved via identifiers and cached in `static` variables to ensure zero database overhead

| Master | This PR |
| --- | --- |
| image | image |

PR #12965, opened 2026-05-14
Show user mention suggestion box

nanaya:
I don't think this can be split further 🫥 

One of the class names should be `...Controller` instead of `...State` but it's rather long already.

The result itself probably needs improvement from #12633. Mainly boosting the active users part.


resources/js/chat/mention-completion-box-state.ts
```diff
@@ -0,0 +1,184 @@
...
+  username: string;
+}
+
+export default class MentionCompletionBoxState {
+  @observable query = '';
+  @observable selectedIndex = 0;
+  @observable visible = false;
+
+  @observable private readonly cache: Partial<Record<string, UserSearchEntry[]>> = {};
+  private readonly debouncedFetchSuggestions = debounce(() => this.fetchSuggestions());
```
notbakaneko: this ends up sending a query every key press?

resources/js/chat/mention-completion-box.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { trans } from 'utils/lang';
+import MentionCompletionBoxEntry from './mention-completion-box-entry';
+import MentionCompletionBoxState from './mention-completion-box-state';
+
+interface Props {
+  state: MentionCompletionBoxState;
+}
+
+@observer
+export default class MentionCompletionBox extends React.Component<Props> {
```
notbakaneko, line 17: need `makeObservable(this)` for `@action` to work

notbakaneko (changes requested):
(still checking)

Flashing the old results doesn't seem right:
https://github.com/user-attachments/assets/baf9c104-c974-48a8-bcf6-8c8f3f239d52


...where's the embed

resources/js/chat/mention-completion-box-state.ts
```diff
@@ -0,0 +1,181 @@
...
+      el.selectionStart = el.selectionEnd = this.newCursorPosition;
+      this.newCursorPosition = undefined;
+    }
+  }
+
+  @action
+  private readonly fetchSuggestions = () => {
+    const query = this.query;
+
+    if (query === '' || query.length >= maxSearchLength || this.cache[query] != null || this.fetching.has(query)) {
```
notbakaneko, line 149: The minimum query length should be 2 since the user index ngram starts at 2, meaning single characters shouldn't match anything, as the minimum username length is 3

resources/js/chat/input-box.tsx
```diff
@@ -183,6 +189,36 @@ export default class InputBox extends React.Component<Props> {
...
+      if (e.shiftKey && this.allowMultiLine) return;
+
+      e.preventDefault();
+      if (!this.sendDisabled) {
+        this.startSendMessage();
+      }
+
+      return;
+    }
+  };
```
notbakaneko, line 230: missing extra newline

resources/js/chat/input-box.tsx
```diff
@@ -183,6 +189,36 @@ export default class InputBox extends React.Component<Props> {
...
+  private readonly handleKeyUp = (e: React.KeyboardEvent<HTMLTextAreaElement>) => {
+    if (['ArrowDown', 'ArrowUp', 'Escape', 'Enter', 'Tab'].includes(e.key)) {
```
notbakaneko: probably should extract this list or logic out; same with the one above

resources/js/chat/mention-completion-box-state.ts
```diff
@@ -0,0 +1,181 @@
...
+    switch (key) {
+      case 'ArrowDown':
+        return void this.shiftSelectedIndex(1);
+      case 'ArrowUp':
+        return void this.shiftSelectedIndex(-1);
+      case 'Escape':
+        return void (this.visible = false);
+      case 'Enter':
+      case 'Tab':
+        return void this.insertUsername(this.users?.[this.selectedIndex]);
```
notbakaneko: the setter is the only one that needs `void`?

resources/js/chat/mention-completion-box-state.ts
```diff
@@ -0,0 +1,181 @@
...
+      : modulo(this.selectedIndex + direction, size);
+  }
+
+  private updateQuery(newQuery: string) {
+    if (this.query !== newQuery) {
+      this.query = newQuery;
+      this.selectedIndex = 0;
+    }
+
+    this.debouncedFetchSuggestions();
```
notbakaneko, line 171: should it be triggering even if the query doesn't change? (since clicking in the box also triggers `updateQuery`)

resources/js/chat/mention-completion-box-state.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    if (this.query.length > 0) {
+      return this.cache[this.query];
+    }
+
+    const channel = this.selectedChannel;
+    if (channel == null) return [];
+
+    const users = new Map<number, UserSearchEntry>();
+    for (const message of channel.messages.slice().reverse()) {
```
notbakaneko: there's `toReversed()` now (also not sure if duplicating all messages to get the last 10 users is a good idea...)

There's also the odd case where a new message causes the list to change and switches the highlighted option while it's open.

resources/js/chat/input-box.tsx
notbakaneko: The list of keys being handled also probably shouldn't be repeated across 2 files

resources/lang/en/chat.php
```diff
@@ -5,6 +5,7 @@
 
 return [
     'loading_users' => 'loading users...',
+    'searching_users' => 'searching users...',
```
notbakaneko: `searching for users...`

notbakaneko (changes requested):

resources/js/chat/mention-completion-box-state.ts
nanaya: It does match users with single letter... thing, though

resources/js/chat/input-box.tsx
nanaya: I didn't even like the other one's split handling so it's going the other way instead

resources/js/chat/mention-completion-box-state.ts
nanaya: yeah you can complete midway (press tab/enter) and then click somewhere after the first autocomplete list and it should fetch the list if not yet

resources/js/chat/mention-completion-box-state.ts
```diff
@@ -0,0 +1,175 @@
...
+import { debounce } from 'lodash';
+import { action, computed, makeObservable, observable, runInAction, untracked } from 'mobx';
+import core from 'osu-core-singleton';
+import * as React from 'react';
+import { modulo } from 'utils/math';
+
+const maxSearchLength = 20;
+const searchRe = /(?<=^|\s|'|"|,|\.|\/)@([A-Za-z0-9-[\]_]*)/;
+
+export const mentionCompletionBoxNavigationKeys = ['ArrowDown', 'ArrowUp', 'Escape', 'Enter', 'Tab'];
```
notbakaneko: what's this export doing here 👀

notbakaneko (approved):

PR #12966, opened 2026-05-14
Skip existence checks for user replay stats

notbakaneko:
#12945 and #12946 take 2, with extra `null` handling

duplicates `fireTransformer` to insert our extra `null` handling


notbakaneko:
~~Causes other problem handling `null`s~~
shouldn't have assumed `transform(null) = null`

nanaya:
is the problem just the counting or fetching the rows as well?

notbakaneko:
it's both; the `exists` against large tables has large overheads in production.

Also, it doesn't make sense if the count doesn't match, although, if it's not rendered.... 🤔

nanaya:
the queries I've been seeing are all on the count though...? (also the matchmaking thing change here doesn't match the pr title?)

the replay count specifically can probably be fixed by doing `->limit(maxResults)->pluck('score_id')->count()` instead

notbakaneko:
we'll hold off on this for now

PR #12972, opened 2026-05-15
Skip checking existence of beatmapset

nanaya:
Scores on beatmap without beatmapset shouldn't ever happen. Such thing should be fixed when encountered instead of being ignored.

...I think?


notbakaneko (changes requested):
needs some conflict resolution now 👀

notbakaneko (approved):

PR #12977, opened 2026-05-16
Use simple count for first place scores

nanaya:
The count may or may not be actually accurate. also it'll be a bit slower in some cases but should still be faster than with the filters

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -876,6 +876,7 @@ class UsersController extends Controller
                 $query = $this
                     ->user
                     ->scoresFirst($this->mode, ScoreSearchParams::showLegacyForUser(\Auth::user()))
+                    ->default()
                     ->with(array_map(
                         fn ($include) => "score.{$include}",
                         ScoreTransformer::USER_PROFILE_INCLUDES_PRELOAD,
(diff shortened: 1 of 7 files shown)


notbakaneko (approved):

peppy:
For those playing along at home: this brings back counts which were removed in recent days from profile pages. Why were they removed? Because as we see more and more automated traffic (search engine, AI, and people just scraping profiles bypassing the api we've created), the load on servers for profile pages continues to increase non-proportional to active users.

The checks in these counts that beatmaps and scores still exist were adding large overheads (think 30% of total system DB load) due to subquery/join considerations. This was not due to code being "old" or "bad", but safeties in place to ensure the correct data is shown to a user in edge cases where the database was not in a pristine state.

We're tidying up the raw data in the background now, and reinvestigating cases where orphaned first-place entries (as one example) still exist, so we can relax the checks going forward. In theory.

For the record, there were only **58 incorrect entries** in the database out of 600,000. So don't believe what you read elsewhere; the counts are still accurate even when this change went live, and will continue to be. The edge cases are **edge cases, (like radiation in the atmosphere causing a failure) not "someone beat your score and it will take longer to update"**.

Yes, we probably could have / should have done a better job messaging this change, but 20/20 hindsight. Especially since it's not obvious to every developer on the team that people actually have reason to *care* about the number in the first place.

PR #12982, opened 2026-05-18
 Combine acl check queries 

nanaya:
One less queries.

- [x] depends on #12981


notbakaneko (dismissed):
queries seem correct 🤔

notbakaneko (approved):

PR #12986, opened 2026-05-19
Add support for LGA banners

notbakaneko:
Called it player supporter because that's what the data is keyed as.
Other clean ups can come later, including remapping the type mapping field name into something more generic...

- [x] change `custom_class` values on production to `player-supporter`

### DIFF
--- a/app/Libraries/Fulfillments/BannerFulfillment.php
+++ b/app/Libraries/Fulfillments/BannerFulfillment.php
@@ -16,6 +16,7 @@ abstract class BannerFulfillment extends OrderFulfiller
     const ALLOWED_TAGGED_NAMES = [
         'owc-supporter',
         'mwc7-supporter',
+        'player-supporter',
         'twc-supporter',
         'cwc-supporter',
         'mwc4-supporter',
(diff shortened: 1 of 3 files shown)


app/Libraries/Fulfillments/LgaSupporterFulfillment.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace App\Libraries\Fulfillments;
+
+use App\Models\Store\OrderItem;
+
+class LgaSupporterFulfillment extends BannerFulfillment
```
nanaya: I wonder if it should be called `TournamentBanner` (`tournament-banner` custom class, or `ProfileBanner`) and the data field `name` (or something generic)

and then (eventually?) get rid of the other banner classes

nanaya:
The product isn't currently enabled and it should be able to be updated accordingly? (for the field name)

peppy:
i disabled it to avoid errors, and yes, will make updates.

peppy (approved):

nanaya:
yeah unless there's another reason to not make the name generic, it should be done here once so no further updates will be needed (or only cleanups instead of adding more things).

notbakaneko:
The label of the dropdown is still derived from the name of the field in the type mappings
image
So unless all the labels are called the same thing, or a renaming mapping field is added to `type_mappings_json`, it's still going to be a separate class.

Also, I just remembered the `custom_class` is used as a convenience url `/store/products/{$class-name}`...

nanaya:
I guess it should be back to lga-supporter if we're keeping the redirect based on the custom class...? ಠ_ಠ

nanaya (approved):
@peppy feel free to merge if the redirect thing isn't needed (yet?) otherwise the custom class thing may need to be renamed back

peppy:
I think it's fine without for now, let's give it a try.

PR #12990, opened 2026-05-20
Preload difficulty ratings for user score api

nanaya:
The score transformer automatically sets `convert` field for beatmap transformer accordingly. The beatmap transformer contains `difficulty_rating` field which creates an extra query to beatmap difficulty table when the `convert` field is set to true.

<sub>and the value isn't even used for display in osu-web ha ha ha... &lt;_&lt;</sub>

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -19,6 +19,7 @@ use App\Models\Beatmap;
 use App\Models\BeatmapDiscussion;
 use App\Models\IpBan;
 use App\Models\Log;
+use App\Models\Solo\Score;
 use App\Models\User;
 use App\Models\UserAccountHistory;
 use App\Models\UserAchievement;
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/UsersController.php
```diff
@@ -869,6 +870,7 @@ private function getExtra($page, array $options, int $perPage = 10, int $offset
                     ScoreTransformer::USER_PROFILE_INCLUDES_PRELOAD
                 );
                 $userRelationColumn = 'user';
+                $collectionFn = fn ($scores) => Score::preloadDifficultyRatings($scores, $this->mode);
```
notbakaneko: this doesn't preload for `scoresBest` because it's already setting `$collection`

notbakaneko (changes requested):

notbakaneko (approved):

PR #12997, opened 2026-05-22
Fix rank value on user profile's ranked play

nanaya:
The `plays > 0` check is required because the rating doesn't start from 0 and thus it affects the rank value if not filtered out.


notbakaneko (approved):
makes sense...

PR #12998, opened 2026-05-22
Adjust ranked play's rank value

nanaya:
Show same rank for the same rating.

Seems to make more sense (and matches the profile page value) and confirmed by @smoogipoo


LiquidPL:
Needs to be updated on `MatchmakingUserStatsTransformer` as well.

nanaya:
This pr is matching* the value returned in that transformer

LiquidPL:
Ah, got it.

smoogipoo (approved):

notbakaneko (approved):

Issue #11585, opened 2024-10-22
Can't change the default game mode on mobile


felipemarins:
Basically, what the title says. It was reported on osu! dev discord and I can confirm the option doesn't seem to appear, at least on Firefox's responsive design mode. I don't know if this is intentional but I thought I would report it as I didn't find anything related by searching on the issue tracker.

Even if this isn't intentional, I think this would be a good addition because more and more people may start playing osu!(lazer) on their phones.

nanaya:
I think it was because there's nowhere to put the button.

peppy:
I'd say shoving it anywhere would be better than not having at all 😉

peppy:
Sounds like a good starting point. Having settings in more than one places allows for better discoverability.

nanaya:
Fixed in #13003 and the api in #13164.

PR #13003, opened 2026-05-24
Add option to change default game mode in account settings

gagahpangeran:
As suggested in https://github.com/ppy/osu-web/issues/11585#issuecomment-2430998578

I'm not sure which section this option should be placed.


resources/lang/en/accounts.php
```diff
@@ -138,6 +138,10 @@
         'tablet' => 'tablet',
         'title' => 'Playstyles',
         'touch' => 'touch',
+
+        'default_game_mode' => [
```
nanaya: `default_ruleset`

and the key can probably just this without additional `_`

nanaya (changes requested):

nanaya (approved):

PR #13006, opened 2026-05-27
Remove deprecated reflection visibility modifier

notbakaneko:
`setAccessible()` is deprecated, and non-public methods and properties can be invoked directly through reflection since php 8.1


nanaya (approved):
More like turned noop in 8.1 and deprecated in 8.5 (it's not actually deprecated yet in 8.4)

PR #13008, opened 2026-05-27
Send notifications to participants of a Beatmap Discussion on replies

### LINKED ISSUE #12885: Receive notifications for replies to modding discussions

notbakaneko:
Sends notifications to all participants of problem or suggestion type Beatmap Discussions, not just the starter.
This is based on the reason the reasoning that if you are replying to a discussion, you are probably involved in it (and doesn't involve adding a watch toggle to every discussion).

~~Also made some of the tests more strict and give more information when the expected receiver ids don't match (because the `Event` and `Queue` assertions don't tell you much which makes find it out annoying...)~~ separated into #13044

closes #12885

- [x] #13057

### DIFF
--- a/database/factories/BeatmapDiscussionFactory.php
+++ b/database/factories/BeatmapDiscussionFactory.php
@@ -7,6 +7,8 @@ namespace Database\Factories;
 
 use App\Models\BeatmapDiscussion;
 use App\Models\BeatmapDiscussionPost;
+use App\Models\Beatmapset;
+use App\Models\User;
 
 class BeatmapDiscussionFactory extends Factory
 {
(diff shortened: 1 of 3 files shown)


app/Jobs/Notifications/BeatmapsetDiscussionPostNotification.php
```diff
@@ -49,8 +49,11 @@ public function getListeningUserIds(): array
         $userIds = $this->beatmapsetDiscussionPost->beatmapset->watches()->pluck('user_id');
 
         $discussion = $this->beatmapsetDiscussionPost->beatmapDiscussion;
-        if ($discussion->canBeResolved() && $discussion->user_id !== null) {
-            $userIds->push($discussion->user_id);
+        if ($discussion->canBeResolved()) {
+            if ($discussion->user_id !== null) {
+                $userIds->push($discussion->user_id);
+            }
```
notbakaneko: I think this might not be necessary since discussions should normally have a post, but the tests currently don't always have the starting post...

nanaya:
I wonder if the #13017 should be done first (splitting manual and automatic watch options)

notbakaneko:
Maybe.... 🤔

notbakaneko:
doing some rejigging

database/factories/BeatmapDiscussionFactory.php
```diff
@@ -63,4 +64,12 @@ public function timeline()
     {
         return $this->state(static::DEFAULTS['timeline']);
     }
+
+    public function withStarter()
```
nanaya: would it make more sense to always have this by default?

nanaya (approved):

PR #13018, opened 2026-05-28
Update Shopify SDK

notbakaneko:
Bump the API version to the latest and removed the migration command which shouldn't work anymore since `Checkout` was removed from their API.

This should actually be `^6.1.1` but that requires `firebase/php-jwt` to be updated which requires `passport` to be updated, which updates other things, etc...
But `6.1.1` will work fine.


nanaya (changes requested):
conflicts 😴

nanaya (commented):
does the api version in `resources/views/store/layout.blade.php` also need an update? 🤔

notbakaneko:
~~so that's why nothing exploded~~

nanaya (approved):

PR #13019, opened 2026-05-28
Remove unused dev dependencies

notbakaneko:
I don't think we use them directly so it doesn't make sense to have them explicitly listed as dependencies?


nanaya (changes requested):
where's the lockfile update ?_? (dom-crawler isn't required anywhere)

nanaya (approved):

PR #13023, opened 2026-05-29
Add index for Passport's findValidToken

notbakaneko:
this gets called via Passport during OAuth auth_code requests
```php
    public function findValidToken($user, $client)
    {
        return $client->tokens()
                      ->whereUserId($user->getAuthIdentifier())
                      ->where('revoked', 0)
                      ->where('expires_at', '>', Carbon::now())
                      ->latest('expires_at')
                      ->first();
    }
```


notbakaneko:
Actually, I don't think the descending index is even needed here? 🤔

nanaya (approved):

PR #13027, opened 2026-06-01
Fix beatmap description error on long title/artist

### LINKED ISSUE #13026: "Topic title exceeded maximum length - can only up to 100 characters" when try to edit beatmapset description
From : https://osu.ppy.sh/community/forums/topics/2210899

---
Problem details:
I just uploaded a new map but i can't even edit my description fields. It always said "Topic title exceeded maximum length - can only up to 100 characters"

Video or screenshot showing the problem:

---

The beatmapset page : [beatmapsets/2562136#osu/5695859](https://osu.ppy.sh/beatmapsets/2562136#osu/5695859)

Because the beatmapset page description is linked with forum post in subforum [Pending Beatmaps](https://osu.ppy.sh/community/forums/6), it automatically creates a new forum topic with title '{artist} - {title}'. The generated title of this beatmapset would be "Kaguya (CV: Yuko Natsuyoshi), Yachiyo Runami (CV: Saori Hayami) - ray (Cosmic Princess Kaguya! Version)" which exceed forum post title limit of 100 characters.

nanaya:
Resolves #13026

It's probably better in the topic itself but that'd involve checking the topic type... which isn't possible when it's created.


gagahpangeran:
Instead of creating the topic with 100 characters title and then updating it if the title is more than 100 characters. Why not just trim it once entirely with maximum 100 characters? :thinking: ([something like this](https://osu.ppy.sh/community/forums/topics/2150055))

People won't visit the forum topic itself, so I guess people won't care either if the title is trimmed.

notbakaneko (approved):

PR #13034, opened 2026-06-03
Fix user modding post listing error on deleted beatmapset

nanaya:
Either this or query with the `has` thing.

### DIFF
--- a/app/Models/DeletedBeatmapset.php
+++ b/app/Models/DeletedBeatmapset.php
@@ -1,0 +1,24 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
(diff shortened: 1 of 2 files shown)


notbakaneko (changes requested):
`BeatmapDiscussionTransformer::includeBeatmapset` should also be updated to handle it

nanaya:
is there any endpoint where it can happen?

notbakaneko:
`/users/{user}/modding`?

nanaya:
I just noticed the beatmapset relation in discussion model already includes deleted beatmapset and the post with error is pointing to a completely nonexistent beatmapset (not even deleted/inactive) which I don't think should happen 🤔

nanaya:
Closing this for now as it's not a good fix. May revisit this later.

PR #13035, opened 2026-06-03
Filter incoming messages

nanaya:
As requested...? 🤷 

(kinda tested)


notbakaneko (approved):
It can cause places with a visible message counter to have messages rejected even when under the limit, but that's a different problem

PR #13038, opened 2026-06-05
Fix query for getting discussion's starting post

nanaya:
I think it's been accidentally working from mysql returning the lowest post id of the discussion

### DIFF
--- a/app/Models/BeatmapDiscussion.php
+++ b/app/Models/BeatmapDiscussion.php
@@ -195,7 +195,7 @@ class BeatmapDiscussion extends Model
         return $this->hasOne(BeatmapDiscussionPost::class)->whereNotExists(function ($query) {
             $table = (new BeatmapDiscussionPost())->getTable();
 
-            $query->from(DB::raw("{$table} d"))
+            $query->from($table, 'd')
                 ->whereColumn('d.beatmap_discussion_id', "{$table}.beatmap_discussion_id")
                 ->whereColumn('d.id', '<', "{$table}.id");
         });
--- a/app/Models/BeatmapDiscussion.php
+++ b/app/Models/BeatmapDiscussion.php
@@ -195,10 +195,9 @@ class BeatmapDiscussion extends Model
         return $this->hasOne(BeatmapDiscussionPost::class)->whereNotExists(function ($query) {
             $table = (new BeatmapDiscussionPost())->getTable();
 
-            $query->selectRaw(1)
-                ->from(DB::raw("{$table} d"))
-                ->where('beatmap_discussion_id', $this->beatmap_discussion_id)
-                ->whereRaw("d.id < {$table}.id");
+            $query->from(DB::raw("{$table} d"))
+                ->whereColumn('d.beatmap_discussion_id', "{$table}.beatmap_discussion_id")
+                ->whereColumn('d.id', '<', "{$table}.id");
         });
     }
 


app/Models/BeatmapDiscussion.php
```diff
@@ -195,10 +195,9 @@ public function startingPost()
         return $this->hasOne(BeatmapDiscussionPost::class)->whereNotExists(function ($query) {
             $table = (new BeatmapDiscussionPost())->getTable();
 
-            $query->selectRaw(1)
-                ->from(DB::raw("{$table} d"))
-                ->where('beatmap_discussion_id', $this->beatmap_discussion_id)
-                ->whereRaw("d.id < {$table}.id");
+            $query->from($table, 'd')
+                ->whereColumn('d.beatmap_discussion_id', "{$table}.beatmap_discussion_id")
+                ->whereColumn('d.id', '<', "{$table}.id");
```
notbakaneko, line 200: can probably use `qualifyColumn` instead of string interpolation

nanaya: that doesn't work because it's a Query Builder, not Eloquent Builder

notbakaneko:
```php
            $model = new BeatmapDiscussionPost();

            $query->from($model->getTable(), 'd')
                ->whereColumn('d.beatmap_discussion_id', $model->qualifyColumn('beatmap_discussion_id'))
                ->whereColumn('d.id', '<', $model->qualifyColumn('id'));
```
?
Not like it really makes much difference

nanaya: The "correct" form gets ugly very quickly

```php
        return $this->hasOne(BeatmapDiscussionPost::class)->where(fn ($query) => $query->whereNotExists(function ($q) use ($query) {
            $subQuery = (new BeatmapDiscussionPost())->query();
            $subQuery->from($subQuery->from, 'd');

            $q->from($subQuery->from)
                ->whereColumn($subQuery->qualifyColumn('beatmap_discussion_id'), $query->qualifyColumn("beatmap_discussion_id"))
                ->whereColumn($subQuery->qualifyColumn('id'), '<', $query->qualifyColumn("id"));
        }));

```

nanaya: wait, it looks like `qualifyColumn` just get the underlying model's table name and prefix it to the column name... ಠ_ಠ

it doesn't consider the query builder at all.

nanaya: yeah the whole thing is a silly farce. It's handy if the table name isn't available anywhere but here the only use of the new class instance is for the table name.

I guess this also works if you think it's better...

```php
        $postClass = BeatmapDiscussionPost::class;

        return $this->hasOne($postClass)->where(fn ($query) =>
            $query->whereNotExists($postClass::query()->from($query->getModel()->getTable(), 'd')
                ->whereColumn('beatmap_discussion_id', $query->qualifyColumn('beatmap_discussion_id'))
                ->whereColumn('id', '<', $query->qualifyColumn('id'))));
```

notbakaneko: That doesn't even use the alias in the final query ಠ_ಠ

Doesn't that mean this would have worked...?
```php
        return $this->hasOne(BeatmapDiscussionPost::class)->whereNotExists(fn ($query) =>
            $query->from(new BeatmapDiscussionPost()->getTable(), 'd')
                ->whereColumn('beatmap_discussion_id', 'd.beatmap_discussion_id')
                ->whereColumn('id', '<', 'd.id')
        );
```

nanaya: it needs to refer the outer table `(inner.id < outer.id)`. without the table name it resolves to the closest table first so in that case (`where not exists (select * from table t where id < t.id)`) both will refer to the same value

...I think

notbakaneko (approved):
let's just go with this for now and fafo separately 👀

Issue #13017, opened 2026-05-28
Separate notification setting for [others replying to my mods]


cihe13375:
Starting from an unknown moment, I started to receive notifications on others replying to my mods, even if I did not watch that map (and I can't even "unwatch" them). I don't know which setting in "delivery options" is controlling that, but I assume it's "beatmap modding" (I only have emails enabled for beatmap modding and private chat, and I am getting emails from them).

Even though notifications on [others replying to my mods] can be helpful sometimes, it's fundamentally different / not nearly important as posts on one's own maps or maps someone watched. For the latter I am opting in at my will, while for the former it's sort of imposed by the system (and I can't even easily opt out). Also in 99% of cases I need to response to the mods on my maps, but in 99% of cases I don't need to do anything for others replying to my mods (usually it's just like "ok" or "sorry I disagree").

Personally, I want to enable email notifications for maps I watched / I owned, but definitely not for others replying to my mods.

Serizawa-Haruki:
Correct me if I'm wrong, but can't you disable the beatmap modding notification toggle and only watch the maps you're interested in (your own maps)?

PR #13044, opened 2026-06-09
Update some Beatmapset notification tests

notbakaneko:
Tests only update part of #13008 while I do more checking for splitting the notification options for #13017
- move dataproviders to top
- switch to attributes
- stricter receiverIds checking


tests/TestCase.php
```diff
@@ -309,9 +310,22 @@ protected function expectExceptionCallable(callable $callable, ?string $exceptio
         }
     }
 
-    protected function inReceivers(Model $model, NewPrivateNotificationEvent|BroadcastNotificationBase $obj): bool
-    {
-        return in_array($model->getKey(), $obj->getReceiverIds(), true);
+    protected function receiversInclude(
+        NewPrivateNotificationEvent|BroadcastNotificationBase $obj,
+        array|Collection|Model $includes = [],
+        array|Collection|Model $excludes = []
```
nanaya: is there any point in checking this `$excludes` when the function is already checking that `$obj` receiver ids are exactly the same as `$includes`'? Seems rather redundant.

nanaya: and might as well `sort($includeIds); sort($receiverIds); $this->assertSame($includeIds, $receiverIds);`...

notbakaneko: It's for checking if something is specifically not included but we don't care about the rest; although it should probably be a separate function atp 🤔  

I originally wrote this to handle both and then discovered `assert(Not)Pushed/Dispatched` didn't quite work the way expected :|

nanaya: > It's for checking if something is specifically not included but we don't care about the rest; typically they're not both going to be used at the same time, although it should probably be a separate function atp 🤔

that won't work here though if there's anything in the receiver ids with empty includes passed?

nanaya: also, the `$includes` name is a bit misleading because currently it needs to be exactly the same as the receiver ids. Might as well call it `$expected`

nanaya (dismissed):

nanaya (approved):

PR #13045, opened 2026-06-09
Force unfollow after blocking user

### LINKED ISSUE #13022: Make blocked users unfollow you when blocking them
as title

Image

Explanation: most social media do this, and there's no point in following someone who has blocked others

notbakaneko:
closes #13022

and add tests for the controller which are mostly a copy of the existing `FriendsControllerTest` but flipped

### DIFF
--- a/app/Http/Controllers/BlocksController.php
+++ b/app/Http/Controllers/BlocksController.php
@@ -52,13 +52,12 @@ class BlocksController extends Controller
         $currentUserId = $currentUser->getKey();
         $relationQuery = $currentUser->relations()->where('zebra_id', $targetId);
         \DB::transaction(function () use ($currentUserId, $relationQuery, $targetId) {
-            if (
-                UserRelation::where([
+            $deleted = UserRelation::where([
                 'friend' => true,
                 'user_id' => $targetId,
                 'zebra_id' => $currentUserId,
-                ])->delete() > 0
-            ) {
+            ])->delete();
+            if ($deleted > 0) {
                 dispatch(new UpdateUserFollowerCountCache($currentUserId))->afterCommit();
             }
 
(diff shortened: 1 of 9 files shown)


nanaya:
Should the blocked user be prevented from following again?

notbakaneko:
...somehow I feel like loading every user's blocklist for the user cards is not a great idea...
or even adding a subquery to include the data just to handle an operation that is more likely to not happen...

nanaya:
The button is still enabled in user card. Although maybe it's fine.

Or alternatively create a `blockedBy` relation/cache for user model that can be used by the current user and have the field included for cards as well.

app/Http/Controllers/BlocksController.php
```diff
@@ -49,24 +49,36 @@ public function store()
...
-            ]);
-        }
+        \DB::transaction(function () use ($currentUserId, $relationQuery, $targetId) {
+            if (
+                UserRelation::where([
+                'friend' => true,
+                'user_id' => $targetId,
+                'zebra_id' => $currentUserId,
+                ])->delete() > 0
+            ) {
```
nanaya: that's kinda hard to read. Assign delete count result to a variable and then compare that variable in the check?

also, wrong indentation?

notbakaneko:
I'm not sure it's worth the overhead of preloading, either as a separate eagerload or inline subquery, for the blocks and adding them to every response with a list of user cards when its more not likely the blocking is not going to be happening there.

nanaya (changes requested):

app/Http/Controllers/BlocksController.php
notbakaneko: funny this is what phpcbf corrected it into ಠ_ಠ

nanaya (approved):

peppy:
Will run a batch fix for this post deploy.

PR #13049, opened 2026-06-11
Don't update Beatmapset watches for problem on qualified beatmap notification

notbakaneko:
Not really noticeable since the qualified problem notification is always queued with a new post notification, but the qualified problem notification isn't sent to watchers so it shouldn't be updating the `last_notified` time.


app/Jobs/Notifications/BeatmapsetDiscussionPostNotification.php
```diff
@@ -44,27 +44,11 @@ public function getDetails(): array
...
-
-        $discussion = $this->beatmapsetDiscussionPost->beatmapDiscussion;
-        if ($discussion->canBeResolved() && $discussion->user_id !== null) {
-            $userIds->push($discussion->user_id);
-        }
-
-        return $userIds->all();
-    }
+    #[\Override]
+    abstract public function getListeningUserIds(): array;
```
notbakaneko, line 48: redeclaring as `abstract` so subclasses have to implement it (also weird if handling `watches` is split across classes)

nanaya (approved):

PR #13054, opened 2026-06-12
Replace homepage livestream embed with static preview

Hiviexd:
current implementation didn't fit internal guidelines (twitch forces ads) so this ended up getting unused..

using static previews just like the livestreams page should address that.

image

### DIFF
--- a/resources/css/bem/user-home-livestream.less
+++ b/resources/css/bem/user-home-livestream.less
@@ -4,24 +4,21 @@
 .user-home-livestream {
   .default-border-radius();
   .default-box-shadow();
+  display: block;
   width: 100%;
   aspect-ratio: 16 / 9;
   position: relative;
   overflow: hidden;
   background-color: @osu-colour-b4;
+  background-size: cover;
   box-sizing: border-box;
+
+  &:link {
+    text-decoration: none;
+  }
 
   &:hover {
     .thicker-box-shadow();
-  }
-
-  &__content {
-    .full-size();
-    background-size: cover;
-
-    &:link {
-      text-decoration: none;
-    }
   }
 
   &__info {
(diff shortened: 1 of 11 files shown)


nanaya:
fwiw the one on dashboard doesn't autoplay

Hiviexd:
iirc @peppy still had issue with that? issue was with ads being forced whe users start the stream from the dashboard.

convo: https://discord.com/channels/90072389919997952/1083320217259749376/1468074549307506709

nanaya:
I think it should use the previous live placeholder design instead (scaled down. see this for reference https://github.com/ppy/osu-web/pull/12679/files)

resources/css/bem/user-home-livestream.less
```diff
@@ -10,4 +10,49 @@
...
+  }
+
+  &__content {
+    .full-size();
+    background-size: cover;
+
+    &:link {
+      text-decoration: none;
+    }
+  }
```
nanaya: this is rather redundant. it can be set directly on the parent itself.

resources/css/variables.less
```diff
@@ -192,6 +192,8 @@
 @font-size--livestream-featured-detail: 15px;
 @font-size--livestream-item-name: 15px;
 @font-size--livestream-item-detail: 10px;
+@font-size--user-home-livestream-name: 20px;
+@font-size--user-home-livestream-detail: 14px;
```
nanaya: should use the existing generic font sizes

resources/css/bem/user-home-livestream.less
```diff
@@ -10,4 +10,49 @@
...
+    color: #fff;
+
+    &--detail {
+      margin-top: 6px;
+      font-size: @font-size--user-home-livestream-detail;
+
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: nowrap;
+    }
```
nanaya: use grid's `gap` instead of margin and set the default size and overflow on `__text` itself. And use existing `.u-ellipisis-overflow()`.

resources/css/bem/user-home-livestream.less
```diff
@@ -10,4 +10,49 @@
...
+    width: 100%;
+    bottom: 0;
+    left: 0;
+    padding: 18px 24px;
+    background-image: linear-gradient(transparent, #000);
+  }
+
+  &__text {
+    margin: 0;
+    color: #fff;
```
nanaya: use `hsl(var(--hsl-c2))`

nanaya (changes requested):

nanaya (approved):

PR #13057, opened 2026-06-12
Add option for receiving notifications for replies to Beatmap Discussions

### LINKED ISSUE #13017: Separate notification setting for [others replying to my mods]

notbakaneko:
Went for the simpler option of just enable/disable reply notifications in the end.

closes #13017


tests/Libraries/BeatmapsetDiscussion/ReplyTest.php
```diff
@@ -154,6 +154,29 @@ public function testReplyQueuesNotificationToStarter(string $messageType, bool $
...
+            'details' => [UserNotificationOption::BEATMAPSET_DISCUSSION_REPLY => false],
+        ]);
+
+        $discussion = BeatmapDiscussion::factory()->general()->create([
+            'beatmapset_id' => $this->beatmapsetFactory(),
+            'message_type' => 'problem',
+            'user_id' => $starter,
+        ]);
+
+        new Reply($user, $discussion, static::TEST_MESSAGE)->handle();
```
nanaya, line 174: I wonder if it's worth checking that the reply is posted successfully

nanaya (approved):

PR #13059, opened 2026-06-13
Color hits based on type in score leaderboard

Hiviexd:
Saw this on a popular userscript so I thought why don't we have this implemented on our side.

Makes the table more readable too IMO

image

<details>
<summary>osu!taiko</summary>
image

</details>
<details>
<summary>osu!catch</summary>
image
</details>
<details>
<summary>osu!mania</summary>
image
</details>


resources/css/bem/beatmap-scoreboard-table.less
```diff
@@ -100,7 +101,9 @@
 
     &:hover,
     &--menu-active {
-      --zero-color: #fff;
+      .@{_top}__cell-content--zero {
+        opacity: 1;
```
nanaya: change it to `--zero-opacity` instead

resources/css/bem/beatmap-scoreboard-table.less
```diff
@@ -194,8 +197,11 @@
       white-space: nowrap;
     }
 
+    .hit-stat-colours();
+
     &--zero {
-      --default-color: var(--zero-color);
+      opacity: 0.4;
+      .fade-element(@hover-transition-duration, ease-in-out, opacity);
```
nanaya: I don't think the original effect was intentional. It doesn't match the background colour change either.

resources/js/utils/score-helper.ts
```diff
@@ -93,6 +104,22 @@ interface ScoreStatistic {
 
 const labelMiss = trans('beatmapsets.show.scoreboard.headers.miss');
 
+const statisticColourByAttribute: Partial<Record<ScoreStatisticsAttribute, ScoreStatisticColour>> = {
```
nanaya: just make the css class/variables match the snake case instead of adding another mapping

resources/css/functions.less
```diff
@@ -351,3 +351,23 @@
...
+    large-tick-hit: @beatmap-score--hsl-hit-large-tick-hit;
+    small-tick-miss: @beatmap-score--hsl-hit-small-tick-miss;
+  };
+
+  each(@colours, {
+    &--@{key} {
+      --default-color: hsl(@value);
+      color: hsl(@value);
+    }
+  });
```
nanaya: make it accept `@rules` parameter and have the caller decide what its content should be

```
// here
&--hit-@{key} {
  @rules();
}

// call example
.hit-stat-colours({
  --default-color: hsl(@value);
});
```

nanaya (changes requested):

resources/css/bem/beatmap-scoreboard-table.less
```diff
@@ -194,8 +198,13 @@
       white-space: nowrap;
     }
 
+    .hit-stat-colours({
+      --default-color: hsl(@value);
+      color: hsl(@value);
```
nanaya: not needed?

resources/css/variables.less
```diff
@@ -260,6 +260,16 @@
...
 
+// cross-reference: https://github.com/ppy/osu/blob/9e5029a5e7b902d984ec0184e4f04b5b0c22079a/osu.Game/Graphics/OsuColour.cs#L114-L145
+@beatmap-score--hsl-hit-perfect: var(--hsl-blue-1);
+@beatmap-score--hsl-hit-great: var(--hsl-blue-2);
+@beatmap-score--hsl-hit-good: var(--hsl-green-1);
+@beatmap-score--hsl-hit-ok: var(--hsl-lime-2);
+@beatmap-score--hsl-hit-meh: var(--hsl-orange-1);
+@beatmap-score--hsl-hit-miss: var(--hsl-red-1);
+@beatmap-score--hsl-hit-large-tick-hit: var(--hsl-blue-2);
+@beatmap-score--hsl-hit-small-tick-miss: var(--hsl-f1);
```
nanaya: define directly in the single place they are used

resources/js/beatmapsets-show/scoreboard/table.tsx
```diff
@@ -62,7 +62,7 @@ export default class Table extends React.Component<Props> {
                   .map((stat) => (
                     <th
                       key={stat.label.short}
-                      className={classWithModifiers(`${bn}__header`, ['hitstat'])}
+                      className={classWithModifiers(`${bn}__header`, ['hitstat', hitStatisticModifier(stat.attributes[0])])}
```
nanaya: I wonder if the helper is worth it compared to just `hit-${attr}` which is shorter...

resources/js/beatmapsets-show/scoreboard/top-card.tsx
```diff
@@ -163,10 +163,10 @@ export default class TopCard extends React.PureComponent<Props> {
             <div className='beatmap-score-top__stats beatmap-score-top__stats--wrappable'>
               {calculateStatisticsFor(this.props.score, 'leaderboard').map((attr) => (
                 <div key={attr.label.short} className='beatmap-score-top__stat'>
-                  <div className='beatmap-score-top__stat-header'>
+                  <div className={classWithModifiers('beatmap-score-top__stat-header', hitStatisticModifier(attr.attribute))}>
                     {attr.label.short}
                   </div>
-                  <div className='beatmap-score-top__stat-value beatmap-score-top__stat-value--smaller'>
+                  <div className={classWithModifiers('beatmap-score-top__stat-value', 'smaller', hitStatisticModifier(attr.attribute))}>
```
nanaya: the one in `/scores/:id` page doesn't have colored values?

nanaya (changes requested):

resources/js/beatmapsets-show/scoreboard/top-card.tsx
Hiviexd: i did add coloring to `/scores/:id` via [`9f450ca` (this PR)](https://github.com/ppy/osu-web/pull/13059/changes/9f450ca030fd4836c7f664b99f8e729088814d7f) tho

image

or is that not what you mean?

nanaya: the values there are still white whereas for this card both are colored.

Hiviexd: Was intentional as it follows lazer score page.

image

I don't feel strongly about this given lazer's score page will be eventually redesigned so I'm fine if we color the numbers, your call.

As of the score table top card, it feels nicer to have it visually consistent with the leaderboard, hence why the numbers are colored there.

resources/js/beatmapsets-show/scoreboard/table.tsx
```diff
@@ -62,7 +62,7 @@ export default class Table extends React.Component<Props> {
                   .map((stat) => (
                     <th
                       key={stat.label.short}
-                      className={classWithModifiers(`${bn}__header`, ['hitstat'])}
+                      className={classWithModifiers(`${bn}__header`, ['hitstat', `hit-${stat.attributes[0]}`])}
```
nanaya, line 65: use helper or not, pick one... (including for all the others using this modifier)

resources/js/beatmapsets-show/scoreboard/table-row.tsx
```diff
@@ -139,7 +139,7 @@ export default class ScoreboardTableRow extends React.Component<Props> {
           <TdLink
             key={stat.label.short}
             href={this.scoreUrl}
-            modifiers={stat.value === 0 ? 'zero' : null}
+            modifiers={mergeModifiers(hitStatisticModifier(stat.attribute), stat.value === 0 ? 'zero' : null)}
```
nanaya: can use `{ zero: stat.value === 0 }` for the other one (it's done the current way because it's the only modifier)

nanaya (changes requested):

resources/js/beatmapsets-show/scoreboard/table.tsx
Hiviexd: will remove helper since it's a basic string concat

nanaya (approved):

ExHix:
It might be late to bring it up, but I recently found out that there are still many players including myself in community not really like or adopt such visual change, voice are that these colors are irruptive when the web text colors are generally white and blue and too many color is causing extra mind burden when viewing score, and it doesn't fit neither stable or lazer score in-game design style currently.
As the PR said it is inspired from web extension, I think these visual change should be treated as opinional and need to be able to get toggled in web settings.

PR #13060, opened 2026-06-13
Paginate contests listing

### LINKED ISSUE #1117: Contest page should be paginated
The index page currently lists all contests. This will be 'less than ideal' as more contests are run.

Hiviexd:
We have 285 contests as of today and it's becoming too much to have in one page.

edit: closes https://github.com/ppy/osu-web/issues/1117

### DIFF
--- a/app/Http/Controllers/ContestsController.php
+++ b/app/Http/Controllers/ContestsController.php
@@ -12,6 +12,8 @@ use Auth;
 
 class ContestsController extends Controller
 {
+    private const PER_PAGE = 25;
+
     public function index()
     {
         $contests = Contest::orderBy('id', 'desc');
@@ -21,7 +23,7 @@ class ContestsController extends Controller
         }
 
         return ext_view('contests.index', [
-            'contests' => $contests->get(),
+            'contests' => $contests->paginate(static::PER_PAGE),
         ]);
     }
 
--- a/app/Http/Controllers/ContestsController.php
+++ b/app/Http/Controllers/ContestsController.php
@@ -12,8 +12,6 @@ use Auth;
 
 class ContestsController extends Controller
 {
-    private const PER_PAGE = 25;
-
     public function index()
     {
         $contests = Contest::orderBy('id', 'desc');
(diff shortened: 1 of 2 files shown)


Joehuu:
This will close https://github.com/ppy/osu-web/issues/1117.

app/Http/Controllers/ContestsController.php
```diff
@@ -12,6 +12,8 @@
 
 class ContestsController extends Controller
 {
+    private const PER_PAGE = 25;
```
nanaya: why is it using custom per page count?

app/Http/Controllers/ContestsController.php
```diff
@@ -21,7 +23,7 @@ public function index()
         }
 
         return ext_view('contests.index', [
-            'contests' => $contests->get(),
+            'contests' => $contests->paginate(static::PER_PAGE),
```
nanaya: I wonder if year based filtering instead would be better but I guess this is simpler for now.

nanaya (changes requested):

app/Http/Controllers/ContestsController.php
Hiviexd: default (50) felt like a bit too much scrolling.

I don't mind using the default page size if you prefer, I don't have strong feelings towards either.

nanaya: I'm not quite sure what "too much scrolling" means...

Hiviexd: referring to how much scroll is needed to reach the bottom, it's purely subjective so I don't mind changing back to the default if you want.

nanaya: yeah just make it the default. If anything it makes it harder for anyone searching for something specific.

resources/views/contests/index.blade.php
```diff
@@ -34,5 +34,7 @@
                 </a>
             @endforeach
         </div>
+
+        @include('objects._pagination_v2', ['object' => $contests])
```
nanaya: I think it looks better aligned inside the contest-list div

inside


current

nanaya (changes requested):

nanaya (approved):

PR #13061, opened 2026-06-13
Replace judge score slider with numeric input

### LINKED ISSUE #11397: Add ability to enter score by hand on contest judging
Probably a click-to-edit on the score display under the slider would be a good way for by-hand filling the values (that can be more handy than using the sliders).

Hiviexd:
most common complaint about judging is the score slider input. A numeric input makes a ton more sense here and feels much more intuitive to use.

image

Resolves https://github.com/ppy/osu-web/issues/11397

### DIFF
--- a/resources/css/bem/contest-judge-entry.less
+++ b/resources/css/bem/contest-judge-entry.less
@@ -12,13 +12,12 @@
     display: flex;
     flex-direction: column;
     gap: 10px;
-    max-width: 500px;
+    max-width: 250px;
   }
 
   &__category-label {
-    align-items: center;
+    align-items: baseline;
     display: flex;
-    font-weight: bold;
     gap: 5px;
   }
 
--- a/resources/css/bem/contest-judge-entry.less
+++ b/resources/css/bem/contest-judge-entry.less
@@ -15,54 +15,11 @@
     max-width: 500px;
   }
 
-  &__category {
+  &__category-label {
+    align-items: center;
     display: flex;
-    flex-direction: column;
-    gap: 10px;
-  }
-
-  &__label {
-    display: flex;
-    gap: 5px;
-    font-size: @font-size--normal;
     font-weight: bold;
-  }
-
-  &__slider {
-    -webkit-appearance: none;
-    appearance: none;
-    background: transparent;
-    cursor: pointer;
-
-    &::-webkit-slider-runnable-track {
-      background: hsla(var(--hsl-pink-1), 50%);
-      height: 8px;
-      border-radius: @border-radius-base;
-    }
-
-    &::-moz-range-track {
-      background: hsla(var(--hsl-pink-1), 50%);
-      height: 8px;
-      border-radius: @border-radius-base;
-    }
-
-    &::-webkit-slider-thumb {
-      -webkit-appearance: none;
-      appearance: none;
-      margin-top: -4px; // initially slightly misaligned
-      background-color: hsl(var(--hsl-pink-1));
-      width: 16px;
-      height: 16px;
-      border-radius: 9999px;
-    }
-
-    &::-moz-range-thumb {
-      border: none;
-      border-radius: 9999px;
-      background-color: hsl(var(--hsl-pink-1));
-      width: 16px;
-      height: 16px;
-    }
+    gap: 5px;
   }
 
   &__title {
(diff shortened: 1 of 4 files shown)


resources/js/contest-judge/entry.tsx
```diff
@@ -111,45 +112,49 @@ export default class Entry extends React.Component<Props> {
...
           data-category-id={category.id}
           disabled={!this.store.canJudge}
+          id={inputId}
+          inputMode='numeric'
           max={category.max_value}
-          onChange={this.handleRangeInputChange}
-          type='range'
-          value={currentScore?.value ?? 0}
+          onChange={this.handleScoreInputChange}
+          value={currentScore != null ? String(currentScore.value) : ''}
```
notbakaneko: just `currentScore?.value ?? ''` is enough

resources/js/contest-judge/entry.tsx
```diff
@@ -111,45 +112,49 @@ export default class Entry extends React.Component<Props> {
...
-    const value = Number(e.currentTarget.value);
+    const value = clamp(Number(e.currentTarget.value), 0, Number(e.currentTarget.max));
 
     const score = { contest_scoring_category_id: categoryId, value };
     this.currentVote.scores.set(categoryId, score);
   };
 
   private readonly renderCategory = (category: ContestScoringCategoryJson) => {
     const currentScore = this.currentVote.scores.get(category.id);
+    const inputId = `contest-judge-score-${category.id}`;
```
notbakaneko: also needs the entry id to be unique

resources/js/contest-judge/entry.tsx
```diff
@@ -111,45 +112,49 @@ export default class Entry extends React.Component<Props> {
   };
 
   @action
-  private readonly handleRangeInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+  private readonly handleScoreInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     const categoryId = Number(e.currentTarget.getAttribute('data-category-id'));
-    const value = Number(e.currentTarget.value);
+    const value = clamp(Number(e.currentTarget.value), 0, Number(e.currentTarget.max));
```
notbakaneko: this needs to check for `Number.isFinite` since `Number('a')` is `NaN`. Also should reject/ignore non-numeric inputs, something like
```
let value = Number(e.currentTarget.value);
if (!Number.isInteger(value)) return;
value = clamp(value, 0, Number(e.currentTarget.max));
```

resources/css/bem/contest-judge-entry.less
```diff
@@ -15,54 +15,11 @@
...
     display: flex;
-    flex-direction: column;
-    gap: 10px;
-  }
-
-  &__label {
-    display: flex;
-    gap: 5px;
-    font-size: @font-size--normal;
     font-weight: bold;
```
notbakaneko: I don't think we need `bold` inside the text container

resources/css/bem/contest-judge-entry.less
```diff
@@ -15,54 +15,11 @@
     max-width: 500px;
   }
 
-  &__category {
+  &__category-label {
+    align-items: center;
```
notbakaneko: `baseline` looks more aligned here? 🤔

resources/css/bem/input-container.less
```diff
@@ -123,4 +127,13 @@
     width: 100%;
     pointer-events: var(--label-pointer-events);
   }
+
+  &__suffix {
+    bottom: var(--input-padding-base);
+    color: hsl(var(--hsl-c2));
+    line-height: 1;
```
notbakaneko: This seems to make it look misaligned with the input text.
Also, the input text should probably be right-aligned, and the 500px input box looks too big as well...

resources/js/contest-judge/entry.tsx
```diff
@@ -111,45 +112,49 @@ export default class Entry extends React.Component<Props> {
...
 
     return (
-      <div key={category.id} className='contest-judge-entry__category'>
-        <div className='contest-judge-entry__label'>
-          <div title={category.description}>
-            <i className='fas fa-question-circle' />
-          </div>
-
-          {category.name}
+      <label
```
notbakaneko: use the updated `InputContainer` from #13099

notbakaneko (changes requested):

resources/js/contest-judge/entry.tsx
Hiviexd: my brain is too hardwired on ternary operations 😭 
will fix

resources/css/bem/input-container.less
Hiviexd: going with 250px because anything smaller could risk messing up with longer labers

notbakaneko:
I'm directly commiting better alignment for the numbers that don't assume the number of digits based on the input in `CreateAnnouncement`

notbakaneko (approved):

PR #13062, opened 2026-06-13
Show judge score save state

Hiviexd:
Saving a judge score gives 0 feedback which feels like bad UX, so now it shows unsaved changes/saved text next to the update button.

### DIFF
--- a/resources/css/bem/contest-judge-entry.less
+++ b/resources/css/bem/contest-judge-entry.less
@@ -4,8 +4,27 @@
 .contest-judge-entry {
   &__button {
     display: flex;
+    align-items: center;
     justify-content: flex-end;
+    gap: 10px;
     text-transform: lowercase;
+  }
+
+  &__status {
+    font-weight: bold;
+
+    &--unsaved {
+      color: hsl(var(--hsl-orange-1));
+    }
+
+    &--saved {
+      color: @green;
+      transition: opacity 120ms ease-in-out;
+    }
+
+    &--fade-out {
+      opacity: 0;
+    }
   }
 
   &__categories {
--- a/resources/css/bem/contest-judge-entry.less
+++ b/resources/css/bem/contest-judge-entry.less
@@ -19,11 +19,6 @@
 
     &--saved {
       color: @green;
-      transition: opacity 120ms ease-in-out;
-    }
-
-    &--fade-out {
-      opacity: 0;
     }
   }
 
--- a/resources/css/bem/contest-judge-entry.less
+++ b/resources/css/bem/contest-judge-entry.less
@@ -12,10 +12,7 @@
 
   &__status {
     font-weight: bold;
-
-    &--unsaved {
-      color: hsl(var(--hsl-orange-1));
-    }
+    color: hsl(var(--hsl-orange-1));
 
     &--saved {
       color: @green;
--- a/resources/js/contest-judge/entry.tsx
+++ b/resources/js/contest-judge/entry.tsx
@@ -29,10 +29,10 @@ const commentsMaxLength = 3000;
 export default class Entry extends React.Component<Props> {
   @observable private readonly currentVote;
   @observable private readonly initialVote;
-  @observable private readonly store;
   @observable private savedFeedback: 'fading' | 'visible' | false = false;
-  @observable private xhr?: JQuery.jqXHR<ContestEntryJson>;
   private savedFeedbackTimeout?: number;
+  @observable private readonly store;
+  @observable private xhr?: JQuery.jqXHR<ContestEntryJson>;
 
   constructor(props: Props) {
     super(props);
@@ -64,6 +64,10 @@ export default class Entry extends React.Component<Props> {
 
   private get commentTooLong() {
     return this.currentVote.comment.length > commentsMaxLength;
+  }
+
+  componentWillUnmount() {
+    window.clearTimeout(this.savedFeedbackTimeout);
   }
 
   render() {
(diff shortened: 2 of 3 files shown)


notbakaneko:
we use the button state and visuals for feedback, the popup is generally a fallback like when the element is no longer visible because of a navigation or a full reload

Hiviexd:
what would you suggest in this case? from my uses (+a judge sharing the same complaint in the past) I kept wondering more than once if things saved given the button state is the only thing that's changing. I guess there's also the checkmark but I think that's just for the initial submission?

edit: I also noticed that forum moderation follows the same UX introduced in this PR, popup is emitted alongside the small button state change.

notbakaneko:
the accounts settings page has something like this that fades out, maybe something similar can be added (and maybe also unsaved changes)?
image

resources/js/contest-judge/entry.tsx
```diff
@@ -93,6 +105,21 @@ export default class Entry extends React.Component<Props> {
         </InputContainer>
 
         <div className='contest-judge-entry__button'>
+          {this.xhr != null && (
+            <div className='contest-judge-entry__status'>
+              <Spinner />
+            </div>
+          )}
```
notbakaneko: having 2 spinners seems very unnecessary

resources/css/bem/contest-judge-entry.less
```diff
@@ -4,10 +4,29 @@
 .contest-judge-entry {
   &__button {
     display: flex;
+    align-items: center;
     justify-content: flex-end;
+    gap: 10px;
     text-transform: lowercase;
   }
 
+  &__status {
```
notbakaneko, line 13: make `transition` and `opacity: 0` the base

resources/js/contest-judge/entry.tsx
```diff
@@ -173,6 +204,15 @@ export default class Entry extends React.Component<Props> {
...
         }
+
+        this.savedFeedback = 'visible';
+        window.clearTimeout(this.savedFeedbackTimeout);
+        this.savedFeedbackTimeout = window.setTimeout(action(() => {
+          this.savedFeedback = 'fading';
+          this.savedFeedbackTimeout = window.setTimeout(action(() => {
+            this.savedFeedback = false;
+          }), 120);
+        }), 3000);
```
notbakaneko: the whole manual fading `savedFeedback` seems unnecessary, we should be able to derive from existing state
```ts
window.clearTimeout(this.timeout);
this.timeout = window.setTimeout(action(() => {
  window.clearTimeout(this.timeout);
  this.timeout = undefined;
}), 3000);
```

resources/js/contest-judge/entry.tsx
```diff
@@ -93,6 +105,21 @@ export default class Entry extends React.Component<Props> {
...
+          {this.xhr == null && this.hasUnsavedChanges && (
+            <div className={classWithModifiers('contest-judge-entry__status', 'unsaved')}>
+              {trans('contest.judge.unsaved_changes')}
+            </div>
+          )}
+          {this.xhr == null && !this.hasUnsavedChanges && this.savedFeedback !== false && (
+            <div className={classWithModifiers('contest-judge-entry__status', 'saved', this.savedFeedback === 'fading' ? 'fade-out' : undefined)}>
+              {trans('common.saved')}
+            </div>
+          )}
```
notbakaneko, line 125: status text and visible can be derived from existing state 
```ts
  @observable
  private get statusText() {
    return this.hasUnsavedChanges
      ? trans('contest.judge.unsaved_changes')
      : this.savedFeedbackTimeout != null
        ? trans('common.saved')
        : null;
  }

  @observable
  private get statusVisible() {
    return this.hasUnsavedChanges || this.savedFeedbackTimeout != null;
  }
```

resources/js/contest-judge/entry.tsx
```diff
@@ -108,6 +135,8 @@ export default class Entry extends React.Component<Props> {
   @action
   private readonly handleCommentChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
     this.currentVote.comment = e.currentTarget.value;
+    this.savedFeedback = false;
+    window.clearTimeout(this.savedFeedbackTimeout);
```
notbakaneko: manually setting isn't needed if the state is derived

notbakaneko (changes requested):

resources/css/bem/contest-judge-entry.less
Hiviexd: i scrapped the opacity/transition stuff due to issue causing the element to unmount before the css animation, and fixing that seems to complicate things more than it'd be worth i think?

resources/js/contest-judge/entry.tsx
```diff
@@ -93,6 +117,14 @@ export default class Entry extends React.Component<Props> {
         </InputContainer>
 
         <div className='contest-judge-entry__button'>
+          {this.statusVisible && (
+            <div className={classWithModifiers('contest-judge-entry__status', {
+              saved: !this.hasUnsavedChanges,
+              unsaved: this.hasUnsavedChanges,
```
notbakaneko: only need one...

resources/js/contest-judge/entry.tsx
```diff
@@ -118,10 +118,7 @@ export default class Entry extends React.Component<Props> {
 
         <div className='contest-judge-entry__button'>
           {this.statusVisible && (
-            <div className={classWithModifiers('contest-judge-entry__status', {
-              saved: !this.hasUnsavedChanges,
-              unsaved: this.hasUnsavedChanges,
-            })}>
+            <div className={classWithModifiers('contest-judge-entry__status', this.hasUnsavedChanges ? undefined : 'saved')}>
```
notbakaneko:
```tsx
classWithModifiers('contest-judge-entry__status', { saved: !this.hasUnsavedChanges })
```

notbakaneko (approved):

PR #13063, opened 2026-06-13
Exclude hidden packs from beatmapset

Hiviexd:
definitely not intended for these to show on beatmap pages/api

### DIFF
--- a/app/Models/Beatmapset.php
+++ b/app/Models/Beatmapset.php
@@ -1600,7 +1600,7 @@ class Beatmapset extends Model implements AfterCommit, CommentableInterface, Ind
                 : explode(',', $rawValue);
         }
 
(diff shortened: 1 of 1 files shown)


app/Models/Beatmapset.php
```diff
@@ -1600,7 +1600,7 @@ private function getPackTags(): array
                 : explode(',', $rawValue);
         }
 
-        return $this->packs()->pluck('tag')->all();
+        return $this->packs()->where('hidden', false)->pluck('tag')->all();
```
nanaya: use the existing `->default()` scope instead

nanaya (changes requested):

nanaya (approved):

PR #13071, opened 2026-06-16
Split profile page detail statistics section component

nanaya:
As I noticed creating a new version of the page is a bit simpler if the `renderNumbers` part is split out.

The empty handling of `profile-detail` can probably be simply handled with direct `is_bot` check? I didn't touch the styling apart of moving out the `profile-detail__stats` and its content.


notbakaneko (approved):
seems to be all there

notbakaneko (approved):

PR #13072, opened 2026-06-16
Add ability to change game modes in popular beatmaps section

### LINKED ISSUE #5514: Popular Beatmaps should forcefully include non osu! gamemodes
For as long as I've been playing, I've never seen a non-osu! beatmap in the most played/popular beatmap section. I don't think that's representative of what the whole community considers popular. Many times non-osu! beatmaps will be very popular in their respesctive, but will never reach the popular beatmaps section because the other gamemodes simply don't have as big of a player-base as osu!standard I think including other gamemodes in the Popular Beatmaps section would help promote these other gamemodes.

I have a couple solutions for this.

 1. Force at least one of the Popular Beatmaps slots to always be a non-osu! gamemode.

    This is in my opinion the best option. It ensures that a non-osu! beatmap will always be in the popular beatmaps, no matter what. The problems I see with it though is that its an arbitrary amount of slots, and it might run into the same issue its trying to fix: one of the non-osu! gamemodes might be popular enough to where the other two gamemodes rarely if ever reach the Popular Beatmaps section.

 2. Weight non-osu! gamemodes more favorably in the Popular Beatmaps formula.

    This solves the issue of one of the gamemodes being more popular than the others and never allowing them to enter the popular beatmaps section, but it seems like a very finicky solution. What is the right ratios and formulas for weighting? I sure don't know.

3. Add separate tabs for non-osu! gamemodes

    This could be integrated nicely with player's default gamemode on their userpage, always showing the gamemode they are interested in. A lot of non-osu! players aren't invested are interested in osu! beatmaps. Only their own gamemode. The problem I see with this is that hybrid beatmaps with osu! diffs are bound to get tons more plays.

Hiviexd:
Resolves https://github.com/ppy/osu-web/issues/5514 I think?

image


resources/js/popular-beatmaps-mode.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+document.addEventListener('turbo:load', () => {
+  for (const container of document.querySelectorAll<HTMLElement>('[data-popular-beatmaps]')) {
```
notbakaneko: use `js-` prefixed class selectors, same for the others

notbakaneko (dismissed):

app/Http/Controllers/HomeController.php
```diff
@@ -130,7 +131,11 @@ public function index()
         if (Auth::check()) {
             $menuImages = json_collection(MenuContent::activeImages(), new MenuImageTransformer());
             $newBeatmapsets = Beatmapset::latestRanked();
-            $popularBeatmapsets = Beatmapset::popular()->get();
+
+            $popularBeatmapsetsByMode = [];
```
nanaya: `ByRuleset` (also everywhere else)

app/Http/Controllers/HomeController.php
```diff
@@ -130,7 +131,11 @@ public function index()
...
+            foreach (Beatmap::MODES as $mode => $_) {
+                $popularBeatmapsetsByMode[$mode] = Beatmapset::popular($mode)->get();
+            }
```
nanaya: no point in caching them one by one when they're all always fetched together

resources/js/entrypoints/app.ts
```diff
@@ -17,3 +17,4 @@ import 'osu-core-singleton';
 import 'main.coffee';
 
 import 'register-components';
+import 'popular-beatmaps-mode';
```
nanaya: this kind of thing is usually defined in `core/` and initialised in `osu-core`

resources/js/popular-beatmaps-mode.ts
```diff
@@ -0,0 +1,23 @@
...
+  for (const container of document.querySelectorAll<HTMLElement>('.js-popular-beatmaps')) {
+    container.addEventListener('click', (event) => {
+      const button = (event.target as Element).closest<HTMLButtonElement>('.js-popular-beatmaps-mode');
+      if (button == null) {
+        return;
+      }
+
+      const mode = button.dataset.popularMode;
+
+      for (const otherButton of container.querySelectorAll<HTMLButtonElement>('.js-popular-beatmaps-mode')) {
```
nanaya: the other variable should be called `currentButton` (or `selectedButton`) and this one just `button` because this also includes the current button

resources/css/bem/user-home.less
```diff
@@ -13,12 +13,38 @@
...
+      gap: 8px;
+    }
+
+    .game-mode-link {
+      font-size: @font-size--title-small-3;
+    }
+
+    button.game-mode-link {
+      .reset-input();
+    }
```
nanaya: no modifying other blocks. Add modifiers to the block itself instead

nanaya (changes requested):

resources/js/core/popular-beatmaps-ruleset.ts
```diff
@@ -0,0 +1,28 @@
...
+import { htmlElementOrNull } from 'utils/html';
+
+export default class PopularBeatmapsRuleset {
+  constructor() {
+    $(document).on('click', '.js-popular-beatmaps-ruleset', this.onClick);
+  }
+
+  private readonly onClick = (event: JQuery.ClickEvent) => {
+    const selectedButton = htmlElementOrNull(event.currentTarget);
+    if (!(selectedButton instanceof HTMLButtonElement)) return;
```
nanaya: no point in using the `htmlElementOrNull` if it's checked for the exact type anyway

or maybe just `?? fail(...)` as it's not supposed to happen in the first place (also for the other check).

resources/js/core/popular-beatmaps-ruleset.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    const container = selectedButton.closest('.js-popular-beatmaps');
+    if (!(container instanceof HTMLElement)) return;
+
+    const ruleset = selectedButton.dataset.popularRuleset;
+
+    for (const button of container.querySelectorAll<HTMLButtonElement>('.js-popular-beatmaps-ruleset')) {
+      button.classList.toggle('game-mode-link--active', button.dataset.popularRuleset === ruleset);
```
nanaya: should be `js-` prefixed class like `js-is-active`

resources/views/home/user.blade.php
```diff
@@ -130,17 +132,45 @@ class="user-home__news-posts-group user-home__news-posts-group--more"
...
+                        <div
+                            @class([
+                                'user-home__beatmapsets',
+                                'js-popular-beatmaps-panel',
+                                'u-hidden' => $ruleset !== $popularRuleset,
+                            ])
+                            data-popular-ruleset="{{ $ruleset }}"
+                        >
+                            @foreach ($popularBeatmapsetsByRuleset[$ruleset] as $beatmapset)
+                                @include('home._user_beatmapset', ['type' => 'popular'])
```
nanaya, line 169: I wonder if the ruleset icon in the entry is redundant because the section itself now has ruleset

resources/css/bem/user-home.less
```diff
@@ -13,12 +13,24 @@
...
+  }
+
   &__beatmap-list-title {
     margin: 0;
     padding: 20px 0 10px;
     font-size: @font-size--title-small-3;
     color: @osu-colour-l1;
     font-weight: bold;
+
+    &--in-header {
```
nanaya: or the other title can be in header too and have this and the duplicated padding removed

resources/css/bem/game-mode-link.less
```diff
@@ -20,6 +20,14 @@
     color: #fff;
   }
 
+  &--button {
+    .reset-input();
+  }
+
+  &--compact {
+    font-size: @font-size--title-small-3;
```
nanaya: should probably be `inherit` if it's intended to match the title (and adjust the size there accordingly)

nanaya (changes requested):

resources/js/core/popular-beatmaps-ruleset.ts
Hiviexd: I added `htmlElementOrNull` because lint failed on its absence

edit: will use `fail(...)`

nanaya: which lint? I removed the function and nothing came up. The following check already type the variable as `HTMLButtonElement`

resources/js/core/popular-beatmaps-ruleset.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export default class PopularBeatmapsRuleset {
+  constructor() {
+    $(document).on('click', '.js-popular-beatmaps-ruleset', this.onClick);
+  }
+
+  private readonly onClick = (event: JQuery.ClickEvent) => {
+    const selectedButton = event.currentTarget instanceof HTMLButtonElement
+      ? event.currentTarget
+      : fail('popular beatmaps ruleset button is missing');
```
nanaya: since a HtmlElement is enough
```suggestion
    const selectedButton = htmlElementOrNull(event.currentTarget) ?? fail('popular beatmaps ruleset button is missing');
```

resources/css/bem/game-mode-link.less
```diff
@@ -15,11 +15,20 @@
...
   &--active {
     font-weight: bold;
     color: #fff;
   }
 
+  &--button {
+    .reset-input();
+  }
+
+  &--compact {
```
nanaya: the name doesn't reflect the content

app/Models/Beatmapset.php
```diff
@@ -182,18 +182,34 @@ public static function coverSizes()
...
+
+        foreach ($idsByRuleset as $ruleset => $ids) {
+            $popularByRuleset[$ruleset] = static::whereIn('beatmapset_id', $ids)->orderByField('beatmapset_id', $ids)->get();
+        }
+
+        return $popularByRuleset;
     }
 
-    public static function popularIds()
+    public static function popularIds(?string $ruleset = null)
```
nanaya: the parameter shouldn't be optional/nullable

resources/js/core/popular-beatmaps-ruleset.ts
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+import { fail } from 'utils/fail';
+
+const activeClass = 'js-is-active';
+
+export default class PopularBeatmapsRuleset {
```
nanaya: just noticed, `Beatmapset`. Also everywhere else

resources/views/home/user.blade.php
```diff
@@ -130,17 +136,45 @@ class="user-home__news-posts-group user-home__news-posts-group--more"
...
+                        <h3 class="user-home__beatmap-list-title">
+                            {{ osu_trans('home.user.beatmaps.popular') }}
+                        </h3>
+
+                        <ul class="game-mode game-mode--inline">
+                            @foreach (Beatmap::MODES as $ruleset => $rulesetId)
+                                <li>
+                                    <button
+                                        type="button"
+                                        class="{{ class_with_modifiers('game-mode-link', 'compact', 'button') }} js-popular-beatmaps-ruleset{{ $ruleset === $popularRuleset ? ' js-is-active' : '' }}"
```
nanaya: why this one uses blade `{{ }}` and the other uses `@class`

resources/css/bem/user-home.less
```diff
@@ -13,10 +13,18 @@
...
     padding: 20px 0 10px;
     font-size: @font-size--title-small-3;
+  }
+
+  &__beatmap-list-title {
+    margin: 0;
+    font-size: inherit;
     color: @osu-colour-l1;
     font-weight: bold;
   }
```
nanaya: probably a bit simpler if the `-header` is a `h3` and the `-title` is `span` instead (or nothing at all since it can just be styled by the `-header`)

```
<h3>Popular Beatmaps <div>game mode selector</div></h3>
```

resources/views/home/user.blade.php
```diff
@@ -130,17 +136,45 @@ class="user-home__news-posts-group user-home__news-posts-group--more"
                     </a>
                 </div>
 
-                <h3 class='user-home__beatmap-list-title'>
-                    {{ osu_trans('home.user.beatmaps.popular') }}
-                </h3>
+                <div class="js-popular-beatmaps">
```
nanaya: can use `u-contents`

nanaya (changes requested):

resources/css/bem/game-mode-link.less
Hiviexd: i assume you're just referring to `--compact`?

nanaya: yes

resources/views/home/user.blade.php
Hiviexd: oops, we do a bit of distracted blundering 😅

app/Models/Beatmapset.php
```diff
@@ -182,18 +182,33 @@ public static function coverSizes()
...
+            }
+
+            return $idsByRuleset;
         });
 
-        return static::whereIn('beatmapset_id', $ids)->orderByField('beatmapset_id', $ids);
+        $popularByRuleset = [];
+
+        foreach ($idsByRuleset as $ruleset => $ids) {
+            $popularByRuleset[$ruleset] = static::whereIn('beatmapset_id', $ids)->orderByField('beatmapset_id', $ids)->get();
```
nanaya: just noticed this is still doing the queries four times over.

resources/css/bem/user-home.less
```diff
@@ -13,14 +13,22 @@
...
+    gap: 10px;
     margin: 0;
     padding: 20px 0 10px;
     font-size: @font-size--title-small-3;
     color: @osu-colour-l1;
     font-weight: bold;
   }
 
+  &__beatmap-list-header-controls {
+    flex: none;
```
nanaya: what this for?

nanaya (changes requested):

nanaya (approved):

PR #13077, opened 2026-06-18
Add lazer only flag on beatmaps

nanaya:
Dunno if the switch mode message should be in the box or in the scoreboard. And if the scoreboard should show the tabs and mods selector if the message is in scoreboard.

Switching lazer mode currently requires full page reload.


resources/js/beatmapsets-show/main.tsx
```diff
@@ -111,6 +133,37 @@ export default class Main extends React.Component<Props> {
     );
   }
 
+  private renderLazerOnlyMessage() {
```
Hiviexd, line 136: IMO design doesn't feel loud enough for a warning of this caliber? I feel it could blend in too much for any user not paying enough attention.

peppy (approved):

notbakaneko (approved):
nice recycling ♻️ 👀

PR #13078, opened 2026-06-18
Move notification icons to main mobile navbar

### LINKED ISSUE #12890: Make notifications more visible on mobile browsers
While using mobile Safari, notifications are not visible until I open the drop-down menu or I expand and looking through tabs. I think notification tabs should be its own section alongside the chevron button similar to how it shows on desktop, that way it's visible at all times.

While on web:
Image

While opening drop-down menu:
Image

While looking through tabs:
Image

Hiviexd:
Resolves https://github.com/ppy/osu-web/issues/12890

image

### DIFF
--- a/resources/css/bem/mobile-menu.less
+++ b/resources/css/bem/mobile-menu.less
@@ -40,4 +40,13 @@
     display: flex;
     background-color: @osu-colour-d4;
   }
+
+  &--notifications {
+    background-color: @osu-colour-b5;
+
+    .mobile-menu__item {
+      max-height: calc((var(--vh, 1vh) * 100) - @navbar-height);
+      overflow-y: auto;
+    }
+  }
 }
--- a/resources/views/layout/_header_mobile.blade.php
+++ b/resources/views/layout/_header_mobile.blade.php
@@ -17,6 +17,36 @@
         </div>
 
         <div class="navbar-mobile__header-section navbar-mobile__header-section--buttons">
+            @if (isset($user))
+                <a
+                    class="navbar-mobile__icon js-click-menu js-react"
+                    data-click-menu-target="mobile-chat-notification"
+                    data-react="chat-icon"
+                    data-turbo-permanent
+                    data-type="mobile"
+                    id="notification-widget-chat-icon-mobile"
+                    href="{{ route('chat.index') }}"
+                >
+                    <span class="notification-icon notification-icon--mobile">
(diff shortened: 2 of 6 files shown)


nanaya (commented):
I haven't looked closer but the notification popup should probably be independent from the mobile menu and tapping on the icon again should close it instead of showing navigation menu

resources/views/layout/_header_mobile.blade.php
```diff
@@ -17,6 +17,36 @@
         </div>
 
         <div class="navbar-mobile__header-section navbar-mobile__header-section--buttons">
+            @if (isset($user))
+                <a
+                    class="navbar-mobile__icon js-click-menu js-react"
```
notbakaneko, line 22: Just use the existing block class.
I don't see the point of copying a block's styles into the element level of another block when everything in it is pretty much the same other than the name.

resources/css/bem/mobile-menu.less
```diff
@@ -40,4 +40,13 @@
     display: flex;
     background-color: @osu-colour-d4;
   }
+
+  &--notifications {
+    background-color: @osu-colour-b5;
```
notbakaneko, line 45: this colour doesn't seem to actually show anywhere?

resources/css/bem/mobile-menu.less
```diff
@@ -40,4 +40,13 @@
...
+
+    .mobile-menu__item {
+      max-height: calc((var(--vh, 1vh) * 100) - @navbar-height);
+      overflow-y: auto;
+    }
```
notbakaneko, line 50: this should be under `&__item` itself like
```
&__item {
  .mobile-menu--notifications & {
  }
}
```
or better just add the modifier to the element 
```
&__item
  &--notifications
```
but it looks more like this shouldn't be changing

notbakaneko (changes requested):
After discussing with @nanaya, it might be better to have a different menu for this instead of trying to hack the existing switch menu to work with it, which isn't implemented with this kind of behaviour in mind. 

---
posting my initial comments:
The menu should not slide up and down when switching between the top level items

notbakaneko:
@Hiviexd I'm reworking the menu to behave the way we expect it to behave

Hiviexd:
Feel free to, I'm away and the soonest I'll be available to fix stuff is next week.

PR #13080, opened 2026-06-18
Generalize UserTagPicker

LiquidPL:
RFC. I'm not sure if this is overkill and it'd be easier/cleaner to just copypaste the tag picker layout to a new component, so I'm PR'ing this in for discussion.

Intended to be used in a future PR allowing trusted users (and possibly other users in the future) to edit beatmap tags through web.


resources/css/bem-index.less
```diff
@@ -46,6 +46,7 @@
 @import "bem/beatmap-discussions";
 @import "bem/beatmap-discussions-header-bottom";
 @import "bem/beatmap-discussions-header-top";
+@import "bem/beatmap-user-tag-picker";
```
nanaya: maybe just call it `beatmap-tag-picker` instead?

resources/js/components/beatmap-tag-picker.tsx
nanaya, line 1: do a two steps commit - one renaming the file, and another one updating the file. Otherwise the history won't get carried over to the new file

resources/css/bem-index.less
LiquidPL: i feel like this would make it too ambiguous whether it concerns mapper tags or user tags. not sure how those are generally named throughout the project

nanaya: right. except the mapper tag is attached to beatmapset, not beatmap. and I think user tag is referred as beatmap tag everywhere

LiquidPL:
Will force push this branch in order to split out the rename into its own commit.

LiquidPL:
Could this PR be looked at again (I'm not sure if I have request review perms)? It's kinda needed to improve beatmap tagging for ranked play.

resources/css/bem-index.less
```diff
@@ -442,7 +442,7 @@
 @import "bem/user-search-card";
 @import "bem/user-session-list";
 @import "bem/user-session-list-session";
-@import "bem/user-tag-picker";
+@import "bem/beatmap-tag-picker";
```
nanaya, line 445: rerun `bin/make_bem_index.sh`

resources/js/components/beatmap-tag-picker.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+const controller = core.beatmapTagPickerController;
+
+const onChange = action((e: React.ChangeEvent<HTMLInputElement>) => controller.query = e.target.value);
+
+interface Props {
+  isTagEnabled: (tag: BeatmapTag) => boolean;
+  onDisabled: (tag: BeatmapTag) => void;
+  onEnabled: (tag: BeatmapTag) => void;
+  showAllRulesets: boolean;
```
nanaya, line 21: how's this supposed to be used with half of the states still being in the singleton controller? `beatmapTagPickerController` itself is using `core.beatmapsetSearchController` internally too.

nanaya (changes requested):

resources/js/components/beatmap-tag-picker.tsx
LiquidPL: leaving `rulesetId` in `BeatmapTagPickerController` is an oversight on my part. this would probably end up being cleaner if `BeatmapTagPickerController` is split into two classes:

- one of them is only a cache for the tags fetched from the backend,
- the other one has all the logic related to filtering/grouping tags for display (would have to be specific per each unique use of the tag picker component, probably an abstract class with a way to specify where the current ruleset is pulled from)

```typescript
class BeatmapTagStore {
  @observable tags: BeatmapTag[];

  protected constructor() {
    makeObservable(this);
    onBecomeObserved(this, 'tags', this.fetchTags);
  }

  private fetchTags() { /* ... */ }
}

abstract class BeatmapTagPickerController {
  @observable query: string = '';

  abstract get rulesetId(): RulesetId|undefined;

  @computed
  get groups() { /* reads from BeatmapTagStore.tags */ }
}

// current controller
class ListingBeatmapTagPickerController extends BeatmapTagPickerController {
  @computed
  get rulesetId() {
    const mode = core.beatmapsetSearchController.filters.mode;

    if (mode !== null) {
      return ensureRulesetId(mode);
    }
  }
} 

// future controller
class InfoBeatmapTagPickerController extends BeatmapTagPickerController {
  @computed
  get rulesetId() { /* ... */ }
}
```

and then pass that to `UserTagPicker` as well.

nanaya: or maybe keep the whole thing except the tag data in the controller and pass that controller as props instead. The controller can then be marked to implement an interface which can be used for other stuff (and have the component require the interface instead).

peppy:
@LiquidPL still around to keep this moving? 😄

LiquidPL:
yes, I will hop to it today, been kinda busy with life recently

resources/js/components/beatmap-tag-picker.tsx
LiquidPL: opted to make the base controller abstract instead of using an interface, since the grouped tag getter isn't going to change between implementations and can be easily reused (still made it so the controller is passed as a parameter though)

resources/js/components/beatmap-tag-picker-controller.ts
```diff
@@ -0,0 +1,36 @@
...
+  abstract rulesetId: RulesetId|undefined;
+
+  @observable query: string = '';
+
+  constructor() {
+    makeObservable(this);
+  }
+
+  @computed
+  get groups() {
```
nanaya: technically this can just be in the react component itself, can't this?

resources/css/bem-index.less
```diff
@@ -38,14 +38,14 @@
 @import "bem/beatmap-discussion-post";
 @import "bem/beatmap-discussion-review-post";
 @import "bem/beatmap-discussion-review-post-embed-preview";
+@import "bem/beatmap-discussions";
```
nanaya: The script didn't enforce the locale for order (until #13345) but the file shouldn't change this much.

resources/js/components/beatmap-tag-picker-controller.ts
LiquidPL: good point

PR #13086, opened 2026-06-19
Add beatmapset cover image to beatmapset page header

gagahpangeran:
Using slim cover variant for this seems correct.

### DIFF
--- a/resources/js/beatmapsets-show/main.tsx
+++ b/resources/js/beatmapsets-show/main.tsx
@@ -153,6 +153,7 @@ export default class Main extends React.Component<Props> {
   private renderPageHeader() {
     return (
       <HeaderV4
+        backgroundImage={this.controller.beatmapset.covers.slimcover}
         links={headerLinks('show', this.controller.beatmapset)}
         linksAppend={this.headerLinksAppend}
         theme='beatmapset'


LiquidPL:
I'm not sure why is this necessary, the cover should already be handled in https://github.com/ppy/osu-web/blob/master/resources/js/beatmapsets-show/main.tsx#L171 and https://github.com/ppy/osu-web/blob/master/resources/js/beatmapsets-show/header.tsx.

gagahpangeran:
I'm using this [figma design](https://www.figma.com/design/HDSXf4LPPV7hnD5wrWYG4jix/Web-Beatmap-Info?node-id=512-11&t=07htFzIWOT4K5Ukd-0) as reference (while updating my PR in #7899). It said that the header background is the same as the cover images, so I just make this small PR to add it. Unless there is updated design document somewhere that says otherwise.

LiquidPL:
My bad, I didn't realize that 'header' means specifically the global page header here.

peppy:
@ppy/team-web maybe we should do this? seems like it's better than a generic header background.

nanaya:
one thing to note would be when the map has broken/missing background, the header will have no background

peppy:
Probably still ok?

resources/js/beatmapsets-show/main.tsx
```diff
@@ -153,6 +153,7 @@ export default class Main extends React.Component<Props> {
   private renderPageHeader() {
     return (
       <HeaderV4
+        backgroundImage={this.controller.beatmapset.covers.slimcover}
```
nanaya, line 156: this needs to check to the visibility similar to the one in BeatmapsetCover component

and then the background needs to be blurred and stuff just like for the profile page

nanaya (changes requested):

nanaya:
I haven't checked but does this also apply to the discussion and version history page?

PR #13090, opened 2026-06-21
change beatmap list component to be more general

gagahpangeran:
While working on rebasing #7899, I notice that beatmap list component now only accept discussion state as its props.

Because I want to reuse this component in new beatmapset show header, I decided to generalize this component and move it to general component folder.

### DIFF
--- a/resources/js/beatmap-discussions/header.tsx
+++ b/resources/js/beatmap-discussions/header.tsx
@@ -12,6 +12,7 @@ import HeaderV4 from 'components/header-v4';
 import NotificationBanner from 'components/notification-banner';
 import PlaymodeTabs from 'components/playmode-tabs';
 import StringWithComponent from 'components/string-with-component';
+import BeatmapExtendedJson from 'interfaces/beatmap-extended-json';
 import BeatmapsetDiscussionsStore from 'interfaces/beatmapset-discussions-store';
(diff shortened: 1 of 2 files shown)


resources/js/beatmap-discussions/header.tsx
```diff
@@ -166,8 +182,12 @@ export class Header extends React.Component<Props> {
...
             <div className={`${bn}__filter-group`}>
               <BeatmapList
-                discussionsState={this.discussionsState}
-                users={this.users}
+                beatmaps={this.beatmaps}
+                beatmapset={this.beatmapset}
+                currentBeatmap={this.currentBeatmap}
+                getCount={this.getCount}
+                makeBeatmapUrl={this.makeBeatmapUrl}
+                onSelectBeatmap={this.onSelectBeatmap}
```
notbakaneko: `onBeatmapSelected`

notbakaneko (commented):
This really should just use or wrap `SelectOptions` if it's going to offload the logic in the options to the parent component. We're trying to move away from having callbacks just to get more data when rendering items and having the data passed in, instead.

notbakaneko:
After doing #13331 I'm not sure if we want to share the beatmap list component; if it's just passing props and callbacks down, it doesn't need to be there. The discussion version of the beatmap list is to handle logic specific to discussions.

PR #13092, opened 2026-06-22
Fix check for external embeds when updating Beatmap Discussion review

notbakaneko:


app/Libraries/BeatmapsetDiscussion/Review.php
```diff
@@ -205,8 +206,10 @@ private function process()
                 );
             } else {
                 // ensure all referenced embeds belong to this discussion
-                $externalEmbeds = BeatmapDiscussion::whereIn('id', $childIds)->where('parent_id', '<>', $this->discussion->getKey())->count();
-                if ($externalEmbeds > 0) {
+                $externalEmbeds = BeatmapDiscussion::whereIn('id', $childIds)
+                    ->where(fn ($q) => $q->where('parent_id', '<>', $this->discussion->getKey())->orWhere('parent_id', null))
+                    ->exists();
```
nanaya: or `->where('parent_id', $reviewId)->count()` and check that it matches? the query would be simpler

also the parent id update down below can probably be moved to after discussion creation above

nanaya (changes requested):

nanaya (approved):

PR #13099, opened 2026-06-23
Support React elements as label for InputContainer

notbakaneko:
More flexibility beyond translation keys.
Also, allow showing length counter without label text.

### DIFF
--- a/resources/js/components/input-container.tsx
+++ b/resources/js/components/input-container.tsx
@@ -10,6 +10,7 @@ import MessageLengthCounter from './message-length-counter';
 interface CommonProps {
   for?: string;
   hasError?: boolean;
+  label?: React.ReactNode;
   labelKey?: string;
   modifiers?: Modifiers;
   showError?: boolean;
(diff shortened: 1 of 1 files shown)


resources/js/components/input-container.tsx
```diff
@@ -27,12 +28,13 @@ type Props = CommonProps & ({
...
   const error = props.hasError && props.showError;
+  const labelText = props.label ?? (props.labelKey != null ? trans(props.labelKey) : null);
 
   return (
     <label className={classWithModifiers('input-container', { error }, props.modifiers)} htmlFor={props.for}>
-      {props.labelKey != null && (
+      {(labelText != null || props.maxLength != null) && (
         <div className='input-container__label'>
-          {trans(props.labelKey)}
+          {labelText}
```
nanaya: should the counter stay on the right or move to the left is there's no label?

staying on the right seems more consistent maybe? 🤔 also less movements when the value changes

notbakaneko: and now I'm reminded the top spacing is from padding and not the presence of elements 🙈

nanaya: might as well remove the check?

nanaya (approved):

PR #13101, opened 2026-06-24
Do not store total score without mods for scores that do not have mods

bdach:
- Part of post-https://github.com/ppy/osu/issues/37818 cleanup
- [x] Should not be merged until https://github.com/ppy/osu-queue-score-statistics/pull/385 is merged and deployed (marking draft to ensure this)

See aforementioned PR and remarks added to API docs for reasoning.


app/Models/Solo/Score.php
```diff
@@ -81,11 +81,18 @@ class Score extends Model implements Traits\ReportableInterface
 
     public static function createFromJsonOrExplode(array $params): static
     {
+        $mods = $params['mods'] ?? [];
+        $totalScoreWithoutMods = $params['total_score_without_mods'] ?? null;
+
+        if (empty($mods)) {
+            $totalScoreWithoutMods = null;
+        }
```
nanaya: this logic should probably be in ScoreData constructor?

bdach: I am not sure I agree; I think moving it to the constructor obscures the location of the logic and also may result in the logic firing where it shouldn't. One current example that may fit that last description is

https://github.com/ppy/osu-web/blob/b7ab85a88249fa14295ff8d4e4633fe3bdfd2fd1/app/Models/LegacyMatch/Score.php#L152

Nevertheless, applied in https://github.com/ppy/osu-web/pull/13101/commits/f17e725d18deb0530e40d2bab9363e54ceefb1ae.

nanaya: I guess that depends on whether or not having non null field with empty mods is valid

bdach: It's an implementation detail by this point. In practice it's probably safe to assume that no mods implies no total score without mods in all cases. Currently there are database rows that do not meet that implication but the eventual plan is to clean them up such that they do meet it.

peppy:
@nanaya this one is good to go

nanaya (approved):

PR #13106, opened 2026-06-25
Fix incorrect casting for account option parameters

nanaya:
Namely `extra_options` shouldn't be casted to string by the default param value handler.

I forgot the default isn't a passthrough/simple null check 🙃


notbakaneko (approved):
🙈

PR #13108, opened 2026-06-25
Include qualified beatmaps in nominated ranked maps user page category

### LINKED ISSUE #9544: Include qualified maps in the Nominated Ranked Maps user page category
Pretty self-explanatory, would be great to include Qualified maps in the same way they are included under the Ranked Maps category.
No Nominated Ranked Maps listed...
image
despite me having nominated a map:
image
Consistency with Ranked Maps would be implemented like so:
image

gagahpangeran:
closes #9544


nanaya (approved):
For reference, qualified is a rather temporary state for it to be included in the ranked section (both nominated and own). There's not really anywhere else to put it though, short of adding a new section or moving it into pending. And both nominated and own ranked sections should use the same list of states so we're going with this (for now?).

PR #13109, opened 2026-06-25
Fix extra stats overflow in score page on mobile

gagahpangeran:
Looks like only osu! ruleset that has extra stats overflow because there are 4 long label.

The other alternative to fix this is shorten the label itself.

| Before | After |
| --- | --- |
|  |  |


nanaya:
making it a flexbox may be better?

gagahpangeran:
> making it a flexbox may be better?

are you talking about changing the whole `score-stats__group-row` from grid to flexbox?

nanaya:
yes

gagahpangeran:
I've tried this. Is this okay?

```diff
diff --git a/resources/css/bem/score-stats.less b/resources/css/bem/score-stats.less
index 99f0422b41..55751c0daa 100644
--- a/resources/css/bem/score-stats.less
+++ b/resources/css/bem/score-stats.less
@@ -35,14 +35,14 @@
 
   &__group-row {
     .@{_top}__group--stats & {
-      display: grid;
-      grid-auto-flow: column;
-      grid-auto-columns: 1fr;
-      grid-gap: 2px;
+      display: flex;
+      gap: 10px 2px;
+      flex-wrap: wrap;
     }
   }
 
   &__stat {
+    flex: 1;
     display: flex;
     flex-direction: column;
     text-align: center;
```

nanaya:
yes but can use max-width and center alignment

resources/css/bem/score-stats.less
```diff
@@ -35,14 +35,16 @@
...
+      display: flex;
+      gap: 10px 2px;
+      flex-wrap: wrap;
+      justify-content: center;
     }
   }
 
   &__stat {
+    flex: 1;
+    max-width: 50%;
```
gagahpangeran, line 47: Not sure about this max-width.

nanaya (approved):

PR #13112, opened 2026-06-26
Refactor store checkout related modules

notbakaneko:
This originally started as converting and cleaning up Xsolla checkout related stuff to prepare to make updates to deprecated (and then undeprecated) libraries.
Then it turned into cleaning out the old unneeded scripts and widgets loading


app/Http/Controllers/Store/CheckoutController.php
```diff
@@ -62,8 +62,9 @@ public function store()
             return ujs_redirect(route('store.cart.show'));
         }
 
-        if ($params['hide_from_activity'] !== null) {
-            $order->setGiftsHidden($params['hide_from_activity']);
+        $hideFromActivity = $params['hide_from_activity'];
+        if ($hideFromActivity) {
+            $order->setGiftsHidden($hideFromActivity);
```
nanaya, line 69: might as well remove the parameter? the function seems to be only called here (and tests)

resources/js/store/index.ts
```diff
@@ -0,0 +1,157 @@
...
+
+      // TODO: less unknown error, disable button
+      // TODO: handle error.message
+      onError(err);
+    }
+  }
+
+  private resumeCheckout(event: TriggeredEvent) {
+    const cartId = event.target.dataset.providerReference;
+    if (cartId == null) throw new Error('cartId is missing');
```
nanaya: can use `?? fail(...)`

resources/js/store/index.ts
```diff
@@ -0,0 +1,157 @@
...
+    this.resumeShopifyCheckout(cartId);
+  }
+
+  private async resumeShopifyCheckout(cartId: string) {
+    showLoadingOverlay();
+    showLoadingOverlay.flush();
+    const response = await getShopifyCart(cartId);
+    const data = response.data as { cart?: Cart };
```
nanaya: type inside the api function instead?

resources/js/store/index.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+  }
+
+  private async beginShopifyCheckout(orderId: string) {
+    showLoadingOverlay();
+    showLoadingOverlay.flush();
+    // create shopify checkout.
+    // error returned will be a JSON string in error.message
+    const response = await createShopifyCart(orderId, Array.from(document.querySelectorAll('.js-store-order-item')));
+    const data = response.data as { cartCreate: CartCreatePayload };
```
nanaya: another one that can be typed in the createShopifyCart function itself

resources/js/store/index.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  private async beginCheckout(event: TriggeredEvent) {
+    if (event.target == null) return;
+
+    const dataset = event.target.dataset;
+    const orderId = dataset.orderId;
+    const shouldShopify = dataset.shopify === '1';
+    if (orderId == null) {
+      throw new Error('orderId is missing');
+    }
```
nanaya: can use `?? fail(...)` on assignment

resources/js/store/index.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    $(document).on('turbo:load', () => {
+      $('.js-store-checkout').prop('disabled', false);
+    });
+
+    $('.js-store-checkout').prop('disabled', false);
+  }
+
+  static init() {
+    window.Store ??= new Store();
```
nanaya: maybe add to `core.store ??=` instead

or shop...

resources/js/store/index.ts
```diff
@@ -0,0 +1,157 @@
...
+    if (provider == null || orderId == null) throw new Error();
+    showLoadingOverlay();
+    showLoadingOverlay.flush();
+
+    if (provider === 'xsolla') {
+      if (orderNumber == null) throw new Error('missing orderNumber');
+      await initXsolla(orderNumber);
+    }
+
+    const hide_from_activity = document.querySelector<HTMLInputElement>('.js-hide-from-activity')?.checked;
```
nanaya: camel case?

resources/js/store/index.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+}
+
+type TriggeredEvent = JQuery.TriggeredEvent<Document, unknown, HTMLElement, HTMLElement>;
+
+export default class Store {
+  private constructor() {
+    $(document).on('click.store', '.js-store-checkout', (event: TriggeredEvent) => void this.beginCheckout(event));
+    $(document).on('click.store', '.js-store-resume-checkout', (event: TriggeredEvent) => this.resumeCheckout(event));
+    $(document).on('click.store', '.js-store-payment-button', (event: TriggeredEvent) => void this.handlePaymentClick(event));
```
nanaya: why don't just pass the function directly...? `.on(..., this.beginCheckout)` etc

also the namespace doesn't do anything?

resources/js/store/store-xsolla.ts
```diff
@@ -0,0 +1,48 @@
...
+const xsollaWidgetUrl = 'https://cdn.xsolla.net/payments-bucket-prod/embed/1.5.4/widget.min.js';
+
+function onXsollaReady(orderNumber: string) {
+  let done = false;
+  window.XPayStationWidget.on(window.XPayStationWidget.eventTypes.STATUS_DONE, () => done = true);
+  window.XPayStationWidget.on(window.XPayStationWidget.eventTypes.CLOSE, () => {
+    if (done) {
+      showLoadingOverlay();
+      showLoadingOverlay.flush();
+      window.location.href = route('payments.xsolla.completed', { foreignInvoice: orderNumber });
```
nanaya: this is from the old one but shouldn't this use turbo navigate...?

resources/js/store/store-xsolla.ts
```diff
@@ -0,0 +1,48 @@
...
+declare global {
+  interface Window {
+    XPayStationWidget: {
+      eventTypes: {
+        CLOSE: string;
+        STATUS_DONE: string;
+      };
+      init: (config: XsollaTokenResponse) => void;
+      on: (event: string, callback: () => void) => void;
+      open(): void;
```
nanaya: can use a bit more consistency instead of having both `fun(): ret` and `fun: () => ret`

resources/js/store/index.ts
```diff
@@ -0,0 +1,157 @@
...
+    showLoadingOverlay();
+    showLoadingOverlay.flush();
+
+    if (provider === 'xsolla') {
+      if (orderNumber == null) throw new Error('missing orderNumber');
+      await initXsolla(orderNumber);
+    }
+
+    const hide_from_activity = document.querySelector<HTMLInputElement>('.js-hide-from-activity')?.checked;
+    await $.post(route('store.checkout.store'), { hide_from_activity, orderId, provider });
```
nanaya: why is the parameter part camel case part snake case 🤨

resources/js/store/index.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    $('.js-store-checkout').prop('disabled', false);
+  }
+
+  static init() {
+    window.Store ??= new Store();
+  }
+
+  private async beginCheckout(event: TriggeredEvent) {
+    if (event.target == null) return;
```
nanaya: can this even be null

nanaya (changes requested):

resources/js/store/index.ts
notbakaneko: it's either this or having a wrapper handler function to call an `async` function because passing an async function fails `@typescript-eslint/no-misused-promises` since it returns a Promise to a handler that expects `void`

notbakaneko:
removing the separate xsolla and paypal token/approval link request steps...

nanaya (approved):
Seems working...?

PR #13114, opened 2026-06-29
Always instantiate transformers directly

notbakaneko:
No more untyped magic string transformer.

This has been on my todo list for a while and always bugs me when I have to look for transformer usages.


nanaya (approved):
I don't think I've ever passed string for `$includes` 🤔

PR #13119, opened 2026-07-03
Clean up handling of line chart margins and sizes

nanaya:
- remove extra margin on profile page rank chart (mainly visible on mobile)
- move hover circle sizing to css variable
- use circle size for the extra hover area paddings
- also add extra vertical padding for hover area

I don't know if the extra vertical padding makes sense but the circle does extend when the value is at the edge...


resources/js/charts/line-chart.ts
```diff
@@ -114,6 +113,7 @@ export default class LineChart<X extends Date | number> {
   constructor(area: HTMLElement, public options: Options<X>) {
     this.area = select(area)
       .classed(classWithModifiers(bn, this.options.modifiers), true);
+    this.hoverCircleRadius = parseInt(this.area.style('--hover-circle-radius'), 10);
```
notbakaneko: this errors when navigating to the page via Turbo, presumably because whatever d3 uses is getting the style (or looking for the element, rather) from the wrong page

notbakaneko (changes requested):

notbakaneko (approved):

PR #13120, opened 2026-07-04
Disallow blocked user to comment under beatmapset page and reply comment

### LINKED ISSUE #12908: disallow blocked users make comments under your comments/beatmaps
Currently while you put user in blacklist he just can't dirrectly contact to you, but he can leave replies and make comments under your uploaded mapsets which is visible by any users (you just simply not receiving notification), is not good state as you don't wanna contact with blocked user and you also can't delete comments under your own maps which is also confusing

gagahpangeran:
closes #12908

### DIFF
--- a/app/Libraries/CommentBundle.php
+++ b/app/Libraries/CommentBundle.php
@@ -115,7 +115,7 @@ class CommentBundle
             'pinned_comments' => json_collection($pinnedComments, 'Comment'),
             'user_votes' => $this->getUserVotes($allComments),
             'user_follow' => $this->getUserFollow(),
-            'users' => json_collection($this->getUsers($allComments), 'UserCompact'),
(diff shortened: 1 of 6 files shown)


app/Libraries/CommentBundle.php
```diff
@@ -115,7 +115,7 @@ public function toArray()
             'pinned_comments' => json_collection($pinnedComments, 'Comment'),
             'user_votes' => $this->getUserVotes($allComments),
             'user_follow' => $this->getUserFollow(),
-            'users' => json_collection($this->getUsers($allComments), 'UserCompact'),
+            'users' => json_collection($this->getUsers($allComments), 'UserCompact', 'current_user_attributes'),
```
nanaya, line 118: this can't be done as is otherwise it'll do as many extra queries as there are the users

nanaya (changes requested):

PR #13123, opened 2026-07-07
Add subdivision for ranked play tiers

nanaya:
Not actually displayed anywhere yet.

The variable names don't seem right so if there's any better name suggestions...

([Same reference as before](https://discord.com/channels/90072389919997952/1458132886946451497/1520020010842591232))


Layendan:
I would rename (in order of appearance):
- `tiers` -> `ranks`
- `mainTitle` -> `currentRank` / `currRank`
- `levels` -> `tiers`
- `level` -> `currentTier` / `currTier`

You could remove current and just use `rank` and `tier`, but then I'd feel like that would be hard to read in the future

nanaya:
`Rank` is already used for the rank number.

Layendan:
You're right, that kinda sucks. The term `leagues` instead of ranks was thrown around in the discord as well, but idk if that's something the team wants to use as a final term.

pacowoc:
> You're right, that kinda sucks. The term `leagues` instead of ranks was thrown around in the discord as well, but idk if that's something the team wants to use as a final term.

"league" kinda imply that a gold player cannot be matched with silver player and vice versa.

pacowoc:
Also I think this is a good opportunity to make the thresholds more sane.

notbakaneko (approved):

PR #13124, opened 2026-07-07
Add endpoints for rating beatmaps

bdach:
Prerequisite for https://github.com/ppy/osu/issues/34735.

Behaviour intended to mirror correspondent osu-web-10 endpoint.


app/Http/Controllers/Beatmapsets/BeatmapsetRatingsController.php
```diff
@@ -0,0 +1,66 @@
...
+
+        $userRating = BeatmapsetUserRating::where([
+            'beatmapset_id' => $beatmapset->getKey(),
+            'user_id' => $user->getKey(),
+        ]);
+        if ($userRating->exists()) {
+            throw new InvariantException("You've already rated this beatmap set.");
+        }
+
+        $rating = get_int($request['rating']);
```
nanaya: can use `get_int(request('rating'))` if it's just one thing. Also the current array access will throw error if the parameter isn't passed.

Also might as well do this the first thing so no extra query is done if the parameter is invalid.

app/Http/Controllers/Beatmapsets/BeatmapsetRatingsController.php
```diff
@@ -0,0 +1,66 @@
...
+
+        if (!$beatmapset->isScoreable()) {
+            throw new InvariantException('Cannot rate this beatmap set.');
+        }
+
+        $userRating = BeatmapsetUserRating::where([
+            'beatmapset_id' => $beatmapset->getKey(),
+            'user_id' => $user->getKey(),
+        ]);
+        if ($userRating->exists()) {
```
nanaya: should update the existing rating instead? at least the web-10 version does that

app/Http/Controllers/Beatmapsets/BeatmapsetRatingsController.php
```diff
@@ -0,0 +1,66 @@
...
+            'rating' => $rating,
+            'user_id' => $user->getKey(),
+        ]);
+        self::recalculateBeatmapsetRating($beatmapset);
+        return response()->json([
+            'updated_rating' => $beatmapset->rating,
+        ]);
+    }
+
+    private static function recalculateBeatmapsetRating(Beatmapset $beatmapset)
```
nanaya: this looks like should be in the beatmapset model itself?

app/Http/Controllers/Beatmapsets/BeatmapsetRatingsController.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        $rating = BeatmapsetUserRating::where([
+            'beatmapset_id' => $beatmapset->getKey(),
+        ])->avg('rating');
+        $beatmapset->update(['rating' => $rating]);
+        $beatmapset->refresh();
```
nanaya: Not needed if only doing simple update.

app/Http/Controllers/Beatmapsets/BeatmapsetRatingsController.php
```diff
@@ -0,0 +1,66 @@
...
+
+        BeatmapsetUserRating::create([
+            'beatmapset_id' => $beatmapset->getKey(),
+            'rating' => $rating,
+            'user_id' => $user->getKey(),
+        ]);
+        self::recalculateBeatmapsetRating($beatmapset);
+        return response()->json([
+            'updated_rating' => $beatmapset->rating,
+        ]);
```
nanaya: just returning the array is enough

tests/Controllers/Beatmapsets/BeatmapsetRatingsControllerTest.php
```diff
@@ -0,0 +1,105 @@
...
+        $this->post(
+            route('api.beatmapsets.ratings.store', ['beatmapset' => $beatmapset->getKey()]),
+            ['rating' => 2]
+        )->assertStatus(422)
+            ->assertJson(['error' => 'Cannot rate this beatmap set.']);
+    }
+
+    /**
+     * @dataProvider dataProviderForInvalidRatings
+     */
```
nanaya: use the new `#[DataProvider...` attribute syntax thing

tests/Controllers/Beatmapsets/BeatmapsetRatingsControllerTest.php
```diff
@@ -0,0 +1,105 @@
...
+        $this->actAsScopedUser($user);
+
+        $this->post(
+            route('api.beatmapsets.ratings.store', ['beatmapset' => $beatmapset->getKey()]),
+            ['rating' => 2]
+        )->assertStatus(422)
+            ->assertJson(['error' => "You've already rated this beatmap set."]);
+    }
+
+    public static function dataProviderForInvalidRatings(): array
```
nanaya: static functions should be at the top

...although it's currently a mess because there's nothing enforcing it (and the previous data provider thing was not static function iirc)

tests/Controllers/Beatmapsets/BeatmapsetRatingsControllerTest.php
```diff
@@ -0,0 +1,105 @@
...
+        $this->actAsScopedUser($firstUser);
+        $request = $this->post(
+            route('api.beatmapsets.ratings.store', ['beatmapset' => $beatmapset->getKey()]),
+            ['rating' => 2]
+        );
+
+        $request->assertSuccessful()
+            ->assertJson(['updated_rating' => 2]);
+        $beatmapset->refresh();
+        $this->assertEquals(2, $beatmapset->rating);
```
nanaya: should use `assertSame` (and 2.0 in this case) unless there's some weird reason to use `assertEquals` (which uses php's amazing `==` loose comparison operator)

tests/Controllers/Beatmapsets/BeatmapsetRatingsControllerTest.php
```diff
@@ -0,0 +1,105 @@
...
+        $this->expectCountChange(fn () => $beatmapset->userRatings()->count(), 2);
+
+        $this->actAsScopedUser($firstUser);
+        $request = $this->post(
+            route('api.beatmapsets.ratings.store', ['beatmapset' => $beatmapset->getKey()]),
+            ['rating' => 2]
+        );
+
+        $request->assertSuccessful()
+            ->assertJson(['updated_rating' => 2]);
```
nanaya: can just chain the whole thing? also the other one down there. weirdly the other tests are already doing that

nanaya (changes requested):

app/Http/Controllers/Beatmapsets/BeatmapsetRatingsController.php
bdach: Unless I'm severely misreading something, the query in web-10 is `REPLACE INTO`, yes, but you can't actually change your rating. Stable doesn't let you, and the web-10 endpoint doesn't either (it early exits if already voted before the replace).

bdach:
Also I forgot to port the check that prevents the owner from voting on own beatmap set (should possibly extend to difficulty owners as well?)

Blocking, will resolve that and review tomorrow.

app/Http/Controllers/Beatmapsets/BeatmapsetRatingsController.php
nanaya: hmm right. I wonder if it's something for timing/toctou problem (or maybe just old code)

bdach:
Unblocking as I've done the above.

In addition I've also added an endpoint to retrieve the user rating as without it it is kind of impossible to know whether it is safe to rate a beatmap. It'll also come in handy for display purposes.

app/Http/Controllers/Beatmapsets/BeatmapsetRatingsController.php
```diff
@@ -0,0 +1,72 @@
...
+        if (!$beatmapset->isScoreable()) {
+            return ['allow_rating' => false];
+        }
+
+        $userRating = BeatmapsetUserRating::where([
+            'beatmapset_id' => $beatmapset->getKey(),
+            'user_id' => $user->getKey(),
+        ])->first();
+        return [
+            'allow_rating' => true,
```
nanaya: What does this "allow rating" mean? Is it in general? Otherwise this looks like should check the `BeatmapsetRate` result instead.

app/Singletons/OsuAuthorize.php
```diff
@@ -648,6 +648,38 @@ public function checkBeatmapsetNominate(?User $user, Beatmapset $beatmapset): st
...
+        }
+
+        if ($user->getKey() === $beatmapset->user_id) {
+            return $prefix.'owner';
+        }
+
+        $beatmapset->loadMissing('beatmaps.beatmapOwners');
+
+        foreach ($beatmapset->beatmaps as $beatmap) {
+            if ($beatmap->isOwner($user)) {
```
nanaya: can query beatmap/owner model directly instead of loading a bunch of extra stuff

Something like `BeatmapOwner::where('user_id', $user->getKey())->whereIn('beatmap_id', $beatmapset->beatmaps()->select('beatmap_id'))->exists()` (and similarly for beatmap)

And maybe combine the owner checks
```php
$isOwner = $user->getKey() === $beatmapset->user_id;
// or as single assignment
$isOwner = $isOwner || $beatmapset->beatmaps()->where...;
$isOwner = $isOwner || BeatmapOwner::...;

if (...) { ... } // or return $isOwner ? ... : ...
```

Also I just noticed it's copied from `checkBeatmapsetNominate` 😐 Maybe move the logic to the beatmapset model.

app/Http/Controllers/Beatmapsets/BeatmapsetRatingsController.php
```diff
@@ -0,0 +1,72 @@
...
+        $user = \Auth::user();
+
+        priv_check('BeatmapsetRate', $beatmapset)->ensureCan();
+
+        $userRating = BeatmapsetUserRating::where([
+            'beatmapset_id' => $beatmapset->getKey(),
+            'user_id' => $user->getKey(),
+        ]);
+        if ($userRating->exists()) {
+            throw new InvariantException("You've already rated this beatmap set.");
```
nanaya: I assume this and the rating value validation aren't supposed to fail in normal case so they don't need to be translated?

tests/Controllers/Beatmapsets/BeatmapsetRatingsControllerTest.php
```diff
@@ -0,0 +1,200 @@
...
+            ->assertJson([
+                'allow_rating' => true,
+                'total_rating' => 4.37,
+            ]);
+    }
+
+    public function testIndexWithUserRating()
+    {
+        $beatmapset = Beatmapset::factory()->create([
+            'approved' => 1,
```
nanaya: there's the `->ranked()`/`->pending()` etc factory

nanaya (changes requested):

app/Http/Controllers/Beatmapsets/BeatmapsetRatingsController.php
bdach: The intention was to give the client *some* indication that it shouldn't even bother to display the rating control. I guess *maybe* it's not strictly necessary because the client should know the beatmap set status, and the other failing cases (being the owner, etc.) will loudly fail on attempts with messages which is probably good for conveying why rating is impossible? Dunno. Will probably end up just removing.

app/Singletons/OsuAuthorize.php
bdach: I'm nowhere near enough fluent in eloquent-fu to know how to do this most efficiently and the documentation is terrible so yes I did just ^C^V from another place that did what I needed.

app/Http/Controllers/Beatmapsets/BeatmapsetRatingsController.php
nanaya: maybe include the priv check result message so the client can show why it can't be rated and doesn't need to do the check manually?

although the check doesn't include existing rating at the moment so it's not quite complete...

bdach: > maybe include the priv check result message so the client can show why it can't be rated and doesn't need to do the check manually?

Works for me.

> although the check doesn't include existing rating at the moment so it's not quite complete...

That's technically fine, the endpoint returns `user_rating`. If it's not null then the user's already voted, so that can be jury-rigged client-side. And that'd require different handling anyway (mirroring stable, client would want to show the client the user's already-set rating but in a disabled state)

app/Http/Controllers/Beatmapsets/BeatmapsetRatingsController.php
bdach: I'll add a translation. Odds are it'll come handy in client anyhow.

app/Http/Controllers/Beatmapsets/BeatmapsetRatingsController.php
peppy: Not sure where we stand on this, but I don't think there's too much harm in allowing users to update their rating after a valid play.

bdach: I can remove the restriction of not being able to change the vote, but "after a valid play" is probably going to need to be completely enforced by the client like it is currently the case with tags.

peppy: If it's a client only endpoint with signing then that's probably fine.

app/Singletons/OsuAuthorize.php
bdach: I'm admittedly a bit scared to be touching nomination code, but I've given this a shot, see https://github.com/ppy/osu-web/pull/13124/commits/2b5191410a16112cb8027c9a24afd48b055dc561.

Of particular note, I noticed that tests of my thing here *and* of nomination don't really cover `BeatmapOwner`. The factory methods involves set `user_id` on the `osu_beatmaps` rows. Therefore this also comes with extra coverage to test both paths (https://github.com/ppy/osu-web/pull/13124/commits/28e86e894db86ac4669bf86b80b6a36f9b2e59ee).

Note that I am relying *very* heavily on the automated testing and have not manually tested this. If you want me to do that, let me know.

app/Http/Controllers/Beatmapsets/BeatmapsetRatingsController.php
bdach: Changing rating is now allowed as of https://github.com/ppy/osu-web/pull/13124/commits/3dfd5799b502e95ac00ab09738a228cb89ca921e

app/Singletons/OsuAuthorize.php
nanaya, line 1: a bit silly but the `BeatmapsetRate` probably should be added to the `alwaysCheck`.

bdach: I tested and it kind of doesn't actually matter as long as this is an OAuth-only endpoint because of

https://github.com/ppy/osu-web/blob/25e080540e1b1a35616b3e24d92dde57387164cf/app/Models/User.php#L1185-L1197

Means that in

https://github.com/ppy/osu-web/blob/2b5191410a16112cb8027c9a24afd48b055dc561/app/Singletons/OsuAuthorize.php#L88

the conditional always fails on `->isAdmin()` and `alwaysCheck()` is never even called.

I can add it for future-proofing though if you think there's a chance this stops being OAuth-only.

nanaya: I guess it's probably fine... if there's a client token check which should probably be added. Although having it anyway may be better

nanaya (approved):

PR #13125, opened 2026-07-07
Add markdown component for user link

### LINKED ISSUE #8932: On-hover user card display for profile links on the wiki
BBCode `[profile]` tags allow for profile links that display a user card when hovering over them, so I thought there could be a tag with the same functionality for use on the wiki and in news post, for the many player listings that can be seen there.

Of course, an alternative solution would be to have any profile link display a user card, since they have a pretty standard format (`osu.ppy.sh/users/{ID}` or `osu.ppy.sh/u/{ID}` if I'm not missing any). To me at least, there's no reason not to display a user card when hovering over a profile link, since it's so ubiquitous in other parts of the website.

gagahpangeran:
Closes #8932 

Syntax:

```
::username::{ user=1234 }
```

Preview:

image

But really, the `username` is simply just plain text, so you can add any inline formatting. The only important attribute is the `user-id`.

This custom inline container with `user-id` attribute will be rendered as link with on-hover user card based on its `user-id`.

Example with formatting:

```
::**bold** _italic_::{ user=1234 }
```


gagahpangeran:
cc : @Walavouchey @Hiviexd

Walavouchey:
also cc @peppy because the wiki-side change is very large

notbakaneko:
I think just `user=` is simpler for wiki editors? It also works with usernames iirc, although we don't recommend using usernames.

gagahpangeran:
> I think just `user=` is simpler for wiki editors? It also works with usernames iirc, although we don't recommend using usernames.

Good idea.

@Walavouchey do we want to support using username or just the user id is okay? (something like `::name::{user="username"}`)

My opinion is we avoid using direct username entirely so it is simpler when the user change their username.

Walavouchey:
supporting username lookup risks links eventually linking to the wrong user. there's no use case for it in the wiki

gagahpangeran:
@Walavouchey I change the syntax `user` attribute instead of `user-id`.

peppy:
Can we not just go with `::{user="12345678"}::`? Do we need hardcoded usernames? There have been updates to the wiki in the past which basically just updated usernames which have changed, which would support *not* hardcoding them, I think.

gagahpangeran:
> Can we not just go with `::{user="12345678"}::`? Do we need hardcoded usernames? There have been updates to the wiki in the past which basically just updated usernames which have changed, which would support _not_ hardcoding them, I think.

From https://github.com/ppy/osu-web/issues/8932#issuecomment-4888934008

> name should be required because of namechages and the wiki intentionally keeps original names for history preservation IIRC

For example this page https://osu.ppy.sh/wiki/en/History_of_osu%21/Online_rankings/osu%21 is using the old username that player using at that time.

peppy:
In that case, I'm not too sure about the formatting of the block. Is there precedent for it looking like this from other usages? 

If it's standard for us then it's probably ok.

Walavouchey:
> Can we not just go with `::{user="12345678"}::`? Do we need hardcoded usernames? There have been updates to the wiki in the past which basically just updated usernames which have changed, which would support _not_ hardcoding them, I think.

apart from displaying old usernames, there are also plenty of cases where people write a phrase referring to someone and link to their profile (mainly in news posts)

we could support both styles and default to the dynamic version, but removing the username presents at least three problems:

- the markdown source becomes unreadable (maybe a minor issue)
- preview tools will need to perform a bunch of api lookups
- if the id has a typo and it's not caught immediately, who knows down the line which user it was *meant* to be linking to, and we wouldn't be able to automatically check for mismatches either

> In that case, I'm not too sure about the formatting of the block. Is there precedent for it looking like this from other usages?

we don't have usages of inline custom containers + attribute lists, but it's apparently in the [markdown spec](https://github.com/xoofx/markdig/blob/main/src/Markdig.Tests/Specs/CustomContainerSpecs.md#inline-custom-container)

Hiviexd:
I initially proposed something like `::{user=12345678 name="username"}::` which feels more intuitive and in line with existing syntax (flags) IMO.

Walavouchey:
> I initially proposed something like `::{user=12345678 name="username"}::` which feels more intuitive and in line with existing syntax (flags) IMO.

this one should also have the advantage of not having to escape special markdown characters in usernames

notbakaneko:
I'd rather have everything in the attribute, it makes it simpler for editors; this custom element was implemented so that we have control over the markup output and transform attributes into the appropriate elements.

Omitting the username also means the rendering has to do lookups to insert the username, and it will also probably be deferred to perform all the lookups in one batch, which means another workaround to reinject the username into the parsed output...

gagahpangeran:
My intention to go with this syntax is so the text (or username in this case) can be used with another inline formatting. For example in [this page](https://osu.ppy.sh/wiki/en/History_of_osu%21/Online_rankings/osu%21).

`[WubWoofWolf (*White Wolf*)](https://osu.ppy.sh/users/39828)`

Can be just change to this component with basically the same structure.

`::WubWoofWolf (*White Wolf*)::{user=39828}`

> I'd rather have everything in the attribute, it makes it simpler for editors; this custom element was implemented so that we have control over the markup output and transform attributes into the appropriate elements.

For me, it feels weird to write the username (that is basically can be any text with formatting) in the attribute itself.

And how about the text formatting like the example above?

> Omitting the username also means the rendering has to do lookups to insert the username, and it will also probably be deferred to perform all the lookups in one batch, which means another workaround to reinject the username into the parsed output...

I don't think there will be any case for omitting the username.

---

Honestly, if we just want to **make all** user link in wiki to have on-hover usercard. We can just add the `js-usercard` class and `data-user-id` attribute via `DocumentProcessor.php`. We don't have to invent or use another syntax or custom attribute for this. I think this is more simple and straightforward.

notbakaneko:
I think the wiki should always use user id lookups since usernames change.

I don't think omitting the username or just using usernames is a good idea since it'd require the deferred lookups and re-injecting the names that was mentioned above. The difference with the `[profile]` bbcode is the bbcode is updated and replaced during save, while the markdown has to do the lookup on render.

If the current syntax is fine then we'll just go with it.

Walavouchey:
> Honestly, if we just want to **make all** user link in wiki to have on-hover usercard. We can just add the `js-usercard` class and `data-user-id` attribute via `DocumentProcessor.php`. We don't have to invent or use another syntax or custom attribute for this. I think this is more simple and straightforward.

i'd also say that this is a valid and sensible direction, as mentioned in the linked issue

gagahpangeran:
> If the current syntax is fine then we'll just go with it.

I'm fine with the current syntax in the OP. cc: @Walavouchey

tests/Libraries/Markdown/html_markdown_examples/attribute_user_id.md
```diff
@@ -0,0 +1,5 @@
+::peppy::{user=2}
```
notbakaneko: `::{ user=1234 }username::` attributes inside is also valid syntax (and possibly more convenient)

notbakaneko (changes requested):

app/Libraries/Markdown/CustomContainerInline/Renderer.php
```diff
@@ -5,35 +5,81 @@
...
+
+        return new HtmlElement('span', $this->attrs->export(), '');
+    }
+
+    private function createProfileElement(string $userId)
+    {
+        $this->attrs->remove('user');
+
+        $blockClass = $this->config->get('osu_extension/block_name');
+        $this->attrs->set('class', "{$blockClass}__link js-usercard");
```
notbakaneko: Should use `->append('class', 'js-usercard')` instead, the default wiki style should be set in `OsuMarkdown::createDefaultAttributesConfig`.
Also, I don't think `__link` actually has any associated style with it, so it wouldn't be needed anyway.

tests/Libraries/Markdown/html_markdown_examples/attribute_user_id.md
Walavouchey: time for polls i guess.. it feels a bit more self-contained but also a bit clunky with the right brace directly against the username

spaces should go inside the braces in any case, since we do that with id containers

notbakaneko: well, the attribute syntax works like that, so both `::{ user=1 }username::` and `::username::{ user=1 }` will be usable

notbakaneko (approved):

PR #13127, opened 2026-07-08
Move out handling of beatmapset description

nanaya:
An attempt at reducing the size of the beatmapset model 🤷 

Also:
- use const for bbcode options
- use const and simple string split for the separator

### DIFF
--- a/app/Libraries/Beatmapset/Description.php
+++ b/app/Libraries/Beatmapset/Description.php
@@ -1,0 +1,84 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+declare(strict_types=1);
+
+namespace App\Libraries\Beatmapset;
+
+use App\Libraries\BBCodeFromDB;
+use App\Models\Beatmapset;
+use App\Models\Forum;
+
+class Description
+{
+    const BBCODE_OPTIONS = [
+        'ignoreLineHeight' => true,
+        'withGallery' => true,
+    ];
+
+    const SEPARATOR = '---------------';
+
+    public function __construct(private readonly Beatmapset $beatmapset)
+    {
+    }
+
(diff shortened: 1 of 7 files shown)


app/Models/Forum/Post.php
```diff
@@ -172,7 +172,7 @@ public function getSearchContentAttribute()
         // unescape html entities
         // strip remaining bbcode
         // strip any html tags left
-        $text = Beatmapset::removeMetadataText($this->post_text);
+        $text = BeatmapsetDescription::removeMetadataText($this);
```
nanaya: why is this done regardless whether or not it's a beatmapset description ಠ_ಠ

and is it even needed

app/Libraries/Beatmapset/Description.php
```diff
@@ -0,0 +1,84 @@
...
+
+    public function bbcode(): BBCodeFromDB
+    {
+        $post = $this->beatmapset->descriptionPost ?? new Forum\Post();
+        $description = trim(static::splitText($post)[1] ?? '');
+
+        return new BBCodeFromDB($description, $post->bbcode_uid, static::BBCODE_OPTIONS);
+    }
+
+    public function update($bbcode, $user)
```
notbakaneko: might as well type hint these?

app/Libraries/Beatmapset/Description.php
```diff
@@ -0,0 +1,84 @@
...
+            $split = static::splitText($post);
+
+            $header = new BBCodeFromDB($split[0], $post->bbcode_uid, static::BBCODE_OPTIONS);
+            $newBody = $header->toEditor().static::SEPARATOR."\n".ltrim($bbcode);
+
+            return $post
+                ->skipBeatmapPostRestrictions()
+                ->update([
+                    'post_text' => $newBody,
+                    'post_edit_user' => $user === null ? null : $user->getKey(),
```
notbakaneko: can use an update to `?->getKey()`; ...is it even possible for `$user` to be null? 🤔

notbakaneko (approved):

PR #13130, opened 2026-07-09
Transparent avatar support.

kevinbudz:
Upon uploading a profile picture that has any alpha/transparency, it immediately hides the gray-ish avatar plate that would normally be placed below it.

Adds a nullable `has_alpha` column on `phpbb_users`. It is set to false by default, and triggers when uploading a new profile picture (.png/gif) that has any non-opaque pixels.

This is a very negligible change. But, it does look pretty for any user who would take advantage of it. Below is an imgur link with relevant screenshots.

https://imgur.com/a/Zi2Gdfk


gagahpangeran:
IMO, I think the simpler solution is adding checklist options on avatar section in account settings. Something like "use transparent avatar background".

kevinbudz:
> IMO, I think the simpler solution is adding checklist options on avatar section in account settings. Something like "use transparent avatar background".

I'd be willing to adapt the PR to fit that approach if need be.

PR #13132, opened 2026-07-09
Return score notice as part of initial oauth endpoint

nanaya:
To inform client about ongoing process.

Changed the message to a static one so it can be translated.

- [x] set the `SCORE_PROCESSING_NOTICE_URL` env (if still relevant) @peppy


notbakaneko (approved):

notbakaneko:
apparently I press approved on the wrong tab 🙈

notbakaneko (changes requested):
and now it conflicts 🙃

notbakaneko (approved):

PR #13141, opened 2026-07-10
Simplify handling of token parsing failure

nanaya:
The beatmap rating and client side medal are fine with 422 according to @bdach (well the former isn't implemented yet iiuc)

The registration endpoint already contains token validation down the path and returning 422 on failure. I'm not quite sure why it's returning the use-the-web response instead. Or maybe it made sense before the additional token validation thing was added (something with the older client not sending the token maybe?).


notbakaneko (approved):

bdach:
Just to be sure I double-checked full-stack and both the medal awarding endpoint and client registration should indeed be fine with this change.

PR #13142, opened 2026-07-10
Hide score notice on bot profiles

LiquidPL:
It's unnecessary since bots don't have ranks or scores, also it looks kinda bad:

image

### DIFF
--- a/resources/js/profile-page/detail.tsx
+++ b/resources/js/profile-page/detail.tsx
@@ -42,10 +42,10 @@ export default class Detail extends React.Component<Props> {
         {!user.is_bot && (
           <div className='profile-detail'>
             <DetailStats user={user} />
+
+            {this.renderScoresNotice()}
           </div>
         )}
-
-        {this.renderScoresNotice()}
 
         <DetailBar user={user} />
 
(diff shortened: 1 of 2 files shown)


diquoks:
i think it's also worth doing a similar PR in lazer

resources/js/profile-page/detail.tsx
```diff
@@ -45,7 +45,7 @@ export default class Detail extends React.Component<Props> {
           </div>
         )}
 
-        {this.renderScoresNotice()}
+        {!user.is_bot && this.renderScoresNotice()}
```
nanaya: probably better to move it inside the `profile-detail` above

nanaya (dismissed):

resources/js/profile-page/detail.tsx
```diff
@@ -39,13 +39,12 @@ export default class Detail extends React.Component<Props> {
...
 
-        {!user.is_bot && (
+        {!user.is_bot && (<>
           <div className='profile-detail'>
             <DetailStats user={user} />
           </div>
-        )}
-
-        {this.renderScoresNotice()}
+          {this.renderScoresNotice()}
```
nanaya: I mean inside the div.

notbakaneko (approved):

PR #13145, opened 2026-07-12
Fix line chart hover

### LINKED ISSUE #13144: Line chart hover is broken when navigating via Turbo
Originally [reported on discord](https://discord.com/channels/188630481301012481/188630652340404224/1525689024692486408).

https://github.com/user-attachments/assets/93387649-fc66-4d8f-b24a-a63070d6a31e

---

Regressed in #13119 

When opening the user profile page via Turbo, the line chart hover is broken. It works if we open it directly or refreshing the page.

After some investigation, the value of `this.hoverCircleRadius` in `line-chart.ts` becomes `NaN` and makes all other value that uses it to also become `NaN`. My guess is there's something wrong when getting the `area` ref when navigating via Turbo.

---

Here is some `console.log` output.

```diff
diff --git a/resources/js/charts/line-chart.ts b/resources/js/charts/line-chart.ts
index 056f2237dd..a48624fc89 100644
--- a/resources/js/charts/line-chart.ts
+++ b/resources/js/charts/line-chart.ts
@@ -114,6 +114,9 @@ export default class LineChart<X extends Date | number> {
     this.area = select(area)
       .classed(classWithModifiers(bn, this.options.modifiers), true);
     this.hoverCircleRadius = parseInt(window.getComputedStyle(area).getPropertyValue('--hover-circle-radius'), 10);
+    console.log(area);
+    console.log(window.getComputedStyle(area));
+    console.log(this.hoverCircleRadius);
 
     this.svg = this.area.append('svg');
```

Access the page directly (full reload).

Image

Via Turbo.

Image

notbakaneko:
closes #13144

would help if my test chart had more than 1 point 👀

### DIFF
--- a/resources/js/charts/line-chart.ts
+++ b/resources/js/charts/line-chart.ts
@@ -11,7 +11,6 @@ import {
   extent,
   Line,
   line,
-  pointer,
   ScaleContinuousNumeric,
   scaleLinear,
   ScaleTime,
@@ -23,7 +22,6 @@ import { clamp } from 'lodash';
 import core from 'osu-core-singleton';
 import { classWithModifiers, Modifiers } from 'utils/css';
 import { fadeIn, fadeOut } from 'utils/fade';
-import { getInt } from 'utils/math';
 
 const bn = 'line-chart';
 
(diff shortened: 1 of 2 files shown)


gagahpangeran (approved):
It works.

resources/js/charts/line-chart.ts
```diff
@@ -270,6 +270,9 @@ export default class LineChart<X extends Date | number> {
   }
 
   private readonly onHover = (event: DragEvent | MouseEvent) => {
+    this.hoverCircleRadius ??= getInt(this.area.style('--hover-circle-radius'));
+    if (this.hoverCircleRadius == null) return;
+
     const relativeX = pointer(event)[0] - this.hoverCircleRadius;
```
nanaya: apparently can do this instead?

```suggestion
    const relativeX = event.clientX - (this.hover.node()?.getBoundingClientRect().left ?? 0);
```

(or `pointer(event, this.hover.node())[0]` but it's internally doing more or less the same thing)

nanaya (commented):
there's now irrelevant comment in the `line-chart.less` 👀

nanaya (approved):

PR #13150, opened 2026-07-14
Add endpoint to generate one time key thing

nanaya:
Something event something. Only to link a string with user id.

https://discord.com/channels/90072389919997952/1458132886946451497/1526427566313967686

The endpoints and return format have since been changed.

```
GET /one-time-key -> create the key (now with button and existing key lookup)
POST /one-time-key/check (parameter: key (string)) -> obtain user id
```

toggle env:
```
OSU_ONE_TIME_KEY=false
```

Removed the key splitting because I miscounted the available bits 🙃

### DIFF
--- a/app/Http/Controllers/OneTimeKeyController.php
+++ b/app/Http/Controllers/OneTimeKeyController.php
@@ -1,0 +1,80 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+declare(strict_types=1);
+
+namespace App\Http\Controllers;
+
+use App\Libraries\OneTimeKey;
+use App\Models\Beatmap;
+use App\Models\User;
+use App\Transformers\UserCompactTransformer;
+
+class OneTimeKeyController extends Controller
+{
+    public function __construct()
+    {
+        $this->middleware('auth', ['except' => ['check']]);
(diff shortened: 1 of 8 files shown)


notbakaneko (commented):
probably should have a page title that's not `unknown`? 🤔

peppy:
cc @smoogipoo

app/Http/Controllers/OneTimeKeyController.php
```diff
@@ -0,0 +1,63 @@
...
+            $matchmakingUserStatsJson[] = [
+                'pool_id' => $entry->pool_id,
+                'rating' => $entry->rating,
+                'ruleset_id' => $entry->pool->ruleset_id,
+                'variant_id' => $entry->pool->variant_id,
+            ];
+        }
+
+        return [
+            'user' => json_item($user, new UserCompactTransformer(), UserCompactTransformer::CARD_INCLUDES),
```
smoogipoo: Can we also include the user's pp here? It would probably need to be something like:
```
'stats': [
  {
    'ruleset_id': 0,
    'pp': 0
  },
  ...
  {
    'ruleset_id': 3,
    'variant': 4,
    'pp': 0
  },
  {
    'ruleset_id': 3,
    'variant': 7,
    'pp': 0
  },
]
```

smoogipoo (changes requested):

notbakaneko (approved):

smoogipoo (approved):
Have been using this in my local testing with no problems 👍

PR #13155, opened 2026-07-17
Add some sort of Device Code OAuth

nanaya:
Conjuring one up manually because the alternative is updating laravel passport which involves updating a bunch of other things. And then the special handling to allow the scope.

It's not very compliant. It doesn't even accept client id but instead has client token check ┐(･.･)┌

Tested with https://oauthplayground.io/

(the background is it's needed for a different part of the event. It's not related to the other one in term of the system. @smoogipoo has more details)

the page design is even more temporary because I don't know if this idea can even pass sanity check

```
create an oauth client and set the id in the env
API_DEVICE_AUTH_CLIENT_ID=

device authorization url:
POST /device/authorisation

token url:
POST /device/token
```


resources/views/device_auth/create.blade.php
```diff
@@ -0,0 +1,67 @@
...
+            </div>
+
+            <div class="dialog-form__row dialog-form__row--title">
+                <div class="dialog-form__logo"></div>
+                <h1 class="dialog-form__title">
+                    Authorise Device
+                </h1>
+            </div>
+
+            <form
```
notbakaneko, line 45: need `data-remote` for the error popup?

app/Libraries/OAuth/DeviceAuth.php
```diff
@@ -0,0 +1,80 @@
...
+            'expires_at' => time() + $expiresIn,
+        ], $expiresIn);
+
+        return [
+            'device_code' => $deviceCode,
+            'expires_in' => $expiresIn,
+            'interval' => 5,
+            'user_code' => $userCode,
+            'verification_uri' => route('device-auth'),
+            'verification_uri_complete' => route('device-auth', ['user_code' => $userCode]),
```
notbakaneko, line 30: does this version of the url have a use?

notbakaneko (changes requested):

app/Libraries/OAuth/DeviceAuth.php
nanaya: it's part of the spec

also the page itself will be updated to show the code as static text instead of input box if the parameter is passed

notbakaneko (changes requested):
I also think the page to enter the code in needs more warnings and messages on what's it for, especially seeing as this flow is granting a `*` token

nanaya:
This doesn't end up being immediately needed. May redo this with updated passport instead if still needed.

PR #13157, opened 2026-07-18
Fill in missing beatmapset event types for search

cl8n:
seems like just an oversight that these weren't written into any type categories


notbakaneko (changes requested):
localisations need to be added under `beatmapset_events.type` as well

notbakaneko (approved):

notbakaneko (approved):

PR #13163, opened 2026-07-21
Move mobile notifications to navigation bar

### LINKED ISSUE #12890: Make notifications more visible on mobile browsers

notbakaneko:
Opted to separate and rework the logic for the mobile menu to make it behave the way we want, instead of trying to squeeze it into the existing click menu.

closes #12890
alternative to #13078

### DIFF
--- a/resources/js/core/mobile-menu.ts
+++ b/resources/js/core/mobile-menu.ts
@@ -8,18 +8,19 @@ import { htmlElementOrNull } from 'utils/html';
 import { isModalShowing } from 'utils/modal-helper';
 
 const menuActiveClass = 'js-click-menu--active';
-const menuElementSelector = '.js-mobile-menu[data-mobile-menu-id]';
-const menuTriggerSelector = '.js-mobile-menu[data-mobile-menu-target]';
+const menuButtonSelector = '.js-mobile-menu--button';
+const menuElementSelector = '.js-mobile-menu';
 const menuRootElementSelector = '.js-mobile-menu--root';
+const navActiveClass = 'js-nav2--active';
 
 function closestMenuId(child: Element | null | undefined) {
   return child != null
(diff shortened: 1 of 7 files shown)


resources/js/core/mobile-menu.ts
```diff
@@ -0,0 +1,155 @@
...
+  private documentMouseEventTarget: unknown;
+
+  constructor() {
+    window.addEventListener('resize', this.handleResize);
+    $(document).on('click', '.js-click-menu--close', this.handleCloseClick);
+    $(document).on('click', menuButtonSelector, this.handleMenuButtonClick);
+    $(document).on('mousedown', this.handleDocumentMousedown);
+    $(document).on('mouseup', this.handleDocumentMouseup);
+    document.addEventListener('turbo:load', this.restoreSaved);
+    document.addEventListener('turbo:before-cache', this.saveCurrent);
```
nanaya, line 37: pretty sure these are only relevant for react-rendered menu

resources/js/core/mobile-menu.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  toggle(target?: string | null) {
+    const actualTarget = (target != null ? menuElement(target)?.dataset.mobileMenuDefault : null) ?? target;
```
nanaya, line 41: I suspect this class would be a lot simpler if it only handles the notification/chat/caret menu (and the overlay) and has the original click menu handles everything else

nanaya (changes requested):

PR #13164, opened 2026-07-22
Add API endpoints for retrieving and modifying user preferences

bdach:
The use case for this is coming from https://github.com/ppy/osu/issues/36527:

> Optimally, sync the setting from osu-web (and other settings using a combined api?)

My starting proposition was https://github.com/ppy/osu/issues/36527#issuecomment-4968547055. As it turns out most of the work was already done. This PR is mostly exposing one extra attribute on `GET /api/v2/me` and exposing an existing website endpoint to also work via `PUT /api/v2/me/options`. I could not see any reason preventing just reusing it - it doesn't seem like that endpoint allows doing something that shouldn't be done by client - but that does not mean there isn't one, so reviewer beware.

I've briefly tested this full-stack with a hastily-written client-side diff and it seems to be enough to make work.

The test is very paranoid and probably unnecessary, I just *really* wanted to make sure I'm not accidentally exposing stuff too widely.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -604,6 +604,8 @@ Route::group(['as' => 'api.', 'prefix' => 'api', 'middleware' => ['api', Throttl
         Route::get('me/download-quota-check', 'HomeController@downloadQuotaCheck')->name('download-quota-check');
         //  GET /api/v2/me
         Route::get('me/{mode?}', 'UsersController@me')->name('me');
+        //  PUT /api/v2/me/options
+        Route::put('me/options', 'AccountController@updateOptions')->name('me.options');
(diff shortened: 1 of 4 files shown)


routes/web.php
```diff
@@ -604,6 +604,8 @@
         Route::get('me/download-quota-check', 'HomeController@downloadQuotaCheck')->name('download-quota-check');
         //  GET /api/v2/me
         Route::get('me/{mode?}', 'UsersController@me')->name('me');
+        //  PUT /api/v2/me/options
```
nanaya, line 613: I don't know if this comment is still useful nowadays

nanaya (approved):

PR #13166, opened 2026-07-22
Adjust font weight for difficulty badge text

nanaya:
Matches client... apparently?


notbakaneko (approved):

stysan:
i kinda feel like because of browser rendering 800 would match more

PR #13167, opened 2026-07-22
Make scores on profile page clickable

nanaya:
As requested offline.


notbakaneko:
This has a problem on mobile where tapping anywhere on any score to make a tooltip go away navigates to the score.

nanaya:
then don't tap on the score...? 🤔 

tapping on any links navigates away from the page already so it doesn't change that behaviour

notbakaneko:
kind of hard to tap on anything else when the whole screen is just scores and more navigation :|

nanaya:
fwiw the page title can be tapped. or maybe add a new behavior of tapping the same thing to hide title

or just scroll away somewhere else...

peppy:
Most of the time on mobile you'd be dragging anyway? Not sure it's a huge issue, having dealt with many other pages that work similarly.

notbakaneko (approved):
🤷

Issue #13170, opened 2026-07-22
SR number and difficulty name misaligned


longnguyen2004:
Example: https://osu.ppy.sh/beatmapsets/2239198#taiko/4758488, look at the SR pill and difficulty name

<img width="206" height="42" alt="Image" src="https://github.com/user-attachments/assets/04b8363e-4e8e-4419-951e-167b64b5d321" />

The SR number is left aligned instead of center aligned, and the pill + difficulty text is bottom aligned instead of middle aligned

Here I added `text-align: center` to `.difficulty-badge__rating` and `vertical-align: middle` to `.difficulty-badge--beatmapset` and it seems to look better?

<img width="199" height="41" alt="Image" src="https://github.com/user-attachments/assets/25e1a8db-f5ef-48d3-991e-741772021e1d" />

Joehuu:
SR being misaligned is a duplicate of https://github.com/ppy/osu-web/issues/8806.

nanaya:
Vertical alignment has been fixed and horizontal alignment is a duplicate issue as mentioned above.

PR #13171, opened 2026-07-22
Update vertical alignment in diff badge

stysan:
Sort of adresses #13170.  

before & after - diff badge is now more centered with diff name, showcased on osu.ppy.sh/b/782310


cloudrac3r:
I suspect the actual correct alignment here is `baseline`.

nanaya (approved):

PR #13172, opened 2026-07-22
Adjust spacing and sizing in details on beatmapset page

prplnorangesoda:
master

 PR
image
</table>

I'm a first time contributor. No AI was used for this.

Wrapped the badge, name and diff owner in a container to make it easier to distance the whole box from everything. also wrapped the text components in a div to more easily margin them closer.
Resizes diff name from 17px to 16px to be able to center the badge

### DIFF
--- a/resources/css/bem-index.less
+++ b/resources/css/bem-index.less
@@ -72,6 +72,7 @@
 @import "bem/beatmapset-beatmap-picker";
 @import "bem/beatmapset-cover";
 @import "bem/beatmapset-cover-admin";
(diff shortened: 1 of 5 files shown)


prplnorangesoda:
dupe of #13171 (my bad), but that pr has a different implementation than this one

stysan:
way better of a solution than what I had

nanaya:
Went with the simpler change for now. Also this one has problem of the badge moving around if the name section is too long

PR #13176, opened 2026-07-24
Adjust entry time filter when creating new build

nanaya:
Hopefully fixes the problem of missing changelog entry on builds 🤷

### DIFF
--- a/app/Models/Build.php
+++ b/app/Models/Build.php
@@ -79,7 +79,9 @@ class Build extends Model implements CommentableInterface
             'allow_bancho' => $stream->default_allow_bancho,
         ]);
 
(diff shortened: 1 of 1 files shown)


notbakaneko:
Is there an example of this?

nanaya:
the most recent tachyon release has `created_at: "2026-07-23T11:56:55Z"` in the payload but the related commit/pr where it was tagged has `merged_at: "2026-07-23T11:56:56Z"`, which is set in the [changelog entry](https://github.com/ppy/osu-web/blob/master/app/Models/ChangelogEntry.php#L133) and thus wasn't included as part of the build.

notbakaneko (approved):
:psy:

PR #13179, opened 2026-07-25
Fix pinned score reordering on User Profile page not working

### LINKED ISSUE #13177: Unable to Reorder Pins on Profile Page
When on the user profile page, hovering over the three bars to change the position of a pinned score indicates that I am hovering over a link to the score. I can click to go to the score details, but clicking and dragging does not allow me to reorder the pinned score. My best guess is that #13167 made the clickable area include too much, but I am not familiar enough with the codebase to make the PR to fix it myself.

Steps to reproduce:
1. Be signed in to the osu website on an account with more than 1 pinned score
2. Click and drag the three bars as in my screenshot below to try and change the location of that pin in the list

Expected behavior:
The pin's position in the list gets changed to reflect where it was dragged

Current behavior:
The pin's position in the list does not change.

Screenshot:

Image

notbakaneko:
closes #13177

### DIFF
--- a/resources/css/bem/sortable-handle.less
+++ b/resources/css/bem/sortable-handle.less
@@ -16,6 +16,7 @@
     display: none;
 
     @media @desktop {
+      position: relative;
       display: flex;
       align-items: center;
       width: var(--pin-sortable-handle-width);


resources/css/bem/sortable-handle.less
```diff
@@ -19,6 +19,9 @@
       display: flex;
       align-items: center;
       width: var(--pin-sortable-handle-width);
+      // keep above score link.
+      // position: absolute doesn't work because the sortable only works within the element's box.
+      z-index: 1;
     }
```
nanaya, line 23: a relative position is probably better. or resizing the bg link itself using the handle width variable

nanaya (changes requested):

resources/css/bem/sortable-handle.less
notbakaneko: that's weird, it didn't work before ?_?

nanaya (approved):

PR #13183, opened 2026-07-27
Add new profile page design

nanaya:
So about that 100 tier badge components... 👀

May or may not split the pr later.

Depends on
- [x] #13182
- [x] #13123
- [x] #13188


notbakaneko:
That's a lot of ~~changes~~files 👀

app/Models/MatchmakingUserStats.php
```diff
@@ -74,6 +74,14 @@ public function scopeWhereRulesetId(Builder $query, int $rulesetId): Builder
         return $query->whereHas('pool', fn ($q) => $q->where('ruleset_id', $rulesetId));
     }
 
+    public function history(): Builder
+    {
+        return MatchmakingUserEloHistory::where([
+            'pool_id' => $this->pool_id,
+            'user_id' => $this->user_id,
+        ]);
+    }
```
notbakaneko, line 83: Why builder instead of relation?

app/Models/MatchmakingUserEloHistory.php
```diff
@@ -29,6 +30,20 @@ class MatchmakingUserEloHistory extends Model
 {
     protected $table = 'matchmaking_user_elo_history';
 
+    public static function daily(int $poolId, int $userId, int $days): Collection
+    {
+        $ids = static
+            ::where(['pool_id' => $poolId, 'user_id' => $userId])
+            ->groupBy('create_date')
+            ->selectRaw('MAX(id) last_id, DATE(created_at) create_date')
+            ->orderByDesc('last_id')
```
notbakaneko: If we're ordering by a group column and applying a limit, wouldn't it make more sense to order by `create_date`?

Is there going to be a limit to the expected number of entries per `(user, pool)` combination (as in it not going to grow forever into 100k rows)?

app/Models/MatchmakingUserEloHistory.php
```diff
@@ -29,6 +30,20 @@ class MatchmakingUserEloHistory extends Model
 {
     protected $table = 'matchmaking_user_elo_history';
 
+    public static function daily(int $poolId, int $userId, int $days): Collection
+    {
+        $ids = static
+            ::where(['pool_id' => $poolId, 'user_id' => $userId])
```
notbakaneko, line 36: missing index on `matchmaking_user_elo_history`?

resources/css/bem/line-chart.less
```diff
@@ -75,6 +75,17 @@
       left: auto;
       right: 0;
     }
+
+    .@{_top}--profile-page-card & {
+      bottom: 100%;
+      top: auto;
+      margin-bottom: 10px;
+
+      &[data-float="right"] {
```
notbakaneko, line 84: actually, why have we been using `data-float` instead of modifiers? 🤔

resources/js/profile-page/detail.tsx
```diff
@@ -54,6 +61,24 @@ export default class Detail extends React.Component<Props> {
     );
   }
 
+  private handleDetailToggle(this: void) {
+    core.userPreferences.toggle('profile_detail_v2');
+  }
+
+  private renderDetailToggle() {
+    return (
+      <div className='profile-detail__type-toggle'>
```
notbakaneko, line 76: could do with a title

notbakaneko (commented):
just some comments first...

app/Models/MatchmakingUserStats.php
nanaya: I don't think relation work with composite keys...?

notbakaneko: could `hasMany()->where()`?

nanaya: not sure what to put for `where` that will work for preload and other actual relation stuff 🤔

resources/css/bem/line-chart.less
nanaya: probably because the value itself is involved in determining whether or not it should be changed

https://github.com/ppy/osu-web/blob/12e1bec34dea025eef3e989f15377830b882b441/resources/js/charts/line-chart.ts#L299-L305

resources/js/profile-page/detail.tsx
nanaya: Not quite sure what to say though. "Switch to new version of details section"? (no)

notbakaneko: `Toggle to new/old details/format`?  Do we even have a name for this? Maybe `toggle more/less details`? 🤔 

Also I feel like we should use a different icon from one that means refresh everywhere else...

nanaya: yeah I couldn't figure out a better icon.

And no, I've got no name for that apart of "v2" 🙃

notbakaneko: maybe `show more/less details` if we want to avoid new/old since the new one is showing more 🤷

resources/js/profile-page/detail-stats-v2/matchmaking.tsx
```diff
@@ -0,0 +1,222 @@
...
+              label={trans('users.show.matchmaking.losses')}
+              modifiers='plain'
+              value={formatNumber(stats.plays - stats.first_placements)}
+            />
+          </div>
+        </div>
+
+        <div className='profile-detail-stats-card__middle profile-detail-stats-card__middle--matchmaking-play'>
+          <div className='matchmaking-result'>
+            <div className='matchmaking-result__title'>Latest Match History</div>
```
notbakaneko: localisation

resources/js/profile-page/detail-stats-v2/matchmaking.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+  @action
+  private readonly handleVariantSelectorClick = (e: React.MouseEvent) => {
+    const poolId = getInt(htmlElementOrNull(e.currentTarget)?.dataset.poolId ?? '')
+      ?? fail('element missing data-pool-id');
+
+    this.props.controller.state.matchmakingPoolId = poolId;
+  };
+
+  private renderVariantSelector() {
```
notbakaneko, line 195: overlapping
Image

notbakaneko (changes requested):

resources/js/profile-page/detail-stats-v2/matchmaking.tsx
nanaya: 😦 

not even quite sure how to fix that

notbakaneko (dismissed):

nanaya:
adjusted stuff so the variant links aren't blocked by the stuff below 👀

notbakaneko:
Maybe we should put a mode switcher somewhere easily accessible :|

notbakaneko (approved):

Givikap120:
pp rankings display rank history, when ranked displays rating history
why this inconsistency?
I would've preferred graph for both rating and rank history (maybe toggleable?), but so it's consistent

PR #13187, opened 2026-07-28
Always use firstOrCreate instead of the other way around

nanaya:
It internally does the createOrFirst after doing the query first. The main problem with doing createOrFirst is it's logged as warning by at least proxysql.


notbakaneko (approved):
I keep forgetting that this is a thing ಠ_ಠ

PR #13188, opened 2026-07-29
Add matchmaking tier badge component

nanaya:
The alternative with css/html text can be checked here (also demo for the usage of both): https://github.com/ppy/osu-web/compare/master...nanaya:matchmaking-tier-badge-css?expand=1

That said, the font weight doesn't match and the positioning is different between firefox and chrome mobile ಠ_ಠ

(and the styling is different as well since it's using inset shadow for the text which doesn't exist in css)

Screenshot_20260729-170554
Screenshot_20260729-170531


nanaya:
Now with png instead because svg sucks (or more like the icons aren't designed natively as svg)

nanaya:
I've updated the html rank number version branch

notbakaneko (commented):
<details>
  <summary>100%</summary>

Chrome macOS
Image
Safari (there's a slight weird alignment on the html version)
Image 
Firefox macOS
Image
Edge Windows
Image
iOS
Image
</details>


<details>
  <summary>200% zoom</summary>

Chrome macOS
Image
Safari
Image
Firefox
Image
Edge Windows
Image
</details>


<details>
  <summary>125% zoom</summary>

Chrome macOS
Image
Safari
Image
Firefox (130%)
Image
Edge
Image
</details>

The text is fairly consistent in shape on macOS, other than the weird vertical alignment on Safari. On Windows however, the png and html numbers are visibly different at different zoom scales, but maybe not enough to matter?

notbakaneko (approved):

PR #13190, opened 2026-07-29
Show pp value of non-preserved scores

bdach:
With the recent roll-out of https://github.com/ppy/osu-queue-score-statistics/pull/392, scores have started getting marked as `preserve=0` much more aggressively, meaning that they started up showing on user profiles in the "recent" section in that state. When they do, the pp value is not displayed which has been noticed pretty much instantly. See [this forum thread](https://osu.ppy.sh/community/forums/topics/2229803?n=1) for an example of a user confused about this.

Now, this is how things worked on the old tables - if a score wasn't the user's best, its pp value would not show anywhere, including the recent scores list. See 3778dec3c3ae26037c331be0c3ca4e7f79bf719e, which is preserved to this day as

https://github.com/ppy/osu-web/blob/ad92032fd45eaf9bef748fd37ebde2eeb0a17370/resources/js/scores/pp-value.tsx#L15-L17

In ade9f0aad4853f0bfc3c46fc0d60b7dfabbfc2fb, seeking parity, I made the treatment of `preserve=0` match the above. However I am told that it is more desirable to keep this value visible as it prevents users wanting to see these values from deploying browser extensions of dubious quality that tend to incur disproportionate load on the API.

This change will affect all places where this component is used, namely:

- beatmap leaderboards (which I would hope to be a dead case anyway, as non-preserved scores should never be on a beatmap leaderboard),
- profile page, both "recent" and "most watched replays" sections,
- individual score pages.

While the scope of the above is somewhat larger than the single place this change was desired (the "recent" profile section), I do not see any particular value in constraining this change to that particular usage site.

| before | after |
| :-: | :-: |
| Screenshot 2026-07-29 at 14 11 47 | Screenshot 2026-07-29 at 14 12 33 |
| Screenshot 2026-07-29 at 14 12 15 | Screenshot 2026-07-29 at 14 12 46 |

### DIFF
--- a/resources/css/bem-index.less
+++ b/resources/css/bem-index.less
@@ -307,6 +307,7 @@
 @import "bem/popup-menu";
 @import "bem/popup-menu-float";
 @import "bem/post-box";
+@import "bem/pp-value";
 @import "bem/product-box";
 @import "bem/profile-badges";
 @import "bem/profile-cover-change-popup";
--- a/resources/css/bem/pp-value.less
+++ b/resources/css/bem/pp-value.less
@@ -1,3 +1,6 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
 .pp-value {
   &--non-preserved {
     opacity: 0.7;
--- a/resources/css/bem/pp-value.less
+++ b/resources/css/bem/pp-value.less
@@ -1,0 +1,5 @@
+.pp-value {
+  &--non-preserved {
+    opacity: 0.7;
+  }
+}
--- a/resources/js/scores/pp-value.tsx
+++ b/resources/js/scores/pp-value.tsx
@@ -5,6 +5,7 @@ import ScoreJson from 'interfaces/score-json';
 import * as React from 'react';
 import { formatNumber } from 'utils/html';
 import { trans } from 'utils/lang';
+import { classWithModifiers } from "../utils/css";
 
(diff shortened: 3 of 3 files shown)


nanaya:
Feels like the tooltip wordings should be adjusted...? Something about the shown pp/score not being counted for the user's total pp. Saying it doesn't award pp while showing a pp value seems a bit weird.

bdach:
I am open to specific suggestions on copy. I almost used https://github.com/ppy/osu-web/blob/4c641d4e35d93d7510f4d623e048dac19b4e0444/resources/lang/en/scores.php#L8 verbatim except it'd a little weird on score pages because the same string would be used in two places (once in the tooltip and once in the yellow warning banner below the score).

nanaya (approved):
yeah any attempt to explain that seems to end up quite a bit longer. also we're already showing non-best pp value with no indicator anyway (for scores with different mods combination) 🤷

cesarila:
I like this a lot for recent scores, but seems like it could be a headache for cases like my top pin which would probably be my top pp on the map if I set it after the combo scaling removal rework. 

My profile: https://osu.ppy.sh/users/4101254
My top pin: https://osu.ppy.sh/scores/2734184070

I could see some players, maybe even myself in the future, being confused from cases like this. Maybe making the tooltip say something like: "This score was pinned before it would have been your top pp score on the map, so it doesn't award pp" would be clearer for such cases. 

Edit: Brain fail because a pin is a preserved score my bad

bdach:
@cesarila your example is not affected by this PR. pinned scores are preserved.

PR #13192, opened 2026-07-30
Only show matchmaking stats with nonzero ranked plays

### LINKED ISSUE #13191: Ranked play stats in profile page show even if only played private/unranked duel
This can be seen on my profile : https://osu.ppy.sh/users/33800057
On mania and catch, I only played unranked duel with a friend, but the ranked stats are still show on my profile.

nanaya:
Resolves #13191


notbakaneko (approved):
I thought this was some other PR 🙈

PR #13193, opened 2026-07-30
Lazy load user avatar in chat channel list

### LINKED ISSUE #13181: Loading chat takes really long due to extensive avatar fetching
when i open chats, it takes a prolonged period to load + send messages. i looked at the network tab and i saw that it was fetching every single past chat's avatar, and it halted everything until it finished fetched all of the avatars

Image

using Firefox 153.0

notbakaneko:
Replaces the use of `background-image` with actual lazy loaded `img` (although, Safari appears to have a very liberal range of what is lazy loaded...)

Includes a workaround to avoid showing the spinner and fade-in for images that are already loaded in cache, most noticeable during browser back/forward navigations.

The component name could use an improvement but I don't want to just blindly replace all the existing `UserAvatar` usages either and find out the layout on something is different 👀  

maybe fixes #13181

### DIFF
--- a/resources/css/bem-index.less
+++ b/resources/css/bem-index.less
@@ -425,6 +425,7 @@
 @import "bem/turbo-progress-bar";
 @import "bem/update-streams-v2";
 @import "bem/user-action-button";
+@import "bem/user-avatar";
 @import "bem/user-card";
 @import "bem/user-card-brick";
 @import "bem/user-cards";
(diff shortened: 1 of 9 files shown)


nanaya:
that's... a lot of code to render an image (and potentially twice because of the first image dom creation and check)

nanaya:
Also apparently the spinner takes up cpu until the image is actually loaded even though it's not visible on viewport... (at least on chrome, with no avatar preloaded)

Maybe just do a simple img for now

notbakaneko:
hmm :|

Issue #2259, opened 2018-01-21
[Wiki search] The list should show articles basing on locale selected


TPGPL:

nanaya:
Is it currently not?

TPGPL:
If you choose Polish, it still searches English ones

TPGPL:
If you are searching for OWC (French locale) the French articles are combined with English ones into one list. It should probably show the English ones on a different list or at least give a notice that it will show the original articles now

gagahpangeran:
I report this issue here because looks like this is related.

Currently the search suggestion is based on the website locale only. When choosing different locale in wiki locale dropdown, it does not show any suggestion based on the choosen locale.

---

Website locale : English
Wiki locale : Bahasa Indonesia

Expected result : the search suggestion shows Bahasa Indonesia article.
Current result : the search suggestion only shows English article.

<img width="943" height="916" alt="Image" src="https://github.com/user-attachments/assets/99e7e9ba-6461-451b-a3a3-65748f4c3ffd" />

---

Another comparison.

Website locale : Bahasa Indonesia
Wiki locale : Bahasa Indonesia

Current result : the search suggestion shows both English and Bahasa Indonesia article.

<img width="939" height="918" alt="Image" src="https://github.com/user-attachments/assets/37d92e3c-2afd-46c5-95a5-db049fda5a2d" />

PR #13208, opened 2026-08-04
Use wiki locale for wiki search suggestions

gagahpangeran:
As I mentioned in https://github.com/ppy/osu-web/issues/2259#issuecomment-5011420172. Currently the search suggestion is using the website locale only. When choosing different locale in wiki locale dropdown, it does not show any suggestion based on the choosen wiki locale. This PR changes the search suggestion behavior to use the wiki locale.

Also I'm not sure how to write test for this controller, since there is not many example to write test with elastic search involved.


resources/js/wiki-search-controller.ts
```diff
@@ -82,12 +84,13 @@ export class WikiSearchController {
   }
 
   @action
-  updateQuery(query: string) {
+  updateQuery(query: string, locale?: string) {
```
notbakaneko: `locale` is never `undefined`

resources/js/wiki-search-controller.ts
```diff
@@ -17,6 +18,7 @@ export class WikiSearchController {
   @observable suggestions: SuggestionJson[] = [];
 
   private readonly debouncedFetchSuggestions = debounce(() => this.fetchSuggestions(), 200);
+  @observable private locale: string | undefined;
```
notbakaneko: passing `locale` into `WikiSearch` means there should always be `locale` set

notbakaneko (changes requested):

resources/js/wiki-search-controller.ts
nanaya: would the value even ever change

notbakaneko (approved):
I'll handle the duplicate language results separately

PR #13210, opened 2026-08-05
Update tests to use the attribute syntax for data provider

nanaya:
The doc comment syntax is deprecated in phpunit 11.

~~There is also `@group` thing but it'll be done separately as it requires some manual tests.~~ Can't use both syntaxes so might as well convert them all.

### DIFF
--- a/tests/Controllers/BeatmapsControllerSoloScoresTest.php
+++ b/tests/Controllers/BeatmapsControllerSoloScoresTest.php
@@ -12,7 +12,6 @@ use App\Models\Beatmap;
 use App\Models\Beatmapset;
 use App\Models\Country;
 use App\Models\Genre;
-use App\Models\Group;
 use App\Models\Language;
 use App\Models\OAuth;
 use App\Models\Solo\Score as SoloScore;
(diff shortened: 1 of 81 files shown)


tests/Controllers/BeatmapsControllerSoloScoresTest.php
```diff
@@ -200,9 +201,9 @@ public static function tearDownAfterClass(): void
     }
 
     /**
-     * @dataProvider dataProviderForTestQuery
      * @group RequiresScoreIndexer
```
notbakaneko: annotations get ignored if there are attributes at the same level, so it needs to be replaced anyway (and the `@group`s on this can probably be moved to the class)

notbakaneko (changes requested):

tests/Controllers/BeatmapsControllerSoloScoresTest.php
nanaya: I just noticed the name may conflict with model Group... 😐 

(also the linter confused annotation with import)

notbakaneko (approved):

PR #13211, opened 2026-08-05
Update phpunit

nanaya:
There are still deprecation but should be easy to clean up afterwards.

- [x] #13209
- [x] #13210

As for the flush handler thing, I'm not quite sure if there's good way to do it.

### DIFF
--- a/tests/CreatesApplication.php
+++ b/tests/CreatesApplication.php
@@ -19,14 +19,4 @@ trait CreatesApplication
 
         return $app;
     }
-
-    /**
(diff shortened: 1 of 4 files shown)


tests/CreatesApplication.php
notbakaneko: incidentally, does this do anything? 🤔

nanaya: oh right, it should be removed. it's used internally but also already defined by the parent class

notbakaneko (approved):

PR #13221, opened 2026-08-13
Use updated php white space lint

nanaya:
-  Squiz.WhiteSpace.LanguageConstructSpacing
   This sniff has been deprecated since v3.3.0 and will be removed in v4.0.0. Use the Generic.WhiteSpace.LanguageConstructSpacing sniff instead.

it says.

They're not the same test and I can't really tell what either of them exactly are ┐( ･_･)┌


notbakaneko (approved):
looks to be the same with more keywords

PR #13224, opened 2026-08-13
Update to Laravel 12

nanaya:
13 involves updating passport which isn't quite trivial. This one looks straightforward enough and removes the need for extra composer option to ignore advisory check when updating anything.


resources/views/home/landing.blade.php
```diff
@@ -163,7 +163,7 @@ class="landing-hero__bg"
 
     <script type="application/ld+json">
     {
-      "@context": "https://schema.org",
+      "@@context": "https://schema.org",
```
gagahpangeran, line 166: Why is this double `@`?

nanaya: `@context` is now a blade directive

(previous screenshot was too large whoops)

notbakaneko (approved):
seems to be working so far...

notbakaneko:
let's see if anything breaks 🙈

PR #13229, opened 2026-08-16
Add wiki suggestion API endpoint

gagahpangeran:
Pre-req for wiki suggestion search box in lazer.

The endpoint is public to match `/wiki/` endpoint and also match the wiki suggestion search box on web (can be used without login).

Use asterisk (`*`) as highlight tags instead of default `<em>` html tag for easier parsing in client.

reference:

- https://www.elastic.co/guide/en/elasticsearch/reference/7.17/highlighting.html#configure-tags
- https://www.elastic.co/guide/en/elasticsearch/reference/7.17/highlighting.html#highlighting-settings

### DIFF
--- a/app/Http/Controllers/SuggestionsController.php
+++ b/app/Http/Controllers/SuggestionsController.php
@@ -42,6 +42,11 @@ class SuggestionsController extends Controller
     public function wiki()
     {
         $search = new WikiSuggestions(new WikiSuggestionsRequestParams(\Request::all()));
+
+        if (is_api_request()) {
+            $search->useMarkdownHighlightTags();
+        }
+
         return json_collection([...$search->response()], new WikiSuggestionsHitTransformer());
     }
 }
--- a/app/Libraries/Elasticsearch/Highlight.php
+++ b/app/Libraries/Elasticsearch/Highlight.php
@@ -12,8 +12,9 @@ namespace App\Libraries\Elasticsearch;
 class Highlight implements Queryable
 {
     protected $fields = [];
-    protected $numberOfFragments = 5;
     protected $fragmentSize = 100;
+    protected $markdownTags = false;
+    protected $numberOfFragments = 5;
 
     /**
      * @return $this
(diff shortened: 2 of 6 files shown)


app/Http/Controllers/SuggestionsController.php
```diff
@@ -42,6 +42,11 @@ public function user()
     public function wiki()
     {
         $search = new WikiSuggestions(new WikiSuggestionsRequestParams(\Request::all()));
+
+        if (is_api_request()) {
+            $search->setHighlightTags(['*'], ['*']);
```
notbakaneko: I don't see us using anything else besides markdown, so this might be better just as a single flag

notbakaneko: This also assumes we're never going to use `*` in titles since we can't effectively escape it here.

gagahpangeran: > This also assumes we're never going to use `*` in titles since we can't effectively escape it here.

Now that you mentioned about this, is there any suggestion how to handle this?

Checking [sitemap page](https://osu.ppy.sh/wiki/en/Sitemap), currently there is not any title with `*` in it.

app/Libraries/Elasticsearch/Highlight.php
```diff
@@ -54,16 +55,35 @@ public function numberOfFragments(int $count)
...
+        return $this;
+    }
+
     /**
      * {@inheritdoc}
      */
     public function toArray(): array
     {
-        return [
+        $highlight = [
```
nanaya, line 75: should the encoder option be adjusted for markdown?

notbakaneko: encoder should be `default` for markdown

app/Http/Controllers/SuggestionsController.php
notbakaneko: I think we'll just deal with it when/if it happens, worst case we'll probably just markdown escape the response and string replace the `<em>` tags

Walavouchey: i don't see why it would ever be needed

notbakaneko (approved):
I wonder if it should just always be set to return markdown (that's a problem for another time)

PR #13232, opened 2026-08-17
Adjust user report requirement

nanaya:
Report comment is optional for reporting comment.

As discussed in [team channel](https://discord.com/channels/90072389919997952/1458132886946451497/1538774486839459892).

### DIFF
--- a/app/Models/UserReport.php
+++ b/app/Models/UserReport.php
@@ -99,6 +99,17 @@ class UserReport extends Model
         return $this->belongsTo(User::class, 'user_id');
     }
 
+    public function isCommentOptional(): bool
+    {
+        static $optionalTypes = [
+            MorphMap::MAP[Chat\Message::class],
+            MorphMap::MAP[Comment::class],
+        ];
+
+        return $this->reason !== 'Other'
+            && in_array($this->reportable_type, $optionalTypes, true);
+    }
+
     public function isRecent(): bool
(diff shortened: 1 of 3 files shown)


app/Models/UserReport.php
```diff
@@ -108,7 +119,7 @@ public function isValid()
     {
         $this->validationErrors()->reset();
 
-        if (!present(trim($this->comments)) && (!($this->reportable instanceof Chat\Message) || $this->reason === 'Other')) {
+        if (!present(trim($this->comments)) && !$this->isCommentOptional()) {
```
nanaya, line 122: `$this->comments` may be null (including existing ones generated outside of osu-web) but the entrypoint for generating the model always fills it with empty string ಠ_ಠ

tests/Models/UserReportTest.php
```diff
@@ -159,7 +160,7 @@ public function testNoComments(string $class): void
         $reportable = static::makeReportable($class);
         $reporter = User::factory()->create();
 
-        if ($class === Message::class) {
+        if ($reportable instanceof Message || $reportable instanceof Comment) {
```
nanaya, line 163: I suppose this should use `$optionalTypes` in the relevant function but I can't figure out a sensible const name for it.

notbakaneko (approved):

PR #13233, opened 2026-08-17
Add custom oauth token decoding flow

nanaya:
Decoding the token is a lot simpler than generating it so just do it manually and get rid of the many layers of abstraction. This also removes the duplicated query for the token.


notbakaneko (commented):
I think I'd add a basic test that verifies `EncodeToken::decodeAccessToken(EncodeToken::encodeAccessToken($token)) === $id`

notbakaneko (approved):

PR #13234, opened 2026-08-18
Cleanup search related classes

notbakaneko:
Add type hints, remove `HasSearch` class since the other class that extended it was deleted a long time ago.
I thought of removing the fluent interfaces, but having them seems useful in some places for chaining and inlining?

I'm pretty sure I missed more stuff but updating turned into a rabbit hole of more changes and updates to fix breakages


app/Libraries/Search/WikiSuggestions.php
```diff
@@ -19,24 +19,20 @@ public function __construct(WikiSuggestionsParams $params)
             $params
         );
 
-        $this->collapse = ['field' => 'path.keyword'];
-        $this->source(['locale', 'title', 'path']);
-        $this->highlight(
-            (new Highlight())
+        $this
+            ->collapse('path.keyword')
```
nanaya, line 23: if anything this adds one extra line? 🙃 

instead of
```
$this->collapse(...);
$this->source(...);
...
```

now we have
```
$this
    ->collapse(...)
    ->source(...)
    ...
```

app/Libraries/Elasticsearch/Search.php
```diff
@@ -308,31 +389,17 @@ private function handleError(ElasticsearchException $e, string $operation)
...
      * Wrapper function to run a query with timing and error reporting.
      *
-     * @param string $operation
-     * @param callable $callable
-     *
-     * @return mixed Returns whatever $callable returns, void with $this->error set on error.
+     * @return mixed Returns whatever $callable returns, null with $this->error set on error.
      */
-    private function runQuery(string $operation, callable $callable)
+    private function runQuery(string $operation, callable $callable): mixed
```
nanaya, line 402: this feels like should just return `[$result, $error]` instead but that's for another day I guess 🤨

nanaya (approved):

PR #13235, opened 2026-08-18
Update timestamp when changing password

notbakaneko:
as requested

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -5,6 +5,7 @@
 
 namespace App\Models;
 
+use App\Casts\TimestampOrZero;
 use App\Exceptions\ChangeUsernameException;
 use App\Exceptions\InvariantException;
 use App\Exceptions\ModelNotSavedException;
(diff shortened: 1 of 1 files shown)


app/Models/User.php
```diff
@@ -2299,6 +2299,7 @@ public function isValid()
 
             if ($this->validationErrors()->isEmpty()) {
                 $this->user_password = Hash::make($this->password);
+                $this->user_passchg = time();
```
nanaya: technically the field should be marked as ~~`datetime`~~ `TimestampOrZero` in the `$casts` array (and the `getAttribute` thing) and this set to `now()` instead... 🤔

app/Models/User.php
```diff
@@ -2299,6 +2301,7 @@ public function isValid()
 
             if ($this->validationErrors()->isEmpty()) {
                 $this->user_password = Hash::make($this->password);
+                $this->user_passchg = Carbon::now();
```
nanaya, line 2304: I wonder if attempting to change to the same password should result in error?

nanaya (approved):

app/Models/User.php
nanaya: I guess can be done later if needed

PR #13237, opened 2026-08-18
Update to Passport 13

nanaya:
That was painful 😐 

- typing adjustments
- overrides for supporting non-hashed client secret
  - maybe update to actually hash them later or something
  - alternatively pre-hash to a different column and swap when updating to this
  - maybe drop the custom hasher first which should only apply to user password
- overrides to support refreshing token with `*` scope
- overrides to support passport-12 grant types by default
  - will need to migrate later
  - maybe allow user to enable specific grant types
- client uuid is disabled
  - I don't know if there's sane migration path for this
  - maybe a new column with copy of id for existing clients and actual uuid for new ones?

depends on...
- [x] #13224


nanaya:
~~Wait, the scope isn't normalised correctly~~ Nvm refreshing token never has its scope normalised.

~~There's different problem with deserialising previous version of the oauth related objects in the session.~~ "Fixed" by... not deserialising them.

notbakaneko:
I _think_ there may be a one time blip if a user visits the auth page on passport 12 and presses a button on 13 but I don't know how likely that will be 🤔

app/Http/Controllers/Passport/AuthorizationController.php
```diff
@@ -23,30 +25,24 @@ class AuthorizationController extends PassportAuthorizationController
...
             }
-            return parent::authorize($this->normalizeRequestScopes($psrRequest), $request, $clients, $tokens);
+            return parent::authorize($this->normalizeRequestScopes($psrRequest), $request, $psrResponse, $viewResponse);
         } catch (AuthenticationException $_e) {
             $cancelUrl = $request->fullUrl();
             $cancelUrl .= strpos($cancelUrl, '?') === false ? '?' : '&';
             $cancelUrl .= 'prompt=none';
 
-            return ext_view('sessions.create', [
+            return (new SimpleViewResponse('sessions.create'))->withParameters([
```
notbakaneko: extra brackets around constructor no longer needed in php 8.4

nanaya:
they can probably retry the process

app/Models/OAuth/Token.php
```diff
@@ -40,17 +43,17 @@ public static function findForVerification(string $id): ?static
...
     }
 
-    public function user()
+    public function user(): BelongsTo
     {
         return $this->belongsTo(User::class, 'user_id');
     }
 
-    public function can($scope)
+    public function can($scope): bool
```
notbakaneko:
```suggestion
    public function can(string $scope): bool
```

app/Models/OAuth/Client.php
```diff
@@ -177,6 +180,31 @@ public function validationErrorsTranslationPrefix(): string
         return 'oauth.client';
     }
 
+    #[\Override]
+    protected function grantTypes(): Attribute
```
notbakaneko, line 185: also add `@property string[] $grant_types` to the class phpdoc

notbakaneko (changes requested):
wonder if we should `#[\Override]` attribute the relations as well 🤔

notbakaneko (approved):

PR #13241, opened 2026-08-20
Add ``description`` to daily challenge queue item

venix12:
So that description field on ``multiplayer_rooms`` can be utilized for daily challenge.


app/Console/Commands/DailyChallengeCreateNext.php
```diff
@@ -45,6 +45,7 @@ public function handle()
             $room = (new Room())->startGame(
                 User::findOrFail($hostId),
                 [
+                    'description' => $nextQueueItem->description,
```
notbakaneko: I'm not sure if we allow setting the description when normally starting a room, so this should probably go in `$extraParams`

peppy (approved):

peppy (approved):

Issue #10313, opened 2023-06-27
add post counter nicknames in forums again


tatitatai:
it was a topic in osu! community forums. if you want to read about their thoughts and opinions: https://osu.ppy.sh/community/forums/topics/1783402?n=1

(these are all people who commonly use the forums)

cl8n:
was this intentionally removed, or just not considered in the new forum design?

peppy:
Bit of both. It was removed as something which didn't really need to be there (and didn't make much sense).

stilettocode:
Would I be allowed to work on this?

peppy:
no one will stop you!

mrafie1:
I'd like to take a crack at this since there hasn't been any kind of activity in a while. If this is still something people would be interested in, I'd be happy to look into putting together a pr 👍

hasanhuseyinkahraman:
This has already been implemented as far as i see. Shouldn't this issue be closed?

PR #13246, opened 2026-08-22
Display forum titles if user has no titles

YassineHaouzane:
Attempt at implementing #10313.

* Adds forum titles by computing it from the user's post count.
* I did not add any translation for these titles since user titles do not have any.
* These titles and rules were found in this wiki page https://osu.ppy.sh/wiki/en/Community/User_title.
* We will display the forum title if the user has no rank title.

image

### DIFF
--- a/app/Libraries/User/UserForumTitles.php
+++ b/app/Libraries/User/UserForumTitles.php
@@ -1,0 +1,30 @@
+<?php
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the GNU Affero General Public License v3.0.
+// See the LICENCE file in the repository root for full licence text.
+
+declare(strict_types=1);
+
+namespace App\Libraries\User;
+
+use App\Models\User;
+
+class UserForumTitles
+{
+    public static function get(User $user): string
+    {
+        $count = $user->user_posts;
+        return match (true) {
+            $count < 5 => 'Rhythm Rookie',
+            $count < 15 => 'Tempo Trainee',
+            $count < 30 => 'Whistle Blower',
+            $count < 50 => 'Cymbal Sounder',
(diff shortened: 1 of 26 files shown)


app/Models/User.php
```diff
@@ -1839,6 +1839,38 @@ public function titleUrl(): ?string
         return $this->rank?->url;
     }
 
+    public function forumTitle(): string
+    {
+        if ($this->user_posts < 5) {
```
nanaya: the class is already too big. this should be moved to its own helper in `libraries/user`

nanaya (changes requested):

app/Libraries/User/UserForumTitles.php
```diff
@@ -0,0 +1,46 @@
...
+
+namespace App\Libraries\User;
+
+use App\Models\User;
+
+class UserForumTitles
+{
+    public static function get(User $user): string
+    {
+        $user_posts_count = $user->user_posts;
```
nanaya: variable name should be camelCase

...or maybe just call it `count` instead.

app/Libraries/User/UserForumTitles.php
```diff
@@ -0,0 +1,46 @@
...
+        if ($user_posts_count < 5) {
```
nanaya: a bit more compact with `match` or array maybe
```php
return match (true) {
    $count < 5 => '...',
    $count < 15 => '...',
    ...
};

// or
static $titles = [
    5 => '...',
    15 => '...',
    ...
    PHP_INT_MAX => '...',
];
foreach ($titles as $maxCount => $title) {
    if ($count < $maxCount) {
        return $title;
    }
}
```

nanaya (changes requested):

nanaya (approved):

PR #13251, opened 2026-08-25
Fix beatmapset exact search

### LINKED ISSUE #12420: Pagination on beatmap listing doesn't always work
Searching for _artist=""THE ORAL CIGARETTES""_ - "Mode: Any, Has leaderboard" beatmapsets only get 50 which aren't all for this artist. 
For example I couldn't find my "Youshi Tanrei na Uso" (https://osu.ppy.sh/beatmapsets/1029803#osu/2221938) although it meets the search requirements.

Switching to **Mode:osu!** gives me my map and many others which were hidden in the previous search.

Sometimes the search works as intended but after clearing the cache the issue persist.

- First URL https://osu.ppy.sh/beatmapsets/search?e=&c=&g=&l=&m=&nsfw=&played=&q=artist%3D%22%22THE%20ORAL%20CIGARETTES%22%22&r=&sort=&s=&cursor_string=eyJfc2NvcmUiOjE0Ljc4MzE3MSwiaWQiOjEyNzQ5NzV9

- Second URL 
https://osu.ppy.sh/beatmapsets/search?e=&c=&g=&l=&m=0&nsfw=&played=&q=artist%3D%22%22THE%20ORAL%20CIGARETTES%22%22&r=&sort=&s=

nanaya:
Looks like es gets very confused when the search relevance score are all the same 🤔 

This doesn't really fix the problem but changes the behaviour when only exact search is involved.

Resolves #12420.

~~(this probably will fail tests)~~

### DIFF
--- a/app/Libraries/Search/BeatmapsetSearch.php
+++ b/app/Libraries/Search/BeatmapsetSearch.php
@@ -409,10 +409,13 @@ class BeatmapsetSearch extends RecordSearch
             $searchFields[] = $field;
             $searchFields[] = "{$field}.*";
 
-            $subQuery->should(['term' => ["{$field}.raw" => ['value' => $value, 'boost' => 100]]]);
+            $valueWithoutQuotes = preg_replace('/^"\s*(.*)\s*"$/', '\1', $value);
+            $subQuery->should(['term' => ["{$field}.raw" => ['value' => $valueWithoutQuotes, 'boost' => 100]]]);
         }
 
-        $subQuery->should(QueryHelper::queryString($value, $searchFields, 'and'));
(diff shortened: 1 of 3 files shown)


nanaya:
It looks like the core problem is the score changes ~~a lot more frequently when doing exact/quoted search~~ between search.

nanaya:
I suppose the correct fix would be [point in time marker](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-open-point-in-time) but that'd be quite a lot more involved changes. And still need to determine the ideal TTL and check how much impact it may cause.

nanaya:
This needs a bit more thinking.

Issue #13185, opened 2026-07-27
Add an option for a Note in modding reviews


fllecc:
Currently there's no option to add a note into a modding review, and it has to be posted separately. Adding the option to include a note would make the modding process more efficient, at least for me. Though, I think the option to post notes on other users' maps is for BNs only, so this would also have to be a BN-only feature.

Basically, adding this option:
<img width="1662" height="336" alt="Image" src="https://github.com/user-attachments/assets/6715466a-669d-45d5-a5be-4f755126c7e9" />
to here:
<img width="1669" height="294" alt="Image" src="https://github.com/user-attachments/assets/9b9469a3-3c84-42a0-85e7-c50402cf59cf" />

PR #13255, opened 2026-08-26
More explicit beatmapset discussion state typings

notbakaneko:
More explicit type discrimination of `Props` between editable and readonly discussions

Part of adding notes to discussion reviews for #13185 that I'm splitting off for sanity checking the types make sense

### DIFF
--- a/resources/js/beatmap-discussions/discussion.tsx
+++ b/resources/js/beatmap-discussions/discussion.tsx
@@ -3,7 +3,7 @@
 
 import BeatmapsetDiscussionJson, { BeatmapsetDiscussionJsonForBundle, BeatmapsetDiscussionJsonForShow } from 'interfaces/beatmapset-discussion-json';
 import BeatmapsetDiscussionPostJson from 'interfaces/beatmapset-discussion-post-json';
-import BeatmapsetDiscussionsStore from 'interfaces/beatmapset-discussions-store';
+import { HasDiscussionsEditable, HasDiscussionsReadOnly } from 'interfaces/has-discussions';
 import { findLast } from 'lodash';
 import { action, computed, makeObservable } from 'mobx';
 import { observer } from 'mobx-react';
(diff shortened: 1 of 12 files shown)


resources/js/beatmap-discussions/discussion.tsx
```diff
@@ -233,20 +229,33 @@ export class Discussion extends React.Component<Props> {
...
-        read={this.isRead(post)}
-        readonly={this.readonly}
-        resolvedStateChangedPostId={this.resolvedStateChangedPostId}
-        store={this.props.store}
-        type={type}
-        user={user}
-      />
-    );
+    // can't discriminate type without this even though the props are the same.
+    return this.props.discussionsState == null
```
notbakaneko: this is silly and I'd like a better way to handling it :|

nanaya: maybe turn the `{ discussionsState, store }` into an object and pass that around instead?

notbakaneko: putting it into an object just seems to move the error into the object but still doesn't accept the prop 🤔

nanaya: I mean turning the two props into single object/prop

resources/js/beatmap-discussions/discussion.tsx
```diff
@@ -233,20 +229,19 @@ export class Discussion extends React.Component<Props> {
...
-        post={post}
-        read={this.isRead(post)}
-        readonly={this.readonly}
-        resolvedStateChangedPostId={this.resolvedStateChangedPostId}
-        store={this.props.store}
-        type={type}
-        user={user}
-      />
-    );
+    const sharedProps = {
```
nanaya, line 232: the thing with doing it this way is typescript won't complain about any [extraneous fields](https://github.com/microsoft/TypeScript/issues/39998)...

You can also do this the other way for "proper" react field name sorting...

```
const differentProps = this.props.discussionsState == null
  ? {
    store: this.props.store,
  } : {
    discussionsState: this.props.discussionsState,
    store: this.props.store,
  };
return <Post key={...} ... {...differentProps} />
```

(Or as I mentioned before, use `this.props.thing` that contains `.discussionsState` and `.store` so they're just single object of type `ThingReadOnly | Thing`)

nanaya (approved):

PR #13257, opened 2026-08-27
Move the custom hasher class to be user-specific

nanaya:
Using this weird hasher for oauth client secrets doesn't seem right? Or does it? 🤔


notbakaneko:
does setting the default to `bcrypt` and using `Hash::driver('osubcrypt')` where it's needed not work?

nanaya:
probably but that seems pointless? None of our auth flow involves laravel's session thing as far as I can tell and the relevant function themselves (`make` and `check`) are very simple compared to all the abstractions that'd be involved when using laravel's Hash thing.

notbakaneko (approved):
I keep forgetting about this 🙈

PR #13260, opened 2026-08-28
Remove check to block self messaging

Loluh:
I want to suggest removing this check as it prevents self hosted song requests bots from existing in Lazer.

Even though we could use already existing ones that have their own accounts, I believe having this check discourages new developers from publishing new bots as making one means you yourself wouldnt be able to use it until you are allowed to create an account for the bot itself.

Bots as these are very practical for streamers to not have to constantly exit the osu client to grab requests from people in their chat. For example if you are using the recommended fullscreen setting in the client, it is extremely annoying to tab out constantly.

I dont know if there was any deep reasoning behind this check, if so, feel free to reject this PR.


peppy:
If this doesn't cause issues, I think it would be a welcomed change.

We would have to make sure it doesn't cause any circular messaging flows to spiral out of control.

app/Models/Chat/Channel.php
```diff
@@ -408,8 +408,24 @@ public function pmTargetFor(?User $user): ?User
...
+
+            if (
+                count($userIds) === 2
+                && $userIds[0] === $userId
+                && $userIds[1] === $userId
+            ) {
+                return $user;
+            }
+
+            return null;
```
nanaya: `$this->userIds()` is cheap for pm so just do this maybe

```suggestion
        return $this->memoize(__FUNCTION__.':'.$userId, function () use ($user, $userId) {
            foreach ($this->userIds() as $targetId) {
                if ($targetId !== $userId) {
                    return $this->users()->find($targetId);
                }
            }
            
            // all ids point to self
            return $user;
```

resources/js/models/chat/channel.ts
```diff
@@ -126,7 +126,15 @@ export default class Channel {
...
-    return this.userIds.find((userId: number) => userId !== core.currentUserOrFail.id);
+    const currentUserId = core.currentUserOrFail.id;
+
+    // self-PMs have the current user as both participants, so there is no
+    // "other" user to find; the conversation targets the current user themselves
+    if (this.userIds.length > 0 && this.userIds.every((userId: number) => userId === currentUserId)) {
+      return currentUserId;
+    }
+
+    return this.userIds.find((userId: number) => userId !== currentUserId);
```
nanaya: can use similar logic as suggested above instead of potentially looping twice

resources/js/stores/channel-store.ts
```diff
@@ -90,16 +90,16 @@ export default class ChannelStore implements DispatchListener {
...
 
     for (const [, channel] of this.channels) {
       if (channel.type !== 'PM') {
         continue;
       }
 
-      if (channel.userIds.includes(userId)) {
+      // self-PMs have the current user as both participants, so match on pmTarget;
+      // regular PMs match the other participant via userIds
+      if (isSelf ? channel.pmTarget === userId : channel.userIds.includes(userId)) {
```
nanaya: ...or always check `pmTarget`?
```suggestion
      if (channel.pmTarget === userId) {
```

nanaya (changes requested):

resources/js/stores/channel-store.ts
nanaya: considering `pmTarget` already returns undefined if not a pm, the channel type check above can probably be removed as well

resources/js/models/chat/channel.ts
Loluh: fixed

resources/js/stores/channel-store.ts
Loluh: done, also removed the isSelf declaration

resources/js/models/chat/channel.ts
```diff
@@ -122,11 +122,18 @@ export default class Channel {
...
+    if (this.type !== 'PM' || $this->userIds.length == 0) {
       return;
     }
 
-    return this.userIds.find((userId: number) => userId !== core.currentUserOrFail.id);
+    foreach ($this->userIds() as $targetId) {
+      if ($targetId !== core.currentUserOrfail.id) {
+        return $targetId;
+      }
+    }
```
nanaya: what

Loluh: my bad, brain didnt switch

PR #13261, opened 2026-08-30
Add mutual filter in home friends page

### LINKED ISSUE #13058: (proposal): add mutual filter to friends page
There's currently no straightforward way to see how many mutuals you have. I'd like to see a new 'Mutual' filter that displays mutual friends and their count on the friends page (or something that indicates mutual count). Below is a quick mock-up of what it might look like in brick view.

Image

YassineHaouzane:
Closes #13058
* Adds a client side filter showing only mutual friends.
* Added fr and en translations for the filter label.

https://github.com/user-attachments/assets/72f01b1c-02f1-4539-a530-c312edc29df5

### DIFF
--- a/app/Models/UserProfileCustomization.php
+++ b/app/Models/UserProfileCustomization.php
@@ -33,7 +33,8 @@ class UserProfileCustomization extends Model
         'legacy_score_only' => false,
         'profile_cover_expanded' => true,
         'scoring_mode' => self::SCORING_MODES[0],
-        'user_list_filter' => self::USER_LIST['filters']['default'],
+        'user_list_filter' => self::USER_LIST['filters']['status']['default'],
+        'user_list_relationship_filter' => self::USER_LIST['filters']['relationship']['default'],
         'user_list_sort' => self::USER_LIST['sorts']['default'],
         'user_list_view' => self::USER_LIST['views']['default'],
     ];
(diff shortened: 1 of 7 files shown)


diquoks:
> - Added fr and en translations for the filter label.

translation strings for languages other than english should be added via crowdin

YassineHaouzane:
> > * Added fr and en translations for the filter label.
> 
> translation strings for languages other than english should be added via crowdin

To make sure I understand, I should revert the change that adds the french translation and add it via crowdin when **these changes get merged** ? 🤔 

Based on this documentation/diagram we can only add new translation keys through crowdin after the english key gets merged. 
[github.com/ppy/osu/wiki/Localising-the-client](https://github.com/ppy/osu/wiki/Localising-the-client)

nanaya:
I think it should be its own filter instead of mixing with online/offline filter...?

YassineHaouzane:
> I think it should be its own filter instead of mixing with online/offline filter...?

What do you mean by its "own filter" ?
We should separate it in the UI ? In the code ? Both ?

Maybe something like that ?  

```ts
export type StatusFilter = 'all' | 'online' | 'offline';
export type RelationshipFilter = 'all' | 'mutual' | 'non-mutual';

interface State {
  statusFilter: StatusFilter;
  relationshipFilter: RelationshipFilter;
  playMode: PlayModeFilter;
  sortMode: SortMode;
  viewMode: ViewMode;
}
```

nanaya:
yes both

YassineHaouzane:
I added a commit which tries to split these 2 filters. Not sure about the UI though.

image

nanaya:
it should be more like the ruleset filter in [bn group page](https://osu.ppy.sh/groups/28)

YassineHaouzane:
Sorry, I didn't find time to work on this. Is this what you had in mind ?
image

PR #13264, opened 2026-08-31
Add option to add notes via Beatmap Discussion reviews

### LINKED ISSUE #13185: Add an option for a Note in modding reviews

notbakaneko:
I feel like the permissions for mapper notes being dependent on the current beatmap may make the insertion bar confusing...

closes #13185


resources/js/interfaces/beatmap-discussion-review.ts
```diff
@@ -22,7 +22,8 @@ export interface NewDocumentIssueEmbed extends DocumentBlock {
   type: 'embed';
 }
 
-export const beatmapReviewDiscussionTypes = ['praise', 'problem', 'suggestion'] as const;
+export const beatmapReviewDiscussionTypes = ['mapper_note', 'praise', 'problem', 'suggestion'] as const;
+export const beatmapReviewDiscussionTypesWithoutNote = ['praise', 'problem', 'suggestion'] as const;
```
nanaya: better typed as `BeatmapReviewDiscussionType[]`?

resources/css/bem/beatmap-discussion-editor-insertion-menu.less
```diff
@@ -3,6 +3,9 @@
 
 .beatmap-discussion-editor-insertion-menu {
   @_top: beatmap-discussion-editor-insertion-menu;
```
nanaya: the variable isn't used anymore

nanaya (changes requested):

nanaya (approved):

PR #13269, opened 2026-09-01
More fix Beatmap Discussions timestamp parsing

### LINKED ISSUE #13268: Beatmap discussion timestamps do not include selected objects
Issue started recently probably due to #13265 
Where the clickable timestamps no longer include the selected objects and treats them as normal string.

Several examples:
https://osu.ppy.sh/beatmapsets/2543018/discussion/5700476/timeline#/5748896 (Standard, should also be applicable to taiko and catch)
https://osu.ppy.sh/beatmapsets/2599213/discussion/5809730/timeline#/5748916 (Mania)

notbakaneko:
~~Limit by space or end of line (non-review posts don't use line breaks) instead since brackets are considered non-words and `12:34.567 (1)` is a valid timestamp.~~

~~`12:34.567 (1)` is `12:34.567 (1)`~~
~~`12:34.567a` is not parsed as a timestamp~~
~~`12:34.567 (1)a` is parsed as `12:34.567`~~

Just revert the regex change from #13265 and probably look at fixing `BeatmapDiscussionPost::parseTimestamp` behaviour to allow `12:34.567a` as a timestamp instead...

Resolves #13268.

### DIFF
--- a/resources/js/utils/beatmapset-discussion-helper.ts
+++ b/resources/js/utils/beatmapset-discussion-helper.ts
@@ -74,7 +74,7 @@ const generalPages = new Set<unknown>(['events', 'generalAll', 'reviews']);
 
 const defaultBeatmapId = '-';
 
-export const timestampRegex = /\b(((\d{2,}):([0-5]\d)[:.](\d{3}))(\s\((?:\d+[,|])*\d+\))?)\b/;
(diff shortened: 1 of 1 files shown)


nanaya:
...period and comma? (`00:00:000,`)

nanaya (approved):

PR #13270, opened 2026-09-02
Add test for parsing Beatmap Discussion timestamps

notbakaneko:
So there's something to refer to in the code of what is a valid timestamp (and not just the regex which is different for decorator and plain timestamp)


tests/Models/BeatmapDiscussionPostTest.php
```diff
@@ -13,10 +13,32 @@
...
 
 class BeatmapDiscussionPostTest extends TestCase
 {
+    public static function dataProviderForParseTimestamp()
+    {
+        return [
+            ['00:00.00', null],
+            ['00.00.000', null],
+            ['00:00.000', 0],
+            ['00:01.000', 1000],
```
nanaya: the int would be a bit easier to read with thousand separator like `_`? (also for the rest)

nanaya (approved):

PR #13272, opened 2026-09-02
Don't share regex with global flag

notbakaneko:
Regex with `/g` flag has `lastIndex` internal state we don't want to be sharing; it's just that we've always been getting all matches each time and avoiding problems 👀

### DIFF
--- a/resources/js/beatmap-discussions/editor.tsx
+++ b/resources/js/beatmap-discussions/editor.tsx
@@ -169,26 +169,16 @@ export default class Editor extends React.Component<Props, State> {
     this.xhr?.abort();
   }
 
-  decorateTimestamps = (entry: NodeEntry) => {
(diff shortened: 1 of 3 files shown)


resources/js/beatmap-discussions/editor.tsx
```diff
@@ -179,7 +179,8 @@ export default class Editor extends React.Component<Props, State> {
 
     let match;
 
-    while ((match = timestampRegexGlobal.exec(node.text)) !== null) {
+    const regex = new RegExp(timestampRegex, 'g');
+    while ((match = regex.exec(node.text)) !== null) {
```
nanaya: can't this be a `for` loop of `[...node.text.matchAll(...)]` instead?

notbakaneko: ..or a `map`...

nanaya (approved):

PR #13277, opened 2026-09-03
Remove unused oauth client details state interface

nanaya:
Must be from before converted to mobx but what's that `[key: string]: unknown` 🤨


notbakaneko:
I'm pretty sure `[key: string]: unknown` was for `setState({ [key]: value })`

notbakaneko (approved):

PR #13283, opened 2026-09-06
Add markdown content in news api

gagahpangeran:
Currently `content` in news API is html output instead of the original markdown content. I added `content_markdown` mainly for future use in lazer news overlay.

### DIFF
--- a/app/Http/Controllers/NewsController.php
+++ b/app/Http/Controllers/NewsController.php
@@ -14,6 +14,29 @@ use App\Transformers\NewsPostTransformer;
  */
 class NewsController extends Controller
 {
+    private static function newsPostContentIncludes(?array $formats): array
+    {
+        static $validFormats = [
+            'html' => 'content',
+            'markdown' => 'content_markdown',
+        ];
+
+        if (is_api_request()) {
+            $ret = [];
+            foreach ($formats ?? [] as $format) {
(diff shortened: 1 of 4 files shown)


nanaya:
one thing to note is news post occasionally contains a bunch of inlined html

[...or even inlined css style block for some reason](https://github.com/ppy/osu-wiki/blob/master/news/2026/2026-05-20-osutaiko-world-cup-2026-concludes.md?plain=1)

gagahpangeran:
> one thing to note is news post occasionally contains a bunch of inlined html
> 
> [...or even inlined css style block for some reason](https://github.com/ppy/osu-wiki/blob/master/news/2026/2026-05-20-osutaiko-world-cup-2026-concludes.md?plain=1)

Yes, I'm aware of this. I'll try to sort this out one by one before fully committing to implement the lazer-side (e.g. by proposing some custom container for those kind of html).

notbakaneko:
I don't think it should be returning both unless explicitly requested.
I think the Changelog has something where the format can be selected, so it should follow something similar.

gagahpangeran:
> I don't think it should be returning both unless explicitly requested. I think the Changelog has something where the format can be selected, so it should follow something similar.

Added this by basically just copy-paste it from changelog controller.

PR #13286, opened 2026-09-07
Fix conflict between passport and game client user relation

nanaya:
But then I noticed the `clients()` relation itself is being [deprecated](https://github.com/laravel/passport/blob/0febfd1b1caaa57a1c4d035da9528929e47aac1d/src/HasApiTokens.php#L24), so... 🤔


notbakaneko:
If we're renaming it this way, doesn't make more sense to rename to the non-deprecated name? 🤔 
Also should probably rename the `oauthClients` validation errors key to match

nanaya:
the non deprecated name uses the owner morph thing

notbakaneko:
oh right 😬

notbakaneko (approved):

PR #13287, opened 2026-09-07
Disable new profile page

nanaya:
Until ready for release I guess? 🤷‍♀️


notbakaneko:
wouldn't it be simpler to just `return null`...? (also what if someone already toggled 🤔 )

nanaya:
huh, there's nothing checking dead code section.

there shouldn't be anyone who have the v2 enabled in the first place.

notbakaneko (approved):

PR #13290, opened 2026-09-08
Enforce int user id for access token

nanaya:
😐


notbakaneko (approved):
:|

bdach:
😐

helps with my problem, thanks 🙏

PR #13293, opened 2026-09-08
Skip laravel attribute casting for user preferences model

nanaya:
I was testing how the `AsArrayObject` performs (spoiler: badly) and figured adding the `getAttribute` function speeds up things.

Setting `->options` directly will end up badly but it shouldn't be done that way 🤷


app/Models/UserProfileCustomization.php
```diff
@@ -67,10 +66,12 @@ class UserProfileCustomization extends Model
     public $incrementing = false;
 
     protected $casts = [
-        'options' => AsArrayObject::class,
+        'options' => 'array',
     ];
     protected $primaryKey = 'user_id';
 
+    private array $cachedOptions;
```
notbakaneko: should clear this on `refresh()` (or `resetMemoized()`)?

app/Models/UserProfileCustomization.php
```diff
@@ -312,7 +270,7 @@ public function setUserListViewAttribute($value)
 
     public function getExtrasOrderAttribute($value)
```
notbakaneko: this can be `private` now?

notbakaneko (changes requested):

notbakaneko (approved):

PR #13302, opened 2026-09-10
Aligning beatmap status text to be vertically centered

Aprixia:
I noticed that the beatmap status text was not properly aligned vertically, see https://ibb.co/7xQ9HFdf
so I decided to fix it, see https://ibb.co/Ccfvw8z

Tested with use of inspect element and does not break anything
Changes the padding of beatmapset-status--panel from `0 5px` to `0 5px 1px 5px` as that 1px bottom padding allows the text to be vertically centered


Aprixia:
forgot to mention, I found the issue to be that the panel had an odd number of pixels meanwhile the text has an even number of pixels, so the 1px bottom padding makes it so the panel also has an even number of pixels

sorry if this is the wrong way to go about it, I know very little of css, I just tested my way to a solution that worked

nanaya (approved):

Issue #13301, opened 2026-09-09
Several languages do not appear in the selector and their translations are thus (presumably) inaccessible


bdach:
Noticed in passing when checking a client change.

The following languages are visible on crowdin, have translated web strings there, have corresponding `lang` folders in this repository, yet do not seem to appear on the web language switcher:

- Spanish, Latin America ([`es-419`](https://github.com/ppy/osu-web/tree/5a41b4486b4483c6f0045595a425c92200006eb1/resources/lang/es-419))
- Persian ([`fa-IR`](https://github.com/ppy/osu-web/tree/5a41b4486b4483c6f0045595a425c92200006eb1/resources/lang/fa-IR))
- Croatian ([`hr-HR`](https://github.com/ppy/osu-web/tree/5a41b4486b4483c6f0045595a425c92200006eb1/resources/lang/hr-HR))
- Kazakh ([`kk-KZ`](https://github.com/ppy/osu-web/tree/5a41b4486b4483c6f0045595a425c92200006eb1/resources/lang/kk-KZ))
- Latvian ([`lv-LV`](https://github.com/ppy/osu-web/tree/5a41b4486b4483c6f0045595a425c92200006eb1/resources/lang/lv-LV))
- Malay ([`ms-MY`](https://github.com/ppy/osu-web/tree/5a41b4486b4483c6f0045595a425c92200006eb1/resources/lang/ms-MY))

Aside from the above, there's also Sinhala ([`si-LK`](https://github.com/ppy/osu-web/tree/5a41b4486b4483c6f0045595a425c92200006eb1/resources/lang/si-LK)) which currently has no translations per crowdin, and also Tajik ([`tg-TJ`](https://github.com/ppy/osu-web/tree/5a41b4486b4483c6f0045595a425c92200006eb1/resources/lang/tg-TJ)) which has very little.

I'm *guessing* that laravel isn't liking the above ISO codes for the languages and that this is going to require nudging the [mapping](https://github.com/ppy/osu-web/blob/5a41b4486b4483c6f0045595a425c92200006eb1/crowdin.yml#L6) in `crowdin.yml` to fix, but what would be the correct incantation to apply is outside of my understanding. Likely this will also require nuking the old folders once the right monikers *are* found.

bdach:
cc @peppy given https://github.com/ppy/osu-web/pull/13300 is in progress (so it may be a convenient time to look at this)

nanaya:
[...]

PR #13303, opened 2026-09-10
Enable Malay locale

nanaya:
One of the locales with relatively complete translation as reported in #13301.

I think calling it `ms` makes more sense?


crowdin.yml
```diff
@@ -24,6 +24,7 @@ files:
         ja: ja
         ko: ko
         lt: lt
+        ms: ms
```
nanaya, line 27: this is rather confusing because we're using `%locale%` (`ms-MY` is the default) instead of whatever crowdin uses for their url (the left part, which is already `ms` - which I think is `%two_letters_code%`?)

notbakaneko (approved):

crowdin.yml
peppy: I'm not sure if this affects anything, but the client is using `ms-MY`:

https://github.com/ppy/osu/blob/30fd40efd16a651a6c00b5c89289a85ffcbe546b/osu.Game/Localisation/Language.cs#L87-L88

PR #13305, opened 2026-09-10
Select less fields for Multiplayer room recent participants list

notbakaneko:
Should use less memory by not filling all the model fields when only a new of them are needed for the transformer.
The nested include means a lot of users are potentially loaded, including all the large text fields we don't need for the transformer include...

(could use some suggestions on where to put it as a `const` 👀 )


nanaya:
maybe the const can be in the `UserCompactTransformer`? That's where the values are used at least.

nanaya (approved):

PR #13308, opened 2026-09-11
Preload beatmap top tag ids

nanaya:
Noticed this when looking at the trace for beatmapset show endpoint.

### DIFF
--- a/app/Models/Beatmap.php
+++ b/app/Models/Beatmap.php
@@ -75,6 +75,8 @@ class Beatmap extends Model implements AfterCommit
         'mania' => 3,
     ];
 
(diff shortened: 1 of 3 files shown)


app/Models/Beatmap.php
```diff
@@ -380,6 +382,11 @@ public function maxCombo()
         return $maxCombo?->value;
     }
 
+    public function preloadTopTagIds(array $data): void
+    {
+        $this->memoized['topTagIds'] = $data;
```
nanaya, line 387: this is rather questionable but I can't find better way to do it short of adding a helper function like `setMemoizeValue` to just pass a value instead of requiring a callback.

notbakaneko (approved):
There are extra `get`s if something is missing from the `mget` but I suppose it is an unusual enough case that is should be fine...

PR #13310, opened 2026-09-12
Add redis caching of expensive ES count lookups

peppy:


app/Http/Controllers/BeatmapsController.php
```diff
@@ -90,7 +90,35 @@ private static function beatmapScores(string $id, ?bool $legacyFormat, ?bool $is
...
+                case 'global':
+                    $cacheKey = 'lb_count_global:'.$beatmap->getKey().'-'.$beatmap->approved.'-'.$legacyMode.'-'.$rulesetId.'-'.$sortedMods;
+                    break;
+                case 'country':
+                    $cacheKey = 'lb_count_country:'.$currentUser->country_acronym.'-'.$beatmap->getKey().'-'.$beatmap->approved.'-'.$legacyMode.'-'.$rulesetId.'-'.$sortedMods;
+                    break;
+            }
+
+            if ($cacheKey !== null) {
+                $count = \Cache::get($cacheKey);
```
nanaya: this returns string on hit and needs to be casted to int with `get_int`

nanaya (changes requested):

app/Http/Controllers/BeatmapsController.php
```diff
@@ -90,7 +90,35 @@ private static function beatmapScores(string $id, ?bool $legacyFormat, ?bool $is
...
+
+                if ($count === null) {
+                    $count = UserRank::getCount($countParams);
+
+                    // use count as TTL, matches `global-rank-lookup-cache`
+                    \Cache::put($cacheKey, $count, $count);
```
nanaya: I wonder if there should be a minimum ttl (for example when there's no hit for specific mods combination on leaderboard with lots of other scores)

nanaya: or maybe even skip querying for example if `count($results['scores']) < $countParams->size`

peppy: Valid point. The previous iteration of this caching never had a cache of "low results count but expensive query" but we do now due to filters.

Maybe 60 or 600 seconds will work well enough.

PR #13315, opened 2026-09-13
Count cache cleanup

nanaya:
Skip querying in some cases and combine assigning the same key suffix.

Also use fetch param country code instead of current user (technically should always be the same for this section).

### DIFF
--- a/app/Http/Controllers/BeatmapsController.php
+++ b/app/Http/Controllers/BeatmapsController.php
@@ -92,21 +92,17 @@ class BeatmapsController extends Controller
         $totalsEnabled = $GLOBALS['cfg']['osu']['scores']['leaderboard_totals_enabled'];
 
         if ($isApi && $totalsEnabled) {
-            $cacheKey = null;
-
-            $sortedMods = implode(',', array_sort($mods));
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/BeatmapsController.php
```diff
@@ -92,34 +92,36 @@ private static function beatmapScores(string $id, ?bool $legacyFormat, ?bool $is
...
+                $cacheKey = match ($type) {
+                    'country' => "lb_count_country:{$esFetch->baseParams->getCountryCode()}",
+                    'global' => 'lb_count_global:',
+                    default => null,
+                };
+
+                if ($cacheKey !== null) {
+                    $sortedMods = implode(',', array_sort($mods));
+                    $legacyMode = $isLegacy ? '1' : '0';
+                    $cacheKey .= "-{$beatmap->getKey()}-{$beatmap->approved}-{$legacyMode}-{$rulesetId}-{$sortedMods}";
```
notbakaneko: should the leading `-` be at the end of the country key prefix? Otherwise this always creates `lb_count_global:-`

notbakaneko (approved):

PR #13318, opened 2026-09-14
Hide restricted user post-query on ranked play leaderboard

### LINKED ISSUE #13262: Rankings: Possible performance related concern with Ranked Play leaderboard
When browsing to https://osu.ppy.sh/rankings/ranked-play/osu and the pages themselves (there's over 2,000 of them), it's much much slower than the usual ranking pages. I'm unsure if this was noticed already or not, I cannot find a duplicate issue.

nanaya:
Sort of resolves #13262

It seems currently fine up to 200-ish ranks and then there are a few restricted users left on the board which then hidden and sort of broke the rank number.

Further speed up would just be limiting the page or forcing cursor based pagination after certain page.


notbakaneko (commented):
is the rank counter supposed to jump?

nanaya:
yeah that's one of the side effects. At least it'd be closer to what's shown on user profile page.

notbakaneko (approved):

PR #13336, opened 2026-09-18
Update trait usage syntax to match psr-12

nanaya:
It's been rather inconsistent already. Sorting them will be done later... maybe.


notbakaneko (approved):
let's just get this out of the way now

